PR #746, opened 2020-03-25
Model name should be included in the inference response for GRPC

deadeyegoodwin:
…REST

Included by mistake from an earlier revision of the proposal.
The GRPC API is alredy correct in this regard.

Also update comment on GRPC requested output tensors to align with
HTTP.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/predict-api/v2/required_api.md
+++ b/docs/predict-api/v2/required_api.md
@@ -653,7 +653,8 @@ failure. The request and response messages for ModelInfer are:
       // The input tensors for the inference.
       repeated InferInputTensor inputs = 5;
 
-      // The requested output tensors for the inference.
+      // The requested output tensors for the inference. Optional, if not
+      // specified all outputs produced by the model will be returned.
       repeated InferRequestedOutputTensor outputs = 6;
     }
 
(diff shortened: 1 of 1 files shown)


docs/predict-api/v2/required_api.md
```diff
@@ -281,14 +281,12 @@ code. The inference response object, identified as
 
     $inference_response =
     {
-      "model_name" : $string,
```
ellis-bigelow: So the spec states to return model_version but not model_name? It may be a nice convenience to have the model_name. WDYT about updating the spec?

deadeyegoodwin: I think it would be good to have the model name, and that is what I originally proposed. But at some point along the way there was the suggestion that because it wasn't absolutely necessary it should be removed. This just fixes the doc issue that I didn't remove it consistently for both HTTP and GRPC. I'm happy to add it back in both places.

ellis-bigelow: Let's keep this in the spec please. Not absolutely necessary, but does mitigate potential pain.

deadeyegoodwin: Updated.

ellis-bigelow:
Nice!
/lgtm
/approve

PR #747, opened 2020-03-25
Triton naming

deadeyegoodwin:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/samples/triton/simple_string/README.md
+++ b/docs/samples/triton/simple_string/README.md
@@ -1,5 +1,5 @@
 
-# Predict on a InferenceService using TensorRT Inference Server
+# Predict on a InferenceService using Triton Inference Server
 ## Setup
 1. Your ~/.kube/config should point to a cluster with [KFServing installed](https://github.com/kubeflow/kfserving/blob/master/docs/DEVELOPER_GUIDE.md#deploy-kfserving).
 2. Your cluster's Istio Ingress gateway must be network accessible.
@@ -7,22 +7,22 @@
 ## Create the InferenceService
 Apply the CRD
 ```
-kubectl apply -f tensorrt.yaml 
+kubectl apply -f triton.yaml 
 ```
 
 Expected Output
 ```
-inferenceservice.serving.kubeflow.org/tensorrt-simple-string created
+inferenceservice.serving.kubeflow.org/triton-simple-string created
 ```
 
 ## Run a prediction
-Uses the client at: https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-guide/docs/client.html#section-client-api
+Uses the client at: https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/client_example.html
 
 
 1. setup vars
 
 ```
-SERVICE_HOSTNAME=$(kubectl get ksvc tensorrt-simple-string-predictor-default -o jsonpath='{.status.url}' | cut -d "/" -f 3)
+SERVICE_HOSTNAME=$(kubectl get ksvc triton-simple-string-predictor-default -o jsonpath='{.status.url}' | cut -d "/" -f 3)
 INGRESS_GATEWAY=istio-ingressgateway
 CLUSTER_IP=$(kubectl -n istio-system get service $INGRESS_GATEWAY -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
 echo $CLUSTER_IP
@@ -31,7 +31,7 @@ echo $CLUSTER_IP
 ```
 curl -H "Host: ${SERVICE_HOSTNAME}" http://${CLUSTER_IP}/api/status
 ```
-3. edit /etc/hosts to map the CLUSTER IP to tensorrt-simple-string-predictor-default.default.example.com
+3. edit /etc/hosts to map the CLUSTER IP to triton-simple-string-predictor-default.default.example.com
 4. run the client
 ```
 docker run -e SERVICE_HOSTNAME:$SERVICE_HOSTNAME -it --rm --net=host kcorer/tensorrtserver_client:19.05
@@ -40,7 +40,7 @@ docker run -e SERVICE_HOSTNAME:$SERVICE_HOSTNAME -it --rm --net=host kcorer/tens
 
 You should see output like:
 ```
-root@trantor:/workspace# ./build/simple_string_client -u tensorrt-simple-string.default.example.com
+root@trantor:/workspace# ./build/simple_string_client -u triton-simple-string.default.example.com
 0 + 1 = 1
 0 - 1 = -1
 1 + 1 = 2
--- a/docs/samples/triton/simple_string/triton.yaml
+++ b/docs/samples/triton/simple_string/triton.yaml
@@ -1,9 +1,9 @@
 apiVersion: "serving.kubeflow.org/v1alpha2"
 kind: "InferenceService"
 metadata:
-  name: "tensorrt-simple-string"
+  name: "triton-simple-string"
 spec:
   default:
     predictor:
-      tensorrt:
-        storageUri: "gs://kfserving-samples/models/tensorrt"
+      triton:
+        storageUri: "gs://kfserving-samples/models/triton"
--- a/docs/samples/triton/simple_string/triton.yaml
+++ b/docs/samples/triton/simple_string/triton.yaml
@@ -6,4 +6,4 @@ spec:
   default:
     predictor:
       triton:
-        storageUri: "gs://kfserving-samples/models/triton"
+        storageUri: "gs://kfserving-samples/models/tensorrt"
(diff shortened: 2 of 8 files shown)


deadeyegoodwin:
@ellis-bigelow @rakelkar 
I'm not sure I got all the references or fixed them correctly. There are other mentions of TensorRT but some seem related to integration with TF and those should stay as TensorRT. It seems like TensorRTSpec should be renamed but not sure.

deadeyegoodwin:
There is also a related PR for KubeFlow website: https://github.com/kubeflow/website/pull/1846

docs/samples/triton/README.md
```diff
@@ -30,16 +30,16 @@ echo $CLUSTER_IP
 ```
 curl -H "Host: ${SERVICE_HOSTNAME}" http://${CLUSTER_IP}/api/status
 ```
-3. edit /etc/hosts to map the CLUSTER IP to tensorrt-simple-string.default.example.com
+3. edit /etc/hosts to map the CLUSTER IP to triton-simple-string.default.example.com
 4. run the client
 ```
-docker run -e SERVICE_HOSTNAME:$SERVICE_HOSTNAME -it --rm --net=host kcorer/tensorrtserver_client:19.05
+docker run -e SERVICE_HOSTNAME:$SERVICE_HOSTNAME -it --rm --net=host nvcr.io/nvidia/tritonserver:20.03-py3
```
yuzisun:
```suggestion
docker run -e SERVICE_HOSTNAME:$SERVICE_HOSTNAME -it --rm --net=host nvcr.io/nvidia/tritonserver:20.03-py3-clientsdk
```

docs/samples/triton/README.md
```diff
@@ -30,16 +30,16 @@ echo $CLUSTER_IP
...
 ./build/simple_string_client -u $SERVICE_HOSTNAME
```
yuzisun:
```suggestion
./workspace/install/bin/simple_string_client -u $SERVICE_HOSTNAME
```

ellis-bigelow:
Yeah we're definitely missing changes for https://github.com/kubeflow/kfserving/blob/master/pkg/apis/serving/v1alpha2/framework_tensorrt.go
I'd run `make test` after making the changes.

yuzisun:
> Yeah we're definitely missing changes for https://github.com/kubeflow/kfserving/blob/master/pkg/apis/serving/v1alpha2/framework_tensorrt.go
> I'd run `make test` after making the changes.

Good point, gonna add a e2e test in my other triton example PR.

deadeyegoodwin:
So do I need to make changes to framework_tensorrt.go, or @yuzisun are you going to handle that in your followup PR?

yuzisun:
> So do I need to make changes to framework_tensorrt.go, or @yuzisun are you going to handle that in your followup PR?

@deadeyegoodwin I can take care of that in my PR, once I am done we can merge together.

rakelkar:
Not sure this change makes sense to me? In my understanding KFServing "Frameworks" are ML frameworks - which inference server we use to inference them should be an implementation detail. 

Triton is an inference server, and we should certainly update the TensorRT framework to use it (and possibly some other or all other frameworks to use it as well e.g. #744 ) but I don't think Triton is an ML framework.

So it seems logical to me to include TensorRT as a framework rather (assuming NVIDIA cares enough to support it) and consider updating the TensorRT sample to use an actual TRT model.plan instead of TF as it currently does.

@ellis-bigelow @cliveseldon @animeshsingh  @yuzisun 

/hold

ellis-bigelow:
@yuzisun , I'd lean away from specifying inferenceserver in framework. This multilayered model seems more complex than it's worth.

rakelkar:
@yuzisun when we say ONNX - it has to be a model serialized in ONNX format.. it doesnt matter how you got to that point... and from user perspective it doesnt matter whether you use the C++ ONNXServer for inference or your own Python Server that hosts the ONNX runtime... that is a serving infra decision...

So IMO framework maps directly to the serialized model format...

rakelkar:
> @yuzisun , I'd lean away from specifying inferenceserver in framework. This multilayered model seems more complex than it's worth.



> I agree with Rakesh's point. QQ though, is the TensorRT name unchanged re: the framework?

I believe they only changed the name of the inference server (which makes sense since it supports multiple model formats)...

yuzisun:
> > @yuzisun , I'd lean away from specifying inferenceserver in framework. This multilayered model seems more complex than it's worth.
> 
> > I agree with Rakesh's point. QQ though, is the TensorRT name unchanged re: the framework?
> 
> I believe they only changed the name of the inference server (which makes sense since it supports multiple model formats)...

So when user want to serve a tensorflow model with triton inference server, how do we specify that on our spec?

rakelkar:
> like user first specify ML framework and then choose which inference server to serve the model

btw you can achieve this today by changing the configmap to point to a different inference server for the framework - but it is platform-wide opinion and not per-model... I guess our stance is the platform chooses for the 90% ... and edge cases are covered via custom? This is what cloud providers do too btw (e.g. sagemaker)

wdyt?

deadeyegoodwin:
TensorRT "inference accelerator" is still called "TensorRT": https://developer.nvidia.com/tensorrt

yuzisun:
> > like user first specify ML framework and then choose which inference server to serve the model
> 
> btw you can achieve this today by changing the configmap to point to a different inference server for the framework - but it is platform-wide opinion and not per-model... I guess our stance is the platform chooses for the 90% ... and edge cases are covered via custom? This is what cloud providers do too btw (e.g. sagemaker)
> 
> wdyt?

I would agree with you, but we are not there yet I think, also currently it is not simply a change in configmap with the image for a given ML framework, the startup command and parameters are pretty hard coded for a given inference server in the controller. Even that could change in the future I would still think giving user the ability to choose which inference server to use per model is needed, like for pytorch what if the model can not be converted to onnx and user just want to serve on a simple KFServing python server without going custom route.

rakelkar:
> model can not be converted to onnx and user just want to serve on a simple KFServing python server without going custom route

But for that we already support pytorch as a framework and have a PYT server (that loads native PYT serialized models and does not require ONNX)?

yuzisun:
> > model can not be converted to onnx and user just want to serve on a simple KFServing python server without going custom route
> 
> But for that we already support pytorch as a framework and have a PYT server (that loads native PYT serialized models and does not require ONNX)?

ya just making an example, or a more relevant question regarding to triton inference server is that since it is a multi-framework inference server how can we allow user to serve tensorflow/pytorch model on triton inference server by leveraging their auto-batching, multi-model co-hosting features without having to convert to the accelerated model format TensorRT. Currently the only way to use triton inference server it to put under `tensorrt` spec but it is confusing and the model actually may not be in tensorRT format, what do you think we can proceed from here?

ellis-bigelow:
I think model server might be a more accurate descriptor. If it's capable of serving many different model formats, triton definitely seems more clear than tensorrt to me.

rakelkar:
> Can we simply treat triton as a slight exception to the framework rule for now and we can resolve this more holistically once we have a clearer understanding of where the industry is moving with model servers

Sounds fair! 

/unhold

yuzisun:
oh needs to move the test model from `gs://kfserving-samples/models/tensorrt` to `gs://kfserving-samples/models/triton`

ellistarn:
Are we planning another 0.3.x release before the v1beta1 API changes? Much of this renaming will happen there.

edenbuaa:
> 
> 
> Are we planning another 0.3.x release before the v1beta1 API changes? Much of this renaming will happen there.

Also want to know.

docs/samples/triton/simple_string/triton.yaml
```diff
@@ -1,9 +1,9 @@
...
 metadata:
-  name: "tensorrt-simple-string"
+  name: "triton-simple-string"
 spec:
   default:
     predictor:
-      tensorrt:
-        storageUri: "gs://kfserving-samples/models/tensorrt"
+      triton:
+        storageUri: "gs://kfserving-samples/models/triton"
```
yuzisun, line 9:
```suggestion
        storageUri: "gs://kfserving-samples/models/tensorrt"
```

yuzisun: let's keep this bucket for now, I somehow lost the permission to move the bucket

docs/samples/triton/simple_string/README.md
```diff
@@ -1,28 +1,28 @@
...
+# Predict on a InferenceService using Triton Inference Server
 ## Setup
 1. Your ~/.kube/config should point to a cluster with [KFServing installed](https://github.com/kubeflow/kfserving/blob/master/docs/DEVELOPER_GUIDE.md#deploy-kfserving).
 2. Your cluster's Istio Ingress gateway must be network accessible.
 
 ## Create the InferenceService
 Apply the CRD
 ```
-kubectl apply -f tensorrt.yaml 
+kubectl apply -f triton.yaml
```
yuzisun, line 10: need to rename the file under `simple_string` as well

edenbuaa:
@yuzisun Could the kfserving v0.3.0 support triton server now?

PR #750, opened 2020-03-26
Triton Renaming and Add bert example for triton inference server

### LINKED ISSUE #645: Upgrate TensorRT to latest valid version
The version of TensorRT in KFServing is 19.05 - We shall upgrade it to the latest version 19.12 (Dec 2019 release)

yuzisun:

Fixes #645 

**Release note**:

```release-note
`tensorrt` is renamed to `triton` on framework choice
```


docs/samples/triton/bert/bert.yaml
```diff
@@ -0,0 +1,35 @@
...
+              memory: 1Gi
+            requests:
+              cpu: "1" 
+              memory: 1Gi
+          command:
+            - "python"
+            - "-m"
+            - "bert_transformer"
+          env:
+            - name: STORAGE_URI
```
salanki, line 24: Unrelated question but is there a good reason this isn't just part of the spec like for most predictors?

ellistarn: I would love to promote some transformers to first class in the spec. Take a look at the feast transformer PR I wrote 6 months ago.

salanki: I was referring to storageUri in the spec vs STORAGE_URI. I read your comment as bundling common transformers? I like that idea too, but doesn’t necessarily have to be part of the API spec.

ellistarn: Oh yeah oops. Definitely misunderstood your question :). 

Transformers don't have a concept of storage uri (it's just a pure customspec), but maybe it makes sense to in the v1alpha3 as a way of interacting with the storage-initializer.

salanki: I’d suggest that for the customSpec. We do a customSpecs and always mount in storage as well to load the models. Building different docker images every time we want to update a model seems like a huge waste.

yuzisun:
@ellistarn @deadeyegoodwin finally passed all the tests with triton inference server, it is good to go now.

deadeyegoodwin:
I took a look at the README and it looks good to me.

PR #777, opened 2020-04-13
fix-775 fix sdk promote missing annotations

### LINKED ISSUE #775: SDK promote method miss annotations
https://github.com/kubeflow/kfserving/blob/8c261457b3ec8017736b882c4ffd3379914471ac/python/kfserving/kfserving/api/kf_serving_client.py#L295-L297

Iamlovingit:
The SDK `promote` method is missing current isvc's annotation
Fixes #775

**Release note**:

```release-note

```

### DIFF
--- a/python/kfserving/kfserving/api/kf_serving_client.py
+++ b/python/kfserving/kfserving/api/kf_serving_client.py
@@ -289,12 +289,18 @@ class KFServingClient(object):
         except KeyError:
             raise RuntimeError("Cannot promote a InferenceService that has no Canary Spec.")
 
+        try:
+            annotations = current_isvc['metadata']['annotations']
+        except KeyError:
+            annotations = dict()
+
         inferenceservice = V1alpha2InferenceService(
             api_version=api_version,
             kind=constants.KFSERVING_KIND,
             metadata=client.V1ObjectMeta(
                 name=name,
-                namespace=namespace),
+                namespace=namespace,
+                annotations=annotations),
             spec=V1alpha2InferenceServiceSpec(
                 default=current_canary_spec,
                 canary=None,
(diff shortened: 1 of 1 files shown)


pugangxa:
From InferenceService point of view, I assume default and canary should have the same annotations right, But can you show the difference and the real issue you find?

yuzisun:
@Iamlovingit Can you help put more description of the issue you found ?

Iamlovingit:
I add my code and output in issue. @pugangxa  @yuzisun 
https://github.com/kubeflow/kfserving/issues/775

Iamlovingit:
use metadata directly. @pugangxa

Iamlovingit:
metadata type of current_isvc is dict(), not match V1ObjectMeta. I can not find a perfact way to init the V1ObjcectMeta, So, only add annotations this time.

pugangxa:
> metadata type of current_isvc is dict(), not match V1ObjectMeta. I can not find a perfact way to init the V1ObjcectMeta, So, only add annotations this time.

Something like  ```obj.__dict__.update``` should work but we need check V1ObjectMeta properties meanings to make sure it can be done like this. In replace function we also copied "resource_version" property. 
Anyway annotation should be the one need change frequently and we can let this go firstly.

python/kfserving/kfserving/api/kf_serving_client.py
```diff
@@ -289,12 +289,18 @@ def promote(self, name, namespace=None, watch=False, timeout_seconds=600): # pyl
         except KeyError:
             raise RuntimeError("Cannot promote a InferenceService that has no Canary Spec.")
 
+        try:
+            annotations = current_isvc['metadata']['annotations']
+        except KeyError:
+            annotations = dict()
```
yuzisun, line 295: @jinchihe do you still recall why for promote we need to do replace instead of patch here?

jinchihe: @yuzisun we must call `replace`, due to we set `canary=None` here, if we use `patch`, the canary will not be removed, and keep old one. To resolve this issue, we implemented `replace` API ...

jinchihe:
/lgtm

Thanks for the enhancement! @Iamlovingit

PR #788, opened 2020-04-17
Use public slack link

rzgry:
Slack link in CONTRIBUTING.md only allows people with a `@google.com`, `@caicloud.io`, or `@microsoft.com` email address to sign up.

image

Update the link with link from kubeflow.org which allows anyone to sign up.


yuzisun:
I did not know there is a shared invite, I know people has been complaining about this. Thanks for fixing it !

/lgtm
/approve

PR #795, opened 2020-04-26
Very hackish dependency upgrade to K8s 1.18

salanki:
This PR makes the webhook work on Kubernetes 1.18.

If someone stumbles on this ticket in a panic after upgrading their cluster to 1.18, you can simply update your statefulset to use this image: [coreweave/kfserving:v0.3-k8s18-0](https://hub.docker.com/repository/docker/coreweave/kfserving). I hope this can save someone a couple of hours of panic.

Without this, the Pod mutator will set no longer existing field types. If you describe the replicaset, it shows:
```
metadata.managedFields.fieldsType: Invalid value: "": must be `FieldsV1` kfserving
```

I only updated the Kubernetes client libs to 1.16, as that the new value is introduced there and I wanted to minimum version jump possible.

The code generators **do not work** as @yuzisun has already discovered: https://github.com/kubernetes-sigs/controller-tools/issues/391. I simply disable these to build the image.

I also manually hacked `vendor/sigs.k8s.io/controller-runtime/pkg/client/apiutil/apimachinery.go`, but this should not be necessary when upgrading to even newer Kubernetes libs.

I do not expect this to be merged, however it might be helpful in some way to do a proper implementation. I regretfully have very limited Go dep experience (this is my second time), and I do not believe I have what the knowledge to bring this fully home.


yuzisun:
@salanki I am migrating KFServing go module https://github.com/kubeflow/kfserving/pulls here, can you give more detail what exactly the issue you are facing with Kubernetes 1.16? Pod mutator is setting a field which is removed in 1.16?

salanki:
1.16 and 1.17 is fine. The issue is in 1.18+. The pod mutator is setting a deprecated field, it seems to be fairly well known that you can’t run 1.15 kubernetes client lib against 1.18 if you google the error I show from the replicaset.

salanki:
This is where I found out that I need at least client-go 1.16: https://github.com/kubedb/issues/issues/739

yuzisun:
@salanki I am getting client-1.16 in this https://github.com/kubeflow/kfserving/pull/796, so are you ok to close this one ?

salanki:
Yup. Closing.

sheep94lion:
Thank you. It really helps!

PR #796, opened 2020-04-26
Migrate KFServing to Go Module 

yuzisun:

Fixes https://github.com/kubeflow/kfserving/issues/404

1. Migrate dep to go module
2. Removed resources validation because the go module does not allow depending on `k8s.io/kubernetes` for some reason, I think this is ok the validation just be done at pod level instead of CRD level.
3. Fix unit test and the CI unit test run issue

**Release note**:
Migrate from Go dep to Go Module
```release-note

```


ellistarn:
Nice -- bit of a difficult review. Looks like some sweeping changes that had some cascading effects. I think I agree with the majority of them, but it's a bit difficult to review.

Makefile
```diff
@@ -1,5 +1,9 @@
 HAS_LINT := $(shell command -v golint;)
-
+ifeq (,$(shell go env GOBIN))
```
ellistarn, line 2: GOBIN=${GOBIN:$GOPATH/bin}

Makefile
```diff
@@ -167,6 +172,21 @@ docker-build-storageInitializer:
 docker-push-storageInitializer: docker-build-storageInitializer
 	docker push ${KO_DOCKER_REPO}/${STORAGE_INIT_IMG}
 
+controller-gen:
+ifeq (, $(shell which controller-gen))
```
ellistarn, line 176: Isn't this covered by users installing dependencies?

config/default/crds/patches/crd_status_condition_patch.yaml
```diff
@@ -1,3 +1,18 @@
+- op: replace
```
ellistarn, line 1: should the name of this file change now that it's both patches?

config/default/webhook/manifests.yaml
```diff
@@ -31,10 +31,10 @@ webhooks:
         path: /mutate-pods
     failurePolicy: Fail
     name: inferenceservice.kfserving-webhook-server.pod-mutator
-    namespaceSelector:
+    objectSelector:
```
ellistarn, line 34: This is a separate feature! Do we want to do this as part of this change??

pkg/apis/serving/v1alpha2/inferenceservice_status.go
```diff
@@ -78,7 +78,7 @@ func (ss *InferenceServiceStatus) GetCondition(t apis.ConditionType) *apis.Condi
 // PropagateDefaultStatus propagates the status for the default spec
 func (ss *InferenceServiceStatus) PropagateDefaultStatus(component constants.InferenceServiceComponent, defaultStatus *knservingv1.ServiceStatus) {
 	if ss.Default == nil {
-		emptyStatusMap := make(ComponentStatusMap)
+		emptyStatusMap := make(map[constants.InferenceServiceComponent]StatusConfigurationSpec)
```
ellistarn, line 81: I had to do this as well for the type generation. +1.

pkg/apis/serving/v1alpha2/inferenceservice_types.go
```diff
@@ -247,7 +247,6 @@ type StatusConfigurationSpec struct {
 	Name string `json:"name,omitempty"`
 	// Host name of the service
 	Hostname string `json:"host,omitempty"`
-	Replicas int    `json:"replicas,omitempty"`
```
ellistarn, line 250: Why is this gone?

yuzisun: this is the field which is not used anywhere, so let's remove it?

config/default/webhook/manifests.yaml
yuzisun: sure, can raise a separate PR

config/default/crds/patches/crd_status_condition_patch.yaml
yuzisun: good point

Makefile
yuzisun: this isn't part of the dependency in `go.mod` and used anywhere in controller go code, it is only used when running `make manifests`

pkg/apis/serving/v1alpha2/inferenceservice_types.go
ellistarn: IIUC we signal current replicas using this field. It may be a bug if it's not updated.

config/default/webhook/manifests.yaml
yuzisun: reverted back

config/default/crds/patches/crd_status_condition_patch.yaml
yuzisun: renamed

Makefile
yuzisun: cleaned up a bit with unnecessary code

pkg/apis/serving/v1alpha2/inferenceservice_types.go
yuzisun: no I checked, we are not doing anything with it, so if we really need it we can add back with a proper implementation.

PR #797, opened 2020-04-27
Add batcher module.

zhangrongguo:
Add batcher module to do auto batching.


zhangrongguo:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fkubeflow%2Fkfserving%2Fpull%2F797) for more info**.

@googlebot I signed it!

cliveseldon:
This is an interesting first step @zhangrongguo 
Can you add some documentation of the design somewhere to help with the review?

zhangrongguo:
> This is an interesting first step @zhangrongguo
> Can you add some documentation of the design somewhere to help with the review?

Ok, I will write a doc.

zhangrongguo:
> > This is an interesting first step @zhangrongguo
> > Can you add some documentation of the design somewhere to help with the review?
> 
> Ok, I will write a doc.

Hi，I have a new commit of this PR. In this commit, the left code and some docs are added.

salanki:
This looks really interesting.

1) What is the point of configuring batcher host and port? Shouldn’t that be done automatically, just like the transformer? There is already a pattern for how “middleware” is Deployer with how the transformer is implemented, is there a reason not to follow the same?

2) Do you have any idea of how this performs in comparison to say TFServings native batcher?

zhangrongguo:
> This looks really interesting.
> 
> 1. If I read the code correctly, the batcher is implemented as a sidecar on the predictor Pod. This is a bad choice, as it will lead to suboptimal batching when there is more than 1 replica as it sits after the load balancer. A better implementation would be to treat the batcher just like the transformer. It will be just another step in the pipeline that does batching and will receive all requests to it, before the load balancing happens. I would put it after the transformer.
> 2. Once it is wired up as a pipeline, the host and port configuration can be removed. It should be fully automatic, just like the transformer.
> 3. I am curious of how this performs in comparison to say TFServings native batcher? Given that batching can happen on a global level, instead of per Pod level, it should be able to be much more effective.

1. If we put the batcher outof the predictor pod, the performance will be decreased due to the large data transfer between pods.
2. The port configuration is for possible future extensions.
3. We haven't tested this with TFServings. However, we have tested this with PyTorch framework, and the the performance is much better than the initial PyTorch Server.

  | PyTorch   Server | PyTorch   Server   With   Batching | PyTorch   Server   With   Batcher Sidecar
-- | -- | -- | --
5   client threads | 5   client threads | 5   client threads
Average(seconds) | 0.7559(s) | 0.1367(s) | 0.062(s)

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+	result, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		errStr = fmt.Sprintf("Response read fail: %v", err)
+		panic(errStr)
+	}
+	err = json.Unmarshal(result, &batcherInfo.Predictions)
+	if err != nil {
+		errStr = fmt.Sprintf("Response unmarshal fail: %v, %s", err, string(result))
+		panic(errStr)
+	}
```
yuzisun, line 157: why do we need to unmarshal and then marshal again?

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+	defer resp.Body.Close()
+	result, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		errStr = fmt.Sprintf("Response read fail: %v", err)
+		panic(errStr)
+	}
+	err = json.Unmarshal(result, &batcherInfo.Predictions)
+	if err != nil {
+		errStr = fmt.Sprintf("Response unmarshal fail: %v, %s", err, string(result))
+		panic(errStr)
```
yuzisun, line 156: shouldn't be panic here

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+
+type ResponseError struct {
+	Message string `json:"message"`
+}
+
+type Predictions struct {
+	Predictions []interface{} `json:"predictions"`
+}
+
+type BatcherInfo struct {
```
yuzisun, line 79: `BatcherInfo` is internal state, does it need to be json serializable ?

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+	if uuidStr, err := uuid.NewV4(); err == nil {
+		return uuidStr.String()
+	}
+	return ""
+}
+
+func CallService() {
+	var errStr string
+	url := fmt.Sprintf("http://%s:%s%s", batcherHandler.SvcHost, batcherHandler.SvcPort, batcherHandler.Path)
+	jsonStr, _ := json.Marshal(Request{
```
yuzisun, line 130: we may need to check header here if the original request is sent in as json

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+		errStr = fmt.Sprintf("NewRequest create fail: %v", err)
+		panic(errStr)
+	}
+	req.Header.Add("Content-Type", batcherHandler.ContentTpye)
+	defer req.Body.Close()
+	client := &http.Client{Timeout: 5 * time.Second}
+	resp, err := client.Do(req)
+	if err != nil {
+		errStr = fmt.Sprintf("NewRequest send fail: %v", err)
+		panic(errStr)
```
yuzisun, line 145: proxy in general should not be panic

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
+/*
```
yuzisun, line 1: need a better name for this file

cmd/batcher/main.go
```diff
@@ -0,0 +1,58 @@
...
+func main() {
+	flag.Parse()
+
+	logf.SetLogger(logf.ZapLogger(false))
+	log := logf.Log.WithName("entrypoint")
+
+	maxBatchsizeInt, err := strconv.Atoi(*maxBatchsize)
+	if err != nil || maxBatchsizeInt <= 0 {
+		log.Info("Invalid max batchsize", "max-batchsize", *maxBatchsize)
+		os.Exit(-1)
```
yuzisun, line 45: convention is to the return positive numbers for exit code

cmd/batcher/main.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	maxLatencyFloat64, err := strconv.ParseFloat(*maxLatency, 64)
+	if err != nil || maxLatencyFloat64 <= 0.0 {
```
yuzisun, line 49: max latency probably should not have a upper bound

salanki: Agreed. This can vary a lot depending on use case. We have some huge models where inference time is 6s per request. I am ok with a 1s latency there for batching.

salanki:
@zhangrongguo thank you for those numbers, that's some pretty good result. I also understand the sidecar setup for CPU based inferencers. When you say large data transfer, are you sending several thousand images per second, or what is the scale we are talking about here?

It would be really great to be able to run the batcher globally, and not only per Pod. For GPU Inferencing, you usually only have 1 GPU per Pod. For large models, where inference time is measured in seconds, batching at the per pod level will be significantly less efficient than batching right at the pipeline input.

For example, in my setup, I have ~50 GPUs that receive inference requests. Each inference takes 2 seconds, leading to a query rate of 25qps. If I run the batcher before the predictors in the pipeline, and want a batch size of 5, I will have to buffer for ~200ms. If I on the other hand run it on the predictor, I'd have to batch for ~10s.

cliveseldon:
@zhangrongguo @salanki Interesting thoughts. 

Sounds like we'd want to be able to choose where the batcher goes, either in front of a particular component or at the start of the pipeline. Not sure how easy it is to put it next to the knatve q-proxy? @yuzisun 

On the numbers,  @salanki can you explain your reasoning a bit more as it seems to assume a fixed 25qps whereas the addition of the batcher would hopefully change the max qps even if in front of the model, so you wouldn't be waiting 10s to construct a batch unless you enforced 25qps externally. It probably depends where requests are queued in knative.

cmd/batcher/main.go
zhangrongguo: That is a lower bound, not a upper bound.

config/default/configmap/inferenceservice.yaml
```diff
@@ -103,3 +103,11 @@ data:
         "cpuRequest": "100m",
         "cpuLimit": "1"
     }
+  batcher: |-
+    {
+        "image" : "gcr.io/kfserving/batcher:latest",
+        "memoryRequest": "8Gi",
```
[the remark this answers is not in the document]
zhangrongguo: We can reduce this.

docs/samples/batcher/basic/README.md
```diff
@@ -0,0 +1,65 @@
...
+kind: "InferenceService"
+metadata:
+  name: "sklearn-iris"
+spec:
+  default:
+    predictor:
+      minReplicas: 1
+      batcher:
+        port: 8082
+        svcHost: "127.0.0.1"
```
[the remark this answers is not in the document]
zhangrongguo: I will remove the svcHost and svcPort.

docs/samples/batcher/basic/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+We can now send a request to the sklearn model. Use `kfserving-ingressgateway` as your `INGRESS_GATEWAY` if you are deploying KFServing as part of Kubeflow install, and not independently.
+
+```
+MODEL_NAME=sklearn-iris
+INPUT_PATH=@./iris-input.json
+INGRESS_GATEWAY=istio-ingressgateway
+CLUSTER_IP=$(kubectl -n istio-system get service $INGRESS_GATEWAY -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+SERVICE_HOSTNAME=$(kubectl get inferenceservice sklearn-iris -o jsonpath='{.status.url}' | cut -d "/" -f 3)
+curl -v -H "Host: ${SERVICE_HOSTNAME}" http://$CLUSTER_IP/v1/models/$MODEL_NAME:predict -d $INPUT_PATH -H "Content-Type: application/json"
```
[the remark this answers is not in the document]
zhangrongguo: I will modify it.

pkg/batcher/controllers/default.go
zhangrongguo: It is the default controller name of Beego.

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+
+var (
+	channelIn = make(chan Input)
+	batcherInfo BatcherInfo
+	batcherHandler *BatcherHandler = nil
+	mutex sync.Mutex
+)
+
+type MainController struct {
+	beego.Controller
```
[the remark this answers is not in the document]
zhangrongguo: I haven't considered about the RPC yet. Beego is a high performance web framework. If net/http/httputil can support both HTTP and RPC, we can use net/http/httputil.

pkg/batcher/controllers/default.go
zhangrongguo: It was used for debug. Shall I remove it?

pkg/batcher/controllers/default.go
zhangrongguo: I will modify it.

pkg/batcher/controllers/default.go
zhangrongguo: Can you explain it in detail?

pkg/batcher/controllers/default.go
zhangrongguo: We have to make sure the response body is like this  "{\"predictions\":[1,2]}".

pkg/batcher/controllers/default.go
zhangrongguo: Uh, that is redundant. I will remove it.

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+		batcherInfo.Info[req.ContextInput] = InputInfo{
+			req.ChannelOut,
+			index,
+		}
+		batcherInfo.CurrentInputLen = len(batcherInfo.Instances)
+	default:
+		time.Sleep(SleepTime)
+	}
+	batcherInfo.Now = GetNowTime()
```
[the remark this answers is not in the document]
zhangrongguo: Even if the consumer got no data from the channel, we still have to check if the time meets the max latency.

pkg/batcher/controllers/default.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+		if batcherHandler.MaxLatency > 0.0 {
+			batcherInfo.MaxLatency = batcherHandler.MaxLatency
+		}
+	}
+	batcherInfo.Info = make(map[*context.Context] InputInfo)
+	batcherInfo.Start = GetNowTime()
+	batcherInfo.Now = batcherInfo.Start
+	batcherInfo.CurrentInputLen = 0
+	for true {
```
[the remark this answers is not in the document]
zhangrongguo: I will modify it.

cmd/batcher/main.go
zhangrongguo: I will modify it.

salanki:
@cliveseldon our use case is real-time prediction. The QPS is therefore fixed, and it depends only on active users on our application. Since it is real-time, latency is important. By doing batching before the load balancer, we could increase throughput without adding a lot of latency. The benefit here is better resource utilization, requiring less GPUs.

cmd/batcher/main.go
```diff
@@ -0,0 +1,58 @@
...
+	"flag"
+	"os"
+	"strconv"
+	"github.com/kubeflow/kfserving/pkg/batcher"
+	"github.com/kubeflow/kfserving/pkg/batcher/controllers"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var (
+	port          = flag.String("port", "8082", "Logger port")
```
pugangxa, line 29: Should be Batcher port rather than "Logger port" right?

pkg/controller/inferenceservice/resources/knative/service.go
```diff
@@ -124,6 +124,39 @@ func addLoggerContainerPort(container *v1.Container) {
...
+			}
+		}
+		return true
+	}
+	return false
+}
+
+func addBatcherContainerPort(container *v1.Container) {
+	if container != nil {
+		if container.Ports == nil {
```
pugangxa, line 149: Here batcher port is added when container ports is nil right? Does that mean we can not use logger and batcher at the same time?
In fact knative can only expose one user port, before if there's logger then logger port will be exposed. Now from code logger will take priority because it's before batcher?

cliveseldon: We should be tie it up in pod mutator to ensure logger->batcher->server ?

ellis-bigelow:
What do we think about combining the logger/batcher into the same go proxy?
https://github.com/kubeflow/kfserving/issues/802

ellis-bigelow:
Hey it's great to see progress on this, but I do want to caution a little bit. Until this design is reviewed, writing up all the code may require a rewrite if we decide to change course on the design.

It's definitely good to get familiar with the code base and implement a prototype in the meantime, but I just want you to be aware.

pugangxa:
> What do we think about combining the logger/batcher into the same go proxy?
> #802

Agreed. One proxy is more extensible,  like Mixer in istio at beginning and integrated in envoy at last.

cmd/batcher/main.go
zhangrongguo: I will modify it.

PR #637, opened 2020-01-20
Introduce parallelism field and make tornado num of workers configurable

yuzisun:
**What this PR does / why we need it**:
- Currently tornado server starts the number of workers with all the available cores on a node, this can be expensive since model server may load large model onto many processes which consumes a lot of memory, we need to make the number of workers configurable.  After the change it still retains the old behavior, but user can overwrite the number of tornado workers at component level. There is also multiprocessing issue regarding to pytorch, cuda can not be initialized in multiprocess env.
- Another issue we currently have is that the scaling target can only be configured globally and we need to make configurable at component (transformer/explainer/predictor) level.

So here we want to introduce the parallelism field at the component level to configure the scaling target and set the proper number of tornado workers.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubeflow/kfserving/issues/432

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfserving/637)

PR #806, opened 2020-05-05
Set ContainerConcurrency to Parallelism

salanki:
The Parallelism field introduced with #637 sets the KNative autoscaler annotation. It does however not set the KNative `ContainerConcurrency` field, which means that the logic in `queue-proxy` is mostly bypassed. It never manages timeouts nor the connection queue. This has a number of drawbacks:

1) The serving instance will receive all connections directly, and will not handle timeouts and queueing as gracefully as one wants (or at all).
2) The load balancing function of KNative Activator does not work as intended. Instead of intelligently directing requests to Pods that have resources available, it will emit them randomly. See https://github.com/knative/serving/issues/7792#issuecomment-622285173 for more detail.

KNative Activator, when correctly set up with `ContainerConcurrency` provides a remarkable latency improvement especially for multi-second inferences.

**Before this PR (Activator disabled)**

**With this PR (Activator enabled)**

I do not see any reason to have a separate configuration knob for this. The implicit intention of the `parallelism` field lines up well with the `ContainerConcurrency` field. I consider this a bug fix more than a new feature. I assumed for a long time that `parallelism` would have configured all relevant settings in the managed KNative resource regarding concurrency.

**Release note**:
```release-note
KNative ContainerConcurrency is now set based on the parallelism defined for the component in InferenceService. This allows queueing to be offloaded from the serving container, and KNative Activator to be used for managing high load situations.
```

### DIFF
--- a/pkg/controller/inferenceservice/resources/knative/service.go
+++ b/pkg/controller/inferenceservice/resources/knative/service.go
@@ -148,7 +148,7 @@ func (c *ServiceBuilder) CreatePredictorService(name string, metadata metav1.Obj
 	if isCanary {
 		endpoint = constants.InferenceServiceCanary
 	}
-
+	concurrency := int64(predictorSpec.Parallelism)
 	service := &knservingv1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -170,7 +170,8 @@ func (c *ServiceBuilder) CreatePredictorService(name string, metadata metav1.Obj
 					Spec: knservingv1.RevisionSpec{
 						// Defaulting here since this always shows a diff with nil vs 300s(knative default)
 						// we may need to expose this field in future
-						TimeoutSeconds: &constants.DefaultPredictorTimeout,
+						TimeoutSeconds:       &constants.DefaultPredictorTimeout,
+						ContainerConcurrency: &concurrency,
 						PodSpec: v1.PodSpec{
 							ServiceAccountName: predictorSpec.ServiceAccountName,
 							Containers: []v1.Container{
@@ -219,6 +220,7 @@ func (c *ServiceBuilder) CreateTransformerService(name string, metadata metav1.O
 		endpoint = constants.InferenceServiceCanary
 	}
 
+	concurrency := int64(transformerSpec.Parallelism)
 	service := &knservingv1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -240,7 +242,8 @@ func (c *ServiceBuilder) CreateTransformerService(name string, metadata metav1.O
 					Spec: knservingv1.RevisionSpec{
 						// Defaulting here since this always shows a diff with nil vs 300s(knative default)
 						// we may need to expose this field in future
-						TimeoutSeconds: &constants.DefaultTransformerTimeout,
+						TimeoutSeconds:       &constants.DefaultTransformerTimeout,
+						ContainerConcurrency: &concurrency,
 						PodSpec: v1.PodSpec{
 							ServiceAccountName: transformerSpec.ServiceAccountName,
 							Containers: []v1.Container{
@@ -285,6 +288,7 @@ func (c *ServiceBuilder) CreateExplainerService(name string, metadata metav1.Obj
 		addLoggerContainerPort(container)
 	}
 
+	concurrency := int64(explainerSpec.Parallelism)
 	service := &knservingv1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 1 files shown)


pugangxa:
Read the content in https://knative.dev/development/serving/configuring-autoscaling/, as mentioned:
```
Note: It is recommended to only use the hard limit if your application has a clear need for it. Low hard limit values can have an impact on the throughput and latency of the application.
Note: If both a soft and a hard limit are specified, the smaller of the two values will be used as to not have the autoscaler target a value that's not even allowed into the replica by the hard limit.
```
So seems we need only use the ContainerConcurrency and remove the target annotation?

salanki:
Yes, it does seem so indeed! I'll make the adjustments.

yuzisun:
> Read the content in https://knative.dev/development/serving/configuring-autoscaling/, as mentioned:
> 
> ```
> Note: It is recommended to only use the hard limit if your application has a clear need for it. Low hard limit values can have an impact on the throughput and latency of the application.
> Note: If both a soft and a hard limit are specified, the smaller of the two values will be used as to not have the autoscaler target a value that's not even allowed into the replica by the hard limit.
> ```
> 
> So seems we need only use the ContainerConcurrency and remove the target annotation?

`ContainerConcurrency` is the hard limit and target annotation is soft limit, I think we should have both as knative does. We can default container concurrency to 0 and user can customize to set a low container concurrency based on their needs.

pugangxa:
> I think we should have both as knative does. We can default container concurrency to 0 and user can customize to set a low container concurrency based on their needs.
Yes ContainerConcurrency is the hard limit while target is the soft one. But in this fix they all come from the same value so keep both seems meaningless. If we really need them think we need place to set these values separately.

ellis-bigelow:
Great! Can you assert this with a quick test?

yuzisun:
@salanki I think this is a critical performance fix for KFServer or any custom python model server, for C++ model server tfserving, triton that you can not pinpoint  a number of concurrent requests to handle then `containerConcurrency 0` might be a good default.

I did a test with this PR for sklearn example, it does give remarkable performance when `containerConcurrency` is set to 4.
```
Container Concurrency 0 without activator
Requests      [total, rate]            3000, 10.00
Duration      [total, attack, wait]    4m59.981365005s, 4m59.899965944s, 81.399061ms
Latencies     [mean, 50, 95, 99, max]  14.454228ms, 5.242094ms, 81.460854ms, 84.099188ms, 98.099025ms
Bytes In      [total, mean]            69000, 23.00
Bytes Out     [total, mean]            246000, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  

Container Concurrency 4 without activator
Requests      [total, rate]            3000, 10.00
Duration      [total, attack, wait]    4m59.903721136s, 4m59.899950721s, 3.770415ms
Latencies     [mean, 50, 95, 99, max]  4.230695ms, 3.985602ms, 4.722888ms, 5.354227ms, 123.64256ms
Bytes In      [total, mean]            69000, 23.00
Bytes Out     [total, mean]            246000, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  

Container Concurrency 0 with activator
Requests      [total, rate]            3000, 10.00
Duration      [total, attack, wait]    4m59.907191426s, 4m59.900006454s, 7.184972ms
Latencies     [mean, 50, 95, 99, max]  10.2892ms, 7.048687ms, 51.114427ms, 69.242763ms, 165.101924ms
Bytes In      [total, mean]            69000, 23.00
Bytes Out     [total, mean]            246000, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  
Error Set:

container concurrency 4 with activator
Requests      [total, rate]            3000, 10.00
Duration      [total, attack, wait]    4m59.906042805s, 4m59.899960704s, 6.082101ms
Latencies     [mean, 50, 95, 99, max]  8.133857ms, 5.941656ms, 25.440998ms, 26.688303ms, 456.370488ms
Bytes In      [total, mean]            69000, 23.00
Bytes Out     [total, mean]            246000, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  
Error Set:
```

yuzisun:
> How many Pods did you load balance over in your benchmark? And is this CPU or GPU?
> 
> We run TFServing on GPU, where it can reasonably only serve one request at a time (`parallelism: 1`). The activator makes a huge difference here, so it is not only for Python.
> 
> If you set `ContainerConcurrency` to 0, activator will load balance on random, it will be worse in most cases than the regular Istio round robin load blancing.
> 
> I am surprised to see that you are getting better tail latencies without activator on concurrency 4 than with. That is strange to me.
> 
> I will get the test done.

I am on knative 0.12, I heard there is some performance issue with activator in that version and latest version has much improved.

salanki:
Possible. I'm on 0.13.

yuzisun:
This is a critical fix for KFServing performance! thanks @salanki !

/lgtm
/approve

salanki:
Thank you! We should have figured out the testing environment and have a test written tomorrow.

yuzisun:
> Thank you! We should have figured out the testing environment and have a test written tomorrow.

cool! follow the develop guide and let me know if you need any help.

PR #807, opened 2020-05-05
Disable retries in Istio VirtualService

salanki:
Currently, the retires in the VirtualService are setup to retry three times on all errors. KNative will also retry three times on all errors between a Transformer and Predictor. These settings, combined with default high timeouts can easily lead to a self-DDoS in an overload condition. See screenshot.

Screen Shot 2020-04-30 at 16 35 04

KNative has removed the retries in their VirtualServices: https://github.com/knative/serving/issues/6549#issuecomment-623800315

KFserving should do one of the following:
1) Disable retries completely (which is what this PR does)
2) Allow retries, but not on the 502/504 that are thrown on timeouts. Retrying on `connect-failure` should be safe.
3) Make the behavior configurable, as users might have different requirements for if they want KFServing to handle retrying or they already have a mechanism to do that higher up in the stack.

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/inferenceservice/resources/istio/virtualservice.go
+++ b/pkg/controller/inferenceservice/resources/istio/virtualservice.go
@@ -228,8 +228,8 @@ func (r *VirtualServiceBuilder) CreateVirtualService(isvc *v1alpha2.InferenceSer
 			},
 			Route: explainRouteDestinations,
 			Retries: &istiov1alpha3.HTTPRetry{
-				Attempts:      3,
-				PerTryTimeout: RetryTimeout,
+				Attempts:      0,
+				PerTryTimeout: nil,
 			},
 		}
 		httpRoutes = append(httpRoutes, explainRoute)
@@ -257,8 +257,8 @@ func (r *VirtualServiceBuilder) CreateVirtualService(isvc *v1alpha2.InferenceSer
 		},
 		Route: predictRouteDestinations,
 		Retries: &istiov1alpha3.HTTPRetry{
-			Attempts:      3,
-			PerTryTimeout: RetryTimeout,
+			Attempts:      0,
+			PerTryTimeout: nil,
 		},
 	}
 	httpRoutes = append(httpRoutes, predictRoute)
--- a/pkg/controller/inferenceservice/controller_test.go
+++ b/pkg/controller/inferenceservice/controller_test.go
@@ -23,7 +23,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kubeflow/kfserving/pkg/controller/inferenceservice/resources/istio"
 	"knative.dev/pkg/network"
 
 	"github.com/kubeflow/kfserving/pkg/controller/inferenceservice/resources/knative"
@@ -274,8 +273,8 @@ func TestInferenceServiceWithOnlyPredictor(t *testing.T) {
 						},
 					},
 					Retries: &istiov1alpha3.HTTPRetry{
-						Attempts:      3,
-						PerTryTimeout: istio.RetryTimeout,
+						Attempts:      0,
+						PerTryTimeout: nil,
 					},
 				},
 			},
@@ -761,8 +760,8 @@ func TestInferenceServiceWithDefaultAndCanaryPredictor(t *testing.T) {
 						},
 					},
 					Retries: &istiov1alpha3.HTTPRetry{
-						Attempts:      3,
-						PerTryTimeout: istio.RetryTimeout,
+						Attempts:      0,
+						PerTryTimeout: nil,
 					},
 				},
 			},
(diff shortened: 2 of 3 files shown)


pkg/controller/inferenceservice/resources/istio/virtualservice.go
```diff
@@ -210,8 +210,8 @@ func (r *VirtualServiceBuilder) CreateVirtualService(isvc *v1alpha2.InferenceSer
 		},
 		Route: predictRouteDestinations,
 		Retries: &istiov1alpha3.HTTPRetry{
-			Attempts:      3,
-			PerTryTimeout: RetryTimeout,
+			Attempts:      0,
+			PerTryTimeout: nil,
 		},
```
pugangxa, line 215: Searched there's 2 definition of "retry" here,  one for predictor and another for explainer, is there a reason we only modify one and keep another?
BTW, there's test cases in controller_test.go and virtualservice_test.go, I'm afraid it will fail after the change.

salanki: No reason, purely oversight from my part. I'll fix the tests as well, having some trouble getting them to run locally.

yuzisun: Agree that this should be retried by end user, let's hold and see how knative's experiment is going if there is any regression.

ellis-bigelow:
Nice!!
+1 caller should be able to control retries and not default to 3.

What do we think about adding #retries into the ComponentSpecs and default them to zero?

salanki:
Assuming that the KNative no-retries stays in, I think that is a good idea @ellis-bigelow. Then there is a use case for the external facing `VirtualService` (which is the KFServing VS) to support retries, if one doesn't want to do it client side.

If KNative would reverse the PR and keep their retries, there is no point to retry in the KFServing `VirtualService` as there are already plenty of retries in the path.

I am all for configurability with sane defaults. Defaulting to `0` should be the most intuitive, and avoid people DoS-ing themselves and having to dig through KFServing internals to understand why. The thought "Oh, I'm getting errors, let me see if there is a retry function I can enable" is easier to reach than "Why did I just spike up to 500 connections, is someone attacking me??".

ellis-bigelow:
> Assuming that the KNative no-retries stays in, I think that is a good idea @ellis-bigelow. Then there is a use case for the external facing `VirtualService` (which is the KFServing VS) to support retries, if one doesn't want to do it client side.
> 
> If KNative would reverse the PR and keep their retries, there is no point to retry in the KFServing `VirtualService` as there are already plenty of retries in the path.
> 
> I am all for configurability with sane defaults. Defaulting to `0` should be the most intuitive, and avoid people DoS-ing themselves and having to dig through KFServing internals to understand why. The thought "Oh, I'm getting errors, let me see if there is a retry function I can enable" is easier to reach than "Why did I just spike up to 500 connections, is someone attacking me??". The retries can also hide some errors that should be troubleshooted at initial rollout.

+1000

ellistarn:
Any word on this @salanki

salanki:
KNatives removal of retries is in and it seems as it will stick. I believe making retries configurable in KFServing is the most reasonable. Would you like me to implement that in this PR, or would that be a larger spec discussion?

If implement here, I'd suggest simply exposing the Istio `HTTPRetry` to the `InferenceService` spec, and defaulting it to disabled.

ellistarn:
Do you have a strong use case for retries? Ideally we would not add this into the spec as it further complicates things. I'd love this to be something Istio could layer in. Perhaps a RetryPolicy that applies to a Pod Label Selector. 

For now, given that Knative has removed it, unless we have a strong usecase to include it, I think it's probably best to leave it out .

salanki:
I'm fine with removing it and letting users who want it create a custom `VirtualService`. It is a bit of the same issue as supporting CORS. The question is, do we want `InferenceService` to fully manage all aspects of the inference service, Ie. be the one place that controls everything related to the API?

Right now, it is a bit awkward to have a VirtualService generated by KFServing that is also the API endpoint retrieved with `kubectl get inferenceservices` and then having to create a separate, manual `VirtualService` if one needs retries.

ellistarn:
Dan and I were chatting about this. It seems like users should be able to
drop in their own VirtualService configs on top of InferenceServices to
configure these things.

I think the long term direction we like is to not generate these configs at
all, but document for users how to layer VirtualServices on top. Do you
want to give this a shot?

salanki:
For the scope of this PR, I think we should do one of two things:

1) Cancel the PR, keeping retries
2) Keep the PR as is, update tests, and merge it, disabling retries.

I'm leaning towards 2.

Then I'd open up a new ticket to discuss the approach we talked about on Slack, to look for an existing VS to modify before creating a new one.

ellistarn:
Agree with #2. /lgtm /approve

salanki:
Tests fixed.

yuzisun:
Nice work! Thanks @salanki !
/lgtm
/approve

PR #822, opened 2020-05-13
Add an easier method to obtain Istio-Dex authentication token

sachua:

Documents an easier method to obtain Dex authentication token by using browser cookies.


yuzisun:
Awesome work @sachua !

/lgtm
/approve

PR #823, opened 2020-05-13
allow configurable region name

harshavardhana:
Configuration region-name to avoid get_bucket_location API calls, some users
may have restricted policies, thus allow for these situations. 

Fixes #

**Release note**:

```release-note

```


ellistarn:
Nice fix! Thanks!

ace-gabriel:
Nice fix! Solved our issue.

Our issue was that KFServing gets access denied error when downloading the model from bucket using Minio client. Turned out that we don't have get_bucket_location access. This will avoid similar issue

yuzisun:
thanks @harshavardhana for the prompt fix!

PR #837, opened 2020-05-21
make default req log url more flexible

ryandawsonuk:
fixes https://github.com/kubeflow/kfserving/issues/836

Don't need to set namespace in the url as the short-form url 'default-broker' will find current namespace. To hit a specific namespace the env var could be set to 'default-broker.specificnamespace'

Parallel to https://github.com/SeldonIO/seldon-core/pull/1861

### DIFF
--- a/config/configmap/inferenceservice.yaml
+++ b/config/configmap/inferenceservice.yaml
@@ -72,7 +72,8 @@ data:
         "memoryRequest": "100Mi",
         "memoryLimit": "1Gi",
         "cpuRequest": "100m",
-        "cpuLimit": "1"
+        "cpuLimit": "1",
+        "defaultUrl": "http://default-broker"
     }
   batcher: |-
     {
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -80,6 +80,11 @@ var (
 	PodMutatorWebhookName                  = KFServingName + "-pod-mutator-webhook"
 )
 
+// Request Logging Constants
+var (
+	LoggerDefaultUrl = getEnvOrDefault("REQUEST_LOGGING_DEFAULT_ENDPOINT", "default-broker")
+)
+
 // GPU Constants
 const (
 	NvidiaGPUResourceType = "nvidia.com/gpu"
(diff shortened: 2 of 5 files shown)


pkg/constants/constants.go
```diff
@@ -76,6 +76,11 @@ var (
 	PodMutatorWebhookName                  = KFServingName + "-pod-mutator-webhook"
 )
 
+// Request Logging Constants
+var (
+	LoggerDefaultUrl = getEnvOrDefault("REQUEST_LOGGING_DEFAULT_ENDPOINT", "http://default-broker")
```
yuzisun, line 81: Can we document this environment variable?

ryandawsonuk: Have now added an explanation to the docs for the main request logging example

yuzisun: @ryandawsonuk can we move this option to the configmap [logger section](https://github.com/kubeflow/kfserving/blob/master/config/configmap/inferenceservice.yaml#L98) ? I think environment variable is a bit hard to discover

ryandawsonuk: Makes sense, will have a look

ryandawsonuk: @yuzisun have now made this update

config/configmap/inferenceservice.yaml
```diff
@@ -72,7 +72,8 @@ data:
         "memoryRequest": "100Mi",
         "memoryLimit": "1Gi",
         "cpuRequest": "100m",
-        "cpuLimit": "1"
+        "cpuLimit": "1",
+        "defaultUrl": "http://default-broker"
```
yuzisun, line 76: So this will require the knative eventing install right?

ryandawsonuk: It defaults to looking for a knative broker but  it will just result in a warning if none is there.  It’s actually the same as current as right now it uses “default-broker.namespace”, which is the same as “default-broker” since having the current namespace in a service address is optional in k8s

PR #888, opened 2020-06-17
Add additional unit tests to event Recorder

### LINKED ISSUE #881: Record events when InferenceService leaves and enters a Ready State
/kind feature

For alerting purposes, it is important for events to be thrown when an InferenceService is `Ready` to receive requests, as is its transition from `Not-Ready` to `Ready`. 

The proposed feature should throw a transition event: 

`InferenceServer is Ready` when the InferenceService goes from `NotReady -> Ready`
`InferenceServer is no longer Ready` when the InferenceService goes from `Ready -> NotReady

ifilonenko:
Adds addition test to showcase event recording when: Ready --> Non-Ready --> Ready

Fixes #881 

**Release note**:

### DIFF
--- a/pkg/controller/inferenceservice/controller_test.go
+++ b/pkg/controller/inferenceservice/controller_test.go
@@ -345,11 +345,11 @@ func TestInferenceServiceWithOnlyPredictor(t *testing.T) {
 		}
 		for _, event := range events.Items {
 			if event.Reason == string(kfserving.InferenceServiceReadyState) &&
-				event.Type == v1.EventTypeNormal {
+				event.Type == v1.EventTypeNormal && event.Count == 1 {
 				return nil
 			}
 		}
-		return fmt.Errorf("Test %q failed: events [%v] did not contain ready", serviceName, events.Items)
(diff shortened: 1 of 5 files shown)


pkg/controller/inferenceservice/controller_test.go
```diff
@@ -374,13 +395,70 @@ func TestInferenceServiceWithOnlyPredictor(t *testing.T) {
 		if len(events.Items) == 0 {
 			return fmt.Errorf("Test %q failed: no events were created", serviceName)
 		}
+		var passes = 0
 		for _, event := range events.Items {
-			if event.Reason == string(kfserving.InferenceServiceNotReadyState) &&
-				event.Type == v1.EventTypeWarning {
-				return nil
+			if event.Reason == string(kfserving.InferenceServiceReadyState) &&
```
yuzisun, line 400: We can make a function `getEvents` which returns an array of `event.Reason` that is sorted by timestamp, so we can assert with the expected state sequence and make the tests less lengthy.

PR #893, opened 2020-06-20
Move KFServing to minimally require kubernetes 1.15

### LINKED ISSUE #891: Move KFServing to minimally require Kubernetes 1.15
/kind feature

- Add object selector for the pod mutator which is supported in kubernetes 1.15
- support new markers for go files or openAPI spec
- update crd install instruction where `--validate=false` flag is needed

yuzisun:

Fixes #891 

a couple of markers are generated by `controller-gen` but it is only supported by kubernetes 1.16 and 1.17, so we need to add `--validate=false` for installing KFServing on kubernetes 1.15.
https://kubernetes.io/docs/reference/using-api/api-concepts/#merge-strategy

**Release note**:

```release-note
KFServing now minimally requires kubernetes 1.15 to use the object selector for pod mutator
```

### DIFF
--- a/config/default/manager/manager.yaml
+++ b/config/default/manager/manager.yaml
@@ -21,7 +21,7 @@ spec:
       containers:
       - command:
         - /manager
-        image: github.com/kubeflow/kfserving/cmd/manager
+        image: ko://github.com/kubeflow/kfserving/cmd/manager
         imagePullPolicy: Always
         name: manager
         env:
--- a/config/default/webhook/manifests.yaml
+++ b/config/default/webhook/manifests.yaml
@@ -31,10 +31,10 @@ webhooks:
         path: /mutate-pods
     failurePolicy: Fail
     name: inferenceservice.kfserving-webhook-server.pod-mutator
-    namespaceSelector:
+    objectSelector:
       matchExpressions:
-        - key: control-plane
-          operator: DoesNotExist
+        - key: serving.kubeflow.org/inferenceservice
+          operator: Exists
     rules:
       - apiGroups:
           - ""
--- a/hack/quick_install.sh
+++ b/hack/quick_install.sh
@@ -68,10 +68,8 @@ helm template --namespace=istio-system \
 kubectl apply -f istio-local-gateway.yaml
 
 # Install Knative
-kubectl apply --selector knative.dev/crd-install=true \
---filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving.yaml \
-
-kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving.yaml \
+kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving-crds.yaml
+kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving-core.yaml
 
 cd ..
 # Install KFServing
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ deploy: manifests
 	kustomize edit remove resource certmanager/certificate.yaml; \
 	else kustomize edit add resource certmanager/certificate.yaml; fi;
 
-	kustomize build config/default | kubectl apply -f -
+	kustomize build config/default | kubectl apply --validate=false -f -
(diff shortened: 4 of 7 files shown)


cliveseldon:
@yuzisun Just wondered why we need `--validate false` is it the controller-gen issue we disucssed?

yuzisun:
> @yuzisun Just wondered why we need `--validate false` is it the controller-gen issue we disucssed?

@cliveseldon yes, do you have better way to solve this?

config/default/manager/manager.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       containers:
       - command:
         - /manager
-        image: github.com/kubeflow/kfserving/cmd/manager
+        image: ko://github.com/kubeflow/kfserving/cmd/manager
```
cliveseldon, line 24: Can you explain this change?

config/default/webhook/manifests.yaml
```diff
@@ -31,10 +31,10 @@ webhooks:
         path: /mutate-pods
     failurePolicy: Fail
     name: inferenceservice.kfserving-webhook-server.pod-mutator
-    namespaceSelector:
+    objectSelector:
       matchExpressions:
-        - key: control-plane
-          operator: DoesNotExist
+        - key: serving.kubeflow.org/inferenceservice
+          operator: Exists
```
cliveseldon, line 37: Does this mean every InferenceService must have this label? If so would we not need to update all our examples?
Also, is this the keyt breaking change for 1.15 k8s? Note kubeflow have 1.14 support for 1.0 but maybe this will be dropped by the time this version of KFServing gets in Kubeflow I assume. https://www.kubeflow.org/docs/started/k8s/overview/#minimum-system-requirements

hack/quick_install.sh
```diff
@@ -68,10 +68,8 @@ helm template --namespace=istio-system \
 kubectl apply -f istio-local-gateway.yaml
 
 # Install Knative
-kubectl apply --selector knative.dev/crd-install=true \
---filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving.yaml \
-
-kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving.yaml \
+kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving-crds.yaml
+kubectl apply --filename https://github.com/knative/serving/releases/download/${KNATIVE_VERSION}/serving-core.yaml
```
cliveseldon, line 73: Does this separation of yamls mean also a new mnimum version of KNative serving?

config/default/webhook/manifests.yaml
yuzisun: @cliveseldon the inference service already has [this label](https://github.com/kubeflow/kfserving/blob/ff54d08749163fc7603e231edcc9dc6019a05f73/pkg/controller/inferenceservice/resources/knative/service.go#L166), we internally already use this with k8s 1.15. We did not add it because currently kubeflow min k8s requirement is 1.14 which does not support object selector, kubeflow is now working on to support 1.16, we can integrate KFServing once kubeflow moves to 1.15 or 1.16. Note that latest Knative min requirement is already [k8s 1.16](https://github.com/knative/serving/releases/tag/v0.15.1)

config/default/manager/manager.yaml
yuzisun: I got this warning when running `make deploy-dev` , so all the image path needs to be prefixed with `ko://` now.
```
We will start requiring ko:// in a coming release.  Please prefix
the following import path for things to continue working:

   github.com/kubeflow/kfserving/cmd/logger
```

hack/quick_install.sh
yuzisun: No, the previous installation has a race condition, so here is to fix using the correct way according to the [knative doc here](https://knative.dev/docs/install/any-kubernetes-cluster), also this quick install is only for demonstration purpose.

config/default/webhook/manifests.yaml
cliveseldon: ok yes. I forgot our webhook is for Pods not InferenceServices

cliveseldon:
ok looks good. We should add add an issue to try and remove the need for `--validate false`?

/approve

PR #894, opened 2020-06-22
Remove default queue proxy resource limit and Add KFServing benchmarking 

### LINKED ISSUE #844: KFServing tail latency issue due to k8s cpu throttle for knative queue proxy
/kind bug

Currently KFserving sets queue proxy resources limit with `queue.sidecar.serving.knative.dev/resourcePercentage: 20` annotation, however due to kubernetes/kubernetes#67577, queue proxy can still get throttled even it does not hit the cpu limit if the OS does not have [kernel fix](https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1), this affects the KFServing tail latency at p95/p99.

related upstream knative issues 
https://github.com/knative/serving/issues/8057
https://github.com/knative/serving/issues/7349

KFServing performance should be comparable to raw kubernetes deployment with knative queue proxy

**Environment:**

- Istio Version:1.1.6
- Knative Version: 0.11.2
- KFServing Version: 0.3
- Kubeflow version: 1.0
- Minikube version:

yuzisun:
- Removing default KFServing queue proxy percentage based resource limit annotation which causes high tail latency due to a [linux kernel bug](https://github.com/kubernetes/kubernetes/issues/67577).
- Add KFServing benchmark result and compare with raw kubernetes service/deployments using HPA.

***before***
```
Requests      [total, rate]            300, 5.02
Duration      [total, attack, wait]    59.840230434s, 59.800002674s, 40.22776ms
Latencies     [mean, 50, 95, 99, max]  6.591095ms, 3.027777ms, 39.89522ms, 40.068618ms, 40.781162ms
Bytes In      [total, mean]            6900, 23.00
Bytes Out     [total, mean]            24600, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:300
```
***after***
```
Requests      [total, rate]            300, 5.02
Duration      [total, attack, wait]    59.804045855s, 59.799961298s, 4.084557ms
Latencies     [mean, 50, 95, 99, max]  4.427072ms, 4.306852ms, 4.902942ms, 5.399814ms, 33.870577ms
Bytes In      [total, mean]            6900, 23.00
Bytes Out     [total, mean]            24600, 82.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:300 
```

Fixes #844 

1. Removing the KFServing default queue proxy resource limit, Knative has made queue proxy resource limit configurable in https://github.com/knative/serving/pull/8195 which should be in the coming release.

**Release note**:

```release-note
For users who has resource quota turned on, we suggest add `queue.sidecar.serving.knative.dev/resourcePercentage` annotation manually on the inference service before knative releases the fix to make queue proxy resource limit configurable.
```


test/benchmark/README.md
```diff
@@ -0,0 +1,223 @@
...
+```bash
+kubectl apply -f ./sklearn_vegeta_cfg.yaml
+```
+- Create the benchmark job
+```bash
+kubectl create -f ./sk_benchmark.yaml
+```
+
+#### CC=8 With queue proxy and activator on the request path
+Create an `InferenceService` with `ContainerCurrency`(cc) set to 8 which equals to the number of codes on the node.
```
cliveseldon, line 42: typo codes->cores

test/benchmark/README.md
```diff
@@ -0,0 +1,223 @@
...
+also effectively acts as a load balancer which distributes the load across all the pods as they become available in a way that
+does not overload them with regards to their concurrency settings. So it protects the app from burst so you do not see messages
+queuing in the user pods.
+
+## Environment Setup
+- K8S: v1.14.10-gke.36
+- Istio: 1.1.6
+- Knative: 0.11.2
+- KFServing: master
+
```
cliveseldon, line 24: Can we say what GKE or other cluster nodes were used so easier for people to replicate?

yuzisun: ye I mentioned right above it is `v1.14.10-gke.36`

cliveseldon: So all defaults - 3 node n1-standard? Would it be possible to run node affinity to make the model run on a separate node to vegeta?

yuzisun: it is 8 nodes n1-standard, the vegeta is run as a kubernetes job so you can specify node anti-affinity to run on a different node.

salanki:
Which cgroup does the queue proxy end up in if there are no resource requests / limits? IIRC, scheduling a Pod with no resource requests / limits will put it in the root of the burstable CGroup where it will have a very small CPU share if the system goes full.

yuzisun:
> Which cgroup does the queue proxy end up in if there are no resource requests / limits? IIRC, scheduling a Pod with no resource requests / limits will put it in the root of the burstable CGroup where it will have a very small CPU share if the system goes full.

@salanki the proper way to configure the queue proxy resource limit is to use the configmap introduced in https://github.com/knative/serving/pull/8195 which allows for setting absolute limits instead of current percentage based limits with hard coded upper/lower bound, we found that by setting CPU with whole number on queue proxy you get the optimal performance than using fractional CPU.

yuzisun:
@pradithya what do you think removing the hard coded queue proxy resource limit here in favor of the configmap introduced in knative/serving#8195  since you run into the same cpu throttling issue with the kernel bug?

pradithya:
I am in favor with this, I did the similar implementation to fix the performance issue in my company.

test/benchmark/sk_benchmark.yaml
```diff
@@ -0,0 +1,37 @@
...
+  parallelism: 1
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
+    spec:
+      containers:
+        - args:
+            - vegeta -cpus=1 attack -duration=10m -rate=5/1s -targets=/var/vegeta/cfg
+              | vegeta report -type=text
```
cliveseldon, line 16: I think its best to rerun these tests again with `-keepalive=false` as I have found for calling internal SVCs you won't get proper round-robin as Vegeta will be reusing TCP connections by default.

yuzisun: I think in the real case user also want `keepalive=true` so vegeta default is good.

cliveseldon: But in the real case users will be usually calling via an Ingress which will ensure load balancing for downstream connections via say Envoy. I think the issue is that you may find in the HPA case that some pods get overloaded when internal SVC calls from Vegeta reuse TCP connections.

yuzisun: For the HPA case if you check the setup above I also create a virtual service which routes to the internal svc, so it is going through the local cluster gateway and envoy is doing the load balancing. The comparison should be fair since both are hitting from the cluster local gateway, for HPA case it directly routes to the user container from kubernetes internal svc while for KPA case it is gateway -> activator -> queue-proxy-> user container.

PR #900, opened 2020-06-28
Fix Test KFServing Installation in the main README

Tomcli:
Fix the Test KFServing Installation in the main README
- In order to run `port-forward` and the rest of the commands in the same shell, port-forward needs to run as background process.
- Run perf.yaml under the same testing namespace
- Update perf.yaml payload to the testing namespace and add `restartPolicy` to fix the below bug https://github.com/kubernetes/kubernetes/issues/41194
   ```
   The Job "load-test284zn" is invalid: spec.template.spec.restartPolicy: Unsupported value: "Always": supported values: "OnFailure", "Never"
   ```

**Release note**:

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,15 +99,15 @@ sklearn-iris   http://sklearn-iris.kfserving-test.example.com/v1/models/sklearn-
 ```
 4) Curl the `InferenceService`
 ```bash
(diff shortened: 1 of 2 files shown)


yuzisun:
Thanks @Tomcli for the fix!

/lgtm
/approve

PR #908, opened 2020-06-29
Create TrainedModel CRD

yuzliu:

**Special notes for your reviewer**: Create TrainedModel CRD for multi-model sharing

**Release note**:

```release-note
Add new TrainedModel CRD
```


pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,48 @@
...
+	// Defaults to requests and limits of 1CPU, 2Gb MEM.
+	Resources v1.ResourceRequirements `json:"resources,omitempty"`
+}
+type ModelVersionPolicy struct {
+	// +kubebuilder:webhooks:Enum={"all","latest","specific"}
+	ModelVersionPolicy string `json:"modelVersionPolicy"`
+	// A list of specific model versions to serve, it is required when model version policy is "specific"
+	ModelVersions []string `json:"modelVersions,omitempty"`
+}
+type TrainedModelDeployStatus struct {
```
ellistarn, line 42: Consider breaking this out into its own definition where we can implement something like propagate status. Also, consider "TrainedModelStatus"

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,48 @@
...
+	ModelVersionPolicy ModelVersionPolicy `json:"modelVersionPolicy,omitempty"`
+	// ML framework name
+	// +kubebuilder:webhooks:Enum={"tensorflow","pytorch","sklearn","onnx","xgboost", "custom"}
+	Framework string `json:"framework"`
+	// Framework version for the trained model
+	FrameworkVersion string `json:"frameworkVersion,omitempty"`
+	// Defaults to requests and limits of 1CPU, 2Gb MEM.
+	Resources v1.ResourceRequirements `json:"resources,omitempty"`
+}
+type ModelVersionPolicy struct {
```
ellistarn, line 36: I'm not convinced on this yet. Do we have use cases for users serving multiple model versions in the same inference service? "All" seems kind of dangerous. It also makes "Status" more complex. Would you be willing to punt on versioning for now and we can add it as we understand this more?

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,48 @@
...
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Default to latest
+	ModelVersionPolicy ModelVersionPolicy `json:"modelVersionPolicy,omitempty"`
+	// ML framework name
+	// +kubebuilder:webhooks:Enum={"tensorflow","pytorch","sklearn","onnx","xgboost", "custom"}
+	Framework string `json:"framework"`
+	// Framework version for the trained model
+	FrameworkVersion string `json:"frameworkVersion,omitempty"`
```
ellistarn, line 32: If we do need framework and version, can we collapse this into one string that corresponds to the git release tag (or subsequent model server docker image)? e.g. tensorflow-1.12

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,48 @@
...
+	// Explainer model spec
+	ExplainerModel ModelSpec `json:"explainerModel,omitempty"`
+}
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Default to latest
+	ModelVersionPolicy ModelVersionPolicy `json:"modelVersionPolicy,omitempty"`
+	// ML framework name
+	// +kubebuilder:webhooks:Enum={"tensorflow","pytorch","sklearn","onnx","xgboost", "custom"}
```
ellistarn, line 29: I think we should avoid restricting this enum to maintain extensibility.

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,48 @@
...
+	// Predictor model spec
+	PredictorModel ModelSpec `json:"predictorModel"`
+	// Explainer model spec
+	ExplainerModel ModelSpec `json:"explainerModel,omitempty"`
+}
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Default to latest
+	ModelVersionPolicy ModelVersionPolicy `json:"modelVersionPolicy,omitempty"`
```
ellistarn, line 27: I'd prefer to leave this out to start, but minimally it should be a pointer (optional)

ifilonenko: I agree, maybe we should keep the scope a little more contained for the first PR.

pkg/apis/serving/v1beta1/trained_model.go
ifilonenko: Isn't that what `custom` is for? or what kind of extensibility are you referring to?

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: I am trying to be consistent with triton server's version policy here: https://docs.nvidia.com/deeplearning/triton-inference-server/master-user-guide/docs/model_configuration.html?highlight=policy#version-policy. Triton provides all, specific and latest options. Tf-serving also allows users to server multiple versions at the same time: https://www.tensorflow.org/tfx/serving/serving_config

I do agree that it will make the status object more complicated and I am willing to start with the "latest" version and add support for "specific" and "all" as we gain more experience.

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: I think this will be an important field as both tf-serving and triton have version policy. But I am fine leave it out to  restrict the scope.

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: Sure, I can make that change.

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: Sounds good.

pkg/apis/serving/v1beta1/trained_model.go
kelaban: Could this be a path toward named  "custom" types? You can imagine implementing the KFServer interface for a proprietary format that has reusability and can be configured similarly to the name->image mapping the other frameworks have.

ellistarn: I've heard from multiple users that they want to be able to extend KFServing without being forced into the "custom" bucket. I.e., they want their own first class models for their organization. One of those users was bloomberg :)

yuzliu: I can see that someone want a "MyAwesomeInternalFramework" here, I can relax the enum check. Since we are on this topic, how do you think we can manage the mapping between machine learning framework and model server?

ellistarn: As long as we're using explicit assignment (i.e. model specifies its model server) we don't need any intelligent mapping. With automatic assignment, I'd expect model servers would advertise supported frameworks, perhaps using a whitelist or regex? Then some process (kfserving controller or kfserving sidecar/agent) would query the model server and do assignment.

yuzisun: exactly the model server would advertise the supported framework. e.g triton supports pytorch, tensorflow and onnx, so `TrainedModel` webhook would reject sklearn model if user tries to deploy that onto a parent triton inference server.

ellistarn: One note on this, ideally we want to make sure we have clean separation between control and data plane components. Ideally the control plane should communicate *only* through the k8s master apis. Happy to go into this further, but it has huge ramifications for the portability of the control plane (i.e. running on another network fabric or even managed cloud provider).

This implies that the InferenceService would need to contain "supported frameworks" in it's status. Perhaps this value can be set by the KFServing Agent? 

KFServing Agent calling K8s Masters (good)
KFServing Controller calling K8s Masters (good)
KFServing Agent communicating with KFServing Controller (bad)

yuzliu: I totally agree that the data pane and control pane should be separate. But why is KFServing Agent communicating with KFServing Controller is bad?

ellistarn: It's about network fabric. In locked down cluster scenarios, the kubernetes API may be the only whitelisted network ingress/egress between the control and data plane. Imagine a scenario where the KFServing controller isn't running in the cluster. Perhaps it's running on another cluster, or another environment entire (like AWS ECS).

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,38 @@
...
+	// Machine Learning <framework name>-<git tag>
+	// The values could be: "tensorflow-v2.2.0","pytorch-v1.5.1","sklearn-0.23.1","onnx-v1.7.0","xgboost-v1.1.1", "myawesomeinternalframework-1.1.0" etc.
```
ellistarn, line 28: Thoughts on making this a colon vs a dash to really reinforce the idea of a git tag?

pkg/apis/serving/v1beta1/trained_model_status.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
+)
+
+type TrainedModelStatus struct {
+	// Condition for "Deployed"
+	duckv1beta1.Status `json:",inline"`
+	// Addressable endpoint for the deployed trained model
+	// http://inferenceservice.metadata.name/v1/models/trainedmodel.metadata.name
```
ellistarn, line 11: nit, can you put something like <inferenceservice.metadata.name> instead of inferenceservice.metadata.name and the same for trainedmodel. Just to show it's a substitution.

pkg/apis/serving/v1beta1/trained_model.go
ellistarn: Ready to approve as soon as you remove this :)

pkg/apis/serving/v1beta1/trained_model.go
yuzisun: Can `supported frameworks` advertised in the KFServing controller configmap?

yuzliu: Yeah, I was going to add that in the configMap. Distracted by too many meetings!

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: I think we can go with ":".  Hyphen is widely used in docker/git tags, it may confuse people. We need to implement a validation webhook to validate the format.

pkg/apis/serving/v1beta1/trained_model_status.go
yuzliu: Sure

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,33 @@
...
+}
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Machine Learning <framework name>:<git tag>
+	// The values could be: "tensorflow:v2.2.0","pytorch:v1.5.1","sklearn:0.23.1","onnx:v1.7.0","xgboost:v1.1.1", "myawesomeinternalframework:1.1.0" etc.
+	Framework string `json:"framework"`
+	// Machine Learning model type, this field is used to match explainer type
+	// +kubebuilder:webhooks:Enum={"tabular","text","image"}
+	Type string `json:"type,omitempty"`
```
yuzliu, line 30: @cliveseldon @ellistarn I added the type field since we need it to match explainer type. WDYT ?

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: @ellistarn @yuzisun @kelaban @ifilonenko What's your opinion about introducing a TrainedModelBinding CRD? Our current approach requires users to create an InferenceService first and then create TrainedModel CR to load models into one of their InferenceServices.

I think introducing the TrainedModelBinding CRD can decouple model and model server.  The advantages are:
1) Users can create TrainedModel CR without creating a model server first.
2) An TrainedModel CR can be loaded into multiple model servers, this approach can reduce duplicate TrainedModels.

yuzliu:
Anyone has ideas about how this unit test timed out? All the unit tests passed locally, it seems that something blocked the ksvc.
```
--- FAIL: TestInferenceServiceWithOnlyPredictor (31.53s)
   controller_test.go:379:
       Timed out after 30.000s.
       Expected
           <bool>: false
       to be true
FAIL
```

config/configmap/inferenceservice.yaml
```diff
@@ -19,27 +19,43 @@ data:
...
+            "supportedFrameworks": [
+              "tensorflow"
             ]
         },
         "onnx": {
             "image": "mcr.microsoft.com/onnxruntime/server",
             "defaultImageVersion": "v0.5.1",
             "allowedImageVersions": [
                "v0.5.1"
+            ],
```
yuzliu, line 32: @yuzisun and @ellistarn I extended the inferenceservice configMap so each model server can advertise the ML frameworks it supports. I know onnx supports a lot more, I put popular ones here for now.

config/configmap/inferenceservice.yaml
```diff
@@ -69,6 +95,11 @@ data:
             "defaultImageVersion": "v0.3.0",
             "allowedImageVersions": [
                "v0.3.0"
+            ],
+            "supportedModelType": [
```
yuzliu, line 99: @cliveseldon I am trying to extend the inferenceservice configMap to allow the explainer server to advertise the model types it can support. Does it look right to you?

config/configmap/inferenceservice.yaml
yuzisun: I do not think onnx currently support multi model serving, @rakelkar can confirm.

yuzisun:
> Anyone has ideas about why this unit test timed out? All the unit tests passed locally, it seems that something blocked the ksvc.
> 
> ```
> --- FAIL: TestInferenceServiceWithOnlyPredictor (31.53s)
>    controller_test.go:379:
>        Timed out after 30.000s.
>        Expected
>            <bool>: false
>        to be true
> FAIL
> ```

@ifilonenko I run into this in my PR to..

pkg/apis/serving/v1beta1/trained_model.go
yuzisun: I think `TrainedModelBinding` blurred the line of model management system and KFServing, KFServing ideally shouldn't be in a position of staging the trained model which isn't ready for deployment yet imo.

yuzliu: I agree that KFServing should not be responsible for models that aren't ready for deployment. I can imagine that there will be a lot of models only stored in a storage system but do not get deployed out.  

Users should only create a TrainedModel CR when they want to deploy it. Having TrainedModelBinding allows users to deploy the same model to multiple model servers and create TrainedModel CR without running a model server first.

I do think it should be a phase 2 things, we don't have to add it now but want to get people to think about it.

ellistarn: I think bindings are too far. Are you worried about duplicating the model's fields? I'm worried about the blast radius of all the models in production if it's served in 4 places and 1 edits the model and the bindings push the change to all consumers.

yuzliu: I see, I got your concern about binding. I was thinking from a user's point of view. Users may have a ML pipeline that process data -> train model -> save model to S3 -> create TrainedModel CR. It seems to me that one model can be deployed to multiple model servers and we shouldn't require users to have a model server running before they can create TrainedModel CRs. 

We can think about it more, it does not need to be part of this PR.

config/configmap/inferenceservice.yaml
yuzliu: Yeah, I didn't find any multi-model serving information from onnx's documentation as well. But this list only advertise the supported frameworks. I added one more boolean field "multiModelServer" to indicate if this server can support multiple models.

config/configmap/inferenceservice.yaml
cliveseldon: So when people create an InferenceService they can optionally add an explainer, then, they can add explainer TrainedModels afterwards? The explainer types for Alibi are AnchorImages, AnchorText, so are <algorithmn>-<input-type>. At present the `gcr.io/kfserving/alibi-explainer` image handles just 1 explainer type and model. Maybe I need to understand how we expect the first version to work with a loaded explainer model and work backwards.

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,33 @@
...
+	Status            TrainedModelStatus `json:"status,omitempty"`
+}
+type TrainedModelSpec struct {
+	// Required field for parent inference service
+	InferenceService string `json:"inferenceService"`
+	// Predictor model spec
+	PredictorModel ModelSpec `json:"predictorModel"`
+	// Explainer model spec
+	ExplainerModel ModelSpec `json:"explainerModel,omitempty"`
+}
```
cliveseldon, line 21: Is this a 1-of-n semantic? So you can just supply either a PredictorModel OR an ExplainerModel but not both? How does this match up to InferenceService which has both?

Maybe the best is to leave out ExplainerModel and you need to define it as now on the inference service so you can only have 1 model artifact with an explainer. The AlibiExplainerSpec already as a StorageURI in its spec.

pkg/apis/serving/v1beta1/trained_model.go
cliveseldon: See above lets decide whether for the first version you have Explaner TrainedModel specs.

config/configmap/inferenceservice.yaml
yuzliu: That's how I think it should work. If the parent InferenceService has an explainer, then the TrainedModel CR should also container an explainer section. Otherwise this TrainedModel CR should be rejected. 

Which explainer type gcr.io/kfserving/alibi-explainer handles at present?

cliveseldon: Alibi handle all 3 text, image, tabular. But the type of the algorithm will need to match,e .g. AnchorText for Text. Also there is the issue of whether the model is classifier, regression see https://docs.seldon.io/projects/alibi/en/latest/overview/algorithms.html#model-explanations for the matrix map of this to available algorithm.

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: I can remove the ExplainerModel for now with the assumption that different predictors need to share the same explainer which is defined in the parent InferenceService.

yuzisun: @cliveseldon actually there is routing problem, when you call `v1/models/model-a:explain` and `v1/models/model-b:explain` will explainer understand it is trying to reuse the same explainer model?

cliveseldon: True. At present the predictorUrl is injected as a single parameter. So you would need a solution to allow 1 explainer to call N models which are assumed to all behave similarly dependant on the request it receives. So looks like som refactoring of the present explainer python server is needed.

ifilonenko:
> > Anyone has ideas about why this unit test timed out? All the unit tests passed locally, it seems that something blocked the ksvc.
> > ```
> > --- FAIL: TestInferenceServiceWithOnlyPredictor (31.53s)
> >    controller_test.go:379:
> >        Timed out after 30.000s.
> >        Expected
> >            <bool>: false
> >        to be true
> > FAIL
> > ```
> 
> @ifilonenko I run into this in my PR too..

meeep, these events should be generating so maybe there is something flaky with the test environment

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,33 @@
...
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Machine Learning <framework name>:<git tag>
+	// The values could be: "tensorflow:v2.2.0","pytorch:v1.5.1","sklearn:0.23.1","onnx:v1.7.0","xgboost:v1.1.1", "myawesomeinternalframework:1.1.0" etc.
+	Framework string `json:"framework"`
+	// Machine Learning model type, this field is used to match explainer type
+	// +kubebuilder:webhooks:Enum={"tabular","text","image"}
+	Type string `json:"type,omitempty"`
+	//Defaults to requests and limits of 1CPU, 2Gb MEM.
+	Resources v1.ResourceRequirements `json:"resources,omitempty"`
```
ellistarn, line 32: Should we even default to 1CPU requested? As I understand it, memory requests make sense here and not much else. Should we simply make this memory rather than full resource requirements?

cliveseldon: I discussed with @yuzisun whether limits are safe as Tensorflow will by default try to use all cores and you might get CPU throttling if you limit to 1CPU. Maybe we should at least remove limits and users can add these as needed? Not sure whether we also want to set:
```
--tensorflow_intra_op_parallelism= 
--tensorflow_inter_op_parallelism=
```
But this is a separate issue.

yuzliu: > Should we even default to 1CPU requested? As I understand it, memory requests make sense here and not much else. Should we simply make this memory rather than full resource requirements?

I think you are right, I am changing it to only has memory now.

yuzliu: > I discussed with @yuzisun whether limits are safe as Tensorflow will by default try to use all cores and you might get CPU throttling if you limit to 1CPU. Maybe we should at least remove limits and users can add these as needed? Not sure whether we also want to set:
> 
> ```
> --tensorflow_intra_op_parallelism= 
> --tensorflow_inter_op_parallelism=
> ```
> 
> But this is a separate issue.

Clive, this resource field is meant to be the amount of resource will be needed by a model. I think users can only estimate memory usage when they create the TrainedModel CR. I will change it to only has memory.

pkg/apis/serving/v1beta1/trained_model.go
yuzliu: I removed the ExplainerModel and I think we should also discuss wether Transformer and Explainer should be an independent CRD. In the world of multi-model inference, each model should be able to choose its transformer and explainer independently.

config/crd/serving.kubeflow.org_inferencerouters.yaml
```diff
@@ -0,0 +1,128 @@
...
+    type: string
+  - JSONPath: .status.conditions[?(@.type=='Ready')].status
+    name: Ready
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
+  group: serving.kubeflow.org
+  names:
+    kind: InferenceRouter
```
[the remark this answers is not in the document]
yuzliu: The CRD is auto generated from make manifest: https://github.com/kubeflow/kfserving/blob/v1beta1/Makefile#L97 I believe it comes from the new InferenceRouter CRD: https://github.com/kubeflow/kfserving/blob/v1beta1/pkg/apis/serving/v1beta1/inference_router.go#L9

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,28 @@
...
+	PredictorModel ModelSpec `json:"predictorModel"`
+}
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Machine Learning <framework name>:<git tag>
+	// The values could be: "tensorflow:v2.2.0","pytorch:v1.5.1","sklearn:0.23.1","onnx:v1.7.0","xgboost:v1.1.1", "myawesomeinternalframework:1.1.0" etc.
+	Framework string `json:"framework"`
+	// Maximum memory this model will consume, this field is used to decide if a model server has enough memory to load this model.
+	Memory v1.ResourceName `json:"memory"`
```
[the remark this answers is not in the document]
yuzliu: I removed ModelVersionPolicy section for now per @ellistarn's suggestion since we should take some time to gain more experience about model versioning. I added this section initially because I think the new MLServer we are implementing should follow Triton's model repo's structure and the model version policy, see my earlier reply here: https://github.com/kubeflow/kfserving/pull/908#discussion_r447124002.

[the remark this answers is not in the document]
yuzliu: Interesting idea to make
>  TrainedModelSpec.InferenceService was made optional/mutable
I also proposed an idea to create a TrainedModelBinding CRD to bind TrainedModel and InferenceService. But it also means that we allow users to create TrainedModel CR without deploying. Per @yuzisun's comment earlier, if we completely decouple TrainedModel and InferenceService CRD then KFServing may become a staging area for un-deployed models. I don't want to make TrainedModelSpec.InferenceService was made optional/mutable as optional for now to avoid turning KFServing into a TrainedModel staging area. However, I think it is interesting to think about how to manage a model's life cycle.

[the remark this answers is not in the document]
yuzliu: About how to get a precise memory usage of model, I think that's something we can add to MLServer's metadata endpoint @cliveseldon wdyt?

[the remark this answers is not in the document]
njhill: Sorry @yuzliu I guess you put responses to all of my comments on this one because the others were on outdated changes. Will reply to all of them here but quote inline...
> I added this section initially because I think the new MLServer we are implementing should follow Triton's model repo's structure and the model version policy, see my earlier reply here: #908 (comment).

Yep probably best for versioning discussion to happen elsewhere, glad that this part was deferred.

> I also proposed an idea to create a TrainedModelBinding CRD to bind TrainedModel and InferenceService. But it also means that we allow users to create TrainedModel CR without deploying. Per @yuzisun's comment earlier, if we completely decouple TrainedModel and InferenceService CRD then KFServing may become a staging area for un-deployed models. I don't want to make TrainedModelSpec.InferenceService was made optional/mutable as optional for now to avoid turning KFServing into a TrainedModel staging area. However, I think it is interesting to think about how to manage a model's life cycle.

Yep I completely agree w.r.t. this separation of concerns. My idea was just a suggested compromise in response to your TrainedModelBinding proposal and motivation, but I think I'm equally happy with the tight coupling.

> About how to get a precise memory usage of model, I think that's something we can add to MLServer's metadata endpoint

That seems like a reasonable approach to me ... having a controller use that to update the `TrainedModel` Status. Though it might be considered more like operational state than model metadata (since memory use might vary between model server versions for example, and may not be known if that particular model isn't loaded at that point in time) so might be more appropriate to go in `ModelReadyResponse`?

ellistarn:
Love how this is shaping up. Great comments as well from @njhill. Let's start iterating! /approve

yuzliu:
Thanks @ellistarn , Dan removed the flaky test so I can pass CI, let me just do a rebase.

PR #918, opened 2020-07-04
Run cluster setup in separate Argo stage

yuzisun:
Cluster setup step might be cpu intensive which affects the unit test step that is running in parallel at the same stage,
run cluster setup as separate stage to avoid unit test `eventually` check timeouts.

Fixes #

**Release note**:

```release-note

```


yuzisun:
fixed in the other PR

PR #921, opened 2020-07-05
update custom/grpc-service sample

Iamlovingit:
update custom/grpc-service sample

* add custom.container.name field to slove the error
`ValidationError(InferenceService.spec.default.predictor.custom.container): missing required field "name" in org.kubeflow.serving.v1alpha2.InferenceService.spec.default.predictor.custom.container`

* change the way of `Get the inference service address:`.
 `kubectl get inferenceservice grpc-service` instead of `kubectl get ksvc grpc-service-predictor-default`

### DIFF
--- a/docs/samples/custom/grpc-server/README.md
+++ b/docs/samples/custom/grpc-server/README.md
@@ -34,6 +34,7 @@ spec:
     predictor:
       custom:
         container:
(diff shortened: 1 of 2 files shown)


yuzisun:
Looks great! 

/lgtm
/approve

PR #925, opened 2020-07-07
Add LivenessProbe to Tensorflow Predictor

salanki:
TensorFlow can in some cases, at least on GPU, end up in a OOM condition at runtime that it never recovers from. This means that each request to the predictor ends up failing. We're observing this sporadically with the 11B T5 model on 16GB GPUs. This shouldn't really happen, but something inside either the model or TensorFlow after several hours of runtime can decide to allocate more input tensors than there is GPU memory.

To manage this, a `LivenessProbe` is added on the Tensorflow Predictor. The probe is fairly generous, requiring three consecutive failures 10 seconds apart.

A constant for the `InitialDelaySeconds` is defined and set to 180s. This is longer than KNative allows a revision to deploy in KNative < 15.0, so it should be safe for every current user of KFServing. KNative 15.0 allows the `progressDeadline` to be configured above the default of 120s so hardcoding it to 180s is not necessarily future proof. 180s is quite a while though to load a model, but it can be debated if it is a good idea to add a hardcoded time limit.

**Release note**
```release-note
Liveness probes will now be configured on TensorFlow predictors. A TensorFlow instance will be declared dead and restarted if it doesn't respond to its internal models metadata endpoint for 30s.
```

### DIFF
--- a/pkg/apis/serving/v1alpha2/framework_tensorflow.go
+++ b/pkg/apis/serving/v1alpha2/framework_tensorflow.go
@@ -49,11 +49,23 @@ func (t *TensorflowSpec) GetContainer(modelName string, parallelism int, config 
 	}
 
 	return &v1.Container{
-		Image:     config.Predictors.Tensorflow.ContainerImage + ":" + t.RuntimeVersion,
-		Name:      constants.InferenceServiceContainerName,
-		Command:   []string{TensorflowEntrypointCommand},
-		Resources: t.Resources,
-		Args:      arguments,
+		Image:     		config.Predictors.Tensorflow.ContainerImage + ":" + t.RuntimeVersion,
+		Name:      		constants.InferenceServiceContainerName,
+		Command:   		[]string{TensorflowEntrypointCommand},
+		Resources: 		t.Resources,
+		Args:      		arguments,
+		LivenessProbe:  &v1.Probe{
+			Handler: v1.Handler {
+				HTTPGet: &v1.HTTPGetAction{
+					Path: "/v1/models/" + modelName,
+				},
+			},
+			InitialDelaySeconds: 180,
+			PeriodSeconds: 10,
+			FailureThreshold: 3,
+			SuccessThreshold: 1,
+			TimeoutSeconds: 1,
+		},
 	}
 }
 
--- a/pkg/apis/serving/v1alpha2/framework_tensorflow.go
+++ b/pkg/apis/serving/v1alpha2/framework_tensorflow.go
@@ -49,22 +49,22 @@ func (t *TensorflowSpec) GetContainer(modelName string, parallelism int, config 
 	}
 
 	return &v1.Container{
-		Image:     		config.Predictors.Tensorflow.ContainerImage + ":" + t.RuntimeVersion,
-		Name:      		constants.InferenceServiceContainerName,
-		Command:   		[]string{TensorflowEntrypointCommand},
-		Resources: 		t.Resources,
-		Args:      		arguments,
-		LivenessProbe:  &v1.Probe{
-			Handler: v1.Handler {
+		Image:     config.Predictors.Tensorflow.ContainerImage + ":" + t.RuntimeVersion,
+		Name:      constants.InferenceServiceContainerName,
+		Command:   []string{TensorflowEntrypointCommand},
+		Resources: t.Resources,
+		Args:      arguments,
+		LivenessProbe: &v1.Probe{
+			Handler: v1.Handler{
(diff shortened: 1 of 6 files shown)


pkg/apis/serving/v1alpha2/framework_tensorflow.go
```diff
@@ -54,6 +54,18 @@ func (t *TensorflowSpec) GetContainer(modelName string, parallelism int, config
 		Command:   []string{TensorflowEntrypointCommand},
 		Resources: t.Resources,
 		Args:      arguments,
+		LivenessProbe: &v1.Probe{
+			Handler: v1.Handler{
+				HTTPGet: &v1.HTTPGetAction{
+					Path: "/v1/models/" + modelName,
```
ellistarn, line 60: Is it possible that if a model load takes ~10 minutes, liveness could kill the pod forever?

Would be great to support healthz https://github.com/tensorflow/serving/issues/671. In the multimodel case, there may not be a model server, but the server is happy and healthy.

salanki: Correct. As per the PR description, you cannot currently use KFServing with models that take longer than 2minutes to load. The question is now that KNative is making it configurable, is 180s going to bite anyone, or do we need to make it configurable?

yuzisun: @salanki @ellistarn In v1beta1 user can configure the custom liveness probe if they are not happy with the defaults since we allow container fields even for prepackaged model servers.  I suggest here do not use `/v1/models` because that won't work with multi model case @yuzliu, does tensorflow have server liveness probe endpoint?

salanki: According to the issue @ellistarn linked, there is no real health endpoint yet. There might be a better REST endpoint to pull that works for multi model as well though. Looking for input from someone with multi model experience.

yuzliu: Hi, I like the idea to add a liveness check, but it should be on the model server level not the model level since we will allow users to load multiple models soon. Can we use it as an opportunity to push tfserving to add a health check endpoint?

yuzisun: I think a simple tcp check can also work here for the liveness probe?

salanki: Do you know if tcp probes work with KNative? It might work, but by checking a model endpoint we can know if TF thinks that model is alive or not.

yuzisun: tcp probe should work as their default readiness probe is tcp check. Definitely agree that model endpoint is better, but we will need to check if it is single model case or multi model.

ellistarn: +1 to tcp check on TFServing. I've done this in a past life and platform ;).

@salanki can you configure this to 10 minutes? I think having a model take 10 minutes to load is more common than this failure mode. A worthwhile tradeoff?

Also, as per Dan's comment, just because we enable passthrough of probes doesn't mean we won't provide defaults. Because of this weird tensorflow issue, we probably want to give users a sane default (like 10 minutes) and they can override if they need.

Will lgtm with a higher value, but I /approve of this change.

salanki: This sounds reasonable, will update.

salanki: Done

salanki: Is multi model serving coming in the nexts release? Ie. do we need to address it here, or can we merge this as is and figure out how to best handle multi model for next release?

yuzliu: MMS won't be in this release, I think we can address it in next release.

yuzisun: For MMS case I think we will want tcp check because we do not want to restart server just because one of the models is not in ready state, we may end up with awkward if/else case.

salanki: I’ll run some tests and we can keep discussing it in another issue for MMS.

yuzisun:
We will figure out how to handle MMS case in v1beta1, it is good for now.

/lgtm
/approve

PR #935, opened 2020-07-12
fix config map format

Iamlovingit:
fix config map format, remove the comma at end of line.


yuzisun:
thanks @Iamlovingit !

/lgtm
/approve

PR #936, opened 2020-07-12
Migrate controller tests to use BDD testing style

yuzisun:
Migrate to use the kubebuilder BDD testing style

Fixes #

1. This is the refactoring to prepare for v1beta1 PR.

**Release note**:

```release-note

```


pkg/testing/envtest_setup.go
```diff
@@ -43,6 +44,7 @@ func SetupEnvTest() *envtest.Environment {
 			filepath.Join("..", "..", "..", "config", "crd"),
 			filepath.Join("..", "..", "..", "test", "crds"),
 		},
+		UseExistingCluster: &useExistingCluster,
 	}
```
ellistarn, line 48: You can use proto.bool for quick primitive booleans

pkg/controller/inferenceservice/controller.go
```diff
@@ -51,100 +50,38 @@ import (
...
 
-var _ reconcile.Reconciler = &ReconcileService{}
+var _ reconcile.Reconciler = &InferenceServiceReconciler{}
 
 // ReconcileService reconciles a Service object
-type ReconcileService struct {
+type InferenceServiceReconciler struct {
 	client.Client
-	scheme   *runtime.Scheme
+	Log      logr.Logger
```
ellistarn, line 69: What's the goal with this new logging pattern

yuzisun: to void the global logger variable which can cause conflicts within package

pkg/testing/envtest_setup.go
yuzisun: oh nice! I have been looking for a quick way to do for long time!

PR #953, opened 2020-07-21
Update GRPC protocol to have raw tensor data at top-level of message

deadeyegoodwin:
This allows a more performant implementation due to the way protobuf
clearing and reuse operates in conjunction with GRPC APIs.

Fixes #

**Release note**:

```release-note

```


deadeyegoodwin:
/assign @ellistarn 

Ellis, this is the performance change to the grpc protocol that I mentioned earlier.

ellistarn:
@rakelkar can you get ONNXRuntime in touch here?

docs/predict-api/v2/required_api.md
```diff
@@ -733,64 +769,64 @@ one-dimensional, row-major order of the tensor elements. Element
...
-      // between the elements. Note that the FP16 data type must be
-      // represented as raw content as there is no standard support for a
-      // 16-bit float type.
-      bytes raw_contents = 1;
-
       // Representation for BOOL data type. The size must match what is
       // expected by the tensor's shape. The contents must be the flattened,
       // one-dimensional, row-major order of the tensor elements.
-      repeated bool bool_contents = 2;
+      repeated bool bool_contents = 1;
```
pugangxa, line 793: So this is just proposal and not implemented by anyone right? I'm not familiar with this part of work, just think we should not change the number field once finalized to avoid the compatibility issues.

deadeyegoodwin: Triton implements it as a beta so small changes like this are ok for it.

pugangxa: Got it, thanks.

ellistarn:
Apologies. I just rewrote history to fix the contributors list. This PR needs a rebase before it can be reopened.

PR #954, opened 2020-07-22
Avoid duplicate hostnames in VirtualService when running in svc.cluster.local domain

akirillov:
When KNative Serving default domain is set to be cluster-local by default (i.e. `svc.cluster.local`), the external host domain is not different from the internal one which creates duplicate hostnames in [virtualservice.go#L278](https://github.com/kubeflow/kfserving/blob/master/pkg/controller/inferenceservice/resources/istio/virtualservice.go#L278) and the created `VirtualService` fails Istio validation with:
```
admission webhook \"pilot.validation.istio.io\" denied the request: configuration is invalid:
duplicate hosts in virtual service: mnist.test.svc.cluster.local & mnist.test.svc.cluster.local
```

This PR adds a check in `virtualservice.go` to avoid duplicates when running cluster-local services.

Fixes https://github.com/kubeflow/kfserving/issues/948


pugangxa:
Straightforward to me, thanks for the fix.

pkg/controller/inferenceservice/resources/istio/virtualservice.go
```diff
@@ -266,6 +266,14 @@ func (r *VirtualServiceBuilder) CreateVirtualService(isvc *v1alpha2.InferenceSer
 	// extract the virtual service hostname from the predictor hostname
 	serviceURL := fmt.Sprintf("%s://%s", "http", serviceHostname)
 
+	serviceInternalHostName := network.GetServiceHostname(isvc.Name, isvc.Namespace)
+	hosts := []string{
+		serviceHostname,
```
yuzisun, line 271: I think the default host should be `serviceInternalHostName`, there is another case we should take in account here is that cluster local service can be also configured at service level with label `serving.knative.dev/visibility=cluster-local` according to [the doc](https://knative.dev/docs/serving/cluster-local-route). So I'd suggest the following change

```
	serviceInternalHostName := network.GetServiceHostname(isvc.Name, isvc.Namespace)
	hosts := []string{
		serviceInternalHostName,
        }
        isClusterLocal := false
        if val, ok := isvc.metadata.labels["serving.knative.dev/visibility"]; ok {
              if val == "cluster-local" {
                     isClusterLocal = true
              } 
        }
        if serviceHostname != serviceInternalHostName && !isClusterLocal {
		hosts = append(hosts, serviceHostname)
	}
```

pkg/controller/inferenceservice/resources/istio/virtualservice_test.go
```diff
@@ -60,6 +64,24 @@ func TestCreateVirtualService(t *testing.T) {
...
 		},
 	}
+	predictorClusterLocalRouteMatch := []*istiov1alpha3.HTTPMatchRequest{
+		{
+			Authority: &istiov1alpha3.StringMatch{
+				MatchType: &istiov1alpha3.StringMatch_Regex{
+					Regex: constants.HostRegExp(constants.InferenceServiceHostName(serviceName, namespace, clusterLocalDomain)),
+				},
+			},
+			Gateways: []string{constants.KnativeIngressGateway},
```
yuzisun, line 74: I think we need to remove the ingress gateway route match if it is configured with cluster local

ellistarn:
Apologies. I just rewrote history to fix the contributors list. This PR needs a rebase before it can be reopened.

akirillov:
@ellistarn, I rebased and pushed the updated branch. Shall I open a separate PR or can you please reopen this one?

pkg/controller/inferenceservice/resources/istio/virtualservice.go
akirillov: Thanks for the pointers, I totally missed the possibility to use labels for that. Updated the code and tests.

yuzisun:
@akirillov please  help recreate the PR

pkg/controller/inferenceservice/resources/istio/virtualservice_test.go
akirillov: This could be another issue because both KNative gateways (ingress and local) are added by default in [virtualservice.go#L240-L257](https://github.com/kubeflow/kfserving/blob/master/pkg/controller/inferenceservice/resources/istio/virtualservice.go#L240-L257).

akirillov:
@yuzisun, here's a new PR: https://github.com/kubeflow/kfserving/pull/1017

PR #963, opened 2020-07-23
Update KNative annotation when modifying minReplicas to 0

### LINKED ISSUE #962: Services cannot be modified to scale-to-zero after being created without it
/kind bug

A `InferenceService` that is first set to have `minReplicas` > 0, and later changed to have `minReplicas` = 0 will never have the KNative Service annotation updated. See https://github.com/kubeflow/kfserving/pull/655#discussion_r459197929.

The annotation should be set to `0` explicitly.

**Environment:**

- Knative Version: 1.16
- KFServing Version: master

salanki:
As described in #962. The `minScale` annotation is never set to 0 when updating an existing `InferenceService`'s `minReplicas` from a positive number to zero. This will give the user the impression of scale-to-zero not working, and leave the KFServing administrator pulling his hair trying to figure out how KNative scale to zero can be broken for only this specific `InferenceService`.

Fixes #962

**Release note**:
```release-note
Fixes a bug where enabling Scale-To-Zero on an existing InferenceService did not take effect.
```


yuzisun:
good catch @salanki !

/lgtm
/approve

PR #978, opened 2020-07-28
Generate SDK for 0.4 release

yuzisun:
Generate SDK for 0.4 release

Fixes #

1. Fixed a path issue for update-openapigen.sh after KFServing migrated to use GO Module

**Release note**:

```release-note

```


python/kfserving/docs/V1alpha2InferenceService.md
```diff
@@ -3,9 +3,9 @@
 ## Properties
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**api_version** | **str** | APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources | [optional] 
-**kind** | **str** | Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds | [optional] 
-**metadata** | [**V1ObjectMeta**](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ObjectMeta.md) |  | [optional]
```
jinchihe, line 8: The link should not be updated, that's need to link to k8s python api docs. Thanks.

python/kfserving/docs/KnativeVolatileTime.md
```diff
@@ -3,7 +3,7 @@
 ## Properties
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**inner** | [**V1Time**](V1Time.md) |  |
```
jinchihe, line 6: Need to double confirm here, is that caused by knative changes? if so, seems the V1Time.md and related code no longer need?

jinchihe:
@yuzisun Would you please also update the `author_email` to correct Ellis Email and add yours if you are willing to do it? Thanks.

https://github.com/kubeflow/kfserving/blob/2e3cfe048e7bcda94b5069c3c3d6dc5efbbf5102/python/kfserving/setup.py#L30

python/kfserving/docs/KnativeVolatileTime.md
jinchihe: Seems that's updated when we generate 0.3 version in https://github.com/kubeflow/kfserving/pull/731, so I think the V1Time.md and v1_time.py is useless, right? of cause you can choose to keep that for safe considering. Thanks.

yuzisun: yes, we can delete delete v1_time.py now, the field should be just a datetime type. Knative creates VolatileTime mainly because they want to make it semantically equal when comparing values https://godoc.org/knative.dev/pkg/apis#VolatileTime.

python/kfserving/docs/V1alpha2InferenceService.md
yuzisun: @jinchihe I did not update this, any idea why the generator did not generate the link?

PR #989, opened 2020-08-01
[POC] Multi-Model Puller

ifilonenko:
This PR is needed when trying to define the multi-model story. The agent is responsible for pulling models into a model directory that the model server will use to host.

Fixes https://github.com/kubeflow/kfserving/issues/1040

**Design:**

We assume the ConfigMap looks like this: 
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: models-config
data:
  models.json: |
    [
      {
        "modelName": "model1",
        "modelSpec": {
          "storageUri": "s3://example-bucket/path/to/model1",
          "framework": "sklearn",
          "memory": "1G"
        }
      },
      {
        "modelName": "model2",
        "modelSpec": {
          "storageUri": "s3://example-bucket/path/to/model2",
          "framework": "sklearn",
          "memory": "1G"
        }
      }
    ]
```
as such `model_name1.json` and `model_name2.json` will each be files in the directory into which `models-config` will be mounted. This puller will watch (or pull) form this directory should any models be added or removed from the cmap, and send the appropriate request to the load / unload methods. 

Tested by running: 

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: puller
spec:
  serviceAccountName: default
  restartPolicy: Never
  containers:
  - name: job
    image: <my_image>
    imagePullPolicy: Always
    env:
    - name: AWS_SECRET_KEY
      value: <my_key>
    - name: AWS_ACCESS_KEY
      value: <my_other_key>
    resources:
      limits:
        cpu: "1"
        memory: 1Gi
      requests:
        cpu: "1"
        memory: 1Gi
    command:
    - /agent
    args:
    - -config-dir
    - /mnt/configs
    - -model-dir
    - /tmp/models
    - -num-workers
    - "1"
    - -s3-endpoint
    - "<my_endpoint>"
    volumeMounts:
    - name: config-volume
      mountPath: /mnt/configs
    - name: model-volume
      mountPath: /tmp/models
  volumes:
    - name: config-volume
      configMap:
        name: models-config
    - name: model-volume
      emptyDir: {}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: models-config
data:
```

by modifying the data with the addition of `model1`, the addition of `model2`:
```
2020/08/17 22:55:03 Entering watch
====
2020/08/19 17:55:10 Name: model1 Spec: {s3://platform-test/flowers/ sklearn {{1 9} {<nil>} 1G DecimalSI}}
2020/08/19 17:55:10 Name: model2 Spec: {s3://platform-test/model.conf sklearn {{1 9} {<nil>} 1G DecimalSI}}
2020/08/19 17:55:10 Need to add model model1
2020/08/19 17:55:10 Sending event {0xc0003141e0 Load true}
2020/08/19 17:55:10 worker 1 for model1 is initialized
2020/08/19 17:55:10 worker 1 model1 started  job {0xc0003141e0 Load true}
2020/08/19 17:55:10 Should download s3://platform-test/flowers/
2020/08/19 17:55:10 Processing: s3://platform-test/flowers/ = 2551120587
2020/08/19 17:55:10 Downloading:  s3://platform-test/flowers/
2020/08/19 17:55:10 Need to add model model2
2020/08/19 17:55:10 Sending event {0xc0003142a0 Load true}
2020/08/19 17:55:10 worker 1 for model2 is initialized
2020/08/19 17:55:10 worker 1 model2 started  job {0xc0003142a0 Load true}
2020/08/19 17:55:10 Should download s3://platform-test/model.conf
2020/08/19 17:55:10 Processing: s3://platform-test/model.conf = 3669683301
2020/08/19 17:55:10 Downloading:  s3://platform-test/model.conf
2020/08/19 17:55:10 Now doing a request on {0xc0003142a0 Load true}
2020/08/19 17:55:16 Now doing a request on {0xc0003141e0 Load true}
```
- [x] manual tests

**Release note**:
```release-note
The initial skeleton of the side-car agent
```


ifilonenko:
@dsun @ellistarn @yuzliu for review

cmd/puller/main.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+var (
+	modelDir = flag.String("model_dir", "/tmp","directory for multi-model config files")
+)
+
+func main() {
+	flag.Parse()
+	puller.OnConfigChange(func(e puller.EventWrapper) {
+		log.Println("Send a request to:", e.LoadState)
```
ifilonenko, line 16: This section will do a load / unload call.

EventWrapper will have information pertaining to:
1. Which model
2. Whether load / unload
3. What information should it be a load

ellistarn:
> We assume the ConfigMap (which will be specified in a separate PR) looks like this:
> 
> ```yaml
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: models-config
>   namespace: <user-namespace>
> data:
>   model_name1.json |
>   - storageUri: xxx
>   - framework: xxx
>   - memory: xxx
> ```

For framework, is this string going relate at all to the image in the Predictor spec, i.e. "tensorflow:v1.12", "patched_tensorflow:v1.12" or will it simply be a well known enum of supported MMS frameworks?

cmd/puller/main.go
```diff
@@ -0,0 +1,20 @@
...
+		log.Println("for model", e.ModelName)
+	})
+	puller.WatchConfig(*modelDir)
```
ellistarn, line 19: Should we use "informers" here? https://firehydrant.io/blog/stay-informed-with-kubernetes-informers/

pkg/puller/watcher.go
```diff
@@ -0,0 +1,126 @@
+package puller
+
+import (
+	"encoding/json"
+	"github.com/fsnotify/fsnotify"
```
ellistarn, line 5: file watchers are notorious for being brittle. I'm a little nervous about relying on this.

puller.Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Build the inference-puller binary
```
ellistarn, line 1: IMO, we should stop creating top level Dockerfiles. Perhaps we need something like
```
/cmd/manager/Dockerfile
/cmd/manager/main.go
/images/logger/Dockerfile
/images/logger/main.go
...
```

puller.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+FROM golang:1.13.0 as builder
```
ellistarn, line 2: We should move to 1.14 asap

pkg/puller/watcher.go
ifilonenko: Viper relies on that as well. But tbh, I’d prefer to have containers look at file-system and be k8s agnostic. Relying on api-server doesn’t sound necessarily better.

ifilonenko: Also, where do you see file watChers being brittle? I purposefully decided to use it instead of configmap informer. My local testing produced good results

puller.Dockerfile
ifilonenko: Sure, lmk how you would prefer to structure it. I was mimicking the previous side-car implementations

puller.Dockerfile
ifilonenko: No worries there, should I do that in this PR? Or is that for another time

ifilonenko:
> For framework, is this string going relate at all to the image in the Predictor spec, i.e. "tensorflow:v1.12", "patched_tensorflow:v1.12" or will it simply be a well known enum of supported MMS frameworks?

I was thinking of doing well-known enums, personally. But the nuances of that decision will be in a separate PR, so we can move that discuss to there. This PR will rely on that one.

pkg/puller/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type State string
+
+const (
+	// State Related
+	ShouldLoad State = "Load"
+	ShouldUnload State = "Unload"
+
+ 	writeOrCreateMask = fsnotify.Write | fsnotify.Create
```
yuzliu, line 45: What does the writeOrCreateMask state mean? Why is it different from the ShouldLoad state?

cmd/puller/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"flag"
+	"github.com/kubeflow/kfserving/pkg/puller"
+	"log"
+)
+
+var (
+	modelDir = flag.String("model_dir", "/tmp","directory for multi-model config files")
```
yuzliu, line 10: Can this variable be moved to [here](https://github.com/kubeflow/kfserving/blob/master/pkg/constants/constants.go), so InferenceService's controller can mount the configMap to the same location for MMS InferenceService?

pkg/puller/watcher.go
yuzliu: nvm, I saw how it is used later  👍

pkg/puller/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				// we only care about the model file:
+				// 1 - if the model file was modified or created
+				// 2 - if the model file was removed as a result of deletion or renaming
+				if p.onConfigChange != nil {
+					ext := filepath.Ext(event.Name)
+					isEdit := event.Op&writeOrCreateMask != 0
```
yuzliu, line 84: Using bitwise & AND makes me nervous here, let's me sure we have good tests for it :)

pkg/puller/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+					isRemove := event.Op&fsnotify.Remove != 0
+					isValidFile := ext == p.fileExtension
+					if isValidFile && (isEdit || isRemove) {
+						fileName := strings.TrimSuffix(filepath.Base(event.Name), ext)
+						if isRemove {
```
yuzliu, line 89: 👍 I like that we have a file for each model in the configMap. We don't need to do a diff on every ConfigMap change which makes the logic cleaner and simpler.

yuzliu: We can add version information in the filename e.g. <modelName>-<version> once we are ready to add VersionPolicy in the TrainedModel CRD.

ifilonenko: Do you think that can be contained in content of the `json` instead of the `filename`? although, we are getting a head of ourselves, that will be for a later PR :)

cmd/puller/main.go
ifilonenko: hmm so this is a flag variable, so its specific to the controller. I believe we should leave it here. The variable that we use will be specified when defining the controller. 

i.e. 
```yaml
- args:
- --model_dir
- <my_shared_variable>
```

puller.Dockerfile
ellistarn: For go, let's try putting them with the main. Our python images are already this way. This way, each cmd/foo is effectively it's own package and can just have a Dockerfile.

ifilonenko: Sure I can move it over, on it

pkg/puller/watcher.go
yuzliu: Yeah we can discuss the details once we add version policy.

cmd/puller/main.go
yuzisun, line 10: Can we default to `/mnt/models` ?

cmd/puller/main.go
yuzisun: @ellistarn We'd like to not depend on k8s for model puller, likewise for model server.

cmd/puller/main.go
ifilonenko: These are config files, I thought we would default `/mnt/models` for model files, but configs would go into a different location. Lmk where you prefer

pkg/puller/downloader.go
```diff
@@ -0,0 +1,28 @@
...
+	"gopkg.in/matryer/try.v1"
+	"log"
+	"time"
+)
+
+func DownloadModel(numRetries int, event EventWrapper) error {
+	err := try.Do(func(attempt int) (bool, error) {
+		err := download(event.ModelDef.StorageUri)
+		if err != nil {
+			time.Sleep(1 * time.Second) // wait a second
```
yuzliu, line 14: Why do we wait for a second here?

cmd/puller/main.go
```diff
@@ -0,0 +1,23 @@
...
+	numWorkers = flag.Int("num_workers", 1,"number of workers for parallel downloads")
```
yuzisun, line 11: the convention for argument is `num-workers`

pkg/puller/downloader.go
ifilonenko: Just a dummy thing for retry purposes, I don’t mind if we retry immediately. It’s a WIP

pkg/puller/puller.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func AddModelToChannel(e EventWrapper) {
+	p.Channel <- e
+}
+
+func InitiatePullers(numWorkers int, numRetries int) {
+	p.NumWorkers = numWorkers
+	p.NumRetries = numRetries
+	p.Channel = make(chan EventWrapper)
```
yuzisun, line 57: We need buffer size here otherwise it is going to block the sender if receiver is still consuming the event.

ifilonenko: Should the buffer be the number of workers?

yuzisun:
Great comments ! thanks @njhill !

>   * What happens if a model is removed while it is downloading? or added while a prior model with same name is still in the process of unloading?

All the events for the same model should be processed by the same worker, we can create a set of channels and assign a worker for each channel, so events for same models are consistent hashed to the same channel and processed by the same worker.

>   * How will changes to existing model definitions be handled in a controlled manner? In some model servers this may entail unloading the existing model followed by loading the new one. As an aside this is partly why I feel that it could be best for `TrainedModel`s to be immutable - maybe for another discussion...

I think this depends on how KFServing manages the  model version, definitely another big discussion.

> * What happens if the puller container dies/restarts and in the meantime models are removed from the configmap? There will be no watcher events for these deletions and so they would remain orphaned with files left on the shared volume and/or loaded in the model server. So an audit of the model data dir would be needed on startup to diff with the current configmap. In support of this I think a special file will need to be written alongside the model data to reflect download completion (along with source URI) and subsequently successfully loading into the model server.

Once the model puller restarts it should reconcile the model dir with the current configmap, yes I agree an audit trail file is needed to facilitate this reconciliation process. Each successful download should write a file with the storage uri there, if the file does not present or storage uri in the file is different from the one in the configmap, we should trigger a new download.

> * What happens if the model server dies/restarts? Depending on its control mechanism, some way of querying its current config might be needed to reconcile state otherwise lost models won't be reinstated. TF-Serving (and maybe others) sidestep this problem via use of a declarative manifest

TFServing can recover from the model configuration during restarts, for other model servers I think model puller need someway to detect the model server status and resync all the models, on the other hand trained model controller also resync the models  periodically which then triggers load/unload to the model server.

> * Related to that last point, we may want to consider both declarative and command-based cases w.r.t. interfacing with different model servers... load/unload events aren't very useful in the TF Serving case since it expects the full manifest to be amended and determines the deltas/actions itself.
> 
> What would make most sense imo is a level-triggered reconciliation loop analogous to a Kube controller, which compares/converges desired/actual states. Where desired state is the configmap mount dir, and actual state includes the model data shared volume contents, model server's own reported state, and any in-progress file download and/or deletion jobs.
> 
> Finally what's the current thinking w.r.t. reporting status back? Ideally the `TrainedModel` CR Status should reflect whether the model is loading / loaded / failed download / failed model-server load.
> 
Correct, we have planned this work to send model status probes from trained model controller so status can be reflected.

> BTW I'd be happy to help with this and apologies if some/all was already planned in some way...

cmd/puller/main.go
ellistarn: Let me know when there is a verdict on this.

njhill: I tend to agree with @yuzisun on this

njhill:
Thanks @yuzisun 
> All the events for the same model should be processed by the same worker, we can create a set of channels and assign a worker for each channel, so events for same models are consistent hashed to the same channel and processed by the same worker.

I'm not sure about the hashing approach. It would be functional but risk delaying work resulting in longer load times than necessary sometimes. There are use cases where the rate of model churn is quite high which are sensitive to total time it takes for a model to become available. For tfserving there is also a question of how best to trigger the config reloads.

> I think this depends on how KFServing manages the model version, definitely another big discussion.

Agree, please include me because I have some thoughts :)

> Correct, we have planned this work to send model status probes from trained model controller so status can be reflected.

Great, but curious from trainedmodel controller to who exactly? Imo this should come from the puller somehow rather directly from model server (whether that is push or pull) since there could be other kinds of failures related to downloading for example, and it will be keeping track of the model-server's state anyhow as mentioned in the model-server restart recovery point.

To me this seems analogous to how Pod CRs are manipulated in the context of scheduling to Nodes. Various actors contribute to the Pod's Status and update it directly IIRC (specifically the Conditions). In particular it's the kubelet that monitors and updates the running status of the Pod, and here I would suggest Puller/Model is kind of equivalent to Kubelet/Pod. But that would probably mean coupling the puller to Kube which I also agree seems to be a nice thing to avoid if possible.

yuzliu:
> > We assume the ConfigMap (which will be specified in a separate PR) looks like this:
> > ```yaml
> > apiVersion: v1
> > kind: ConfigMap
> > metadata:
> >   name: models-config
> >   namespace: <user-namespace>
> > data:
> >   model_name1.json |
> >   - storageUri: xxx
> >   - framework: xxx
> >   - memory: xxx
> > ```
> 
> For framework, is this string going relate at all to the image in the Predictor spec, i.e. "tensorflow:v1.12", "patched_tensorflow:v1.12" or will it simply be a well known enum of supported MMS frameworks?

@ellistarn, @yuzisun and are considering to remove version information from the ModelSpec's framework enum and from this model configuration file. The reason is that we can validate framework against different model servers, but validating if X model server supports Y framework's Z version is pretty difficult. But if we allow users to put ML framework version in the TrainedModel CR they will expect that KFServing can validate framework versions against model server versions properly.

cmd/puller/main.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	"flag"
+	"github.com/kubeflow/kfserving/pkg/puller"
+)
+
+var (
+	configDir  = flag.String("config-dir", "/mnt/configs", "directory for multi-model config files")
```
yuzliu, line 9:
```suggestion
	configDir  = flag.String("config-dir", "/mnt/configs", "directory for model config files")
```

According to the comments in other PRs we want to can the config model config instead of multi-model config. The rational is that we want to mount this sidecar and configmap to all the InferenceService pods so a single model InferenceService is really a syntactic sugar of MMS.

pkg/modelconfig/configmap.go
```diff
@@ -13,8 +13,8 @@ var logger = log.Log.WithName("ModelConfig")
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
```
yuzliu, line 14: Since Yao's PR is merged you can remove this file and use https://github.com/kubeflow/kfserving/tree/master/pkg/modelconfig

pkg/puller/downloader.go
```diff
@@ -0,0 +1,173 @@
...
+	ModelDir   string
+}
+
+type Protocol string
+
+const (
+	// Supported Protocols
+	S3    Protocol = "s3://"
+	GS    Protocol = "gs://"
+	PVC   Protocol = "pvc://"
```
yuzliu, line 32: I would suggest to only keep s3 and gs protocol for now and add other protocols when we do support them.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,173 @@
...
+func init() {
+	d = NewDownloader()
+}
+
+func NewDownloader() *Downloader {
+	d := new(Downloader)
+	return d
+}
+
+func SetEnvs(numRetries int, modelDir string) {
```
yuzliu, line 49: This function name indicate that it is setting an environment variable. Can we rename it to something less ambiguous ?

pkg/puller/downloader.go
```diff
@@ -0,0 +1,173 @@
...
+	d.ModelDir = modelDir
+}
+
+func DownloadModel(event EventWrapper) error {
+	modelSpec := event.ModelSpec
+	if modelSpec != nil {
+		modelUri := modelSpec.StorageURI
+		hashString := hash(modelUri)
+		log.Println("Processing:", modelUri, "=", hashString)
+		successFile := filepath.Join(d.ModelDir, fmt.Sprintf("SUCCESS.%d", hashString))
```
yuzliu, line 60: Should we just use the model file and append `.success` to it as a success indicator? My concern about using a separate success file is that it is more error prone when we update, delete and reload the same model multiple times.
Consider the situation model that the TrainedModel model-1 is deleted and then add back how do we make sure the success file still indicates the right status?

ifilonenko: It is based on the storageURI. If you upload a different file to the same storageURI, then that is the only thing that breaks the "caching" contract of the SUCCESS file. I am only hashing the storageURI, not the modelName.

pkg/modelconfig/configmap.go
ifilonenko: I am, as per https://github.com/kubeflow/kfserving/pull/989/files#diff-5b2482c98bf5c61625b6fb19a69b7477R87, these changes are just after running `make fmt vet`

pkg/puller/downloader.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+		}
+		file, err := create(fileName)
+		if err != nil {
+			return fmt.Errorf("file is already created: %v", err)
+		}
+		defer file.Close()
+
+		// TODO: Should the S3 client be shared?
+		sess, err := session.NewSession()
```
yuzliu, line 110: According to https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ each session refers to one set of access keys. If one InferenceServce's models all come from the same s3 tenancy we should be able to share it.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,173 @@
...
+		if err != nil {
+			return fmt.Errorf("unable to get new s3 session %v", err)
+		}
+		s3Svc := s3.New(sess)
+		downloader := s3manager.NewDownloaderWithClient(s3Svc, func(d *s3manager.Downloader) {
+			d.Concurrency = 1             // TODO: Override?
```
yuzliu, line 116: The default concurrency is 5: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/, any reason to not use the default setting?

pkg/puller/downloader.go
ifilonenko: Yeah, I am just unsure if it can handle concurrency. i.e. multiple go-routines calling the same session. For now, I will just leave it as is, if that is okay.

pkg/puller/downloader.go
ifilonenko: Yeah, I can leave default to 5, it just launches 5 go-routines for the download (I was thinking that there are a lot of go-routines so i wanted to keep it at a minimum, but I guess they are cheap. so i can leave it at 5, and have this value be overrideable?)

pkg/puller/puller.go
```diff
@@ -0,0 +1,76 @@
...
+	"log"
+)
+
+var (
+	p *Puller
+)
+
+type Puller struct {
+	// TODO: Should this be a syncmap?
+	ChannelMap map[string]Channel
```
yuzliu, line 13: I think we need to use sync.Map here: https://golang.org/pkg/sync/#Map not syncmap. syncmap is just a prototype for sync.Map. And yes, I think we should use sync.Map here. The puller has different goroutines to call AddModel and RemoveModel which can read and write the same map concurrently.

yuzliu: See the description of sync.Map 
``` Map is like a Go map[interface{}]interface{} but is safe for concurrent use by multiple goroutines without additional locking or coordination. Loads, stores, and deletes run in amortized constant time.

The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

The zero Map is empty and ready for use. A Map must not be copied after first use.
```

pkg/puller/downloader.go
yuzliu: What is a user create a TrainedModel with different name but use the same storageUri? The real question is what identifies a model. It seems that we are trying to use the storageUri to identity a model which doesn't sound right :(

pkg/puller/downloader.go
yuzliu: For the record ```
Sessions are safe to use concurrently as long as the session isn't being modified. The SDK doesn't modify the session once the session is created. Creating service clients concurrently from a shared session is safe.
```

File: pkg/puller/puller.go [hunk 21]
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type Channel struct {
+	EventChannel chan EventWrapper
+}
+
+func (*Puller) AddModel(modelName string, numWorkers int) Channel {
+	// TODO: Figure out the appropriate buffer-size for this
+	eventChannel := make(chan EventWrapper)
```
yuzliu, line 22: Check if a channel for this model exists first.

pkg/puller/puller.go
```diff
@@ -0,0 +1,76 @@
...
+	channel := Channel{
+		EventChannel: eventChannel,
+	}
+	// TODO: Figure out parallelism
+	for workers := 1; workers <= numWorkers; workers++ {
+		go worker(workers, modelName, channel.EventChannel)
```
yuzliu, line 28:
```suggestion
		go modelProcessor(workers, modelName, channel.EventChannel)
```
How about rename worker to modelProcessor so it is more specific?

pkg/puller/watcher.go
```diff
@@ -0,0 +1,184 @@
...
+					file, err := ioutil.ReadFile(filepath.Join(symlink, constants.ModelConfigFileName))
+					modelConfigs := make(modelconfig.ModelConfigs, 0)
+					if err != nil {
+						log.Println("Error in reading file", err)
+					}
+					err = json.Unmarshal([]byte(file), &modelConfigs)
+					if err != nil {
+						log.Println("unable to marshall for", event, "with error", err)
+					}
+					for _, modelConfig := range modelConfigs {
```
yuzliu, line 95: I think we need another loop that loops over the all the keys in ModelTracker and check it if is delete from the modelConfigs. It it is deleted, then remove from the ModelTracker.

pkg/puller/puller.go
ifilonenko: Ah I meant sync.Map when I wrote that. I just didn’t know if it was necessary to make it a sync.Map. As we are writing to the channel. Not quite convinced it’s necessary, but I can be easily persuaded :)

cmd/puller/puller.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN go mod download
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o puller ./cmd/puller
+
+# Copy the inference-puller into a thin image
+FROM gcr.io/distroless/static:latest
+COPY third_party/ third_party/
+WORKDIR /
+COPY --from=builder /go/src/github.com/kubeflow/kfserving/puller .
```
ellistarn, line 20: Can we call this the agent instead of puller? I think it's responsibilities should grow.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+	"regexp"
+	"strings"
+)
+
+var (
+	d *Downloader
+)
+
+type Downloader struct {
+	NumRetries int
```
ellistarn, line 22: I think we probably never want to stop retrying. Keep retrying, but backoff to some slow interval if it keeps failing. These systems need to self heal.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+	//PVC   Protocol = "pvc://"
+	//File  Protocol = "file://"
+	//HTTPS Protocol = "https://"
+)
+
+// TODO: Add more supported protocols
+var SupportedProtocols = []Protocol{S3, GCS}
+
+func init() {
+	d = NewDownloader()
```
ellistarn, line 40: I don't think we want this to be a global singleton.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func NewDownloader() *Downloader {
+	d := new(Downloader)
+	return d
+}
+
+func SetDownloadSettings(numRetries int, modelDir string) {
```
ellistarn, line 48: I'd probably avoid setter/ctor patterns in golang unless you need a builder/factory pattern. This is one of the nice things about go object initialization.

Just set the variables directly at creation

downloader := Downloader {
 NumRetries: ...
 ModelDir: ...
}

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+
+func DownloadModel(event EventWrapper) error {
+	modelSpec := event.ModelSpec
+	if modelSpec != nil {
+		modelUri := modelSpec.StorageURI
+		hashString := hash(modelUri)
+		log.Println("Processing:", modelUri, "=", hashString)
+		successFile := filepath.Join(d.ModelDir, fmt.Sprintf("SUCCESS.%d", hashString))
+		ok := fileExists(successFile)
+		if !ok {
```
ellistarn, line 61: Any reason to not simply inline fileExists(...)

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+			downloadErr := download(modelUri)
+			if downloadErr != nil {
```
ellistarn, line 63: Inline here.

if err:= foo(); err != nil {
  return fmt.Errorf
}

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+				return fmt.Errorf("download error: %v", downloadErr)
+			} else {
+				file, createErr := os.Create(successFile)
+				if createErr != nil {
+					return fmt.Errorf("create file error: %v", createErr)
```
ellistarn, line 68: consider an annotation pattern that will yield cleaner error messages:

"while creating file: %v"

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+	d := new(Downloader)
+	return d
+}
+
+func SetDownloadSettings(numRetries int, modelDir string) {
+	d.NumRetries = numRetries
+	d.ModelDir = modelDir
+}
+
+func DownloadModel(event EventWrapper) error {
```
ellistarn, line 53: IMO you should use a pointer receiver here.

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func download(storageUri string) error {
+	log.Println("Downloading: ", storageUri)
+	prefix, err := validateStorageURI(storageUri)
+	if err != nil {
+		return err
+	}
+	switch prefix {
+	case S3:
```
ellistarn, line 86: Can we break each of these up into different structs, each implementing a download interface?

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	return nil
+}
+
+func download(storageUri string) error {
+	log.Println("Downloading: ", storageUri)
+	prefix, err := validateStorageURI(storageUri)
+	if err != nil {
+		return err
```
ellistarn, line 83: annotate your errors for easier debugging

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+			return fmt.Errorf("file is already created: %v", err)
+		}
+		defer file.Close()
+
+		// TODO: Should the S3 client be shared?
+		sess, err := session.NewSession()
+		if err != nil {
+			return fmt.Errorf("unable to get new s3 session %v", err)
+		}
+		s3Svc := s3.New(sess)
```
ellistarn, line 114: You should be initializing this one time, rather than once per call. Use dependency injection.

ifilonenko: I agree, I was worried about concurrency when initially writing it

pkg/puller/downloader.go
ifilonenko: StorageURI is for caching. If a different model name used the same storageURI it should use the already cached model. The only issue in this design is that you can’t update an existing storageURI and expect it to update. Maybe we can set a “hard-override” property to always re-download

pkg/puller/puller.go
```diff
@@ -0,0 +1,76 @@
...
+func (*Puller) RemoveModel(modelName string) {
+	channel, ok := p.ChannelMap[modelName]
+	if ok {
+		close(channel.EventChannel)
+		delete(p.ChannelMap, modelName)
+	}
+}
+
+func init() {
+	p = NewPuller()
```
ellistarn, line 44: Can you run this logic in main to avoid global singletons

pkg/puller/watcher.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+type EventWrapper struct {
+	ModelSpec      *v1beta1.ModelSpec
+	LoadState      LoadState
+	ShouldDownload bool
+}
+
+type ModelWrapper struct {
+	ModelSpec *v1beta1.ModelSpec
```
yuzisun, line 51: use no-pointer type as it is small struct and required field

pkg/puller/watcher.go
```diff
@@ -0,0 +1,184 @@
...
+								log.Println("Sending event", event)
+								channel.EventChannel <- event
+							}
+						} else {
+							if modelWrapper.Stale {
+								channel, ok := p.ChannelMap[modelName]
+								if !ok {
+									log.Println("Need to add model", modelName)
+									// TODO: Maybe have more workers per Channel?
+									channel = p.AddModel(modelName, w.numWorkers)
```
yuzisun, line 155: I think we should only allow one worker per model channel, otherwise we are going to run into race condition when processing downloading and deleting events at same time.

pkg/puller/watcher.go
```diff
@@ -0,0 +1,184 @@
...
+					w.ModelTracker.Range(func(key interface{}, value interface{}) bool {
+						modelName, modelWrapper := key.(string), value.(ModelWrapper)
+						if modelWrapper.Time.Before(timeNow) {
+							w.ModelTracker.Delete(modelName)
+							channel, ok := p.ChannelMap[modelName]
+							if !ok {
+								log.Println("Model", modelName, "was never added to channel map")
+							} else {
+								event := EventWrapper{
+									ModelSpec:      nil,
```
yuzisun, line 142: Can we use more explicit action enum for deletion/adding/update?

pkg/puller/downloader.go
```diff
@@ -0,0 +1,172 @@
...
+	return nil
+}
+
+func hash(s string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+func validateStorageURI(storageURI string) (Protocol, error) {
```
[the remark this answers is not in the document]
ifilonenko: We have it in 2 other places, but I would prefer to have it return the Protocol, which is different from the other methods that just do a validate method

pkg/puller/watcher.go
ifilonenko: I set it for 1 right now, I'm not quite clear on the use-case for more than 1. but maybe some fancy go-routine `.Merge()` logic can be done?

pkg/puller/watcher.go
ifilonenko: Do we care? I thought we only care about Load and Unload when it comes to a request

pkg/puller/watcher.go
ifilonenko: not necessarily, If its an unload command, we only know the modelName, there is no ModelSpec

cmd/agent/main.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	if *s3Endpoint != "" {
+		sess, err := session.NewSession(&aws.Config{
+			Endpoint: aws.String(*s3Endpoint),
+			Region: aws.String(*s3Region)},
+		)
+		if err != nil {
+			panic(err)
+		}
+		s3Svc := s3.New(sess)
```
ellistarn, line 35: Consider inlining

cmd/agent/main.go
```diff
@@ -0,0 +1,53 @@
...
+		downloader.ProtocolManagers[protocols.S3] = &protocols.S3Manager{
+			S3: s3Svc,
+		}
+	}
+	puller := agent.Puller{
```
ellistarn, line 40: consider inlining

cmd/agent/main.go
```diff
@@ -0,0 +1,53 @@
...
+		ChannelMap: map[string]agent.Channel{},
+		Downloader: downloader,
+	}
+
+	modelTracker := new(syncmap.Map)
```
ellistarn, line 45: consider inlining

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+				return fmt.Errorf("download error: %v", downloadErr)
+			} else {
+				file, createErr := os.Create(successFile)
+				if createErr != nil {
+					return fmt.Errorf("create file error: %v", createErr)
+				}
+				defer file.Close()
+			}
+		} else {
+			log.Println("Model", modelSpec.StorageURI, "exists already")
```
ellistarn, line 42: We should use a logger like zap.S().Infof() or something similar to what we have in the controller.

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
+package agent
+
+import (
+	"fmt"
+	pro "github.com/kubeflow/kfserving/pkg/agent/protocols"
```
ellistarn, line 5: nit: I'd probably just leave this without a name and let it default to the package name "protocols"

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+type Downloader struct {
+	// TODO: Add back-off retries
+	ModelDir   string
+	ProtocolManagers map[pro.Protocol]pro.ProtocolManager
+}
+
+var SupportedProtocols = []pro.Protocol{pro.S3}
+
+func (d *Downloader) DownloadModel(event EventWrapper) error {
+	modelSpec := event.ModelSpec
```
ellistarn, line 23: If nothing else is using event's other fields, why not just pass ModelSpec directly into this method?

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+	if modelSpec != nil {
+		modelUri := modelSpec.StorageURI
+		hashString := hash(modelUri)
+		log.Println("Processing:", modelUri, "=", hashString)
+		successFile := filepath.Join(d.ModelDir, fmt.Sprintf("SUCCESS.%d", hashString))
```
ellistarn: This seems like an odd mechanism for checking completion. Can't you just check the existence of the file you're trying to download?

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+		if !pro.FileExists(successFile) {
+			downloadErr := d.download(modelUri)
```
ellistarn: Consider inlining. There are some super common go idioms for this

```
if err := d.download(modelUri); err != nil {
 // annotate and return error
}
```

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+	log.Println("Downloading: ", storageUri)
+	protocol, err := validateStorageURI(storageUri)
+	if err != nil {
+		return fmt.Errorf("unsupported protocol: %v", err)
+	}
+	manager, ok := d.ProtocolManagers[protocol]
+	if !ok {
+		return fmt.Errorf("protocol manager for %s is not initialized", protocol)
+	}
+	if err = manager.Download(d.ModelDir, storageUri); err != nil {
```
ellistarn: nit, use := instead of =, otherwise you're mutating the error from line 49 and this code will break if you remove that validation logic.

pkg/agent/downloader.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil
+}
+
+func hash(s string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+func validateStorageURI(storageURI string) (pro.Protocol, error) {
```
ellistarn, line 71: I might name this extractProtocol or something similar. It's doing more than validation.

cmd/agent/main.go
```diff
@@ -0,0 +1,53 @@
...
+		sess, err := session.NewSession(&aws.Config{
+			Endpoint: aws.String(*s3Endpoint),
+			Region: aws.String(*s3Region)},
+		)
+		if err != nil {
+			panic(err)
+		}
+		s3Svc := s3.New(sess)
+		downloader.ProtocolManagers[protocols.S3] = &protocols.S3Manager{
+			S3: s3Svc,
```
ellistarn, line 37: Thinking further, it would probably be cleaner to encapsulate the s3 initialization logic inside the S3Manager object. That helps achieve a cleaner main file like you mentioned before.

pkg/agent/protocols/manager.go
```diff
@@ -0,0 +1,15 @@
...
+
+type Protocol string
+
+const (
+	S3    Protocol = "s3://"
+	//GCS    Protocol = "gs://"
+	//PVC   Protocol = "pvc://"
+	//File  Protocol = "file://"
+	//HTTPS Protocol = "https://"
+)
```
ellistarn, line 15: You editor should be adding newlines on save here.

pkg/agent/protocols/manager.go
```diff
@@ -0,0 +1,15 @@
+package protocols
+
+type ProtocolManager interface {
```
ellistarn, line 3: I'd probably call this "Provider". 
```
storageProvider := storage.Provider{}
storageProvider.Download(storageUri, modelDir)
```

pkg/agent/protocols/manager.go
```diff
@@ -0,0 +1,15 @@
...
+	Download(modelDir string, storageUri string) error
```
ellistarn: Typically a download or copy or move API would be <src, dst>.
I'd probably have your storage implementation be model agnostic w/ it's names
e.g. Download(src, dest)

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	s3ObjectLoactor := &S3ObjectDownloader{
+		ModelDir: modelDir,
+		Bucket: bucket,
+		Item: item,
+	}
+	objects, err := s3ObjectLoactor.GetAllObjects(m.S3)
+	if err != nil {
+		return fmt.Errorf("unable to get batch objects %v", err)
+	}
+	err = s3ObjectLoactor.Download(m.S3, objects)
```
ellistarn, line 33: same w.r.t error patterns and inlining

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+type S3Manager struct {
+	S3 *s3.S3
+}
+
+func (m *S3Manager) Download(modelDir string, storageUri string) error {
+	s3Uri := strings.TrimPrefix(storageUri, string(S3))
+	path := strings.SplitN(s3Uri, "/", 2)
+	bucket := path[0]
+	item := path[1]
+	s3ObjectLoactor := &S3ObjectDownloader{
```
ellistarn, line 23: nit: loactor

pkg/agent/protocols/manager.go
```diff
@@ -0,0 +1,15 @@
+package protocols
```
ellistarn, line 1: I might call this package "storage" for better readability.

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	return results, nil
+}
+
+func (s *S3ObjectDownloader) Download(s3Svc *s3.S3, objects []s3manager.BatchDownloadObject) error {
+	iter := &s3manager.DownloadObjectsIterator{Objects: objects}
+	downloader := s3manager.NewDownloaderWithClient(s3Svc, func(d *s3manager.Downloader) {
+		d.Concurrency = 5             // TODO: Override?
+		d.PartSize = 64 * 1024 * 1024 // TODO: Override?
+	})
+	if err := downloader.DownloadWithIterator(aws.BackgroundContext(), iter); err != nil {
```
ellistarn, line 97: +1

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+				return nil
+			},
+		}
+		results = append(results, object)
+	}
+	return results, nil
+}
+
+func (s *S3ObjectDownloader) Download(s3Svc *s3.S3, objects []s3manager.BatchDownloadObject) error {
+	iter := &s3manager.DownloadObjectsIterator{Objects: objects}
```
ellistarn, line 92: consider inlining if only used once.

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type S3Manager struct {
```
ellistarn, line 14: Consider S3StorageProvider or just S3

var storage storage.Provider = storage.S3{}

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	S3 *s3.S3
+}
+
+func (m *S3Manager) Download(modelDir string, storageUri string) error {
+	s3Uri := strings.TrimPrefix(storageUri, string(S3))
+	path := strings.SplitN(s3Uri, "/", 2)
+	bucket := path[0]
```
ellistarn, line 21: consider inlining

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	item := path[1]
```
ellistarn, line 22: consider inlining

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+)
+
+type S3Manager struct {
+	S3 *s3.S3
+}
+
+func (m *S3Manager) Download(modelDir string, storageUri string) error {
+	s3Uri := strings.TrimPrefix(storageUri, string(S3))
+	path := strings.SplitN(s3Uri, "/", 2)
```
ellistarn, line 20: Simpler to just call .Split(s3Uri, "/')

pkg/agent/protocols/s3.go
ellistarn: Why are you giving two names to this thing? locator and downloader. Which is it?

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+		Prefix:    aws.String(s.Item),
+	})
+	if err != nil {
+		return nil, err
+	}
+	results := make([]s3manager.BatchDownloadObject, 0)
+
+	for _, object := range resp.Contents {
+		fileName := filepath.Join(s.ModelDir, *object.Key)
+		ok := FileExists(fileName)
```
ellistarn, line 61: nit: inline

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+		Bucket:    aws.String(s.Bucket),
+		Prefix:    aws.String(s.Item),
+	})
+	if err != nil {
+		return nil, err
+	}
+	results := make([]s3manager.BatchDownloadObject, 0)
+
+	for _, object := range resp.Contents {
+		fileName := filepath.Join(s.ModelDir, *object.Key)
```
ellistarn: Is key the basename of the path? What if the user passes us a storageuri with subfolders?

pkg/agent/protocols/s3.go
```diff
@@ -0,0 +1,99 @@
...
+		results = append(results, object)
+	}
+	return results, nil
+}
+
+func (s *S3ObjectDownloader) Download(s3Svc *s3.S3, objects []s3manager.BatchDownloadObject) error {
+	iter := &s3manager.DownloadObjectsIterator{Objects: objects}
+	downloader := s3manager.NewDownloaderWithClient(s3Svc, func(d *s3manager.Downloader) {
+		d.Concurrency = 5             // TODO: Override?
+		d.PartSize = 64 * 1024 * 1024 // TODO: Override?
```
ellistarn, line 95: consider leaving these out until you want to tune them and know what they should be.

pkg/agent/downloader.go
ifilonenko: We create a file and then use the downloader to write to it. if the downloader fails, we might have a partially filled writer. i.e. data corruption or whatnot. So I think it is safer to use a success file. but I don't mind

pkg/agent/protocols/s3.go
ifilonenko: this handles sub-folders, I can show a demo. But that is handled by the list objects command.

pkg/agent/protocols/s3.go
ifilonenko: Yes :) I initially had 3 :P

pkg/agent/downloader.go
ifilonenko: I had issues with getting it to print in the controller, but we can do that last. I will punt on this, _this_ iteration

pkg/agent/downloader.go
ifilonenko: I will now need the modelName

cmd/agent/main.go
ifilonenko: I am okay with how the `main.go` file is structured atm. It should be doing the heavy lifting of going through the flagged variables. I think its appropriate as is.

pkg/agent/protocols/manager.go
ifilonenko: Yeah, I hadn't ran `fmt vet` yet :)

pkg/agent/protocols/manager.go
ifilonenko: I would have prefered this as well, I can do storageUri, and build the `dest` for you. But the `dest` is a combo of modelDir and modelName

pkg/agent/protocols/s3.go
ifilonenko: The line is too long to inline

pkg/agent/puller.go
```diff
@@ -0,0 +1,104 @@
...
+			if err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+				event.NumRetries += 1
+				if event.NumRetries > 5 {
+					// TODO: Maybe use a done channel from another event coming in?
+					log.Println("Too many retries")
+				} else {
+					log.Println("Adding", event, "back to queue")
+					time.Sleep(time.Duration(event.NumRetries) * time.Second)
+					channel, ok := p.ChannelMap[modelName]
```
yuzisun, line 71: I am not sure what's the behavior when you looping over the channel and at the same time you are sending more events to it

pkg/agent/puller.go
```diff
@@ -0,0 +1,104 @@
...
+	for event := range events {
+		log.Println("worker", id, modelName, "started  job", event)
+		switch event.LoadState {
+		case ShouldLoad:
+			log.Println("Should download", event.ModelSpec.StorageURI)
+			err = p.Downloader.DownloadModel(event)
+			if err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+				event.NumRetries += 1
+				if event.NumRetries > 5 {
```
yuzisun, line 65: I think this is not the right place for retry, what if a delete event comes in before the retry?

pkg/agent/puller.go
```diff
@@ -0,0 +1,104 @@
...
+			if err := p.RemoveModel(event.ModelName); err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+				event.NumRetries += 1
+				if event.NumRetries > 5 {
+					// TODO: Maybe use a done channel from another event coming in?
+					log.Println("Too many retries")
+				} else {
+					log.Println("Adding", event, "back to queue")
+					time.Sleep(time.Duration(event.NumRetries) * time.Second)
+					channel, ok := p.ChannelMap[modelName]
```
yuzisun, line 89: I feel we should remove this retry here and at trained model reconcilation process handles it

pkg/agent/watcher.go
```diff
@@ -0,0 +1,163 @@
...
+								Time:       timeNow,
+								Stale:      isStale,
+								Redownload: reDownload,
+							})
+						}
+					}
+					// TODO: Maybe make parallel and more efficient?
+					w.ModelTracker.Range(func(key interface{}, value interface{}) bool {
+						modelName, modelWrapper := key.(string), value.(ModelWrapper)
+						if modelWrapper.Time.Before(timeNow) {
```
yuzisun, line 111: I think we can build the modelTracker state before starting watching the configmap changes, so you can easily know what are the models are deleted instead of relying on timestamp which is hard to debug. Upon startup the first thing we should do is to load the model state from configmap.

ifilonenko: This is for diff checking, but I agree on the “on-startup” building of the in-memory map. I will add that

pkg/agent/puller.go
ifilonenko: I was just musing here. I agree, can remove

pkg/agent/puller.go
ifilonenko: Another worker picks it up

pkg/agent/puller.go
ifilonenko: That is an excellent question, that’s why I was thinking of creating a done channel to lock out those events. But that can be a follow up PR, as the logic there might get complicated

ifilonenko:
@yuzisun @ellistarn @njhill comments have been resolved and basic functionality is here. I think it might be wise to scope this PR to the current functionality that is provided and leave the rest of the TODOs for follow-up tickets.

pkg/agent/downloader.go
```diff
@@ -0,0 +1,86 @@
...
+	log.Println("Downloading: ", storageUri)
+	protocol, err := extractProtocol(storageUri)
+	if err != nil {
+		return fmt.Errorf("unsupported protocol: %v", err)
+	}
+	manager, ok := d.Providers[protocol]
+	if !ok {
+		return fmt.Errorf("protocol manager for %s is not initialized", protocol)
+	}
+	if err := manager.Download(d.ModelDir, modelName, storageUri); err != nil {
```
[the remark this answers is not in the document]
ifilonenko: They will

pkg/agent/puller.go
```diff
@@ -0,0 +1,77 @@
...
+			log.Println("Should unload", event.ModelName)
+			if err := p.RemoveModel(event.ModelName); err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+			}
+		}
+		// If there is an error, we will NOT send a request. As such, to know about errors, you will
+		// need to call the error endpoint of the puller
+		if err == nil {
+			// TODO: Do request logic
+			log.Println("Now doing a request on", event)
```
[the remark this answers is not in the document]
ifilonenko: We assume calls are idempotent

pkg/agent/watcher.go
```diff
@@ -0,0 +1,165 @@
...
+				isCreate := event.Op&fsnotify.Create != 0
+				eventPath := filepath.Clean(event.Name)
+				isDataDir := filepath.Base(eventPath) == "..data"
+				// TODO: Should we use atomic integer or timestamp??
+				if isDataDir && isCreate {
+					symlink, _ := filepath.EvalSymlinks(eventPath)
+					file, err := ioutil.ReadFile(filepath.Join(symlink, constants.ModelConfigFileName))
+					modelConfigs := make(modelconfig.ModelConfigs, 0)
+					if err != nil {
+						log.Println("Error in reading file", err)
```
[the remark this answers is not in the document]
ifilonenko: Agreed, that was handled

pkg/agent/puller.go
```diff
@@ -0,0 +1,77 @@
...
+			err = p.Downloader.DownloadModel(event)
+			if err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+			}
+		case ShouldUnload:
+			log.Println("Should unload", event.ModelName)
+			if err := p.RemoveModel(event.ModelName); err != nil {
+				log.Println("worker", id, "failed on", event, "because: ", err)
+			}
+		}
```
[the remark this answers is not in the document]
ifilonenko: Agreed

pkg/agent/puller.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil
+}
+
+func (p *Puller) modelProcessor(id int, modelName string, events chan EventWrapper) {
+	log.Println("worker", id, "for", modelName, "is initialized")
+	var err error
+	// TODO: Instead of going through each event, one-by-one, we need to drain and combine
+	// this is important for handling Load --> Unload requests sent in tandem
+	// Load --> Unload = 0 (cancel first load)
+	// Load --> Unload --> Load = 1 Load (cancel second load?)
```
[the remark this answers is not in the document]
ifilonenko: +1

pkg/agent/downloader.go
```diff
@@ -0,0 +1,86 @@
...
+	modelSpec := event.ModelSpec
+	modelName := event.ModelName
+	if modelSpec != nil {
+		modelUri := modelSpec.StorageURI
+		hashString := hash(modelUri)
+		log.Println("Processing:", modelUri, "=", hashString)
+		successFile := filepath.Join(d.ModelDir, modelName, fmt.Sprintf("SUCCESS.%d", hashString))
+		// Download if the event there is a success file and the event is one which we wish to Download
+		if !storage.FileExists(successFile) && event.ShouldDownload {
+			// TODO: Handle retry logic
```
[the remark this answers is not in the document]
ifilonenko: We purge each of the files (objects) that we are downloading

ifilonenko:
@njhill @ellistarn @yuzisun I have resolved almost all of the comments. This current work is fully working e2e, I have run it across a multitude of scenarios, standard pulling, total failure, misconfigured SUCCESS file, and all the works.  @njhill you had a few comments relating to using an in-flight mechanism and removing timestamp. Could I possibly merge this and have that be additive work on-top of this current PR? I feel as though I don't want this to handle all cases, on the first round, and would prefer to not keep it open for more than a month (around 26 days atm).

pkg/agent/syncer.go
```diff
@@ -0,0 +1,29 @@
...
+	"io/ioutil"
+	"log"
+	"path/filepath"
+)
+
+type Syncer struct {
+	Watcher Watcher
+}
+
+func (s *Syncer) Start() {
```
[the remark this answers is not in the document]
ifilonenko: Agreed, I did that here

njhill:
Hey @ifilonenko the latest updates look good but quite a few of my comments weren't addressed (as you mentioned). I still think some of these are important but have no preference w.r.t. merging first, I'm fine with it if others are. If you like I can have a go at a PR after yours to address some of them...

pkg/agent/syncer.go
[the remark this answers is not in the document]
njhill: I still think it would be good to have a single function in `watcher` with the common part of the logic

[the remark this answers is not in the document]
ifilonenko: Ah I decided that because of the model directory sync, keeping it separate had some merit.

yuzisun:
thanks @njhill ! @ifilonenko Can you create a follow up issue with unaddressed comments?

cmd/agent/agent.Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Build the inference-agent binary
+FROM golang:1.13.0 as builder
+
+# Copy in the go src
+WORKDIR /go/src/github.com/kubeflow/kfserving
+COPY pkg/    pkg/
```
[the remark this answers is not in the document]
ifilonenko: ^ that is neat, will need to re-organize for that. good call, because the build took forevahh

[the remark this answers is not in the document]
yuzisun: Can you elaborate ? move before the pkg?

pkg/agent/watcher.go
```diff
@@ -0,0 +1,165 @@
...
+				Redownload: reDownload,
+			})
+		}
+	}
+	// TODO: Maybe make parallel and more efficient?
+	w.ModelTracker.Range(func(key interface{}, value interface{}) bool {
+		modelName, modelWrapper := key.(string), value.(ModelWrapper)
+		if modelWrapper.Time.Before(timeNow) {
+			w.ModelTracker.Delete(modelName)
+			channel, ok := w.Puller.ChannelMap[modelName]
```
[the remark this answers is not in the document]
yuzisun: @njhill I think we should do this, thanks for the suggestion.

PR #991, opened 2020-08-03
InferenceService v1beta1 API

yuzisun:
InferenceService v1beta1 API and InferenceRouter v1alpha1 preview

Fixes #

I am putting inference service and router together mainly for folks to review with a wholistic perspective. 

1. v1beta1 inference service api should be good for review.
2. v1alpha1 inference router api is for preview, mainly focused on the pipeline spec.

**Release note**:

```release-note

```


pkg/apis/routing/v1alpha1/inference_router.go
```diff
@@ -0,0 +1,111 @@
+/*
```
ellistarn, line 1: Consider the path pkg/apis/router/v1alpha1

ellistarn:
Any way to separate some of these types to review them separately in a smaller PR?

yuzisun:
> Any way to separate some of these types to review them separately in a smaller PR?

yes will separate router out to separate PR once getting initial feedback, I want to have a single place for people to review both as an overall plan.

Makefile
```diff
@@ -22,7 +22,8 @@ all: test manager logger batcher
 
 # Run tests
 test: fmt vet lint manifests
-	go test ./pkg/... ./cmd/... -coverprofile cover.out
+	go test ./pkg/... ./cmd/... -coverprofile cover.out.tmp
+	cat cover.out.tmp |grep -v "_generated.go" > cover.out
```
ellistarn, line 26: Why? O.o

docs/samples/v1beta1/custom/simple.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: "serving.kubeflow.org/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "custom-simple"
+spec:
+  predictor:
+    containers:
+      - image: codait/max-object-detector
```
ellistarn, line 8: Is this indented an extra layer too far?

docs/samples/v1beta1/tensorflow/liveness_probe.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: "serving.kubeflow.org/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "flowers-readiness"
+spec:
+  predictor:
+    tensorflow:
+      storageUri: "gs://kfserving-samples/models/tensorflow/flowers"
+      livenessProbe:
```
ellistarn, line 9: Should this live in the custom example?

pkg/apis/routing/v1alpha1/routing_strategy_multiarmbandit.go
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+package v1alpha1
+
+//
+type MultiArmBanditSpec struct {
+	EpsilonGreedy *EpsilonGreedySpec `json:"epsilonGreedy,omitempty"`
+}
+
+type EpsilonGreedySpec struct {
```
ellistarn, line 24: Not confident about this yet, consider leaving this out. Would love to see a proposal and how this evolves.

pkg/apis/serving/v1alpha2/explainer.go
```diff
@@ -20,8 +20,7 @@ import (
 	"k8s.io/klog"
 )
 
-// +k8s:openapi-gen=false
-// +k8s:deepcopy-gen=false
+// +kubebuilder:object:generate=false
```
ellistarn, line 23: We should consider moving these interfaces into pkg/controller or somewhere else. The interfaces can be implemented by our types, but defined in a separate package to keep the generation simple.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
+/*
```
ellistarn, line 1: love it.

docs/samples/v1beta1/custom/simple.yaml
yuzisun: what do you mean by that?

docs/samples/v1beta1/tensorflow/liveness_probe.yaml
yuzisun: sure

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
...
+			return nil, fmt.Errorf("Unable to unmarshall %v json string due to %v ", TransformerConfigKeyName, err)
+		}
+	}
+	return transformerConfig, nil
+}
+
+func getExplainersConfigs(configMap *v1.ConfigMap) (*ExplainersConfig, error) {
+	explainerConfig := &ExplainersConfig{}
+	if data, ok := configMap.Data[ExplainerConfigKeyName]; ok {
+		err := json.Unmarshal([]byte(data), &explainerConfig)
```
ellistarn, line 153: yaml.DisallowUnknownFields

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -0,0 +1,125 @@
...
+	// Spec for TorchServe (https://pytorch.org/serve)
+	PyTorch *TorchServeSpec `json:"pytorch,omitempty"`
+	// Spec for Triton Inference Server (https://github.com/NVIDIA/triton-inference-server)
+	Triton *TritonSpec `json:"triton,omitempty"`
+	// Spec for ONNX runtime (https://github.com/microsoft/onnxruntime)
+	ONNX *ONNXRuntimeSpec `json:"onnx,omitempty"`
+	// Passthrough Pod fields or specify a custom container spec
+	*CustomPredictor `json:",inline"`
+	// Extensions available in all components
+	ComponentExtensionSpec `json:",inline"`
```
ellistarn, line 56: Thoughts on making this guy optional?

docs/samples/v1beta1/custom/simple.yaml
ellistarn: Maybe I'm crazy.
```    
containers:
- image: codait/max-object-detector
```
vs
```
containers:
  - image: codait/max-object-detector
```

pkg/apis/routing/v1alpha1/inference_router.go
```diff
@@ -0,0 +1,111 @@
...
+	// probability e and best model with probability 1-e.
+	// +optional
+	MultiArmBandit *MultiArmBanditSpec `json:"multiArmBandit,omitempty"`
+	// Ensemble is an inference router node which can combine results from previous steps
+	// +optional
+	Ensemble *EnsembleSpec `json:"ensemble,omitempty"`
+	// Pipeline defines a series of steps which are depending on previous step.
+	// Only one root node is allowed and all other routes should specify dependencies with the consumes field.
+	// The graph should form a valid DAG, no cycles are allowed, branches are allowed but they needs
+	// be mutually exclusive. Each route on the pipeline can be either an InferenceService or an Inference Router.
```
cliveseldon, line 50: Would we not allow any Destination (that supports the Data Plane)?

pkg/apis/serving/v1beta1/predictor.go
yuzisun: it was `*ComponentExtensionSpec` before and it gives me big trouble which I spent a week to debug the panic issue with Solly Ross.

pkg/apis/routing/v1alpha1/inference_router.go
```diff
@@ -0,0 +1,111 @@
...
+	// e.g. _x-request-id_.
+	//
+	// Header values are case-sensitive and formatted as follows:
+	// - `exact: "value"` for exact string match
+	// - `prefix: "value"` for prefix-based match
+	// - `regex: "value"` for regex-based match
+	Headers map[string]*StringMatch `json:"headers,omitempty"`
+	// This is a list of pipeline route names that should be executed before the current route.
+	// The field is only valid for pipeline routing type
+	Consumes []string `json:"consumes,omitempty"`
```
cliveseldon, line 71: Not sure I like the Consumes which is only used in Pipeline. Seems confusing. Maybe can rethink later?

pkg/apis/serving/v1beta1/configmap.go
yuzisun: is this an unmarshal option?

pkg/apis/routing/v1alpha1/inference_router.go
```diff
@@ -0,0 +1,111 @@
...
+	// The destination of the route
+	Destination duckv1.Destination `json:",inline"`
+	// The header keys must be lowercase and use hyphen as the separator,
+	// e.g. _x-request-id_.
+	//
+	// Header values are case-sensitive and formatted as follows:
+	// - `exact: "value"` for exact string match
+	// - `prefix: "value"` for prefix-based match
+	// - `regex: "value"` for regex-based match
+	Headers map[string]*StringMatch `json:"headers,omitempty"`
```
cliveseldon, line 68: Just focusing on headers with simple matches makes sense. I assume this is the "Condition" field of the Pipeline design v1?

docs/samples/v1beta1/custom/simple.yaml
yuzisun: oh got it, good catch!

pkg/apis/routing/v1alpha1/inference_router.go
yuzisun: correct, I'd like to see how far header matching can go before we introduce more advanced matching.

pkg/apis/routing/v1alpha1/inference_router.go
yuzisun: sure, open for ideas. The advantage is to reuse the same `RouteSpec` and not introducing a new pipeline CRD, or you can think of other routes types all have an implicit `consumes` for the root.

njhill:
Just a note that I would like to review this but may take a bit of time to get through it!

pkg/apis/serving/v1beta1/configmap.go
ellistarn: Yep

pkg/apis/routing/v1alpha1/inference_router.go
ellistarn: I might simply remove Consumes, for now. I think the whole model of defining dependencies might be the wrong direction. It results in a really challenging to understand semantic. You can build dags as follows:
```
Pipeline1 { Pipeline 2, Pipeline 3}
Pipeline2 { Transformer1, InferenceService1 }
Pipeline3{ Transformer1, InferenceService2 }
```

pkg/apis/serving/v1beta1/configmap.go
yuzisun: does not look like that json unmarshal has this option

ellistarn: https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields

docs/samples/v1beta1/tensorflow/canary.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: "serving.kubeflow.org/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "tensorflow-canary"
+spec:
+  predictor:
+    canaryTrafficPercent: 20
```
ellistarn, line 7: This is sweet. I'm a huge fan.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
...
+type TransformerConfig struct {
+	// transformer docker image name
+	ContainerImage string `json:"image"`
+	// default transformer docker image version
+	DefaultImageVersion string `json:"defaultImageVersion"`
+}
+
+// +kubebuilder:object:generate=false
+type TransformersConfig struct {
+	Feast TransformerConfig `json:"feast,omitempty"`
```
ellistarn, line 80: Love to see this.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, err
+	}
+	return &InferenceServicesConfig{
+		Predictors:   predictorsConfig,
+		Transformers: transformersConfig,
+		Explainers:   explainersConfig,
+	}, nil
+}
+
+func getPredictorsConfigs(configMap *v1.ConfigMap) (*PredictorsConfig, error) {
```
ellistarn, line 128: You could abstract this for all 3 like
```
func getComponentConfig(config interface, key string) error {}
```
In fact, is there any reason you can't let the serialization library do all of this generically and just do
```
json.Unmarshal(InferenceServicesConfig, configmap.Data)
```

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ExplainerSpec defines the container spec for a model explanation server,
+// The following fields follow a "1-of" semantic. Users must specify exactly one spec.
+type ExplainerSpec struct {
+	// Spec for alibi explainer
+	Alibi *AlibiExplainerSpec `json:"alibi,omitempty"`
+	// Pass through Pod fields or specify a custom container spec
+	*CustomExplainer `json:",inline"`
+	// Extensions available in all components
+	*ComponentExtensionSpec `json:",inline"`
```
ellistarn, line 38: Why can this one be a pointer but the PredictorExtensionSpec can't?

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -0,0 +1,99 @@
...
+	} {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (e *ExplainerSpec) GetExplainer() (Explainer, error) {
+	handlers := []Explainer{}
```
ellistarn, line 87: You could do some cool generic looping here. 
```
for _, handler := []Explainer{
  e.CustomExplainer,
  e.Alibi
} { 
   ...
}
```

pkg/apis/serving/v1beta1/inference_service.go
```diff
@@ -0,0 +1,132 @@
...
+	// Minimum number of replicas, defaults to 1 but can be set to 0 to enable scale-to-zero.
+	// +optional
+	MinReplicas *int `json:"minReplicas,omitempty"`
+	// Maximum number of replicas for autoscaling.
+	// +optional
+	MaxReplicas int `json:"maxReplicas,omitempty"`
+	// ContainerConcurrency specifies how many requests can be processed concurrently, this sets the hard limit of the container
+	// concurrency(https://knative.dev/docs/serving/autoscaling/concurrency).
+	// +optional
+	ContainerConcurrency int `json:"containerConcurrency,omitempty"`
```
ellistarn, line 49: Should this be a *int

pkg/apis/serving/v1beta1/inference_service.go
```diff
@@ -0,0 +1,132 @@
...
+	// TimeoutSeconds specifies the number of seconds to wait before timing out a request to the component.
+	// +optional
+	TimeoutSeconds int `json:"timeout,omitempty"`
```
ellistarn, line 52: Should this be a *int

pkg/apis/serving/v1beta1/inference_service.go
```diff
@@ -0,0 +1,132 @@
...
+	// CanaryTrafficPercent defines the traffic split percentage between the candidate revision and the last ready revision
+	// +optional
+	CanaryTrafficPercent int `json:"canaryTrafficPercent,omitempty"`
```
ellistarn, line 55: Should this be a *int

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -0,0 +1,99 @@
...
+		validateLogger(e.LoggerSpec),
+	} {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (e *ExplainerSpec) GetExplainer() (Explainer, error) {
```
ellistarn, line 86: I've been thinking about this function. We have a lot of code that needs to get the explainer and needs to do error checking every time this is called. The error only necessary for the validation check and after that, it would be quite a bit cleaner for other code to just call GetExplainer and be guaranteed the correct explainer.

```
func GetExplainer() Explainer
func ValidateExplainerImplementation() error
```
Likewise for Predictor/Transformer.

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -0,0 +1,105 @@
...
+	for k, _ := range alibi.Config {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		args = append(args, "--"+k)
+		args = append(args, alibi.Config[k])
+	}
+
+	return &v1.Container{
```
ellistarn, line 88: Now that AlibiExplainerSpec is a ContainerSpec, should we just populate the ContainerSpec's fields and return it? i.e.

```
...
alibi.ContainerSpec.Image = config.Explainers.AlibiExplainer.ContainerImage + ":" + alibi.RuntimeVersion,
...
return alibi.ContainerSpec
```

pkg/apis/serving/v1beta1/explainer_custom.go
```diff
@@ -0,0 +1,59 @@
...
+	return nil
+}
+
+// Default sets defaults on the resource
+func (c *CustomExplainer) Default(config *InferenceServicesConfig) {
+	c.Name = constants.InferenceServiceContainerName
+	setResourceRequirementDefaults(&c.Spec.Containers[0].Resources)
+}
+
+func (c *CustomExplainer) GetStorageUri() *string {
```
ellistarn, line 46: Can we genericize this implementation to be the same for all explainers, transformers, predictors. I think having a hierarchy of checking spec.StorageUri and then falling back on a env is a good way of doing this for all explainers.

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -0,0 +1,99 @@
...
+	Alibi *AlibiExplainerSpec `json:"alibi,omitempty"`
+	// Pass through Pod fields or specify a custom container spec
+	*CustomExplainer `json:",inline"`
+	// Extensions available in all components
+	*ComponentExtensionSpec `json:",inline"`
+}
+
+// Explainer interface is implemented by all explainers
+// +kubebuilder:object:generate=false
+type Explainer interface {
```
ellistarn, line 43: I think we should be able to abstract this interface into component.go that implements:
```
GetContainer()
Default()
Validate()
```
Predictor/Explainer/Transformer can all implement this.


We can have a second interface 
explainer.go
that has the specialized function
```
GetStorageUri()
```

I think this should clean up a ton of validation/defaulting/retrieval logic.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -0,0 +1,84 @@
...
+	configMap, err := GetInferenceServicesConfig(cli)
+	if err != nil {
+		panic(err)
+	}
+	if predictor, err := isvc.Spec.Predictor.GetPredictor(); err == nil {
+		predictor.Default(configMap)
+	}
+
+	if isvc.Spec.Transformer != nil {
+		if transformer, err := isvc.Spec.Transformer.GetTransformer(); err == nil {
```
ellistarn, line 74: As discussed above, if this GetTransformer() call just returns a Transformer instead of an error as well, it cleans things up. Additionally, if you generalize these components behind a common interface (component.go), you should be able to express this as:
```
for _, component := range []Component{
  isvc.Spec.Predictor,
  isvc.Spec.Explainer,
  isvc.Spec.Transformer,
} {
  if component != nil {
    component.Default(configmap)
  }
}
```
Likewise for validation.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -0,0 +1,178 @@
...
+	SupportedStorageURIPrefixList = []string{"gs://", "s3://", "pvc://", "file://"}
+	AzureBlobURIRegEx             = "https://(.+?).blob.core.windows.net/(.+)"
+	// logger for the validation webhook.
+	validatorLogger = logf.Log.WithName("inferenceservice-v1beta1-validation-webhook")
+)
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (isvc *InferenceService) ValidateCreate() error {
+	validatorLogger.Info("validate create", "name", isvc.Name)
+
```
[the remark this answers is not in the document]
ellistarn: Passing over this again, see what I wrote re: defaulter.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -0,0 +1,178 @@
...
+	if logger != nil {
+		if !(logger.Mode == LogAll || logger.Mode == LogRequest || logger.Mode == LogResponse) {
+			return fmt.Errorf(InvalidLoggerType)
+		}
+	}
+	return nil
+}
+
+// GetIntReference returns the pointer for the integer input
+func GetIntReference(number int) *int {
```
ellistarn, line 175: proto.Int(int) will do this for ya.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -0,0 +1,125 @@
...
+	// +optional
+	RuntimeVersion string `json:"runtimeVersion,omitempty"`
+	// Container enables overrides for the predictor.
+	// Each framework will have different defaults that are populated in the underlying container spec.
+	// +optional
+	v1.Container `json:",inline"`
+}
+
+// GetPredictor returns the framework for the Predictor
+func (p *PredictorSpec) GetPredictor() (Predictor, error) {
```
ellistarn, line 73: Same thing re: splitting one-of error handling and GetPredictor() -> Predictor logic.

pkg/apis/serving/v1beta1/predictor_sklearn.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if containerConcurrency != 0 {
+		arguments = append(arguments, fmt.Sprintf("%s=%s", constants.ArgumentWorkers, strconv.Itoa(containerConcurrency)))
+	}
+	if k.Container.Image == "" {
+		k.Container.Image = config.Predictors.SKlearn.ContainerImage + ":" + k.RuntimeVersion
+	}
+	k.Container.Name = constants.InferenceServiceContainerName
+	k.Container.Args = arguments
+	return &k.Container
```
ellistarn, line 62: Nice. Glad to see you utilizing the inlined container spec. See comment above in explainer where I'm looking for something like this.

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -0,0 +1,86 @@
...
+	TensorflowServingGRPCPort            = "9000"
+	TensorflowServingRestPort            = "8080"
+	TensorflowServingGPUSuffix           = "-gpu"
+	InvalidTensorflowRuntimeVersionError = "Tensorflow RuntimeVersion must be one of %s"
+	InvalidTensorflowRuntimeIncludesGPU  = "Tensorflow RuntimeVersion is not GPU enabled but GPU resources are requested. " + InvalidTensorflowRuntimeVersionError
+	InvalidTensorflowRuntimeExcludesGPU  = "Tensorflow RuntimeVersion is GPU enabled but GPU resources are not requested. " + InvalidTensorflowRuntimeVersionError
+)
+
+// TFServingSpec defines arguments for configuring Tensorflow model serving.
+type TFServingSpec struct {
```
ellistarn, line 37: After our discussion on frameworks, I think we decided to go back to calling this Tensorflow instead of TFServing. The same way TorchServe will be called PyTorch.

pkg/apis/serving/v1beta1/predictor_triton.go
```diff
@@ -0,0 +1,72 @@
...
+		t.RuntimeVersion = config.Predictors.Triton.DefaultImageVersion
+	}
+	setResourceRequirementDefaults(&t.Resources)
+}
+
+// GetContainers transforms the resource into a container spec
+func (t *TritonSpec) GetContainer(modelName string, containerConcurrency int, config *InferenceServicesConfig) *v1.Container {
+	arguments := []string{
+		"trtserver",
+		fmt.Sprintf("%s=%s", "--model-store", constants.DefaultModelLocalMountPath),
```
ellistarn, line 54: Consider writing a helper function (useful for all predictor implementations)

```
func ToContainerArguments(arguments map[string]string) []string
```

pkg/apis/serving/v1beta1/predictor_triton.go
```diff
@@ -0,0 +1,72 @@
...
+	if t.RuntimeVersion == "" {
+		t.RuntimeVersion = config.Predictors.Triton.DefaultImageVersion
+	}
+	setResourceRequirementDefaults(&t.Resources)
+}
+
+// GetContainers transforms the resource into a container spec
+func (t *TritonSpec) GetContainer(modelName string, containerConcurrency int, config *InferenceServicesConfig) *v1.Container {
+	arguments := []string{
+		"trtserver",
```
ellistarn, line 53: Should this line be part of command rather than arguments?

pkg/apis/serving/v1beta1/predictor_xgboost.go
```diff
@@ -0,0 +1,66 @@
...
+	if x.RuntimeVersion == "" {
+		x.RuntimeVersion = config.Predictors.XGBoost.DefaultImageVersion
+	}
+	setResourceRequirementDefaults(&x.Resources)
+}
+
+// GetContainer transforms the resource into a container spec
+func (x *XGBoostSpec) GetContainer(modelName string, containerConcurrency int, config *InferenceServicesConfig) *v1.Container {
+	arguments := []string{
+		fmt.Sprintf("%s=%s", constants.ArgumentModelName, modelName),
```
ellistarn, line 49: That helper would be useful here :)

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -0,0 +1,73 @@
...
+	// +required
+	PredictorModel ModelSpec `json:"predictorModel"`
+}
+
+// ModelSpec describes a trained model
+type ModelSpec struct {
+	// Storage URI for the model repository
+	StorageURI string `json:"storageUri"`
+	// Machine Learning framework which is used for trained model
+	// Valid values are:
```
ellistarn, line 53: Does this enum map to our Predictor implementations? Can we codify this in code somehow?

pkg/apis/serving/v1beta1/transformer.go
```diff
@@ -0,0 +1,85 @@
...
+// +kubebuilder:object:generate=false
+type Transformer interface {
+	GetContainer(metadata metav1.ObjectMeta) *v1.Container
+	GetStorageUri() *string
+	Default()
+	Validate() error
+}
+
+func (t *TransformerSpec) GetTransformer() (Transformer, error) {
+	transformers := []Transformer{}
```
ellistarn, line 48: We're missing feast here.

pkg/apis/serving/v1beta1/v1beta1.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NOTE: Boilerplate only.  Ignore this file.
+
+// Package v1beta1 contains API Schema definitions for the serving v1beta1 API group
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=package,register
+// +k8s:conversion-gen=kfserving/pkg/apis/serving
+// +k8s:defaulter-gen=TypeMeta
+// +groupName=serving.kubeflow.org
+package v1beta1
```
ellistarn, line 25: I was discussing this on another PR from Bloomberg, but I'll mention too you too.

Given how complex each of our specs are, I think we should be striving to separate these components into different packages to keep the code maintainable. However, they all need to be part of the same API Group. Everything should be part of `serving.kubeflow.org` and different resources will implement `v1alpha2` or `v1beta1`. The folder name "serving" used to corresp
ond to the group name, but we've evolved beyond that now.

Therefore, I think it probably makes sense to do something like this.
```
pkg/apis/v1alpha2.go // Defines alpha apis/schemes/GVK (group version kind)
pkg/apis/v1beta1.go // Defines beta apis/schemes/GVK
pkg/apis/inferencerouter/v1alpha1
pkg/apis/inferenceservice/v1alpha2
pkg/apis/inferenceservice/v1beta1
pkg/apis/trainedmodel/v1beta1
```
What do you think?

pkg/apis/serving/v1beta1/inference_service_validation.go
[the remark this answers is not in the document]
yuzisun: I have refactored to use the `Component` interface.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: I think this will not work? We need to detect exact one spec.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: oh forgot to fix for explainer and transformer.

pkg/apis/serving/v1beta1/configmap.go
yuzisun: We are not storing the entire config as one json, it is multiple data items in the configmap.

pkg/apis/serving/v1beta1/predictor_tfserving.go
yuzisun: the name on yaml is `tensorflow` but I'd like to keep internal struct still as `TFServingSpec` as it is really creating tfserving container under the hood.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: you are right, I have consolidated them into the `Component` interface

pkg/apis/serving/v1beta1/v1beta1.go
yuzisun: sgtm,  as this is going to involve changes in many places can we do it in a follow up PR?

pkg/apis/serving/v1beta1/transformer.go
yuzisun: not sure what to do with feast, do you have any plan?

pkg/apis/serving/v1beta1/trained_model.go
yuzisun: We will have trained model validation webhook which codifies this, see details [here](https://github.com/kubeflow/kfserving/issues/1006)

pkg/apis/serving/v1beta1/transformer.go
ellistarn: Let's discuss in WG.

pkg/apis/serving/v1beta1/v1beta1.go
ellistarn: Yeah definitely.

pkg/apis/serving/v1beta1/configmap.go
ellistarn: Ah I see. configMap.Data is a map[string]string. Still, can be done generically like:
```
cfg: = InferenceServicesConfig{}
for _, err := []error{
  getComponentConfig(PredictorConfigKeyName, cfg.Predictors),
  getComponentConfig(ExplainerConfigKeyName, cfg.Explainers),
  getComponentConfig(TransformerConfigKeyName, cfg.Transformers),
} {
  if err != nil {
    return nil, err 
  } 
}
return cfg, nil

...
func getComponentConfig(key string, cfg interface) error { 
...
}
```
Just have to make sure InferenceServicesConfig doesn't use pointers, which it probably shouldn't anyways.

pkg/apis/serving/v1beta1/predictor.go
ellistarn: It's not clear to me why. There are other specs that use inline pointers.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
...
+
+// +kubebuilder:object:generate=false
+type TransformersConfig struct {
+	Feast TransformerConfig `json:"feast,omitempty"`
+}
+
+// +kubebuilder:object:generate=false
+type InferenceServicesConfig struct {
+	// Transformer configurations
+	Transformers *TransformersConfig `json:"transformers"`
```
ellistarn, line 86: IMO, we probably don't want these to be pointers.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -0,0 +1,159 @@
...
+	// Predictor configurations
+	Predictors *PredictorsConfig `json:"predictors"`
+	// Explainer configurations
+	Explainers *ExplainersConfig `json:"explainers"`
+}
+
+func GetInferenceServicesConfig(client client.Client) (*InferenceServicesConfig, error) {
```
ellistarn, line 93: Consider reorienting this code around an object. Whenever I see passing a client as a function parameter, I think that something isn't encapsulated correctly.

What if we attached this behavior to the InferenceServicesConfig object?
cfg := InferenceServicesConfig{ client: client } 

// On configmap watch event.
err := cfg.load()

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -0,0 +1,80 @@
...
+		validateLogger(e.LoggerSpec),
+	} {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (e *ExplainerSpec) GetExplainer() (Component, error) {
```
ellistarn, line 67: What happened to the comment about simply returning (Component) and doing error checking elsewhere?

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -0,0 +1,115 @@
...
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		args = append(args, "--"+k)
+		args = append(args, alibi.Config[k])
+	}
+	if alibi.Container.Image == "" {
+		alibi.Image = config.Explainers.AlibiExplainer.ContainerImage + ":" + alibi.RuntimeVersion
+	} else {
```
ellistarn, line 97: dead else?

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -0,0 +1,115 @@
...
+
+	}
+	alibi.Name = constants.InferenceServiceContainerName
+	alibi.Args = args
+	return &alibi.Container
+}
+
+func (alibi *AlibiExplainerSpec) Default(config *InferenceServicesConfig) {
+	alibi.Name = constants.InferenceServiceContainerName
+	if alibi.RuntimeVersion == "" {
```
ellistarn, line 107: Cleaner to set runtime version to a pointer and nil check here, I think.

ellistarn:
@googlebot I consent.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: just note that we need to use `reflect.ValueOf(explainer).IsNil()` for the interface nil check

pkg/apis/serving/v1beta1/configmap.go
yuzisun: The config structs for predictor, explainer and transformer are fairly different, not sure how we can use the cfg interface.

pkg/apis/serving/v1beta1/explainer_custom.go
yuzisun: The problem is that `StorageUri` is not at the explainer or predictor spec level, it is a field on specific predictor or explainer spec, maybe we should really move the field one level up and totally get rid of the magic environment variable.

pkg/apis/serving/v1beta1/inference_service_validation.go
yuzisun: proto only supports int32, in64, still need a small utility for int

pkg/apis/serving/v1beta1/configmap.go
ellistarn: The helper function doesn't do anything beyond pass the struct to ```err := json.Unmarshal([]byte(data), &predictorConfig)```. This function takes an interface https://golang.org/pkg/encoding/json/#Unmarshal, so you don't actually need to know what kind of config it is when you serialize it, it will use reflection to load the variables. 

If this doesn't compile, I may be confused about something in golang, but I'm 99% sure you can do this.

pkg/apis/serving/v1beta1/explainer.go
ellistarn: Correct. Kind of a bummer in golang, but I see this all over in OSS go.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: addressed, it is returning []component now

pkg/apis/serving/v1beta1/configmap.go
yuzisun: I think you are right, fixed.

wengyao04:
@googlebot I consent

pkg/apis/serving/v1beta1/component.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Components interface is implemented by predictor, transformer and explainer
+// +kubebuilder:object:generate=false
+type Component interface {
+	GetContainer(metadata metav1.ObjectMeta, containerConcurrency *int64, config *InferenceServicesConfig) *v1.Container
```
ellistarn, line 11: Now that containerconcurrency is in the component itself, do we need to pass this through?

yuzisun: it is on the `ComponentExtensionSpec` not the model server spec, ModelServerSpec maps to the container.

PR #992, opened 2020-08-03
Add multimodel configmap

wengyao04:

Special notes for your reviewer: Create multimodelconfig for multi-model sharing 

The multi-model configmap will be mounted to the volumes, and be used as files which are watched by the sidecar, the sidecar will load or unload the models based on the change of the files.

The  multi-model configmap is like
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: models-config
  namespace: <user-model-namespace>
data:
  models.json: |
      [
        {
          "modelName": "model1",
          "modelSpec": {
            "storageUri": "s3://example-bucket/path/to/model1",
            "framework": "sklearn",
            "memory": "1G"
          }
        },
        {
          "modelName": "model2",
          "modelSpec": {
            "storageUri": "s3://example-bucket/path/to/model2",
            "framework": "sklearn",
            "memory": "1G"
          }
        }
  ]
```

multimodelconfig package contains
```
type ModelConfig struct {
	Name string             `json:"modelName"`
	Spec *v1beta1.ModelSpec `json:"modelSpec"`
}

type ModelConfigs []*ModelConfig

type ConfigsDelta struct {
	updated map[string]*ModelConfig
	deleted []string
}
```

```
func NewConfigsDelta(updatedConfigs ModelConfigs, deletedConfigs []string) *ConfigsDelta
func (config *ConfigsDelta) Process(configMap *v1.ConfigMap) (err error) 
```
Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.
 Release note:
Add new multimodel configmap


pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+type MultiModelConfig struct {
+	fileName    string
+	fileContent ModelDefinition
+}
+
+type ModelDefinition struct {
```
yuzliu, line 14: Nice, we can add dimension information required by triton server here later.

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
...
+	StorageUri string `json:"storageUri"`
+	Framework  string `json:"framework"`
+	Memory     string `json:"memory"`
```
ifilonenko, line 17: Isn't memory another struct? i.e. cpu, gpu, memory?

yuzliu: No, we decided to change it to memory only in my previous pr

yuzisun: the type should be `resource.Quantity` not string

yuzisun: also this struct seems to be the same as `ModelSpec` defined on `TrainedModel`, should we use that instead?

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"encoding/json"
+	"k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+type MultiModelConfig struct {
+	fileName    string
+	fileContent ModelDefinition
```
yuzisun, line 11: I think the member should be a list of model config items

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
...
+//      memory: 1G
+//    }
+//  example_model_name2.json: |
+//    {
+//      storageUri: s3://example-bucket/path/to/model_name2
+//      framework: sklearn
+//      memory: 2G
+//    }
+
+func AddOrUpdateMultiModelConfigMap(configMap *v1.ConfigMap, multiModelConfig ...MultiModelConfig) error {
```
yuzisun, line 40: suggest using receiver function
`func (MultiModelConfig *) AddOrUpdate(configMap *v1.ConfigMap)`

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, filename := range modelFileNames {
+		if _, ok := configMap.Data[filename]; ok {
+			delete(configMap.Data, filename)
+		} else {
+			klog.Warningf("%s does not exist in configmap %s", filename, configMap.Name)
+		}
+	}
+}
+
+func convert(from ...MultiModelConfig) (map[string]string, error) {
```
yuzisun, line 68: this can be made as constructor of `MultiModelConfig`

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,78 @@
+package multimodelconfig
+
+import (
+	"encoding/json"
+	"k8s.io/api/core/v1"
+	"k8s.io/klog"
```
yuzisun, line 6: we are trying to use [controller runtime logger](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/log) more consistently across the repo

pkg/multimodelconfig/configmap.go
yuzliu: > No, we decided to change it to memory only in my previous pr

oops sorry I changed it to ```Memory resource.Quantity `json:"memory,omitempty"`` `` later

yuzliu: Discussed with @yuzisun offline, we can keep the ModelDefinition struct but we don't need to put the `Memory` field since model servers does not need this information to load/unload. The memory field will be used by the TrainedModel controller to decide if the parent inferenceservice has enough memory.

yuzliu:
LGTM

pkg/multimodelconfig/configmap.go
ifilonenko: I agree with removing `memory` for now

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+//    {
+//      storageUri: s3://example-bucket/path/to/model_name1
+//      framework: sklearn
+//      memory: 1G
+//    }
+//  example_model_name2.json: |
+//    {
+//      storageUri: s3://example-bucket/path/to/model_name2
+//      framework: sklearn
+//      memory: 2G
```
njhill, line 55: `memory` still in this example

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+type ModelDefinition struct {
+	StorageUri string `json:"storageUri"`
+	Framework  string `json:"framework"`
+}
+
+func NewConfigsDelta(updatedCfg[]ModelConfig, deletedCfg []ModelConfig) (*ConfigsDelta, error) {
+	var configsDelta ConfigsDelta
+	var err error
```
ellistarn, line 28: Consider typing your return types to avoid these vars

```func NewConfigsDelta(updatedCfg[]ModelConfig, deletedCfg []ModelConfig) (delta *ConfigsDelta, err error)```

File: pkg/multimodelconfig/configmap.go [hunk 9]
```diff
@@ -0,0 +1,110 @@
+package multimodelconfig
```
ellistarn, line 1: As I understand it, this isn't only for multimodel and in the degenerate case, we will use the same mechanism for a single model. What do you think about putting this code in the TrainedModel api package?

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,110 @@
...
+	updatedCfg map[string]string
+	deletedCfg map[string]string
+}
+
+type ModelConfig struct {
+	fileName    string
+	fileContent *ModelDefinition
+}
+
+type ModelDefinition struct {
```
ellistarn, line 21: Should this be the same as the TrainedModel struct?

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,110 @@
...
+//      storageUri: s3://example-bucket/path/to/model_name2
+//      framework: sklearn
+//    }
+
+func (config *ConfigsDelta) Process(configMap *v1.ConfigMap) {
+	config.addOrUpdate(configMap)
+	config.delete(configMap)
+}
+
+func (config *ConfigsDelta) addOrUpdate(configMap *v1.ConfigMap) {
```
ellistarn, line 61: consider addOrUpdate -> apply

python/kfserving/test/test_storage.py
```diff
@@ -73,7 +73,7 @@ def test_no_permission_buckets(mock_connection, mock_minio):
     bad_gcs_path = "gs://random/path"
     # Access private buckets without credentials
     mock_minio.return_value = Minio("s3.us.cloud-object-storage.appdomain.cloud", secure=True)
-    mock_connection.side_effect = error.AccessDenied(None)
+    mock_connection.side_effect = error.AccessDenied()
```
ellistarn, line 76: Should this be a separate PR?

wengyao04: I will send a separate PR to fix it

pkg/multimodelconfig/configmap.go
wengyao04: We will replace ModelDefinition with ModelSpec in TrainModel after https://github.com/kubeflow/kfserving/pull/991 is merged to master. I add a comment here.

pkg/multimodelconfig/configmap.go
wengyao04: We prefer to put configmap in pkg/multimodelconfig but not in TrainedModel api package since we it is used internally and we don't want to expose this package externally.

yuzisun:
@wengyao04 Can you rebase the minio fix?

/lgtm

pkg/multimodelconfig/configmap.go
ellistarn: If this is used for both multimodel and single model, what do you think about
```
/pkg/modelconfig
```

pkg/multimodelconfig/configmap.go
ellistarn: I recommend just waiting for that to submit first, or stacking your PR on top of #991

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"encoding/json"
+	"k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var logger = log.Log.WithName("MultiModelConfig")
+
+type ConfigsDelta struct {
+	updatedCfg map[string]string
```
ellistarn, line 12: consider:
```
type ModelConfigDelta struct {
  updated ...
  deleted ...
}
```
configsdelta.updatedcfg stutters with the word "config"

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+//after we merge https://github.com/kubeflow/kfserving/pull/991 to master
+type ModelDefinition struct {
+	StorageUri string `json:"storageUri"`
+	Framework  string `json:"framework"`
+}
+
+func NewConfigsDelta(updatedCfg[]ModelConfig, deletedCfg []ModelConfig) (*ConfigsDelta, error) {
+	updatedcfgs, err := convert(updatedCfg...)
+	if err != nil {
+		return nil, err
```
ellistarn, line 31: Consider annotating your error message e.g.
```return nil, fmt.Errorf("while parsing configuration to map, %v, err)```
This will give you nicely annotated error messages if you follow this pattern all the way up the stack.

"Failed to do something, while doing something else, while making an api call, while parsing the json"

File: pkg/multimodelconfig/configmap.go (hunk 13 above)
Comment (ellistarn) [inline pkg/multimodelconfig/configmap.go:12]: Why use map[string]string instead of the objects themselves?
```
type ModelConfigDelta struct {
  updated []ModelConfig
  delete []ModelCnofig
}
```

File: pkg/multimodelconfig/configmap.go [hunk 15]
```diff
@@ -0,0 +1,112 @@
...
+		if _, ok := configMap.Data[filename]; ok {
+			delete(configMap.Data, filename)
+		} else {
+			logger.Info("Model filename does not exist in configmap",
+				"model file name", filename, "configmap name", configMap.Name)
+		}
+	}
+}
+
+func isEmpty(input map[string]string) bool {
```
ellistarn, line 94: len(nil) is 0, so you can just use len instead of writing this function.

pkg/multimodelconfig/configmap.go
yuzliu: @ellistarn and @wengyao04 I submitted https://github.com/kubeflow/kfserving/pull/1009 to merge ModelSpec into master branch.

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return nil
+}
+
+func (config *ConfigsDelta) apply(configMap *v1.ConfigMap) (err error) {
+	if len(config.updated) == 0 {
+		return nil
+	}
+	if configMap.Data == nil {
+		configMap.Data = map[string]string{}
```
yuzisun, line 62: suggest `configMap.Data = make(map[string]string)`

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,128 @@
...
+				}
+			}
+			if to, err := encode(data); err == nil {
+				configMap.Data[constants.MultiModeConfigFileName] = to
+			} else {
+				err = fmt.Errorf("while deleting %s err %v", configMap.Name, err)
+			}
+		} else {
+			err = fmt.Errorf("while deleting %s err %v", configMap.Name, err)
+		}
```
yuzisun, line 102: I think it is easier to read if coding this way
```
if data, err := decode(configData); err == nil {
   return err
}
for name, _ := range config.deleted {
}

if to, err := encode(data); err != nil {
     return err
}
configMap.Data[constants.MultiModeConfigFileName] = to

```

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,128 @@
...
+		if to, err := encode(data); err == nil {
+			configMap.Data[constants.MultiModeConfigFileName] = to
+		} else {
+			err = fmt.Errorf("while updating %s err %v", configMap.Name, err)
+		}
+	} else {
+		err = fmt.Errorf("while updating %s err %v", configMap.Name, err)
+	}
+	return err
```
yuzisun, line 79: likewise the comment for delete block

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (config *ConfigsDelta) apply(configMap *v1.ConfigMap) (err error) {
+	if len(config.updated) == 0 {
+		return nil
+	}
+	if configMap.Data == nil {
+		configMap.Data = map[string]string{}
+	}
+	if data, err := decode(configMap.Data[constants.MultiModeConfigFileName]); err == nil {
+		if len(data) == 0 {
```
yuzisun, line 65: it is cleaner to check whether the entry exists?

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (config *ConfigsDelta) apply(configMap *v1.ConfigMap) (err error) {
+	if len(config.updated) == 0 {
+		return nil
+	}
+	if configMap.Data == nil {
+		configMap.Data = map[string]string{}
+	}
+	if data, err := decode(configMap.Data[constants.MultiModeConfigFileName]); err == nil {
```
yuzliu, line 64:
```suggestion
	if data, err := decode(configMap.Data[constants.ModelConfigFileName]); err == nil {
```

As Ellis suggested in pr https://github.com/kubeflow/kfserving/pull/1013, let's change MultiModeConfigFileName o ModelConfigFileName

pkg/constants/constants.go
```diff
@@ -43,6 +43,11 @@ var (
 	InferenceServiceConfigMapName = "inferenceservice-config"
 )
 
+// InferenceService MultiModel Constants
+var (
+	MultiModeConfigFileName = "multi-models-config.json"
```
yuzliu, line 48:
```suggestion
	ModelConfigFileName = "multi-models-config.json"
```

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,134 @@
...
+			if len(to) == 0 {
+				to = ModelConfig{}
+			}
+			err = nil
+			break
+		} else if err != nil {
+			err = fmt.Errorf("fail to decode %s", from)
+			break
+		}
+	}
```
njhill, line 123: What's the loop here needed for?

How about replacing function contents with
```go
err = json.Unmarshall([]byte(from), &to)
```
this should be more efficient since no intermediate allocations ... decoders are pooled/reused, same for `encode` (`Marshall`).

Also can get a big speedup by dropping in something like https://github.com/json-iterator/go

pkg/multimodelconfig/configmap.go
```diff
@@ -0,0 +1,134 @@
...
+//      }
+//    }
+
+func (config *ConfigsDelta) Process(configMap *v1.ConfigMap) (err error) {
+	if err = config.apply(configMap); err != nil {
+		return err
+	}
+	if err = config.delete(configMap); err != nil {
+		return err
+	}
```
njhill, line 53: It would be more efficient to combine apply/delete or at least lift the json encoding/decoding so that it isn't done twice in the case there are both updates and deletions

pkg/modelconfig/configmap.go
```diff
@@ -0,0 +1,124 @@
...
+	"k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var logger = log.Log.WithName("ModelConfig")
+var json = jsoniter.ConfigCompatibleWithStandardLibrary
+
+type ModelConfig struct {
+	Name string             `json:"modelName"`
+	Spec *v1beta1.ModelSpec `json:"modelSpec"`
```
yuzisun, line 17: we can make this non-pointer type and change `ConfigsDelta` to 
```
type ConfigsDelta struct {
	updated map[string]*ModelConfig
	deleted []string
}
```

pkg/modelconfig/configmap.go
```diff
@@ -0,0 +1,124 @@
...
+
+func slice2Map(from ModelConfigs) map[string]*ModelConfig {
+	to := make(map[string]*ModelConfig)
+	for _, config := range from {
+		to[config.Name] = config
+	}
+	return to
+}
+
+func map2slice(from map[string]*ModelConfig) ModelConfigs {
```
yuzisun, line 102: nit: map2slice->map2Slice

pkg/modelconfig/configmap_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+func getSortedConfigData(input string) (output ModelConfigs, err error) {
+	if err = json.Unmarshal([]byte(input), &output); err != nil {
+		return nil, err
+	}
+	sort.Slice(output, func(i, j int) bool {
+		return output[i].Name < output[j].Name
+	})
+	return output, nil
+}
```
yuzisun, line 226: nit: newline

pkg/constants/constants.go
```diff
@@ -43,6 +43,11 @@ var (
 	InferenceServiceConfigMapName = "inferenceservice-config"
 )
 
+// InferenceService MultiModel Constants
+var (
+	ModelConfigFileName = "models-config.json"
```
ellistarn, line 48: Consider models.json

PR #1017, opened 2020-08-11
Avoid duplicate hostnames in VirtualService when running in cluster-local domain

akirillov:
When KNative Serving default domain is set to be cluster-local by default (i.e. `svc.cluster.local`), the external host domain is not different from the internal one which creates duplicate hostnames in [virtualservice.go#L278](https://github.com/kubeflow/kfserving/blob/master/pkg/controller/inferenceservice/resources/istio/virtualservice.go#L278) and the created `VirtualService` fails Istio validation with:
```
admission webhook \"pilot.validation.istio.io\" denied the request: configuration is invalid:
duplicate hosts in virtual service: mnist.test.svc.cluster.local & mnist.test.svc.cluster.local
```

This PR adds a check in `virtualservice.go` to avoid duplicates when running cluster-local services.

This PR is a follow-up of https://github.com/kubeflow/kfserving/pull/954.

Fixes https://github.com/kubeflow/kfserving/issues/948

### DIFF
--- a/pkg/controller/inferenceservice/resources/istio/virtualservice.go
+++ b/pkg/controller/inferenceservice/resources/istio/virtualservice.go
@@ -266,6 +266,14 @@ func (r *VirtualServiceBuilder) CreateVirtualService(isvc *v1alpha2.InferenceSer
 	// extract the virtual service hostname from the predictor hostname
 	serviceURL := fmt.Sprintf("%s://%s", "http", serviceHostname)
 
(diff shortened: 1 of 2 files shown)


yuzisun:
@akirillov we have addressed this in v1beta1, please help check master if that fixed your issue.

akirillov:
closing, as the issue was fixed in v1beta1.

PR #1055, opened 2020-08-30
Harden model puller logic and add tests

### LINKED ISSUE #1048: Harden Puller logic in the side-car Agent
/kind feature

With the merging of #989 a variety of enhancements are still required before we can remove the [POC] label from the multi-model puller. These enhancements include but are not limited to:
* Handle retries
* Move logger to zapr
* Set proper buffer size
* Handle request logic
* Implement a variety of protocols: HDFS, GCS, FILE, HTTPS
* Handle proper config overrides of s3 clients
* Remove need for timestamp for config diffing by using a bool
* Figure out if mid-download failures can be checkpointed and recovered *optional
* loop through map in a parallel way instead of sequential *optional
* Handle load, unload, load events *optional

* Unit tests
* End to End tests

yuzisun:

Fixes #1048 

1. Improve model config sync logic and add event done channels
2. Mock model downloader and S3 client and add model config sync tests
3. Change to use controller runtime logger

**Release note**:

```release-note

```


njhill:
Hey @yuzisun I actually also had made some similar changes (now opened in https://github.com/kubeflow/kfserving/pull/1057) which look like they overlap to some degree (sorry for the miscommunication, I was a bit vague when I suggested that I might [here](https://github.com/kubeflow/kfserving/pull/989#issuecomment-682012944)).

I can merge/rebase mine once this is in or vice versa, or can just close it if now redundant (it does have some additional simplification/refactoring).

pkg/agent/watcher.go
```diff
@@ -113,7 +120,10 @@ func (w *Watcher) parseConfig(modelConfigs modelconfig.ModelConfigs) {
...
 			w.modelAdded(name, &spec)
 		} else if !cmp.Equal(spec, *existing.Spec) {
-			existing.Spec, existing.stale = &spec, false
+			w.modelTracker[name] = modelWrapper{
+				Spec:  existing.Spec,
+				stale: false,
+			}
 			// Changed - replace
 			w.modelRemoved(name)
 			w.modelAdded(name, &spec)
```
ifilonenko, line 129: Are we keeping this or allowing for updates? I still believe this is v-inefficient

pkg/agent/watcher_test.go
```diff
@@ -0,0 +1,274 @@
...
+						Providers: map[storage.Protocol]storage.Provider{
+							storage.S3: &storage.S3Provider{
+								Client:     &mockS3Client{},
+								Downloader: &mockS3Downloder{},
+							},
+						},
+					},
+				}
+				go puller.processCommands(watcher.ModelEvents)
+				logger.Printf("starting go routing")
```
ifilonenko, line 188: go routine

yuzliu (approved):
/LGTM

PR #1056, opened 2020-08-30
Use github action for kfserving controller tests 

yuzisun:

Fixes #

1. Use github action for KFServing go controller unit tests 

**Release note**:

```release-note

```


.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
```
pugangxa, line 25: Seems need not do this, but I guess this is for vendor mode right?

yuzisun: you are right, this is not needed

Makefile
```diff
@@ -184,6 +184,19 @@ docker-build-storageInitializer:
 docker-push-storageInitializer: docker-build-storageInitializer
 	docker push ${KO_DOCKER_REPO}/${STORAGE_INIT_IMG}
 
+kubebuilder:
```
ellistarn, line 187: Why do we need kb as a dependency? This is for generating controllers, right?

yuzisun: this is for `make test` which requires local apiserver/etcd which is installed part of kubebuilder

PR #1057, opened 2020-08-30
Puller streamlining/simplification

njhill:

Follow-on changes to #989 based on remaining review suggestions.

- Simplified configmap change diffing
- Connect watcher and puller with event channel
- Have puller track in-progress ops per model via op completion channel and tie lifecycle of per-model channel+goroutine pairs to this

### DIFF
--- a/pkg/agent/puller.go
+++ b/pkg/agent/puller.go
@@ -1,73 +1,126 @@
 package agent
 
 import (
-	"fmt"
 	"github.com/kubeflow/kfserving/pkg/agent/storage"
+	v1 "github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1"
 	"log"
 	"path/filepath"
 )
 
+type OpType string
+
+const (
+	Add    OpType = "Add"
+	Remove OpType = "Remove"
+)
+
+// sentinel to represent model absence
+var remove v1.ModelSpec = v1.ModelSpec{}
+
 type Puller struct {
-	ChannelMap map[string]Channel
-	Downloader Downloader
+	channelMap  map[string]ModelChannel
+	completions chan string
+	Downloader  Downloader
 }
 
-type Channel struct {
-	EventChannel chan EventWrapper
+type ModelOp struct {
+	ModelName string
+	Op        OpType
+	Spec      *v1.ModelSpec
 }
 
-func (p *Puller) AddModel(modelName string) Channel {
-	// TODO: Figure out the appropriate buffer-size for this
-	// TODO: Check if event Channel exists
-	eventChannel := make(chan EventWrapper, 20)
-	channel := Channel{
-		EventChannel: eventChannel,
+func StartPuller(downloader Downloader, commands <-chan ModelOp) {
+	puller := Puller{
+		channelMap:  map[string]ModelChannel{},
+		completions: make(chan string, 128),
+		Downloader:  downloader,
 	}
-	go p.modelProcessor(modelName, channel.EventChannel)
-	p.ChannelMap[modelName] = channel
-	return p.ChannelMap[modelName]
+	go func() {
+		var finished bool
+		for {
+			select {
+			case modelOp, ok := <-commands:
+				if !ok {
+					finished = true
+				} else {
+					switch modelOp.Op {
+					case Add:
+						puller.enqueueModelOp(modelOp.ModelName, modelOp.Spec)
+					case Remove:
+						puller.enqueueModelOp(modelOp.ModelName, &remove)
+					}
+				}
+			case completed := <-puller.completions:
+				puller.modelOpComplete(completed, finished)
+			}
+		}
+	}()
 }
 
-func (p *Puller) RemoveModel(modelName string) error {
-	channel, ok := p.ChannelMap[modelName]
+type ModelChannel struct {
+	modelOps    chan *v1.ModelSpec
+	opsInFlight int
+}
+
+func (p *Puller) enqueueModelOp(modelName string, modelSpec *v1.ModelSpec) {
+	modelChan, ok := p.channelMap[modelName]
+	if !ok {
+		modelChan = ModelChannel{
+			modelOps: make(chan *v1.ModelSpec, 8),
+		}
+		go p.modelProcessor(modelName, modelChan.modelOps)
+		p.channelMap[modelName] = modelChan
+	}
+	modelChan.opsInFlight += 1
+	modelChan.modelOps <- modelSpec
+}
+
+func (p *Puller) modelOpComplete(modelName string, closed bool) {
+	modelChan, ok := p.channelMap[modelName]
 	if ok {
-		close(channel.EventChannel)
-		delete(p.ChannelMap, modelName)
+		modelChan.opsInFlight -= 1
+		if modelChan.opsInFlight == 0 {
+			close(modelChan.modelOps)
+			delete(p.channelMap, modelName)
+			if closed && len(p.channelMap) == 0 {
+				// this was the final completion, close the channel
+				close(p.completions)
+			}
+		}
+	} else {
+		//TODO log warning, shouldn't happen
 	}
-	if err := storage.RemoveDir(filepath.Join(p.Downloader.ModelDir, modelName)); err != nil {
-		return fmt.Errorf("failing to delete model directory: %v", err)
-	}
-	return nil
 }
 
-func (p *Puller) modelProcessor(modelName string, events chan EventWrapper) {
+func (p *Puller) modelProcessor(modelName string, ops <-chan *v1.ModelSpec) {
 	log.Println("worker for", modelName, "is initialized")
 	// TODO: Instead of going through each event, one-by-one, we need to drain and combine
 	// this is important for handling Load --> Unload requests sent in tandem
 	// Load --> Unload = 0 (cancel first load)
 	// Load --> Unload --> Load = 1 Load (cancel second load?)
-	for event := range events {
-		log.Println("worker", modelName, "started  job", event)
-		switch event.LoadState {
-		case ShouldLoad:
-			log.Println("Should download", event.ModelSpec.StorageURI)
-			err := p.Downloader.DownloadModel(event)
+	for modelSpec := range ops {
+		if modelSpec != &remove {
+			// Load
+			log.Println("Should download", modelSpec.StorageURI)
+			err := p.Downloader.DownloadModel(modelName, modelSpec)
 			if err != nil {
-				log.Println("worker failed on", event, "because: ", err)
+				log.Println("Download of model", modelName, "failed because: ", err)
 			} else {
 				// If there is an error, we will NOT send a request. As such, to know about errors, you will
 				// need to call the error endpoint of the puller
 				// TODO: Do request logic
-				log.Println("Now doing a request on", event)
+				log.Println("Now doing load request for", modelName)
 			}
-		case ShouldUnload:
+		} else {
+			// Unload
 			// TODO: Do request logic
-			log.Println("Now doing a request on", event)
+			log.Println("Now doing unload request for", modelName)
 			// If there is an error, we will NOT do a delete... that could be problematic
-			log.Println("Should unload", event.ModelName)
-			if err := p.RemoveModel(event.ModelName); err != nil {
-				log.Println("worker failed on", event, "because: ", err)
+			log.Println("Should unload", modelName)
+			if err := storage.RemoveDir(filepath.Join(p.Downloader.ModelDir, modelName)); err != nil {
+				log.Printf("failing to delete model directory: %v", err)
 			}
 		}
+		p.completions <- modelName
 	}
 }
--- a/pkg/agent/puller.go
+++ b/pkg/agent/puller.go
@@ -14,12 +14,9 @@ const (
 	Remove OpType = "Remove"
 )
 
-// sentinel to represent model absence
-var remove v1.ModelSpec = v1.ModelSpec{}
-
 type Puller struct {
 	channelMap  map[string]ModelChannel
-	completions chan string
+	completions chan *ModelOp
 	Downloader  Downloader
 }
 
@@ -32,7 +29,7 @@ type ModelOp struct {
 func StartPuller(downloader Downloader, commands <-chan ModelOp) {
 	puller := Puller{
 		channelMap:  make(map[string]ModelChannel),
-		completions: make(chan string, 128),
+		completions: make(chan *ModelOp, 4),
 		Downloader:  downloader,
 	}
 	go puller.processCommands(commands)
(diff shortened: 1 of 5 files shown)


njhill:
This overlaps a bit with #1055 but I had already finished the changes so opening for consideration.

cc @ifilonenko @yuzisun

yuzisun:
Thanks @njhill! This looks great! yes there is some overlap I think we can reconcile on this PR and then I can rebase once yours is merged in. I have some other refactoring regarding testing.

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
+)
+
+// sentinel to represent model absence
+var remove v1.ModelSpec = v1.ModelSpec{}
+
 type Puller struct {
-	ChannelMap map[string]Channel
-	Downloader Downloader
+	channelMap  map[string]ModelChannel
+	completions chan string
```
yuzisun, line 22: I think it is useful to know the command add or remove on the completion channel(good for me for asserting in tests),  in my PR I also included `error` field to track the download failures but now I think there is a common pattern to have separate error channel. WDYT?

pkg/agent/watcher.go
```diff
@@ -84,78 +86,47 @@ func (w *Watcher) Start() {
...
-					ModelName:      modelName,
-					ModelSpec:      nil,
-					LoadState:      ShouldUnload,
-					ShouldDownload: false,
-				}
-				log.Println("Sending event", event)
-				channel.EventChannel <- event
-			}
+	for name, wrapper := range w.modelTracker {
+		if wrapper.stale {
```
yuzisun, line 109: I converted the modelconfig to a map and use that to find the removed ones, I think this works too and slightly more efficient but a bit hard to follow the logic.

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
+				switch modelOp.Op {
+				case Add:
+					p.enqueueModelOp(modelOp.ModelName, modelOp.Spec)
+				case Remove:
+					p.enqueueModelOp(modelOp.ModelName, &remove)
+				}
+			}
+		case completed := <-p.completions:
+			p.modelOpComplete(completed, finished)
+		}
```
yuzisun, line 60: I liked this a lot!

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
+}
+
+func (p *Puller) modelOpComplete(modelName string, closed bool) {
+	modelChan, ok := p.channelMap[modelName]
+	if ok {
+		modelChan.opsInFlight -= 1
+		if modelChan.opsInFlight == 0 {
+			close(modelChan.modelOps)
+			delete(p.channelMap, modelName)
+			if closed && len(p.channelMap) == 0 {
```
yuzisun, line 89: Does this mean we close the completion channel when all model events are processed? how do we create this channel again as I saw it is only created when starting the puller.

pkg/agent/watcher.go
```diff
@@ -84,78 +86,47 @@ func (w *Watcher) Start() {
...
+		existing, exists := w.modelTracker[name]
+		if !exists {
+			// New - add
+			w.modelTracker[name] = modelWrapper{Spec: &spec}
+			w.modelAdded(name, &spec)
+		} else if !cmp.Equal(spec, *existing.Spec) {
+			existing.Spec, existing.stale = &spec, false
+			// Changed - replace
+			w.modelRemoved(name)
+			w.modelAdded(name, &spec)
```
yuzisun, line 105: Should this be just Added/Reload ? Remove then Add triggers unload and load which would cause some downtime of the model on the model server side?

njhill: My thinking was that if TrainedModels are immutable (as I think was agreed), this case is unlikely to occur since it would imply the TrainedModel CR itself had been deleted and another with the same name subsequently created. In that case continuity would not be expected anyhow.

pkg/agent/puller.go
njhill: This is just for clean teardown of the channels but I guess likely won't be needed since I assume the watcher will keep running until the container is going down anyhow.

The idea is that the ModelEvents channel would first be closed and then this logic would ensure the completions channel is closed as soon as the last in-progress/queued operation completes after that.

pkg/agent/watcher.go
njhill: I was aiming to minimze iterations over the map/slices (i.e. 1 pass over each). Agree the way that the stale flag is toggled could be a bit confusing to read... let me add a couple more comments and if it's still not clear then could change to use an intermediate map

pkg/agent/puller.go
njhill: Right, I hadn't included proper handling of the error cases in this pass, was thinking that could be done in a next iteration :) Since there's quite a lot to consider and I expect it will require some adjustments of the state management... maybe something like making the per-model goroutines more like reconciliation loops.

Having the completion events be/include the original op for now sounds like a good idea though!

pkg/agent/watcher.go
```diff
@@ -84,78 +86,47 @@ func (w *Watcher) Start() {
...
-				}
-				log.Println("Sending event", event)
-				channel.EventChannel <- event
-			}
+			wrapper.stale = true // reset for next iteration
 		}
 	}
 }
+
+func (w *Watcher) modelAdded(name string, spec *v1beta1.ModelSpec) {
```
yuzliu, line 119: Oh, this is sweet.

pkg/agent/watcher.go
```diff
@@ -47,7 +41,14 @@ func (w *Watcher) Start() {
 		log.Fatal(err)
 	}
 	defer watcher.Close()
-
+	if err = watcher.Add(w.configDir); err != nil {
+		log.Fatal(err)
+	}
+	// Add a first create event to the channel to force initial sync
```
yuzliu, line 47: Is the intention of this initial event to replace ```syncer.Start()``` in main.go?

pkg/agent/puller.go
yuzliu: I am still a little confused about why we need to close the model channel and completion channel whenn the in-flight model operation is 0. Shouldn't we only close those channels when a model is removed?

pkg/agent/watcher.go
njhill: Not exactly, the syncer logic still runs first but moved into `NewWatcher` function above since I thought it kind of made sense to be part of the watcher. It just populates the initial state of the `modelTracker` map.

Injecting the event here just forces an intial sync of the map with the configmap. It's done after the fsnotify watch is started to avoid missing any changes, otherwise in theory at least we could sync the configmap, then it changes, then the watch starts ... and so we've missed that change in between.

Oh hmm though I just realized that this might not work if the Events chan is not buffered, I guess it may hang here if so. But that should be fixable by just moving this to right after the goroutine below is started.

pkg/agent/puller.go
njhill: This is my pretty much my first time writing go so I'm not sure of certain best practices etc. please don't hesitate to point out things that look odd/wrong in general :)

The idea here is to keep the goroutine and channel around only as long as it's needed to serialize multiple events that might be produced for a given model (most likely case model is removed while still in the process of being added). Once the last one of those finishes then they are cleaned up, and recreated on the next event. Otherwise if you had say 1000 models loaded, you would have the same number of channels and goroutines just sitting around doing nothing. But maybe that's not a big deal?

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
 	"github.com/kubeflow/kfserving/pkg/agent/storage"
+	v1 "github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1"
 	"log"
 	"path/filepath"
 )
 
+type OpType string
+
+const (
+	Add    OpType = "Add"
```
ifilonenko, line 13: Does `Add` include `Update`? That was the initial design.
I remember you mentioned assuming that `TrainedModel` objects should be treated as immutable. 
But i disagreed. Example: 
```
TrainedModel 
name: foo
resource: 1g
storage: s3://here/1
```
if i modify resource I want to send an update command while not re-downloading. If i want to modify storage to:
`s3://there/1` and then change it back to `s3://here/1` I don't want to re-download `s3://here/1`. Example, what if I mis-type something, don't want to delete, would rather only delete up on `TrainedModel` deletion

pkg/agent/puller.go
ifilonenko: I 100% would agree on a seperate error channel. I think that beyond for testing it would be important for the error status endpoint.

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
-func (p *Puller) AddModel(modelName string) Channel {
-	// TODO: Figure out the appropriate buffer-size for this
-	// TODO: Check if event Channel exists
-	eventChannel := make(chan EventWrapper, 20)
-	channel := Channel{
-		EventChannel: eventChannel,
+func StartPuller(downloader Downloader, commands <-chan ModelOp) {
+	puller := Puller{
+		channelMap:  make(map[string]ModelChannel),
+		completions: make(chan string, 128),
```
ifilonenko, line 35: Why `128`, is there a reason for this buffer-size? Also, can we move that to a constant

pkg/agent/puller.go
ifilonenko: I would consider not to close the channel after all in-flight events are done. I'd consider keeping it open and only close when the model is removed, as yuzhui mentioned above. Bringing up and down a channel when in-flights finish and a new one comes in doesn't seem to be most efficient. But then again, do any go experts know the cost of doing that?

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
+		if modelChan.opsInFlight == 0 {
+			close(modelChan.modelOps)
+			delete(p.channelMap, modelName)
+			if closed && len(p.channelMap) == 0 {
+				// this was the final completion, close the channel
+				close(p.completions)
+			}
+		}
+	} else {
+		//TODO log warning, shouldn't happen
```
ifilonenko, line 95: Can we add this, any reason to leave as TODO?

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
 	}
-	return nil
 }
 
-func (p *Puller) modelProcessor(modelName string, events chan EventWrapper) {
+func (p *Puller) modelProcessor(modelName string, ops <-chan *v1.ModelSpec) {
```
ifilonenko, line 99: The idea to make the channel able to receive and send events is so that I can handle retries (by writing back to the same channel), is there a strategy to handle retries? I guess writing back to same channel is a bad design, maybe?

pkg/agent/puller.go
```diff
@@ -1,73 +1,130 @@
...
+			// Unload
 			// TODO: Do request logic
-			log.Println("Now doing a request on", event)
+			log.Println("Now doing unload request for", modelName)
 			// If there is an error, we will NOT do a delete... that could be problematic
-			log.Println("Should unload", event.ModelName)
-			if err := p.RemoveModel(event.ModelName); err != nil {
-				log.Println("worker failed on", event, "because: ", err)
+			log.Println("Should unload", modelName)
+			if err := storage.RemoveDir(filepath.Join(p.Downloader.ModelDir, modelName)); err != nil {
```
ifilonenko, line 124: Should we not keep this logic in a method that will probably be expanded upon, or is the extent of the storage deleting a model, going to be deleting the `dir`

pkg/agent/watcher.go
ifilonenko: Im confused by this events add... I though we should be looping through the modelDir upon start, not the `..data`. The reason is that: maybe the configmap hasn't been mounted yet, but the modelDir is still filled up.

pkg/agent/watcher.go
ifilonenko: As I said above, I think that we should handle updates to the TrainedModel.

pkg/agent/puller.go
njhill: > `TrainedModel` objects should be treated as immutable.

I thought there was some agreement on this on the last call we had. I feel fairly strongly that it's needed, it is analogous to why pods are immutable. If you change a pod spec the pod does not reconfigure itself, a new pod with a new unique name replaces it, orchestrated via replica sets. This goes hand-in-hand with the proposal for InferenceRouters to be the vehicle for versioning and stable inferencing endpoints.

So it's important to close on this one way or the other I think since it has significant design implications.

cc @yuzisun @ellistarn @yuzliu

pkg/agent/puller.go
njhill: Kind of arbitrary, I just wanted sending to never block. It could be made much smaller now that there's a dedicated goroutine for managing the channel map.

pkg/agent/puller.go
ifilonenko: I might have missed that, I am willing to be convinced otherwise :) I just think that the creation of a new model, of the same name would be common if users update their `TrainedModel`. Example: users update InferenceService when trying to patch their version, they would probably want assume the same behavior for the `TrainedModel`s.

pkg/agent/watcher.go
njhill: I didn't think there would necessarily be such an event. The watched files (symlink) will be there from the outset and so I don't _think_ will necessarily produce any create/modification fsnotify events.

pkg/agent/puller.go
njhill: Probably yes, maybe even as a function on the Downloader for symmetry. I can't envisage what else would be done for this apart from deleting the model's dir, but it will probably be important to ensure that the "success" file is deleted first before the rest of it.

pkg/agent/puller.go
njhill: Right, I agree that appears kind of inconsistent with being able to directly change the `storageUri` of a single-model InferenceService. But it ultimately comes down to the fact that there isn't any indirection of the model names so there's no way that this could be implemented non-disruptively in the general case. Changing the `storageUri` like that would entail a small outage, possibly indefinite if there's some issue with the new URI/model data, which is a non-starter i.m.o.

pkg/agent/watcher.go
njhill: I don't think I tested with fsnotify, but I had experimented with a different inotify-based lib where from what I recall there wasn't an event to reflect the initial state. That's cool though, if it's not needed then this line can be removed. I'll try to also test when I get a chance

ifilonenko: yeah before a lgtm, I will run this in a docker container and do a test. lmk when you have tested so I can do a follow-up as well

pkg/agent/puller.go
```diff
@@ -1,73 +1,121 @@
...
-		delete(p.ChannelMap, modelName)
+func (p *Puller) processCommands(commands <-chan ModelOp) {
+	// channelMap accessed only by this goroutine
+	for {
+		select {
+		case modelOp, ok := <-commands:
+			if ok {
+				p.enqueueModelOp(&modelOp)
+			} else {
+				commands = nil
```
yuzisun, line 46: I think we can save the closed channel as bool variable here ? I think setting channel as nil is not a good practice, as sending event to a nil channel blocks forever while sending event to a closed channel results in panic.

yuzisun:
@njhill  I made a comment on setting the channel to nil see https://dave.cheney.net/2014/03/19/channel-axioms, I think we can merge this after that is addressed.

pkg/agent/puller.go
njhill: @yuzisun the use here was deliberate. I had the behaviour of nil channels in mind, in particular as a way to disable the corresponding branch of the `select` after the channel is closed by the sender. See for example https://medium.com/justforfunc/why-are-there-nil-channels-in-go-9877cc0b2308.

The link you gave describes the properties of nil channels but I don't see anywhere that says it's bad practice to use them, I assume they were designed that way for a reason.

Happy to change it if you would still like me to or alternatively I could add a comment to make things a bit more explicit?

PR #1058, opened 2020-08-31
KFServing v1beta1 API (trigger e2e tests)

yuzliu:
Trigger e2e tests for v1beta1 branch

**Release note**:

```release-note

```


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -0,0 +1,98 @@
...
+		Name:      isvc.Name + "-" + string(v1beta1.PredictorComponent),
+		Namespace: isvc.Namespace,
+		Labels: utils.Union(isvc.Labels, map[string]string{
+			constants.InferenceServicePodLabelKey: isvc.Name,
+			constants.KServiceComponentLabel:      string(v1beta1.PredictorComponent),
+		}),
+		Annotations: annotations,
+	}
+	if isvc.Spec.Predictor.CustomPredictor == nil {
+		container := predictor.GetContainer(isvc.ObjectMeta, isvc.Spec.Predictor.GetExtensions(), p.inferenceServiceConfig)
```
ellistarn, line 74: nit inline?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -0,0 +1,98 @@
...
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						*container,
+					},
+				},
+			},
+		}
+	}
+	// Here we allow switch between knative and vanilla deployment
+	r := knative.NewKsvcReconciler(p.client, p.scheme, objectMeta, &isvc.Spec.Predictor.ComponentExtensionSpec,
```
ellistarn, line 86: Can you use DI instead of reinstantiating here? Avoids recreatnig clients.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -0,0 +1,98 @@
...
+		&isvc.Spec.Predictor.CustomPredictor.Spec)
+
+	if err := controllerutil.SetControllerReference(isvc, r.Service, p.scheme); err != nil {
+		return err
+	}
+	if status, err := r.Reconcile(); err != nil {
+		return err
+	} else {
```
ellistarn, line 94: I'd drop this else block. Makes changing this logic awkward. Cleaner to just use common if err patterns

PR #983, opened 2020-07-30
Make KFServer HTTP requests asynchronous

salanki:
This is a re-open of #791. Please check discussion in #791.

**What this PR does / why we need it**:

Requests from KFServer are currently block. This leads to a transformer not being able to make more than number of Tornado workers concurrent requests to the predictor.

As predictors are usually the bottleneck, and transformers are fast, the transformer becomes an unnecessarily bottleneck.

This PR applies Tornado best practices, making HTTP requests async. This drastically improves the transformer performance.

**Before**
Here we scale transformer instances with 40 worker threads each, to handle a total of 300 requests
Screen Shot 2020-04-21 at 14 40 29

**After**
On one transformer instance, with 1 thread. The target number of concurrent requests is much quicker achieved on the predictor, while consuming less resources.
Screen Shot 2020-04-21 at 14 39 08

**Which issue(s) this PR fixes** *
Fixes #789

**Special notes for your reviewer**:

**Release note**:
```release-note
Improve transformer concurrency and performance by making Asynchronous HTTP calls.
```

PR #1069, opened 2020-09-02
Fix thread safety issue in KFServer asyncio

salanki:
#983 got merged while me and @yuzisun were nailing down the sporadic build issues. There seems to be a lot of situations where multi process workers break with Tornado in Python 3.5+ due to asyncio not being thread (process) safe. These issues are realized with the async-mangling needed to get the synchronous Alibi implementation to work with the now async KFServer.

The recommendation from the Tornado team seems to be to run only one process and handle load balancing elsewhere. Instead of making that change, as I do not know what impact it would have on other usages of KFServer I'm trying to shift the IOLoop and process fork order to make sure all IOLoop actions happen after the fork.

### DIFF
--- a/python/alibiexplainer/alibiexplainer/__main__.py
+++ b/python/alibiexplainer/alibiexplainer/__main__.py
@@ -17,7 +17,6 @@ import kfserving
 import logging
 import os
 import sys
-import nest_asyncio
 from alibiexplainer import AlibiExplainer
 from alibiexplainer.explainer import ExplainerMethod  # pylint:disable=no-name-in-module
 from alibiexplainer.parser import parse_args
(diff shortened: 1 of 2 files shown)


ellistarn:
Feel free to submit a PR for membership. 
/ok-to-test

salanki:
I’m working on this as we speak.

salanki:
Running the test a couple of times to make sure it's stable.
/retest

salanki:
/retest

Out of GPUs

salanki:
This seems stable now.

salanki:
@yuzisun: Ready for review now.

yuzisun:
We might want to consider default worker=1,  multi-processing is quite tricky to get right in a few cases.

/lgtm
/approve

salanki:
> We might want to consider defaulting worker=1, multi-processing is quite tricky to get right in a few cases.

Yes, that is what I was thinking initially and would have been a much simpler solution. What I do not know if people who use KFServer for CPU based inference rely on the multiple workers spun up by default for concurrency.

PR #1077, opened 2020-09-03
Support Kubernetes 1.18 (Update controller-gen and crds accordingly)

### LINKED ISSUE #1070: Cannot deploy kf-serving on kubernetes 1.18
/kind bug

```
make deploy
/Users/pugang/go/bin/controller-gen "crd:maxDescLen=0" paths=./pkg/apis/serving/v1alpha2/... output:crd:dir=config/crd
/Users/pugang/go/bin/controller-gen rbac:roleName=kfserving-manager-role paths=./pkg/controller/inferenceservice/... output:rbac:artifacts:config=config/rbac
............
issuer.cert-manager.io/selfsigned-issuer unchanged
validatingwebhookconfiguration.admissionregistration.k8s.io/inferenceservice.serving.kubeflow.org configured
The CustomResourceDefinition "inferenceservices.serving.kubeflow.org" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[default].properties[explainer].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[default].properties[predictor].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[default].properties[transformer].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[canary].properties[transformer].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[canary].properties[explainer].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[canary].properties[predictor].properties[custom].properties[container].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```
Can be installed successfully.

**Environment:**
```
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

pugangxa:
Update controller-gen to add default value for protocol. CRDs are updated accordingly too. 
Notice we must add "preserveUnknownFields=false" to spec or another error will be reported, see https://github.com/kubernetes-sigs/controller-tools/issues/476

Fixes #1070

### DIFF
--- a/config/crd/serving.kubeflow.org_trainedmodels.yaml
+++ b/config/crd/serving.kubeflow.org_trainedmodels.yaml
@@ -1,23 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.1-0.20200528125929-5c0c6ae3b64b
+    controller-gen.kubebuilder.io/version: v0.4.1-0.20200901165317-c7154a3a4603
   creationTimestamp: null
   name: trainedmodels.serving.kubeflow.org
 spec:
-  additionalPrinterColumns:
-  - JSONPath: .status.url
-    name: URL
-    type: string
-  - JSONPath: .status.conditions[?(@.type=='Ready')].status
-    name: Ready
-    type: string
-  - JSONPath: .metadata.creationTimestamp
-    name: Age
-    type: date
   group: serving.kubeflow.org
   names:
     kind: TrainedModel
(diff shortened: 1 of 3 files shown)


config/crd/serving.kubeflow.org_trainedmodels.yaml
```diff
@@ -1,23 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
yuzisun, line 3: how do you generate with v1 crd?

pugangxa: Now controller-gen will generate v1 crd by default, unless "crdVersions=v1beta1" is added. I'm not sure in Makefile some places have "$(CRD_OPTIONS)" followed "$(CONTROLLER_GEN)" but some does not. I think we can add "$(CRD_OPTIONS)" for all the places?

pugangxa:
/hold
Unfortunately it still does not work for if I use v1beta1. I can generate v1 CRD and deploy, or generate v1beta1 CRD by removing lines in $GOPATH/pkg/mod/k8s.io/api@v0.18.6/core/v1/types.go

// +listType=map
// +listMapKey=containerPort
// +listMapKey=protocol

I think the fix is still not complete by controller-tool team, will append for their awareness.

pugangxa:
So based on my test, move to new version of controller-gen  while keep ```apiVersion: apiextensions.k8s.io/v1beta1``` does not work. Since the defaulter is not allowed in v1beta1 so it's treated as unknown fields, the defaulter feature require "preserveUnknownFields=false" so that default values can be applied but during apply the new error like below is reported.
```
The CustomResourceDefinition "inferenceservices.serving.kubeflow.org" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[predictor].properties[triton].properties[ports].items.properties[protocol].default: Forbidden: must not be set
...........
```

So the solution to this issue can be: 
1: Just pinned the controller-gen to the new version "c7154a3a46038eaffeef9ff8d37bbc775728c60d" which has added the default value "TCP" for protocol, no need to do any other change.
This works for me but the CRD version changed to v1 which is the new default of controller-tools.
2: Still use old controller-gen and edit the generated CRD and add protocol as required(as we already do using ```perl -pi -e```(need test more)
3: Wait the solution from community, though I'm not sure whether there will be one since for k8s 1.18 the api-version moved to v1 and the new version of controller-tools seems works by design...

I think the correct way is to move to v1 for "apiVersion: apiextensions.k8s.io", but I do not have all the supported versions of k8s clusters so not sure whether it will break previous deployment. WDYT @yuzisun

pugangxa:
The code changes now works on K8S 1.18, but I'm not sure whether it works on previous versions.

jinchihe:
> The code changes now works on K8S 1.18, but I'm not sure whether it works on previous versions.

I think the CI test is 1.16, which can ensure the previous versions.

pugangxa:
Yes, though I cannot see the details of the build log(something wrong?), I think it cannot work since defaulter is a new feature. So maybe we must add protocol as required for now to work on all editions.

yuzisun:
thanks @pugangxa for looking into this!  I think kubernetes 1.16+ supports v1 crd and latest knative already requires minimal kubernetes version to be 1.16, maybe we can just make that requirement for kfserving too when we release KFServing v1beta1 in 0.5 release.

yuzisun:
@pugangxa Can you help rebase? What do you think we go for CRD v1 as knative minimal requirement is already 1.17...

pugangxa:
> @pugangxa Can you help rebase? What do you think we go for CRD v1 as knative minimal requirement is already 1.17...

I agree. I am on vacation and will try this once I get back.

PR #1078, opened 2020-09-03
support generic filename for ONNX runtime model (#1004)

### LINKED ISSUE #1004: Generic naming for ONNX model objects
/kind feature

Currently in order to turn an ONNX runtime model into a servable, the file must have the name `model.onnx`, [here](https://github.com/kubeflow/kfserving/blob/ff02ea49b399565a8abd4a556c75e1fbb003a6e2/pkg/apis/serving/v1alpha2/framework_onnx.go#L25). This poses restrictions when multiple models are stored in the same directory, or with the potential extension of a `storageUri` that exists as a path over the internet. For example, the models in the ONNX [model zoo](https://github.com/onnx/models) are not directly servable under the current structure.   

It would be good to extend the specification & downloading of ONNX models to have an arbitrary name. Whether this can be specified with another field in the CRD (like Pytorch does with `modelClassName`, or via some other mechanism is an open question.

tduffy000:
Add the support for parsing the `storageUri` in an `InferenceService` CRD to take in a generic filename such that the below:
```yaml
apiVersion: serving.kubeflow.org/v1alpha2
kind: InferenceService
metadata:
  name: onnx-from-uri
spec:
  default:
    predictor:
      onnx:
        storageUri: s3://some_bucket/some/key/path/my_great_model.onnx
```
will also be valid. Here, if the path does not specify a file name, e.g. `s3://some_bucket/some/key/path`, then we'd continue to use the default `model.onnx`.

Fixes #1004 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/serving/v1alpha2/framework_onnx.go
+++ b/pkg/apis/serving/v1alpha2/framework_onnx.go
@@ -14,15 +14,20 @@ limitations under the License.
 package v1alpha2
 
 import (
+	"net/url"
+	"path"
+
+	"strconv"
+
 	"github.com/kubeflow/kfserving/pkg/constants"
 	v1 "k8s.io/api/core/v1"
-	"strconv"
 )
 
 var (
 	ONNXServingRestPort            = "8080"
 	ONNXServingGRPCPort            = "9000"
-	ONNXModelFileName              = "model.onnx"
+	DefaultONNXModelName           = "model.onnx"
+	ONNXFileExt                    = "onnx"
 	InvalidONNXRuntimeVersionError = "ONNX RuntimeVersion must be one of %s"
 )
 
@@ -36,8 +41,15 @@ func (s *ONNXSpec) GetResourceRequirements() *v1.ResourceRequirements {
 }
 
 func (s *ONNXSpec) GetContainer(modelName string, parallelism int, config *InferenceServicesConfig) *v1.Container {
+	uri, _ := url.Parse(s.StorageURI)
+	var filename string
+	if path.Ext(uri.Path) != ONNXFileExt {
+		filename = DefaultONNXModelName
+	} else {
+		filename = path.Base(uri.Path)
+	}
 	arguments := []string{
-		"--model_path", constants.DefaultModelLocalMountPath + "/" + ONNXModelFileName,
+		"--model_path", constants.DefaultModelLocalMountPath + "/" + filename,
 		"--http_port", ONNXServingRestPort,
 		"--grpc_port", ONNXServingGRPCPort,
 	}
--- a/pkg/apis/serving/v1alpha2/framework_onnx.go
+++ b/pkg/apis/serving/v1alpha2/framework_onnx.go
@@ -28,6 +28,7 @@ var (
 	ONNXServingRestPort            = "8080"
 	ONNXServingGRPCPort            = "9000"
 	ONNXFileExt                    = ".onnx"
+	DefaultONNXFileName            = "model.onnx"
 	InvalidONNXRuntimeVersionError = "ONNX RuntimeVersion must be one of %s"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/apis/serving/v1alpha2/framework_onnx.go
```diff
@@ -36,8 +41,15 @@ func (s *ONNXSpec) GetResourceRequirements() *v1.ResourceRequirements {
 }
 
 func (s *ONNXSpec) GetContainer(modelName string, parallelism int, config *InferenceServicesConfig) *v1.Container {
+	uri, _ := url.Parse(s.StorageURI)
```
tduffy000, line 44: I'm wondering what the best way to handle `err` is here.

yuzisun: I think the validation logic can be in the `Validate` and here we can safely extract the filename
`func (s *ONNXSpec) Validate(config *InferenceServicesConfig) error `.

For onnx the ext should always be `.onnx ` ?

yuzisun:
@rakelkar to review also

pkg/apis/serving/v1alpha2/framework_onnx.go
tduffy000: it's possible that someone does 
```yaml
storageUri: s3://some_bucket/some/key/path/
```
too though right?

rakelkar:
sweet! this is awesome @tduffy000 

/approve

rakelkar:
@yuzisun  can take a look so that the ONNX example can be fixed?

pkg/apis/serving/v1alpha2/framework_onnx.go
```diff
@@ -36,8 +41,15 @@ func (s *ONNXSpec) GetResourceRequirements() *v1.ResourceRequirements {
...
+	var filename string
+	if path.Ext(uri.Path) != ONNXFileExt {
```
yuzisun, line 46: I think we can validate if the file ext is `.onnx` otherwise reject and `DefaultONNXModelName` is no longer needed?

yuzisun (commented):
@tduffy000 Thanks for the change, can you help port this to v1beta1 as well ?

pkg/apis/serving/v1alpha2/framework_onnx.go
tduffy000: I agree that would probably be more robust, but wouldn't that be a breaking change here https://github.com/kubeflow/kfserving/blob/67e3d1ac21da0ba5867ef2a88c4ff78ea35a6d94/docs/samples/pytorch/pytorch.yaml#L9? It's possible that people have not specified the actual file name in blob storage in their CRD but it still works at the present moment because the storage initializer will then list them out.

tduffy000:
@yuzisun sure can also port over to v1beta1 once we're happy with the changes on v1alpha2.

pkg/apis/serving/v1alpha2/framework_onnx.go
yuzisun: My concern is that if user specified  some other ext e.g `gs://kfserving-samples/models/pytorch/cifar10/model.pkl` but we silently convert to model.onnx and in this case the load is going to fail for sure, so we may want to give that error up front.

tduffy000: That sounds reasonable to me, better to give users helpful and clear error messages. The other case to consider is with the generic URI functionality, someone could also place their onnx model in a `zip` file:
```yaml
storageUri: s3://some_bucket/some/key/path/onnx_model.zip
```
or in a tarball. Do we want to reserve those MIME types to the models that require multiple artifacts (a la Tensorflow & Pytorch)?

yuzisun: Currently S3 or GCS like object storage all does not support zip files in KFServing as it is not a common pattern to store models that way. We support zip for http in your last PR more for the reason that  http is not object store and it is also only for experiment and education purpose. I feel here we can validate if there is file ext then it needs to be `.onnx`, otherwise it is considered as model store and expect `model.onnx` under it, WDYT?

yuzisun:
> @yuzisun sure can also port over to v1beta1 once we're happy with the changes on v1alpha2. Is that okay to place in this PR or better off in a separate one?

yes separate PR for v1beta1, thanks!

pkg/apis/serving/v1alpha2/framework_onnx.go
tduffy000: Totally agree, just want to make sure I'm not introducing anything that leads to downstream frustrations for folks. I'll update it, thanks :smile: .

tduffy000: updated in https://github.com/kubeflow/kfserving/pull/1078/commits/85d2a688113425c0b80a5b4c5dbe3b7763131c3e

yuzisun:
Great work always :), please help create another PR to v1beta1, thanks @tduffy000 ! 

/lgtm

PR #1087, opened 2020-09-12
[POC] Make inline podspec concrete instead of a pointer

ellistarn:
Resolves an API issue with embedded pointer types.

1. Custom specs are now `json:",inline,omitempty"`
2. Validation logic now only includes custom if containers are specified


ellistarn:
This is merged into Dan's PR

PR #1105, opened 2020-09-27
Add GCS support and tests

### LINKED ISSUE #1048: Harden Puller logic in the side-car Agent

mszacillo:
This PR adds GCS Protocol support for downloading models. 

Fixes #1048 

**Overview**:
I've added GCS as an accepted protocol to provider.go, along with the necessary model download logic using gcs which resides in gcs.go. For unit testing, I needed to find a way to mock the gcs client. The best resource I could find was googleapis/google-cloud-go-testing, but when I tried adding this library as a dependency, there was a conflict that prevented the project from pulling it. Because of this, I added in some interface classes, which I use to adapt the gcs client to a general type, which can be used for mocking. Test coverage for the gcs class is currently at 75%, but this can be increased with a couple more tests. 

Note: There were conflicts when importing google.cloud.com/go/storage alongside the storage package within agent. Due to this, I renamed the storage class to kfstorage (but I'm happy to alter this name to something else more fitting if needed). Also noticed that there are a bunch of conflicts with #1055 - so I can resolve those after that PR merges. :) 

**Testing**

For testing, I deployed the puller pod manually and checked the logs to make sure that the models were being successfully downloaded from my GCS bucket. I did this using the following yaml:

```
apiVersion: v1
kind: Pod
metadata:
  name: puller
spec:
  serviceAccountName: default
  restartPolicy: Never
  containers:
  - name: job
    image: mszacillo/korepo:gcssupporttest
    imagePullPolicy: Always
    resources:
      limits:
        cpu: "1"
        memory: 1Gi
      requests:
        cpu: "1"
        memory: 1Gi
    command:
    - /agent
    args:
    - -config-dir
    - /mnt/configs
    volumeMounts:
    - name: config-volume
      mountPath: /mnt/configs
    - name: models-volume
      mountPath: /mnt/models
    - name: gcs-serviceaccount-volume
      mountPath: /gcssecret
  volumes:
    - name: config-volume
      configMap:
        name: models-config
    - name: models-volume
      emptyDir: {}
    - name: gcs-serviceaccount-volume
      configMap:
        name: gcs-config
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: models-config
data:
  models.json: |
    [
      {
        "modelName": "model1",
        "modelSpec": {
          "storageUri": "gs://mskfservingmodels/iris.py",
          "framework": "sklearn",
          "memory": "1G"
        }
      }
    ]
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: gcs-config
data:
  gcloud-application-credentials.json: |
     { *** credential info removed *** }
```

After the puller pod comes up, I can see the following logs print which show that the download was successful, however, I will also run a full e2e test with the kfserving setup. Below you can see a failure when trying to connect to localhost:8080, this is because I wasn't proxyporting the ingress gateway.

```
{"level":"info","ts":1606770594.7795548,"logger":"modelAgent","msg":"Initializing model agent with","config-dir":"/mnt/configs","model-dir":"/mnt/models"}
{"level":"info","ts":1606770594.7796307,"logger":"modelAgent","msg":"Setting credential file for GCS."}
{"level":"info","ts":1606770594.7796636,"logger":"modelAgent","msg":"Initializing gcs client, using existing GOOGLE_APPLICATION_CREDENTIALS variable."}
{"level":"info","ts":1606770595.8202243,"logger":"Syncer","msg":"Syncing model directory..","modelDir":"/mnt/models"}
&{mskfservingmodels [{project-owners-810000861461  OWNER   0xc00052d9c0} {project-editors-810000861461  OWNER   0xc00052da00} {project-viewers-810000861461  READER   0xc00052da20}] {false 0001-01-01 00:00:00 +0000 UTC} {false 0001-01-01 00:00:00 +0000 UTC} [{project-owners-810000861461  OWNER   0xc00052da40} {project-editors-810000861461  OWNER   0xc00052da80} {project-viewers-810000861461  READER   0xc00052daa0}] false   US 1 STANDARD 2020-11-30 18:28:08.213 +0000 UTC false map[] false {[]} <nil> [] <nil> <nil> <nil> CAE= multi-region}
{"level":"info","ts":1606770595.8207524,"logger":"modelWatcher","msg":"adding model","modelName":"model1"}
{"level":"info","ts":1606770595.8211873,"logger":"Watcher","msg":"Start to watch model config event"}
{"level":"info","ts":1606770595.8213172,"logger":"Watcher","msg":"Watching","modelConfig":"/mnt/data/models.json"}
{"level":"info","ts":1606770595.8212771,"logger":"modelProcessor","msg":"worker is started for","model":"model1"}
{"level":"info","ts":1606770595.8214014,"logger":"modelProcessor","msg":"Downloading model","storageUri":"gs://mskfservingmodels/iris.py"}
{"level":"info","ts":1606770595.8214495,"logger":"Downloader","msg":"Downloading to model dir","modelUri":"gs://mskfservingmodels/iris.py","modelDir":"/mnt/models"}
{"level":"info","ts":1606770595.8214915,"logger":"Downloader","msg":"Success file does not exist"}
{"level":"info","ts":1606770595.82158,"logger":"modelAgent","msg":"Downloading model ","modelName":"model1","storageUri":"gs://mskfservingmodels/iris.py","modelDir":"/mnt/models"}
{"level":"info","ts":1606770596.0174067,"logger":"modelAgent","msg":"Getting file /mnt/models/model1/iris.py"}
{"level":"info","ts":1606770596.2042253,"logger":"modelWatcher","msg":"Downloaded model."}
{"level":"error","ts":1606770596.213174,"logger":"modelProcessor","msg":"Failed to Load model","modelName":"model1","error":"Post http://localhost:8080/v2/repository/models/model1/load: dial tcp [::1]:8080: connect: connection refused","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132\ngithub.com/kubeflow/kfserving/pkg/agent.(*Puller).modelProcessor\n\t/go/src/github.com/kubeflow/kfserving/pkg/agent/puller.go:143"}
{"level":"info","ts":1606770596.2133465,"logger":"modelOnComplete","msg":"completion event for model","modelName":"model1","inFlight":0}
```

**Release note**:

```release-note
Adding in GCS as an accepted storage provider
```


mszacillo:
@googlebot I fixed it.

mszacillo:
@googlebot I fixed it.

yuzisun:
@mszacillo This is great! Can you help rebase the master and sign the Google CLA?

mszacillo:
@yuzisun Yes! I can take care of rebasing the branch - for the CLA I'm just waiting to hear back for approval from legal. Should be able to get that done in the next couple days.

ifilonenko:
has this been tested outside of unit tests? i.e. an e2e example of a pod with the appropriate labels pulling down from gcs? Can the results of that test be put in the PR description please

mszacillo:
Good point, I can do some manual verifications and post the results on the PR description.

pkg/agent/test/mockapi/storageinterface.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"context"
+	"io"
+	"cloud.google.com/go/iam"
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+type Client interface {
```
yuzisun, line 17: @mszacillo Is there any reason to copy over the interface ? Seems like we can use https://github.com/googleapis/google-cloud-go-testing/blob/master/storage/stiface/interfaces.go

pkg/apis/serving/v1beta1/trained_model.go
```diff
@@ -65,3 +65,31 @@ type ModelSpec struct {
 	// +optional
 	Memory resource.Quantity `json:"memory,omitempty"`
 }
+
+type TrainedModelYaml struct {
```
yuzisun, line 69: Why we do need the yaml version?

cmd/agent/main.go
```diff
@@ -19,24 +23,43 @@ var (
...
 		if err != nil {
 			panic(err)
 		}
-		sessionClient := s3.New(sess)
-		downloader.Providers[storage.S3] = &storage.S3Provider{
-			Client: sessionClient,
-			Downloader: s3manager.NewDownloaderWithClient(sessionClient, func(d *s3manager.Downloader) {
-			}),
+		downloader.Providers[kfstorage.GCS] = &kfstorage.GCSProvider{
+			Client: mockapi.AdaptClient(client),
```
yuzisun, line 62: here should use the real client?

pkg/agent/downloader.go
```diff
@@ -77,14 +80,41 @@ func (d *Downloader) download(modelName string, storageUri string) error {
 	return nil
 }
 
+func (d* Downloader) GetProvider() (string, error) {
+	storageUri := ""
+	matches, _ := filepath.Glob(d.ModelDir + "/*.yaml")
```
yuzisun, line 85: not sure if I understand this, the storage uri should be passed on `ModelSpec` and does not need to be parsed from the yaml.

pkg/agent/test/mockapi/storageinterface.go
mszacillo: @yuzisun That was the original plan, but there were conflicts upon importing the google-cloud-go-testing library. So I temporarily included these as a workaround in order to create some unit tests. However, after updating and rebasing on master I noticed that the import works now - so I'll go ahead and remove these!

pkg/agent/downloader.go
mszacillo: My understanding was that we'd configure the storage provider before any of the models are pulled, however I think I was over complicating things. We can just add each supported provider to the Provider map and then upon an add event, the correct provider will be parsed from the `ModelSpec` like you said. I'll go ahead and remove this logic.

cmd/agent/main.go
mszacillo: The real client is defined by `storage.NewClient(ctx)` on line 54. This line instead adapts the client to fix the mock client interface which is used to allow unit testing. At first I had set the client type for the GCS provider as storage.Client, but this was causing type errors when I went and mocked the client for unit tests - so I used a common interface between the mocked version and the real client. 

I might be missing a better way to do this, so I'll look into it further.

pkg/agent/kfstorage/gcs.go
```diff
@@ -0,0 +1,118 @@
...
+	rc, err := client.Bucket(attrs.Bucket).Object(attrs.Name).NewReader(g.Context)
+	if err != nil {
+		return fmt.Errorf("failed to create reader for object(%s) in bucket(%s): %v",
+			attrs.Name,
+			attrs.Bucket,
+			err,
+		)
+	}
+	defer rc.Close()
+	data, err := ioutil.ReadAll(rc)
```
ifilonenko, line 95: Is there a way to send this directly to disk? i.e. zero copy transfer or some more efficient way to write to disk directly? or is there BatchDownloader that takes a FileWriter? and what is `rc` here?

pkg/apis/serving/v1beta1/trained_model.go
ifilonenko: +1

ifilonenko (changes requested):
a few nits but thanks for writing up a much need storage protocol!

cmd/agent/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Build the inference-agent binary
```
yuzliu: This dockerfile can be removed since we already have a dockerfile for agent here: https://github.com/kubeflow/kfserving/blob/master/agent.Dockerfile

cmd/agent/main.go
```diff
@@ -1,28 +1,33 @@
...
+	"github.com/kubeflow/kfserving/pkg/agent/kfstorage"
+	gcscredential "github.com/kubeflow/kfserving/pkg/credentials/gcs"
 	s3credential "github.com/kubeflow/kfserving/pkg/credentials/s3"
 	"os"
 )
 
 var (
-	configDir = flag.String("config-dir", "/mnt/configs", "directory for model config files")
-	modelDir  = flag.String("model-dir", "/mnt/models", "directory for model files")
+	configDir = flag.String("config-dir", "/Users/michael.szacillo/mnt/configs", "directory for model config files")
```
yuzliu, line 20: Please revert the changes for local testing.

cmd/agent/Dockerfile
mszacillo: Fixed this along with your other comment during my most recent rebase.

pkg/agent/kfstorage/provider.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package storage
+package kfstorage
```
ifilonenko, line 17: is there a reason for package rename?

pkg/agent/test/gcs_test.go
```diff
@@ -0,0 +1,195 @@
+package test
```
ifilonenko, line 1: why not move your test to: `pkg/agent/watcher_test.go` as there is already a test suite that is available there, you would only need to add a mocked provider as you had done in your unit test. I think moving the providers into a package called: `mocks` might be useful however for organization. The presence of s3 mocks in `pkg/agent/watcher_test.go` seems incorrect if there are more than one protocols

. For isolation, maybe you could leave

pkg/agent/kfstorage/provider.go
mszacillo: I was getting conflicts with the `cloud.google.com/go/storage` import, since storage was resolving to the package instead of the gcs library. However, I have now changed the gcs import to have a unique name, so the library will be resolved without changing the package name.

ifilonenko: you can do: `import gstorage cloud.google.com/go/storage` which functions in go similar to: 
`import cloud.google.com/go/storage as gstorage`

pkg/agent/watcher_test.go
```diff
@@ -240,8 +211,8 @@ var _ = Describe("Watcher", func() {
 						ModelDir: modelDir + "/test3",
 						Providers: map[storage.Protocol]storage.Provider{
 							storage.S3: &storage.S3Provider{
-								Client:     &mockS3Client{},
-								Downloader: &mockS3Downloder{},
+								Client:     &testutils.MockS3Client{},
```
ifilonenko, line 214: maybe we could table test here?

ifilonenko (changes requested):
i think this is very close @yuzisun what do you think

pkg/agent/storage/gcs.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+)
+
+type GCSProvider struct {
+	Client stiface.Client
+}
+
+func (p *GCSProvider) DownloadModel(modelDir string, modelName string, storageUri string) error {
+	gcsUri := strings.TrimPrefix(storageUri, string(GCS))
+	path := strings.Split(gcsUri, "/")
```
yuzisun, line 22: this does not work for the case where only bucket is specified like `gs://bucket`, we fixed this in the S3Provider please checkout there.

pkg/agent/watcher_test.go
```diff
@@ -332,4 +297,39 @@ var _ = Describe("Watcher", func() {
 			})
 		})
 	})
+
+	Describe("Use GCS Downloader", func() {
```
yuzisun, line 301: Nice test! could you add a test for failing to download ?

yuzisun:
@mszacillo This looks great! Can you help resolve the conflicts and we should be good to merge!

yuzliu:
@mszacillo I think there must be a bug in the gcs downloader. I tested using ```gs://kfserving-samples/models/sklearn/iris```  The logs shows that it downloaded the model but when I exec into the pod the model was not downloaded.
```
{"level":"info","ts":1607388819.7210934,"logger":"modelProcessor","msg":"Downloading model","storageUri":"gs://kfserving-samples/models/sklearn/iris"}
{"level":"info","ts":1607388819.7213714,"logger":"Downloader","msg":"Downloading to model dir","modelUri":"gs://kfserving-samples/models/sklearn/iris","modelDir":"/mnt/models"}
{"level":"info","ts":1607388819.7258606,"logger":"modelOnComplete","msg":"completion event for model","modelName":"model3","inFlight":0}
```

mszacillo:
> Could you please try
> 
> ```
> kubectl exec -it <pod_name> -- /bin/bash
> cd /mnt/models
> ```
> 
> And verify that models are downloaded from gcs?

@yuzliu I added some debug logs to double check that the file is getting downloaded successfully - seems like the download works, the file gets written to successfully, and that the file exists after the download:

```
{"level":"info","ts":1607454256.6356146,"logger":"modelAgent","msg":"Downloading model ","modelName":"model1","storageUri":"gs://mskfservingmodels/iris.py","modelDir":"/mnt/models"}
{"level":"info","ts":1607454257.6121614,"logger":"modelAgent","msg":"Object iris.py exists."}
{"level":"info","ts":1607454257.7909303,"logger":"modelAgent","msg":"Created filename /mnt/models/model1/iris.py"}
{"level":"info","ts":1607454258.070437,"logger":"modelAgent","msg":"Wrote to file /mnt/models/model1/iris.py"}
{"level":"info","ts":1607454258.0704966,"logger":"modelAgent","msg":"File /mnt/models/model1/iris.py exists."}
{"level":"info","ts":1607454258.070524,"logger":"modelWatcher","msg":"Downloaded model."}
```
This is however with only the puller pod deployed by itself. Next I'll try creating a trainedmodel.

pkg/agent/storage/gcs.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (g *GCSObjectDownloader) GetObjectIterator(client stiface.Client) (stiface.ObjectIterator, error) {
+	query := &gstorage.Query{Prefix: g.Item}
+	return client.Bucket(g.Bucket).Objects(g.Context, query), nil
+}
+
+func (g *GCSObjectDownloader) Download(client stiface.Client, it stiface.ObjectIterator) error {
+	var errs []error
+	// flag to help determine if query prefix returned an empty iterator
+	var foundObject = false
```
mszacillo, line 64: I added in a flag to determine if the returned iterator is empty (meaning that the storage uri prefix used to search for objects returned nothing). If the flag is still false after iteration, that means no objects were found and we should log a warning for the user. This was done since the object iterator does not have a way to check size. That said I'm not entirely happy with this solution - any suggestions?

yuzisun:
@mszacillo Thanks for your awesome contribution!

/lgtm
/approve

Issue #977, opened 2020-07-28
make deploy fails with  The CustomResourceDefinition "inferenceservices.serving.kubeflow.org" is invalid:


navneet1075:
/kind bug

**What steps did you take and what happened:**
1. installed the kf serving using the developer guide from this  location: [developer guide](https://github.com/kubeflow/kfserving/blob/master/docs/DEVELOPER_GUIDE.md)
[A clear and concise description of what the bug is.
while running the make deploy step i am getting below errors : 
 **The CustomResourceDefinition "inferenceservices.serving.kubeflow.org" is invalid:
 spec.validation.openAPIV3Schema.properties[status].properties[address].properties[url].$ref: Forbidden: $ref is not supported
 spec.validation.openAPIV3Schema.properties[status].properties[conditions].$ref: Forbidden: $ref is not supported**

**What did you expect to happen:**
crds properly getting installed. 

**Anything else you would like to add:**
I am using the certificate manager as recommended in the developer guide.

[Miscellaneous information that will assist in solving the issue.]
Full logs of the make deploy command :   

make deploy

/Users/i318056/go/bin/controller-gen "crd:maxDescLen=0" paths=./pkg/apis/... output:crd:dir=config/crd
/Users/i318056/go/bin/controller-gen rbac:roleName=kfserving-manager-role paths=./pkg/controller/inferenceservice/... output:rbac:artifacts:config=config/rbac
/Users/i318056/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths=./pkg/apis/serving/v1alpha2
#TODO Remove this until new controller-tools is released
perl -pi -e 's/storedVersions: null/storedVersions: []/g' config/crd/serving.kubeflow.org_inferenceservices.yaml
perl -pi -e 's/conditions: null/conditions: []/g' config/crd/serving.kubeflow.org_inferenceservices.yaml
perl -pi -e 's/Any/string/g' config/crd/serving.kubeflow.org_inferenceservices.yaml
# Remove the certmanager certificate if KFSERVING_ENABLE_SELF_SIGNED_CA is not false
cd config/default && if [ true != false ]; then \
kustomize edit remove resource certmanager/certificate.yaml; \
else kustomize edit add resource certmanager/certificate.yaml; fi;
kustomize build config/default | kubectl apply --validate=false -f -
BATCHER_IMG ?= batcher:latest
namespace/kfserving-system unchanged
role.rbac.authorization.k8s.io/leader-election-role unchanged
[...]

PR #1115, opened 2020-10-07
Add instruction from #977 to the developer guide

theofpa:

Adds the instruction from #977 to the developer guide, so people with that problem will find it easier to update their `controller-gen` when that error occurs.


yuzisun:
thanks @theofpa for adding the instruction!

/lgtm
/approve

PR #1103, opened 2020-09-27
Make standalone installation kustomizable.

jazzsir:
**What this PR does / why we need it**:
To allow KFServing standalone to install in kubeflow namespace, I made standalone installation kustomizable, and updated README.md

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #976

**Special notes for your reviewer**:
This is my second attempt to merge PR, because of confliction issue at #1019

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #1126, opened 2020-10-10
use "kustomize build" instead of "kubectl kustomize"

jazzsir:
We use "kustomize build" instead of "kubectl kustomize" to avoid many changes.
To use existing one, some keyword should be changed for compatibility.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -41,7 +41,8 @@ generatorOptions:
 # kfserving namespace in webhook configurations
 configMapGenerator:
   - name: kfserving-config
-    env: params.env
+    envs: 
+    - params.env
 vars:
 - name: ingressGateway
   objref:
--- a/config/overlays/kubeflow/kustomization.yaml
+++ b/config/overlays/kubeflow/kustomization.yaml
@@ -20,4 +20,5 @@ generatorOptions:
 configMapGenerator:
   - name: kfserving-config
     behavior: replace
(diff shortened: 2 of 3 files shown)


pugangxa:
When I run ```kustomize build``` it reports: 
```
Error: json: unknown field "env"
```
So I also did the changes in this PR and it passed. 

But there's other issues(see issue #1127 ) . 
And also I found this https://github.com/kubernetes-sigs/kustomize/issues/2052 and seems it's not recommended to use vars. So maybe we should revert PR #1103 )?
@yuzisun

yuzisun:
> When I run `kustomize build` it reports:
> 
> ```
> Error: json: unknown field "env"
> ```
> 
> So I also did the changes in this PR and it passed.
> 
> But there's other issues(see issue #1127 ) .
> And also I found this [kubernetes-sigs/kustomize#2052](https://github.com/kubernetes-sigs/kustomize/issues/2052) and seems it's not recommended to use vars. So maybe we should revert PR #1103 )?
> @yuzisun

I think @jazzsir borrowed this from kubeflow manifests, will read a bit why vars are not recommended. But I do not see the issue reported in #1127, is that kustomize version issue ?

yuzisun:
@pugangxa i think this is ok and it is passing the CI, pretty sure the CI does kustomize build and the conversion also.

/lgtm
/approve

PR #1128, opened 2020-10-10
Support Kubernetes 1.18

### LINKED ISSUE #1070: Cannot deploy kf-serving on kubernetes 1.18

pugangxa:
Add protocol as required. CRDs are updated accordingly too.
Notice we must add "preserveUnknownFields=false" to spec or another error will be reported, see kubernetes-sigs/controller-tools#476

Fixes #1070

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -11,3 +11,11 @@ resources:
 - serving.kubeflow.org_inferenceservices.yaml
 - serving.kubeflow.org_trainedmodels.yaml
 
+patchesJson6902:
+  # Fix for https://github.com/kubernetes/kubernetes/issues/91395
+  - target:
+      group: apiextensions.k8s.io
+      version: v1beta1
+      kind: CustomResourceDefinition
+      name: inferenceservices.serving.kubeflow.org
+    path: patches/protocol.yaml
--- a/config/crd/patches/protocol.yaml
+++ b/config/crd/patches/protocol.yaml
@@ -1,0 +1,80 @@
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/canary/properties/explainer/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/canary/properties/predictor/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/canary/properties/transformer/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/default/properties/explainer/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/default/properties/predictor/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/default/properties/transformer/properties/custom/properties/container/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/triton/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/pytorch/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/onnx/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/sklearn/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/xgboost/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/tensorflow/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/containers/items/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/predictor/properties/initContainers/items/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/transformer/properties/containers/items/properties/ports/items/required/1
+  value: protocol
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/transformer/properties/initContainers/items/properties/ports/items/required/1
+  value: protocol
+
+
+- op: add
+  path: /spec/versions/1/schema/openAPIV3Schema/properties/spec/properties/explainer/properties/containers/items/properties/ports/items/required/1
+  value: protocol
+
+- op: add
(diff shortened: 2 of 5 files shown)


pugangxa:
Recreate the PR, I can deploy and apply inferenceservice with the change, though I'm afraid it may not pass version 1.16 validation.

Tomcli:
I think the `trainedmodel` crd is failing with the crd v1 spec, is it something you can fix it?

pugangxa:
> I think the `trainedmodel` crd is failing with the crd v1 spec, is it something you can fix it?

Checked the error, it's because after changing controller-gen version a ```default``` field will be added to the CRD, see https://github.com/kubernetes-sigs/controller-tools/pull/480 . But the ```default``` feature is a new feature which is not supported by Kubernetes 1.16 version(used in test cluster and kubebuilder for go test).

The reason I changed to add defaulter is due to the issue #1070, this is due to a change in https://github.com/kubernetes/kubernetes/pull/70100 that cause the port vaule failed since it's added as a listMapKey so it must be required or have default value in new Kubernetes version. Ideally if do not consider compatiblity we can add default value(it's also suggested fix in controller-tool), but it does not work in previous k8s version. 

I think another way is to use yq in Makefile to add protocol as required after CRD generated. If you can help on this it would be great, if not I will do it later since I've got caught up in something else, thanks Tommy.

yuzisun:
> > I think the `trainedmodel` crd is failing with the crd v1 spec, is it something you can fix it?
> 
> Checked the error, it's because after changing controller-gen version a `default` field will be added to the CRD, see [kubernetes-sigs/controller-tools#480](https://github.com/kubernetes-sigs/controller-tools/pull/480) . But the `default` feature is a new feature which is not supported by Kubernetes 1.16 version(used in test cluster and kubebuilder for go test).
> 
> The reason I changed to add defaulter is due to the issue #1070, this is due to a change in [kubernetes/kubernetes#70100](https://github.com/kubernetes/kubernetes/pull/70100) that cause the port vaule failed since it's added as a listMapKey so it must be required or have default value in new Kubernetes version. Ideally if do not consider compatiblity we can add default value(it's also suggested fix in controller-tool), but it does not work in previous k8s version.
> 
> I think another way is to use yq in Makefile to add protocol as required after CRD generated. If you can help on this it would be great, if not I will do it later since I've got caught up in something else, thanks Tommy.

@pugangxa Thanks for the investigation, let's do the yq approach then.

cliveseldon:
> > > I think the `trainedmodel` crd is failing with the crd v1 spec, is it something you can fix it?
> > 
> > 
> > Checked the error, it's because after changing controller-gen version a `default` field will be added to the CRD, see [kubernetes-sigs/controller-tools#480](https://github.com/kubernetes-sigs/controller-tools/pull/480) . But the `default` feature is a new feature which is not supported by Kubernetes 1.16 version(used in test cluster and kubebuilder for go test).
> > The reason I changed to add defaulter is due to the issue #1070, this is due to a change in [kubernetes/kubernetes#70100](https://github.com/kubernetes/kubernetes/pull/70100) that cause the port vaule failed since it's added as a listMapKey so it must be required or have default value in new Kubernetes version. Ideally if do not consider compatiblity we can add default value(it's also suggested fix in controller-tool), but it does not work in previous k8s version.
> > I think another way is to use yq in Makefile to add protocol as required after CRD generated. If you can help on this it would be great, if not I will do it later since I've got caught up in something else, thanks Tommy.
> 
> @pugangxa Thanks for the investigation, let's do the yq approach then.

You can check out this Kustomize patch in Seldon that does what @yuzisun suggests: 

https://github.com/SeldonIO/seldon-core/blob/3b042a42092c0b164c26f7c75940ecc615aa64f3/operator/config/crd/kustomization.yaml#L60-L66

pugangxa:
The error is due to migrate version to CRD v1. Think we should spilit it to be 2 PRs, one to fix the protocol issue in k8s 1.18, another to migrate to CRD v1.

pugangxa:
See 
https://github.com/kubernetes-sigs/controller-tools/issues/302#issuecomment-558623779 
https://github.com/kubernetes-sigs/controller-tools/issues/302#issuecomment-561609196

Seems it only occur in envtest since I can deploy this without error.

cliveseldon:
The only downside to this solution is that in a normal PodSpec `protocol` is not required. But this seems the only solution for v1beta1 CRD so we have initially adopted it in Seldon for PodSpec's defined inside our v1beta1 CRD. 

In v1 CRD you can do it properly.

pugangxa:
> The only downside to this solution is that in a normal PodSpec `protocol` is not required. But this seems the only solution for v1beta1 CRD so we have initially adopted it in Seldon for PodSpec's defined inside our v1beta1 CRD.
> 
> In v1 CRD you can do it properly.

I agree. But I'm afraid even in v1 CRD there's no good method because still we are going to support 1.16 and 1.17. Controller tools solve this issue by adding default value for protocol but the defaulter is a new feature that does not work on k8s 1.16. We can remove this until we migrate to 1.18.

yuzisun:
> The only downside to this solution is that in a normal PodSpec `protocol` is not required. But this seems the only solution for v1beta1 CRD so we have initially adopted it in Seldon for PodSpec's defined inside our v1beta1 CRD.
> 
> In v1 CRD you can do it properly.

@cliveseldon @pugangxa Does this mean we need to add protocol field for this gRPC example?
https://github.com/kubeflow/kfserving/blob/master/docs/samples/v1beta1/tensorflow/grpc.yaml

cliveseldon:
> > The only downside to this solution is that in a normal PodSpec `protocol` is not required. But this seems the only solution for v1beta1 CRD so we have initially adopted it in Seldon for PodSpec's defined inside our v1beta1 CRD.
> > In v1 CRD you can do it properly.
> 
> @cliveseldon @pugangxa Does this mean we need to add protocol field for this gRPC example after this change?
> https://github.com/kubeflow/kfserving/blob/master/docs/samples/v1beta1/tensorflow/grpc.yaml

I would imagine so.

yuzisun:
thanks @pugangxa ! Looks like we have to live with this hack for now.

/lgtm
/approve

animeshsingh:
@pugangxa will this be the only PR needed to enabled support for 1.18 with last release of KFServing (0.4.1)?

pugangxa:
> @pugangxa will this be the only PR needed to enabled support for 1.18 with last release of KFServing (0.4.1)?

Yes, with this PR I can deploy on 1.18 and run the sklearn example. But suggest also include PR #1144 since I plan to migrate to CRD v1 which is the new default on 1.18 and that will depend on PR #1144.

PR #1133, opened 2020-10-13
Fix rollout_canary API

Peilun-Li:
Should the Python `rollout_canary` API only update `canary` section instead of the `default` (primary) section? Or am I missing some changes in the v1beta1 spec.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/python/kfserving/kfserving/api/kf_serving_client.py
+++ b/python/kfserving/kfserving/api/kf_serving_client.py
@@ -269,7 +269,6 @@ class KFServingClient(object):
 
         current_isvc['spec']['canaryTrafficPercent'] = percent
         if canary:
-            current_isvc['spec']['default'] = canary
             current_isvc['spec']['canary'] = canary
 
         return self.patch(name=name, inferenceservice=current_isvc, namespace=namespace,


pugangxa:
This api is to rollout canary so just use canary to replace default configuration.

Peilun-Li:
> This api is to rollout canary so just use canary to replace default configuration.

That usecase sounds like the `promote` API which got removed in the same commit. The questions then are:
1. The functionality of `rollout_canary` seems to be changed, from creating canary route to promoting canary route.
2. Why do we still need a `percent` parameter here since all traffic will just be routed to the same model.
3. How do we create canary route on existed default route using Python SDK then? Previously we can use `rollout_canary` to do so and `promote` to promote it (https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/kfserving/src/kfservingdeployer.py#L222-L236), but now it seems we could only achieve so by calling `replace`/`patch` upon inferenceservice built manually.

I'm got a bit confused by the SDK changes, or is there anything forthcoming in v0.5 release which can help explain

yuzisun:
@Peilun-Li great questions! let me explain this.

The implementation of canary is changed, previously canary is implemented at inference service spec level, in v1beta1 we removed the default/canary level and pushed the canary implementation down to knative revision level. As you modify the inference service spec the change is already tracked by knative revisions and now you can do traffic split on two knative revisions of the same knative service instead of splitting on two different knative services. 

For backwards compatibility we convert the v1alpha2 spec to v1beta1, v1beta no longer has the canary section so if you still have canary spec on v1alpha2 we update default with the canary which results in a new knative revision so rollout_canary and trafficPercent still makes sense, however promote is no longer needed as now you can simply change trafficPercent to 100 to rollout your canary version.

We are going to generate the v1beta1 sdk api and totally rewrite the API against v1beta1, again the main purpose of the v1alpha2 sdk change is to maintain the API backwards compatibility.

Peilun-Li:
@yuzisun That makes sense then. Thanks for the explanation and looking forward to the new experience!

PR #1144, opened 2020-10-19
upgrade k8s and kn go library versions

ryandawsonuk:
Aiming to upgrade the go dependencies to a set of compatible versions:

* [k8s](https://github.com/kubernetes/client-go/blob/release-1.18/go.mod) 1.18.8
* [istio](https://github.com/istio/istio/blob/1.6.12/go.mod) 1.16.2
* [knative](https://github.com/knative/serving/blob/v0.18.0/go.mod) 0.18.0

Not especially related to https://github.com/kubeflow/kfserving/pull/1128 as I believe that PR is not upgrading any go library versions.

Key code changes resulting:
* Many k8s golang APIs now require a context parameter.
* Needed to change loading of CRDs for [envtest](https://book.kubebuilder.io/reference/envtest.html) as it now errors if you point it at a kustomize dir. So instead copying just CRD manifests into existing directory for test CRDs.
* Have to explicitly create namespaces in tests now.


ryandawsonuk:
Current status: 
* Go tests pass.
* Presubmit failed but failures there look like unrelated build failures.
* I can run with `make deploy-dev` and run the [sklearn sample](https://github.com/kubeflow/kfserving/tree/master/docs/samples/sklearn) successfully to get predictions.

yuzisun:
> Current status:
> 
> * Go tests pass.
> * Presubmit failed but failures there look like unrelated build failures (?).
> * I can run with `make deploy-dev` (after `./hack/quick_install.sh` and `make undeploy`) and run the [sklearn sample](https://github.com/kubeflow/kfserving/tree/master/docs/samples/sklearn) successfully to get predictions.

@ryandawsonuk the builds are fine, after retry it is fine, the problem is the e2e test which has been stuck for hours now.

ryandawsonuk:
Thanks @yuzisun - have the e2e tests been timing out on other PRs too?

Makefile
```diff
@@ -115,6 +115,8 @@ manifests: controller-gen
 	yq d -i config/crd/serving.kubeflow.org_inferenceservices.yaml 'spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.*.properties.*.properties.livenessProbe.properties.httpGet.required'
 	yq d -i config/crd/serving.kubeflow.org_inferenceservices.yaml 'spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.*.properties.*.properties.readinessProbe.properties.tcpSocket.required'
 	yq d -i config/crd/serving.kubeflow.org_inferenceservices.yaml 'spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.*.properties.*.properties.livenessProbe.properties.tcpSocket.required'
+	cp -u config/crd/serving.kubeflow.org_inferenceservices.yaml test/crds/serving.kubeflow.org_inferenceservices.yaml
```
yuzisun, line 118: Why do we need to copy the crd files?

pkg/apis/serving/v1alpha2/openapi_generated.go
```diff
@@ -50,10 +50,6 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 		"./pkg/apis/serving/v1alpha2.TransformerSpec":         schema_pkg_apis_serving_v1alpha2_TransformerSpec(ref),
 		"./pkg/apis/serving/v1alpha2.TritonSpec":              schema_pkg_apis_serving_v1alpha2_TritonSpec(ref),
 		"./pkg/apis/serving/v1alpha2.XGBoostSpec":             schema_pkg_apis_serving_v1alpha2_XGBoostSpec(ref),
-		"knative.dev/pkg/apis.URL":                            schema_knativedev_pkg_apis_URL(ref),
```
yuzisun, line 53: Any idea why this is deleted? I think we are still using them.

ryandawsonuk: I guess it's the generation. I didn't take it out manually. If I put it in manually it still seems to compile. My best guess right now is it's a knock-on from some upstream change in knative that's coming out of upgrading the version. Interesting that kservices are being created and their URLs do show in the status section even without these lines.

yuzisun: ye this is to generate the openapi schema for KFServing python SDK. kubectl would still work but the python SDK client may not show the urls. I guess there is something wrong when generating knative types. We can revert this for now and create an issue to investigate the knative type generation issue.

ryandawsonuk: Have manually overridden in the code for now with a note and can create an issue to follow up.

Makefile
pugangxa: If there's other files(here it's kustomize) an error will be reported. But I think we need not copy CRD files because we can specify to path level through Environment.CRDInstallOptions

go.mod
```diff
@@ -3,53 +3,70 @@ module github.com/kubeflow/kfserving
...
+	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
 	k8s.io/klog v1.0.0
-	k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6
-	knative.dev/pkg v0.0.0-20191217184203-cf220a867b3d
-	knative.dev/serving v0.11.0
-	sigs.k8s.io/controller-runtime v0.4.0
+	k8s.io/kube-openapi v0.0.0-20200410145947-bcb3869e6f29
+	knative.dev/pkg v0.0.0-20200922164940-4bf40ad82aab
+	knative.dev/serving v0.18.0
+	sigs.k8s.io/controller-runtime v0.6.3
```
pugangxa, line 42: I upgraded controller-runtime to 0.5.0, why has it to be 0.6.3? Also @yuzisun , what compatibility level do we support? In my opinion at least k8s 1.17?

Makefile
ryandawsonuk: Yeah the issue is about [envtest](https://book.kubebuilder.io/reference/envtest.html) as it now errors if you point it at a kustomize dir. In master it points [at the config dir](https://github.com/kubeflow/kfserving/pull/1144/files#diff-68abf77fa6ead6f002a59acab8552c391eaf159cdfe897c83096ddae128f30e6L54) and in the new version envtest tries to load [kustomization.yaml](https://github.com/kubeflow/kfserving/blob/master/config/crd/kustomization.yaml) and errors that it doesn't contain a Kind. I can change the envtest path but it needs a path, it can't be pointed at single files. And the path with the crds does contain that kustomization. I figured that since it's already pointing at the [test directory](https://github.com/kubeflow/kfserving/tree/4bddfbc0b650261cb867d50512149a07e37f7fab/test/crds) (which also contains istio crds) it would make sense to just copy the built crds to there every time they're built and keep that test directory up to date and make it the [only place that the tests load CRDs from](https://github.com/kubeflow/kfserving/pull/1144/files#diff-5cefe84774ca5f18960f72fe12ad5e8f7722b1ba535ab68402c797837ce7385dR40)

go.mod
ryandawsonuk: The idea of upgrading to 0.6.3 is to get to kubernetes 1.18 - the [0.6.X tree of controller-runtime is for 1.18](https://github.com/kubernetes-sigs/controller-runtime/issues/1033#issuecomment-706562042)

Makefile
pugangxa: Thanks, I understand since I encounter the same error and must move kustomize.yaml out to make it work. But after investigation I found when setting up envtest there's CRDInstallOptions can be passed and can point to file:
https://github.com/kubernetes-sigs/controller-runtime/blob/bae8fdb711031aa10dbf7361dbe77a5a650aa8d9/pkg/envtest/crd.go#L44

It's said this can take effect together with CRDDirectoryPaths:
https://github.com/kubernetes-sigs/controller-runtime/blob/bae8fdb711031aa10dbf7361dbe77a5a650aa8d9/pkg/envtest/server.go#L111-L114

go.mod
pugangxa: Yes I understand. So if we require 1.18 then it should be fine. Now I just upgraded to 0.5.0 to support 1.17.

Makefile
ryandawsonuk: Did you manage to get it to work with files? I tried it and found I got an error as in the controller-runtime [code block that tries to handle it](https://github.com/kubernetes-sigs/controller-runtime/blob/bae8fdb711031aa10dbf7361dbe77a5a650aa8d9/pkg/envtest/crd.go#L304) the if and the else both assume a Dir. I can try again.

pugangxa: Yes, though there's error due to test case failed that I did not get time to figure out, but at least the install CRD part works for me with below code in envtest_setup.go:
```
func SetupEnvTest() *envtest.Environment {
	t := &envtest.Environment{
		CRDInstallOptions: envtest.CRDInstallOptions{
			Paths: []string{
				filepath.Join("..", "..", "..", "..", "..", "..", "config", "crd", "serving.kubeflow.org_inferenceservices.yaml"),
				filepath.Join("..", "..", "..", "..", "..", "..", "config", "crd", "serving.kubeflow.org_trainedmodels.yaml"),
				filepath.Join("..", "..", "..", "..", "config", "crd", "serving.kubeflow.org_inferenceservices.yaml"),
				filepath.Join("..", "..", "..", "..", "config", "crd", "serving.kubeflow.org_trainedmodels.yaml"),
			},
		},
		// The relative paths must be provided for each level of test nesting
		// This code should be illegal
		CRDDirectoryPaths: []string{
			//filepath.Join("..", "..", "..", "..", "..", "..", "config", "crd"),
			filepath.Join("..", "..", "..", "..", "..", "..", "test", "crds"),
			//filepath.Join("..", "..", "..", "..", "config", "crd"),
			filepath.Join("..", "..", "..", "..", "test", "crds"),
		},
		UseExistingCluster: proto.Bool(false),
	}
```

ryandawsonuk: @pugangxa you were right, I made a mistake before. It does work with files. Have changed so no need for a copy now. Thanks for the tip.

pugangxa:
After PR #1128 kfserving can work on k8s 1.18. But to migrate to CRD v1 we need most of the changes, take a look at the code and in fact I did a lot of similar changes in PR #1147 (not committed yet due to test error). 
I think we can merge these 2 PRs together, or make this PR merged firstly, and PR #1147 can depend on this, thanks.

go.mod
yuzisun: @pugangxa @ryandawsonuk I think we should support at least Kubernetes 1.16 which is the min requirement of knative 0.15 and latest knative 0.18 release bumped the min requirement to 1.17. Based on the fact that this PR passed the CI  it does actually seem to work with 1.16, not sure if there could be issues which are not covered by e2e tests.

pugangxa: Thanks, then I think it should be fine and I think these libs maintain the compatibility.

ryandawsonuk: I've also been checking backward-compatibility - been running this version on a k8s 1.15 cluster running knative 0.11 and istio 1.1.6. Been able to create, call and delete and not found any problems yet.

go.mod
```diff
@@ -3,53 +3,70 @@ module github.com/kubeflow/kfserving
...
+	k8s.io/kubectl => k8s.io/kubectl v0.18.8
+	k8s.io/kubelet => k8s.io/kubelet v0.18.8
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.18.8
+	k8s.io/metrics => k8s.io/metrics v0.18.8
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.8
+	k8s.io/utils => k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
+	k8s.io/test-infra => k8s.io/test-infra v0.0.0-20200803112140-d8aa4e063646
+	k8s.io/klog => k8s.io/klog v1.0.0
+	// based on https://github.com/openshift/cluster-ingress-operator/pull/429/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R34
+	sigs.k8s.io/controller-runtime => github.com/zchee/sigs.k8s-controller-runtime v0.6.1-0.20200623114430-46812d3a0a50
```
yuzisun, line 71: why do we use the controller-runtime fork ?

ryandawsonuk: That’s due to https://github.com/kubernetes-sigs/controller-runtime/issues/1033

ryandawsonuk:
@yuzisun Is there anything further needed in order to move forward with this?

yuzisun:
Thanks @ryandawsonuk ! Awesome job!

/lgtm
/approve

go.mod
pugangxa: Is the fork still there? Seems not work for me:
```
➜  kfserving git:(migrate-crd-v1) go mod verify
go: github.com/zchee/sigs.k8s-controller-runtime@v0.6.1-0.20200623114430-46812d3a0a50: reading https://goproxy.io/github.com/zchee/sigs.k8s-controller-runtime/@v/v0.6.1-0.20200623114430-46812d3a0a50.mod: 404 Not Found
	server response: not found: unknown revision 46812d3a0a50
```

ryandawsonuk: Hmm, `go mod verify` works for me and the fork seems to be there - https://github.com/openshift/cluster-ingress-operator/pull/429/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R34

Also the makefile commands like `make manager` work. I'm on go1.14.4. Maybe it's a go version thing?

pugangxa: Thanks. I am on go1.14.7 and not sure why. But checked https://github.com/zchee/sigs.k8s-controller-runtime/tags and not found v0.6.1. Maybe because it's already under your go mod cache...

Issue #13, opened 2019-04-10
Non-KNative Resource Option


cliveseldon:
We want to allow the user to run kfserving components without the dependency on KNative.
An initial implementation could :

  * Allow non-knative option via annotations
  * Create raw k8s deployments and services for model serving components.

gaocegege:
I suggest listing the feature in a high priority. We (Caicloud) have an internal version of Serving CRD, which is based on Kubernetes Ingress/Service and Istio. I think we should support different serving backends to be more general.

How about define an interface for serving backends and use annotations or CLI flags in controller to control what backends (KNative, Istio, Linkerd, Kubernetes Native Ingress/Service) will be used

yuzisun:
we actually discussed about the annotation approach in our WG meeting today.

cliveseldon:
@gaocegege  That sounds good. To handle canaries and blue/green routing the annotation would indeed need to specify the routing resources to utilize:
 * KNative (default?)
 * Istio
 * Linkerd
 * Envoy

gaocegege:
@cliveseldon KNative by default SGTM. Using Kubernetes native resources to handle canaries and blue/green routing may be hard, while it is feasible to use istio/linkerd/knative. Thus the list LGTM.

@yuzisun Is there any calendar for the SIG meeting? I am interested in it, too.

ellis-bigelow:
@gaocegege, I've added you to the meetings.

ellis-bigelow:
I want to minimize the number of implementations. I'm all for pluggability, but we need to make sure we're delivering clear customer value. Knative itself is working on the pluggability question, so we may be able to simply rely on their improvements.

Short term, as we discussed in our KFSWG meeting, the best value for effort will be:

- Default Knative impl w/ canarying & all the bells
- Annotated raw k8s impl w/o canarying

ellis-bigelow:
Also @gaocegege, in our discussions so far, we've limited the networking responsibilities of KFServing to in-cluster communication. If you have Knative/Istio installed, you can use their Loadbalancer Service for Ingress, but we're not providing an Ingress solution ourselves.

I'm not sure we want to be creating ingress resources, as they don't map 1:1 with a KFService. If there's some disagreement here, this might be a great topic for a KFSWG Special Topics meeting.

rakelkar:
[...]

PR #1147, opened 2020-10-20
Migrage to CRD version v1

pugangxa:
- Update controller-gen version to v0.4
- Update conversion webhook format


yuzisun:
@pugangxa Are you still on this? We'd like to get v1 as this is going away in kubernetes 1.22

yuzisun:
going to work off under https://github.com/kserve/kserve/issues/13

PR #1149, opened 2020-10-20
Introduce defaultTimeout for predictor and apply in tfserving

theofpa:
Serving models with large number of parameters and classes takes longer than a typical API response. For indication, object detection or image classification models based on Open Images Dataset v4 built in Tensorflow takes between 20-50 seconds to respond. The `tensorflow/serving` timeout is defined by the [--rest_api_timeout_in_ms](https://www.tensorflow.org/tfx/serving/serving_config#miscellaneous_flags) flag and by default has the value of 30 seconds. This barrier is not in line with the Knative/Istio/Ingress timeouts across the kfserving project, which are usually configured to 600 seconds.

This PR introduces the `defaultTimeout` parameter in the `PredictorConfig` of the `inferenceservice-config` ConfigMap, and implements it in the arguments of the tfserving pod. This way, ML engineers who are managing KFServing environments will be able to change it across all `inferenceservices` using tfserving.

* Implemented as a parameter in the ConfigMap and not in Spec, to keep the `InferenceService` CR simple.
* Tests exit successfully on `make test`, the `defaultTimeout` is passed to the pod arguments.
* The gen-crd-apidocs generated additional elements for AIX explainer which were missed by #1094

**Release note**:

```release-note
TFServing timeout can be configured in inferenceservice-config configmap
```


pugangxa:
Good feature, thanks for your PR. @yuzisun Do you think we need a common timeout feature for other models too?

salanki:
Please note that the overall max timeout is governed by https://github.com/kubeflow/kfserving/blob/deba8df156718f8af67cc0c262511c53bb1f1ad6/pkg/constants/constants.go#L73 that is set on the `KNativeService`. Making the timeout configurable on the predictor spec and having that drive both the `KNativeService` timeout, the Istio `VirtualService` and the TFServing (and other framework) timeouts would be the best solution. There is an issue for it here: https://github.com/kubeflow/kfserving/issues/809

pugangxa:
Thanks for explanation. I think it's a good starting to apply on TFServing firstly, and later can be applied to other models.

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -88,6 +88,7 @@ func (t *TFServingSpec) GetContainer(metadata metav1.ObjectMeta, extensions *Com
 		fmt.Sprintf("%s=%s", "--rest_api_port", TensorflowServingRestPort),
 		fmt.Sprintf("%s=%s", "--model_name", metadata.Name),
 		fmt.Sprintf("%s=%s", "--model_base_path", constants.DefaultModelLocalMountPath),
+		fmt.Sprintf("%s=%s", "--rest_api_timeout_in_ms", config.Predictors.Tensorflow.DefaultTimeout),
```
yuzisun, line 91: @theofpa [timeout](https://github.com/kubeflow/kfserving/blob/master/pkg/apis/serving/v1beta1/component.go#L80) is already added in v1beta1 API, the value needs to match user specified value, the defaultTimeout should be applied on mutating webhook.

pkg/apis/serving/v1alpha2/configmap.go
```diff
@@ -53,6 +53,8 @@ type PredictorConfig struct {
 
 	DefaultImageVersion    string `json:"defaultImageVersion"`
 	DefaultGpuImageVersion string `json:"defaultGpuImageVersion"`
+	// Default timeout of predictor for serving a request, in milliseconds
+	DefaultTimeout string `json:"defaultTimeout,omitempty"`
```
yuzisun, line 57: should be int64 instead of string

yuzisun: @theofpa I am thinking we probably do not need this as TFServing already has a default 30s timeout right?

theofpa: We need to have it inherit the timeout from the inferenceservice though. When it comes to large timeout values, it's a problem.

theofpa: addressed in afaf6f9

pkg/apis/serving/v1beta1/predictor_tfserving.go
theofpa: addressed in 565d3ab

theofpa:
Pinging @pugangxa @salanki @yuzisun for review after the implementation of the recommended improvements.

pkg/apis/serving/v1alpha2/framework_tensorflow_test.go
```diff
@@ -110,6 +110,7 @@ func TestTensorflowContainer(t *testing.T) {
 			"--rest_api_port=" + TensorflowServingRestPort,
 			"--model_name=someName",
 			"--model_base_path=/mnt/models",
+			"--rest_api_timeout_in_ms=0",
```
yuzisun, line 113: This should get the default value from configmap right?

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -83,11 +85,18 @@ func (t *TFServingSpec) GetStorageUri() *string {
 
 // GetContainers transforms the resource into a container spec
 func (t *TFServingSpec) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
+	// Get the timeout from user input or else use the default timeout
+	if extensions.TimeoutSeconds != nil {
+		TimeoutMilliSeconds = 1000 * *extensions.TimeoutSeconds
+	} else {
+		TimeoutMilliSeconds = 1000 * constants.DefaultPredictorTimeout
```
yuzisun, line 92: should get the configured default value instead of the constant?

pkg/controller/v1alpha2/inferenceservice/reconcilers/knative/service_reconciler_test.go
```diff
@@ -144,6 +144,7 @@ func TestKnativeServiceReconcile(t *testing.T) {
 												"--rest_api_port=" + v1alpha2.TensorflowServingRestPort,
 												"--model_name=mnist",
 												"--model_base_path=" + constants.DefaultModelLocalMountPath,
+												"--rest_api_timeout_in_ms=0",
```
yuzisun, line 147: should be 60000?

pkg/apis/serving/v1alpha2/framework_tensorflow_test.go
theofpa: thanks for keeping me honest with the tests, fixed in 2c3b5de

pkg/apis/serving/v1beta1/predictor_tfserving.go
theofpa: indeed that should have been the case, I've fixed it in 2c3b5de

pkg/controller/v1alpha2/inferenceservice/reconcilers/knative/service_reconciler_test.go
theofpa: corrected in 2c3b5de

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -32,6 +33,7 @@ var (
 	TensorflowServingGRPCPort            = "9000"
 	TensorflowServingRestPort            = "8080"
 	TensorflowServingGPUSuffix           = "-gpu"
+	TimeoutMilliSeconds                  int64
```
yuzisun, line 36: This should be defined as a local variable in `GetContainer`

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -83,11 +85,18 @@ func (t *TFServingSpec) GetStorageUri() *string {
 
 // GetContainers transforms the resource into a container spec
 func (t *TFServingSpec) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
+	// Get the timeout from user input or else use the default timeout
+	if extensions.TimeoutSeconds != nil {
+		TimeoutMilliSeconds = 1000 * *extensions.TimeoutSeconds
+	} else {
+		TimeoutMilliSeconds = 1000 * config.Predictors.Tensorflow.DefaultTimeout
+	}
```
yuzisun, line 93: suggest the following 
```
TimeoutMilliSeconds := 1000 * config.Predictors.Tensorflow.DefaultTimeout
if extensions.TimeoutSeconds != nil {
    TimeoutMilliSeconds = 1000 * *extensions.TimeoutSeconds
}
```

pkg/apis/serving/v1beta1/predictor_tfserving.go
theofpa: addressed in 61de56e

pkg/apis/serving/v1beta1/predictor_tfserving.go
theofpa: addressed in 61de56e

yuzisun:
Looks great! thanks @theofpa !

/lgtm
/approve

Issue #1111, opened 2020-09-30
Add support to deploy SKLearn and XGBoost models with MLServer


adriangonz:
/kind feature

**Describe the solution you'd like**
Add support to deploy SKLearn and XGBoost models compatible with the [V2 Dataplane](https://github.com/kubeflow/kfserving/tree/master/docs/predict-api/v2), leveraging [MLServer](https://github.com/seldonio/mlserver).

**Anything else you would like to add:**
For the initial step, we could just focus on the serving aspect (i.e. the basic V2 Dataplane) and leave the MMS endpoints (and other extensions) for later.

PR #1155, opened 2020-10-23
Add support for MLServer in the SKLearn predictor

adriangonz:

Adds support for MLServer (and the V2 dataplane) in the `v1beta1` version of the SKLearn predictor. Note that this PR will still default all `v1beta1` SKLearn `InferenceServices` to use the V1 dataplane. To enable the V2 protocol, there is a new `protocolVersion` field in the predictor spec which can be set to `v1`.

This PR also adds an example worth checking under `./docs/samples/v1beta1/sklearn` showcasing how the integration works. 

Works towards #1111 *(it doesn't fix it yet, as it still needs support for XGBoost)*.

The aim of this PR is to kickstart the discussion on how do we want to shape the integration between KFServing and MLServer. This has already surfaced some questions, e.g.:

- ~Do we want to maintain backwards compatibility in the data plane with previous versions?~ *Yes, for now.*
- Moving to MMS, do we want to restrict a single framework in each MLServer pod? In other words, do we need a `mlserver` predictor which supports multiple frameworks? 

As such, it would be great to get people's thoughts on this proposal and which aspects would they change. 

That's also why it only focuses on SKLearn for now. Once we are happy with how it looks, it should pretty straightforward to extend to XGBoost. 

**Release note**:

```release-note
Added support for the V2 dataplane in the SKLearn predictor when using the `v1beta1` version of the `InferenceService` CRDs.
```


pkg/apis/serving/v1beta1/predictor_sklearn.go
```diff
@@ -53,22 +52,60 @@ func (k *SKLearnSpec) Default(config *InferenceServicesConfig) {
...
 func (k *SKLearnSpec) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
-	arguments := []string{
-		fmt.Sprintf("%s=%s", constants.ArgumentModelName, metadata.Name),
-		fmt.Sprintf("%s=%s", constants.ArgumentModelDir, constants.DefaultModelLocalMountPath),
-		fmt.Sprintf("%s=%s", constants.ArgumentHttpPort, constants.InferenceServiceDefaultHttpPort),
-	}
-	if extensions.ContainerConcurrency != nil {
-		arguments = append(arguments, fmt.Sprintf("%s=%s", constants.ArgumentWorkers, strconv.FormatInt(*extensions.ContainerConcurrency, 10)))
-	}
+	k.Container.Env = append(
```
yuzisun, line 55: Does MLServer support command arguments ? If we can use the same set of the arguments that helps maintain the backwards compatibility.

config/configmap/inferenceservice.yaml
```diff
@@ -24,8 +24,8 @@ data:
             "multiModelServer": "false"
         },
         "sklearn": {
-            "image": "gcr.io/kfserving/sklearnserver",
-            "defaultImageVersion": "v0.4.1",
+            "image": "docker.io/seldonio/mlserver",
+            "defaultImageVersion": "0.1.2",
```
yuzisun, line 28: Let's still keep the default to old v1 version for a few releases and then we can start defaulting to mlserver once they migrate away from v1. With v1beta1 user can specify the image and version on the inference service spec itself if they want to use mlserver.

yuzisun:
thanks @adriangonz, Thanks is a great start! 
- I think for the initial v1beta1 release we still default to kfserver, but we can implement in a way that both kfserver and mlserver work with the sklearn spec, e.g if user specify mlserver image and version on inference service spec like following then it creates mlserver to load the sklearn model.
```
sklearn:
    image: seldon.io/mlserver
    runtimeVersion: v0.1.2
    storageUri: gs://kfserving-examples/sklearn/iris
```
Does mlserver model repo works the same way as the current kfserver ?
- for MMS I think we can punt on supporting multi frameworks when use case comes up.

adriangonz:
Thanks for your comments @yuzisun! To make it more explicit for the end user, do you think it would make sense to have a `protocol` flag in `PredictorExtensionSpec`? We could potentially leverage the same flag in Triton to enable / disable the V2 protocol (although, I'm not sure if Triton allows that in recent images).

yuzisun:
> Thanks for your comments @yuzisun! To make it more explicit for the end user, do you think it would make sense to have a `protocol` flag in `PredictorExtensionSpec`? We could potentially leverage the same flag in Triton to enable / disable the V2 protocol (although, I'm not sure if Triton allows that in recent images).

I think that's good idea, we make it more explicit on the spec and user knows which protocol model server supports.
- tensorflow defaults to v1, no v2 support
- sklearn/xgboost defaults to v1 for now but user can specify `protocol: v2` to use mlserver
- triton defaults to v2 @deadeyegoodwin triton no longer supports v1 correct?

deadeyegoodwin:
Triton only supports KFServing V2 protocol. V1 support was dropped a couple of months ago.

config/configmap/inferenceservice.yaml
```diff
@@ -24,12 +24,22 @@ data:
...
             ],
             "multiModelServer": "false"
+          },
+          "v2": {
+            "image": "docker.io/seldonio/mlserver",
+            "defaultImageVersion": "0.1.2",
+            "supportedFrameworks": [
+              "sklearn"
+            ],
+            "multiModelServer": "false"
```
yuzisun, line 41: @adriangonz Can you help add these config to the test overlay as well https://github.com/kubeflow/kfserving/blob/master/config/overlays/test/configmap/inferenceservice.yaml ? The CI crashed on 

```
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1.(*SKLearnSpec).getContainerV1(0xc0008a7080, 0xc0008be2e0, 0x16, 0x0, 0x0, 0xc0008be300, 0x15, 0xc000453b20, 0x6c, 0xc000280c30, ...)
	/go/src/github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1/predictor_sklearn.go:86 +0x356
github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1.(*SKLearnSpec).GetContainer(0xc0008a7080, 0xc0008be2e0, 0x16, 0x0, 0x0, 0xc0008be300, 0x15, 0xc000453b20, 0x6c, 0xc000280c30, ...)
	/go/src/github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1/predictor_sklearn.go:68 +0xb3
github.com/kubeflow/kfserving/pkg/controller/v1beta1/inferenceservice/components.(*Predictor).Reconcile(0xc000926180, 0xc0009d7000, 0x1, 0x1)
```

adriangonz: Done!

adriangonz:
Hey @yuzisun , I just amended the overlay to add the new image. I've also added a new integration test for the SKLearn predictor with the V2 protocol.

yuzisun:
This is really awsome work! thanks @adriangonz !
/lgtm

@cliveseldon If this looks good to you, can you approve?

cliveseldon:
Indeed. Great addition. /approve

PR #1157, opened 2020-10-23
Add yq as dep in developer guide 

iancoffey:

The Makefile requires yq to deploy, so this just adds it as a dep.


yuzisun:
that's my oversight! thanks @iancoffey !

/lgtm
/approve

PR #1193, opened 2020-11-07
Update KFServing SDK Example link. Old link no longer exists

tcnt:
The original link referred to a notebook which is no longer in the docs/samples/client folder.
There are now 2 new notebooks. Hence I changed the link to use docs/samples/client
instead of referring to a specific notebook in the folder.

Also a space is removed on 2 lines at the end by my editor.


yuzisun:
Thanks @tcnt for the fix!

/lgtm
/approve

PR #1194, opened 2020-11-09
config: Fix port number of webhook in manager Pod

yanniszark:

The webhook-server port in the manager Pod was set to 443, while it was
changed to 9443 [1]. Fix the discrepancy and refer to the port by name.

**Release note**:
```release-note
Fix the webhook-server's port number in the manager Pod.
```


yuzisun (commented):
good catch! Thanks @yanniszark !

/lgtm
/approve

PR #1195, opened 2020-11-09
Create github action for python testing and pypi publishing

yuzisun:
- Add github action for python packages pytest
- Add github action for publishing pypi artifacts

Fixes #

1. Regenerate TrainedModel SDK after downgrading to v1alpha1
2. Add github action for python test CI and remove from workflow

**Release note**:

```release-note
KFServing python SDK should be automatically published when tags are created
```


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
jinchihe, line 25: I wonder where the env set? Thanks.

jinchihe (commented):
That's looks great @yuzisun please resolve conflicts.

jinchihe:
@yuzisun for each publishing, the version in setup.py should be updated with the tag? right?

.github/workflows/python-publish.yml
yuzisun: github settings can configure the secrets and they are encrypted, the value can be referenced like above with the configured variable names.

yuzisun:
> @yuzisun for each publishing, the version in setup.py should be updated with the tag? right?

the version in setup.py should match the release tag, you updated the version in setup.py last time to [0.5.0](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/setup.py#L28),

jinchihe:
OK， Thanks

/lgtm

PR #1206, opened 2020-11-14
[Docs] Add OpenShift installation guide.

pvaneck:
These docs cover basic installation of KFServing on OpenShift.

Also changed the KFServing version in `quick_install.sh` to 0.4.1.

### DIFF
--- a/docs/OPENSHIFT_GUIDE.md
+++ b/docs/OPENSHIFT_GUIDE.md
@@ -1,0 +1,86 @@
+# KFServing on OpenShift
+
+To install standalone KFServing on OpenShift, the easiest way is to use the [`quick_install.sh`](./hack/quick_install.sh) script provided
+in this repository. This assumes you do not already have Istio and Knative running on your cluster.
+
+**Note**: These instructions were tested on OpenShift 4.5.15, with KFServing 0.4.1, Istio 1.6.2, and Knative 0.15.0 which
+are in the quick install script. Additionally, we have tested it with Kubeflow 1.2 recommended versions for Istio and
+Knative, i.e. Istio 1.3.1 and Knative 0.14.3.
+
+## Clone repository
+
+```bash
+git clone https://github.com/kubeflow/kfserving
+```
+
+## Add Security Context Constraint (SCC)
+
+Run the following to enable containers to run with UID 0 for Istio’s service accounts. As we make further progress,
+we will be revisiting these instructions to enable it to work without having to run containers with UID 0.
+
+
+```bash
+oc adm policy add-scc-to-group anyuid system:serviceaccounts:istio-system
+```
+
+## Run install script
+
+From the root of the `kfserving` directory, execute the following:
+
+```bash
+./hack/quick_install.sh
+```
+
+This [script](./hack/quick_install.sh) will install Istio, Knative, Cert Manager, and then finally KFServing.
+
+## Verify KFServing installation
+
+Check that the KFserving controller is running:
+
+```bash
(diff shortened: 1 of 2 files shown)


docs/OPENSHIFT_GUIDE.md
```diff
@@ -0,0 +1,83 @@
+# KFServing on OpenShift
+
+To install standalone KFServing on OpenShift, the easiest way is to use the [`quick_install.sh`](./hack/quick_install.sh) script provided
+in this repository. This assumes you do not already have Istio and Knative running on your cluster.
+
+**Note**: These instructions were tested on OpenShift 4.5.15.
```
animeshsingh, line 6:
```suggestion
**Note**: These instructions were tested on OpenShift 4.5.15, with Istio 1.3.1 and KNative 0.14.3. Our recommendation is to move to Istio 1.6+ and KNative 0.15+
```

docs/OPENSHIFT_GUIDE.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Clone repository
+
+```bash
+git clone https://github.com/kubeflow/kfserving
+```
+
+## Add Security Context Constraint (SCC)
+
+Run the following to enable containers to run with UID 0 for Istio’s service accounts.
```
animeshsingh, line 16:
```suggestion
Run the following to enable containers to run with UID 0 for Istio’s service accounts. As we make further progress, we will be revisiting these instructions to enable it to work without having to run containers with UID 0.
```

animeshsingh (commented):
thanks @pvaneck 

cc @moficodes

docs/OPENSHIFT_GUIDE.md
pvaneck: Wouldn't be accurate to say these versions since I tested standalone installation with whatever is in the quick_install script (i.e. Istio 1.6.2 and Knative 0.15.0)

animeshsingh: how about the changed line

pvaneck: Updated with both suggestions.

animeshsingh:
One concern
./hack/quick_install.sh was meant for local install on laptop. Additionally it installs the code in master?

Standalone installation recommends using kubectl
https://github.com/kubeflow/kfserving#standalone-kfserving-installation

animeshsingh:
Or it can be suggested as the first option if it works out with the tests.

docs/OPENSHIFT_GUIDE.md
```diff
@@ -0,0 +1,86 @@
...
+oc -n istio-system expose svc/istio-ingressgateway --port=http2
+```
+
+## Test KFServing installation
+
+Now, create an inference service. From the root of the `kfserving` directory, run:
+
+```bash
+oc create ns kfserving-test
+oc apply -f docs/samples/sklearn//sklearn.yaml -n kfserving-test
```
Tomcli, line 61: typo
```suggestion
oc apply -f docs/samples/sklearn/sklearn.yaml -n kfserving-test
```

pvaneck:
@animeshsingh  I think the quick_install script is good first option and probably the easiest option for installing kfserving on a fresh OpenShift, since it installs istio and knative for you.

The standalone installation kubectl command is what is run at the end of the quick install script. We can suggest that users can run that command by itself, but tell them they need to make sure that their clusters have the prereqs already installed (istio, knative, cert-manager).

moficodes (approved):

PR #1223, opened 2020-11-18
Consolidate logger to model agent sidecar and add readiness probe to model container

### LINKED ISSUE #1124: Unexpected (false-positive) model ready status when using batcher
/kind bug

Usually when deploying kfserving endpoint, the inferenceservices won't be marked as ready=True (can be monitored through `kubectl get inferenceservices` or python sdk `KFServing.get(watch=True)`) until the model gets ready (i.e., warmup finished through load function). However, this is not the case when using batcher: the inferenceservices will be marked as ready as soon as the batcher sidecar is ready and all containers running, before the model warmup finished. So in case there's a bug in model load function, if using batcher, we'll get false-positive and thus can't get the correct ready status.

Here I created a simple custom model docker (model will never be ready) and uploaded to `lplrobert/kfserving-not-ready`

```
# Dockerfile
FROM python:3.6-alpine

WORKDIR /app

COPY app.py /app

RUN apk update
RUN apk add make automake gcc g++ subversion python3-dev libffi-dev openssl-dev

RUN pip install kfserving==0.4.0

CMD ["python", "app.py"]
```

```
# app.py
import kfserving
import time

class KFServingSampleModel(kfserving.KFModel):
    def __init__(self, name):
        super().__init__(name)
        self.ready = False

    def load(self):
        time.sleep(60)
        raise RuntimeError("Model load failed!")

    def predict(self, request):
        return request

if __name__ == "__main__":
    model = KFServingSampleModel("kfserving-custom-model")
    model.load()
    kfserving.KFServer(workers=1).start([model])
```

```
# custom.yaml
apiVersion: serving.kubeflow.org/v1alpha2
kind: InferenceService
metadata:
  labels:
    controller-tools.k8s.io: "1.0"
  name: kfserving-custom-model
spec:
  default:
    predictor:
      batcher:
        maxBatchSize: 32
        maxLatency: 5000
        timeout: 60
      custom:
        name: kfserving-custom-model
        container:
          image: lplrobert/kfserving-not-ready
```

If we apply above CRD (with batcher enabled) and watch at `kubectl get inferenceservices`, we'll see `kfserving-custom-model` gets ready pretty quickly
```
NAME                              URL                                                              READY   DEFAULT TRAFFIC   CANARY TRA
FFIC   AGE
kfserving-custom-model            http://kfserving-custom-model.namespace.example.com            True    100
       15s
```

Then become unready only after waiting for a few (>2) minutes when the pod restarted and failed
```
NAME                              URL                                                              READY   DEFAULT TRAFFIC   CANARY TRA
FFIC   AGE
kfserving-custom-model                                                                             False
       2m15s
```

Then become ready again quickly when the pod retry again
```
NAME                              URL                                                              READY   DEFAULT TRAFFIC   CANARY TRA
FFIC   AGE
kfserving-custom-model            http://kfserving-custom-model.namespace.example.com            True    100
       2m27s
```

And after that, the ready status will change back and forth when the pod fails and restarts. In that case, we'll be misled to think the model works fine at first and can't be alerted until waiting for quite a long time.

While in comparison, if we apply the above CRD with batcher section removed. The inferenceservice will never be in ready status, and thus we know at first there's a bug somewhere.

If using batcher, inferenceservice should also return ready=True only if model warmup succeed, rather than when all containers are running.

**Environment:**

- Istio Version:
- Knative Version:
- KFServing Version: 0.4.0
- Kubeflow version:
- Minikube version:

yuzisun:
- Consolidate logger and model puller to single sidecar agent
- Add readiness probe for agent to ensure main model container health as queue proxy now probes the agent sidecar

**Which issue(s) this PR fixe** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1124 

1. Remove logger binary and consolidate into agent image

**Release note**:

```release-note

```


yuzisun:
@cliveseldon Can you review this, this adds the readiness probe for the logger and consolidates to a single sidecar.

cmd/agent/main.go
```diff
@@ -1,29 +1,238 @@
...
+	healthState := &health.State{}
+	ctx := signals.NewContext()
+	// Setup probe to run for checking user-application healthiness.
+	probe := buildProbe(logger, env.ServingReadinessProbe, *componentPort)
+	mainServer := buildServer(ctx, *port, *componentPort, loggerArgs, healthState, probe, logger)
+	servers := map[string]*http.Server{
+		"main": mainServer,
+	}
+	errCh := make(chan error)
+	listenCh := make(chan struct{})
```
cliveseldon, line 120: Should we reference the copy of code from https://github.com/knative/serving/blob/master/cmd/queue/main.go ?

yuzisun: sure

PR #1244, opened 2020-12-10
Adversarial robustness explainer

drewbutlerbb4:
**What this PR does / why we need it**:
Adds an Explainer tool from the Adversarial Robustness Toolbox (ART) project that can create adversarial examples and trick a model into misclassifying. The ART explainer uses a technique called SquareAttack to create the adversarial examples. A sample has been made which tests the ART explainer on a model trained on the MNIST dataset.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1243 

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #1245, opened 2020-12-10
Fix minio python client version

Tomcli:

Fixes #

The new [minio client 7.0.0](https://github.com/minio/minio-py/releases/tag/7.0.0) release introduced some breaking changes. Since some users might still using the previous version of minio for their other Python packages, we cap the minio version for now until everyone is comfortable to move beyond minio 7.0.0.

related #1244 

**Release note**:

```release-note

```


python/kfserving/requirements.txt
```diff
@@ -6,7 +6,7 @@ urllib3>=1.15.1
 kubernetes>=12.0.0
 tornado>=6.0.0
 argparse>=1.4.0
-minio>=4.0.9
+minio>=4.0.9,<7.0.0
```
yuzisun, line 9: I think this effectively means `minio==6.0.2`, so make it more explicit?

Tomcli: I think having `minio==6.0.2` will make some users have conflicts with their other packages. How about `minio>=4.0.9,<=6.0.2`?

yuzisun: I think only storage initializer currently are depending on minio.

Tomcli: yes, but kfserving SDK also using the same `requirements.txt` where the conflicts will kick in.

Tomcli: I brought up this because pip is changing how they handle the package dependencies
https://pip.pypa.io/en/latest/user_guide/#deprecation-timeline

yuzisun: ok the pip is going to be more strict.

PR #1251, opened 2020-12-15
Upgrade minio client version to 7.0.0

### LINKED ISSUE #1246: Upgrading from storage initializer 0.4.0 to 0.4.1 breaks AWS auth via IAM Roles
/kind bug

Upgrading the storage initializer image from `gcr.io/kfserving/storage-initializer:v0.4.0` to `gcr.io/kfserving/storage-initializer:v0.4.1` breaks AWS auth via IAM roles, which should otherwise be implicitly available through EC2 instance metadata.

We are running kfserving on EKS in a managed node group. Since our kfserving workloads are restricted to these nodes, we granted permissions on the instance level to read from a certain s3 bucket. All pods spawned on the nodes implicitly have S3 read permissions. So we should not need to provide credentials to the storage initializer at all. This works for all other pods spawned on the node group.

We were able have this behavior in v0.4.0 by creating a secret for the storage initializer like this:
```
apiVersion: v1
data:
  AWS_ACCESS_KEY_ID: ""
  AWS_SECRET_ACCESS_KEY: ""
kind: Secret
metadata:
  annotations:
    serving.kubeflow.org/s3-endpoint: s3.ap-east-1.amazonaws.com
    serving.kubeflow.org/s3-region: ap-east-1
    serving.kubeflow.org/s3-usehttps: "1"
    serving.kubeflow.org/s3-verifyssl: "1"
  name: storage-initializer-creds
  namespace: default
type: Opaque
```

This secret is attached to the SA that the storage initializer uses, and it works.

Now after upgrading to 0.4.1 we see a "permissions denied" error from S3. This is actually to be expected, because if you have anything set in the two credential variables, AWS SDKs will try those creds first. You usually have to initialize the client without access keys entirely to force it to try and look in other places.

After following other threads, I have tried to set other env variables in this secret, like `MINIO_ACCESS_KEY` and `MINIO_SECRET_KEY`. Setting those gets me an "illegal location constraint ap-east-1" on my bucket which is definitely in ap-east-1.

To get around this error, we are using everything 0.4.1, except for the storage initializer which we rolled back to 0.4.0
It works that way for now.

The ideal behavior would be that the storage initializer just follows the standard practices here. We should be able to omit the AWS access keys entirely, and see it pull auth from the EC2 metadata.

I believe this is purely a minio problem, but I cannot for the life of me find a thread where someone has a fix. And the fact that the auth handling changed this drastically in a patch version is a smell.

The kfserving example docs provide sample ways of doing this using AWS access keys only. But using keys instead of roles is dangerous. Handling this with roles ought to be the default example.

**Environment:**

- Istio Version:
   client version: 1.7.2
   control plane version: 1.7.2
   data plane version: 1.7.2
- Knative Version: v0.17.0
- KFServing Version: 0.4.1
- Kubernetes version: (use `kubectl version`): Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
- OS (e.g. from `/etc/os-release`): Amazon Linux EKS CPU AMI

yuzisun:

Fixes #1246 

**Release note**:

```release-note

```


PatrickXYS:
/test kubeflow-kfserving-presubmit

Test to see if E2E test works fine without Unauthorized issue

If this test works fine, and 401 Unauthorized error should only relate to this PR's test code https://github.com/kubeflow/kfserving/pull/1306

PR #1254, opened 2020-12-15
Reference local copy of predict v2 grpc proto file

kylegc:
This PR adds a grpc inferencing proto file which follows the spec https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md

Then this PR updates the doc to point to the local spec rather than pointing to the triton-inference-server version.

This is needed because some organizations my not be keen on or able to pull in the proto file with the NVIDIA copyright.

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/predict-api/v2/README.md
+++ b/docs/predict-api/v2/README.md
@@ -9,7 +9,7 @@ standardized around this protocol.
 The protocol is composed of a required set of APIs that must be
 implemented by a compliant server. This required set of APIs is
 described in [required_api.md](./required_api.md). The [GRPC proto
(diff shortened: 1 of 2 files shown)


kylegc:
@googlebot I signed it!

animeshsingh:
thanks @kylegc 
/ok-to-test

animeshsingh:
/lgtm

@deadeyegoodwin any comments?

pugangxa:
I think we are going to generate grpc server/client and even rest code based on this, so maybe we can move it to another directlry(say api/) rather than under /docs.

PR #1258, opened 2020-12-16
Fixes the broken build - logger

### LINKED ISSUE #1257: make all fails with logger not found
/kind bug

The `make all` fails because this fails to run ` go build -o bin/logger ./cmd/logger` . The `cmd` does not have `logger`

`make` to succeed. 

**Environment:**

- Istio Version:
- Knative Version:
- KFServing Version:
- Kubeflow version:
- Minikube version:

naveensrinivasan:
The make fails with the broken build because the logger is not present in the repository.

Fixes the broken build
**Which issue(s) this PR fixes will close the issue(s) when PR gets merged)**:
Fixes #1257

NONE


pugangxa:
Thanks for your PR, would you please also remove other useless target for logger, like docker-build-logger, docker-push-logger in Makefile, and also the dockerfile for logger?

naveensrinivasan:
> Thanks for your PR, would you please also remove other useless target for logger, like docker-build-logger, docker-push-logger in Makefile, and also the dockerfile for logger?

@pugangxa  Took care of it. Thanks

logger.Dockerfile
```diff
@@ -1,22 +0,0 @@
...
-
-# Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o agent ./cmd/agent
-
-# Copy the inference-agent into a thin image
-FROM gcr.io/distroless/static:latest
-COPY third_party/ third_party/
-WORKDIR /
-COPY --from=builder /go/src/github.com/kubeflow/kfserving/agent .
-ENTRYPOINT ["/agent"]
```
yuzisun, line 22: Do not remove this yet, it is still part of the CI although it produces the same image as agent. We will take care of this.

pugangxa: Got it. So we can remove it later after CI fixed.

naveensrinivasan:
@PatrickXYS Thanks! 

Looks like the build succeeded!

logger.Dockerfile
PatrickXYS: @naveensrinivasan 

Seems like you should not remove this as of now

naveensrinivasan: Thanks

yuzisun:
Thanks @naveensrinivasan ! We have fixed this, so the PR is no longer needed, thanks for reporting the issue.

PR #1262, opened 2020-12-18
Remove redundant code in ONNX InferenceService sample

nuka137:

Remove redundant code in ONNX InferenceService sample notebook.
Some modules are imported in multiple times.

### DIFF
--- a/docs/samples/v1alpha2/onnx/mosaic-onnx.ipynb
+++ b/docs/samples/v1alpha2/onnx/mosaic-onnx.ipynb
@@ -134,10 +134,6 @@
    "outputs": [],
    "source": [
     "# Create request message to be sent to the predictor\n",
-    "import assets.onnx_ml_pb2 as onnx_ml_pb2\n",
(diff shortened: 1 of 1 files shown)


adriangonz:
Hey @nuka137, aren't those imports used down the line within the same notebook?

nuka137:
@adriangonz 

Yes, that's right.
But, these import lines also exist in third code block.

```python
from PIL import Image
import numpy as np
import google.protobuf.json_format as json_format
import assets.onnx_ml_pb2 as onnx_ml_pb2
import assets.predict_pb2 as predict_pb2
import requests
```

So, I think we can remove redundant import lines.

nuka137:
@adriangonz 

> But, these import lines also exist in third code block.
> So, I think we can remove redundant import lines.

Is there any comments about my idea?

adriangonz:
Good spot @nuka137 ! Looking good from my side, although there seems to be a failing test. I've kicked it off again. If it passes, this looks good from my side.

/test kubeflow-kfserving-presubmit

PR #1270, opened 2020-12-26
Generate v1beta1 API docs

yuzisun:

Fixes #

1. Use `kubectl replace` instead of `kubectl apply` as it takes long time to do the diff on client side.
2. Review generated v1beta1 api docs

**Release note**:

```release-note

```


Makefile
```diff
@@ -52,7 +52,8 @@ deploy: manifests
 	cd config/default && if [ ${KFSERVING_ENABLE_SELF_SIGNED_CA} != false ]; then \
 	kustomize edit remove resource certmanager/certificate.yaml; \
 	else kustomize edit add resource certmanager/certificate.yaml; fi;
-
+	kubectl get crd inferenceservices.serving.kubeflow.org || kustomize build config/crd | kubectl create --validate=false -f -
+	kustomize build config/crd | kubectl replace --validate=false -f -
```
cliveseldon, line 56: Could this not be added to first part above to make it a single op?

cliveseldon: Also, won't the apply below for config/default do an "apply" again on the CRD as its a base of that kustomize?

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -42,19 +42,22 @@ const (
...
 // AlibiExplainerSpec defines the arguments for configuring an Alibi Explanation Server
 type AlibiExplainerSpec struct {
-	// The type of Alibi explainer
-	// Valid values are:
-	// - "AnchorTabular";
-	// - "AnchorImages";
-	// - "AnchorText";
-	// - "Counterfactuals";
-	// - "Contrastive";
+	// The type of Alibi explainer <br />
```
cliveseldon, line 45: So new lines are not added to generated docs? This seems annoying we need to do everywhere.

Makefile
yuzisun: Combined to the single command, it will apply the crd again but it is pretty fast as the crd created or replaced does not have the `last-applied-annotation`.

pkg/apis/serving/v1beta1/explainer_alibi.go
yuzisun: it is a bit annoying but most of fields have single line comment.

PR #1284, opened 2021-01-08
Workaround for kustomize bug triggered in metrics and monitoring instructions

sriumcp:

The sample instructions for getting started with metrics monitoring for KFServing models (see [here](https://github.com/kubeflow/kfserving/tree/master/docs/samples/metrics-and-monitoring)) triggers a bug in Kustomize.

The Kustomize bug is reported here: https://github.com/kubernetes-sigs/kustomize/issues/3430

Meanwhile, this simple one-field change in the kustomization resources provides a workaround which enables Prometheus monitoring to work as claimed.

@Tomcli @animeshsingh 

This is a pure documentation PR.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
+++ b/docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
@@ -9,4 +9,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: prometheus
-  namespace: kfserving-monitoring
+  namespace: default
--- a/docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
+++ b/docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
@@ -9,4 +9,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: prometheus
-  namespace: default
+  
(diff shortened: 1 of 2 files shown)


docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
```diff
@@ -9,4 +9,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: prometheus
-  namespace: kfserving-monitoring
\ No newline at end of file
+  namespace: default
```
Tomcli, line 12: Maybe we can remove the namespace section and let kustomize to handle it? It is possible for kustomize v3 unless kustomize did some changes to the latest version.

e.g. https://github.com/kubeflow/manifests/blob/master/kfserving/kfserving-install/base/cluster-role-binding.yaml

Tomcli:
thanks @sriumcp 
/lgtm
/assign @animeshsingh

docs/samples/metrics-and-monitoring/prometheus/clusterrolebinding.yaml
sriumcp: Yes. The latest commit removes namespace field altogether. It is still Kustomizable at the top-level so all good.

PR #1309, opened 2021-01-20
add dependabot config script

DavidSpek:
Inspired by https://github.com/kubeflow/pipelines/issues/4682 I created a script that will create a config file for depandabot so that it knows what directories to scan. It will scan the repository for files named `*ockerfile*`, `package*.json`, `*requirements.txt` and `go.*`. It is setup for dockerfiles, gomod, npm and python at the moment, and I believe this should cover almost all the code in the repo. It is trivial to further customize what folders are selected if further customization is needed. It also parses the closest OWNERS file for a given directory and assigns the approvers and adds the reviewers to the PRs it creates. 

This is a sibling PR to https://github.com/kubeflow/pipelines/pull/5015 and https://github.com/kubeflow/kubeflow/pull/5542. 

As it stands now, there are about 56 PRs (excluding go as it is still busy scanning) that will be created with this configuration, so it might be advisable to have some form of plan to implement it in stages or be ready to quickly go through lots of the PRs. Another option is to create a target branch for all these PRs so they can be merged into that first rather than master.

For reference, the PRs that will be created can be found here: https://github.com/DavidSpek/kfserving/pulls


hack/create_dependabot.py
```diff
@@ -0,0 +1,100 @@
...
+                go_clean_list.append(go_file.parents[0])
+    return go_clean_list
+
+def append_updates(ecosystem, directory, assignees, reviewers=None):
+    config = {}
+    config['package-ecosystem'] = ecosystem
+    config['directory'] = directory
+    config['schedule']= {}
+    config['schedule']['interval'] = 'daily'
+    config['open-pull-requests-limit'] = 10
```
pugangxa, line 59: So this limit is just for the folder right? And it's still one PR for one update so I guess it will be a lot of PRs created at first time. 
Can we do it manually for the initial upgrade and then we can depend on the bot and there should not be too much at one time.

DavidSpek: The limitation is for 1 folder and for that specific package system yes. There will be one PR per dependency. The first batch is 56 PRs, which I think is still manageable. The way to reduce it is to delete part yaml at first.

yuzisun:
Closing staled PR

PR #1343, opened 2021-02-02
[#1159] Support cloud events for KFServing custom framework

### LINKED ISSUE #1159: Support cloud events for KFServing custom framework
/kind feature

As a user when doing feature transformation the payload can be protobuf/avro or some other binary format delivered via HTTP or KAFKA in cloudevent binary format(https://github.com/cloudevents/spec/blob/v1.0.1/kafka-protocol-binding.md#32-binary-content-mode) but currently KFServing custom framework [only supports JSON](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/kfserving/handlers/http.py#L51).

- I'd like to send cloud event to KFServer via KAFKA in following format and KFServer pass the decoded cloud event to `preprocess` or `predict` handler
  
```
------------------ Message -------------------
Topic Name: mytopic
------------------- key ----------------------
Key: mykey
------------------ headers -------------------
ce_specversion: "1.0"
ce_type: "com.example.someevent"
ce_source: "/mycontext/subcontext"
ce_id: "1234-1234-1234"
ce_time: "2018-04-05T03:56:24Z"
content-type: application/protobuf
------------------- value --------------------
      ... application data encoded in protobuf ...
-----------------------------------------------

             consume                       http(cloudevent)
KAFKA ----------> Kafka Event Source -----------------> InferenceService

```

```python
def has_ce_headers(headers: typing.Dict[str, str]) -> bool:
    return (
        "ce-specversion" in headers
        and "ce-source" in headers
        and "ce-type" in headers
        and "ce-id" in headers
    )
    

 if has_ce_headers(self.request.headers):
        event = from_http(self.request.headers, self.request.body)
        body = event['data']
 else:
        try:
            body = json.loads(self.request.body)
        except json.decoder.JSONDecodeError as e:
            raise tornado.web.HTTPError(
                 status_code=HTTPStatus.BAD_REQUEST,
                 reason="Unrecognized request format: %s" % e)
 ```                   
                    
- Currently HTTP headers are not exposed to the `preprocess` or `predict` handler, if a cloud event is sent to KFServer it is not possible for user to check the http headers which can be useful for tracing and examining the original event source(kafka or http).
`async def predict(self, request: Dict, headers: Dict) -> Dict:`
`async def preprocess(self, request: Dict, headers: Dict) -> Dict:`

- Also it would be nice if KFServing can generate cloud event response so that it works with knative eventing channels seamlessly.

### PR DESCRIPTION
**Pull request by mtickoobb**
Support cloud events for KFServing custom framework

Fixes #1159 

**Release note**:

```release-note

```


python/kfserving/kfserving/handlers/http.py
```diff
@@ -43,21 +47,46 @@ def validate(self, request):
             )
         return request
 
+    def has_ce_headers(self, headers: typing.Dict[str, str]) -> bool:
+        return (
+            "ce-specversion" in headers
+            and "ce-source" in headers
+            and "ce-type" in headers
+            and "ce-id" in headers
+        )
```
wengyao04, line 56: could we call=```has_binary_headers``` https://github.com/cloudevents/sdk-python/blob/master/cloudevents/sdk/converters/util.py#L4

```from cloudevents.sdk.converters.util import has_binary_headers```

File: python/kfserving/kfserving/handlers/http.py [hunk 2]
```diff
@@ -43,21 +47,46 @@ def validate(self, request):
...
 
 class PredictHandler(HTTPHandler):
     async def post(self, name: str):
+        if self.has_ce_headers(self.request.headers):
+            body = from_http(self.request.headers, self.request.body)
```
wengyao04, line 61: I think we also need to add catch exception for decoding error
from_http raises exceptions https://github.com/cloudevents/sdk-python/blob/master/cloudevents/http/http_methods.py#L14

yuzisun: @wengyao04 is right, if we do not catch this then we tornado by default returns HTTP 500 for uncaught exception but here we really want HTTP 400(bad request).

python/kfserving/kfserving/handlers/http.py
mtickoobb: Thanks @wengyao04 for pointing this out. 
Have updated now.

python/kfserving/kfserving/handlers/http.py
mtickoobb: Good catch @wengyao04 .
I've updated this.

mtickoobb:
@googlebot I fixed it

mtickoobb:
Upstream lib deps fixed in [PR](https://github.com/kubeflow/kfserving/pull/1349)
Will try again after it is merged

yuzisun:
@PatrickXYS failed multiple times on this PR too :( 

```
W
                    namespace=namespace,
                    timeout_seconds=timeout_seconds)
            else:
                try:
                    return self.api_instance.get_namespaced_custom_object(
                        constants.KFSERVING_GROUP,
                        version,
                        namespace,
                        constants.KFSERVING_PLURAL,
                        name)
                except client.rest.ApiException as e:
>                   raise RuntimeError(
                        "Exception when calling CustomObjectsApi->get_namespaced_custom_object:\
                        %s\n" % e)
E                   RuntimeError: Exception when calling CustomObjectsApi->get_namespaced_custom_object:                        (401)
E                   Reason: Unauthorized
E                   HTTP response headers: HTTPHeaderDict({'Audit-Id': '1d6b84f2-bae5-4215-9ab0-d5af1c4cccab', 'Cache-Control': 'no-cache, private', 'Content-Type': 'appl
ication/json', 'Date': 'Thu, 04 Feb 2021 21:05:45 GMT', 'Content-Length': '129'})
```

PatrickXYS:
@yuzisun Can you send another PR to add retry mechanism? 

Ref: https://github.com/kubeflow/kfserving/pull/1339#issuecomment-773448921

yuzisun:
This is really awesome work! Thanks @mtickoobb !

/lgtm
/approve

PR #1361, opened 2021-02-05
Allow any .onnx file name in v1beta1 api

markwinter:

Allow any model file name with `.onnx` extension or a storageURI with no extension.

This was added in `v1alpha2` but was not ported to `v1beta1` https://github.com/kubeflow/kfserving/pull/1078


yuzisun:
@markwinter Good catch! thanks for the contribution!

yuzisun:
@markwinter Can you help do a rebase from master, I think you missed some commits for the CI migration.

markwinter:
@yuzisun Ah I missed that. I've rebased and pushed it now.

Thanks for reviewing my PR

PR #1363, opened 2021-02-06
Make kative local gateway configurable

### LINKED ISSUE #1340: Istio gateways are hard coded and anti-pattern
kfserving
/kind feature

You should be able to specify any (or none!) ingress-gateway/cluster-local-gateway when using kfserving

Right now the values are hardcoded in `pkg/constants/constants.go`:
```
// Knative constants
const (
    KnativeLocalGateway   = "knative-serving/cluster-local-gateway"
    KnativeIngressGateway = "knative-serving/knative-ingress-gateway"
    VisibilityLabel       = "serving.knative.dev/visibility"
)

var (
    LocalGatewayHost = "cluster-local-gateway.istio-system.svc." + network.GetClusterDomainName()
)
```

It is a misnomer because in knative these are not strictly constants; you can use any gateway you want

You ought to be able to configure it like so:
```
// Knative constants
const (
    KnativeLocalGateway   = "<namespace>/<gateway name>"
    KnativeIngressGateway = "<namespace>/<gateway name>"
    VisibilityLabel       = "serving.knative.dev/visibility"
)

var (
    LocalGatewayHost = "<gateway/envoy name>.<namespace>.svc." + network.GetClusterDomainName()
)
```

Moreover when a InferenceService creates an istio VirtualService it sets the route from the host (example `sklearn-iris.default.example.com`) to `cluster-local-gateway.istio-system.svc.cluster.local` and sets the host header as `iris-predictor-default.default.svc.cluster.local`. In my understanding this is fundamentally an incorrect use of the cluster local gateway. 

The source says:
```
    # A cluster local gateway to allow pods outside of the mesh to access
    # Services and Routes not exposing through an ingress.  If the users
    # do have a service mesh setup, this isn't required and can be removed.
    #
    # An example use case is when users want to use Istio without any
    # sidecar injection (like Knative's istio-ci-no-mesh.yaml).  Since every pod
    # is outside of the service mesh in that case, a cluster-local  service
    # will need to be exposed to a cluster-local gateway to be accessible.
 ```

The cluster local gateway should be an entrance point for services which *do not* run in the mesh. If you are doing everything according to best practices you don't even need it. So a possible fix would be to route the traffic directly to the service, namely `iris-predictor-default.default.svc.cluster.local` and not through the cluster local gateway at all.

TLDR Kfserving should let you BYO knative and istio with no assumptions. Moreover, by forcing the need for a cluster-local-gateway, this breaks potential for mesh-only setups.

Imho it is strange that kfserving by default wants to serve on a ingress rather than only in the mesh - it makes much more sense in a production setting to have models served only internally and let some other api do the dealing with callers.

### PR DESCRIPTION
**Pull request by yuzisun**

Fixes #1340 

**Release note**:

```release-note
Added knative local gateway to the inference service config file
```


config/configmap/inferenceservice.yaml
```diff
@@ -149,6 +149,8 @@ data:
     {
         "ingressGateway" : $(ingressGateway)
         "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local"
+        "localGateway" : "cluster-local-gateway"
+        "localGatewayService" : "cluster-local-gateway.istio-system.svc.cluster.local"
```
theofpa, line 153:
```suggestion
        "localGateway" : "knative-local-gateway"
        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local"
```

config/overlays/test/configmap/inferenceservice.yaml
```diff
@@ -117,7 +117,9 @@ data:
   ingress: |-
     {
         "ingressGateway" : "knative-ingress-gateway.knative-serving",
-        "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local"
+        "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
+        "localGateway": "cluster-local-gateway.knative-serving",
+        "localGatewayService" : "cluster-local-gateway.istio-system.svc.cluster.local"
```
theofpa, line 122:
```suggestion
        "localGateway": "knative-local-gateway.knative-serving",
        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local"
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -86,7 +86,9 @@ var _ = Describe("v1beta1 inference service controller", func() {
             }`,
 			"ingress": `{
                "ingressGateway": "knative-serving/knative-ingress-gateway",
-               "ingressService": "test-destination"
+               "ingressService": "test-destination",
+               "localGateway": "knative-serving/cluster-local-gateway",
+               "localGatewayService": "cluster-local-gateway.istio-system.svc.cluster.local"
```
theofpa, line 91:
```suggestion
               "localGateway": "knative-serving/knative-local-gateway",
               "localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local"
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler_test.go
```diff
@@ -479,8 +479,10 @@ func TestCreateVirtualService(t *testing.T) {
 				testIsvc.Spec.Explainer = &v1beta1.ExplainerSpec{}
 			}
 			ingressConfig := &v1beta1.IngressConfig{
-				IngressGateway:     constants.KnativeIngressGateway,
-				IngressServiceName: "someIngressServiceName",
+				IngressGateway:          constants.KnativeIngressGateway,
+				IngressServiceName:      "someIngressServiceName",
+				LocalGateway:            constants.KnativeLocalGateway,
+				LocalGatewayServiceName: "cluster-local-gateway.istio-system.svc.cluster.local",
```
theofpa, line 485:
```suggestion
				LocalGatewayServiceName: "knative-local-gateway.istio-system.svc.cluster.local",
```

theofpa (changes requested):
Shouldn't we use the new localGateway convention of knative (`knative-local-gateway`) as the default value, and users of old versions can change that to `cluster-local-gateway`?

yuzisun:
> Shouldn't we use the new localGateway convention of knative (`knative-local-gateway`) as the default value, and users of old versions can change that to `cluster-local-gateway`?

@theofpa Yes let’s do that in your PR, I want to first make it configurable and run the e2e rests to confirm it works with the old value.

theofpa:
> > Shouldn't we use the new localGateway convention of knative (`knative-local-gateway`) as the default value, and users of old versions can change that to `cluster-local-gateway`?
> 
> @theofpa Yes let’s do that in your PR, I want to first make it configurable and run the e2e rests to confirm it works with the old value against older knative version as you are upgrading knative version in e2e.

ok!
/lgtm

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -110,8 +110,10 @@ type InferenceServicesConfig struct {
 
 // +kubebuilder:object:generate=false
 type IngressConfig struct {
-	IngressGateway     string `json:"ingressGateway,omitempty"`
-	IngressServiceName string `json:"ingressService,omitempty"`
+	IngressGateway          string `json:"ingressGateway,omitempty"`
+	IngressServiceName      string `json:"ingressService,omitempty"`
+	LocalGateway            string `json:"localGateway,omitempty"`
+	LocalGatewayServiceName string `json:"localGatewayService,omitempty"`
```
theofpa, line 116: @yuzisun, I've missed that, but it generates a name mismatch violation in the `make generate`. We can either match the names or add it in the violation exceptions.

config/configmap/inferenceservice.yaml
```diff
@@ -149,6 +149,8 @@ data:
     {
         "ingressGateway" : $(ingressGateway)
         "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local"
+        "localGateway" : "cluster-local-gateway"
```
theofpa, line 152: missing commas for valid json

PR #1372, opened 2021-02-10
Feature to include headers in http/https service requests (non-MMS)

### LINKED ISSUE #1351: Support headers in http(s) uri requests for inference service
/kind feature

## InferenceService headers support in http(s) uri requests
Downloading model files from a http/https service may require authorization passed through the header. The problem can be solved by storing header details into secrets and extracting them to use when the storage initializer is downloading the model. This will be for non MMS use case.

### Summary
Inference service currently is unable to pass along credentials when making http/https GET requests. Http/https services that were normally unable to be queried due to lack of credentials can work by including the necessary headers. Anyone that requires authentication or additional details passed along with a query will benefit from this change.

### Motivation
There is no method for allowing credentials in the headers when making requests to an http/https service. There should be a flexible solution that makes use of k8s secrets/annotations to store this information.

### Goals
- Storage initializer will be able to download models (tar/zip) from a http/https service that requires header information. (ie: account_name and/or api_key)
- Stores headers in a k8s secret.
- Dynamically determines if a uri has corresponding headers to include.
	
### Non-Goals
- Model agent will be able to download models (tar/zip) from http/https services that require header information.

#### Option 1: Host URI in annotation and headers in data of secret

The idea is to store the host uri of the website in the annotation and the headers in the data of a k8s secret. These values will be stored as environment variables. More specific details will be shown below about the environment variable. In the storage initializer, a request to get the environment variable of the host uri which will return the list of headers. These headers will then get stored in the headers dictionary and added to the request.

##### Option 1.0.1: Prefixed values
**Secret.yml**
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  annotations:
    serving.kubeflow.org/https-host-uri: “example.com”
type: Opaque
data:
  header.account-name: “some_account_name”
  header.secret-key: “some_secret_key”
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
secrets:
  - name: mysecret
```

##### Option 1.0.2: Dedicated fields for header
**Secret.yml**
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  annotations:
    serving.kubeflow.org/https-host-uri: “example.com”
type: Opaque
data:
  headers: |-
    account-name: “some_account_name”
    secret-key: “some_secret_key”
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
secrets:
  - name: mysecret
```

The below shows different suggestions on how to store the headers in environment variables.

Note that both suggestions data will follow the same format <type>.<name>: <value>. This will allow for future flexibility in specifying what to do with the data element key. (ie proxy.proxy_key:proxy_value)

##### Option 1.1: Comma separated headers
Environment variables will be created as the following:
```
<https-host-uri-value>: <comma-separated-data-keys>
<data-element-key>: <data-element-value> # For each data element
```
And in this case:
```
example.com: “header.account-name,header.secret-key”
header.account-name: “some_account_name”
header.secret-key: “some_secret_key”
```

The benefits of this approach is obtaining the list of data keys can quickly be obtained by splitting the string by the comma. The drawback is that this list may be extremely long.

##### Option 1.2: Numbered headers count
Environment variables will be created as the following:
```
<https-host-uri-value>: <number-of-data-element>
<https-host-uri-#>:<data-element-key-#> # For each data element key
<data-element-key>: <data-element-value> # For each data element
```
And in this case:
```
example.com: 2
example.com-1: “header.account-name”
example.com-2: “header.secret-key”
header.account-name: “some_account_name”
header.secret-key: “some_secret_key”
```

The benefit of this approach is there will not be any issues with having too many headers/data keys. But this approach requires iterating through 0 to N (where N is the number of data keys) which can be costly.

Limitations for both cases:
- Does not support same uri but multiple credentials

#### Option 2: Modifying the inference service CRD
The storageURI can be modified to store potential headers. In the inference service, the storageURI can be modified to look like:
```
storageURI:
uri: “example.com”
headers:
account-name: “some_account_name”
secret-key: “some_secret_key”
```
The benefit is being able to explicitly set headers for each URI but changing the definition for inference service does not seem ideal.

I'm currently working on a PR for option 1.1

abchoo:
**What this PR does / why we need it**:  Currently there is no header support for inference service uri request. It would be beneficial as authentication for requests may require credentials to be passed into the header.

Fixes #1351

**Release note**:

```release-note

```

### DIFF
--- a/docs/samples/storage/uri/README.md
+++ b/docs/samples/storage/uri/README.md
@@ -19,7 +19,7 @@ metadata:
   name: mysecret
 type: Opaque
 data:
-  https-host-uri: ZXhhbXBsZS5jb20=
+  https-host: ZXhhbXBsZS5jb20=
   headers: |-
     YWNjb3VudC1uYW1lOiBzb21lX2FjY291bnRfbmFtZQpzZWNyZXQta2V5OiBzb21lX3NlY3JldF9rZXk=
 ---
@@ -30,9 +30,9 @@ metadata:
 secrets:
   - name: mysecret
 ```
-Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host uri.
+Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host.
 
-You will need to base64 encode the headers and host uri. Make sure each header is on a newline with the format `header_key: header_value`.
+You will need to base64 encode the headers and host. Make sure each header is on a newline with the format `header_key: header_value`.
 ```text
 example.com
 # echo -n "example.com" | base64
--- a/docs/samples/storage/uri/README.md
+++ b/docs/samples/storage/uri/README.md
@@ -8,6 +8,39 @@ This `storageUri` option supports single file models, like `sklearn` which is sp
 2. Your cluster's Istio Ingress gateway must be network accessible.
 3. Your cluster's Istio Egress gateway must [allow http / https traffic](https://knative.dev/docs/serving/outbound-network-access/)
 
+## Create HTTP/HTTPS header Secret and attach to Service account
+If you do not require headers in your HTTP/HTTPS service request then you can skip this step.
+You can define headers by following the following format:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+  annotations:
+    serving.kubeflow.org/https-host-uri: “example.com”
+type: Opaque
+data:
+  headers: |-
+    YWNjb3VudC1uYW1lOiBzb21lX2FjY291bnRfbmFtZQpzZWNyZXQta2V5OiBzb21lX3NlY3JldF9rZXk=
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa
+secrets:
+  - name: mysecret
+```
+Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host uri.
+
+You will need to base64 encode the headers
+```
+account-name: some_account_name
+secret-key: some_secret_key
+# Base64 encoded into:
+YWNjb3VudC1uYW1lOiBzb21lX2FjY291bnRfbmFtZQpzZWNyZXQta2V5OiBzb21lX3NlY3JldF9rZXk=
+```
+
 ## Sklearn
 ### Train and freeze the model
 Here, we'll train a simple iris model. Please note that `kfserving` requires `sklearn==0.20.3`. 
--- a/docs/samples/storage/uri/README.md
+++ b/docs/samples/storage/uri/README.md
@@ -33,8 +33,8 @@ secrets:
 ```
 Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host uri.
 
-You will need to base64 encode the headers
-```
+You will need to base64 encode the headers. Make sure each header is on a newline with the format `header_key: header_value`.
+```text
 account-name: some_account_name
 secret-key: some_secret_key
 # Base64 encoded into:
--- a/docs/samples/storage/uri/README.md
+++ b/docs/samples/storage/uri/README.md
@@ -10,7 +10,7 @@ This `storageUri` option supports single file models, like `sklearn` which is sp
 
 ## Create HTTP/HTTPS header Secret and attach to Service account
 If you do not require headers in your HTTP/HTTPS service request then you can skip this step.
-You can define headers by following the following format:
+You can define headers using the following format:
 
 ```yaml
 apiVersion: v1
@@ -21,7 +21,7 @@ type: Opaque
 data:
   https-host: ZXhhbXBsZS5jb20=
   headers: |-
-    YWNjb3VudC1uYW1lOiBzb21lX2FjY291bnRfbmFtZQpzZWNyZXQta2V5OiBzb21lX3NlY3JldF9rZXk=
+    ewoiYWNjb3VudC1uYW1lIjogInNvbWVfYWNjb3VudF9uYW1lIiwKInNlY3JldC1rZXkiOiAic29tZV9zZWNyZXRfa2V5Igp9
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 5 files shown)


docs/samples/storage/uri/README.md
```diff
@@ -8,6 +8,39 @@ This `storageUri` option supports single file models, like `sklearn` which is sp
...
+  - name: mysecret
+```
+Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host uri.
+
+You will need to base64 encode the headers. Make sure each header is on a newline with the format `header_key: header_value`.
+```text
+account-name: some_account_name
+secret-key: some_secret_key
+# Base64 encoded into:
+YWNjb3VudC1uYW1lOiBzb21lX2FjY291bnRfbmFtZQpzZWNyZXQta2V5OiBzb21lX3NlY3JldF9rZXk=
```
yuzliu, line 41: Could you please add a code example? like echo -n 'account:key' | base64

python/kfserving/kfserving/storage.py
```diff
@@ -232,12 +232,23 @@ def _download_from_uri(uri, out_dir=None):
...
 
-        with requests.get(uri, stream=True) as response:
+        # Get header information from host url
+        headers = {}
+        host_uri = url.hostname
+        fields = os.getenv(host_uri, "").split(",")
+
+        for field in fields:
+            if field.startswith("header."):
+                headers[field[7:]] = os.getenv(field)
```
yuzliu, line 242: is 7 the length of "header" ? Could you please put it in a a constant and use the constant here? This will be a lot more readable.

python/kfserving/kfserving/storage.py
```diff
@@ -232,12 +232,23 @@ def _download_from_uri(uri, out_dir=None):
         if filename == '':
             raise ValueError('No filename contained in URI: %s' % (uri))
 
-        with requests.get(uri, stream=True) as response:
+        # Get header information from host url
+        headers = {}
+        host_uri = url.hostname
+        fields = os.getenv(host_uri, "").split(",")
```
yuzliu, line 238: Can you rename fields to something more meaningful?

pkg/credentials/https/https_secret.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"github.com/kubeflow/kfserving/pkg/constants"
+	v1 "k8s.io/api/core/v1"
+	"strings"
+)
+
+// Create constants -- baseURI
+const (
+	HTTPSHostURI = "https-base-uri"
+	HEADER       = "header"
```
ifilonenko, line 28: why is key "header" here but "headers" in the readme?

ifilonenko (changes requested):

pkg/credentials/https/https_secret.go
abchoo: Good catch! Thanks

pkg/credentials/https/https_secret.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/kubeflow/kfserving/pkg/constants"
+	v1 "k8s.io/api/core/v1"
+	"strings"
+)
+
+// Create constants -- baseURI
+const (
+	HTTPSHostURI = "https-base-uri"
```
jzhoucliqr, line 27: readme has 'https-host-uri', here is https-base-uri

pkg/credentials/https/https_secret.go
```diff
@@ -0,0 +1,78 @@
...
+	HeaderPrefix                 = HEADER + "."
+	CommaSeparator               = ","
+	ColonSeparator               = ": "
+)
+
+// Can be used for http and https uris
+func BuildSecretEnvs(secret *v1.Secret) []v1.EnvVar {
+	var fieldKeys []string
+	envs := []v1.EnvVar{}
+	hostURI, ok := secret.Annotations[InferenceServiceHTTPSHostURI]
```
jzhoucliqr, line 44: wdyt about using secret type VS annotation ? https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

abchoo: I think using secret's data type may be better in this case for consistency with the [other conditions](https://github.com/kubeflow/kfserving/blob/master/pkg/credentials/service_account_credentials.go#L98). Provides more confidentiality I suppose.

docs/samples/storage/uri/README.md
```diff
@@ -8,6 +8,39 @@ This `storageUri` option supports single file models, like `sklearn` which is sp
...
+If you do not require headers in your HTTP/HTTPS service request then you can skip this step.
+You can define headers by following the following format:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+  annotations:
+    serving.kubeflow.org/https-host-uri: “example.com”
```
yuzisun, line 21: I think this is just host part not the uri right?

docs/samples/storage/uri/README.md
```diff
@@ -8,6 +8,39 @@ This `storageUri` option supports single file models, like `sklearn` which is sp
...
+secrets:
+  - name: mysecret
+```
+Make sure you have serviceAccountName specified in your predictor in your inference service. These headers will be applied to any http/https requests that have the same host uri.
+
+You will need to base64 encode the headers. Make sure each header is on a newline with the format `header_key: header_value`.
+```text
+account-name: some_account_name
+secret-key: some_secret_key
+# echo -n 'account-name: some_account_name\nsecret-key: some_secret_key' | base64
```
yuzisun, line 40: Can we directly encode as json, so later you can easily decode back as a python dict? 
```json
{
   account-name: some_account_name,
   secret-key: some_secret_key
}
```

pkg/credentials/https/https_secret_test.go
```diff
@@ -0,0 +1,94 @@
...
+					Value: headerValue1,
+				},
+				{
+					Name:  HeaderPrefix + header2,
+					Value: headerValue2,
+				},
+				{
+					Name:  hostUri,
+					Value: HeaderPrefix + header1 + CommaSeparator + HeaderPrefix + header2,
+				},
```
yuzisun, line 82: I think it is cleaner to collapse into one env.

docs/samples/storage/uri/README.md
abchoo: I will try this out! Would probably make my implementation cleaner. Thanks.

python/kfserving/kfserving/storage.py
```diff
@@ -232,12 +234,24 @@ def _download_from_uri(uri, out_dir=None):
...
-        with requests.get(uri, stream=True) as response:
+        # Get header information from host url
+        headers = {}
+        host_uri = url.hostname
+
+        headers_json = os.getenv(host_uri + _HEADERS_SUFFIX, "{}")
+        headers_keys_values = json.loads(headers_json)
+
+        for key in headers_keys_values:
+            headers[key] = headers_keys_values[key]
```
yuzisun, line 245: This block seems redundant

abchoo: Oh.. you're right. Thanks!

PR #1375, opened 2021-02-11
Prevent experimental features from being set by default setting multimodelserver flags to false 

abchoo:
**What this PR does / why we need it**: By default the flags in inferenceservice.yaml for SKLearn, XGBoost, LightGMB, and Triton have been set to true which means MMS is enabled without users knowing. We want to prevent this from happening.

Fixes #

**Release note**:

```release-note
Using MMS requires flags to be set in inferenceservice.yaml
```

### DIFF
--- a/config/configmap/inferenceservice.yaml
+++ b/config/configmap/inferenceservice.yaml
@@ -31,7 +31,7 @@ data:
             "supportedFrameworks": [
               "sklearn"
             ],
-            "multiModelServer": true
+            "multiModelServer": false
           },
           "v2": {
             "image": "docker.io/seldonio/mlserver",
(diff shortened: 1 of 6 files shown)


yuzliu (approved):

yuzisun:
@abchoo what happens when user submits trained model when the flag is off?

yuzisun:
@abchoo Can you help regenerate the 0.5 release yaml by running `hack/generate-install.sh v0.5.0` ?

yuzliu:
> @abchoo Can you help regenerate the 0.5 release yaml by running `hack/generate-install.sh v0.5.0` ?

@yuzisun  Should we cut release after the validation webhook is in?

yuzisun:
Thanks @abchoo!

/lgtm
/approve

PR #1382, opened 2021-02-16
Use custom local gateway for isvc external service

### LINKED ISSUE #1383: Use custom local gateway in isvc external name
/kind feature

The kfserving isvc exposes an option to use a custom local gateway. However, the external service name still uses the hard coded cluster local gateway. It should be changed to use the custom one instead. It's unclear to me if there should be a fallback to the hardcoded one in case the configmap gets  messed up on something. If that's not necessary, it should be a simple fix.

adamkgray:

InferenceService expose an option to customise the istio local gateway that is used with them, however the reconviler still points to the hard coded local gateway specified in constants. This fixes that by using the local gateway from config instead.

Fixes #1383

**Release note**:

```release-note

```


yuzisun:
@adamkgray Good catch!

PR #1388, opened 2021-02-18
Re-add crd as base to default kustomization

pvaneck:
The CRD installation was separated from the default install to get around an issue with the CRD size being too large to store in the `metadata.annotations.last-applied-configuration` field of the resource (max 256kb).

Currently, with the removal of `initContainers` schema spec from the inferenceservice CRD, the size is now small enough to fit into the annotations (with ~15kb to spare). This commit re-adds the CRD yaml to the default yaml, returning it to how it was originally. 

Keeping it this structure simplifies the install process and helps with the planned `kubeflow/manifests` setup for KFServing (Ref: https://github.com/kubeflow/kfserving/issues/1376)

Issue about reducing CRD size: https://github.com/kubeflow/kfserving/issues/1296


Makefile
```diff
@@ -43,8 +43,8 @@ deploy: manifests
 	cd config/default && if [ ${KFSERVING_ENABLE_SELF_SIGNED_CA} != false ]; then \
 	kustomize edit remove resource certmanager/certificate.yaml; \
 	else kustomize edit add resource certmanager/certificate.yaml; fi;
-	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/default/crd | kubectl replace --validate=false -f - || \
-	kustomize build config/default/crd | kubectl create --validate=false -f -
+	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/crd | kubectl replace --validate=false -f - || \
```
Tomcli, line 46: if we are going to apply the crd below, there's no purpose for create/replace the crd over here.

Makefile
```diff
@@ -55,8 +55,8 @@ deploy-dev: manifests
 	kustomize edit remove resource certmanager/certificate.yaml; \
 	else kustomize edit add resource certmanager/certificate.yaml; fi;
 
-	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/default/crd | kubectl replace --validate=false -f - || \
-	kustomize build config/default/crd | kubectl create --validate=false -f -
+	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/crd | kubectl replace --validate=false -f - || \
```
Tomcli, line 58: same as above

Makefile
```diff
@@ -89,8 +89,8 @@ deploy-dev-storageInitializer: docker-push-storageInitializer
 	kustomize build config/overlays/dev-image-config | kubectl apply --validate=false -f -
 
 deploy-ci: manifests
-	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/default/crd | kubectl replace --validate=false -f - || \
-	kustomize build config/default/crd | kubectl create --validate=false -f -
+	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/crd | kubectl replace --validate=false -f - || \
```
Tomcli, line 92: same as above

Tomcli (commented):
Thanks @pvaneck, the kustomization looks good. just few minor comments on the makefile.

Makefile
pvaneck: Should be fine to just remove the added lines here and just keep the `kustomize build config/default | kubectl apply --validate=false -f -` since that  will now install the crd, right?

Tomcli: yes, this replace/create command is for replacing the crd without storing the crd spec in the annotations field. However, since we put the crd in `config/default`, we will be always applying the crd with annotations.

Tomcli:
thanks @pvaneck 
/lgtm

Makefile
```diff
@@ -43,8 +43,6 @@ deploy: manifests
 	cd config/default && if [ ${KFSERVING_ENABLE_SELF_SIGNED_CA} != false ]; then \
 	kustomize edit remove resource certmanager/certificate.yaml; \
 	else kustomize edit add resource certmanager/certificate.yaml; fi;
-	kubectl get crd inferenceservices.serving.kubeflow.org && kustomize build config/default/crd | kubectl replace --validate=false -f - || \
```
yuzisun, line 46: The problem currently with `kubectl apply` is that it takes long time(almost 10-15minutes) to install the crd because of the last applied annotation, `kubectl replace` is much faster way.

pvaneck: Ah, okay did not know this. I went ahead and put it back.

However, apply commands for the KFServing CRDS have always been pretty quick for me. Is the 10-15 min time just on the testing infra? I found this issue: https://github.com/kubernetes/kubernetes/issues/91615 which outlines what is possibly making it slow, so I downgraded to a kubectl version that didn't have the patch for this issue (v1.18.3) and sure enough, I found myself with a 12 min apply time versus what I had using version v1.18.8 (<5 second apply time). Seems like kubectl should be updated on the testing infra?

yuzisun: wow, good finding did not know that it is fixed in newer kubectl version, ye looks like I am using v1.18.3 as well.

pvaneck: Since https://github.com/kubeflow/kfserving/pull/1401 was merged, I went ahead and removed the create/replace lines

yuzisun:
Thanks @pvaneck !

/lgtm
/approve

PR #1401, opened 2021-02-23
Update kubectl version used in e2e tests

pvaneck:
Based off  https://github.com/kubeflow/kfserving/blob/master/test/scripts/create-cluster.sh#L31 , the test cluster uses Kubernetes 1.17.

This PR updates the `kubectl` version used to a matching version. This version also has the fix for the problem outlined in this issue: https://github.com/kubernetes/kubernetes/issues/91615 where re-applying a CRD with deeply nested properties took a long time. 

This PR also changes the README instructions to tell users to use a version of `kubectl` which has a fix for the above issue.


yuzisun:
Thanks @pvaneck! that's great finding!

/lgtm
/approve

PR #1402, opened 2021-02-23
Validate the parent inference service is ready in trained model controller

abchoo:
**What this PR does / why we need it**: Prevents the trained model from being deployed if the inference service is not ready

Fixes #

**Special notes for your reviewer**: I plan on adding a few other validation checks in controllerafter this PR

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/apis/serving/v1alpha1/trained_model_status.go
+++ b/pkg/apis/serving/v1alpha1/trained_model_status.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"k8s.io/api/core/v1"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
@@ -32,4 +33,49 @@ type TrainedModelStatus struct {
 	// Addressable endpoint for the deployed trained model
 	// http://<inferenceservice.metadata.name>/v1/models/<trainedmodel>.metadata.name
 	Address *duckv1.Addressable `json:"address,omitempty"`
+}
+
+// ConditionType represents a Service condition value
+// TODO: Add conditions for framework supported and memory availability
+const (
+	// InferenceServiceReady is set when inference service reported readiness
+	InferenceServiceReady apis.ConditionType = "InferenceServiceReady"
+)
+
+// TrainedModel Ready condition is depending on inference service readiness condition
+// TODO: Similar to above, add the constants here
+var conditionSet = apis.NewLivingConditionSet(
+	InferenceServiceReady,
+)
+
+var _ apis.ConditionsAccessor = (*TrainedModelStatus)(nil)
+
+func (ss *TrainedModelStatus) InitializeConditions() {
+	conditionSet.Manage(ss).InitializeConditions()
+}
+
+// IsReady returns if the service is ready to serve the requested configuration.
+func (ss *TrainedModelStatus) IsReady() bool {
+	return conditionSet.Manage(ss).IsHappy()
+}
+
+// GetCondition returns the condition by name.
+func (ss *TrainedModelStatus) GetCondition(t apis.ConditionType) *apis.Condition {
+	return conditionSet.Manage(ss).GetCondition(t)
+}
+
+// IsConditionReady returns the readiness for a given condition
+func (ss *TrainedModelStatus) IsConditionReady(t apis.ConditionType) bool {
+	return conditionSet.Manage(ss).GetCondition(t) != nil && conditionSet.Manage(ss).GetCondition(t).Status == v1.ConditionTrue
+}
+func (ss *TrainedModelStatus) SetCondition(conditionType apis.ConditionType, condition *apis.Condition) {
+	switch {
+	case condition == nil:
+	case condition.Status == v1.ConditionUnknown:
+		conditionSet.Manage(ss).MarkUnknown(conditionType, condition.Reason, condition.Message)
+	case condition.Status == v1.ConditionTrue:
+		conditionSet.Manage(ss).MarkTrue(conditionType)
+	case condition.Status == v1.ConditionFalse:
+		conditionSet.Manage(ss).MarkFalse(conditionType, condition.Reason, condition.Message)
+	}
 }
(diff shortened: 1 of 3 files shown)


pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -81,6 +80,18 @@ func (r *TrainedModelReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	if !isvc.Status.IsReady() {
+		log.Info("Parent InferenceService is not ready", "TrainedModel", tm.Name, "InferenceService", isvc.Name)
+		tm.Status.SetCondition(v1alpha1api.InferenceServiceReady, &apis.Condition{
+			Type:    v1alpha1api.InferenceServiceReady,
+			Status:  v1.ConditionFalse,
+			Reason:  "InferenceServiceNotReady",
+			Message: "Inference Service needs to be ready before Trained Model can be ready",
+		})
+	}
+
```
yuzliu, line 94: Need to do update status here

pkg/apis/serving/v1alpha1/trained_model_status.go
```diff
@@ -33,3 +34,48 @@ type TrainedModelStatus struct {
...
+
+// GetCondition returns the condition by name.
+func (ss *TrainedModelStatus) GetCondition(t apis.ConditionType) *apis.Condition {
+	return conditionSet.Manage(ss).GetCondition(t)
+}
+
+// IsConditionReady returns the readiness for a given condition
+func (ss *TrainedModelStatus) IsConditionReady(t apis.ConditionType) bool {
+	return conditionSet.Manage(ss).GetCondition(t) != nil && conditionSet.Manage(ss).GetCondition(t).Status == v1.ConditionTrue
+}
```
yuzisun, line 70: newline

pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -190,6 +199,45 @@ func (r *TrainedModelReconciler) updateStatus(req ctrl.Request, desiredModel *v1
...
+	} else {
+		log.Info("Parent InferenceService is not ready", "TrainedModel", tm.Name, "InferenceService", isvc.Name)
+		tm.Status.SetCondition(v1alpha1api.InferenceServiceReady, &apis.Condition{
+			Type:    v1alpha1api.InferenceServiceReady,
+			Status:  v1.ConditionFalse,
+			Reason:  "InferenceServiceNotReady",
+			Message: "Inference Service needs to be ready before Trained Model can be ready",
+		})
+
+		conditionErr = utils.FirstNonNilError([]error{
```
yuzisun, line 225: why can't directly assign 
`conditionErr := fmt.Errorf(InferenceServiceNotReady, isvc.Name, tm.Name)` ?

pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -190,6 +199,45 @@ func (r *TrainedModelReconciler) updateStatus(req ctrl.Request, desiredModel *v1
...
+			conditionErr,
+			fmt.Errorf(InferenceServiceNotReady, isvc.Name, tm.Name),
+		})
+	}
+
+	if conditionErr != nil {
```
yuzisun, line 231: should be updated regardless right?

pkg/controller/v1alpha1/trainedmodel/controller_test.go
```diff
@@ -370,6 +501,18 @@ var _ = Describe("v1beta1 TrainedModel controller", func() {
...
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, serviceKey, inferenceService)
+				if err != nil {
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
+
+			inferenceService.Status.Status = statusConditions
+			Expect(k8sClient.Status().Update(context.TODO(), inferenceService)).To(BeNil())
```
yuzisun, line 514: Add test to check the trained model status condition?

pkg/controller/v1alpha1/trainedmodel/controller.go
abchoo: I wanted to make it FirstNonNilError because I could easily insert another condition check above it without having to worry about modifying that part. The reason why i declared it as a var is because `conditionErr` could potentially be used by other condition check.

pkg/controller/v1alpha1/trainedmodel/controller.go
abchoo: I suppose. I was thinking we could update it when the other statuses are being updated but this also works.

pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -190,6 +199,45 @@ func (r *TrainedModelReconciler) updateStatus(req ctrl.Request, desiredModel *v1
...
+		if statusErr := r.Status().Update(context.TODO(), tm); statusErr != nil {
+			r.Recorder.Eventf(tm, v1.EventTypeWarning, "UpdateFailed",
+				"Failed to update status for TrainedModel %q: %v", tm.Name, conditionErr)
```
jzhoucliqr, line 234: should this event contain conditionErr or statusErr?  Also the event is already on tm, so adding tm.Name in the message seems redundant.

pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -190,6 +199,45 @@ func (r *TrainedModelReconciler) updateStatus(req ctrl.Request, desiredModel *v1
...
+		}
+	}
+
+	return conditionErr
```
jzhoucliqr, line 238: feels a bit not intuitive to treat condition not ready the same way as a generic Get/Update error, because the caller just ignores the error and return without requeue. ideally for Get/Update error should return err and requeue immediately.

pkg/controller/v1alpha1/trainedmodel/controller.go
abchoo: It should contain statusErr -- my mistake. Thanks for point that out! I've removed tm.Name as you suggested.

pkg/controller/v1alpha1/trainedmodel/controller.go
abchoo: That's another mistake I made. It shouldn't have ignored the returned error. Changed it to do so. Just to clarify, returning an error will make the it requeue?

pkg/controller/v1alpha1/trainedmodel/controller_test.go
abchoo: Added it here: https://github.com/kubeflow/kfserving/pull/1402/files#diff-91393e839bab87f65a6a1b532750bb1b0753ca29cf828612b9d8ac8cdf2ba7c1R428

pkg/controller/v1alpha1/trainedmodel/controller.go
jzhoucliqr: yes, there will be requeue with exponential backoff when reconciler return error.

PR #1407, opened 2021-02-24
Validation for storage uri in webhook

abchoo:
**What this PR does / why we need it**: The webbook will validate the storageURI passed in the trained model

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/apis/serving/v1alpha1/trainedmodel_webhook.go
+++ b/pkg/apis/serving/v1alpha1/trainedmodel_webhook.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"fmt"
 	"github.com/kubeflow/kfserving/pkg/agent/storage"
+	"github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1"
 	"github.com/kubeflow/kfserving/pkg/utils"
 	"k8s.io/apimachinery/pkg/runtime"
 	"regexp"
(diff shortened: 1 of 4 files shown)


pkg/apis/serving/v1alpha1/trainedmodel_webhook.go
```diff
@@ -27,16 +27,20 @@ import (
 
 // regular expressions for validation of isvc name
 const (
-	TmNameFmt                   string = "[a-z]([-a-z0-9]*[a-z0-9])?"
-	InvalidTmNameFormatError           = "the Trained Model \"%s\" is invalid: a Trained Model name must consist of lower case alphanumeric characters or '-', and must start with alphabetical character. (e.g. \"my-name\" or \"abc-123\", regex used for validation is '%s')"
-	InvalidTmMemoryModification        = "the Trained Model \"%s\" memory field is immutable. The memory was \"%s\" but it is updated to \"%s\""
+	TmNameFmt                    string = "[a-z]([-a-z0-9]*[a-z0-9])?"
+	StorageUriFmt                string = "(gs|s3)://(([a-z]([-a-z0-9]*[a-z0-9])?/))*([a-z]([-a-z0-9]*[a-z0-9]))"
```
yuzliu, line 31: We probably should check against https://github.com/kubeflow/kfserving/blob/master/pkg/agent/storage/provider.go#L26 . So next time when we add a new provider we don't need to update both places.

pkg/apis/serving/v1alpha1/trainedmodel_webhook.go
```diff
@@ -18,25 +18,37 @@ package v1alpha1
...
 var (
 	// log is for logging in this package.
 	tmLogger = logf.Log.WithName("trainedmodel-alpha1-validator")
-	// regular expressions for validation of isvc name
+	// regular expressions for validation of tm name
 	TmRegexp = regexp.MustCompile("^" + TmNameFmt + "$")
+	// protocols that are accepted by storage uri
+	StorageUriProtocols = strings.Join(storage.GetAllProtocol(), CommaSpaceSeparator)
+	// prefix allowed in storage uri
+	StorageUriPrefixes = "(" + strings.Join(storage.GetAllProtocol(), "|") + ")"
```
yuzliu, line 49: Nice change!

yuzliu (commented):
/LGTM

pkg/apis/serving/v1alpha1/trainedmodel_webhook.go
```diff
@@ -18,25 +18,37 @@ package v1alpha1
...
 
 // regular expressions for validation of isvc name
 const (
-	TmNameFmt                   string = "[a-z]([-a-z0-9]*[a-z0-9])?"
-	InvalidTmNameFormatError           = "the Trained Model \"%s\" is invalid: a Trained Model name must consist of lower case alphanumeric characters or '-', and must start with alphabetical character. (e.g. \"my-name\" or \"abc-123\", regex used for validation is '%s')"
-	InvalidTmMemoryModification        = "the Trained Model \"%s\" memory field is immutable. The memory was \"%s\" but it is updated to \"%s\""
+	CommaSpaceSeparator                 = ", "
+	AlphaNumericDash                    = "[a-z]([-a-z0-9]*[a-z0-9])"
+	TmNameFmt                    string = AlphaNumericDash + "?"
+	StorageUriFmt                string = "((" + AlphaNumericDash + "?)/)*(" + AlphaNumericDash + "?)"
```
yuzisun, line 35: Can we share the same validation logic as inference service ? I know prefix set might be different but the url validation logic should be same?

abchoo: The inference service only really checks for the prefix so I could extract that logic out and use it for both validators. The validateStorageURI doesn't check for content and I suppose that is because some protocols can contain different characters. I took a look at [GCS Object name requirements](https://cloud.google.com/storage/docs/naming-objects#objectnames) and it seems that vastly differs to [S3 Object key naming](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html). So now I am considering not validating the non-prefix part. What do you think?

PR #1412, opened 2021-02-28
reorg multi-model serving doc

yuzliu:
**What this PR does / why we need it**: Reorg multi-model serving docs into one folder


abchoo:
Could you please help fix this markdown issue https://github.com/yuzliu/kfserving/blob/mmsdoc/docs/samples/multimodelserving/sklearn/README.md?

Adding a space should fix it #Multi-Model Serving with Sklearn ->
# Multi-Model Serving with Sklearn

PR #1414, opened 2021-02-28
Check if isvc.Status.Default is not nil before trying to access pointer

abchoo:
**What this PR does / why we need it**: We currently do not check if default is non-nil and try to dereference the pointer. This could potentially lead to invalid memory address or nil pointer dereference errors.

Fixes #

**Release note**:

```release-note
None
```


yuzisun:
good catch!
/approve

PR #1415, opened 2021-02-28
WIP: Support collocate transformer and predictor in the same pod

yuzisun:
While deploying transformer as separate pod gives nice property to scale differently and run on different device, in some cases you might want to deploy them together if they are coupled and need to collocate in the same pod.

Fixes #

**Release note**:

```release-note

```


docs/samples/v1beta1/transformer/torchserve_image_transformer/transformer.Dockerfile
```diff
@@ -5,6 +5,6 @@ RUN apt-get update \
 
 COPY . .
 RUN pip install --upgrade pip
-RUN pip install kfserving==0.5.0rc0
+RUN pip install kfserving==0.5.0
```
abchoo, line 8: Should this be 0.5.1?

abchoo, line 8: Don't forget to update this! :)

pkg/apis/serving/v1beta1/predictor_custom.go
```diff
@@ -65,8 +65,12 @@ func (c *CustomPredictor) Default(config *InferenceServicesConfig) {
 	if len(c.Containers) == 0 {
 		c.Containers = append(c.Containers, v1.Container{})
 	}
-	c.Containers[0].Name = constants.InferenceServiceContainerName
-	setResourceRequirementDefaults(&c.Containers[0].Resources)
+	for i, _ := range c.Containers {
+		if c.Containers[i].Name == "" {
+			c.Containers[i].Name = constants.InferenceServiceContainerName
```
Iamlovingit, line 70: If Containers are not be set the name, then they will be set the same container name. I suggest adding the index of `c.Containers` behind the `constants.InferenceServiceContainerName`

PR #1417, opened 2021-03-01
Logging in IsMMSPredictor

abchoo:
**What this PR does / why we need it**: Adds a log that notifies users if the inference service is using multi-model serving or single model serving.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -31,9 +31,10 @@ func IsMMSPredictor(isvc *v1beta1api.InferenceService, isvcConfig *v1beta1api.In
 	isMMS := predictor.GetImplementation().IsMMS(isvcConfig) && predictor.GetImplementation().GetStorageUri() == nil
 
 	if isMMS {
-		log.Info("Predictor is configured for multi-model serving", "InferenceService", isvc.Name)
+		log.V(5).Info("Predictor is configured for multi-model serving", "InferenceService", isvc.Name)
 	} else {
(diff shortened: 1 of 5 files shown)


pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -18,9 +18,22 @@ package utils
...
 
 // Only enable MMS predictor when predictor config sets MMS to true and storage uri is not set
-func IsMMSPredictor(predictor *v1beta1api.PredictorSpec, isvcConfig *v1beta1api.InferenceServicesConfig) bool {
-	return predictor.GetImplementation().IsMMS(isvcConfig) && predictor.GetImplementation().GetStorageUri() == nil
+func IsMMSPredictor(isvc *v1beta1api.InferenceService, isvcConfig *v1beta1api.InferenceServicesConfig) bool {
+	predictor := isvc.Spec.Predictor
+	isMMS := predictor.GetImplementation().IsMMS(isvcConfig) && predictor.GetImplementation().GetStorageUri() == nil
+
+	if isMMS {
+		log.Info("Predictor is configured for multi-model serving", "InferenceService", isvc.Name)
```
yuzisun, line 34: this might be a bit excessive for info level, do we need this ?

abchoo: I'm debating on whether or not it's better to have: `log.V(5).Info("Determining if predictor is configured for multi-model serving", "InferenceService", isvc.Name, "IsMMS", isMMS)` rather than the condition statements. What do you think?

PR #1424, opened 2021-03-04
Update kubeflow overlay

pvaneck:
Additional resources are created when KFServing is installed as a part of Kubeflow in the manifests repository. This commits updates the kubeflow overlay to make it more akin to what is installed with the [manifests repo.](https://github.com/kubeflow/manifests/tree/master/apps/kfserving/upstream/installs/generic)

A user can generate a YAML to install KFServing into a Kubeflow environment using the following command:

```
kustomize build config/overlays/kubeflow 
```

Ref: https://github.com/kubeflow/kfserving/issues/1376


config/overlays/kubeflow/application.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - group: rbac.authorization.k8s.io/v1
+      kind: RoleBinding
+  descriptor:
+    type: kfserving
+    version: v1beta1
+    description: |-
+      KFServing provides a Kubernetes Custom Resource Definition for serving ML Models on arbitrary frameworks
+    maintainers:
+      - name: Johnu George
+        email: johnugeo@cisco.com
```
yuzisun, line 30: update the names here?

pvaneck: Yea, I'll just do what pipelines did [here](https://github.com/kubeflow/manifests/blob/master/apps/pipeline/upstream/installs/generic/application.yaml#L19).

config/overlays/kubeflow/namespace-patch.yaml
```diff
@@ -0,0 +1,8 @@
+# This keeps the kubeflow namespace labels how they were initially.
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: kubeflow
+    istio-injection: enabled
```
Tomcli, line 7: quick question, will the `istio-injection: enabled` apply to single user kubeflow when it runs on Kubeflow namespace? Or kubeflow/manifests has another overlay to overwrite this for single user setup?

pvaneck: The intent behind this patch is to return to what the  `kubeflow` namespace is deployed as [here](https://github.com/kubeflow/manifests/blob/master/common/namespaces/base/namespaces.yaml). However, I just looked more into kustomize, and apparently it's possible to just remove the namespace resource from the generated YAML with a `$patch: delete`. This  way the kubeflow namespace won't be touched and is assumed to already exist.

Tomcli: Yeah that namespace.yaml in the manifests is only used in the multi-user mode.

yanniszark:
@pvaneck thanks, this is fantastic! :)
Could you also add the `commonLabels` to the `kubeflow` overlay?
```yaml
commonLabels:
  app: kfserving
  kustomize.component: kfserving
  app.kubernetes.io/component: kfserving
  app.kubernetes.io/name: kfserving
```

I believe it's the only thing missing from the old `installs/generic` overlay.

pvaneck:
@yanniszark Good catch. Added the common labels.

pvaneck:
@Tomcli @yuzisun I believe this should be good to go.

Tomcli:
thanks @pvaneck 
/lgtm

animeshsingh:
/lgtm
/approve

cc @yuzisun

PR #1446, opened 2021-03-11
Validate trained model memory based on parent inference service

abchoo:
**What this PR does / why we need it**: Checks to see if the inference service has enough memory resources for all children trained models. Will reject child that asks for more resources than parent can provide.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -45,6 +45,11 @@ var (
 	InferenceServiceConfigMapName = "inferenceservice-config"
 )
 
+// TrainedModel Constants
+var (
+	TrainedModelAllocated = KFServingAPIGroupName + "/" + "trainedmodel-allocated"
+)
+
 // InferenceService MultiModel Constants
 var (
 	ModelConfigFileName = "models.json"
--- a/pkg/controller/v1alpha1/trainedmodel/controller.go
+++ b/pkg/controller/v1alpha1/trainedmodel/controller.go
@@ -49,8 +49,9 @@ import (
 )
 
 const (
-	InferenceServiceNotReady = "Inference Service \"%s\" is not ready. Trained Model \"%s\" cannot deploy"
-	FrameworkNotSupported    = "Inference Service \"%s\" does not support the Trained Model \"%s\" framework \"%s\""
+	InferenceServiceNotReady   = "Inference Service \"%s\" is not ready. Trained Model \"%s\" cannot deploy"
+	FrameworkNotSupported      = "Inference Service \"%s\" does not support the Trained Model \"%s\" framework \"%s\""
+	MemoryResourceNotAvailable = "Inference Service \"%s\" memory resources are not available. Trained Model \"%s\" cannot deploy"
 )
 
 var log = logf.Log.WithName("TrainedModel controller")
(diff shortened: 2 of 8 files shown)


pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -251,6 +252,52 @@ func (r *TrainedModelReconciler) updateConditions(req ctrl.Request, tm *v1alpha1
 		}
 	}
 
+	// Get trained models with same inference service
+	var trainedModels v1alpha1api.TrainedModelList
+	if err := r.List(context.TODO(), &trainedModels, client.MatchingLabels{constants.ParentInferenceServiceLabel: isvc.Name, constants.TrainedModelMemoryIncluded: isvc.Name}); err != nil {
```
yuzisun, line 257: Need to include the namespace in the ListOption, otherwise it may find the trained model in other namespaces with same inference service name.

pkg/controller/v1alpha1/trainedmodel/controller_test.go
```diff
@@ -671,4 +676,128 @@ var _ = Describe("v1beta1 TrainedModel controller", func() {
 			}, timeout, interval).Should(Equal(expected.Data))
 		})
 	})
+
+	Context("When creating a new TrainedModel that requeires more memory than available", func() {
```
yuzisun, line 680:
```suggestion
	Context("When creating a new TrainedModel that requires more memory than available", func() {
```

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -18,9 +18,27 @@ package utils
...
+func IsMemoryResourceAvailable(isvc *v1beta1api.InferenceService, totalReqMemory resource.Quantity, isvcConfig *v1beta1api.InferenceServicesConfig) bool {
+	if isvc.Spec.Predictor.GetExtensions() == nil || len(isvc.Spec.Predictor.GetImplementations()) == 0 {
+		return false
+	}
+
+	container := isvc.Spec.Predictor.GetImplementation().GetContainer(isvc.ObjectMeta, isvc.Spec.Predictor.GetExtensions(), isvcConfig)
+
+	if constants.InferenceServiceContainerName == container.Name {
+		predictorMemoryLimit := container.Resources.Limits.Memory()
+		return predictorMemoryLimit.Cmp(totalReqMemory) >= 0
```
yuzisun, line 40: wondering if we should leave some room here as the model server process itself consumes certain amount of memory.

pkg/constants/constants.go
```diff
@@ -44,6 +44,11 @@ var (
 	InferenceServiceConfigMapName = "inferenceservice-config"
 )
 
+// TrainedModel Constants
+var (
+	TrainedModelMemoryIncluded = "trainedmodel-memory-included"
```
yuzisun, line 49: I think the name is too long, remember that the length limit for label is only 63 chars
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

yuzisun:
@abchoo Can you help squash your commits?

yuzisun:
@googlebot I consent.

abchoo:
@googlebot I consent.

yuzisun:
@abchoo  Looks like you did not rebase from master correctly it is missing [this configuration](https://github.com/abchoo/kfserving/blob/master/test/scripts/run-e2e-tests.sh#L130) in your branch.

PR #1449, opened 2021-03-11
Fix broken links to SKLearn Server README

kevinbazira:
SKLearn Server links point to `page not found`. Fixed links to point to the correct `README` file. Easy navigation and easy reading for others.


yuzisun:
thanks @kevinbazira !

/lgtm
/approve

PR #1454, opened 2021-03-13
fix broken image-net downloader

### LINKED ISSUE #1453: Alibiexplainer python test requesting url not found
/kind bug

Run `pytest alibiexplainer/tests/test_anchor_images.py`

Test to pass

I believe that the explainer's dataset location has been changed. My PR fails when the python test tries to curl http://www.image-net.org/api/text/imagenet.synset.geturls?wnid=n02123394.

My PR for reference:
https://github.com/kubeflow/kfserving/pull/1452/checks?check_run_id=2098354600#step:5:1064

I am using the latest version of KFServing.

theofpa:
This fixes the alibiexplainer test which fails because it downloads the images from the image-net API [ref](http://www.image-net.org/update-mar-11-2021.php) which is no longer available.

Fixes #1453


yuzisun:
@theofpa Thanks so much for the quick fix!

/lgtm
/approve

PR #1469, opened 2021-03-18
Add mms support for custom spec

### LINKED ISSUE #1427: multi-model-serving support for custom container in predictorSpec
/kind feature

multimodel feature support currently depends on "multiModelServer" : true in config/configmap/inferenceservice.yaml
this feature works for standard supported predictors such as sklearn, triton and NOT for a custom container-based preditor spec. 

- ASsumption is that custom container is a model server exposing predict endpoint along with load, unload endpoints (defined in v2 protocol). If not, then agent will not be able to call load/unload endpoints. a mutlimodel server like AWS MMS can load models by polling /mnt/models/ directory without the need for load/unload endpoints.

Proposed solution:
- can we add new field "multiModelServer" under CoponentExtensionSpec struct (pkg/apis/serving/v1beta1/component.go)
this field in predictorSpec will also be useful to override default config behaviour

- further (c *CustomPredictor) IsMMS(config *InferenceServicesConfig) under pkg/apis/serving/v1beta1/predictor_custom.go  can be adjusted to use the multiModelServer flag to return if inferenceService is multi-model-server

- if MMS, then agent will be injected as sidecar along the custom container, agent watches for configMap changes and downloads models to mnt/models shared directory and call /load and /unload endpoints of model server

shydefoo:
Enables Custom predictors to make use of MMS feature

Fixes #1427 

* Introduced a new field in `ComponentExtensionSpec` as discussed in #1427 
* Modify `IsMMSPredictor()` to check if field is set 
* Eg inference-service spec:
```yaml
kind: "InferenceService"
metadata:
  name: "custom-sklearn"
spec:
  predictor:
    minReplicas: 1
    containers:
    - image: gcr.io/kfserving/sklearnserver
      name: kfserving-container
      ports:
        - containerPort: 8080
          protocol: TCP
      args:
      - --model_dir=/mnt/models
      - --http_port=8080
    multiModelServer: true
---
apiVersion: "serving.kubeflow.org/v1alpha1"
kind: "TrainedModel"
metadata:
  name: "model1-sklearn"
spec:
  inferenceService: "custom-sklearn"
  model:
    storageUri: "gs://kfserving-samples/models/sklearn/iris"
    framework: "sklearn"
    memory: "256Mi"
```

**Release note**:

```release-note
A new field 'MultiModelServer' has been added in ComponentExtensionSpec. When set 'true', it indicates that the Custom container supports multi model serving, or that the 'multiModelServer' field of the predictor in the inferenceservice-config configmap (eg. Sklearn, Triton) should be overridden.
```

### DIFF
--- a/pkg/apis/serving/v1beta1/component.go
+++ b/pkg/apis/serving/v1beta1/component.go
@@ -89,6 +89,10 @@ type ComponentExtensionSpec struct {
 	// Activate request batching and batching configurations
 	// +optional
 	Batcher *Batcher `json:"batcher,omitempty"`
+	// MultiModelServer can be used to override default mms behaviour specified in predictor config
+	// or to indicate Custom runtimes support MMS
+	// +optional
+	MultiModelServer *bool `json:"multiModelServer,omitempty"`
 }
 
 // Default the ComponentExtensionSpec
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -24,9 +24,9 @@ import (
 // is set or predictor config sets MMS to true and storage uri is not set
 func IsMMSPredictor(predictor *v1beta1api.PredictorSpec, isvcConfig *v1beta1api.InferenceServicesConfig) bool {
 	mmsExtension := predictor.GetExtensions().MultiModelServer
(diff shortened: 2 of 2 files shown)


shydefoo:
Drafting a PR to see if this approach is viable first

PR #1471, opened 2021-03-18
Update pipelines sample code

pvaneck:
The pipelines KFServing component was recently updated for the v0.5 release and the v1beta1 API, so here, the samples were updated accordingly.


docs/samples/pipelines/sample-tf-pipeline.py
```diff
@@ -23,32 +25,21 @@
...
+                             framework=framework)
+
 
 if __name__ == '__main__':
-    import kfp.compiler as compiler
-    compiler.Compiler().compile(kfservingPipeline, __file__ + '.tar.gz')
\ No newline at end of file
+    compiler.Compiler().compile(kfservingPipeline, __file__ + '.tar.gz')
+    client = Client()
+    experiment = client.create_experiment(name='My Experiment', namespace='anonymous')
```
Tomcli, line 44: we probably don't need to create the run as part of this example because users might want to run it in different namespace, experiment etc on the ui.

pvaneck: Sure, I will revert it to just having the compile function.

Tomcli:
thanks @pvaneck 
/lgtm

Issue #1320, opened 2021-01-27
Kubernetes Native Deployment on KFServing


Iamlovingit:
/kind feature

**thanks @yuzisun & @cliveseldon** 

# Kubernetes Native Deployment on KFServing
## Problem:
KFServing totally depends on Knative to bring serving features like scale to zero, Knative Pod Autoscaler(KPA) and canary rollouts to ML/DL deployments, and the serving runtime contract requires HTTP as the inbound transport.
There are use cases where Knative would be a great solution but where HTTP is not used.  Especially, due to safety concerns, some cases need alternative transports like UDP or TCP or even custom protocols. So, we need an effective scenario to solve the problem, make Knative Serving and native serving should be both available for the Kubernetes cluster.
## KNative Limitiations
* Doesn’t allow TCP/UDP
* Doesn’t allow multiple volumes mounted
  * User might want configmap loaded as volume etc

## Challenges
* Ingress
  * We can use K8S Ingress. Plus native Deployments and SVCs
* Canarying
  * Difficult to keep same workflow as Knative functionality uses Revisions
* Autoscaling
  * Can use KEDA, but can’t scale from 0
* Readiness Checks
* Image immutability (connected with canarying/revisions)
* Make sure logger and batcher can be used as now
## Targets:
support all features of raw kubernetes container
support all frameworks of KFServing
support multi-arm-bandits
support KPA, similar to Knative

For more detail, read [Kubernetes Native Deployment on KFServing](https://docs.google.com/document/d/1GZrfUSqYz5zbQGY26VFYZq_lQqRlkDdxsOTrJ6TQdi8/edit?usp=sharing)

PR #1508, opened 2021-03-27
1320 add raw demployment ability in kfserving

Iamlovingit:
relative issue: #1320 

# Kubernetes Native Deployment on KFServing
## Problem:
KFServing totally depends on Knative to bring serving features like scale to zero, Knative Pod Autoscaler(KPA) and canary rollouts to ML/DL deployments, and the serving runtime contract requires HTTP as the inbound transport.
There are use cases where Knative would be a great solution but where HTTP is not used.  Especially, due to safety concerns, some cases need alternative transports like UDP or TCP or even custom protocols. So, we need an effective scenario to solve the problem, make Knative Serving and native serving should be both available for the Kubernetes cluster.
## KNative Limitiations
* Doesn’t allow TCP/UDP
* Doesn’t allow multiple volumes mounted
  * User might want configmap loaded as volume etc

## Challenges
* Ingress
  * We can use K8S Ingress. Plus native Deployments and SVCs
* Canarying
  * Difficult to keep same workflow as Knative functionality uses Revisions
* Autoscaling
  * Can use KEDA, but can’t scale from 0
* Readiness Checks
* Image immutability (connected with canarying/revisions)
* Make sure logger and batcher can be used as now
## Targets:
support all features of raw kubernetes container
support all frameworks of KFServing
support multi-arm-bandits
support KPA, similar to Knative

For more detail, read [Kubernetes Native Deployment on KFServing](https://docs.google.com/document/d/1GZrfUSqYz5zbQGY26VFYZq_lQqRlkDdxsOTrJ6TQdi8/edit?usp=sharing)


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -79,7 +79,8 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 			constants.InferenceServicePodLabelKey: isvc.Name,
 			constants.KServiceComponentLabel:      string(v1beta1.PredictorComponent),
 		}),
-		Annotations: annotations,
+		Annotations:     annotations,
+		ResourceVersion: "0",
```
yuzisun, line 83: I think this should be set by kubernetes.

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+
+import (
+	"context"
+
+	"github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1"
+	v1beta1api "github.com/kubeflow/kfserving/pkg/apis/serving/v1beta1"
+	"github.com/kubeflow/kfserving/pkg/constants"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	exv1beta1 "k8s.io/api/extensions/v1beta1"
```
yuzisun, line 24: this package is going to be deprecated check https://github.com/kubernetes/kubernetes/pull/74057/files

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
```diff
@@ -0,0 +1,387 @@
+/*
+Copyright 2020 kubeflow.org.
```
yuzisun, line 2:
```suggestion
Copyright 2021 kubeflow.org.
```

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+	case checkResult == CheckResultUnknown:
+		return nil, err
+	}
+	return deployment, nil
+}
+
+//checkNativeDeploymentExist checks is the deployment is existed?
+func (r *NativeReconciler) checkNativeDeploymentExist(client *kubernetes.Clientset) (CheckResultType, error) {
+	//get deployment
+	existingDeployment, err := client.AppsV1().Deployments(r.Deployment.Namespace).Get(context.TODO(), r.Deployment.Name, metav1.GetOptions{})
```
yuzisun, line 272: Can we use the controller runtime [client API](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go)? want to be consistent with other places.

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+	case checkResult == CheckResultExisted:
+		log.Info("Reconcile", "Service", "Existed. Will do nothing.")
+	case checkResult == CheckResultUpdate:
+		_, err = client.CoreV1().Services(r.Service.Namespace).Update(context.TODO(), r.Service, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, err
+		}
+	case checkResult == CheckResultUnknown:
+		return nil, err
+	}
```
yuzisun, line 246: I think we can simplify this block and avoid switch cases, we only need create and update, if there is an error just return back.

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+		_ = r.Ingress
+	case checkResult == CheckResultUpdate:
+		_, err = client.ExtensionsV1beta1().Ingresses(r.Ingress.Namespace).Update(context.TODO(), r.Ingress, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, err
+		}
+	case checkResult == CheckResultUnknown:
+		return nil, err
+	}
+	return deployment, nil
```
yuzisun, line 266: please help split ingress, service, deployment reconcile into a separate function

pkg/controller/v1beta1/inferenceservice/reconcilers/kubernetes/kube_reconciler.go
Iamlovingit: ok, get it

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Iamlovingit: ye, I forgot this.

pradithya:
I am looking forward to see this feature in KFserving!
Also, I am wondering if transformer will be in scope for this PR?

pkg/constants/constants.go
```diff
@@ -220,6 +221,31 @@ var (
 	}
 )
 
+var (
+	//RawIngressNamePrefix ...
+	RawIngressNamePrefix = KFServingName + "-ingress"
```
yuzisun, line 226: Do we need the prefix? dns has length limit for 63

pkg/constants/constants.go
```diff
@@ -220,6 +221,31 @@ var (
...
+)
+
+// GetRawServiceLabel generate native service label
+func GetRawServiceLabel(service string, namespace string) string {
+	return "isvc." + namespace + "." + service
+}
+
+// GetRawServiceIngressName generate native service ingress name
+func GetRawServiceIngressName(service string, namespace string) string {
+	return RawIngressNamePrefix + "-" + namespace + "-" + service
```
yuzisun, line 246: I think we probably need a `domainTemplate` like in knative, the convention is `service.namespace` or `server-namespace`.

pkg/constants/constants.go
```diff
@@ -220,6 +221,31 @@ var (
...
+const (
+	CheckResultCreate  CheckResultType = 0
+	CheckResultUpdate  CheckResultType = 1
+	CheckResultExisted CheckResultType = 2
+	CheckResultUnknown CheckResultType = 3
+)
+
+// GetRawServiceLabel generate native service label
+func GetRawServiceLabel(service string, namespace string) string {
+	return "isvc." + namespace + "." + service
```
yuzisun, line 241: I think we do not need the prefix and namespace, the services are namespace scoped.

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+		Namespace: r.Deployment.Namespace,
+		Name:      r.Deployment.Name,
+	}, existingDeployment)
+	if err != nil {
+		if apierr.IsNotFound(err) {
+			return constants.CheckResultCreate, nil
+		}
+		return constants.CheckResultUnknown, err
+	}
+	//existed, check equalment
```
yuzisun, line 111: nit: typo

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	return constants.CheckResultUpdate, nil
+}
+
+func semanticDeploymentEquals(desired, existing *appsv1.Deployment) bool {
+	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec)
+}
+
+func setDefaultPodSpec(podSpec *corev1.PodSpec) {
+	if podSpec.DNSPolicy == "" {
+		podSpec.DNSPolicy = corev1.DNSClusterFirst
```
yuzisun, line 124: I think these defaults are set by kubernetes?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	if podSpec.RestartPolicy == "" {
+		podSpec.RestartPolicy = corev1.RestartPolicyAlways
+	}
+	if podSpec.TerminationGracePeriodSeconds == nil {
+		TerminationGracePeriodSeconds := int64(corev1.DefaultTerminationGracePeriodSeconds)
```
yuzisun, line 130: Knative defaults this to timeout value as you want to wait for all the in flight requests to finish before terminating the pod.

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	if podSpec.SchedulerName == "" {
+		podSpec.SchedulerName = corev1.DefaultSchedulerName
+	}
+	for i, _ := range podSpec.Containers {
+		container := &podSpec.Containers[i]
+		if container.TerminationMessagePath == "" {
+			container.TerminationMessagePath = "/dev/termination-log"
+		}
+		if container.TerminationMessagePolicy == "" {
+			container.TerminationMessagePolicy = corev1.TerminationMessageReadFile
```
yuzisun, line 145: Can we default to `FallbackToLogsOnError` to the error gets propagated back to inference service status

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+//GenerateIngressHost return the config domain in configmap.IngressDomain
+func GenerateIngressHost(client client.Client, metadata metav1.ObjectMeta) string {
+	ingressConfig, err := v1beta1api.NewIngressConfig(client)
+	if err != nil {
+		log.Error(err, "GenerateIngressHost")
+		if ingressConfig == nil {
+			ingressConfig = &v1beta1api.IngressConfig{}
+		}
+		if ingressConfig.IngressDomain == "" {
+			ingressConfig.IngressDomain = "1.2.3.4.xip.io"
```
yuzisun, line 65: get the domain from config

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+		Spec: netv1.IngressSpec{
+			Rules: []netv1.IngressRule{
+				{
+					Host: ingressHost,
+					IngressRuleValue: netv1.IngressRuleValue{
+						HTTP: &netv1.HTTPIngressRuleValue{
+							Paths: []netv1.HTTPIngressPath{
+								{
+									Path:     "/" + string(constants.Predict),
```
yuzisun, line 90: we have other paths so do not need to limit to predict endpoint?

yuzisun:
Let's add a simple sklearn e2e test first in [here](https://github.com/kubeflow/kfserving/blob/master/test/e2e/predictor/test_sklearn.py) and then we can add others in a separate PR, I think the spec should be mostly the same with the additional annotation for raw deployment right?

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,138 @@
...
+		ObjectMeta: componentMeta,
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": constants.GetRawServiceLabel(componentMeta.Name, componentMeta.Namespace),
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name: componentMeta.Name,
+					Port: podSpec.Containers[0].Ports[0].ContainerPort,
+					TargetPort: intstr.IntOrString{
```
yuzisun, line 72: Add a doc and comment that user knows we always point service to the first container.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+		Ingress:         createRawIngress(client, componentMeta, componentExt, podSpec, componentStatus),
+		componentExt:    componentExt,
+		componentStatus: componentStatus,
+	}
+}
+
+//GenerateIngressHost return the config domain in configmap.IngressDomain
+func GenerateIngressHost(client client.Client, metadata metav1.ObjectMeta) string {
+	ingressConfig, err := v1beta1api.NewIngressConfig(client)
+	if err != nil {
```
yuzisun, line 59: It could be configuration parsing error also, prefer not masking those issues. We can set the default value in config instead.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+		log.Error(err, "GenerateIngressHost")
+		if ingressConfig == nil {
+			ingressConfig = &v1beta1api.IngressConfig{}
+		}
+		if ingressConfig.IngressDomain == "" {
+			ingressConfig.IngressDomain = "1.2.3.4.xip.io"
+		}
+	}
+	return metadata.Name + "." + metadata.Namespace + "." + ingressConfig.IngressDomain
```
yuzisun, line 68: We may want the `domainTemplate` configuration also as it can be `Name-Namespace.Domain` if using wildcard certs

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -148,6 +152,56 @@ func (ss *InferenceServiceStatus) IsConditionReady(t apis.ConditionType) bool {
...
+	}
+	ss.SetNativeCondition(deployment)
+	ss.Components[component] = statusSpec
+}
+
+func getDeploymentCondition(deployment *appsv1.Deployment, conditionType appsv1.DeploymentConditionType) *appsv1.DeploymentCondition {
+	var condition *appsv1.DeploymentCondition
+	condition = nil
+	for _, con := range deployment.Status.Conditions {
+		if con.Type == appsv1.DeploymentAvailable {
```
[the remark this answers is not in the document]
Iamlovingit: deployment conditions: DeploymentAvailable, DeploymentProgressing, DeploymentReplicaFailure
now I matched the DeploymentAvailable to isvc's PredictorReady, what about the other conditions?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -148,6 +152,56 @@ func (ss *InferenceServiceStatus) IsConditionReady(t apis.ConditionType) bool {
...
+			condition = &con
+			break
+		}
+	}
+	return condition
+}
+
+func (ss *InferenceServiceStatus) SetNativeCondition(deployment *appsv1.Deployment) {
+	//get deployment available status
+	condition := getDeploymentCondition(deployment, appsv1.DeploymentAvailable)
```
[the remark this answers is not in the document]
Iamlovingit: Ye, I will modify the args

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -17,12 +17,16 @@ limitations under the License.
...
-	"k8s.io/api/core/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var log = logf.Log.WithName("inference service status")
```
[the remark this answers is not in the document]
Iamlovingit: OK

config/configmap/inferenceservice.yaml
```diff
@@ -150,7 +150,8 @@ data:
         "ingressGateway" : "$(ingressGateway)",
         "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
         "localGateway" : "knative-local-gateway.knative-serving",
-        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local"
+        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
+        "ingressDomain"  : "1.2.3.4.xip.io"
```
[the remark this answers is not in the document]
Iamlovingit: OK

pkg/constants/constants.go
```diff
@@ -220,6 +221,31 @@ var (
...
+)
+
+// GetNativeServiceLabel generate native service label
+func GetNativeServiceLabel(service string, namespace string) string {
+	return "isvc." + namespace + "." + service
+}
+
+// GetNativeServiceIngressName generate native service ingress name
+func GetNativeServiceIngressName(service string, namespace string) string {
+	return NativeIngressNamePrefix + "-" + namespace + "-" + service
```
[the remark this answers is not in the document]
Iamlovingit: This is the ingress name, not the ingress host, the ingress host is formated by `{{.Name}}.{{.Namespace}}.{{.Domain}} `

pkg/constants/constants.go
Iamlovingit: right, this is useless

pkg/constants/constants.go
Iamlovingit: ok

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: So, we change the ingress host template to `Name-Namespace.Domain`? I think it doesn't need to support two format

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,138 @@
...
+	service := &corev1.Service{
+		ObjectMeta: componentMeta,
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": constants.GetRawServiceLabel(componentMeta.Name),
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name: componentMeta.Name,
+					Port: podSpec.Containers[0].Ports[0].ContainerPort,
```
yuzisun, line 71: I think the service port can be just 80?

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,138 @@
...
+					TargetPort: intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: podSpec.Containers[0].Ports[0].ContainerPort,
```
yuzisun, line 74: same here

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,176 @@
...
+						HTTP: &netv1.HTTPIngressRuleValue{
+							Paths: []netv1.HTTPIngressPath{
+								{
+									Path:     "/" + getIngressPath(componentType),
+									PathType: &pathType,
+									Backend: netv1.IngressBackend{
+										Service: &netv1.IngressServiceBackend{
+											Name: componentMeta.Name,
+											Port: netv1.ServiceBackendPort{
+												Number: podSpec.Containers[0].Ports[0].ContainerPort,
```
yuzisun, line 98: If service port is set to 80 here can use the same.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,176 @@
...
+func createRawIngress(client client.Client, componentMeta metav1.ObjectMeta,
+	componentExt *v1beta1.ComponentExtensionSpec,
+	podSpec *corev1.PodSpec,
+	componentStatus v1beta1.ComponentStatusSpec,
+	componentType v1beta1.ComponentType) *netv1.Ingress {
+	ingressHost := GenerateIngressHost(client, componentMeta)
+	pathType := netv1.PathType(netv1.PathTypeImplementationSpecific)
+	ingress := &netv1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      componentMeta.Name,
```
yuzisun, line 81: For Istio Ingress gateway to work with Ingress CRD here we need the [additional annotation kubernetes.io/ingress.class: istio](https://istio.io/latest/docs/tasks/traffic-management/ingress/kubernetes-ingress/) I think

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,199 @@
...
+	if checkResult == constants.CheckResultCreate {
+		err = r.client.Create(context.TODO(), r.Deployment)
+	} else if checkResult == constants.CheckResultUpdate { //CheckResultUpdate
+		err = r.client.Update(context.TODO(), r.Deployment)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	err = r.client.Get(context.TODO(), types.NamespacedName{
```
yuzisun, line 191: We can return the deployment in `checkDeploymentExist` so we do not have to call twice

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+		Spec: netv1.IngressSpec{
+			Rules: []netv1.IngressRule{
+				{
+					Host: ingressHost,
+					IngressRuleValue: netv1.IngressRuleValue{
+						HTTP: &netv1.HTTPIngressRuleValue{
+							Paths: []netv1.HTTPIngressPath{
+								{
+									Path:     "/" + getIngressPath(componentType),
```
yuzisun, line 99: I think we should just do `/` which means all paths
https://kubernetes.io/docs/concepts/services-networking/ingress/#examples

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,145 @@
...
+					},
+					Protocol: corev1.ProtocolTCP,
+				},
+			},
+		},
+	}
+	return service
+}
+
+//checkNativeDeploymentExist checks is the deployment is existed?
```
yuzisun, line 91: /Deployment/Service/

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,145 @@
...
+func (r *ServiceReconciler) checkServiceExist(client client.Client) (constants.CheckResultType, error) {
+	//get deployment
```
yuzisun, line 93: /deployment/service/

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,145 @@
...
+		Name:      r.Service.Name,
+	}, exsitingService)
+	if err != nil {
+		if apierr.IsNotFound(err) {
+			return constants.CheckResultCreate, nil
+		}
+		return constants.CheckResultUnknown, err
+	}
+
+	//existed, check equalment
```
yuzisun, line 106: /equalment/equivalent/

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -69,6 +71,23 @@ type ComponentStatusSpec struct {
 	Address *duckv1.Addressable `json:"address,omitempty"`
 }
 
+type RawConditionType string
+type RawCondition struct {
```
yuzisun, line 75: Is there any reason not to use the buildin condition type?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
yuzisun: if users are using wildcard certs for all the services then they want `Name-Namespace.Domain`, but if user want one certificate per service then you need `Name.Namespace.Domain`

test/e2e/predictor/test_raw_deployment.py
```diff
@@ -0,0 +1,64 @@
...
+    isvc = V1beta1InferenceService(
+        api_version=api_v1beta1_version,
+        kind=constants.KFSERVING_KIND,
+        metadata=client.V1ObjectMeta(
+            name=service_name, namespace=KFSERVING_TEST_NAMESPACE, annotations=annotations,
+        ),
+        spec=V1beta1InferenceServiceSpec(predictor=predictor),
+    )
+
+    KFServing.create(isvc)
```
yuzisun, line 63: Can you add the test to send a request and assert the expected response?

Iamlovingit: Is there a ingress in the test env?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
Iamlovingit: Got it

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
Iamlovingit: thanks.

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
Iamlovingit: thanks.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: ok, this way is better.

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
Iamlovingit: thanks

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: ye, I pass the metadata from predictor

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: sorry, I can not get this meaning

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -104,15 +104,15 @@ func (r *DeploymentReconciler) checkDeploymentExist(client client.Client) (const
 	}, existingDeployment)
 	if err != nil {
 		if apierr.IsNotFound(err) {
-			return constants.CheckResultCreate, nil
+			return constants.CheckResultCreate, existingDeployment, nil
```
yuzisun, line 107: return nil in this case as the deployment is not found

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -117,20 +114,9 @@ func createRawIngress(client client.Client, componentMeta metav1.ObjectMeta,
...
-	} else if componentType == v1beta1.ComponentType(constants.Explainer) {
-		return "explain"
-	} else if componentType == v1beta1.ComponentType(constants.Transformer) {
-		return "transform"
-	}
-	return ""
-}
-
-//checkNativeDeploymentExist checks is the deployment is existed?
+//checkRawIngressExist checks is the ingress is existed?
```
yuzisun, line 117: /is the ingress is existed/if the ingress exists/

pkg/controller/v1beta1/inferenceservice/reconcilers/raw/raw_reconciler.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	knapis "knative.dev/pkg/apis"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//NativeReconciler is the struct of Native K8S Object
+type RawReconciler struct {
+	client          client.Client
+	scheme          *runtime.Scheme
+	Deployment      *deployment.DeploymentReconciler
```
yuzisun, line 33: lower case as this does not need to be exposed and can be named as `deploymentReconciler`

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+	if checkResult == constants.CheckResultCreate {
+		err = r.client.Create(context.TODO(), r.Service)
+	} else if checkResult == constants.CheckResultUpdate { //CheckResultUpdate
+		err = r.client.Update(context.TODO(), r.Service)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return r.Service, nil
```
yuzisun, line 136: r.Service should be returned from `checkServiceExist`?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	if checkResult == constants.CheckResultCreate {
+		err = r.client.Create(context.TODO(), r.Ingress)
+	} else if checkResult == constants.CheckResultUpdate { //CheckResultUpdate
+		err = r.client.Update(context.TODO(), r.Ingress)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return r.Ingress, nil
```
yuzisun, line 160: `r.Ingress` should be returned from `checkRawIngressExist` ?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,191 @@
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": constants.GetRawServiceLabel(componentMeta.Name),
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app": constants.GetRawServiceLabel(componentMeta.Name),
+					},
```
yuzisun, line 87: We need to pass the labels and annotations from `componentMeta` to the template, the pod mutator is checking [the annotations](https://github.com/kubeflow/kfserving/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L81) to inject the storage initializer.

Iamlovingit: got it, thanks

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+	service := &corev1.Service{
+		ObjectMeta: componentMeta,
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": constants.GetRawServiceLabel(componentMeta.Name),
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name: componentMeta.Name,
+					Port: int32(port),
```
yuzisun, line 78: I think here we might use port 80 otherwise when user is calling from in cluster they need to know the internal port the container exposes, we can do the mapping from 80 -> user port here.

pkg/apis/serving/v1beta1/transformer_custom.go
```diff
@@ -66,6 +66,15 @@ func (c *CustomTransformer) GetStorageUri() *string {
 // GetContainers transforms the resource into a container spec
 func (c *CustomTransformer) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
 	container := &c.Containers[0]
+	var argumentPredictorHost string
+	annotations := utils.Filter(metadata.Annotations, func(key string) bool {
+		return !utils.Includes(constants.ServiceAnnotationDisallowedList, key)
+	})
+	if value, ok := annotations[constants.RawDeploymentAnnotationKey]; ok && value == "true" {
```
yuzisun, line 73: I think here should be the same for raw deployment, namespace isn't strictly needed if all the resources are in the same namespace but better keeps the logic consistent.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+			Rules: rules,
+		},
+	}
+	if err := controllerutil.SetControllerReference(isvc, ingress, scheme); err != nil {
+		return nil, err
+	}
+	return ingress, nil
+}
+
+//checkRawIngressExist checks is the ingress is existed?
```
yuzisun, line 180:
```suggestion
//checkRawIngressExist checks if the ingress exists?
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+	} else {
+		return isvc.Name + "-" + isvc.Namespace + "." + ingressConfig.IngressDomain
+	}
+}
+
+func createRawIngress(client client.Client,
+	scheme *runtime.Scheme,
+	isvc *v1beta1api.InferenceService,
+	ingressConfig *v1beta1api.IngressConfig) (*netv1.Ingress, error) {
+	topLevelFlag := false
```
yuzisun, line 143: What does this flag do?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+func GenerateIngressHost(ingressConfig *v1beta1api.IngressConfig,
+	isvc *v1beta1api.InferenceService,
+	componentType string,
+	topLevelFlag bool) string {
+	if ingressConfig.IngressDomain == "" {
+		ingressConfig.IngressDomain = "example.com"
+	}
+	metadata := generateMetadata(isvc, constants.InferenceServiceComponent(componentType))
+	if topLevelFlag {
+		return metadata.Name + "-" + metadata.Namespace + "." + ingressConfig.IngressDomain
```
yuzisun, line 133: Do we need to generate the host at component level?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+	ingressHost string,
+	componentName string) netv1.IngressRule {
+	pathType := netv1.PathType(netv1.PathTypeImplementationSpecific)
+	rule := netv1.IngressRule{
+		Host: ingressHost,
+		IngressRuleValue: netv1.IngressRuleValue{
+			HTTP: &netv1.HTTPIngressRuleValue{
+				Paths: []netv1.HTTPIngressPath{
+					{
+						PathType: &pathType,
```
yuzisun, line 79: if there is explainer, we need to support both `:predict` and `:explain` which requires path based routing to predictor and explainer.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
```diff
@@ -0,0 +1,231 @@
...
+		host := GenerateIngressHost(ingressConfig, isvc, string(constants.Explainer), topLevelFlag)
+		//set topLevelFlag as true
+		topLevelFlag = true
+		podSpec := corev1.PodSpec(isvc.Spec.Explainer.PodSpec)
+		rules = append(rules, generateRule(&podSpec, host, constants.DefaultExplainerServiceName(isvc.Name)))
+	}
+	//add predictor rule
+	host := GenerateIngressHost(ingressConfig, isvc, string(constants.Predictor), topLevelFlag)
+	podSpec := corev1.PodSpec(isvc.Spec.Predictor.PodSpec)
+	rules = append(rules, generateRule(&podSpec, host, constants.DefaultPredictorServiceName(isvc.Name)))
```
yuzisun, line 162: How does this work when there are transformer and predictor, I think it creates the conflicting rules right?

test/e2e/transformer/test_raw_transformer.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_transformer():
+    service_name = 'raw'
+    predictor = V1beta1PredictorSpec(
+        min_replicas=1,
+        pytorch=V1beta1TorchServeSpec(
+            storage_uri='gs://kfserving-examples/models/torchserve/image_classifier',
+            resources=V1ResourceRequirements(
+                requests={'cpu': '1', 'memory': '2Gi'},
```
yuzisun, line 46: Can we set lower cpu request  like 100m as we have limited resource to run all the e2e tests

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+
+var log = logf.Log.WithName("DeploymentReconciler")
+
+//DeploymentReconciler is the struct of Raw K8S Object
+type DeploymentReconciler struct {
+	client          client.Client
+	scheme          *runtime.Scheme
+	Deployment      *appsv1.Deployment
+	componentExt    *v1beta1.ComponentExtensionSpec
+	componentStatus v1beta1.ComponentStatusSpec
```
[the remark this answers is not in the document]
Iamlovingit: the componentExt is used, the componentStatus is unused, will remove this one, thanks Dan

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+	if checkResult == constants.CheckResultCreate {
+		err = r.client.Create(context.TODO(), r.Deployment)
+	} else if checkResult == constants.CheckResultUpdate { //CheckResultUpdate
+		err = r.client.Update(context.TODO(), r.Deployment)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return deployment, nil
```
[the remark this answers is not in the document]
Iamlovingit: We have already returned the deployment object if `create` or `update` succeed. When failed, the deployment object is unused.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: This is a top level service flag, if transformerSpec is existed, inferenceservice host points to transformer, and set `toplevelFlag` with `true`, so the `explainer` and `predictor` will be set the component adress as the ingress host.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: I think we need this, because the transformer and predictor have same path.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: explainer and predictor have different host, if there are explainer and predictor, the top level will point to explainer.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
Iamlovingit: the host is different, so there is no conflicting.

test/e2e/transformer/test_raw_transformer.py
Iamlovingit: sure.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/rawingress_reconciler.go
yuzisun: there could be two verbs for explainer case and it is totally depending on the request path user specifies, when user calls `:predict` KFServing should send the request directly to predictor, when calling `:explain` path KFServing should send to the explainer.

pkg/constants/constants.go
```diff
@@ -193,7 +194,8 @@ const (
 
 // InferenceService container name
 const (
-	InferenceServiceContainerName = "kfserving-container"
+	InferenceServiceContainerName       = "kfserving-container"
+	InferenceServiceCustomContainerName = "kfserving-custom-container"
```
yuzisun, line 198: This constant is not used.

yuzisun:
@googlebot I consent

pkg/constants/constants.go
Iamlovingit: yeah, I forgot this.

yuzisun:
@Iamlovingit Thanks for the great work! This is exciting!
/lgtm
/approve

PR #1510, opened 2021-03-29
Introducing HTTPS protocol for MMS

abchoo:
**What this PR does / why we need it**: Introduces HTTPS protocol for multi-model serving. Now mms will be able to download files from s3, gcs, and https.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/agent/storage/https.go
+++ b/pkg/agent/storage/https.go
@@ -1,0 +1,210 @@
+/*
+Copyright 2021 kubeflow.org.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"archive/tar"
+	"archive/zip"
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	HEADER_SUFFIX = "-headers"
+)
+
+type HTTPSProvider struct {
+	Client *http.Client
+}
+
+func (m *HTTPSProvider) DownloadModel(modelDir string, modelName string, storageUri string) error {
+	log.Info("Download model ", "modelName", modelName, "storageUri", storageUri, "modelDir", modelDir)
+	uri, err := url.Parse(storageUri)
+	if err != nil {
+		return fmt.Errorf("unable to parse storage uri: %v", err)
+	}
+	HTTPSDownloader := &HTTPSDownloader{
+		StorageUri: storageUri,
+		ModelDir:   modelDir,
+		ModelName:  modelName,
+		Uri:        uri,
+	}
+	if err := HTTPSDownloader.Download(*m.Client); err != nil {
+		return err
+	}
+	return nil
+}
+
+type HTTPSDownloader struct {
+	StorageUri string
+	ModelDir   string
+	ModelName  string
+	Uri        *url.URL
+}
+
+func (h *HTTPSDownloader) Download(client http.Client) error {
+	// Create request
+	req, err := http.NewRequest("GET", h.StorageUri, nil)
+	if err != nil {
+		return err
+	}
+
+	headers, err := h.extractHeaders()
+	if err != nil {
+		return err
+	}
+	for key, element := range headers {
+		req.Header.Add(key, element)
+	}
+
+	// Query request
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to make a request: %v", err)
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("URI: %s returned a %d response code", h.StorageUri, resp.StatusCode)
+	}
+
+	// Write content into file(s)
+	contentType := resp.Header.Get("Content-type")
+	fileDirectory := filepath.Join(h.ModelDir, h.ModelName)
+
+	if strings.Contains(contentType, "application/zip") {
+		if err := extractZipFiles(resp.Body, fileDirectory); err != nil {
+			return err
+		}
+	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/gzip") {
+		if err := extractTarFiles(resp.Body, fileDirectory); err != nil {
+			return err
+		}
+	} else {
+		paths := strings.Split(h.Uri.Path, "/")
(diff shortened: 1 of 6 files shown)


pkg/apis/serving/v1alpha1/trainedmodel_webhook_test.go
```diff
@@ -96,9 +96,9 @@ func TestValidateCreate(t *testing.T) {
 		"invalid storageURI prefix": {
 			tm: makeTestTrainModel(),
 			update: map[string]string{
-				storageURI: "https://kfserving/sklearn/iris",
+				storageURI: "pvc://kfserving/sklearn/iris",
```
adriangonz, line 99: Should replace this by an obviously fake schema (i.e. `foo://`)?

pkg/agent/storage/utils.go
```diff
@@ -132,7 +133,11 @@ func GetProvider(providers map[Protocol]Provider, protocol Protocol) (Provider,
 				}),
 			}
 		}
-
+	case HTTPS:
+		var httpsClient HTTPSClient = &http.Client{}
```
adriangonz, line 137: Thinking out loud here, but how would you see the agent supporting HTTPS endpoints with self-signed certificates? 

I can see how this may be out of scope of this PR, but would be interesting to see how this could be implemented as a next step.

pkg/agent/storage/https.go
```diff
@@ -0,0 +1,218 @@
...
+		return fmt.Errorf("failed to make a request: %v", err)
+	}
+
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("URI: %s returned a %d response code", h.StorageUri, resp.StatusCode)
+	}
+
+	// Write content into file(s)
+	if resp.Uncompressed {
```
adriangonz, line 99: Doesn't the `http.Transport` class deal with compressed responses under the hood?

abchoo: I believe [Transport sets the compression to gzip and uncompresses the response (check out DisableCompression)](https://golang.org/pkg/net/http/#Transport) when DisableCompression is false. Seems to me like the actual file content will still stay in its original file format (ie `tar` or `zip`) so we will need to implement extraction ourselves.

pkg/apis/serving/v1alpha1/trainedmodel_webhook_test.go
abchoo: I wanted to set the prefix to one of the non-implemented protocols in [Provider.go](https://github.com/kubeflow/kfserving/blob/c1b9fa07d1e6bee4ad2fde2d7cb394089c29b732/pkg/agent/storage/provider.go#L25) but this makes sense as well. Will make that change to avoid future problems.

pkg/agent/storage/https.go
abchoo: Yeah! Otherwise the trained model files within would not be accessible. 

Btw I added zip/gzip/tar to mimic how [single model serving downloads from https](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/kfserving/storage.py#L229).

pkg/agent/storage/utils.go
abchoo: Could you elaborate more on this? 

I'm not too familiar with self-signed certificates. Would the agent behaviour be drastically different?

adriangonz: By default, the `net/http` package in Go will reject sending requests to `https://` endpoints signed with self-signed certificates (as it can't verify them). To override that behaviour, you need to change the `http.Transport` used, by one which has the `InsecureSkipVerify` flag enabled in the `TLSClientConfig` object. 

However, this is only useful for self-signed certs. With regular ones, you do want to verify them. I'm wondering whether this could be a field in one of the credentials secrets which gets picked by the agent?

pkg/agent/storage/https.go
adriangonz: Got it! It now makes sense.

adriangonz:
Looks good from my side!

/lgtm

pkg/agent/storage/utils.go
adriangonz: I'm wondering whether we should open this as a separate issue?

pkg/agent/mocks/httpsmock.go
```diff
@@ -0,0 +1,31 @@
+package mocks
```
yuzliu, line 1: https://github.com/golang/mock/blob/master/gomock/call.go#L111 can this existing mock library be used here?

pkg/agent/storage/utils.go
yuzisun: I think if the nodes are installed with root certificates then the https client should just work, otherwise you'd need to load the certificate and set the TLS transport on the client.

pkg/agent/mocks/httpsmock.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+	"github.com/kubeflow/kfserving/pkg/agent/storage"
+	"net/http"
+	"net/url"
+)
+
+type mockHTTPSClient struct {
```
yuzisun, line 10: I think you can use the `"net/http/httptest"` package

pkg/agent/storage/https.go
```diff
@@ -0,0 +1,214 @@
...
+	if strings.Contains(contentType, "application/zip") {
+		if err := extractZipFiles(resp.Body, fileDirectory); err != nil {
+			return err
+		}
+	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/gzip") {
+		if err := extractTarFiles(resp.Body, fileDirectory); err != nil {
+			return err
+		}
+	} else {
+		paths := strings.Split(h.Uri.Path, "/")
```
yuzisun, line 111: Do we validate if the url path always include a file?

pkg/agent/storage/provider.go
```diff
@@ -27,9 +27,14 @@ const (
 	GCS Protocol = "gs://"
 	//PVC   Protocol = "pvc://"
 	//File  Protocol = "file://"
-	//HTTPS Protocol = "https://"
+	HTTPS Protocol = "https://"
```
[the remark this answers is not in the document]
abchoo: Good idea! Will add this protocol!

pkg/agent/mocks/httpsmock.go
abchoo: I'll check it out and see if I can use that package instead. Thanks!

pkg/agent/storage/https.go
abchoo: We do not validate if the url path always include a file. Should we? It seems difficult to do so as the file can technically be in any format in this case.

pkg/agent/mocks/httpsmock.go
abchoo: I'll check it out! Thanks!

pkg/agent/storage/utils.go
```diff
@@ -132,7 +133,16 @@ func GetProvider(providers map[Protocol]Provider, protocol Protocol) (Provider,
 				}),
 			}
 		}
-
+	case HTTPS:
```
yuzisun, line 136: nit: can be combined to `case HTTPS, HTTP:`

yuzisun:
Awesome work! thanks @abchoo !

/lgtm
/approve

PR #1511, opened 2021-03-29
Add a GH action for the ui

### LINKED ISSUE #1504: web-app: Implement a GitHub action for building the web app
/kind feature

We will need to add a new GitHub action, to the list of [existing workflows](https://github.com/kubeflow/kfserving/tree/master/.github/workflows), for building the InferenceServices web app #1327. The action will first test that the image can be built and then it will publish it.

My main question about this is what `$IMAGE_NAME` should we use in this case? I can think of the following, but would really like to hear your thoughts as well:
* `inferenceservices-web-app`
* `models-web-app`
* `web-app`

Will get to this right after the code for the app is merged #1328

cc @kubeflow/wg-serving-leads 
/assign @kimwnasptd

kimwnasptd:

Fixes #1504 

I've made the following assumptions in this PR:
1. The app is named `models-web-app`, since the `InferenceService` objects are model servers
2. It's using the `kfserving/models-web-app` image name
3. I've used the yaml from [alibiexplainer-docker-publish.yml](https://github.com/kubeflow/kfserving/blob/master/.github/workflows/alibiexplainer-docker-publish.yml) and slightly modified for this action

I don't have any hard preference for the name so if you'd like to use a different name feel then please let me know and I'll change it accordingly. 

/cc @yuzisun @animeshsingh

### DIFF
--- a/.github/workflows/web-app-docker-publish.yml
+++ b/.github/workflows/web-app-docker-publish.yml
@@ -1,0 +1,79 @@
+name: Models web app Docker Publisher
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  IMAGE_NAME: models-web-app
+
+jobs:
+  # Run tests.
+  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run tests
+        run: |
+          if [ -f docker-compose.test.yml ]; then
+            docker-compose --file docker-compose.test.yml build
+            docker-compose --file docker-compose.test.yml run sut
+          else
+            cd web-app
+            docker build . --file Dockerfile
+          fi
+
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+    # Ensure test job passes before pushing image.
+    needs: test
+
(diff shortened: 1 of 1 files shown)


.github/workflows/web-app-docker-publish.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  IMAGE_NAME: models-web-app
```
yuzisun, line 17: @animeshsingh Any preference with the image name?

pvaneck: `models-web-app` works for me.

kimwnasptd: @animeshsingh I'll respond here to your comment https://github.com/kubeflow/kfserving/pull/1511#issuecomment-815306317 to keep the conversation in one place.

I don't have a strong opinion on the name. The only reason I named it `models-web-app` was because I was thinking of placing the app under `Models` in the KF CentralDashboard in the future. 

I think the argument for shorter names is a good one. Looking at the repo folders and the docker images I see that  all of the components are named as one word. So it would make sense to keep this pattern and name it just `ui`, or something similar. The docker image would end up being `kfserving/ui`. We also follow this code naming with Katib [[code](https://github.com/kubeflow/katib/tree/master/pkg/ui)].

@yuzisun @pvaneck @animeshsingh what do you think? If you'd prefer the `ui` name I can make the change to the code as well. If you don't have a hard preference I guess we can keep `models-web-app`(?)

kimwnasptd: @yuzisun @pvaneck @animeshsingh `models-web-app` it is then?

PR #1513, opened 2021-03-30
[xgboostserver] Convert list input to numpy array before creating DMatrix

pradithya:
Due to https://github.com/dmlc/xgboost/pull/3970 , prediction calculated by current xgboost model is incorrect for some cases. The result deviates quite significantly compared to training set. 

Fix incorrect prediction result in xgboost server.

- Additional dependency (numpy) is added to xgboost server

**Release note**:

```release-note

```

### DIFF
--- a/python/xgbserver/setup.py
+++ b/python/xgbserver/setup.py
@@ -33,7 +33,6 @@ setup(
     python_requires='>3.4',
     packages=find_packages("xgbserver"),
     install_requires=[
-        "numpy>=1.0"
         "kfserving>=0.5.1",
         "xgboost == 0.82",
         "scikit-learn == 0.20.3",
--- a/python/xgbserver/setup.py
+++ b/python/xgbserver/setup.py
@@ -33,6 +33,7 @@ setup(
     python_requires='>3.4',
     packages=find_packages("xgbserver"),
     install_requires=[
+        "numpy>=1.0"
         "kfserving>=0.5.1",
         "xgboost == 0.82",
         "scikit-learn == 0.20.3",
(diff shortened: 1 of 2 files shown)


pradithya:
It seems the CI failure is unrelated to the PR

```
        testset = torchvision.datasets.CIFAR10(root='./data', train=False,
>                                              download=True, transform=transform)
```

python/xgbserver/setup.py
```diff
@@ -33,6 +33,7 @@
     python_requires='>3.4',
     packages=find_packages("xgbserver"),
     install_requires=[
+        "numpy>=1.0"
```
yuzisun, line 36: kfserving package already include the numpy dependency

pradithya: Updated in new commit

python/xgbserver/xgbserver/model.py
```diff
@@ -43,7 +44,7 @@ def load(self) -> bool:
     def predict(self, request: Dict) -> Dict:
         try:
             # Use of list as input is deprecated see https://github.com/dmlc/xgboost/pull/3970
```
pradithya, line 46: @yuzisun I am curious why this hasn't been fixed before even though there is a comment pointing to the issue in xgboost.

yuzisun: @pradithya We are mostly moving to v2 protocol at this point for xgboost and I think it is fixed there https://github.com/SeldonIO/MLServer/blob/739ebbaf92f1c135c789df009f2c386c6ab6f49e/runtimes/xgboost/mlserver_xgboost/xgboost.py#L49.

yuzisun:
@pradithya If you want this to be released, please help cherry pick onto the release-0.5 branch.

python/xgbserver/xgbserver/model.py
pradithya: Got it, thanks for clarifying @yuzisun

PR #1522, opened 2021-04-05
Adds condition for Trained Model to check if isvc predictor supports MMS

abchoo:
**What this PR does / why we need it**: Trained Model requires the parent inference service to enable multi-model serving. This PR adds a condition to let users know that their model is not Ready if parent's predictor is not configured to mms.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/serving/v1alpha1/trained_model_status.go
+++ b/pkg/apis/serving/v1alpha1/trained_model_status.go
@@ -43,6 +43,8 @@ const (
 	FrameworkSupported apis.ConditionType = "FrameworkSupported"
 	// MemoryResourceAvailable is set when inference service reported resources availability
 	MemoryResourceAvailable apis.ConditionType = "MemoryResourceAvailable"
+	// InferenceServiceMMSPredictor is set when inference service predictor is set to multi-model serving
+	InferenceServiceMMSPredictor apis.ConditionType = "InferenceServiceMMSPredictor"
 )
 
 // TrainedModel Ready condition is depending on inference service readiness condition
@@ -51,6 +53,7 @@ var conditionSet = apis.NewLivingConditionSet(
 	InferenceServiceReady,
 	FrameworkSupported,
 	MemoryResourceAvailable,
+	InferenceServiceMMSPredictor,
 )
 
 var _ apis.ConditionsAccessor = (*TrainedModelStatus)(nil)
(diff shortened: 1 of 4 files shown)


pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -101,6 +101,11 @@ func (s *PredictorSpec) GetImplementation() ComponentImplementation {
 	return s.GetImplementations()[0]
 }
 
+// IsEmpty returns if all components does not exist
+func (s *PredictorSpec) IsEmpty() bool {
```
yuzisun, line 105: How this is different from `GetPredictor` ?

pkg/apis/serving/v1alpha1/trained_model_status.go
```diff
@@ -43,6 +43,8 @@ const (
 	FrameworkSupported apis.ConditionType = "FrameworkSupported"
 	// MemoryResourceAvailable is set when inference service reported resources availability
 	MemoryResourceAvailable apis.ConditionType = "MemoryResourceAvailable"
+	// InferenceServiceMMSPredictor is set when inference service predictor is set to multi-model serving
+	InferenceServiceMMSPredictor apis.ConditionType = "InferenceServiceMMSPredictor"
```
yuzisun, line 47: Since this is a boolean condition, maybe better to call `IsMMSPredictor` ?

pkg/apis/serving/v1beta1/predictor.go
abchoo: Hmm I may need to refractor the names for clarity. `GetPredictor` will return a `PredictorImplementation` if one exists. I wanted a way to retrieve a `ComponentImplementation` if one exists and I decided to create a function `IsEmpty` to determine if any component implementation exists.

pkg/controller/v1alpha1/trainedmodel/controller.go
```diff
@@ -229,28 +230,47 @@ func (r *TrainedModelReconciler) updateConditions(req ctrl.Request, tm *v1alpha1
...
-	// Update Framework Supported condition
 	isvcConfig, err := v1beta1api.NewInferenceServicesConfig(r.Client)
 	if err != nil {
 		return err
+	}
+
+	// Update Is MMS Predictor condition
+	implementations := isvc.Spec.Predictor.GetImplementations()
+	if len(implementations) > 0 && v1beta1utils.IsMMSPredictor(&isvc.Spec.Predictor, isvcConfig) {
+		log.Info("Predictor is multi-model serving", "TrainedModel", tm.Name, "InferenceService", isvc.Name)
```
yuzisun, line 241: Is this logging needed? it can spew the log when you deploy large amount of trained models.

abchoo: I wanted to be consistent with the rest of the condition updates but I suppose on an Info level that would be excessive.

yuzisun (commented):
/lgtm
/approve

PR #1533, opened 2021-04-14
Fix logger for error response case

### LINKED ISSUE #1530: Model server response status code not propagated when using logger
/kind bug

First, I deployed a tensorflow model following the sample [here](https://github.com/kubeflow/kfserving/blob/master/docs/samples/v1beta1/tensorflow/tensorflow.yaml). Making inference with a **valid** payload returns _200 OK_ and the corresponding predictions. Trying with an invalid payload (e.g replacing 'instances' with 'invalid' or using an empty value) returns _400 Bad Request_ together with the error message (e.g { "error": "Missing 'inputs' or 'instances' key" })

Then, I added inference logging following the sample [here](https://github.com/kubeflow/kfserving/blob/master/docs/samples/logger/basic/message-dumper.yaml). Using valid payloads returns _200 OK_ and the predictions as expected. However, when using an invalid payload the model server response is not propagated and always returns 200 OK and the error message in the response body.

I expect the model server response status code to be propagated as well as the response body and get _400 Bad Request_ instead of _200 OK_.

The same happens for sklearn models. I believe the bug is in the [logger handler](https://github.com/kubeflow/kfserving/blob/master/pkg/logger/handler.go).

**Environment:**

- Istio Version: 1.7.2
- Knative Version: 0.17.0
- KFServing Version: 0.5.1
- Kubeflow version: --
- Kfdef: k8s_istio
- Minikube version: --
- Kubernetes version: Client Version: 1.18, GitVersion: v1.18.8 //  Server Version: 1.18, GitVersion: v1.18.8
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.5 LTS

yuzisun:
There is a regression to propagate the error response when we consolidate the logger to the agent.

Fixes #1530 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/logger/handler.go
+++ b/pkg/logger/handler.go
@@ -125,7 +125,11 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	} else {
 		eh.log.Info("Failed to proxy request", "status code", rr.Code)
 	}
-
+	contentType := rr.Header().Get("Content-Type")
+	if contentType != "" {
+		w.Header().Set("Content-Type", contentType)
+	}
+	w.WriteHeader(rr.Code)
 	_, err = w.Write(rr.Body.Bytes())
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/pkg/logger/handler.go
+++ b/pkg/logger/handler.go
@@ -82,13 +82,13 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Get or Create an ID
 	id := getOrCreateID(r)
-
+	contentType := r.Header.Get("Content-Type")
 	// log Request
 	if eh.logMode == v1beta1.LogAll || eh.logMode == v1beta1.LogRequest {
 		if err := QueueLogRequest(LogRequest{
 			Url:              eh.logUrl,
 			Bytes:            &body,
-			ContentType:      "application/json", // Always JSON at present
+			ContentType:      contentType,
 			ReqType:          InferenceRequest,
 			Id:               id,
 			SourceUri:        eh.sourceUri,
@@ -105,13 +105,17 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	rr := httptest.NewRecorder()
 	eh.next.ServeHTTP(rr, r)
 	responseBody := rr.Body.Bytes()
+	contentType = rr.Header().Get("Content-Type")
+	if contentType != "" {
+		w.Header().Set("Content-Type", contentType)
+	}
 	// log response if OK
 	if rr.Code == http.StatusOK {
 		if eh.logMode == v1beta1.LogAll || eh.logMode == v1beta1.LogResponse {
 			if err := QueueLogRequest(LogRequest{
 				Url:              eh.logUrl,
 				Bytes:            &responseBody,
-				ContentType:      "application/json", // Always JSON at present
+				ContentType:      contentType,
 				ReqType:          InferenceResponse,
 				Id:               id,
 				SourceUri:        eh.sourceUri,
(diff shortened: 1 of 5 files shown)


pkg/logger/handler.go
```diff
@@ -125,7 +125,11 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	} else {
 		eh.log.Info("Failed to proxy request", "status code", rr.Code)
 	}
-
+	contentType := rr.Header().Get("Content-Type")
```
theofpa, line 128: we can move this before line 109 and replace the condition:
```
	if rr.Code == http.StatusOK {
```
with
```
	if contentType == "application/json" {
```
By doing so, we'll actually write the error log in the cloudevent as well, as long as it's json, as per the request of @javierdlrm

theofpa (changes requested):
I created a new e2e test and the current change doesn't write the error response to the cloudevent, see below. Can I edit the PR to add the test, see theofpa/kfserving@fa0caf60f0d228769993903bc6049c905902aa48

pkg/logger/handler.go
javierdlrm: It looks good! Just another remark, having a look here https://github.com/theofpa/kfserving/commit/fa0caf60f0d228769993903bc6049c905902aa48, in case the response is OK but `Content-Type: application/json` is not present in the headers, `Failed to proxy request` will appear in the logs although the request was successful. In that case, the log is misleading.

javierdlrm: Maybe we can move line 127 ([commit](https://github.com/theofpa/kfserving/commit/fa0caf60f0d228769993903bc6049c905902aa48)) into a different if-condition checking for OK status code, just below the content-type condition.

yuzisun:
> I created a new e2e test and the current change doesn't write the error response to the cloudevent, see below and [theofpa/kfserving@fa0caf6](https://github.com/theofpa/kfserving/commit/fa0caf60f0d228769993903bc6049c905902aa48). Can I edit the PR to add the test?

Do we need to capture the error response?

pkg/logger/handler_test.go
```diff
@@ -89,3 +89,41 @@ func TestLogger(t *testing.T) {
 	// get logResponse
 	<-responseChan
 }
+
+func TestBadResponse(t *testing.T) {
+
+	g := gomega.NewGomegaWithT(t)
+
+	predictorRequest := []byte(`{"instances":[[0,0,0]]}`)
+	predictorResponse := "BadRequest\n"
```
theofpa, line 98: The case of tensorflow-model-server that returns 400 with json which can be logged is not captured by this plain text response. We can try forcing an application-json content-type like in theofpa/kfserving@9330ec7af9036915e985aa8743a387173aa4827d

theofpa:
> > I created a new e2e test and the current change doesn't write the error response to the cloudevent, see below and [theofpa/kfserving@fa0caf6](https://github.com/theofpa/kfserving/commit/fa0caf60f0d228769993903bc6049c905902aa48). Can I edit the PR to add the test?
> 
> 
> 
> Do we need to capture the error response?

Yes, when producing statistics about the failures, based on logs, that would be useful.

yuzisun:
> > > I created a new e2e test and the current change doesn't write the error response to the cloudevent, see below and [theofpa/kfserving@fa0caf6](https://github.com/theofpa/kfserving/commit/fa0caf60f0d228769993903bc6049c905902aa48). Can I edit the PR to add the test?
> > 
> > 
> > Do we need to capture the error response?
> 
> Yes, when producing statistics about the failures, based on logs, that would be useful.

There was an [issue](https://github.com/kubeflow/kfserving/issues/778) and we did not log error response intentionally, I think this needs to be an option.
```
Some use cases for the logger are to store the core request, responses and allow them to be further analyzed for outlier detection, drift detection, adversarial detection etc. If we allow error payloads then this creates a different and unexpected payloads and is a different use case I think.
We could add error_response as a new type and it would need to be given a different cloud event type so downstream processes like above could ignore them.
```

PR #1541, opened 2021-04-19
fix(web-app): Fix Grafana detection due to HEAD request

### LINKED ISSUE #1540: web-app: METRICS tab shown even if Grafana is not exposed in Kubeflow
/kind bug

I launched the web app, alongside Kubeflow 1.3 RC, using the manifests from #1512 without [exposing Grafana](https://docs.google.com/document/d/1MaCqHeqd7NiP1nseWkzqqAntRr3PiEH7k_7WnKjQSPU/edit#heading=h.f8a7gznpw4dr) outside the cluster and I still saw a `METRICS` tab. The graphs had errors in them, since Grafana does not exist in my cluster.

The web app to not show the `METRICS` tab at all.

This is because currently the UI makes a [HEAD request](https://github.com/kubeflow/kfserving/blob/5705bab63c7483b7b77712791394cd12713ac771/web-app/frontend/src/app/pages/server-info/server-info.component.ts#L91-L102) to `/grafana` to detect if the service is exposed. But when deployed with Kubeflow, under the `/models/` prefix, then the Central Dashboard will be returning it's `index.html` file for that route even if the endpoint doesn't exist. This makes the app think that it found Grafana, since the request succeeded.

The UI should instead make a GET request to a Grafana API endpoint, like `/api/search` and ensure that the response is formatted as expected.

cc @pvaneck who also bumped into this https://github.com/kubeflow/kfserving/pull/1512#pullrequestreview-635140387

kimwnasptd:
Makes the app's logic for detecting Grafana more robust. Instead of only relying on whether the request failed or the UI should also check if the response is formatted as expected.

This will mitigate the current issue where, in Kubeflow, the Central Dashboard will be returning its own `index.html` for non-matched routes. The Central Dashboard also listens to the root URL `/` and that's why it will return a response for the `/grafana` endpoint as well.

Fixes #1540

In order to test the changes you will have to deploy the app from it's manifests:
1. Checkout the code from #1512, or navigate to `/config` once it gets merged
2. Run `kustomize build config/overlays/kubeflow | kubectl apply -f -`
3. Connect to `$KF_IP/models/`
4. Click on the name of an InferenceService to see its details

At that point you should **not** see the `METRICS` tab in the web app.

/cc @pvaneck @StefanoFioravanzo @tasos-ale

### DIFF
--- a/web-app/frontend/src/app/pages/server-info/server-info.component.ts
+++ b/web-app/frontend/src/app/pages/server-info/server-info.component.ts
@@ -89,14 +89,27 @@ export class ServerInfoComponent implements OnInit, OnDestroy {
     });
 
     // don't show a METRICS tab if Grafana is not exposed
+    console.log('Checking if Grafana endpoint is exposed');
+    const grafanaApi = environment.grafanaPrefix + '/api/search';
+
     this.http
-      .head(environment.grafanaPrefix)
+      .get(grafanaApi)
       .pipe(timeout(1000))
       .subscribe({
-        next: () => {
+        next: resp => {
+          if (!Array.isArray(resp)) {
+            console.log(
+              'Response from the Grafana endpoint was not as expected.',
+            );
+            this.grafanaFound = false;
+            return;
+          }
+
+          console.log('Grafana endpoint detected. Will expose a metrics tab.');
           this.grafanaFound = true;
         },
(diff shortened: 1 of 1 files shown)


pvaneck (commented):
/lgtm

PR #1544, opened 2021-04-19
Remove application cr from kubeflow overlay

pvaneck:
All Kubeflow components should have this removed per: https://github.com/kubeflow/manifests/issues/1715


Tomcli:
thanks @pvaneck 
/lgtm

PR #1549, opened 2021-04-20
Prevents /mnt/models/<model name> from being converted into a file

abchoo:
**What this PR does / why we need it**: The model directory will be converted into a file whenever the fileName (Model dir, model name, objectName) is equal to the model directory. This will prevent that override from happening.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/agent/storage/gcs.go
+++ b/pkg/agent/storage/gcs.go
@@ -62,6 +62,7 @@ func (g *GCSObjectDownloader) Download(client stiface.Client, it stiface.ObjectI
 	var errs []error
 	// flag to help determine if query prefix returned an empty iterator
 	var foundObject = false
+	filePath := filepath.Join(g.ModelDir, g.ModelName)
 	for {
 		attrs, err := it.Next()
 		if err == iterator.Done {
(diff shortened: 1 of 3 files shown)


abchoo:
The GCS test was writing the model content into `/mnt/models/model1` which should be a directory but became a file. Updated this to write to `/mnt/models/model1/testModel1` to reflect the expected behaviour.

pkg/agent/storage/gcs.go
```diff
@@ -70,9 +71,12 @@ func (g *GCSObjectDownloader) Download(client stiface.Client, it stiface.ObjectI
 		if err != nil {
 			return fmt.Errorf("an error occurred while iterating: %v", err)
 		}
-		foundObject = true
 		objectValue := strings.TrimPrefix(attrs.Name, g.Item)
 		fileName := filepath.Join(g.ModelDir, g.ModelName, objectValue)
+		if fileName == filePath {
```
yuzisun, line 76: So the case here is that objectValue is empty string?

abchoo: This case covers when objectValue is `` or `/`.

PR #1573, opened 2021-04-28
mwa(front): Don't hardcode the base-url

### LINKED ISSUE #1572: Can't configure the prefix of the web app
/kind bug

Set the `APP_PREFIX: /` env var in the app and exposed it locally via port-forward, while testing the commands for #1571.

After I port-forwarded and connected the page was blank.

See the app loading as expected

The `index.html` file had the `<base href="/">` element, which means the backend correctly set the prefix. But the issue is in the fact that the [build process](https://github.com/kubeflow/kfserving/blob/master/web-app/frontend/package.json#L7) hardcodes the `/models/` prefix in front of the script files [via the `--deploy-url` arg]. This results in the browser trying to fetch files from a wrong prefix.

/label kfserving/webapp
/assign @kimwnasptd

kimwnasptd:
Makes the `APP_PREFIX` env var to properly configure the prefix of the app. The script tags will be using relative urls so the browser will be appending the prefix found from the `<base href=...">` element.

Fixes #1572 

You can try out this PR by following the steps to expose the app locally from #1571 

cc @yuzisun
/cc @tasos-ale @pvaneck

### DIFF
--- a/web-app/frontend/package.json
+++ b/web-app/frontend/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "ng": "ng",
     "format": "prettier --write './**/*.{ts,html,css}'",
-    "build": "npm run copyLibAssets && ng build --prod --base-href /models/ --deploy-url /models/static/",
+    "build": "npm run copyLibAssets && ng build --prod --deploy-url static/",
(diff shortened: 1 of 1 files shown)


yuzisun (commented):
/lgtm

PR #1592, opened 2021-05-03
Update Azure Blob to 12.1.0

maganaluis:

Fixes # https://github.com/kubeflow/kfserving/issues/923

**Release note**:
```
1. Upgraded Azure Blob to 12.1.0 as part of the storage initializer in KFServing. 
```

### DIFF
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -180,19 +180,8 @@ The path or model %s does not exist." % uri)
         blobs = container_client.list_blobs(prefix=prefix)
         for blob in blobs:
             dest_path = os.path.join(out_dir, blob.name)
-            if "/" in blob.name:
-                head, tail = os.path.split(blob.name)
-                if prefix is not None:
-                    head = head[len(prefix):]
-                if head.startswith('/'):
-                    head = head[1:]
-                dir_path = os.path.join(out_dir, head)
-                dest_path = os.path.join(dir_path, tail)
-                if not os.path.isdir(dir_path):
-                    os.makedirs(dir_path)
-
-            logging.info("Downloading: %s to %s", blob.name, dest_path)
             Path(os.path.dirname(dest_path)).mkdir(parents=True, exist_ok=True)
+            logging.info("Downloading: %s to %s", blob.name, dest_path)
             downloader = container_client.download_blob(blob.name)
             with open(dest_path, "wb+") as f:
                 f.write(downloader.readall())
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -24,12 +24,13 @@ import tarfile
 import tempfile
 import zipfile
 from urllib.parse import urlparse
+import requests
+from pathlib import Path
+from azure.storage.blob import BlobServiceClient
 
 from botocore.client import Config
 from botocore import UNSIGNED
 import boto3
-import requests
-from azure.storage.blob import BlockBlobService
 from google.auth import exceptions
 from google.cloud import storage
 
@@ -38,6 +39,7 @@ from kfserving.kfmodel_repository import MODEL_MOUNT_DIRS
 _GCS_PREFIX = "gs://"
 _S3_PREFIX = "s3://"
 _BLOB_RE = "https://(.+?).blob.core.windows.net/(.+)"
+_ACCOUNT_RE = "https://(.+?).blob.core.windows.net"
 _LOCAL_PREFIX = "file://"
 _URI_RE = "https?://(.+)/(.+)"
 _HTTP_PREFIX = "http(s)://"
@@ -160,6 +162,7 @@ The path or model %s does not exist." % uri)
     @staticmethod
     def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
         match = re.search(_BLOB_RE, uri)
+        account_url = re.search(_ACCOUNT_RE, uri).group(0)
         account_name = match.group(1)
         storage_url = match.group(2)
         container_name, prefix = storage_url.split("/", 1)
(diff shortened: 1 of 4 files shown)


yuzisun:
@maganaluis Thanks for taking care of this issue! Can you check the CLA seems like the commit includes some other author.

maganaluis:
@yuzisun CLA should be good now.

python/kfserving/kfserving/storage.py
```diff
@@ -173,7 +172,9 @@ def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
                     os.makedirs(dir_path)
 
             logging.info("Downloading: %s to %s", blob.name, dest_path)
-            block_blob_service.get_blob_to_path(container_name, blob.name, dest_path)
+            downloader = blob_service_client.get_container_client(container_name).download_blob(blob.name)
```
yuzisun, line 175: `get_container_client` gets called twice here.

python/kfserving/kfserving/storage.py
```diff
@@ -150,15 +149,15 @@ def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
                      container_name,
                      prefix)
         try:
-            block_blob_service = BlockBlobService(account_name=account_name)
-            blobs = block_blob_service.list_blobs(container_name, prefix=prefix)
-        except Exception:  # pylint: disable=broad-except
+            connect_str = os.getenv('AZURE_STORAGE_CONNECTION_STRING')
```
yuzisun, line 152: how is this environment variable going to be set?

yuzisun:
Thanks @maganaluis ! would be highly appreciable if you can help add an e2e test for azure.

python/kfserving/kfserving/storage.py
maganaluis: Good question. The connection string, is highlighted in the documentation, for best practices.

https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python

However the storage initializer sidecar will not have a way to provide this information or a way to support it. Do we want to add support here? I think that would be the best approach. 

https://github.com/kubeflow/kfserving/blob/master/docs/samples/storage/azure/README.md
https://github.com/kubeflow/kfserving/blob/42007c532286a1e43893ef2be03b15e104bfd7a4/pkg/credentials/azure/azure_secret.go

Let me know your thoughts.

python/kfserving/kfserving/storage.py
maganaluis: Yup, will take care of this. Let's continue the discussion on how to allow users to setup the AZURE_STORAGE_CONNECTION_STRING, as this will slightly broaden the scope of this PR.

yuzisun:
@maganaluis can you help resolve the conflicts here?

maganaluis:
I corrected the conflicts. Could you check this https://github.com/kubeflow/kfserving/issues/1671 we are working on an event driven architecture with KFServing :)

maganaluis:
@yuzisun I corrected the conflicts. Unrelated, but could you provide input on https://github.com/kubeflow/kfserving/issues/1671?  :)

yuzisun:
> @yuzisun I corrected the conflicts. Unrelated, but could you provide input on #1671? :)
> 
> Also note I removed checking for env variables, given this code runs within the storage sidecar there is currently no way to set this up.

@maganaluis You should be able to easily test storage initializer with `make deploy-dev-storageInitializer`! Can you also help fix the lint issue?

yuzisun:
@maganaluis Can you help fix the lint and tests?

maganaluis:
@yuzisun I corrected the test cases, there is a lot of lint errors not related to my change. Is there a particular way to run the lint check? 

Also, I switched to use azure-identity to create the token. This seems to be the best way to authenticate with the latest version of Azure Blob. I tested this on a minikube cluster with the latest version of KFServing.

yuzisun:
> @yuzisun I corrected the test cases, there is a lot of lint errors not related to my change. Is there a particular way to run the lint check?

./python/kfserving/kfserving/storage.py:27:16: W291 trailing whitespace
./python/kfserving/test/test_azure_storage.py:48:1: E302 expected 2 blank lines, found 1

@maganaluis there are two lint errors which should be an easy fix, thanks.

maganaluis:
I pushed some more changes, after the previous commit some test cases had to change.

yuzisun:
@maganaluis this looks good to me, are we able to add an e2e test for this? I am not sure how to verify this change.

maganaluis:
> @maganaluis this looks good to me, are we able to add an e2e test for this? I am not sure how to verify this change.

I can add a e2e test case, but I don't see e2e test cases for the other storage options, you would have to have a storage account for each cloud provider.

python/kfserving/kfserving/storage.py
```diff
@@ -168,16 +171,13 @@ def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
...
-        except Exception:  # pylint: disable=broad-except
-            token = Storage._get_azure_storage_token()
-            if token is None:
-                logging.warning("Azure credentials not found, retrying anonymous access")
-            block_blob_service = BlockBlobService(account_name=account_name, token_credential=token)
-            blobs = block_blob_service.list_blobs(container_name, prefix=prefix)
+        token = Storage._get_azure_storage_token()
+        if token is None:
+            logging.warning("Azure credentials not found, retrying anonymous access")
+        blob_service_client = BlobServiceClient(account_url, credential=token)
```
yuzisun, line 177: @maganaluis if the token is None, this should still work right? Does azure support public storage blobs?

yuzisun: oh I found the doc https://docs.microsoft.com/en-us/python/api/overview/azure/storage-blob-readme?view=azure-python#types-of-credentials

python/kfserving/kfserving/storage.py
```diff
@@ -192,7 +192,10 @@ def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
                     os.makedirs(dir_path)
 
             logging.info("Downloading: %s to %s", blob.name, dest_path)
-            block_blob_service.get_blob_to_path(container_name, blob.name, dest_path)
+            Path(os.path.dirname(dest_path)).mkdir(parents=True, exist_ok=True)
```
yuzisun, line 195: With this do we still need the above code from 183 to 192 to create the directories?

maganaluis: That's right, they were unnecessary, I removed those lines.

python/kfserving/kfserving/storage.py
maganaluis: Correct, however that's unlikely to be used. I did verify that Azure AD works. The current documentation still holds well.

yuzisun:
Thanks @maganaluis for your awesome work!

/lgtm
/approve

PR #1599, opened 2021-05-05
fix(python): regenerate python part based upon current master

NikeNano:
It seems like the python part currently is not regenerated after the latest changes, running: 
```
 ./hack/python-sdk/client-gen.sh
```
from master indicates that there are changes that are not updated. 

Fixes #

**Release note**:

```release-note

```


yuzisun:
@NikeNano Seems like there are a few test failures
https://github.com/kubeflow/kfserving/pull/1599/checks?check_run_id=2512705306

python/kfserving/test/test_v1alpha1_model_spec.py
```diff
@@ -51,12 +51,13 @@ def make_instance(self, include_optional):
         if include_optional :
             return V1alpha1ModelSpec(
                 framework = '0', 
-                memory = None,
```
NikeNano, line 54: This test breaks otherwise due to that memory is required.

python/kfserving/test/test_v1beta1_predictor_protocols.py
```diff
@@ -57,11 +57,13 @@ def make_instance(self, include_optional):
                     default_gpu_image_version="0",
                     default_image_version="0",
                     image="0",
+                    supported_frameworks=['0']
```
NikeNano, line 60: What is the logic behind the if statement here, as I understand the result is the same(no mater my current change) @yuzisun

NikeNano:
> @NikeNano Seems like there are a few test failures
> https://github.com/kubeflow/kfserving/pull/1599/checks?check_run_id=2512705306

Should be fixed now @yuzisun

yuzisun:
Thanks @NikeNano for this great work!

/lgtm
/approve

PR #1603, opened 2021-05-10
fix certificate unknown authority error using self-sign-ca scripts

haijohn:
After run self-signed-ca.sh instead using cert-manager and apply the sample InferenceService yaml, there is certificate-signed-by-unknown-authority error. this is because the self-signed-ca scripts misses some webhook patches for 0.6.0-rc0 release.

### DIFF
--- a/hack/self-signed-ca.sh
+++ b/hack/self-signed-ca.sh
@@ -117,9 +117,9 @@ echo "Encoded CA:"
 echo -e "${caBundle} \n"
 
 # Patch CA Certificate to webhooks
-mutatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/1/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
+mutatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/1/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/2/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
 mutatingPatchString=$(echo ${mutatingPatchString} | sed "s|{{CA_BUNDLE}}|${caBundle}|g")
-validatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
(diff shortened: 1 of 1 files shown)


yuzisun:
thanks @haijohn !

hack/self-signed-ca.sh
```diff
@@ -117,9 +117,9 @@ echo "Encoded CA:"
 echo -e "${caBundle} \n"
 
 # Patch CA Certificate to webhooks
-mutatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/1/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
+mutatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/1/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/2/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
```
yuzisun, line 120: looks like it was causing the issue after we added v1beta1, not sure if there is better way to do this as I can imagine it is going to be broken again as we are deprecating v1alpha2.

haijohn: maybe we can get webhooks in each configuration using kubectl and then build the patch string using webhook counts

yuzisun: right need some way to get the count of the webhooks

hack/self-signed-ca.sh
```diff
@@ -116,16 +116,32 @@ caBundle=$(cat ${tmpdir}/ca.crt | openssl enc -a -A)
 echo "Encoded CA:"
 echo -e "${caBundle} \n"
 
-# Patch CA Certificate to webhooks
-mutatingPatchString='[{"op": "replace", "path": "/webhooks/0/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}, {"op": "replace", "path": "/webhooks/1/clientConfig/caBundle", "value":"{{CA_BUNDLE}}"}]'
+# Patch CA Certificate to mutatingWebhook
+mutatingWebhookCount=$(kubectl get mutatingwebhookconfiguration | grep ${webhookConfigName} | awk '{print $2}')
```
yuzisun, line 120: how about using the following command to get the count?
`kubectl get mutatingwebhookconfiguration inferenceservice.serving.kubeflow.org -ojson | jq -r '.webhooks' | jq length`

haijohn: this is better, but if jq is not installed in the system, the scripts should exits or fallback using awk？

yuzisun: we can exit the script so user know it needs to install jq.

yuzisun:
Thanks @haijohn for the contribution ! Looks great!

/lgtm
/approve

PR #1609, opened 2021-05-12
relax .tar.gz files Content-Type check

Ruminateer:
When `storageUri` is a URI with a suffix of `.tar.gz` or `.tgz`, the current Kfserving requires `response.header.Content-Type=application/x-tar`. However, [IANA didn't define a MIME type for .tar.gz files](https://www.iana.org/assignments/media-types/media-types.xhtml), and many sites are using different `Content-Type`. Therefore, I opened the pull request to allow some other `Content-Type` for `.tar.gz` files.

Ref
- https://www.iana.org/assignments/media-types/media-types.xhtml
- https://en.wikipedia.org/wiki/List_of_archive_formats
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
- https://github.com/python/cpython/blob/3.9/Lib/mimetypes.py
- https://stackoverflow.com/questions/28396703/mime-type-for-tgz
- https://superuser.com/questions/901962/what-is-the-correct-mime-type-for-a-tar-gz-file

**Release note**:

```release-note
Do I need to write release notes?
```

### DIFF
--- a/pkg/agent/storage/https.go
+++ b/pkg/agent/storage/https.go
@@ -99,7 +99,8 @@ func (h *HTTPSDownloader) Download(client http.Client) error {
 		if err := extractZipFiles(resp.Body, fileDirectory); err != nil {
 			return err
 		}
-	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/gzip") {
+	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/x-gtar") ||
+		strings.Contains(contentType, "application/x-gzip") || strings.Contains(contentType, "application/gzip") {
 		if err := extractTarFiles(resp.Body, fileDirectory); err != nil {
 			return err
 		}
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -252,11 +252,11 @@ The path or model %s does not exist." % (uri))
             if mimetype == 'application/zip' and not response.headers.get('Content-Type', '')\
                     .startswith('application/zip'):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/zip\'" % uri)
+            tar_content_types = ('application/x-tar', 'application/x-gtar', 'application/x-gzip', 'application/gzip')
             if mimetype == 'application/x-tar' and not response.headers.get('Content-Type', '')\
-                    .startswith(('application/x-tar', 'application/x-gtar', 'application/x-gzip', 'application/gzip')):
-                raise RuntimeError("URI: %s did not respond with \'Content-Type\': "
-                                   "\'application/x-tar\', \'application/x-gtar\', "
-                                   "\'application/x-gzip\', or \'application/gzip\'" % uri)
+                    .startswith(tar_content_types):
+                raise RuntimeError("URI: %s did not respond with any of following \'Content-Type\': " % uri +
+                                   ", ".join(tar_content_types))
             if (mimetype != 'application/zip' and mimetype != 'application/x-tar') and \
                     not response.headers.get('Content-Type', '').startswith('application/octet-stream'):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/octet-stream\'"
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -253,7 +253,7 @@ The path or model %s does not exist." % (uri))
                     .startswith('application/zip'):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/zip\'" % uri)
             if mimetype == 'application/x-tar' and not response.headers.get('Content-Type', '')\
-                    .startswith('application/x-tar'):
+                    .startswith(('application/x-tar', 'application/x-gtar', 'application/x-gzip', 'application/gzip')):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/x-tar\'" % uri)
             if (mimetype != 'application/zip' and mimetype != 'application/x-tar') and \
                     not response.headers.get('Content-Type', '').startswith('application/octet-stream'):
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -254,7 +254,8 @@ The path or model %s does not exist." % (uri))
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/zip\'" % uri)
             if mimetype == 'application/x-tar' and not response.headers.get('Content-Type', '')\
                     .startswith(('application/x-tar', 'application/x-gtar', 'application/x-gzip', 'application/gzip')):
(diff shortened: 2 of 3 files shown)


Ruminateer:
@googlebot I signed it!

Ruminateer:
fixed flake8 errors

yuzisun:
@Ruminateer Could you help add a test [here](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/test/test_storage.py) ?

Ruminateer:
> @Ruminateer Could you help add a test [here](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/test/test_storage.py) ?

Added, hope how I did it looks good to you 🤨

python/kfserving/test/test_storage.py
```diff
@@ -94,61 +96,62 @@ def test_https_uri_path(_):
...
+    def test_fail(_):
+        with pytest.raises(expected_error):
+            kfserving.Storage.download(uri)
+
+    if expected_error is None:
+        return test_success
+    return test_fail
+
+
+test_http_uri_path_with_xtar = gen_download_tests(HTTPS_URI_TARGZ,
```
yuzisun, line 116: I think we can use `pytest.mark.parametrize` here for different combinations of HTTP_URI and HttpResponse ?

Ruminateer:
Used `pytest.mark.parametrize`.

Do you want me to use [it](https://github.com/Ruminateer/kfserving/blob/06f6dfcf38d75dbbf963f93685692e0ac8660223/python/kfserving/test/test_storage.py#L116-L130) for [these 2 tests](https://github.com/Ruminateer/kfserving/blob/06f6dfcf38d75dbbf963f93685692e0ac8660223/python/kfserving/test/test_storage.py#L79-L96)? I haven't done so, since technically they are not equivalent.

pkg/agent/storage/https.go
```diff
@@ -99,7 +99,8 @@ func (h *HTTPSDownloader) Download(client http.Client) error {
 		if err := extractZipFiles(resp.Body, fileDirectory); err != nil {
 			return err
 		}
-	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/gzip") {
+	} else if strings.Contains(contentType, "application/x-tar") || strings.Contains(contentType, "application/x-gtar") ||
+		strings.Contains(contentType, "application/x-gzip") || strings.Contains(contentType, "application/gzip") {
```
yuzisun, line 103: I think this condition can be simplified to 
```python
tar_content_types = ["application/x-tar", "application/x-gtar", "application/x-gzip"]
else if any(ct in contentType for ct in tar_content_types):
```

python/kfserving/kfserving/storage.py
```diff
@@ -253,8 +253,10 @@ def _download_from_uri(uri, out_dir=None):
                     .startswith('application/zip'):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/zip\'" % uri)
             if mimetype == 'application/x-tar' and not response.headers.get('Content-Type', '')\
-                    .startswith('application/x-tar'):
-                raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/x-tar\'" % uri)
+                    .startswith(('application/x-tar', 'application/x-gtar', 'application/x-gzip', 'application/gzip')):
```
yuzisun, line 256: prefer using a variable for the content type list so we do not have to repeat in a lot of the places

python/kfserving/kfserving/storage.py
```diff
@@ -253,8 +253,10 @@ def _download_from_uri(uri, out_dir=None):
...
+                raise RuntimeError("URI: %s did not respond with \'Content-Type\': "
+                                   "\'application/x-tar\', \'application/x-gtar\', "
```
yuzisun, line 258: same here we can use the content type list variable

python/kfserving/test/test_storage.py
```diff
@@ -94,61 +96,38 @@ def test_https_uri_path(_):
...
+    (HTTPS_URI_TARGZ, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/x-gzip'), None),
+    (HTTPS_URI_TARGZ, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/gzip'), None),
+    (HTTPS_URI_TARGZ_WITH_QUERY, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/x-tar'), None),
+    (HTTPS_URI_TARGZ_WITH_QUERY, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/x-gtar'), None),
+    (HTTPS_URI_TARGZ_WITH_QUERY, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/x-gzip'), None),
+    (HTTPS_URI_TARGZ_WITH_QUERY, MockHttpResponse(200, FILE_TAR_GZ_RAW, 'application/gzip'), None),
+    ('https://foo.bar/model.zip', MockHttpResponse(200, FILE_ZIP_RAW, 'application/zip'), None),
+    ('https://foo.bar/model.zip?foo=bar', MockHttpResponse(200, FILE_ZIP_RAW, 'application/zip'), None),
+    ('https://theabyss.net/model.joblib', MockHttpResponse(404), RuntimeError),
+    ('https://some.site.com/test.model', MockHttpResponse(status_code=200, content_type='text/html'), RuntimeError),
```
yuzisun, line 111: If the response is 200, the exception should be None right?

python/kfserving/test/test_storage.py
```diff
@@ -94,61 +96,38 @@ def test_https_uri_path(_):
...
+    ('https://foo.bar/test/', MockHttpResponse(200), ValueError),
```
yuzisun, line 112: same here

python/kfserving/test/test_storage.py
Ruminateer: These are not new tests. These are equivalent to the [original tests](https://github.com/kubeflow/kfserving/blob/a61d146b5c46706b9e9be85bf6b169262a122c95/python/kfserving/test/test_storage.py#L140-L151). The original tests expect the code to throw exceptions on bad `content_type`/bad URL even with a 200.

python/kfserving/test/test_storage.py
Ruminateer: same as above

pkg/agent/storage/https.go
Ruminateer: I don't think Go supports such syntax...

yuzisun: oops, sorry was thinking about storage initializer..

yuzisun:
Thanks @Ruminateer !

/lgtm
/approve

haijohn:
can you also extend  **application/zip**  to **[application/zip, application/x-zip-compressed, application/zip-compressed]**
Becase I encountered an error when download a zip file using http, and I found out the actual Content-Type is **application/x-zip-compressed** instead of **application/zip**. Please see this doc https://mimetype.io/application/zip

vdanyliv:
@Ruminateer is there any possibility to unzip from S3?

PR #1322, opened 2021-01-27
Add E2E tests for MMS with the V2 protocol

adriangonz:
**What this PR does / why we need it**:

Add integration tests for MMS when using the V2 protocol with SKLearn and XGBoost models.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

There are some changes due to `black`'s autoformatting rules. 

**Release note**:

```release-note
NONE
```

PR #1623, opened 2021-05-23
Update MMS docs for 0.6 release

### LINKED ISSUE #1621: Several Issues with the KFServing MMS Triton README
/kind bug

Following the steps in the README is not a seamless experience. It seems like several commands/instructions are outdated.

1) README mentions “ You must set Triton's multiModelServer flag in inferenceservice.yaml to true to enable multi-model serving for Triton” -- but MMS only worked for me when the multimodelserver flag was set in the kfserving.yaml file used in installation (eg. install/v0.5.1/kfserving.yaml).

2) The README mentions “Setup Minio for storing the models as currently model agent sidecar only supports S3 protocol.”, but it looks like gs:// protocol is also supported now as I was able to specify model location in a GCS bucket.

3) In minio.yaml, the apiVersion is specified as extensions/v1beta1. However, that gives an error:
"**error: unable to recognize "minio.yaml": no matches for kind "Deployment" in version "extensions/v1beta1**"

From https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ :
“Deployment in the extensions/v1beta1, apps/v1beta1, and apps/v1beta2 API versions is no longer served. Migrate to use the apps/v1 API version, available since v1.9.”
So, in minio.yaml the apiVersion should be updated 
`apiVersion: extensions/v1beta1 → apiVersion: apps/v1`

4) The README mentions “**Setup Minio for storing the models as currently model agent sidecar only supports S3 protocol.**”, but it looks like gs:// protocol is also supported now as I was able to specify model location in a gcs bucket

5) Copying cifar10 model to minio gives the following error:
`mc cp -r cifar10 myminio/triton/torchscript`
**mc: <ERROR> Failed to copy `cifar10/1/model.pt`. Bucket `triton` does not exist.**
**mc: <ERROR> Failed to copy `cifar10/config.pbtxt`. Bucket `triton` does not exist.**
We may need to create a Triton bucket before copying

6) The command to copy the simple_string model to mc is missing the destination argument.
`gsutil cp -r gs://kfserving-samples/models/triton/simple_string`

7) It looks like the model “gs://kfserving-samples/models/triton/simple_string” has moved, or does not exist at that path.
There is a simple_string model at "gs://kfserving-samples/models/tensorrt/simple_string/” -- is this the one to use in this example?

8) In the section to Deploy the "Cifar10 Torchscript Model", the command provided is the one to start the InferenceService, instead of specifying a TrainedModel:
`kubectl apply -f multi_model_triton.yaml` 
should be
`kubectl apply -f trained_model.yaml`

9) For the simple_string Tensorflow model, the framework specified in it's trained_model2.yaml is "pytorch"

Other Suggestions:
1) It would be great to indicate that the following command should be run from another terminal.
**kubectl port-forward $(kubectl get pod --selector="app=minio" --output jsonpath='{.items[0].metadata.name}') 9000:9000**

2) To make the example simpler, is it possible to directly use gs:// instead of minio s3?

The list of steps provided for the examples should execute without any errors to provide a seamless experience to new users.

**Environment:**
- Istio Version: 1.9.0
- Knative Version: 0.18.0
- KFServing Version: 0.5.1
- Kubeflow version: -
- Minikube version: 1.17.1
- Kubernetes version: (use `kubectl version`): Client: v1.20.4, Server: v1.20.2
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.2 LTS (Focal Fossa)

yuzisun:

Fixes #1621 

1. Turn on MMS flags for triton, sklearn/xgboost model server
2. Link MMS doc on feature map
3. Update outdated triton mms example

**Release note**:

```release-note

```


config/configmap/inferenceservice.yaml
```diff
@@ -31,7 +31,7 @@ data:
             "supportedFrameworks": [
               "sklearn"
             ],
-            "multiModelServer": false
+            "multiModelServer": true
```
Aaronchoo, line 34: Should sklearn v2 `multiModelServer` flag be true as well?

docs/samples/README.md
```diff
@@ -72,6 +72,16 @@ Additionally KFServing also integrates with The [AI Explainability 360 (AIX360)]
 | Deploy Alibi Text Explainer| [Alibi Text Explainer](./explanation/alibi/moviesentiment) |
 | Deploy AIX360 Image Explainer| [AIX360 Image Explainer](./explanation/aix/mnist/README.md) |
 
+### Deploy InferenceService with Multiple Models(Alpha)
+Multi Model Serving allows deploying `TrainedModels` at scale without bounded by the Kubernetes service/pod limits and reducing the TCO,
```
Aaronchoo, line 76: without `being` bounded?

config/configmap/inferenceservice.yaml
yuzisun: There is no doc and e2e test currently for v2 MLServer mms, @adriangonz What's the current mms status for MLServer?

docs/samples/README.md
```diff
@@ -72,6 +72,16 @@ Additionally KFServing also integrates with The [AI Explainability 360 (AIX360)]
...
 
+### Deploy InferenceService with Multiple Models(Alpha)
+Multi Model Serving allows deploying `TrainedModels` at scale without being bounded by the Kubernetes compute resources(CPU/GPU/Memory), 
+service/pod limits and reducing the TCO, see [Multi Model Serving](../MULTIMODELSERVING_GUIDE.md) for more details. 
+Multi Model Serving is supported for Triton, SKLearn/XGBoost as well as Custom KFServer.
+ 
+| Features  | Examples |
+| ------------- | ------------- |
+| Deploy multiple models with Triton Inference Server| [Multi Model Triton InferenceService](./mutlimodelserving/triton/README.md)  |
+| Deploy multiple models with SKLearn/XGBoost KFServer| [Multi Model SKLearn InferenceService](./mutlimodelserving/sklearn/README.md)  |
```
Aaronchoo, line 83: Should be `./multimodelserving/` -- I believe there is a typo and the `l` and `t` need to be switched.

config/configmap/inferenceservice.yaml
adriangonz: There should be e2e tests for both v1 and v2 in this file (they got introduced in #1322): 

https://github.com/kubeflow/kfserving/blob/master/test/e2e/predictor/test_multi_model_serving.py

Regarding the docs, that's a good question. I remember adding an example for the individual models, but I don't think we got around adding an example with MMS. What would be the best place to add these? Should we open this as a separate issue?

yuzisun: @adriangonz Nice ! Can you help add the examples [here](https://github.com/kubeflow/kfserving/tree/master/docs/samples/multimodelserving/sklearn), move the current one to v1 folder and added yours in v2?

docs/samples/multimodelserving/triton/README.md
```diff
@@ -88,37 +76,66 @@ spec:
...
+  - lastTransitionTime: "2021-05-22T20:58:56Z"
+    status: "True"
+    type: Ready
+  url: http://triton-mms.default.example.com/v2/models/cifar10/infer
+```
+
+Now you can curl the model metadata endpoint
 ```bash
 MODEL_NAME=cifar10
 INPUT_PATH=@./input.json
```
shashank3959, line 141: Does this input.json data file exist in this location?

yuzisun: @shotarok I think here it is only calling the metadata endpoint so this input is not needed and I can remove it, I already included this input in the perf test job.

yuzliu:
Awesome! MMS it getting more and more official now!

yuzliu (approved):

theofpa (commented):
/lgtm

PR #1637, opened 2021-05-30
Parallel inference support

### LINKED ISSUE #1074: Investigate how to achieve parallelism in KFServing python model server
/kind feature

As the KFServing WG is implementing the load/unload endpoints in the SKLearn and XGBoost model servers to support multi-model serving, I'd like to start a conversion about how to achieve parallelism. 
The goals are:
1)The Python model server can do load(IO bound) and prediction(CPU bound) at the same time for SKLearn and XGBoost models
2)The Python model server can run multiple predictions at the same time for SKLearn and XGBoost 

Our may concerns are:
1. Python is designed with the assumption that each computer only has one core, so it placed a GIL(global interpreter lock) which does not allow multiple threads running in parallel
2. The asyncio based concurrency solutions (fastapi and tornado) use multiple coroutines to achieve concurrency but they share the same thread.  If one prediction coroutine takes a long time, the IO coroutine will be blocked. In addition to that, it does not allow multiple predictions running in parallel.

Some work items to go from here:
1. Investigate a way to share memory between processes and use multiple processes in the model server to achieve parallelism
2. Collect data to understand how slow model server can be if we load hundreds to thousands models into it and share the same process.

yuzisun:
# Parallel Inference
The key feature here is to use RayServe [Python ServeHandle APIs](https://docs.ray.io/en/master/serve/package-ref.html#servehandle-apil) to scale up the model inference and offload the heavy computation to RayServe. 
- This allows keeping using the existing tordano webserver and KFModel/KFServer APIs, all user needs to do is to add the RayServe annotations to configure the remote Ray Actor and replicas and everything else happens seamlessly.
- For multi model serving this allows creating Ray Actor per model, so inference for different models can be done in parallel.

Architecture Review:
Screen Shot 2021-06-01 at 9 39 02 AM

RAY Dashboard
Screen Shot 2021-06-01 at 9 42 12 AM

Fixes #1074 

**Release note**:

```release-note

```


python/kfserving/requirements.txt
```diff
@@ -16,3 +16,5 @@ cloudevents>=1.2.0
 avro>=1.10.1
 boto3>=1.17.32
 botocore>=1.20.32
+ray[serve]>=1.3.0
```
simon-mo, line 19: 1.4.0 was just released so feel free to bump the version :-)

python/kfserving/kfserving/kfmodel_repository.py
```diff
@@ -31,19 +32,28 @@ def __init__(self, models_dir: str = MODEL_MOUNT_DIRS):
...
 
     def is_model_ready(self, name: str):
         model = self.get_model(name)
-        return False if model is None else model.ready
+        if not model:
+            return False
+        if isinstance(model, KFModel):
+            return False if model is None else model.ready
+        else:
+            # TODO find out how to get the remote model status
```
simon-mo, line 48: For Ray Serve, the models are guaranteed to be ready after the deploy call. So `ModelClass.deploy()` itself is blocking and it will wait for `num_replicas` models to be ready. You can use `_wait_for_goal` API to make it async:
```
from serve.api import _get_global_client

client = serve.start()
goal_id = Model.deploy(_blocking=False)
is_ready: bool = client._wait_for_goal(goal_id=goal_id, timeout=1) # This timeout is configurable, and it will return True or False indicating whether or not the goal has been completed. It can even be 0.
```

simon-mo (commented):
This is awesome! @yuzisun, please let me know if there any feedback or questions around the Ray Serve API and what can we do to help accelerate this.

yuzisun:
@Ruminateer I am also getting this error now
```
E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"admission webhook \"inferenceservice.kfserving-webhook-serv
er.v1beta1.validator\" denied the request: Exactly one of [SKLearn, XGBoost, Tensorflow, PyTorch, Triton, ONNX, PMML, LightGBM, PodSpec] must be specified in PredictorSpe
c","reason":"Exactly one of [SKLearn, XGBoost, Tensorflow, PyTorch, Triton, ONNX, PMML, LightGBM, PodSpec] must be specified in PredictorSpec","code":403}
```

python/kfserving/requirements.txt
yuzisun: @simon-mo Looks like 1.4.0 has some breaking change regarding namespaces, so I pinned to 1.3.0 for now

simon-mo: Hi @yuzisun, you can specify a namespace inside ray.init(..., namespace="..."). See 
https://ray--16786.org.readthedocs.build/en/16786/serve/deployment.html#id3

cc @edoakes

maganaluis:
@yuzisun This is pretty cool, I have a question in how replicas can share models or embeddings, pre-train transformers, etc ...  As you imagine there might be cases where models are quite large, so it would be beneficial to show an example how to share a model between replicas. It seems that a good way would be to create an Actor for the model itself?

```python
import numpy as np
import ray
import requests
from ray import serve

serve.start()


@ray.remote
class Fortune(object):
    def __init__(self):
        self.n = np.random.randint(0, 9)

    def num(self):
        return self.n


fortune_actor = Fortune.remote()


@serve.deployment(num_replicas=2)
def hello(request):
    name = request.query_params["name"]
    return f"Hello {name}! {ray.get(fortune_actor.num.remote())}"


hello.deploy()

# Query our endpoint over HTTP.
response = requests.get("http://127.0.0.1:8000/hello?name=serve").text
print(response)
# Query our endpoint over HTTP.
response = requests.get("http://127.0.0.1:8000/hello?name=serve").text
print(response)
```

PR #1673, opened 2021-06-20
extend zip content type check when download zip use uri

haijohn:
application/zip may also appear as application/x-zip-compressed, application/zip-compressed, this pr adds these two content type to content type check when download zip files from uri.


python/kfserving/kfserving/storage.py
```diff
@@ -268,8 +268,9 @@ def _download_from_uri(uri, out_dir=None):
         with requests.get(uri, stream=True, headers=headers) as response:
             if response.status_code != 200:
                 raise RuntimeError("URI: %s returned a %s response code." % (uri, response.status_code))
+            zip_content_types = ('application/x-zip-compressed', 'application/zip', 'application/zip-compressed')
             if mimetype == 'application/zip' and not response.headers.get('Content-Type', '')\
-                    .startswith('application/zip'):
+                    .startswith(zip_content_types):
                 raise RuntimeError("URI: %s did not respond with \'Content-Type\': \'application/zip\'" % uri)
```
yuzisun, line 274: Can you help fix the error message with the variable `zip_content_types`?

haijohn: Fixed

yuzisun:
thanks @haijohn !

/lgtm
/approve

PR #1681, opened 2021-06-22
Remove "default" suffix from generated component name

### LINKED ISSUE #1397: Reason for still using "default" in revision name?
In kfserving v0.4 we have primary route with name `{isvc_name}-predictor-default` and canary route with name `{isvc_name}-predictor-canary`, which makes sense to distinguish them.

However in v0.5, from what I see [here](https://github.com/kubeflow/kfserving/tree/master/docs/samples/v1beta1/torchserve/canary#check-the-traffic-split-between-the-two-revisions), we are actually using DNS label as `latest-{isvc_name}-predictor-default` indicating canary, and `prev-{isvc_name}-predictor-default` indicating primary. Do we really still need the suffix `-default` here given it seems redundant (we are using `default` for both primary and canary)?

The reason for asking this is I'm a bit worried since DNS label has a maximum character limitation of 63, including such redundant information will reduce the effective `isvc_name` we can use, especially in the case that e.g., we are using knative domain template as `{{.Name}}-{{.Namespace}}.{{.Domain}}` for HTTPS wildcard certificates, with the new v0.5 the DNS label would be something like `latest-{isvc_name}-{namespace}-predictor-default`, only leaving 37 characters for `isvc_name` and `namespace`, and with a long `namespace` there's essentially not much space left for `isvc_name`.

Haven't tested yet (need some time to upgrade to v0.5) and maybe I'm missing something like maybe the URL in `ksvc` is not actually used for routing? Or is there any way we can customize our installation to reduce the DNS label length (such as update the suffix `-predictor-default` to `-pred-df`)? Thanks for any help/advice!

yuzisun:

Fixes #1397 

1. Default name suffix was left from legacy reason when we were creating default/canary pair in v1alpha2, we should remove it now as it adds to the dns name limits
2. In order to to backwards compatible for existing deployed service, we do not change the generated if the controller detects that the service exists.

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/serving/v1beta1/component.go
+++ b/pkg/apis/serving/v1beta1/component.go
@@ -55,7 +55,7 @@ var (
 type ComponentImplementation interface {
 	Default(config *InferenceServicesConfig)
 	Validate() error
-	GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container
+	GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig, predictorHost ...string) *v1.Container
 	GetStorageUri() *string
 	GetProtocol() constants.InferenceServiceProtocol
 	IsMMS(config *InferenceServicesConfig) bool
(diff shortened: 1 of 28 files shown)


Peilun-Li:
Hey just checking any plan to have this released sooner? We are exploring this [tag header based routing](https://github.com/kserve/kserve/tree/master/docs/samples/v1beta1/rollout#tag-based-routing) feature and it looks like with that feature enabled, we are adding `latest-` prefix making the DNS 63 character length restriction even tighter. Having the `-default` suffix removed will help a lot there. Thanks!

yuzisun:
close for https://github.com/kserve/kserve/pull/2508

PR #1682, opened 2021-06-22
Append "--no-cache-dir" to "pip install" in Dockerfiles

xuhdev:
See the rationale at https://github.com/hadolint/hadolint/wiki/DL3042

In addition to space saving, it also ensures that pip doesn't install
from local cache (especially in case of a rebuild) and reduces the
chance of creating bugs during development.

**What this PR does / why we need it**: See above

**Special notes for your reviewer**: It's difficult for me to test every image, but they shouldn't make functional changes to the built Docker images for clean build.

**Release note**:

```release-note
Use `pip install --no-cache-dir` instead of naked `pip install` in Dockerfiles.
```

### DIFF
--- a/docs/samples/explanation/aix/mnist/rfserver/rf.Dockerfile
+++ b/docs/samples/explanation/aix/mnist/rfserver/rf.Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.7
 
 COPY . .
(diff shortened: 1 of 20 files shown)


xuhdev:
Any chance to review this? :)

yuzisun:
Thanks @xuhdev !

/lgtm
/approve

PR #1696, opened 2021-07-01
sklearnserver: add safeguard if multiple model files are present

xuhdev:
Similar to the safeguard that exists in pytorchserver.

**What this PR does / why we need it**: It'll be more robust and better guard against inadvertent error (which may take a long time to debug out)

**Release note**:

```release-note
sklearnserver now errors out when multiple model files or no model file are present
```

### DIFF
--- a/python/sklearnserver/sklearnserver/model.py
+++ b/python/sklearnserver/sklearnserver/model.py
@@ -15,7 +15,7 @@
 import kfserving
 import joblib
 import numpy as np
-import os
+import pathlib
 from typing import Dict
 
 MODEL_BASENAME = "model"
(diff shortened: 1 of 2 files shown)


pugangxa:
/lgtm
Seems for every model with similar issue we need this?

xuhdev:
I would agree, if the model script scans for multiple files.

xuhdev:
The error doesn't seem to relate to this PR

xuhdev:
@pugangxa 
I've rebased (no other changes) and all tests have passed. Could you approve again? Thanks!

yuzisun:
@xuhdev @pugangxa There could be use cases where for example KFP saves the multiple artifact files(model metadata etc) in the same model storage directory, so not sure if we want to restrict to only one file.

xuhdev:
@yuzisun If we allow multiple files, then we need to develop a principled way to select a file.

yuzisun:
> > @xuhdev @pugangxa There could be use cases where for example KFP saves the multiple artifact files(model metadata etc) in the same model storage directory, so not sure if we want to restrict to only one file.
> 
> Yes, I think there can be multiple other files but only 1 model?

ye validating only one model file is fine.

yuzisun:
Thanks @xuhdev !
/lgtm
/approve

PR #1715, opened 2021-07-09
Use index instead of element when searching for a volume

xuhdev:
`range` copies the elements by value. Avoid using it when searching for
volumes. This should potentially improve overall performance when
`AppendVolumeIfNotExists` is used.

**What this PR does / why we need it**: above

**Release note**:

```release-note
Performance improvements for AppendVolumeIfNotExists.
```


yuzisun:
thanks @xuhdev !

/lgtm
/approve

PR #1721, opened 2021-07-14
Make canary name match tensorflow.yaml

markwinter:

The canary should have the same name as the base tensorflow.yaml so it correctly applies as a canary


yuzisun:
Thanks @markwinter !

PR #1750, opened 2021-08-05
Azure Blob Upgrade Bug Fix

### LINKED ISSUE #923: azure-storage-blob package update to 12.x
/kind feature

**Update to the current azure blob service to 12.x**
Update the current package from the legacy 2.1 version to the current 12.x series.

Hi, what would need to be done to do the upgrade? Is it only the files in the python kfserving package?

maganaluis:

This PR addresses a bug which was introduced when we upgraded to Azure Blob 12.1.X or greater. Previously we thought that the container_client will filter blobs with a prefix but that's not the case. Per the [documentation](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-list?tabs=python#use-a-hierarchical-listing) you need to implement your own logic. 

This PR introduces the use of a stack to iterate over the blob directories trees, and ensures blobs are saved without the prefix so it can work with inference services which expect certain artifacts to be in a standard format.

Typically you'll see storageUris which are structured like these: 
```
storageUri: https://<account>.blob.core.windows.net/private/models/iris-example/
storageUri: https://<account>.blob.core.windows.net/private/models/ner-tensorflow/1/
```

The prefix will be anything after the container name, the container name in this case is ```private```. Furthermore the artifacts would need to be written to the ```/mnt/models/``` directory without the prefix to ensure that the predictors can read the artifacts in the expected format.

Fixes #923 

This should be addressed and merged prior the next release.

**Release note**:

```release-note

```

### DIFF
--- a/python/kfserving/kfserving/storage.py
+++ b/python/kfserving/kfserving/storage.py
@@ -27,6 +27,7 @@ from urllib.parse import urlparse
 import requests
 from pathlib import Path
 from azure.storage.blob import BlobServiceClient
+from azure.storage.blob._list_blobs_helper import BlobPrefix
 
 from botocore.client import Config
 from botocore import UNSIGNED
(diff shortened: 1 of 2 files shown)


yuzisun:
@maganaluis Let's definitely add an e2e test for this, I am not sure how to verify this change.

maganaluis:
> @maganaluis Let's definitely add an e2e test for this, I am not sure how to verify this change.

Unless we have an azure account, I think we could add more unit tests, which cover expected usage.

yuzisun:
> @yuzisun Dan is there anyway to move this forward? Can we add a end 2 end test case ourselves or is it perhaps better to see if we can engage someone from Azure?

@maganaluis any suggestion how we can add an e2e test for this?

yuzisun:
> > @yuzisun Dan is there anyway to move this forward? Can we add a end 2 end test case ourselves or is it perhaps better to see if we can engage someone from Azure?
> 
> @maganaluis any suggestion how we can add an e2e test for this? we'd like to include this in kserve 0.7 release

maganaluis:
@yuzisun I think the only way I see we can test this is if we create a public blob that we can load, and test that the code is able to extract it based on the prefix.

We can use a custom predictor, and load a text-file with a string, then print the content of the string. This is to avoid you guys having to maintain a model down the road.

Does KFServing has access to an Azure account? Is this something you guys need to add it to the end 2 end test cases?

CC @marcjimz

yuzisun:
@maganaluis Can we help resolve the conflicts as the project is now renamed to kserve, thanks.

PR #1755, opened 2021-08-09
Ignore istio dir downloaded by hack install

markwinter:

Add the istio directory which is downloaded by `quick_install.sh` to the ignore list


theofpa:
Thank you @markwinter !
/lgtm

PR #1758, opened 2021-08-10
add default deployment mode

Iamlovingit:
# deploy.defaultDeploymentMode
Add **deploy.defaultDeploymentMode** in **inferenceservice-config configmap**. **defaultDeploymentMode** now supports two modes: "Serverless" and "RawDeployment", by default, **defaultDeploymentMode** is "Serverless".
If user wants to change the deployment mode, there are two ways:
* 1. modify the **inferenceservice-config** config map, change the **defaultDeploymentMode**
```
deploy: |-
    {
    }
```
* 2. add annotation to inference service deploy yaml.
```
kind: "InferenceService"
metadata:
  name: "flower-sample"
  annotations:
    "serving.kubeflow.org/deploymentMode" : "RawDeployment"
spec:
  predictor:
    tensorflow:
      storageUri: "gs://kfserving-samples/models/tensorflow/flowers"
```


pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -88,10 +88,26 @@ func (e *Explainer) Reconcile(isvc *v1beta1.InferenceService) error {
...
+
 	podSpec := v1.PodSpec(isvc.Spec.Explainer.PodSpec)
 
 	// Here we allow switch between knative and vanilla deployment
-	if value, ok := annotations[constants.RawDeploymentAnnotationKey]; ok && value == "true" {
+	/*
+		add an item in inferenceservice configmap, called deploy.
+		case 1:
+		if deploy.DefaultDeploymentMode is Serverless, that means the default deployment mode is serverless(or knative),
+		if you want raw deployment, add an annotation "SwitchDeploymentMode" with value "true"
```
yuzisun, line 103: I think `switchDeploymentMode` might not be clear for what mode you are deploying, how about just be explicit for the `DeploymentMode`, if it is `serverless` then deploy knave otherwise deploy `raw`. If knative is not installed then `serverless` is not supported.

Iamlovingit: OK

yuzisun:
moved to https://github.com/kserve/kserve/pull/57

Issue #1766, opened 2021-08-12
storage-initializer: support AWS_CA_BUNDLE 


elukey:
/kind feature

**Describe the solution you'd like**
While following [this tutorial](https://github.com/kubeflow/kfserving/blob/master/docs/samples/storage/s3/README.md#create-s3-secret-and-attach-to-service-account) about how to configure the storage-initializer to fetch from an S3 endpoint I found it difficult to find how to add a custom value to the `AWS_CA_BUNDLE` environment variable. The main motivation for me is to avoid TLS validation issues while trying to contact an internal S3 endpoint (we have Openstack Swift exposing an S3 API internally, that uses a TLS certificate signed by our internal CA). I tried first via standard container ENV variables but due to how the storage-initializer is configured (via a ConfigMap IIUC) I didn't find a way to do it easily, so I had to add it to the Docker image itself (we have our own images in our internal registry, very close to official kfserving ones). This worked nicely but it would be way better if there was a more friendly way, maybe adding an annotation to [s3_secret.go](https://github.com/kubeflow/kfserving/blob/master/pkg/credentials/s3/s3_secret.go)? Something like `s3-ca-bundle` or similar, that in turn adds the AWS_CA_BUNDLE environment variable to storage-initializer.

If there is another way please let me know!

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

DontStopMeNow:
@elukey Hi, did you try setting `serving.kubeflow.org/s3-verifyssl: "0"` in your secrets? Or do you have to verify the connection?

elukey:
@DontStopMeNow I tried as test (and it didn't work), but yes I have to verify the connection :)

umka1332:
For anyone, who will try to use
`    serving.kubeflow.org/s3-verifyssl: "0"`
It doesn't work, but this helps me to disable SSL verification - just set serving.kserve.io/s3-cabundle to an empty string in the same secret annotation:
`    serving.kserve.io/s3-cabundle: ""`

Issue #1765, opened 2021-08-12
S3 storage-initializer, secret/annotations and AWS_REGION


elukey:
/kind bug

**What steps did you take and what happened:**
Hi everybody! I am trying to deploy kfserving 0.6.0 standalone with the storage-initializer to fetch models from a local Swift cluster (exposing an S3 API compatible with v4 signature). I followed the [tutorial](https://github.com/kubeflow/kfserving/blob/master/docs/samples/storage/s3/README.md) and added the annotation for the `s3-region` as described (in my case, it needs to be `US` to avoid signature validation failures when contacting Swift). I checked after the k8s deploy that they were applied correctly to the container, and I found the following (as expected):

```
Init Containers:
  storage-initializer:
    [..]
    Limits:
      cpu:     1
      memory:  1Gi
    Requests:
      cpu:     100m
      memory:  100Mi
    Environment:
      AWS_ACCESS_KEY_ID:      <set to the key 'AWS_ACCESS_KEY_ID' in secret 'test-secret'>      Optional: false
      AWS_SECRET_ACCESS_KEY:  <set to the key 'AWS_SECRET_ACCESS_KEY' in secret 'test-secret'>  Optional: false
      S3_USE_HTTPS:           1
      S3_ENDPOINT:            thanos-swift.discovery.wmnet
      AWS_ENDPOINT_URL:       https://thanos-swift.discovery.wmnet
      AWS_REGION:             US
    [..]
```
But the storage-initializer container still failed to fetch the model from Swift, due to the following error:

```
botocore.exceptions.ClientError: An error occurred (SignatureDoesNotMatch) when calling the ListObjects operation: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```
The above can be easily reproduced via a python-venv with boto3, testing something like (of course in this case it is my internal endpoint, to repro one needs to use another one like AWS etc..):
```
import boto3

# Retrieve the list of existing buckets
s3 = boto3.client('s3', aws_access_key_id="XXXXXXX", 
                  aws_secret_access_key="XXXXXXXX",
                  endpoint_url="https://thanos-swift.discovery.wmnet")
response = s3.list_buckets()
```
The above doesn't work if I export `AWS_REGION="US"`, but it does if I export `AWS_DEFAULT_REGION="US"`. This seems to be supported by https://github.com/boto/boto3/issues/2574 (if you check the attached pull request the code mentions the AWS_DEFAULT_REGION env variable only).

[...]

Issue #44, opened 2019-04-20
Race condition on kfservice_controller_test


ellis-bigelow:
```
➜  kfserving git:(admission) ✗ make docker-build
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
CRD manifests generated under '/Users/ellisbigelow/go/src/github.com/kubeflow/kfserving/config/crds'
RBAC manifests generated under '/Users/ellisbigelow/go/src/github.com/kubeflow/kfserving/config/rbac'
go test ./pkg/... ./cmd/... -coverprofile cover.out
?   	github.com/kubeflow/kfserving/pkg/apis	[no test files]
?   	github.com/kubeflow/kfserving/pkg/apis/serving	[no test files]
ok  	github.com/kubeflow/kfserving/pkg/apis/serving/v1alpha1	8.408s	coverage: 42.5% of statements
?   	github.com/kubeflow/kfserving/pkg/controller	[no test files]
--- FAIL: TestReconcile (0.29s)
    kfservice_controller_test.go:185:
        Unexpected error:
            <*errors.StatusError | 0xc0002d8b40>: {
                ErrStatus: {
                    TypeMeta: {Kind: "", APIVersion: ""},
                    ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
                    Status: "Failure",
                    Message: "Operation cannot be fulfilled on services.serving.knative.dev \"foo\": the object has been modified; please apply your changes to the latest version and try again",
                    Reason: "Conflict",
                    Details: {
                        Name: "foo",
                        Group: "serving.knative.dev",
                        Kind: "services",
                        UID: "",
                        Causes: nil,
                        RetryAfterSeconds: 0,
                    },
                    Code: 409,
                },
            }
            Operation cannot be fulfilled on services.serving.knative.dev "foo": the object has been modified; please apply your changes to the latest version and try again
        occurred
FAIL
coverage: 72.0% of statements
FAIL	github.com/kubeflow/kfserving/pkg/controller/kfservice	7.865s
?   	github.com/kubeflow/kfserving/pkg/frameworks/tensorflow	[no test files]
ok  	github.com/kubeflow/kfserving/pkg/reconciler/ksvc	7.927s	coverage: 81.8% of statements
[...]

PR #1780, opened 2021-08-25
Fix AWS_REGION and add AWS_CA_BUNDLE to s3_secrets.go

elukey:
The boto3 package seems to look for AWS_DEFAULT_REGION, meanwhile
kfserving uses AWS_REGION, that doesn't seem to be working
(tested following
https://github.com/kubeflow/kfserving/blob/master/docs/samples/storage/s3/README.md).

This change doesn't replace all occurrences of AWS_REGION, but only
the ones that set the environment variables.

GH-1765

Add also the possibility to force boto to use a custom CA Bundle path when validating the TLS certificate of the S3 endpoint. Our use case is related to using an internal S3 endpoint (based on Openstack Swift) offering a TLS certificate signed by our internal CA.

GH-1766

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/credentials/s3/s3_secret.go
+++ b/pkg/credentials/s3/s3_secret.go
@@ -27,7 +27,7 @@ const (
 	AWSAccessKeyIdName     = "awsAccessKeyID"
 	AWSSecretAccessKeyName = "awsSecretAccessKey"
 	AWSEndpointUrl         = "AWS_ENDPOINT_URL"
-	AWSRegion              = "AWS_REGION"
+	AWSRegion              = "AWS_DEFAULT_REGION"
 	S3Endpoint             = "S3_ENDPOINT"
 	S3UseHttps             = "S3_USE_HTTPS"
 	S3VerifySSL            = "S3_VERIFY_SSL"
--- a/pkg/credentials/s3/s3_secret.go
+++ b/pkg/credentials/s3/s3_secret.go
@@ -21,6 +21,11 @@ import (
 	"k8s.io/api/core/v1"
 )
 
+/*
+For a quick reference about AWS ENV variables:
+AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
+Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables
+*/
 const (
 	AWSAccessKeyId         = "AWS_ACCESS_KEY_ID"
 	AWSSecretAccessKey     = "AWS_SECRET_ACCESS_KEY"
@@ -33,6 +38,7 @@ const (
 	S3VerifySSL            = "S3_VERIFY_SSL"
 	S3UseVirtualBucket     = "S3_USER_VIRTUAL_BUCKET"
 	AWSAnonymousCredential = "awsAnonymousCredential"
+	AWSCABundle            = "AWS_CA_BUNDLE"
 )
 
 type S3Config struct {
@@ -49,6 +55,7 @@ var (
 	InferenceServiceS3SecretHttpsAnnotation      = constants.KFServingAPIGroupName + "/" + "s3-usehttps"
 	InferenceServiceS3UseVirtualBucketAnnotation = constants.KFServingAPIGroupName + "/" + "s3-usevirtualbucket"
 	InferenceServiceS3UseAnonymousCredential     = constants.KFServingAPIGroupName + "/" + "s3-useanoncredential"
+	InferenceServiceS3CABundleAnnotation         = constants.KFServingAPIGroupName + "/" + "s3-cabundle"
 )
 
 func BuildSecretEnvs(secret *v1.Secret, s3Config *S3Config) []v1.EnvVar {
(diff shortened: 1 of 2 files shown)


elukey:
Hi everybody, any feedback?

python/kfserving/kfserving/api/creds_utils.py
```diff
@@ -67,7 +67,7 @@ def set_s3_credentials(namespace, credentials_file, service_account,
                               is specified, will attach created secret with the service account,
                               otherwise will create new one and attach with created secret.
         s3_endpoint(str): S3 settings variable S3_ENDPOINT.
-        s3_region(str): S3 settings variable AWS_REGION.
+        s3_region(str): S3 settings variable AWS_DEFAULT_REGION.
```
yuzisun, line 70: Also need to add the s3_cabundle argument here?

pkg/credentials/s3/s3_secret.go
```diff
@@ -49,6 +50,7 @@ var (
 	InferenceServiceS3SecretHttpsAnnotation      = constants.KFServingAPIGroupName + "/" + "s3-usehttps"
 	InferenceServiceS3UseVirtualBucketAnnotation = constants.KFServingAPIGroupName + "/" + "s3-usevirtualbucket"
 	InferenceServiceS3UseAnonymousCredential     = constants.KFServingAPIGroupName + "/" + "s3-useanoncredential"
+	InferenceServiceS3CABundlePathAnnotation     = constants.KFServingAPIGroupName + "/" + "s3-cabundlepath"
```
yuzisun, line 53: `s3-cabundlepath` -> `s3-cabundle` just to be consistent with the env variable name?

yuzisun:
> Hi everybody, any feedback?

Thanks @elukey ! The change mostly lgtm, Can you help link the AWS S3 doc regarding the region variable so we can have a reference?

python/kfserving/kfserving/api/creds_utils.py
elukey: Totally missed this, going to follow up!

pkg/credentials/s3/s3_secret.go
elukey: I was unsure about naming, will change it :)

elukey:
> > Hi everybody, any feedback?
> 
> Thanks @elukey ! The change mostly lgtm, Can you help link the AWS S3 doc regarding the region variable so we can have a reference?

Sure! Do you want only a reference in here or do you prefer also a reference in a code's comment? 

AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

yuzisun:
> > > Hi everybody, any feedback?
> > 
> > 
> > Thanks @elukey ! The change mostly lgtm, Can you help link the AWS S3 doc regarding the region variable so we can have a reference?
> 
> Sure! Do you want only a reference in here or do you prefer also a reference in a code's comment?
> 
> AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
> Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables

code comments will be helpful, thanks!

elukey:
The test to upload something to AWS s3 fails for some reason, will dig a bit more into it. Tried to modify the tests but didn't succeed. Any suggestion is welcome :)

elukey:
I see that in https://github.com/kubeflow/testing/search?q=AWS_REGION the variable is mentioned a lot, so this may explain why I am struggling with tests. But boto doesn't support it, more info: https://github.com/boto/boto3/issues/2574

elukey:
The error that I see is:

```
 INFO|2021-08-31T10:01:27|/src/kubeflow/testing/py/kubeflow/testing/argo_client.py|48| Parsing ApiException body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
INFO|2021-08-31T10:01:27|/src/kubeflow/testing/py/kubeflow/testing/argo_client.py|60| ApiException code=401 
```

Initially I thought that it was related to the s3 api, but maybe it is something (un)related to the k8s api?

elukey:
@yuzisun I found the same error mentioned above in other pull requests, is it a known issue?

elukey:
> > @yuzisun I found the same error mentioned above in other pull requests, is it a known issue?
> 
> ye might be due to the auth token expiration which throws the authorization error, rerunning the test.

Still seeing the issue in the last presubmit job :(

elukey:
Created https://github.com/kubeflow/kfserving/issues/1788

yuzisun:
> Created #1788

@elukey we found the issue, the image built from CI is conflicting with the PR from kserve repo.. see https://github.com/kserve/kserve/pull/44

elukey:
> > Created #1788
> 
> @elukey we found the issue, the image built from CI is conflicting with the PR from kserve repo.. see [kserve/kserve#44](https://github.com/kserve/kserve/pull/44)

Thanks a lot! I've re-run the test but it still fails :(

elukey:
@yuzisun this time it worked!

yuzisun:
Thanks @elukey ! 
/lgtm
/approve

umka1332:
Hi @elukey,

How do you put the CA bundle file(s) inside storage-initializer container? Is it built into image or were you able to mount it somehow into the container? If the second, could you please share steps to do this? I've tried a lot of things but non of them works with default image.

elukey:
@umka1332 Hi! Correct yes, we build our own versions of the Kserve docker images, including the bundle. Then we simply add the environment variables and that's it :)

PR #1807, opened 2021-09-20
add wikimedia in the list of adopters

theofpa:
Wikimedia is building the `liftwing` platform based on kfserving. There are already models in production that score wikipedia edits for quality/damage checks in some languages. More information about the inferenceservices in [wikimedia/machinelearning-liftwing-inference-services](https://github.com/wikimedia/machinelearning-liftwing-inference-services) and https://ores.wikimedia.org.

/cc @elukey @accraze @kevinbazira


elukey:
I am wondering if we should add only @chrisalbon as contact for the project, otherwise I'd keep the current list and I'd add @chrisalbon and https://github.com/klausman. Thanks!

theofpa:
> I am wondering if we should add only @chrisalbon as contact for the project, otherwise I'd keep the current list and I'd add @chrisalbon and https://github.com/klausman. Thanks!

thanks @elukey, you're right, let's keep one contact only for consistency with the rest of the list, see 8bcbb85

elukey:
LGTM! Let's wayt @chrisalbon's approval too just in case :)

chrisalbon:
nice! Thanks Theo!

PR #1834, opened 2021-09-26
Migration job

### LINKED ISSUE #1818: KFServing -> KServe migration script
/kind feature

For users that are migrating from kfserving, kserve installs in its own namespace kserve-system and the user needs to scale down the old controller if there are existing services running in the cluster. The migration script converts the CR from kubeflow.org to kserve.io and gets reconciled in the new controller, however they should still create the exact same knative resources which would not affect the running inference services.

Suresh-Nakkeran:

Fixes #1818 

**Release note**:

```release-note

```

### DIFF
--- a/hack/kserve_migration/kserve_migration.sh
+++ b/hack/kserve_migration/kserve_migration.sh
@@ -72,6 +72,7 @@ log INFO "getting inference services config"
 inference_services=$(kubectl get inferenceservice.serving.kubeflow.org -A -o jsonpath='{.items[*].metadata.namespace},{.items[*].metadata.name}')
 declare -a isvc_names
 declare -a isvc_ns
+declare -A kfserving_isvc_status
 if [ ! -z "$inference_services" ]; then
     mkdir -p ${ISVC_CONFIG_DIR}
     IFS=','; isvc_split=($inference_services); unset IFS;
@@ -82,6 +83,7 @@ isvc_count=${#isvc_names[@]}
 for (( i=0; i<${isvc_count}; i++ ));
 do
     kubectl get inferenceservice.serving.kubeflow.org ${isvc_names[$i]} -n ${isvc_ns[$i]} -o yaml > "${ISVC_CONFIG_DIR}/${isvc_names[$i]}.yaml"
+    kfserving_isvc_status[${isvc_names[$i]}]=$(kubectl get inferenceservice.serving.kubeflow.org ${isvc_names[$i]} -n ${isvc_ns[$i]} -o json | jq --raw-output '.status.conditions | map(select(.type == "Ready"))[0].status')
 done
 
 # Get knative services names
@@ -182,10 +184,12 @@ sleep 5
 log INFO "verifying inference services are migrated and ready"
 for (( i=0; i<${isvc_count}; i++ ));
 do
-    (
-        trap 'log ERROR "inference service ${isvc_names[$i]} did not migrate properly. migration job exits with code 1."' ERR
-        kubectl wait --for=condition=ready --timeout=10s inferenceservice.serving.kserve.io/${isvc_names[$i]} -n ${isvc_ns[$i]}
-    )
+    if [ "${kfserving_isvc_status[${isvc_names[$i]}]}" == "True" ]; then
+        (
+            trap 'log ERROR "inference service ${isvc_names[$i]} did not migrate properly. migration job exits with code 1."' ERR
+            kubectl wait --for=condition=ready --timeout=10s inferenceservice.serving.kserve.io/${isvc_names[$i]} -n ${isvc_ns[$i]}
+        )
+    fi
 done
 
 # Start kfserving controller for clean up
--- a/hack/kserve_migration/kserve_migration.sh
+++ b/hack/kserve_migration/kserve_migration.sh
@@ -1,0 +1,236 @@
+#!/bin/bash
+# Usage: kserve_migration.sh
+
+set -o errexit
+set -o errtrace
+
+export KSERVE_VERSION=v0.7.0-rc0
+export CONFIG_DIR="config"
+export ISVC_CONFIG_DIR="${CONFIG_DIR}/isvc"
+export KSVC_CONFIG_DIR="${CONFIG_DIR}/ksvc"
+
+CLEAN_KFSERVING="true"
+
+# custom logger
+log() {
+    level=$1
+    msg=$2
+    ts=$(date -u +'%F %T')
+    echo "${ts} [${level}] $msg"
+}
+
+# Validates whether controller manager and models web app service running 
+# on this machine for the given namespace or not.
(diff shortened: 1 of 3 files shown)


hack/kserve_migration/kserve_migration.sh
```diff
@@ -0,0 +1,236 @@
...
+            exit 1;
+        fi
+    done
+    po_names=$(kubectl get po -n $namespace -o jsonpath='{.items[*].metadata.name}')
+    for po_name in "${prefix}-controller-manager" "${prefix}-models-web-app"; do
+        if [ ! -z "${po_names##*$po_name*}" ]; then
+            log ERROR "${prefix} controller services are not installed completely."
+            exit 1;
+        fi
+    done
```
yuzisun: I think we do not necessarily need to check web app, for our use we do not run web app which is an optional component.

hack/kserve_migration/kserve_migration.sh
```diff
@@ -0,0 +1,236 @@
...
+fi
+ksvc_count=${#ksvc_names[@]}
+
+(
+    # Stop kfserving controller
+    log INFO "stopping kfserving controller"
+    kubectl scale --replicas=0 statefulset.apps kfserving-controller-manager -n kfserving-system
+    sleep 30
+
+    trap 'kubectl scale --replicas=1 statefulset.apps kfserving-controller-manager -n kfserving-system' ERR
```
yuzisun, line 108: what does this do?

hack/kserve_migration/kserve_migration.sh
```diff
@@ -0,0 +1,236 @@
...
+done
+sleep 5
+
+# Verify that all inference services are migrated and ready
+log INFO "verifying inference services are migrated and ready"
+for (( i=0; i<${isvc_count}; i++ ));
+do
+    (
+        trap 'log ERROR "inference service ${isvc_names[$i]} did not migrate properly. migration job exits with code 1."' ERR
+        kubectl wait --for=condition=ready --timeout=10s inferenceservice.serving.kserve.io/${isvc_names[$i]} -n ${isvc_ns[$i]}
```
yuzisun: what happens if the original service was in `NotReady` state before migration?

hack/kserve_migration/kserve_migration.sh
```diff
@@ -0,0 +1,236 @@
...
+    )
+done
+
+# Start kfserving controller for clean up
+log INFO "starting kfserving controller for clean up"
+kubectl scale --replicas=1 statefulset.apps kfserving-controller-manager -n kfserving-system
```
yuzisun, line 197: starting up the controller might have side effects, do we need to bring up the controller to clean up them?

hack/kserve_migration/kserve_migration.sh
```diff
@@ -0,0 +1,236 @@
...
+    kubectl delete ValidatingWebhookConfiguration trainedmodel.serving.kubeflow.org
+fi
+
+log INFO "deleting kfserving revisions"
+for (( i=0; i<${isvc_count}; i++ ));
+do
+    revisions=$(kubectl get deployment.apps -n ${isvc_ns[$i]} -o json | jq -r '.items[] | select(.status.availableReplicas==null and .metadata.ownerReferences[0].kind=="Revision") | .metadata.ownerReferences[0].name')
+    for revision in "${revisions[@]}"
+    do
+        kubectl delete revision.serving.knative.dev ${revision} -n ${isvc_ns[$i]}
```
yuzisun: why do we need to delete the revisions?

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: Yes @yuzisun . We need controller to clean up older inference service.

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: @yuzisun If the new inference service is not ready, after timeout it throws timeout error, prints error message in trap command and exit from script execution.

hack/kserve_migration/kserve_migration.sh
yuzisun: would kfserving controller still reconciles the older inference service once it comes up ?

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: @yuzisun If we face any issues in deploying new inference service, It restores kfserving controller.

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: Ok @yuzisun . I will remove web app from validation.

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: It's optional. The old revision are active even after deleting kfserving namespace. I added steps clean up revisions as well.

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: yes @yuzisun

yuzisun: So that can cause problem as two controllers are trying to update the knative service with race condition?

hack/kserve_migration/kserve_migration.sh
yuzisun: knative revisions are managed by knative services so we should't need to worry about its lifecycle.

Suresh-Nakkeran: Okay @yuzisun . I will remove this step.

hack/kserve_migration/kserve_migration.sh
Suresh-Nakkeran: @yuzisun I do not face any issues during testing. we are removing kfserving controller as well right after inference service clean up.

hack/kserve_migration/kserve_migration.sh
yuzisun: I think we should only have one controller running, if we scale back `kfserving-controller` then we probably should scale down `kserve-controller`

hack/kserve_migration/kserve_migration.sh
yuzisun: I mean there could be some existing bad services in the cluster, so when deploying to kserve it will not be in ready state

hack/kserve_migration/kserve_migration.sh
yuzisun: ye I feel it might be safer to remove the finalizer so you do not need to scale back the controller to delete the resources

yuzisun:
Thanks @Suresh-Nakkeran ! Great job !
/lgtm
/approve

PR #1836, opened 2021-09-27
adding tgz/zip support for all storages

### LINKED ISSUE #1569: Support unpacking of model as tar/zip file for s3
/kind feature

We are trying to marry  "KF Pipelines + KF Artifacts + KF Serving" in a nice use-case where we train a (sklearn) model using KFP, store it as an output artifact to the artifact store of pipelines (minio) and then deploy that model directly from the artifact store using KFServing. The problem we are facing is that the output artifacts are by default stored as `.tgz` , so even a `.joblib` blob model object is still wrapped into `.tgz`  at creation time of the artifact in minio. On the other end, the default KF serving application server expects a `.joblib` file, not a `.tgz`. 
We're using the `s3` protocol to load the model file from minio to the kfserving pod. 

We already discussed with @yuzisun in the Kubeflow slack, and he pointed out that unpacking of a model (in the storage-initializer container) is currently only supported for http: 
[see this example](https://github.com/kubeflow/kfserving/tree/master/docs/samples/storage/uri#specify-and-create-the-inferenceservice-1)
[see the code](https://github.com/kubeflow/kfserving/blob/47362081224d1b9855648c7fcbd50a3d8037e1cc/python/kfserving/kfserving/storage.py#L233)

I'm opening this issue to first of all start a discussing on how to go about this. 
I can already add that, if it's decided to add support to s3 for the extraction of a model file, we at Credo are happy to help :slightly_smiling_face:

Suresh-Nakkeran:

Fixes https://github.com/kserve/kserve/issues/1569

**Release note**:

```release-note

```

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -109,6 +109,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
         bucket_name = parsed.netloc
         bucket_path = parsed.path.lstrip('/')
 
+        count = 0
         bucket = s3.Bucket(bucket_name)
         for obj in bucket.objects.filter(Prefix=bucket_path):
             # Skip where boto3 lists the directory as an object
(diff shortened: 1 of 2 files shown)


yuzisun:
Thanks @Suresh-Nakkeran ! Great work!

/lgtm
/approve

yuzisun (approved):

PR #1844, opened 2021-09-30
Update quick install script to set up modelmesh

chinhuang007:
Update quick install script to set up modelmesh. The script uses
the modelmesh-serving release assets to deploy modelmesh, so the
version number needs to be updated for each release.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/hack/quick_install.sh
+++ b/hack/quick_install.sh
@@ -34,6 +34,7 @@ export ISTIO_VERSION=1.9.0
 export KNATIVE_VERSION=v0.22.0
 export KSERVE_VERSION=v0.7.0
 export CERT_MANAGER_VERSION=v1.3.0
+export KSERVE_MODELMESH_VERSION=v0.7.0
 
 KUBE_VERSION=$(kubectl version --short=true)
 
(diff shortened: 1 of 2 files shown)


chinhuang007:
@pvaneck Please review. It works using my github account and I will change it to kserve once modelmesh-serving v0.7.0 is released with the expected yaml files.

pvaneck (commented):
I think this looks good. We can merge this once we cut modelmesh-serving `v0.7.0` early next week.
/hold

chinhuang007:
@yuzisun I would like to have this PR in release 0.7.0 so that ModelMesh can be installed as part of quick install. I have tested with the modelmesh-serving v0.7.0-rc1. Since the formal kserve v0.7.0 will depend on modelmesh-serving v0.7.0, I put  KSERVE_MODELMESH_VERSION=v0.7.0 in script.
@pvaneck Please update us when modelmesh-serving v0.7.0 is released. Thanks!

hack/quick_install.sh
```diff
@@ -106,3 +107,14 @@ if [ ${KSERVE_VERSION:3:1} -gt 6 ]; then KSERVE_CONFIG=kserve.yaml; fi
 for i in 1 2 3 4 5 ; do kubectl apply -f install/${KSERVE_VERSION}/${KSERVE_CONFIG} && break || sleep 15; done
 # Clean up
 rm -rf istio-${ISTIO_VERSION}
+
+# Install KServe Modelmesh
+kubectl create ns modelmesh-serving || true
+CURRENT_NAMESPACE=$(kubectl config  get-contexts $(kubectl config current-context) |tail -1|awk '{ print $5 }')
+kubectl config set-context --current --namespace=modelmesh-serving
+
+kubectl apply -f https://github.com/kserve/modelmesh-serving/releases/download/${KSERVE_MODELMESH_VERSION}/modelmesh-optional-dependencies.yaml
```
yuzisun: What's in the optional dependencies ? can this be installed optionally?

hack/quick_install.sh
```diff
@@ -106,3 +107,14 @@ if [ ${KSERVE_VERSION:3:1} -gt 6 ]; then KSERVE_CONFIG=kserve.yaml; fi
 for i in 1 2 3 4 5 ; do kubectl apply -f install/${KSERVE_VERSION}/${KSERVE_CONFIG} && break || sleep 15; done
 # Clean up
 rm -rf istio-${ISTIO_VERSION}
+
+# Install KServe Modelmesh
+kubectl create ns modelmesh-serving || true
+CURRENT_NAMESPACE=$(kubectl config  get-contexts $(kubectl config current-context) |tail -1|awk '{ print $5 }')
+kubectl config set-context --current --namespace=modelmesh-serving
```
yuzisun, line 114: Is this a bit confusing that if users are trying out kserve for single model serving but it defaults to `modelmesh-serving` namespace?

hack/quick_install.sh
chinhuang007: This cannot be optionally installed during quick install, because we assumes the user has not installed prerequisites. We will change the yaml name in modelmesh-serving and the script here shortly.

hack/quick_install.sh
chinhuang007: Right now, the quick start installs modelmesh-serving into the default modelmesh-serving namespace for users to experiment. As we discussed separately, the controller is namespaced and will need to be installed into other namespaces following  modelmesh-serving install instructions. This is definitely not ideal and we are working to fix it.

pvaneck: Yea, I think the assumption with the quick install is that we have an instance of ModelMesh Serving running in some namespace, but the caveat is that if a user wants to deploy on ModelMesh, they must deploy in that namespace. If anything, we should maybe have something in the "Get Started" section of the website which elaborates on this and explains deploying an Isvc/Predictor on ModelMesh.

yuzisun:
merged https://github.com/kserve/kserve/pull/2032, closing this.

PR #1850, opened 2021-10-04
deprecate v1alpha2 api group

### LINKED ISSUE #1618: deprecate inference service v1alpha2
/kind feature

inference service v1alpha2 version should no longer be used after 0.6, we should remove v1alpha2 version support from crd and corresponding webhooks, controller and api types.

Suresh-Nakkeran:

Fixes #1618 

**Release note**:

```release-note

```


config/rbac/role.yaml
```diff
@@ -6,19 +6,6 @@ metadata:
   creationTimestamp: null
   name: kserve-manager-role
 rules:
-- apiGroups:
```
yuzisun: these role bindings should not be removed?

Suresh-Nakkeran: @yuzisun These role are removed when I ran make `deploy-dev` so I added.

yuzisun: oh maybe we should look at why `deploy-dev` removed those

Suresh-Nakkeran: @yuzisun I added missing rbac kubebuilder steps in beta1 controller. Now the roles are not ignored.

yuzisun:
@Suresh-Nakkeran looks [here](https://github.com/kserve/kserve/blob/master/test/workflows/components/workflows.libsonnet#L392) we are still using the transformer dockerfile from v1alpha2, can we change to use the [dockerfile](https://github.com/kserve/kserve/blob/master/docs/samples/v1beta1/transformer/torchserve_image_transformer/transformer.Dockerfile) in v1beta1? We need to fix to install [kserve](https://github.com/kserve/kserve/blob/master/docs/samples/v1beta1/transformer/torchserve_image_transformer/transformer.Dockerfile#L8), also fix the [e2e test](https://github.com/kserve/kserve/blob/master/test/e2e/transformer/test_transformer.py) to use the [v1beta1 transformer example](https://github.com/kserve/kserve/blob/master/docs/samples/v1beta1/transformer/torchserve_image_transformer/transformer.yaml).

Makefile
```diff
@@ -119,17 +118,17 @@ manifests: controller-gen
 	perl -pi -e 's/conditions: null/conditions: []/g' config/crd/serving.kserve.io_trainedmodels.yaml
 	perl -pi -e 's/Any/string/g' config/crd/serving.kserve.io_trainedmodels.yaml
 	#remove the required property on framework as name field needs to be optional
-	yq d -i config/crd/serving.kserve.io_inferenceservices.yaml 'spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.*.properties.*.required'
+	yq d -i config/crd/serving.kserve.io_inferenceservices.yaml 'spec.validation.openAPIV3Schema.properties.spec.properties.*.properties.*.required'
```
yuzisun, line 121: I think `validation` field is actually from v1beta1 crd which is deprecated, see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/

@andyi2it can confirm.

yuzisun:
Thanks @Suresh-Nakkeran !

/lgtm
/approve

PR #1873, opened 2021-10-20
Handle migration for kubeflow based installs

markwinter:

KFServing can be installed in `kfserving-system` when standalone, or in `kubeflow` namespace when installed as part of Kubeflow.

I have updated the migration to handle kubeflow based installs.

I've tested this on our dev cluster. I used the v0.7 kserve_kubeflow.yaml file to install KServe alongside existing v0.6.1 kfserving_kubeflow install

### Changes

I updated the script to handle a different kfserving namespace

I added a separate migration job yaml that is for kubeflow installations
  - It installs the migration roles etc. in the `kubeflow` namespace
  - Sets `KFSERVING_NAMESPACE` job env var to `kubeflow`
  - Disables istio sidecar injection for the job pod because this broke contacting the k8s api server

I can update the migration documentation after this PR is merged.

We should also add cleaning up the migration resources after it is complete

```
kubectl delete ClusterRoleBinding cluster-migration-rolebinding
kubectl delete ClusterRole cluster-migration-role
kubectl delete ServiceAccount cluster-migration-svcaccount -n kubeflow
```

### DIFF
--- a/hack/kserve_migration/kserve_migration.sh
+++ b/hack/kserve_migration/kserve_migration.sh
@@ -10,6 +10,7 @@ export ISVC_CONFIG_DIR="${CONFIG_DIR}/isvc"
 export KSVC_CONFIG_DIR="${CONFIG_DIR}/ksvc"
 
 CLEAN_KFSERVING="true"
+KFSERVING_NAMESPACE=${KFSERVING_NAMESPACE:-kfserving-system}
 
 # custom logger
 log() {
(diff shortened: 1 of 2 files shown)


markwinter:
Hey @Suresh-Nakkeran @yuzisun could you take a look at this please?

Suresh-Nakkeran:
@markwinter The changes look good to me. We can keep one migration job yaml file. We can update env variables value and run if we run for kubeflow.
@yuzisun your thoughts?

markwinter:
@Suresh-Nakkeran 

I think using the same migration yaml wont work because of the other changes:
- Disable istio sidecar injection
- Install clusterrole, serviceaccount, clusterrolebinding in kubeflow namespace

Unless I'm missing a nice way to do this?

Suresh-Nakkeran:
@markwinter sounds good to me.

yuzisun:
Thanks @markwinter !
/lgtm
/approve

markwinter:
@yuzisun @Suresh-Nakkeran 

Thank you for reviewing.

Is `kserve/kserve-migration:latest` docker image updated automatically or does someone need to update it?

PR #1472, opened 2021-03-19
Change storage-init to use boto3

matty-rose:
**What this PR does / why we need it**:
Uses the boto3 library instead of minio for storage-initializer, mainly in order to allow use of IAM Roles for Service Accounts for AWS users, also changes storage tests to mock boto3 instead of minio

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #831

**Special notes for your reviewer**:

Cherrypicks the first commit from https://github.com/kubeflow/kfserving/pull/1130 since there doesn't seem to be much continuing activity there

I've also tested this by building/pushing the new image to an ECR registry, changing the image location in the inference config map, and using service accounts both with secret credentials, and IAM roles.

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note
- Replaces minio library with boto3 for storage-initializer container to pull models from S3 so that IRSA can be used
```

PR #1883, opened 2021-10-26
storage.py: add more logging and an error condition for s3

### LINKED ISSUE #1882: Improve storage-initializer's messages/logs
/kind bug

Hi everybody! I spent quite some time investigating a configuration issue for the storage-initializer, namely a `STORAGE_URI` env value set like:

```
s3://wmf-ml-models/articlequality/enwiki/20211022183902/model.bin
```
instead of
```
s3://wmf-ml-models/articlequality/enwiki/20211022183902/
```

It was clearly a misconfiguration issue but the log messages from the container made it more confusing:

```
[I 211025 16:39:52 storage-initializer-entrypoint:13] Initializing, args: src_uri [s3://wmf-ml-models/articlequality/enwiki/20211022183902/model.bin] dest_path[ [/mnt/models]
[I 211025 16:39:52 storage:52] Copying contents of s3://wmf-ml-models/articlequality/enwiki/20211022183902/model.bin to local
[I 211025 16:39:52 credentials:1102] Found credentials in environment variables.
[I 211025 16:39:52 storage:85] Successfully copied s3://wmf-ml-models/articlequality/enwiki/20211022183902/model.bin to /mnt/models
```
The above seems to suggest that `model.bin` has been copied successfully, meanwhile what happens is that nothing is copied due to the wrong `STORAGE_URI` (that is an object path, rather than something than contains objects).

Same thing if I put a `STORAGE_URI` that points to no objects or to a non-existent path on s3. I believe that the code in https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L114 should check if any object is contained in the URI provided, and report an error if nothing is returned (ideally skipping the "Successfully copied ..."). In this way the pod creation would fail at the storage-initializer bootstrap, and not later on (in my case, the kserve-container failed, rightfully, due to the absence of the model.bin file that it was looking for).

elukey:
The storage initializer for s3 ends up into the following
two use cases:
- If no object is present in the bucket, it doesn't raise
  any error, and later on the kserve-container fails since
  no model binary is present.
- If a full object path is specified (like /a/b/../model.bin)
  then the s3 downloader will create the same path under
  /mnt/models, and if the kserve-container is not smart enough
  it will fail to find the model.bin.
The current storage-initializer logging doesn't give any hint
or errors about the above use case, ending up in difficult-to-debug
scenarios (see the related issue for more info).

Fixes: #1882

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -117,8 +117,20 @@ class Storage(object):  # pylint: disable=too-few-public-methods
                 continue
             # In the case where bucket_path points to a single object, set the target key to bucket_path
             # Otherwise, remove the bucket_path prefix, strip any extra slashes, then prepend the target_dir
+            # Example:
+            # s3://test-bucket
+            # Objects: /a/b/c/model.bin /a/model.bin /model.bin
+            #
+            # If 'uri' is set to "s3://test-bucket", then the dowloader will
+            # download all the objects listed above, re-creating their subpaths
+            # under the temp_dir.
+            # If 'uri' is set to "s3://test-bucket/a", then the downloader will
+            # add to temp_dir: b/c/model.bin and model.bin.
+            # If 'uri' is set to "s3://test-bucket/a/b/c/model.bin", then
+            # the downloader will add to temp dir: model.bin
+            # (without any subpaths).
             target_key = (
-                obj.key
+                obj.key.rsplit("/", 1)[-1]
                 if bucket_path == obj.key
                 else obj.key.replace(bucket_path, "", 1).lstrip("/")
             )
@@ -126,7 +138,11 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             if not os.path.exists(os.path.dirname(target)):
                 os.makedirs(os.path.dirname(target), exist_ok=True)
             bucket.download_file(obj.key, target)
+            logging.info('Downloaded object %s to %s' % (obj.key, target))
             count = count + 1
+        if count == 0:
+            raise RuntimeError(
+                "Failed to fetch model. No model found in %s." % bucket_path)
 
         # Unpack compressed file, supports .tgz, tar.gz and zip file formats.
         if count == 1:
@@ -164,8 +180,8 @@ class Storage(object):  # pylint: disable=too-few-public-methods
                 blob.download_to_filename(dest_path)
             count = count + 1
         if count == 0:
-            raise RuntimeError("Failed to fetch model. \
-The path or model %s does not exist." % uri)
+            raise RuntimeError(
+                "Failed to fetch model. No model found in %s." % uri)
 
         # Unpack compressed file, supports .tgz, tar.gz and zip file formats.
         if count == 1:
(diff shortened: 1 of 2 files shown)


elukey:
@yuzisun Hi! Do you think that the PR is eligible for a test?

python/kserve/kserve/storage.py
```diff
@@ -134,6 +135,11 @@ def _download_s3(uri, temp_dir: str):
             if mimetype in ["application/x-tar", "application/zip"]:
                 Storage._unpack_archive_file(target, mimetype, temp_dir)
 
+        # Raise an error if nothing is retrieved from the bucket
+        # provided in input.
+        if count == 0:
+            raise RuntimeError('No object found in %s' % bucket_path)
```
yuzisun: Thanks for fixing this! Can you help make the error message consistent with other storage type?

yuzisun:
> * If a full object path is specified (like /a/b/../model.bin)
>   then the s3 downloader will create the same path under
>   /mnt/models, and if the kserve-container is not smart enough
>   it will fail to find the model.bin.

This seems wrong we should fix that, the [code snippet](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L118) here is problematic.

python/kserve/kserve/storage.py
elukey: Done! Let me know if it is ok now, I tried also to simply the common error msg :)

elukey:
> > * If a full object path is specified (like /a/b/../model.bin)
> >   then the s3 downloader will create the same path under
> >   /mnt/models, and if the kserve-container is not smart enough
> >   it will fail to find the model.bin.
> 
> This seems wrong we should fix that, the [code snippet](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L118) here is problematic.

I thought the same too, and I have updated the patch with a simple fix. I didn't do it before since the logic seems to be repeated elsewhere, for example download_gcs seems to do something similar afaics. At this point we probably need to decide what is the intended semantic for the downloader, and apply it to all occurrences of it.

yuzisun:
@elukey I think the unit test needs a fix as well.

elukey:
> @elukey I think the unit test needs a fix as well.

Definitely, I'll try to fix it. It feels though that the semantic that we are trying to change was intended, afaics the last big change was done in https://github.com/kserve/kserve/pull/1472/files. I am ok to modify the semantic of course, but I am wondering if there are people already relying on it (that we may break with this change).

elukey:
@yuzisun Hi! I fixed the unit test, and also added more comments related to what I think the code does. Let me know!

python/kserve/kserve/storage.py
```diff
@@ -117,16 +117,32 @@ def _download_s3(uri, temp_dir: str):
...
+            # download all the objects listed above, re-creating their subpaths
+            # under the temp_dir.
+            # If 'uri' is set to "s3://test-bucket/a", then the downloader will
+            # add to temp_dir: b/c/model.bin and model.bin.
+            # If 'uri' is set to "s3://test-bucket/a/b/c/model.bin", then
+            # the downloader will add to temp dir: model.bin
+            # (without any subpaths).
             target_key = (
-                obj.key
+                obj.key.rsplit("/", 1)[1]
```
yuzisun: would this be an issue when `storageUri` is `s3://test-bucket/model.bin` and `obj.key` is `model.bin`?

elukey: @yuzisun good point! I replaced the [1] with [-1], it should work fine even when the rsplit list is one element:
```
>>> "model.bin".rsplit("/",1)[-1]
'model.bin'
>>> "/a/b/c/model.bin".rsplit("/",1)[-1]
'model.bin'
```

Added also a specific unit test for it, lemme know :)

yuzisun:
Thanks @elukey !! This is great and fixed the long standing bug!
 /lgtm
 /approve

PR #1899, opened 2021-11-05
Add phase 1 implementation of new storage spec

### LINKED ISSUE #1814: Propose new storage spec to support next-gen multi-model serving
/kind feature

We want to introduce a new storage spec for inference service and trained model, so it can support the next-gen model mesh multi-model serving with different storage provider. The storageUri can continue to work with this new storage spec.
Example:
```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/deploymentMode: ModelMesh
spec:
  predictor:
    sklearn:
      # Use either storageUri or storage spec
      # storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
      storage:
        key: localMinIO
        path: sklearn/mnist-svm.joblib
        # schemaPath: null
        parameters:
          bucket: modelmesh-example-models
```

Proposed new storage spec design doc: https://docs.google.com/document/d/1rYNh93XNMRp8YaR56m-_zK3bqpZhhGKgpVbLPPsi-Ow/edit#
Phase 1 implementation for modelmesh serving controller: https://github.com/kserve/modelmesh-serving/pull/56

Tomcli:
The new modelmesh serving method needs a new storage definition to unify the model storage for both single and multi-model serving on different framework server. The Phase 1 is to show what the new storage definition looks like without breaking any regression test. In this phase 1 implementation, we initially has S3 storage, but will add more storage support when the modelmesh agent able to support other storage types. 

Design docs: https://docs.google.com/document/d/1rYNh93XNMRp8YaR56m-_zK3bqpZhhGKgpVbLPPsi-Ow/edit#

Fixes #1814

**Release note**:

```release-note

```


config/configmap/inferenceservice.yaml
```diff
@@ -140,7 +140,8 @@ data:
         "memoryRequest": "100Mi",
         "memoryLimit": "1Gi",
         "cpuRequest": "100m",
-        "cpuLimit": "1"
+        "cpuLimit": "1",
+        "storageSpecSecret": "storage-config"
```
yuzisun: this secret is global scoped?

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
...
+  name: example-sklearn-isvc
+spec:
+  predictor:
+    sklearn:
+      storage:
+        key: localMinIO # Credential key for the destination storage in the common secret
+        path: sklearn # Model path inside the bucket
+        # schemaPath: null # Optional schema files for payload schema
+        parameters: # Parameters to override the default values inside the common secret.
+          default-bucket: example-models
```
yuzisun: Why not including the `bucket` on the path? I found this is an extra step for user to think about and changes the conventional s3 path.

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
+# Predict on a InferenceService with StorageSpec
+## Setup
+1. Your ~/.kube/config should point to a cluster with [KServe installed](https://github.com/kserve/kserve#installation).
+2. Your cluster's Istio Ingress gateway must be [network accessible](https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/).
+3. Your cluster's Istio Egresss gateway must [allow accessing S3 Storage](https://knative.dev/docs/serving/outbound-network-access/)
+
+## Create Common Storage Secret
+Create a common secret with your [storage credential](https://console.aws.amazon.com/iam/home#/users), `KServe` reads the secret key to inject
```
yuzisun, line 8: User may still have multiple secrets, how do we specify that on the spec?

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
...
+```yaml
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: example-sklearn-isvc
+spec:
+  predictor:
+    sklearn:
+      storage:
+        key: localMinIO # Credential key for the destination storage in the common secret
```
yuzisun, line 53: Do we provide a default key name?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -44,7 +44,7 @@ const (
 
 // Constants
 var (
-	SupportedStorageURIPrefixList = []string{"gs://", "s3://", "pvc://", "file://", "https://", "http://"}
+	SupportedStorageURIPrefixList = []string{"gs://", "s3://", "pvc://", "file://", "https://", "http://", "secret://"}
```
yuzisun: Why do we need to introduce this artificial scheme type? Can we construct back the s3 or other url from the storage spec?

python/kserve/kserve/storage.py
```diff
@@ -77,6 +78,8 @@ def download(uri: str, out_dir: str = None) -> str:
             # Don't need to download models if this InferenceService is running in the multi-model
             # serving mode. The model agent will download models.
             return out_dir
+        elif uri.startswith(_STROAGE_SPEC_PREFIX):
```
yuzisun: instead of using `secret://`, can we check if the secret volume exists and then load the volume, download using storage spec?

config/configmap/inferenceservice.yaml
Tomcli: Secrets are namespace scope in Kubernetes. In KFP, they have a profile controller that replicates all the necessary namespace scoped resource when creating a new user. We can leverage the same capability to replicate the common secret for Kubeflow users. For standalone users, we are expecting them to create their own common secrets for their namespaces because we cannot authorize those users.
https://github.com/kubeflow/manifests/blob/master/apps/pipeline/upstream/base/installs/multi-user/pipelines-profile-controller/sync.py

docs/samples/storage/storageSpec/README.md
Tomcli: In the next phase we can have kserve controller or create a new controller to sync multiple secrets into one common secret in each namespace. The reason we want to have a common secret so that when using modelmesh or multi-model serving, the model serving pod doesn't have to roll out again to mount on new secret volume.

docs/samples/storage/storageSpec/README.md
Tomcli: We are planning to support the StorageUri format with the new storage Spec. I didn't include it in this PR because the StorageUri can override the new Storage parameters and break our current behavior. I can bring this up in this week's community call and discuss what parameters are immutable and add the StorageUri support in my next PR.

Currently this PR didn't break any backward compatibility so you can still use the old S3 secret mounting method.

docs/samples/storage/storageSpec/README.md
Tomcli: We don't have a default storage key for standalone deployment. Maybe I can add something for the Kubeflow deployment and borrow the KFP minio to store our models.

pkg/apis/serving/v1beta1/component.go
Tomcli: At the KServe controller level, the controller may not know the bucket name because the `default-bucket` in the storage API is optional. Unless the controller reads the secret in the user namespace to get the default-bucket value for every model deployment.

python/kserve/kserve/storage.py
Tomcli: Sure, I can do that. My original intent is try to not break any backward compatibility until we introduced all storage support in the new storage spec.

pkg/apis/serving/v1beta1/component.go
yuzisun: that's why I strongly suggest removing `default-bucket` and have user explicitly specify the bucket on the path
```
spec:
  predictor:
    sklearn:
      storage:
        key: localMinIO
        path: example-models/sklearn # path including the bucket
```

Tomcli: So I think @njhill has this requirement where the default-bucket is a parameter in the secret. @njhill what do you think?

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
+# Predict on a InferenceService with StorageSpec
+## Setup
+1. Your ~/.kube/config should point to a cluster with [KServe installed](https://github.com/kserve/kserve#installation).
```
njhill:
```suggestion
1. Your `~/.kube/config` should point to a cluster with [KServe installed](https://github.com/kserve/kserve#installation).
```

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
...
+apiVersion: v1
+data:
+  localMinIO: ewogICJ0eXBlIjogInMzIiwKICAiYWNjZXNzX2tleV9pZCI6ICJtaW5pbyIsCiAgInNlY3JldF9hY2Nlc3Nfa2V5IjogIm1pbmlvMTIzIiwKICAiZW5kcG9pbnRfdXJsIjogImh0dHA6Ly9taW5pby1zZXJ2aWNlLmt1YmVmbG93OjkwMDAiLAogICJkZWZhdWx0X2J1Y2tldCI6ICJtbHBpcGVsaW5lIiwKICAicmVnaW9uIjogInVzLXNvdXRoIgp9
+kind: Secret
+metadata:
+  name: storage-config
+type: Opaque
+```
+
+The above `localMinIO` key is encoded with the Kubeflow minio storage credentials. The format of the above credentials is:
```
njhill: Could simplify the example above to use `stringData` instead of `data` and then the json could be embedded/readable directly (no need for separate snippet).

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Apply the secret and service account
+```bash
+kubectl apply -f common_secret.yaml
+```
+
+Then, download the [sklearn model](https://storage.cloud.google.com/kfserving-samples/models/sklearn/iris/model.joblib) and store the model at the path `sklearn/model.joblib` inside the a new bucket called `example-models`.
+
+Note: if you are running kserve with istio sidecars enabled, there can be a race condition between the istio proxy being ready and the agent pulling models.
+This will result in a `tcp dial connection refused` error when the agent tries to download from s3.
```
njhill, line 37: Kind of tangential but we could consider adding some tolerance to the agent for this - to retry some number of times (with delay in between) if such an error is encountered close to pod startup time.

docs/samples/storage/storageSpec/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+The above `localMinIO` key is encoded with the Kubeflow minio storage credentials. The format of the above credentials is:
+```
+{
+  "type": "s3",
+  "access_key_id": "minio",
+  "secret_access_key": "minio123",
+  "endpoint_url": "http://minio-service.kubeflow:9000",
+  "default_bucket": "mlpipeline",
```
njhill: Now that we are planning to allow overriding of storage parameters arbitrarily in the InferenceService, we should change the name of this one to just `bucket`.

docs/samples/storage/storageSpec/README.md
njhill: @Tomcli as discussed on the call we can just have the controller treat the spec as invalid if both `storageUri` and the storage spec are provided and conflict with each other.

I think it would be good to retain support for it in this PR so as to make it clear the existing approach won't be broken, and to include another small example here which uses `storageUri` instead of the spec.

docs/samples/storage/storageSpec/README.md
njhill: We should probably have an implicit default key per type of storage. This would be used for example (if present) if only a `storageUri` is provided, and could be automatically populated from the existing kfserving style storage secrets.

njhill: I'm also wondering whether there's a better name to use for this than `key`, though not sure I can think of one :)

pkg/apis/serving/v1beta1/component.go
njhill: As commented above, I feel like it might be better for the `storage.path` to not include the bucket since it would be used for other kinds of storage too and so having it be "just the path" seems more consistent / less ambiguous. The `s3://` storage uris will still include bucket as a prefix of course.

But I'm also not sure that a new `secret://` uri scheme is needed.

pkg/apis/serving/v1beta1/explainer_aix360.go
```diff
@@ -44,10 +45,20 @@ type AIXExplainerSpec struct {
...
+	}
+	return nil
+}
+
+func (s *AIXExplainerSpec) GetStorageSpec() *StorageSpec {
+	if s.Storage != nil && s.Storage.StorageKey != nil {
+		return s.Storage
 	}
-	return &s.StorageURI
+	return nil
```
njhill: If we keep this approach it would be good to deduplicate the logic in the above two functions across the various component impls.

njhill (commented):
Thanks @Tomcli this is a great start, and sorry for the delayed review, I had to re-familiarize myself with how the existing storage settings are orchestrated.

Part of the impetus for introducing a more structured/consistent storage spec is to improve some of the inconsistency around how the different kinds of storage are configured, and I think that can be done for both the internals and implementation. I will add some more detailed thoughts on this as a separate comment in this PR.

njhill:
Here's my proposal on how to standardize:

The `ComponentInterface.GetStorageUri()` func is currently used only:
1. As part of the determination of whether the InferenceService is a (non-modelmesh) MMS
2. To set the storageUri annotation on the corresponding component's kube resources, to be picked up by the storage initializer injector

For (1) I'd propose to leave existing behaviour as-is so as not to break backwards compatibility (don't change impl).
For (2) I think we should make a change to set an annotation containing the "resolved" storage spec in json form instead of the storage uri. In this case that only a storageUri is provided, this would be converted to a storage spec first.

This gets passed to the storage initializer on the command line, I would propose to instead just set it as an environment variable (the json contents). Maybe the storageUri command line arg could continue to be supported, but I presume that it it's not used as a "standalone" utility?

The storage-type specific env vars / volumes from existing secrets are separately added to the storage initializer / agent via the `CredentialBuilder.CreateSecretVolumeAndEnv()` func. I propose we change this function to instead populate/update the shared/common storage secret, and then always mount just that secret to the agent / storage initializer. So no environment variables are set for this (we could consider one env var which points to the mounted storage secret dir). Note that we would probably need for this common storage secret to be per- service account so may need to think about the naming (couldn't be a single well-defined name in that case).

`storage.py` then does not need to have any knowledge of storageUris and would work only with the provided storage spec. The "key" would be read from the spec or otherwise default based on the type, and then this key would be used to look up a corresponding entry (file) in the mounted storage secret dir. The combination of the provided spec and the contents of that storage entry would then be used to configure the appropriate storage client (rather than env variables).

@Tomcli @yuzisun WDYT?

pkg/apis/serving/v1beta1/component.go
Tomcli: I have the `secret://` uri scheme because the controller doesn't know what type of storages are in the secret. To know the type of storage, we either need to let the controller pull the user secret every time a new model is deployed/updated or ask them to provide the storageUri that has the storage type.

docs/samples/storage/storageSpec/README.md
Tomcli: done

docs/samples/storage/storageSpec/README.md
Tomcli: This is more an issue when deployed with istio sidecar. For modelmesh it should be addressed in the modelmesh sidecar.

pkg/apis/serving/v1beta1/component.go
Tomcli: Seems like the credentials package already listing all the secrets in the user namespace. We can have add some extra logics to also check the storage `type` fields in the common secret to know the storage type in controller time if that's okay with everyone.

Tomcli:
> Here's my proposal on how to standardize:
> 
> The `ComponentInterface.GetStorageUri()` func is currently used only:
> 
> 1. As part of the determination of whether the InferenceService is a (non-modelmesh) MMS
> 2. To set the storageUri annotation on the corresponding component's kube resources, to be picked up by the storage initializer injector
> 
> For (1) I'd propose to leave existing behaviour as-is so as not to break backwards compatibility (don't change impl). For (2) I think we should make a change to set an annotation containing the "resolved" storage spec in json form instead of the storage uri. In this case that only a storageUri is provided, this would be converted to a storage spec first.
> 
> This gets passed to the storage initializer on the command line, I would propose to instead just set it as an environment variable (the json contents). Maybe the storageUri command line arg could continue to be supported, but I presume that it it's not used as a "standalone" utility?
> 
> The storage-type specific env vars / volumes from existing secrets are separately added to the storage initializer / agent via the `CredentialBuilder.CreateSecretVolumeAndEnv()` func. I propose we change this function to instead populate/update the shared/common storage secret, and then always mount just that secret to the agent / storage initializer. So no environment variables are set for this (we could consider one env var which points to the mounted storage secret dir). Note that we would probably need for this common storage secret to be per- service account so may need to think about the naming (couldn't be a single well-defined name in that case).
> 
> `storage.py` then does not need to have any knowledge of storageUris and would work only with the provided storage spec. The "key" would be read from the spec or otherwise default based on the type, and then this key would be used to look up a corresponding entry (file) in the mounted storage secret dir. The combination of the provided spec and the contents of that storage entry would then be used to configure the appropriate storage client (rather than env variables).
> 
> @Tomcli @yuzisun WDYT?

Thanks @njhill 

For (1) we didn't change the behavior of how isvc determine MMS. I only changed the implementation logic when the model needed the storage initializer. 

For (2) Let the controller to create a common storage secret per service account is a good approach. This can help KServe to sync the current s3/gcs/azure storage secrets into a common storage secret. We originally planned to do this in phases 2-3 because it changes the implementation and inputs for storage.py.

So for (2) I have two suggestions,
A) I can implement what you have suggested above and let the storage initializer always mount on the common secret to get the credentials. It can be implemented in a way that's seamlessly for Kubernetes users. However, developers who are testing the KServe python server on their local machine need to be aware of the parameter changes in the python storage class. 

B) If we don't want to change the current storage.py implementation, I can move all the new storage spec credential loading logics to the controller side. This means the controller will read the common secret to extra the storage type, access key, and secret access key. Then map them as environment variables or Uri the same way as the current storage.py. This way, no changes need to be done on the python server and the initializer script, but the controller will have storage-specific logic for parsing the common secret and mapping them to the current storageUri and secret format.

Both solution above can remove the `secret://` schema since the controller is going to parse the secret. @njhill probably wants the (A) approach. @yuzisun What do you think?

njhill:
@Tomcli maybe (B) is least invasive for a first pass (at least to support just S3 via the new config mechanism) and defers having to worry about the secret transfer part.

But (A) imo would be better destination since we'll then have consistency of how storage config is consumed by the storage initializer/puller, i.e. a single secret with a consistent schema which can work the same for multi and single model (and same code used by both). storageUri parsing will then only be done in the controller, as outlined in my prior comment.

Question for @yuzisun and others: Do we think it could be an issue supporting the existing secret formats via the controller copying them to the new common secret? I.e. from a security/permission pov since the controller's service account would require secret write access. It would only be done within the context of the same service account.

pkg/apis/serving/v1beta1/explainer_aix360.go
```diff
@@ -44,10 +45,20 @@ type AIXExplainerSpec struct {
...
-	if s.StorageURI == "" {
-		return nil
+	if s.StorageURI != "" {
+		return &s.StorageURI
+	} else if s.Storage != nil && s.Storage.StorageKey != nil && s.Storage.Path != nil {
+		storageSpecUri := "secret://" + *s.Storage.StorageKey + "/" + strings.TrimPrefix(*s.Storage.Path, "/")
+		return &storageSpecUri
+	}
+	return nil
+}
```
njhill: @Tomcli I don't think @pvaneck's first iteration of the `ServingRuntime` stuff is touching this as you thought it might, so as discussed I think we should just move these func implementations to `PredictorExtensionSpec` and `ExplainerExtensionSpec`.

Tomcli: done

docs/samples/storage/storageSpec/README.md
Tomcli: I add a default key for the common secret. We can have the default key per storage in next phase when we migrate the storage initializer to new golang based image.

python/kserve/kserve/storage.py
Tomcli: removed

Tomcli:
@njhill sorry for the late update. I addressed your comments and implemented with the (A) method where all the mapping logics are in the controller now. Since we are not modifying the python storage initializer for this PR, we are pushing all the environments variables as `value` for now. This will be changed to secret mounting in phase 2 when we bring in the modelmesh golang storage client.

pkg/agent/storage/gcs.go
```diff
@@ -88,7 +88,7 @@ func (g *GCSObjectDownloader) Download(client stiface.Client, it stiface.ObjectI
 		}
 		objectValue := strings.TrimPrefix(attrs.Name, g.Item)
 		fileName := filepath.Join(g.ModelDir, g.ModelName, objectValue)
-		
+
```
njhill: Revert unrelated whitespace changes in this file?

pkg/agent/storage/s3.go
```diff
@@ -94,7 +94,7 @@ func (s *S3ObjectDownloader) GetAllObjects(s3Svc s3iface.S3API) ([]s3manager.Bat
 		}
 		subObjectKey := strings.TrimPrefix(*object.Key, s.Prefix)
 		fileName := filepath.Join(s.ModelDir, s.ModelName, subObjectKey)
-		
+
```
njhill: Revert unrelated whitespace changes in this file?

pkg/apis/serving/v1beta1/explainer_aix360.go
```diff
@@ -44,10 +44,11 @@ type AIXExplainerSpec struct {
...
 func (s *AIXExplainerSpec) GetStorageUri() *string {
-	if s.StorageURI == "" {
-		return nil
-	}
-	return &s.StorageURI
+	return s.ExplainerExtensionSpec.GetStorageUri()
+}
+
+func (s *AIXExplainerSpec) GetStorageSpec() *StorageSpec {
+	return s.ExplainerExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `ExplainerExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -56,10 +56,11 @@ type AlibiExplainerSpec struct {
...
 func (s *AlibiExplainerSpec) GetStorageUri() *string {
-	if s.StorageURI == "" {
-		return nil
-	}
-	return &s.StorageURI
+	return s.ExplainerExtensionSpec.GetStorageUri()
+}
+
+func (s *AlibiExplainerSpec) GetStorageSpec() *StorageSpec {
+	return s.ExplainerExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `ExplainerExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/explainer_art.go
```diff
@@ -44,10 +44,11 @@ type ARTExplainerSpec struct {
...
 func (s *ARTExplainerSpec) GetStorageUri() *string {
-	if s.StorageURI == "" {
-		return nil
-	}
-	return &s.StorageURI
+	return s.ExplainerExtensionSpec.GetStorageUri()
+}
+
+func (s *ARTExplainerSpec) GetStorageSpec() *StorageSpec {
+	return s.ExplainerExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `ExplainerExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_lightgbm.go
```diff
@@ -78,7 +78,11 @@ func (x *LightGBMSpec) GetContainer(metadata metav1.ObjectMeta, extensions *Comp
 }
 
 func (x *LightGBMSpec) GetStorageUri() *string {
-	return x.StorageURI
+	return x.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (x *LightGBMSpec) GetStorageSpec() *StorageSpec {
+	return x.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -55,7 +55,14 @@ func (m *ModelSpec) Validate() error {
 func (m *ModelSpec) Default(config *InferenceServicesConfig) {}
 
 func (m *ModelSpec) GetStorageUri() *string {
-	return m.StorageURI
+	return m.PredictorExtensionSpec.GetStorageUri()
+}
```
njhill: Can delete this function (will be inherited)

pkg/apis/serving/v1beta1/predictor_onnxruntime.go
```diff
@@ -91,7 +91,11 @@ func (o *ONNXRuntimeSpec) GetContainer(metadata metav1.ObjectMeta, extensions *C
 }
 
 func (o *ONNXRuntimeSpec) GetStorageUri() *string {
-	return o.StorageURI
+	return o.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (o *ONNXRuntimeSpec) GetStorageSpec() *StorageSpec {
+	return o.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_paddle.go
```diff
@@ -68,7 +68,11 @@ func (p *PaddleServerSpec) GetContainer(metadata metav1.ObjectMeta, extensions *
 }
 
 func (p *PaddleServerSpec) GetStorageUri() *string {
-	return p.StorageURI
+	return p.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (p *PaddleServerSpec) GetStorageSpec() *StorageSpec {
+	return p.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_pmml.go
```diff
@@ -70,7 +70,11 @@ func (p *PMMLSpec) GetContainer(metadata metav1.ObjectMeta, extensions *Componen
 }
 
 func (p *PMMLSpec) GetStorageUri() *string {
-	return p.StorageURI
+	return p.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (p *PMMLSpec) GetStorageSpec() *StorageSpec {
+	return p.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_sklearn.go
```diff
@@ -172,7 +172,11 @@ func (k *SKLearnSpec) getDefaultsV2(metadata metav1.ObjectMeta) []v1.EnvVar {
 }
 
 func (k *SKLearnSpec) GetStorageUri() *string {
-	return k.StorageURI
+	return k.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (k *SKLearnSpec) GetStorageSpec() *StorageSpec {
+	return k.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_tfserving.go
```diff
@@ -81,7 +81,11 @@ func (t *TFServingSpec) Default(config *InferenceServicesConfig) {
 }
 
 func (t *TFServingSpec) GetStorageUri() *string {
-	return t.StorageURI
+	return t.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (t *TFServingSpec) GetStorageSpec() *StorageSpec {
+	return t.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_torchserve.go
```diff
@@ -160,7 +160,11 @@ func (t *TorchServeSpec) GetContainerV2(metadata metav1.ObjectMeta, extensions *
 }
 
 func (t *TorchServeSpec) GetStorageUri() *string {
-	return t.StorageURI
+	return t.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (t *TorchServeSpec) GetStorageSpec() *StorageSpec {
+	return t.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/apis/serving/v1beta1/predictor_triton.go
```diff
@@ -84,7 +84,11 @@ func (t *TritonSpec) GetContainer(metadata metav1.ObjectMeta, extensions *Compon
 }
 
 func (t *TritonSpec) GetStorageUri() *string {
-	return t.StorageURI
+	return t.PredictorExtensionSpec.GetStorageUri()
+}
+
+func (t *TritonSpec) GetStorageSpec() *StorageSpec {
+	return t.PredictorExtensionSpec.GetStorageSpec()
```
njhill: No need to include these functions here, they are already inherited from `PredictorExtensionSpec` ... can just delete both of them.

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -86,7 +86,7 @@ func createHPA(componentMeta metav1.ObjectMeta,
 	} else {
 		minReplicas = int32(*componentExt.MinReplicas)
 	}
-	
+
```
njhill: Revert unrelated whitespace change?

njhill (commented):
@Tomcli I'm still reviewing the rest of your latest changes, just thought I would post some low-hanging comments first which should help to reduce the size of the PR.

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -48,10 +52,32 @@ type ExplainerExtensionSpec struct {
...
+	}
+	return nil
+}
+
+// GetStorageSpec returns the predictor storage spec object
+func (e *ExplainerExtensionSpec) GetStorageSpec() *StorageSpec {
+	if e.Storage != nil {
+		return e.Storage
+	}
+	return nil
```
njhill:
```suggestion
    return e.Storage
```

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -114,6 +134,25 @@ func (s *PredictorSpec) GetExtensions() *ComponentExtensionSpec {
...
+	}
+	return nil
+}
+
+// GetStorageSpec returns the predictor storage spec object
+func (p *PredictorExtensionSpec) GetStorageSpec() *StorageSpec {
+	if p.Storage != nil {
+		return p.Storage
+	}
+	return nil
```
njhill:
```suggestion
    return p.Storage
```

pkg/apis/serving/v1beta1/predictor_xgboost.go
```diff
@@ -177,7 +178,20 @@ func (x *XGBoostSpec) getDefaultsV2(metadata metav1.ObjectMeta) []v1.EnvVar {
...
+		return &storageSpecUri
+	}
+	return nil
+}
+
+func (x *XGBoostSpec) GetStorageSpec() *StorageSpec {
+	if x.Storage != nil && x.Storage.StorageKey != nil {
+		return x.Storage
+	}
+	return nil
```
njhill: @Tomcli why is the logic in the functions above different/special for this kind of predictor?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -209,13 +211,44 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+	storageKey, storageKey_exist := pod.ObjectMeta.Annotations[constants.StorageSpecKeyAnnotationKey]
+	if !storageKey_exist {
+		storageKey = mi.config.DefaultCredentialKey
+	}
+	// Inject Storage Spec credentials if exist
+	if ok && hasStorageSpec == "true" {
+		storageSpecSecret := mi.config.StorageSpecSecret
+		var overrideParams map[string]string
+		if storageSpecParam, ok := pod.ObjectMeta.Annotations[constants.StorageSpecParamAnnotationKey]; ok {
+			err := json.Unmarshal([]byte(storageSpecParam), &overrideParams)
```
njhill: nit: can inline this check

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -209,13 +211,44 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
 	// Inject credentials
-	if err := mi.credentialBuilder.CreateSecretVolumeAndEnv(
-		pod.Namespace,
-		pod.Spec.ServiceAccountName,
-		initContainer,
-		&pod.Spec.Volumes,
-	); err != nil {
-		return err
+	hasStorageSpec, ok := pod.ObjectMeta.Annotations[constants.StorageSpecAnnotationKey]
+	storageKey, storageKey_exist := pod.ObjectMeta.Annotations[constants.StorageSpecKeyAnnotationKey]
```
njhill: nit: Better to use standard go camelcase convention e.g. `storageKeyExists`

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -209,13 +211,44 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+	if !storageKey_exist {
+		storageKey = mi.config.DefaultCredentialKey
+	}
+	// Inject Storage Spec credentials if exist
+	if ok && hasStorageSpec == "true" {
```
njhill: nit: no need for the `ok` check here

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -37,11 +37,13 @@ const (
...
-	CpuLimit      string `json:"cpuLimit"`
-	MemoryRequest string `json:"memoryRequest"`
-	MemoryLimit   string `json:"memoryLimit"`
+	Image                string `json:"image"`
+	CpuRequest           string `json:"cpuRequest"`
+	CpuLimit             string `json:"cpuLimit"`
+	MemoryRequest        string `json:"memoryRequest"`
+	MemoryLimit          string `json:"memoryLimit"`
+	StorageSpecSecret    string `json:"storageSpecSecret"`
+	DefaultCredentialKey string `json:"defaultCredentialKey"`
```
njhill: I think we should rename this to something like `DefaultStorageSecretKey` or `DefaultStorageConfigKey`

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -209,13 +211,44 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
-		pod.Namespace,
-		pod.Spec.ServiceAccountName,
-		initContainer,
-		&pod.Spec.Volumes,
-	); err != nil {
-		return err
+	hasStorageSpec, ok := pod.ObjectMeta.Annotations[constants.StorageSpecAnnotationKey]
+	storageKey, storageKey_exist := pod.ObjectMeta.Annotations[constants.StorageSpecKeyAnnotationKey]
+	if !storageKey_exist {
+		storageKey = mi.config.DefaultCredentialKey
```
njhill: Actually our thinking was to default this to a key name based on the type, for example if s3 it would be `s3_default`, etc. So probably better to just leave the key empty here (no need for the separate _exist var) and have the logic in `CreateStorageSpecSecretVolumeAndEnv` sort that out.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -37,11 +37,13 @@ const (
...
-	CpuRequest    string `json:"cpuRequest"`
-	CpuLimit      string `json:"cpuLimit"`
-	MemoryRequest string `json:"memoryRequest"`
-	MemoryLimit   string `json:"memoryLimit"`
+	Image                string `json:"image"`
+	CpuRequest           string `json:"cpuRequest"`
+	CpuLimit             string `json:"cpuLimit"`
+	MemoryRequest        string `json:"memoryRequest"`
+	MemoryLimit          string `json:"memoryLimit"`
+	StorageSpecSecret    string `json:"storageSpecSecret"`
```
njhill: Would suggest `StorageSpecSecretName` to be slightly clearer.

pkg/controller/v1beta1/inferenceservice/components/component.go
```diff
@@ -27,6 +28,23 @@ type Component interface {
 	Reconcile(isvc *v1beta1.InferenceService) error
 }
 
+func addStorageSpecAnnotations(storageSpec *v1beta1.StorageSpec, annotations map[string]string) bool {
+	if storageSpec != nil {
+		annotations[constants.StorageSpecAnnotationKey] = "true"
+		if storageSpec.Parameters != nil {
+			jsonParam, err := json.Marshal(storageSpec.Parameters)
+			if err == nil {
```
njhill: nit: can inline this if

pkg/credentials/service_account_credentials.go
```diff
@@ -62,6 +63,58 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+				})
+			}
+			if secretRegion, ok := storageDataJson["region"]; ok {
+				container.Env = append(container.Env, v1.EnvVar{
+					Name:  "AWS_DEFAULT_REGION",
+					Value: secretRegion,
+				})
+			}
+			if secretBucket, ok := storageDataJson["bucket"]; ok {
+				container.Args[0] = strings.Replace(container.Args[0], "<bucket-placeholder>", secretBucket, -1)
```
njhill: this bucket placeholder thing is kind of nasty :) but I guess it is somewhat temporary

pkg/credentials/service_account_credentials.go
```diff
@@ -62,6 +63,58 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
 	}
 }
 
+func (c *CredentialBuilder) CreateStorageSpecSecretVolumeAndEnv(namespace string, storageKey string,
+	storageSecretName string, overrideParams map[string]string, container *v1.Container, volumes *[]v1.Volume) error {
+	secret := &v1.Secret{}
```
njhill, line 78: As mentioned in other comment I think we want to change the logic here a bit:

If secret key is non-empty the look it up, otherwise get the type from `overrideParams` and based on that look up a secret key using that type name, like `s3_default`. If the secret is not found, we should only fail in the former case (where the secretKey was specified explicitly), otherwise, just make the overrideParams be the actual params.

pkg/credentials/service_account_credentials.go
```diff
@@ -62,6 +63,58 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+			return err
+		}
+		// override secret values if parameters are provided
+		for key, element := range overrideParams {
+			storageDataJson[key] = element
+		}
+		if storageDataJson["type"] == "s3" {
+			if endpointURL, ok := storageDataJson["endpoint_url"]; ok {
+				container.Env = append(container.Env, v1.EnvVar{
+					Name:  "AWS_ENDPOINT_URL",
```
njhill: We should use the existing constants for these env var values (defined in s3_secret.go).

Do we also need to set the `S3_ENDPOINT` and `S3_USE_HTTPS` env vars as is done in `s3_secret.go`?

pkg/credentials/service_account_credentials.go
```diff
@@ -62,6 +63,58 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+					Name:  "AWS_SECRET_ACCESS_KEY",
+					Value: secretAccessKey,
+				})
+			}
+			if secretRegion, ok := storageDataJson["region"]; ok {
+				container.Env = append(container.Env, v1.EnvVar{
+					Name:  "AWS_DEFAULT_REGION",
+					Value: secretRegion,
+				})
+			}
```
njhill, line 106: We can also include the "AWS_CA_BUNDLE" env var here, comes from the "certificate" parameter.

pkg/credentials/service_account_credentials.go
```diff
@@ -62,6 +63,58 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+			if accessKeyID, ok := storageDataJson["access_key_id"]; ok {
+				container.Env = append(container.Env, v1.EnvVar{
+					Name:  "AWS_ACCESS_KEY_ID",
+					Value: accessKeyID,
+				})
+			}
+			if secretAccessKey, ok := storageDataJson["secret_access_key"]; ok {
+				container.Env = append(container.Env, v1.EnvVar{
```
njhill: I just realized that the current logic uses `secretKeyRef` environment variable definitions to map those secret keys directly. We can't do that since we aren't using explicit keys for these values, but including the literal secret values here may have security implications.

So unfortunately we might need to rethink this interim approach a bit. I can't think of an alternative unfortunately to making some modification to the storage initializer - it can be passed the json from the secret in one env var and the override parameters in another, and it will have to do the merging and mapping to env vars itself upfront.

Of course that's all still "temporary" until we replace with our other puller logic, but that will hopefully be ready soon.

njhill (commented):
Thanks @Tomcli, please see my other comments. Unfortunately I came across one thing which may be a show-stopper to this approach and mean we need to move at least a bit of the logic into the storage initializer after all.

pkg/credentials/service_account_credentials.go
Tomcli: `S3_ENDPOINT` and `S3_USE_HTTPS` is for the minio python client. I think KServe changed the S3 client to boto3 in KFServing 0.5+ due to some AWS IAM is not working with minio python client.

Tomcli: So we only need `AWS_ENDPOINT_URL` in this case

pkg/credentials/service_account_credentials.go
Tomcli: done

pkg/webhook/admission/pod/storage_initializer_injector.go
Tomcli: done

pkg/webhook/admission/pod/storage_initializer_injector.go
Tomcli: done

pkg/apis/serving/v1beta1/predictor_xgboost.go
Tomcli: This was removed in the previous commit.

pkg/apis/serving/v1beta1/predictor.go
Tomcli: done

pkg/apis/serving/v1beta1/explainer.go
Tomcli: done

pkg/webhook/admission/pod/storage_initializer_injector.go
Tomcli: done

pkg/webhook/admission/pod/storage_initializer_injector.go
Tomcli: renamed to `DefaultStorageSecretKey`

pkg/webhook/admission/pod/storage_initializer_injector.go
Tomcli: done

pkg/controller/v1beta1/inferenceservice/components/component.go
Tomcli: done

pkg/credentials/service_account_credentials.go
Tomcli: I updated with your new logic, please take a look

pkg/credentials/service_account_credentials.go
Tomcli: @njhill I'm planning to pass the whole JSON value in the secret key as env variable using `secretKeyRef`. Then update the python initializer image to handle this new env variable. What do you think?

njhill: @Tomcli sounds good to me, thanks! (as already agreed on slack)

pkg/credentials/service_account_credentials.go
njhill: @Tomcli it doesn't quite cover everything that I was suggesting... might be easier for me to just push a small commit, will aim to do that soon. But most of this logic will be moving to the storage initializer anyhow right?

pkg/credentials/service_account_credentials.go
Tomcli: @njhill I updated the to let the container to obtain the whole storage secret value using `secretKeyRef`. Then update the S3/AWS environment variables inside the python server.

pkg/credentials/service_account_credentials.go
Tomcli: I kept the secret checking logics inside the golang code. If the secret is found, it will pass the secret value as a new env variable to the pod. It also passes all the `overrideParams` as another env variable to the pod if exist.

njhill:
@Tomcli I did a bit more refactoring but was unable to push to this branch so I pushed a commit to my own fork. The commit message has a summary of what's changed, PTAL!

https://github.com/Tomcli/kfserving/compare/storagespec...njhill:storagespec?expand=1

pkg/apis/serving/v1beta1/transformer_custom.go
```diff
@@ -62,6 +62,10 @@ func (c *CustomTransformer) GetStorageUri() *string {
 	return nil
 }
 
+func (c *CustomTransformer) GetStorageSpec() *StorageSpec {
```
yuzisun, line 66: So for custom transformer the storage spec is not supported? Currently it is actually supported via `STORAGE_URI` env variable

njhill: @yuzisun I'm not sure... TBH I had to look again at how custom predictors/transformers are handled. It seems to me that some of the things in `PredictorExtensionSpec` kind of belong up a level i.e. would be applicable to _any_ predictor including custom ones (at least `StorageURI`, `StorageSpec`, `ProtocolVersion`).
Similarly wouldn't it make sense to have StorageURI/Spec as strongly typed fields in the transformer spec too?

Semi-related, is the intention to keep a separate `Container` embedded in the `PredictorExtensionSpec` in addition to the `PodSpec`? Was wondering what the interplay might be with the `ServingRuntime` case where the SR might have multiple containers.

Tomcli: For the `custom` framework, because it doesn't use the `PredictorExtensionSpec`, the new storageSpec will not be reflected over here. The new `model` spec for predictor should have `PredictorExtensionSpec` for all frameworks, so we need to have the same `model` spec for explainer and transformer so that their API spec won't get diverged based on the framework.

njhill (approved):
Thanks @Tomcli, looks good to me now

pkg/apis/serving/v1beta1/transformer_custom.go
yuzisun: @njhill The original design thinking is that `predictor` maps to the `pod` level fields while `model` maps to `container` level fields, that's why we embed a single `Container` on `PredictorExtensionSpec` and does not allow user set container fields on `PodSpec` if model spec presents. We want user to only be able to customize model container and sidecar should be always managed by KServe, but I guess there might be still use case  as you said when someone wants to define SR with their own sidecar implementation.

Regarding `StorageURI`/spec, we originally had the debate of making this field strongly typed for custom predictor/transformer cases but realized the spec looks a bit weird if we level up the field.
```yaml
spec:
   predictor:
     storageUri: s3://xxx
     sklearn:
```
vs
```yaml
spec:
    predictor:
       sklearn:
           storageUri: s3://xxxx
```
As we introduced model spec, this actually look a bit better, still feel storage should be part of the model spec but maybe this is not a big deal as we do name it generically as `storage`.
```yaml
spec:
    predictor:
       storageUri: s3://xxxx
       model:
          modelFormat:
               name: sklearn
```

python/kserve/kserve/storage.py
```diff
@@ -51,6 +51,7 @@
 class Storage(object):  # pylint: disable=too-few-public-methods
     @staticmethod
     def download(uri: str, out_dir: str = None) -> str:
+        uri = Storage._update_with_storage_spec(uri)
```
yuzisun, line 54: hmm will this break user if someone is calling this API directly to download model?

yuzisun: nvm looks like it returns back the uri if `STORAGE_CONFIG` or `STORAGE_OVERRIDE_CONFIG` is not specified

pkg/credentials/service_account_credentials.go
```diff
@@ -63,6 +71,87 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+			},
+		})
+	}
+
+	if stype == "" {
+		return errors.New("unable to determine storage type")
+	}
+
+	if strings.HasPrefix(container.Args[0], UriSchemePlaceholder+"://") {
+		container.Args[0] = stype + container.Args[0][20:]
```
yuzisun: what's `container.Args[0][20:]` here?

Tomcli: This is the StorageURI without the storage type prefix, so we can replace the UriSchemePlaceholder  `<scheme-placeholder>` to stype `s3`. Sorry let me change back from `20` to `len(UriSchemePlaceholder)`

python/kserve/kserve/storage.py
Tomcli: Yes that's correct. It only updates the uri if STORAGE_CONFIG is specified, so there's no impact for the current storageURI behavior.

pkg/apis/serving/v1beta1/transformer_custom.go
njhill: OK thanks @yuzisun

I guess in my mind it makes sense details about the persisted model (format, storage location, etc.) are kind of separate to properties related to how the model is deployed (i.e. the kind of predictor, framework-specific deployment config params, pod/container customizations), and so make sense for these things to sit alongside each other. I guess this will become cleaner when we move away from the strongly typed frameworks in favour of the serving runtimes.

> I guess there might be still use case as you said when someone wants to define SR with their own sidecar implementation.

Yes it could be the case that the overall runtime implementation takes the form of multiple containers. In particular if someone wanted to integrate some existing model server that we don't currently support, they could implement their own thin adapter container to be used in conjunction with it.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -163,7 +156,15 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 		} else {
 			podSpec.Containers[0] = *container
 		}
+	}
 
+	// Knative does not support INIT containers or mounting, so we add annotations that trigger the
+	// StorageInitializer injector to mutate the underlying deployment to provision model data
+	if sourceURI := predictor.GetStorageUri(); sourceURI != nil {
+		if _, ok := annotations[constants.StorageInitializerSourceUriInternalAnnotationKey]; ok {
+			return errors.New("must provide only one of storageUri and storage.path")
```
yuzisun, line 165: Not sure if I understand this check here.

pkg/credentials/service_account_credentials.go
```diff
@@ -63,6 +71,87 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
 	}
 }
 
+func (c *CredentialBuilder) CreateStorageSpecSecretEnvs(namespace string, storageKey string,
+	storageSecretName string, overrideParams map[string]string, container *v1.Container) error {
+
+	stype, ok := overrideParams["type"]
+	if ok && stype != "s3" {
+		return errors.New("only S3 type storage is currently supported with storage spec")
```
yuzisun: I think this should be done in the validation webhook instead

pkg/credentials/service_account_credentials.go
```diff
@@ -63,6 +71,87 @@ func NewCredentialBulder(client client.Client, config *v1.ConfigMap) *Credential
...
+	}
+	if u, err := url.Parse(container.Args[0]); err == nil {
+		if stype = u.Scheme; stype != "s3" {
+			return errors.New("only S3 type storage is currently supported with storage spec")
```
yuzisun: same here let's move these to the inference service validation webhook

docs/samples/storage/storageSpec/common_secret.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+stringData:
+  localMinIO: |
+    {
+      "type": "s3",
```
yuzisun, line 5: I think the storage type should be on the `storageSpec` ?

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -81,6 +81,25 @@ type PredictorExtensionSpec struct {
 	// Each framework will have different defaults that are populated in the underlying container spec.
 	// +optional
 	v1.Container `json:",inline"`
+	// Storage Spec for model location
+	// +optional
+	Storage *StorageSpec `json:"storage,omitempty"`
+}
+
+type StorageSpec struct {
```
yuzisun, line 90: Is there any reason why storage type is not on the spec?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Tomcli: If storageSpec with path exist, the `addStorageSpecAnnotations` function at line 74 will set the uri annotation. Therefore if the storageURI is defined again, we will throw an error because there's conflict when users define both storageSpec.path and storageURI.

docs/samples/storage/storageSpec/common_secret.yaml
Tomcli: In the original modelmesh design, users don't need to be aware of the storage type because the user management service are responsible for the credential creation (in Kubeflow would be the profile controller). We still allow users to override the type here using `storage.parameters.type`. @njhill are you aware of other reasons why modelmesh stays with this design?

pkg/apis/serving/v1beta1/predictor.go
Tomcli: see https://github.com/kserve/kserve/pull/1899#discussion_r787158486

docs/samples/storage/storageSpec/common_secret.yaml
njhill: @yuzisun @Tomcli Yes that's right, the type is a parameter in the storage entries in the common secret that are referenced in the predictors via their key, so the type is generally redundant there (you could still name the storage key itself `s3` or `s3public1` or whatever). It can still be specified explicitly though as a parameter called "type", for example if you are specifying the storage completely without reference to a secret entry, or you can specify only the type in which case it will look for a storage entry named `default_<type>`.

I guess we could promote this to an optional field of the storage spec but I'm not sure we want to encourage its use. But I'm fine with it if you think we should.

pkg/credentials/service_account_credentials.go
Tomcli: I added the Storage Spec check for these 2 use cases on all the current framework. I also abstracted the framework uri validation function since most framework only validates storageUri only.

docs/samples/storage/storageSpec/common_secret.yaml
yuzisun: @njhill @Tomcli I am just thinking the public s3 bucket case where secret is not required, not sure how the storage spec can handle it as there is no storage type. 
```
kind: InferenceService
metadata:
  name: example-sklearn-isvc
spec:
  predictor:
    sklearn:
      storage:
        path: sklearn # Model path inside the bucket
        # schemaPath: null # Optional schema files for payload schema
        parameters: # Parameters to override the default values inside the common secret.
          bucket: example-models
```

yuzisun: There is also use case when user is using AWS IAM roles, secret is not used and you annotate the role name on service account.

Tomcli: I updated the python server to also take `anonymous` as the parameters. The below definition should able to run anonymous mode for aws
```
kind: InferenceService
metadata:
  name: example-sklearn-isvc
spec:
  predictor:
    sklearn:
      storage:
        path: sklearn # Model path inside the bucket
        parameters: # Parameters to override the default values inside the common secret.
          anonymous: True
```

Tomcli: For AWS IAM roles, right now users probably have to migrant the iam secret credentials to the common secret. Since S3 could have different providers (e.g. aws, ibm cloud, minio), we can look into adding some extra reconcile logics for automate the migration of s3 secrets from different providers to the common secret.

njhill: > @njhill @Tomcli I am just thinking the public s3 bucket case where secret is not required, not sure how the storage spec can handle it as there is no storage type.

@yuzisun @Tomcli sorry for delayed response, have been off sick and still recovering. The type can still be provided in the `parameters` map, so it should still work fine like this in the public s3 bucket case with no secret:
```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc
spec:
  predictor:
    sklearn:
      storage:
        path: sklearn # Model path inside the bucket
        parameters:
          type: s3
          bucket: example-models
```
As I discussed [above](https://github.com/kserve/kserve/pull/1899#discussion_r787283720) we could consider promoting this to an explicit field of the storage spec, but I'd vote against that since in _most_ cases it shouldn't be needed.

> There is also use case when user is using AWS IAM roles, secret is not used and you annotate the role name on service account.

@yuzisun part of the reason for the new approach was to standardize how the various storage parameters are configured for the various storage types, so yes like @Tomcli said the "new" way to do this would be to set these parameters in either an explicit entry in the common storage secret (referenced by key in the specs) or a default entry (which would give the same behaviour from the InfereneService spec pov). But you're right that this specific case would require values to be set in a secret when they weren't before - is that a problem though? IMO it's better for things to be consistent. Of course you could technically still avoid use of a secret by including those params explicitly in the `parameters` section of the storage spec.

As we had discussed before (and @Tomcli mentions above), we can have backwards-compatibility logic which populates/updates default entries in the new storage secret based on various different ways that the storage parameters are currently set (i.e. watches the both the service account(s) and associated secrets)

Tomcli:
Hi @yuzisun, do you think we can merge this phase 1 implementation right after the 0.8 release? The AWS IAM integration is probably similar to GCP's metacontroller that can mutate the serving pod with credentials, so it should still work with the current set of annotations. 

We are planning to have the phase 2 implementation that built on top of this with the new modelmesh pullman library that can also support GCP, HTTPS, AZURE, etc. We can probably ask users on different cloud providers to verify all the IAM integration during phase 2.

Tomcli:
In phase 1 we mainly want to introduce the storageSpec API, the S3 support is more like a preview on what it can look like.

yuzisun:
> In phase 1 we mainly want to introduce the storageSpec API, the S3 support is more like a preview on what it can look like.

Thanks @Tomcli ! Going to merge after 0.8 is released, can you also help follow up with an e2e test for the new storage spec? we can install minio on the test cluster if needed.

Tomcli:
Thanks @yuzisun, I added the e2e test and regenerated the python client sdk with the latest 0.8 release commit.

yuzisun:
Awesome work @Tomcli ! Looking forward to see the next phase of storage API unification.

/lgtm
/approve

Tomcli:
thanks @yuzisun and @njhill for all the reviews.

chintha-v:
Hi, Which apiVersion of kind "TrainedModel" has PVC support for storageUri? Please provide example yaml file if available.

PR #1901, opened 2021-11-09
Add preliminary ServingRuntime support

### LINKED ISSUE #1816: Incorporate ServingRuntime into InferenceService
/kind feature

Want to start transitioning the concept of a ServingRuntime into the InferenceService CRD and the KServe controller.

Working copy of proposal doc can be found here: https://docs.google.com/document/d/1LcnBobEaHTe_WpwqwrJG9N1zIFg61zqV1I9JErp_Eqo/edit?usp=sharing

pvaneck:
 
This PR adds preliminary Serving Runtime support as outlined in [this proposal document](https://docs.google.com/document/d/1LcnBobEaHTe_WpwqwrJG9N1zIFg61zqV1I9JErp_Eqo/edit?usp=sharing).

Example InferenceService will look like:

```yaml
kind: "InferenceService"
metadata:
  name: "sklearn-iris"
spec:
  predictor:
    model:
      framework:
        name: sklearn
        version: "0"
      # runtime: kserve-sklearnserver
      storageUri: "gs://kfserving-samples/models/sklearn/iris"
      env:
        - name: SOME_ENV_VAR
          value: foo
    minReplicas: 2
    maxReplicas: 3

```

Will be adding some tests shortly.

Followup PRs can be done to:

Fixes #1816

**Release note**:

```release-note
Add ClusterServingRuntime and ServingRuntime CRD support
```


pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -0,0 +1,193 @@
...
+	Resources       corev1.ResourceRequirements `json:"resources,omitempty"`
+	Env             []corev1.EnvVar             `json:"env,omitempty"`
+	ImagePullPolicy corev1.PullPolicy           `json:"imagePullPolicy,omitempty"`
+	WorkingDir      string                      `json:"workingDir,omitempty"`
+
+	// Periodic probe of container liveness.
+	// Container will be restarted if the probe fails.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	LivenessProbe *corev1.Probe `json:"livenessProbe,omitempty"`
```
yuzisun: We should add the readiness probe as well which is setup for all of the single model serving runtimes.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -60,6 +63,24 @@ type PredictorSpec struct {
 	ComponentExtensionSpec `json:",inline"`
 }
 
+type ModelSpec struct {
+	// Framework of the model being served.
+	Framework Framework `json:"framework,omitempty"`
```
yuzisun: @theofpa Any thoughts on the wording here, in the ML literature what should be the proper name?

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -60,6 +63,24 @@ type PredictorSpec struct {
...
+	Runtime *string `json:"runtime,omitempty"`
+
+	PredictorExtensionSpec `json:",inline"`
+}
+
+type Framework struct {
+	// +required
+	Name string `json:"name"`
+	// +optional
+	Version *string `json:"version,omitempty"`
```
yuzisun: what's the semantics of the version here? is it only the major version portion ?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -54,20 +58,82 @@ func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceService
...
 
+	// Reconcile modelConfig
+	configMapReconciler := modelconfig.NewModelConfigReconciler(p.client, p.scheme)
+	if err := configMapReconciler.Reconcile(isvc); err != nil {
+		return err
+	}
+
+	// If Model is specified, prioritize using that. Otherwise, we will assume a framework object was specified.
+	if isvc.Spec.Predictor.Model != nil {
+		runtimes, err := servingruntimes.GetRuntimesSupportingModelType(p.client, isvc.Namespace, isvc.Spec.Predictor)
```
yuzisun: I think we only need modelSpec here and the util function can be moved to a member function of `ModelSpec`, so you can call like `svc.Spec.Predictor.Model.GetRuntimesSupportingModelType(p.client, isvc.Namespace)`

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -54,20 +58,82 @@ func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceService
...
+		if err != nil {
+			return err
+		}
+		if len(runtimes) == 0 {
+			return errors.New("No runtime found to support predictor")
```
yuzisun: should this error be exposed in the status condition?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -54,20 +58,82 @@ func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceService
...
+		}
+		// Get first supporting runtime.
+		sRuntime := runtimes[0]
+		if len(sRuntime.Containers) == 0 {
+			return errors.New("No container configuration found in selected serving runtime")
```
yuzisun: same here, how do we expose this error to user if the serving runtime is misconfigured.

pkg/servingruntimes/runtime.go
```diff
@@ -0,0 +1,128 @@
...
+func (ss StringSet) Add(s string) {
+	ss[s] = exists
+}
+
+func (ss StringSet) Contains(s string) bool {
+	_, found := ss[s]
+	return found
+}
+
+func GetRuntimesSupportingModelType(cl client.Client, namespace string, p v1beta1.PredictorSpec) ([]v1alpha1.ServingRuntimeSpec, error) {
```
yuzisun: this method can be a member function of `ModelSpec`.

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: kserve/kserve-controller:latest
+      - image: pvaneckw/kserve-controller:latest
```
yuzisun: you can run `make deploy-dev` which should patch the image automatically for you

pkg/apis/serving/v1beta1/predictor.go
joerunde: This is sort of analogous to the model type parameter in model mesh, and in practice is used to determine an appropriate runtime to use if `spec.predictor.model.runtime` isn't given, right?

We have some use cases within IBM of larger model-mesh clusters that serve multiple runtimes, and multiple model types within those runtimes. We use the model type parameter to both route models to the correct runtime, and to pull up some different configs per model within each runtime. This is most important for reporting the correct size and concurrency limits for each model, so that we can have some very heterogeneous models running smoothly alongside each other.

I think "ModelType" makes more sense as a name that captures that level of granularity, but I also think it would make sense to leave something at the granularity of "Framework" or "RuntimeName" here and provide an optional "ModelType" field as well.

(Also hello, nice to meet you guys!)

pkg/apis/serving/v1beta1/predictor.go
pvaneck: This can cover multiple levels of granularity. Could just be major, major.minor, or even major.minor.patch. Currently the granularity has to exactly match a version specified in a ServingRuntime `supportedModelTypes` list, but logic can be added to add more flexibility so that an Isvc with a framework version of "1.4" could match a supportedModelType version of  just "1".

pkg/controller/v1beta1/inferenceservice/components/predictor.go
pvaneck: Did some refactoring to make these functions with ModelSpec as a receiver.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
pvaneck: Hmm, currently this is treated like the other errors returned in this function and is listed in the InferenceService events:
```bash
  Warning  InternalError  47s (x15 over 2m9s)  v1beta1Controllers  No runtime found to support predictor
```
As you can see it continues to try to reconcile with the exponential backoff which I suppose is a waste to keep retrying. Will need to investigate more to see how I can populate `InferenceServiceStatus.Conditions` and halt the reconciliation.

pkg/apis/serving/v1alpha1/servingruntime_types.go
pvaneck: Done

pkg/apis/serving/v1beta1/predictor.go
pvaneck: Thanks for your input @joerunde.  Right, this is for automatically finding a supporting ServingRuntime if `runtime` isn't explicitly given. I believe in the last community call there was an inclination towards keeping the "ModelType" naming that ModelMesh uses (cc @njhill). Wondering now if `spec.predictor.model.modelType`  is alright or is "model" repeating too much. Though having it as `spec.predictor.model.type` maybe seems ambiguous?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -58,3 +67,99 @@ func GetDeploymentMode(annotations map[string]string, deployConfig *v1beta1api.D
...
+// Get a ServingRuntime by name. First, ServingRuntimes in the given namespace will be checked.
+// If a resource of the specified name is not found, then ClusterServingRuntimes will be checked.
+func GetServingRuntime(cl client.Client, name string, namespace string) (*v1alpha1.ServingRuntimeSpec, error) {
+
+	runtime := &v1alpha1.ServingRuntime{}
+	err := cl.Get(context.TODO(), client.ObjectKey{Name: name, Namespace: namespace}, runtime)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	} else if err == nil {
+		return &runtime.Spec, nil
```
njhill: nit: could be simplified slightly (same below)
```suggestion
	if err == nil {
		return &runtime.Spec, nil
	} else if !errors.IsNotFound(err) {
		return &runtime.Spec, nil
	}
```

pkg/apis/serving/v1beta1/predictor.go
yuzisun: Let's document the field, version is used to match with the support version list from `ServingRuntime` which can be major, major.minor or major.minor.patch.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// List all cluster-scoped runtimes.
+	clusterRuntimes := &v1alpha1.ClusterServingRuntimeList{}
+	if err := cl.List(context.TODO(), clusterRuntimes); err != nil {
+		return nil, err
+	}
+
+	srSpecs := make([]v1alpha1.ServingRuntimeSpec, 0, len(runtimes.Items)+len(clusterRuntimes.Items))
+	for _, rt := range runtimes.Items {
```
njhill: nit: can avoid struct copying by using index
```go
for i := range runtimes.Items {
    rt := &runtimes.Items[i]
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -54,20 +54,82 @@ func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceService
...
 
+	// Reconcile modelConfig
+	configMapReconciler := modelconfig.NewModelConfigReconciler(p.client, p.scheme)
+	if err := configMapReconciler.Reconcile(isvc); err != nil {
+		return err
+	}
+
+	// If Model is specified, prioritize using that. Otherwise, we will assume a framework object was specified.
+	if isvc.Spec.Predictor.Model != nil {
+		runtimes, err := isvc.Spec.Predictor.Model.GetSupportingRuntimes(p.client, isvc.Namespace)
```
njhill: We should check for `Spec.Runtime` here first right? and if set use only that runtime (I guess if the type is supported we can also verify it's compatible with the specified runtime).

I see you already created a `GetServingRuntime` func for this, but it's not yet used...

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -0,0 +1,200 @@
...
+type BuiltInAdapter struct {
+	// ServerType can be one of triton/mlserver and the runtime's container must have the same name
+	ServerType ServerType `json:"serverType,omitempty"`
+	// Port which the runtime server listens for model management requests
+	RuntimeManagementPort int `json:"runtimeManagementPort,omitempty"`
+	// Fixed memory overhead to subtract from runtime container's memory allocation to determine model capacity
+	MemBufferBytes int `json:"memBufferBytes,omitempty"`
+	// Timeout for model loading operations in milliseconds
+	ModelLoadingTimeoutMillis int `json:"modelLoadingTimeoutMillis,omitempty"`
+}
```
njhill, line 152: We had been discussing internally that it would be good to also have a generic `map[string]string` parameters field in this struct to allow more fine-grained config of the built in adapter.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -0,0 +1,200 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Framework struct {
+	// +required
+	Name string `json:"name"`
+	// +optional
+	Version *string `json:"version,omitempty"`
+}
```
njhill, line 33: I was wondering whether it would make sense to include some kind of preference ordinal in this struct. To use when there are multiple runtimes satisfying a predictor's model type requirement. _Or_ maybe just a disabled flag here... so that it can be "turned off" for non-preferred runtimes without having to remove from the list.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
njhill: Agree, all "invalid spec" kind of errors should be reflected in the status. We may need to make sure however that re-reconciliation is triggered _somehow_ for some classes of error like this that e.g. involve `ServingRuntime` state. Maybe adding `ServingRuntime` as a type that the isvc controller watches, possibly with a handler to filter which set of `InferenceServices` are reconciled when a particular `ServingRuntime` changes. Or worst-case could requeue at some conservative interval I guess.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -54,20 +54,82 @@ func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceService
...
+
+		if len(isvc.Spec.Predictor.PodSpec.Containers) == 0 {
+			isvc.Spec.Predictor.PodSpec.Containers = []v1.Container{
+				*container,
+			}
+		} else {
+			isvc.Spec.Predictor.PodSpec.Containers[0] = *container
+		}
+
+		podSpec = v1.PodSpec(isvc.Spec.Predictor.PodSpec)
```
njhill: nit: could move this up a few lines so that the others can reference `podSpec` instead of `isvc.Spec.Predictor.PodSpec`?

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Configuration for this runtime's use of the storage helper (model puller)
+	// It is enabled unless explicitly disabled
+	// +optional
+	StorageHelper *StorageHelper `json:"storageHelper,omitempty"`
+
+	// Provide the details about built-in runtime adapter
+	// +optional
+	BuiltInAdapter *BuiltInAdapter `json:"builtInAdapter,omitempty"`
+}
```
njhill, line 125: We should probably make it clear that this spec is still provisional / very alpha, since there's some details still to be decided upon I think especially w.r.t. differences between single and multi-model cases.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -0,0 +1,200 @@
...
+	Env             []corev1.EnvVar             `json:"env,omitempty"`
+	ImagePullPolicy corev1.PullPolicy           `json:"imagePullPolicy,omitempty"`
+	WorkingDir      string                      `json:"workingDir,omitempty"`
+
+	// Periodic probe of container liveness.
+	// Container will be restarted if the probe fails.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	LivenessProbe  *corev1.Probe `json:"livenessProbe,omitempty"`
+	ReadinessProbe *corev1.Probe `json:"readinessProbe,omitempty"`
```
njhill, line 50: We had intentionally omitted the readiness probe here since the runtime containers are expected to conform to the KServe or model-mesh management API and this already defines status endpoints that could be used without every impl having to separately specify the probe (the controller can add the probe and/or we rely on model-mesh to report ready once it has received confirmation itself from pinging the runtime during startup): https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#health

Could consider the same thing for liveness too, since the KServe API has both `/ready` and `/live` endpoints.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -58,3 +67,99 @@ func GetDeploymentMode(annotations map[string]string, deployConfig *v1beta1api.D
...
+	coreContainer.Args = argCopy
+	coreContainer.Env = envCopy
+
+	return &coreContainer, nil
+
+}
+
+// Merge the predictor PodSpec struct with the runtime PodSpec struct, allowing users
+// to override runtime PodSpec settings from the predictor spec.
+func MergePodSpec(runtimePodSpec *v1alpha1.ServingRuntimePodSpec, predictorPodSpec *v1beta1.PodSpec) (*v1.PodSpec, error) {
```
njhill, line 123: Were you thinking we'll end up with a single PodSpec definition and if so which one? I'm trying to remember why we went with more restricted versions of the Pod and Container specs, there was some reason I think that we wanted to be able to not allow _full_ control.

njhill (commented):
Thanks @pvaneck looks like a great start

pkg/controller/v1beta1/inferenceservice/components/predictor.go
yuzisun: Reconciling all the `InferenceService` at once upon `ServingRuntime` change (especially single model case) might be disruptive, according to the code looks like we are  already updating the `InferenceService` with latest `ServingRuntime` when it gets reconciled.

Suresh-Nakkeran:
@pvaneck The inferenceservices are registered with name "model" instead of inferenceservice name.
image

pkg/apis/serving/v1alpha1/servingruntime_types.go
pvaneck: I was thinking of something similar, something like `weight` or `priority` for the ServingRuntime, but having this at the Framework/ModelType level seems to be more flexible. Would having an integer `weight` field where a higher number indicates a higher preference be fine?

pkg/apis/serving/v1alpha1/servingruntime_types.go
pvaneck: I believe this is need to support configuration for some of the existing single model serving runtimes which don't adhere to the v2 protocol. Not sure if we want to prevent these types of configuration just yet.

pkg/apis/serving/v1alpha1/servingruntime_types.go
pvaneck: Sure, this can be added. Would this be something like:
```go
Parameters map[string]string `json:"parameters,omitempty"`
```
?

pkg/apis/serving/v1beta1/predictor_model.go
pvaneck: Thanks. Will update.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
pvaneck: Yea, I can add that check to save some API calls. The current way does work as I borrowed it from ModelMesh-Serving. It's just perhaps sub-optimal for cases where a runtime is explicitly specified since it compares with all ServingRuntimes.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
pvaneck: Sure, I will do that.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
pvaneck: Thanks, this simplifies it. Though, I am assuming you meant to have `return nil, err` in the last return,

pkg/controller/v1beta1/inferenceservice/utils/utils.go
pvaneck: Right, for now just trying to allow the level of configuration that KServe currently provides since the Isvc predictor spec has the full podspec as available fields. PodSpec fields in the Isvc will override podspec fields in the SR. I realize a lot of this is more single model serving centric, as in the modelmesh case, I suppose PodSpec fields specified in the Isvc won't have any effect.

pvaneck:
@Suresh-Nakkeran Thanks for pointing that out. Hmm, one way to fix that from the Isvc spec is under `spec.predictor.model` add:
```yaml
      args:
        - --model_name=sklearn-iris
```
However, I realize that having to add this quite hampers the user experience.

Need to figure out a good way to automatically pass in the Isvc name to whatever arbitrary flag or env variable the runtime expects. the KFServer-based runtimes require that `--model-name` be passed in with the Isvc name. And I think MLServer requires that the `MLSERVER_MODEL_NAME` env var be set to the Isvc name.

Perhaps having some known placeholder/template variable be used (e.g. `${{ INFERENCESERVICE_NAME }}` or `${{ metadata.name }}`) in the ServingRuntime args/env that is processed and substituted with the InferenceService name. So args for the sklearn serving runtime might look something like:

```yaml
args:
    - --model_name=${{ INFERENCESERVICE_NAME }}
    - --model_dir=/mnt/models
    - --http_port=8080
```
But that placeholder will eventually be replaced before passing off the container spec.
@yuzisun @njhill WDYT?

Suresh-Nakkeran:
@pvaneck @yuzisun I just tried the below approach in my local that we can have default model name "model" in runtime. In Predictor reconcile, we can update model name where we update runtime container. The model name can also be part of environment.
```
// Update default model name with inferenceservice name in args and envs if it exists
isvcutils.UpdateModelName(container, isvc.ObjectMeta.Name)
```

```
func UpdateModelName(container *v1.Container, name string) {
	// update model name in args if exists
	args := container.Args
	for i, vv := range args {
		if strings.Contains(vv, "model_name") {
			args[i] = fmt.Sprintf("%s=%s", constants.ArgumentModelName, name)
			break
		}
	}

	// update model name in envs if exists
	envs := container.Env
	for i, vv := range envs {
		if vv.Name == constants.MLServerModelNameEnv {
			envs[i] = v1.EnvVar{Name: constants.MLServerModelNameEnv, Value: name}
			break
		}
	}

	container.Args = args
	container.Env = envs
	return
}
```

pvaneck:
@Suresh-Nakkeran That approach is a bit specific towards the KFServer and MLServer runtimes. Would like to keep it flexible enough to where a user can inject the inferenceservice name into the runtime even if the argument or env variable is named something else (i.e. not `--model_name` or `MLSERVER_MODEL_NAME`). This  is why I am leaning towards introducing a template variable that can be used in the ServingRuntime yaml. Though, I am open to other solutions.

pkg/apis/serving/v1alpha1/servingruntime_types.go
yuzisun: We can always generate default readiness/liveness probes for runtimes that are conforming to the v2 protocol, but leaving this gives the flexibility if user wants to define their totally customized model server.

yuzisun:
seems like there is a race condition during installation, the `ClusterServingRuntime` crd is not ready yet when creating the serving runtimes.

```
unable to recognize "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
unable to recognize "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
njhill: I _think_ the model-mesh logic only sets the runtime for placement purposes if it's provided. I don't remember if we also validate that the runtime supports the specified model type if both are provided.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -121,11 +124,6 @@ func (m *ModelSpec) runtimeSupportsModel(runtimeName string, srSpec *v1alpha1.Se
 }
 
 func (m *ModelSpec) getModelTypeLabel() string {
-	if m.Runtime != nil {
-		// constrain placement to specific runtime
-		return "rt:" + *m.Runtime
-	}
-	// constrain placement based on model type
```
njhill: @pvaneck why this change and the one below? This is the logic I was referring to in model-mesh case where we use the runtime for placement if it's provided.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -75,15 +78,30 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+				return fmt.Errorf("Specified runtime %s does not support specified framework/version", *isvc.Spec.Predictor.Model.Runtime)
+			}
+			sRuntime = *r
+		} else {
+			runtimes, err := isvc.Spec.Predictor.Model.GetSupportingRuntimes(p.client, isvc.Namespace)
+			if err != nil {
+				return err
+			}
+			if len(runtimes) == 0 {
+				return errors.New("No runtime found to support predictor")
```
njhill: Would be good to include the model type info in the error message here

yuzisun:
> @Suresh-Nakkeran That approach is a bit specific towards the KFServer and MLServer runtimes. Would like to keep it flexible enough to where a user can inject the inferenceservice name into the runtime even if the argument or env variable is named something else (i.e. not `--model_name` or `MLSERVER_MODEL_NAME`). This is why I am leaning towards introducing a template variable that can be used in the ServingRuntime yaml. Though, I am open to other solutions.

ye template approach(e.g go template) sounds good to me. @pvaneck, @Suresh-Nakkeran is working on a follow up PR after this is merged in, so I think he can help implement the template there.

pkg/apis/serving/v1beta1/predictor_model.go
pvaneck: In the predictor reconciler, I had added the check for `Runtime` being explicitly given. If it is, then I perform the call to `GetServingRuntime` and do checks on what is retrieved. 

Since that runtime checking logic was added in the predictor reconciler, I shifted the focus of these functions to only be concerned with model type labels. Currently, these are used to verify a runtime supports a predictors model type which needed the runtime checking logic here to be removed (otherwise `RuntimeSupportsModel` would always return true because of the `rt:` labels.

pvaneck:
> ye template approach(e.g go template) sounds good to me. @pvaneck, @Suresh-Nakkeran is working on a follow up PR after this is merged in, so I think he can help implement the template there.

Sounds good. Can try to get this one in asap so follow-up PRs and enhancements can be performed in parallel. Trying to get the E2E tests to pass. 

I believe a list of some of the to-dos are:

1. Add template variable support for enabling passing in isvc name to runtime container.
2. Add more default ClusterServingRuntimes
3. Adjust the InferenceService mutator to convert old schema into the new schema.
4. Add ability to disable a Framework/ModelType, in a ServingRuntime and add associated logic?
6. Add E2E tests using new format
7. Eventually add default ClusterServingRuntimes as a part of the default installation (perhaps generate separate yaml file for default runtimes as including the runtimes in the main default install causes that race condition with the CRD not existing yet).
8. Apply ServingRuntimes to transformer/explainer cases.

Also, wasn't quite sure where we landed with the Framework vs ModelType naming inside the ModelSpec. ModelMesh uses ModelType, but I currently still have it as Framework here. Does `predictor.model.modelType` make sense?

pvaneck:
Need to figure out why E2E tests are failing with:
```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Internal error occurred: failed calling webhook \"inferenceservice.kserve-webhook-server.defaulter\": Post \"https://kserve-we
bhook-server-service.kserve.svc:443/mutate-serving-kserve-io-v1beta1-inferenceservice?timeout=10s\": no endpoints available for service \"kserve-webhook-server-service\"","reason":"InternalError","details":{"causes":[{"message":"failed c
alling webhook \"inferenceservice.kserve-webhook-server.defaulter\": Post \"https://kserve-webhook-server-service.kserve.svc:443/mutate-serving-kserve-io-v1beta1-inferenceservice?timeout=10s\": no endpoints available for service \"kserve
-webhook-server-service\""}]},"code":500}
```

yuzisun:
> Need to figure out why E2E tests are failing with:
> 
> ```
> HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Internal error occurred: failed calling webhook \"inferenceservice.kserve-webhook-server.defaulter\": Post \"https://kserve-we
> bhook-server-service.kserve.svc:443/mutate-serving-kserve-io-v1beta1-inferenceservice?timeout=10s\": no endpoints available for service \"kserve-webhook-server-service\"","reason":"InternalError","details":{"causes":[{"message":"failed c
> alling webhook \"inferenceservice.kserve-webhook-server.defaulter\": Post \"https://kserve-webhook-server-service.kserve.svc:443/mutate-serving-kserve-io-v1beta1-inferenceservice?timeout=10s\": no endpoints available for service \"kserve
> -webhook-server-service\""}]},"code":500}
> ```

@pvaneck this is the crash stack from controller

```
E1122 09:46:23.242494       1 runtime.go:78] Observed a panic: runtime.boundsError{x:0, y:0, signed:true, code:0x0} (runtime error: index out of range [0] with length 0)
goroutine 511 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1aff6e0, 0xc00116a540)
        /go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/runtime/runtime.go:48 +0x82
panic(0x1aff6e0, 0xc00116a540)
        /usr/local/go/src/runtime/panic.go:969 +0x166
github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/components.(*Predictor).Reconcile(0xc001152940, 0xc0006b9800, 0x13, 0x2010a60)
        /go/src/github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/components/predictor.go:171 +0x1b01
github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice.(*InferenceServiceReconciler).Reconcile(0xc000745f40, 0x2004c60, 0xc00086ce10, 0xc000a2a040, 0x12, 0xc000a2a020, 0
x14, 0xc00086ce10, 0x40941b, 0xc00084e000, ...)
        /go/src/github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/controller.go:166 +0x622
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000a8fe00, 0x2004ba0, 0xc000cccb00, 0x1a4b5c0, 0xc0009e73e0)
```

pkg/apis/serving/v1alpha1/servingruntime_types.go
njhill: WDYT about just a flag to disable? I'd guess the normal case would be where it's known which the preferred runtime is for a given kind of model, so maybe better to be explicit rather than based on an indirect weight comparison?

pvaneck: Hmm, yea, I see what you mean and agree that it is better to be explicit. I am okay with just a disabled flag . As you said, it'd be useful for turning off a model type without removing it from the supported list (which would indicate that the runtime doesn't support a framework/version).

yuzisun: @pvaneck I think `framework` is ok as I can see a few references using the same wording e.g
https://azure.microsoft.com/en-us/overview/ml-frameworks.

pkg/apis/serving/v1beta1/predictor.go
yuzisun: @pvaneck I think `framework` is ok as I can see a few references using the same wording e.g
https://azure.microsoft.com/en-us/overview/ml-frameworks.

yuzisun:
Thanks @pvaneck ! This is awesome work !
/lgtm
/approve

yuzisun:
We can follow up the work in https://github.com/kserve/kserve/issues/1924

PR #1905, opened 2021-11-11
Made changes to use probe port injected from agent injector

### LINKED ISSUE #1536: When using batcher in a custom inference service the the containerPort isn't respected
/kind bug

```
kind: InferenceService
metadata:
  name: bran-relation-bento-batched
  namespace: kfserving
spec:
  predictor:
    minReplicas: 1
    batcher:
      maxBatchSize: 32
      maxLatency: 5000
      timeout: 60
    containers:
    - image: foo.dkr.ecr.us-east-1.amazonaws.com/bran-relation-bento:20210327T151131-0700
      name: bento
      ports:
        - containerPort: 5000
          protocol: TCP
```

With that inference service deployed the queue proxy still tries to use port 8080

It should use port 500 as it's specified in the containerPort argument

Making the server listen on port 8080 rescues the service.

**Environment:**

- Istio Version:
- Knative Version: 0.22.0
- KFServing Version: 0,5,1
- Kubeflow version: ??
- Minikube version:

andyi2it:
Made changes to use probe port injected from agent injector

Fixes #1536 

**Release note**:

```release-note
NONE
```


pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -196,23 +197,59 @@ func (ag *AgentInjector) InjectAgent(pod *v1.Pod) error {
...
+		if container.Name == "kserve-container" {
+			containerPort = fmt.Sprint((container.Ports[0].ContainerPort))
 		}
 	}
 
-	if !queueProxyAvailable {
-		readinessProbeJson, err := json.Marshal(pod.Spec.Containers[0].ReadinessProbe)
-		if err != nil {
-			return err
+	containerReadinessProbe := &v1.Probe{
```
yuzisun: Why do we need to define the readiness probe here? 
- For serverless mode readiness probe can be specified on kserve container and queue proxy gets that via Env `SERVING_READINESS_PROBE`.
- For native mode,  we inject the `SERVING_READINESS_PROBE` env for model agent container.

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -196,23 +197,59 @@ func (ag *AgentInjector) InjectAgent(pod *v1.Pod) error {
...
+		return err
+	}
+
+	for i, envVar := range queueProxyEnvs {
+
+		if envVar.Name == "SERVING_READINESS_PROBE" {
+			readinessProbeAvailable = true
+			envVar.Value = string(readinessProbeJson)
+			queueProxyEnvs[i] = envVar
+			queueProxyEnvs = append(queueProxyEnvs, v1.EnvVar{Name: "COMPONENT_PORT", Value: containerPort})
```
yuzisun: We also need to set the component port for the RawDeployment mode right?

cmd/agent/main.go
```diff
@@ -63,6 +63,7 @@ type config struct {
 	QueueServingPort       int    `split_words:"true"`
 	UserPort               int    `split_words:"true"`
 	RevisionTimeoutSeconds int    `split_words:"true"`
+	ComponentPort          int    `split_words:"true"`
```
yuzisun: any reason change to use the env instead of argument on line 36?

pkg/webhook/admission/pod/agent_injector_test.go
```diff
@@ -152,7 +152,7 @@ func TestAgentInjector(t *testing.T) {
 								},
 							},
 							Args: []string{"--enable-puller", "--config-dir", "/mnt/configs", "--model-dir", "/mnt/models"},
-							Env:  []v1.EnvVar{{Name: "SERVING_READINESS_PROBE", Value: "{\"tcpSocket\":{\"port\":8080},\"timeoutSeconds\":1,\"periodSeconds\":10,\"successThreshold\":1,\"failureThreshold\":3}"}},
+							Env:  []v1.EnvVar{{Name: "SERVING_READINESS_PROBE", Value: "{\"tcpSocket\":{\"port\":\"8080\"},\"timeoutSeconds\":1,\"periodSeconds\":10,\"successThreshold\":1,\"failureThreshold\":3}"}},
```
yuzisun: Let's add a test case for the non default container port

cmd/agent/main.go
andyi2it: Its not getting passed as argument. So used environment. We will discuss it in chat.

test/e2e/batcher/test_batcher_custom_port.py
```diff
@@ -0,0 +1,80 @@
...
+    predictor = V1beta1PredictorSpec(
+        batcher=V1beta1Batcher(
+            max_batch_size=32,
+            max_latency=5000,
+        ),
+        min_replicas=1,
+        pytorch=V1beta1TorchServeSpec(
+            args=["--http_port=5000"],
+            storage_uri='gs://kfserving-samples/models/pytorch/cifar10',
+            model_class_name='Net',
```
yuzisun: @andyi2it can you help rebase Jagadeesh’s recent change for torchserve ? The torchserve test has been rewritten.

andyi2it: Sure...

test/e2e/batcher/test_batcher_custom_port.py
```diff
@@ -0,0 +1,79 @@
...
+def test_batcher_custom_port():
+    service_name = 'isvc-pytorch-batcher-custom'
+    predictor = V1beta1PredictorSpec(
+        batcher=V1beta1Batcher(
+            max_batch_size=32,
+            max_latency=5000,
+        ),
+        min_replicas=1,
+        pytorch=V1beta1TorchServeSpec(
+            args=["--http_port=5000"],
```
yuzisun, line 48: I think the torchserve actually does not support this argument, we can change to use sklearnserver instead.

andyi2it: Yes, made changes to use `env=[V1EnvVar(name='TS_INFERENCE_ADDRESS', value='"http://127.0.0.1:5000"')]`
but it only works if config.properties entry is added to use environment variables... Let me see if there is another way, if not will use sklearn...

pkg/webhook/admission/pod/agent_injector.go
yuzisun: @andyi2it Can you help remind me why we still need to define the readiness probe here ?

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -196,19 +197,52 @@ func (ag *AgentInjector) InjectAgent(pod *v1.Pod) error {
...
+	readinessProbeJson, err := json.Marshal(containerReadinessProbe)
+	if err != nil {
+		return err
+	}
+
+	for i, envVar := range queueProxyEnvs {
+
+		if envVar.Name == "SERVING_READINESS_PROBE" {
+			readinessProbeAvailable = true
+			envVar.Value = string(readinessProbeJson)
```
yuzisun: So this will overwrite the user defined readiness probe?

pkg/webhook/admission/pod/agent_injector.go
yuzisun: so the env is injected correctly with the port already 
```yaml
    - name: USER_PORT
      value: "5000"
    - name: SERVING_READINESS_PROBE
      value: '{"httpGet":{"path":"/v1/models/hello-lambda","port":5000,"host":"127.0.0.1","scheme":"HTTP","httpHeaders":[{"name":"K-Kubelet-Probe","value":"queue"}]},"successThreshold":1}'
```

pkg/webhook/admission/pod/agent_injector.go
andyi2it: Not exactly sure, looks like it from the code. Let me check, should be overriding only when default probe is attached.

pkg/controller/v1beta1/inferenceservice/components/component.go
```diff
@@ -78,7 +78,7 @@ func addAgentAnnotations(isvc *v1beta1.InferenceService, annotations map[string]
 func addAgentContainerPort(container *v1.Container) {
```
yuzisun: I think this function can be deleted now

pkg/webhook/admission/pod/agent_injector.go
yuzisun: Here actually mutates the queue proxy's `SERVING_READINESS_PROBE` and `USER_PORT` as `queueProxyEnvs` holds a reference to the queue proxy's env vars. As we are now inject agent between `queue-proxy` and `kserve-container`, `queue-proxy` should send probes to agent while agent sends probes to `kserve-container`.

yuzisun: We can take a copy of `queueProxyEnvs` and that should be what's set on the agent container before we mutate the `queue-proxy` envs

test/e2e/batcher/test_batcher_custom_port.py
```diff
@@ -0,0 +1,85 @@
...
+
+    predictor = V1beta1PredictorSpec(
+        batcher=V1beta1Batcher(
+            max_batch_size=32,
+            max_latency=5000,
+        ),
+        min_replicas=1,
+        sklearn=V1beta1SKLearnSpec(
+            args=["--http_port=5000"],
+            storage_uri="gs://kfserving-samples/models/sklearn/iris",
```
yuzisun: we have upgraded the sklearn version, so we should use `gs://kfserving-examples/models/sklearn/1.0/model` now.

yuzisun:
@andyi2it can you help rebase the PR?

cmd/agent/main.go
```diff
@@ -275,11 +275,16 @@ func buildProbe(logger *zap.SugaredLogger, probeJSON string) *readiness.Probe {
...
+	newProbe := readiness.NewProbe(coreProbe)
+	newProbe.InitialDelaySeconds = 10
+	return newProbe
 }
 
 func buildServer(ctx context.Context, port string, userPort string, loggerArgs *loggerArgs, batcherArgs *batcherArgs,
 	probeContainer func() bool, logging *zap.SugaredLogger) (server *http.Server, drain func()) {
+
+	logging.Info("Building server ", userPort)
+	logging.Info("Port ", port)
```
yuzisun: I think you need to use` logging.Infof` to format the log here, the two lines can also be combined into one.

cmd/agent/main.go
```diff
@@ -275,11 +275,16 @@ func buildProbe(logger *zap.SugaredLogger, probeJSON string) *readiness.Probe {
 		logger.Fatalw("Agent failed to parse readiness probe", zap.Error(err))
 		panic("Agent failed to parse readiness probe")
 	}
-	return readiness.NewProbe(coreProbe)
+	newProbe := readiness.NewProbe(coreProbe)
+	newProbe.InitialDelaySeconds = 10
```
yuzisun: just curious the reason to add `InitialDelaySeconds` here

andyi2it: There were errors in logs where it was aggressively probing, so wanted to give time for the service to be fully up...
The probe becomes successful on average after 10 seconds, that's why...

cmd/agent/main.go
andyi2it: Sure will do.

cmd/agent/main.go
yuzisun: This can be configured on the inference service yaml so might need to check if it is set then give this default value.

cmd/agent/main.go
```diff
@@ -275,11 +275,15 @@ func buildProbe(logger *zap.SugaredLogger, probeJSON string) *readiness.Probe {
...
+	return newProbe
 }
 
 func buildServer(ctx context.Context, port string, userPort string, loggerArgs *loggerArgs, batcherArgs *batcherArgs,
 	probeContainer func() bool, logging *zap.SugaredLogger) (server *http.Server, drain func()) {
+
+	// logging.Infof("Building server user port %s port %s", userPort, port)
```
yuzisun: Let's remove this line if it is not necessary

andyi2it: It would be better to have the log, just commented to debug test failure coz this was the only change.

PR #1910, opened 2021-11-13
Initial inference graph API implementation

### LINKED ISSUE #846: Multi stage model inference pipelines
/kind feature
In many situations, inferenceservice is contained many small services. If we want to deploy this big servcie by KFServing, we should create a server to connect these small services together, and use `custom` model to deploy this big service. If we want to change the connection, we should recode the server and rebuild the image. If KFServing can handle this by edit the deploy yaml, it'll save much developer time.
```
                                                            +-----------------+
                                                            |   big-service   |
                                                            |                 |
                                                        +--->                 +-----+
          +---------------+        +----------------+   |   |    step 3-1/4   |     |      +-----------------+
          |  big-service  |        | big-service    |   |   |                 |     |      |    big-service  |
          |               |        |                |   |   +-----------------+     |      |                 |
          |               +------->+                +---+                           +----->+                 |
          |   step 1/4    |        |    step 2/4    |   |   +-----------------+     |      |                 |
          |               |        |                |   |   |   big-service   |     |      |    step 4/4     |
          +---------------+        +----------------+   |   |                 |     |      |                 |
                                                        +--->                 +-----+      +-----------------+
                                                            |   step 3-2/4    |
                                                            |                 |
                                                            +-----------------+
```

yuzisun:

Fixes #846 

Please see the detailed proposal in https://docs.google.com/document/d/13VHfOxa72pgoy5Eg5c-gGHZfEL7uBDc7rTLa2pAi4Ko

1. tested examples

```yaml
---
kind: "InferenceService"
metadata:
  name: "sklearn-iris"
spec:
  predictor:
    sklearn:
---
kind: "InferenceService"
metadata:
  name: "xgboost-iris"
spec:
  predictor:
    xgboost:
      storageUri: "gs://kfserving-examples/models/xgboost/iris"
---
kind: InferenceGraph
metadata:
  name: ensemble-model
spec:
  nodes:
    ensembleModel:
      routerType: Ensemble
      routes:
        - service: sklearn-iris
        - service: xgboost-iris
---
kind: InferenceGraph
metadata:
  name: model-chainer
spec:
  nodes:
    root:
      routerType: Single 
      routes:
        - service: sklearn-iris
      nextRoutes:
        - nodeName: treeModel
          data: $request
    treeModel:
      routerType: Single
      routes:
        - service: xgboost-iris
---
kind: InferenceGraph
metadata:
  name: splitter-model
spec:
  nodes:
    TrafficSplit:
      routerType: Splitter 
      routes:
        - service: sklearn-iris
          weight: 20
        - service: xgboost-iris
          weight: 80
```

**Release note**:

```release-note
alpha version of inference graph with basic model chaining, traffic splitting, ensemble capability.
```


pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -0,0 +1,266 @@
...
+	graph := &v1alpha1api.InferenceGraph{}
+	if err := r.Get(ctx, req.NamespacedName, graph); err != nil {
+		if apierr.IsNotFound(err) {
+			// Object not found, return.  Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
```
pugangxa, line 112: Question:
Is there any validation on the graph configuration? I mean should the service url being existing inferenceservice endpoint or just any string is accepted? 
Also I saw in the router image there's no validation too so it's just behave like proxy and user need make the request following our request spec? 
In fact if there's no any validation user may chain any knative service and make calls as long as the request is valid for the service accordingly. That may be an additional flexibility.

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		r.Log.Error(err, "Failed to find config map", "name", constants.InferenceServiceConfigMapName)
+		return reconcile.Result{}, err
+	}
+	routerConfig, err := getRouterConfigs(configMap)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	// Create service if does not exist
+	desired := createKnativeService(graph.ObjectMeta, graph, routerConfig)
```
pugangxa: Before we already dropped the hard requirement for knative by deployment and HPA, so for graph do we consider add adaptation to this? Or I think this will add back the dependency on knative.

pkg/controller/v1alpha1/inferencegraph/controller.go
yuzisun: The idea is to support both `InferenceService` name and url, if `InferenceService` name is provided then the controller fills in the url from the status object. Will also add the graph validation and graph status based on the individual inference service status.

Graph orchestrator serves as the proxy and  I think the request validation should be done in the destination service and the proxy can propagate the error back.

pkg/controller/v1alpha1/inferencegraph/controller.go
yuzisun: We still support both Knative and raw deployment mode, will add the raw deployment handling as well.

pugangxa: Yeah, great!

pkg/controller/v1alpha1/inferencegraph/controller.go
pugangxa: That's clear to me now, thanks for response and look forward to the feature.

config/rbac/role.yaml
```diff
@@ -264,3 +264,23 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - serving.kubeflow.org
```
Iamlovingit: `apiGroups` should be `serving.kserve.io`

rachitchauhan43:
@yuzisun : When is this PR scheduled to get merged ? Will it be part of v0.9 release ?

cmd/router/main.go
```diff
@@ -0,0 +1,165 @@
...
+		if route == nil {
+			return fmt.Errorf("The pickuped servcie is nil.")
+		}
+		result := make(chan []byte)
+		go callService(route.ServiceUrl, input, result)
+		responseBytes := <-result
+		var res map[string]interface{}
+		json.Unmarshal(responseBytes, &res)
+		response = res
+	} else if currentStep.RouterType == v1alpha1.Switch {
```
yuzisun: This has not been implemented yet, we plan to use [k8s.io/client-go/util/jsonpath](http://k8s.io/client-go/util/jsonpath) to evaluate the jsonpath and route the request based on condition

pkg/controller/v1alpha1/inferencegraph/controller.go
yuzisun: if `InferenceService` name is provided we resolve to the service url and in the router it is always use service url to invoke the request.

yuzisun:
> @yuzisun : When is this PR scheduled to get merged ? Will it be part of v0.9 release ?

@rachitchauhan43 this PR is ready for review, please help test out if you are interested, the plan is to release the alpha version in 0.9.

rachitchauhan43:
> > @yuzisun : When is this PR scheduled to get merged ? Will it be part of v0.9 release ?
> 
> @rachitchauhan43 this PR is ready for review, please help test out if you are interested, the plan is to release the alpha version in 0.9.

@yuzisun : Happy to help out in testing this. Where can I start at to test it out ? Or any particular scenarios documented to test it out ?

yuzisun:
@njhill @pvaneck Would like to get your feedback on this.

Makefile
```diff
@@ -183,9 +193,15 @@ docker-push:
 docker-build-agent:
 	docker build -f agent.Dockerfile . -t ${KO_DOCKER_REPO}/${AGENT_IMG}
 
+docker-build-router:
+	docker build -f router.Dockerfile . -t ${KO_DOCKER_REPO}/${ROUTER_IMG}
```
pvaneck, line 202: Doesn't look like `router.Dockerfile` was added?

cmd/manager/main.go
```diff
@@ -158,6 +159,20 @@ func main() {
...
 
+	//Setup Inference graph controller
+	inferenceGraphEventBroadcaster := record.NewBroadcaster()
+	setupLog.Info("Setting up InferenceGraph controller")
+	inferenceGraphEventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: clientSet.CoreV1().Events("")})
+	if err = (&graphcontroller.InferenceGraphReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("v1alpha1Controllers").WithName("InferenceGraph"),
+		Scheme: mgr.GetScheme(),
+		Recorder: eventBroadcaster.NewRecorder(mgr.GetScheme(), v1.EventSource{Component: "InferenceGraph	Controller"}),
```
pvaneck: Should the EventSource Component be `v1alpha1Controllers`?

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+KServe inference graph is designed for this.
+
+## **2. KServe Inference Graph** 
+![image](graph.png)
+### **2.1 Inference Graph**
+As above image shows, inference graph is made up of a list of `nodes` , and each `node` consists of several `isvcs`. Every graph must have a root node named `root`. When an inference request hits the graph it excutes the `root` node from the DFS, if the graph has other `nodes`, it will pass the `$request` or `$response` of the root service as input data to the `next node`. There are four types of `node` are supported: ***Single***, ***Switch***, ***Ensemble***, ***Splitter***.
```
pvaneck:
```suggestion
As the above image shows, an inference graph is made up of a list of `nodes`, and each `node` consists of several `isvcs`. Every graph must have a root node named `root`. When an inference request hits the graph, it executes the `root` node from the DAG. If the graph has other `nodes`, it will pass the `$request` or `$response` of the root service as input data to the `next node`. There are four `node` types that are supported: ***Single***, ***Switch***, ***Ensemble***, ***Splitter***.
```

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+
+### **2.2 Single Node**
+**Single Node** makes user can connect 2 isvcs in a sequence relationship. The
```
pvaneck:
```suggestion
**Single Node** allows users to connect two `isvcs` in a sequence relationship. The
```

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+  routerType: Single 
+  routes:
+  - service: isvc1
+  nextRoutes:
+  - nodeName: isvc2
+    data: $request
+...
+```
+### **2.3 Switch Node**
+**Switch Node** makes user can select an `isvc` to handle the request by setting the `condition`. Usually, user doesn't need to set the `serviceURL`, kserve will fill it with `isvc.status.address.URL` of `isvc`, but if you want to specified the `serviceURL`, you can set it manually.
```
pvaneck:
```suggestion
**Switch Node** allows users to select an `isvc` to handle the request by setting the `condition`. Usually, users don't need to set the `serviceURL` as KServe will fill it in with `isvc.status.address.URL` of the `isvc`, but if desired, `serviceURL` can be specified explicitly.
```

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+![image](ensembleNode.png)
+```yaml
+root:
+  routerType: Ensemble
+  routes:
+  - service: sklearn-iris
+  - service: xgboost-iris
+```
+### **2.5 Splitter Node**
+**Splitter Node** make user can
```
pvaneck: Looks like a description here was mistakenly left out? I'm assuming this would say something like:

**Splitter Node** allows users to split traffic to multiple `isvcs` using a weighted distribution

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+kubectl get isvc
+NAME              URL                                                    READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                       AGE
+sklearn-iris      http://sklearn-iris.default.10.166.15.29.sslip.io      True           100                              sklearn-iris-predictor-default-00001      80m
+xgboost-iris      http://xgboost-iris.default.10.166.15.29.sslip.io      True           100                              xgboost-iris-predictor-default-00001      80m
+
+kubectl get ig
+NAME            URL                                                  READY   AGE
+model-chainer   http://model-chainer.default.10.166.15.29.sslip.io   True    5s
+```
+3. Tesing `graph`.
```
pvaneck: s/Tesing/Testing/

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+kubectl get isvc
+NAME              URL                                                    READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                       AGE
+blue              http://blue.default.10.166.15.29.sslip.io              True           100                              blue-predictor-default-00002              3d7h
+green             http://green.default.10.166.15.29.sslip.io             True           100                              green-predictor-default-00002             3d7h
+
+kubectl get ig
+NAME            URL                                                  READY   AGE
+model-switch     http://model-switch.default.10.166.15.29.sslip.io     True    35s
+```
+3. Tesing `graph`.
```
pvaneck: s/Tesing/Testing/

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+```shell
+curl http://model-switch.default.10.166.15.29.sslip.io
```
pvaneck, line 119: I think an example where the user passes in an explicit target (i.e. target == "blue") would be more beneficial in terms of showcasing usage.

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+kubectl get isvc
+NAME              URL                                                    READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                       AGE
+sklearn-iris      http://sklearn-iris.default.10.166.15.29.sslip.io      True           100                              sklearn-iris-predictor-default-00001      80m
+xgboost-iris      http://xgboost-iris.default.10.166.15.29.sslip.io      True           100                              xgboost-iris-predictor-default-00001      80m
+
+kubectl get ig
+NAME            URL                                                  READY   AGE
+splitter-model   http://splitter-model.default.10.166.15.29.sslip.io   True    15m
+```
+3. Tesing `graph`.
```
pvaneck: s/Tesing/Testing/

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+kubectl get isvc
+NAME              URL                                                    READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                       AGE
+sklearn-iris      http://sklearn-iris.default.10.166.15.29.sslip.io      True           100                              sklearn-iris-predictor-default-00001      80m
+xgboost-iris      http://xgboost-iris.default.10.166.15.29.sslip.io      True           100                              xgboost-iris-predictor-default-00001      80m
+
+kubectl get ig
+NAME            URL                                                  READY   AGE
+ensemble-model   http://ensemble-model.default.10.166.15.29.sslip.io   True    15m
+```
+3. Tesing `graph`.
```
pvaneck: s/Tesing/Testing/

docs/samples/graph/sequence.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: model-chainer
+spec:
+  nodes:
+    root:
+      routerType: Single 
+      routes:
+        - service: sklearn-iris
+      nextRoutes:
+        - nodeName: treeModel
+          data: $request
```
pvaneck, line 31: Would be cool to also have a sample where root.nextRoutes.data is $response instead of $request (doesn't need to be this PR)

pvaneck (commented):
Hey, @yuzisun. This looks great so far. Did an initial review with some comments inline. 

Also noticed that there are several line changes to the servingruntime and isvc crd yamls. I'm assuming because of differing versions of yq shifting the indentation when `make manifests` is run, creating quite huge diffs. In the storage spec PR a while back, an older version of yq was used in the generation, but now, it looks like a newer version of 'yq' is used, re-shifting the indentation.

Should we have the crd regeneration done in a separate PR to keep this PR cleaner? And should we do something like we do with ModelMesh-Serving where we [perform](https://github.com/kserve/modelmesh-serving/blob/main/scripts/develop.sh) tests/generation using a [developer image](https://github.com/kserve/modelmesh-serving/blob/main/Dockerfile.develop) with correct dependencies installed?

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+```shell
+curl http://ensemble-model.default.10.166.15.29.sslip.io -d @./iris-input.json
+``` 
+***Expect result***
+```shell
+{"http://sklearn-iris.default.svc.cluster.local/v2/models/sklearn-iris/infer":{"predictions":[1,1]},"http://xgboost-iris.default.svc.cluster.local/v2/models/xgboost-iris/infer":{"predictions":[1,1]}}
```
rachitchauhan43: Why keys for predictions are complete service URLs ? Why not just use the service name ike `sklearn-iris` and `xgboost-iris` as the key ?
Won't output be more readable like that ? 
```
{
  "sklearn-iris": {
    "predictions": [
      1,
      1
    ]
  },
  "xgboost-iris": {
    "predictions": [
      1,
      1
    ]
  }
}
```

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+	"math/rand"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	flag "github.com/spf13/pflag"
+)
+
+var log = logf.Log.WithName("InferenceGraphRouter")
+
+func callService(serviceUrl string, input []byte, res chan<- []byte) error {
+	resp, err := http.Post(serviceUrl, "application/json", bytes.NewBuffer(input))
```
rachitchauhan43, line 42: I created an IG with a non-ISVC node and IG was able to invoke that.
```
kind: InferenceGraph
metadata:
  name: ext-service-caller
spec:
  nodes:
    root:
      routerType: Single
      routes:
        - serviceUrl: http://helloworld.kserve-test.svc.cluster.local:5000/hello
```

So, one thing we observed while invoking external services as a node in IG is that, IG will only make calls assuming those nodes are ISVCs hence the HTTP verb is always POST.(https://github.com/yuzisun/kserve/blob/graph/cmd/router/main.go#L42)

But think about scenarios where, someone would want to invoke some external services like AWS or Google cloud services or in-house feature store service they have in their org that has other semantics like HTTP GET to fetch data.

In that case what would be the right solution ? One solution I see is, creating a pseudo-isvc that internally just makes calls to other services.

Can we think of something more general ? Or allowing extending how the node should be invoked ?

docs/samples/graph/README.md
rachitchauhan43: Think about the scenario when ensemble’s output is passed onto downstream model in the graph. Then working on the json with URL as keys is not going to a great developer/user experience.

cmd/router/main.go
njhill: I remember someone saying that we might use some other json expression library for this since k8s jsonpath is quite limited?

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil
+	}
+	// process outgoing edges
+	jobs := map[string]chan []byte{}
+	for _, routeTo := range currentStep.NextRoutes {
+		job := make(chan []byte)
+		jobs[routeTo.NodeName] = job
+		if router, ok := graph.Nodes[routeTo.NodeName]; ok {
+			if routeTo.Data == "$request" {
+				go routeStep(routeTo.NodeName, router, graph, input, job)
```
njhill: This case seems odd, basically we must be throwing away the result of the prior step (`jsonRes`) so why perform it at all? (unless it has some other side effect like payload logging, maybe that's reasonable)

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+		return err
+	}
+
+	if len(currentStep.NextRoutes) == 0 {
+		res <- jsonRes
+		return nil
+	}
+	// process outgoing edges
+	jobs := map[string]chan []byte{}
+	for _, routeTo := range currentStep.NextRoutes {
```
njhill: The multiple nextRoutes logic here doesn't appear to be documented. The docs/examples in this PR imply there should only be a single one.

Reading what is implemented here, it looks similar/equivalent to the ensemble case, except that the parallel branches are other nodes rather than services, and there's no way to do any kind of post-processing of the result (I guess it would have to be the output of the inference graph as a whole).

docs/samples/graph/sequence.yaml
njhill: Related to my other comment in the code corresponding to this, what will happen to the output (response) from the `sklearn-iris` service here? It seems like it will essentially be dropped, in which case why even include it?

docs/samples/graph/switch.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - service: green
+        serviceUrl: http://green.default.10.166.15.29.sslip.io/splitter
+      nextRoutes:
+      - nodeName: mymodel
+        data: $response
+    mymodel:
+      routerType: Switch
+      routes:
+      - service: blue
+        serviceUrl: http://blue.default.10.166.15.29.sslip.io/splitter
```
njhill: The CRD doc states `InferenceService URL, mutually exclusive with Service`, so this example appears to be invalid?

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+```
+[***Demo yaml***](switch.yaml)
+### **3.3 [Splitter](splitter.yaml)**
+***Test steps***
+
+1. Deploy the demo `isvc` and `graph` 
+```shell 
+kubectl apply -f splitter.yaml
+```
+2. Waiting for `isvc` and `graph` up.
```
njhill:
```suggestion
2. Waiting for `isvc` and `graph` to come up.
```

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -0,0 +1,224 @@
...
+			// Object not found, return.  Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	r.Log.Info("Reconciling inference graph", "apiVersion", graph.APIVersion, "graph", graph.Name)
+	configMap := &v1.ConfigMap{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: constants.InferenceServiceConfigMapName, Namespace: constants.KServeNamespace}, configMap)
```
njhill: Can use the passed-in `ctx` for this (and others below).

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,252 @@
...
+// metadata:
+//   name: dog-breed-classification
+// spec:
+//   nodes:
+//     top-level-classifier:
+//       routes:
+//       - service: cat-dog-classifier
+//       nextRoutes:
+//         nodeName: breed-classifier
+//         data: "{ $input }"
```
njhill: Syntax for this looks inconsistent with other code/docs. (`{ $input }` vs `$request`)

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,252 @@
...
+// InferenceRouter defines the router for each InferenceGraph node with one or multiple models
+// and where it routes to as next step
+//
+// ```yaml
+// kind: InferenceGraph
+// metadata:
+//   name: canary-route
+// spec:
+//   nodes:
+//     mymodel:
```
njhill: There's no node named `root` in these examples

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func pickupRoute(routes []v1alpha1.InferenceRoute) *v1alpha1.InferenceRoute {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	//generate num [0,100)
+	point := r.Intn(99)
+	start, end := 0, 0
+
+	for _, route := range routes {
+		start += end
```
njhill: I think this line is meant to be `start = end`?

Also there is no need to keep track of `start` at all, can just remove that var and related check.

njhill (commented):
@yuzisun thanks for this and sorry for taking so long to look at it closely, I've taken a (non-exhaustive) first pass and added some comments.

The design doc appears to be a bit out of date, there's inconsistencies between the design doc, examples/doc in the PR, and actual implemented code. The resulting API seems a bit muddled to me, and I think has some limitations in terms of the kind of compositions possible.

I have some thoughts about how the API design could be streamlined a bit, I've created a discussion for those [here](https://github.com/kserve/kserve/discussions/2170) to avoid derailing this PR.

shrinandj:
I'd love to try this out. Any idea when this could get merged?

config/webhook/manifests.yaml
```diff
@@ -108,3 +108,30 @@ webhooks:
...
+    rules:
+      - apiGroups:
+          - serving.kserve.io
+        apiVersions:
+          - v1alpha1
+        operations:
+          - CREATE
+          - UPDATE
+        resources:
+          - inferencegraphs
```
alexagriffith: nit: ied complains no line at eof

pkg/apis/serving/v1alpha1/inference_graph_validation.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if err := validateInferenceGraphSplitterWeight(ig); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (ig *InferenceGraph) ValidateUpdate(old runtime.Object) error {
+	validatorLogger.Info("validate update", "name:", ig.Name)
```
alexagriffith: nit: can we remove `:` in `name:` to be consistent with the other logging

pkg/controller/v1alpha1/inferencegraph/controller.go
alexagriffith: also question: why not use something like 
```
ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
defer cancel()
```

for the client call?

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+
+func pickupRouteByCondition(input []byte, routes []v1alpha1.InferenceRoute) *v1alpha1.InferenceRoute {
+	//convert input to Input
+	data, err := convertInput(input)
+	if err != nil {
+		log.Error(err, "converInput failed.")
+		return nil
+	}
+	for _, route := range routes {
+		//new jsonpath object
```
alexagriffith: obvious comment - could delete it

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+		j := jsonpath.New("Parser")
+		//j.AllowMissingKeys(true)
```
alexagriffith: did you mean to leave this here?

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+	var data Input
+	data.Items = append(data.Items, inputData)
+	return data, nil
+}
+
+func convertCondition(origin string) string {
+	//remove whitespaces
+	str := strings.Replace(origin, " ", "", -1)
+	//remove {}
+	str = str[1 : len(str)-1]
```
alexagriffith: do we want to validate that the leading and trailing brackets exist?

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		for name, result := range ensembleRes {
+			responseBytes := <-result
+
+			var res map[string]interface{}
+			json.Unmarshal(responseBytes, &res)
+			response[name] = res
+		}
+	} else { //routeType == Single
```
alexagriffith: just to be explicit, could we change else to `else if currentStep.RouterType == v1alpha1.Single` ? so that when the code changes, it is clear and if something is changed, it will break if it doesn't match one of the conditions.

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+func timeTrack(start time.Time, name string) {
+	elapsed := time.Since(start)
+	log.Info("elapsed time", "node", name, "time", elapsed)
+}
+
+func routeStep(nodeName string, currentStep v1alpha1.InferenceRouter, graph v1alpha1.InferenceGraphSpec, input []byte, res chan<- []byte) error {
+	log.Info("current step", "nodeName", nodeName, "URL", currentStep.Routes[0].ServiceUrl)
+	defer timeTrack(time.Now(), nodeName)
+	response := map[string]interface{}{}
+	//For splitter and ABNTest call virtual service
```
alexagriffith: nit: can you add a bit of "why" to this comment? like - why is splitter different here?

config/webhook/manifests.yaml
alexagriffith: this happens in a couple of files

docs/samples/graph/README.md
```diff
@@ -0,0 +1,251 @@
...
+    - [**2.5 Splitter Node**](#25-splitter-node)
+  - [**3 Examples**](#3-examples)
+    - [**3.1 Single**](#31-single)
+    - [**3.2 Switch**](#32-switch)
+    - [**3.3 Splitter**](#33-splitter)
+    - [**3.4 Ensemble**](#34-ensemble)
+# **Inference Graph**
+## **1. Problem** 
+
+Often the time on a production inference pipeline, it is not a single inference service, models need to be chained together to produce the final prediction result. For example a face recognition pipeline may need to find the face area first and then compute the features of the faces to match the face database, these two models are depending on each other and the first model’s output is the second model’s input. Another example is the NLP pipeline, it is very common that you would need to do some document classification first followed by downstream named entity detection or text summary tasks.[[...]](https://docs.google.com/document/d/13VHfOxa72pgoy5Eg5c-gGHZfEL7uBDc7rTLa2pAi4Ko/edit#heading=h.x9snb54sjlu9)
```
alexagriffith:
```suggestion
Production inference pipelines are often composed of multiple inference services, and models need to be chained together to produce the final prediction result. For example, a face recognition pipeline may need to find the face area first and then compute the features of the faces to match the face database. These two models depend on each other, and the first model’s output is the second model’s input. Another example is the NLP pipeline. It is very common to do some document classification and then have downstream tasks for named entity detection or text summary.[[...]](https://docs.google.com/document/d/13VHfOxa72pgoy5Eg5c-gGHZfEL7uBDc7rTLa2pAi4Ko/edit#heading=h.x9snb54sjlu9)
```

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+		log.Error(err, "error while reading the response")
+		return err
+	}
+	res <- body
+	return nil
+}
+
+func pickupRoute(routes []v1alpha1.InferenceRoute) *v1alpha1.InferenceRoute {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	//generate num [0,100)
```
alexagriffith, line 57: nit: can you add "why" for this comment too? why do we need to gen random num? I think that will add value to the code/comment

cmd/router/main.go
```diff
@@ -0,0 +1,248 @@
...
+	Items []interface{} `json:"items"`
+}
+
+func convertInput(input []byte) (interface{}, error) {
+	var inputData interface{}
+	if err := json.Unmarshal(input, &inputData); err != nil {
+		return nil, err
+	}
+	var data Input
+	data.Items = append(data.Items, inputData)
```
alexagriffith: if we set data here to the zero val of Input, why do we need to append instead of say 

```
data.Items := inputData
```
?

cmd/router/main.go
yuzisun: we are going to use gjson https://github.com/tidwall/gjson which is used by argo for condition check.

cmd/router/main.go
yuzisun: right, this is for sub-pipeline ensemble case, e.g parallel branches for different model pipelines. I liked your idea in the [discussion](https://github.com/kserve/kserve/discussions/2170), it is more general and you can also apply the condition and weights for nodes.
```
     | - transformer1 -> model1 -|
-->                              | ->
     | - transformer2 -> model2 -|
```

cmd/router/main.go
yuzisun: there could be case that you want to add the condition based on the response but still send the input to next model with the previous request, for example you first classify the text with high level category and then send the original text as input to next model to classify sub categories based on the previous model prediction result.

docs/samples/graph/README.md
yuzisun: though service name is optional and in case only service url is provided we won't be able to populate that, but agreed we can at least optimize for the service name case.

docs/samples/graph/sequence.yaml
yuzisun: this example is over-simplified, often the time you would need to add condition based on the response and then forward the request to the next model.

docs/samples/graph/switch.yaml
yuzisun: right, should remove the serviceUrl

yuzisun:
> Hey, @yuzisun. This looks great so far. Did an initial review with some comments inline.
> 
> Also noticed that there are several line changes to the servingruntime and isvc crd yamls. I'm assuming because of differing versions of yq shifting the indentation when `make manifests` is run, creating quite huge diffs. In the storage spec PR a while back, an older version of yq was used in the generation, but now, it looks like a newer version of 'yq' is used, re-shifting the indentation.
> 
> Should we have the crd regeneration done in a separate PR to keep this PR cleaner? And should we do something like we do with ModelMesh-Serving where we [perform](https://github.com/kserve/modelmesh-serving/blob/main/scripts/develop.sh) tests/generation using a [developer image](https://github.com/kserve/modelmesh-serving/blob/main/Dockerfile.develop) with correct dependencies installed?

Sounds good, we can also do something like https://github.com/kserve/kserve/blob/master/hack/verify-codegen.sh to verify if the PR generates the files correctly by comparing the generation in CI.

docs/samples/graph/README.md
yuzisun: I have added a new field `stepName` for this, so the merged response will be populated with `stepName` as the key instead of using `serviceUrl`.

rachitchauhan43: `stepName` field would work fine too.

cmd/router/main.go
rachitchauhan43: As per our discussion in community meting, transformers should be used for this prupose.

rachitchauhan43:
Did observe this bug while testing:

Created this:
```
kind: InferenceService
metadata:
  name: custom-model
spec:
  predictor:
    containers:
      - name: kserve-container
        image: rachitchauhan885/custom-model:v4
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
---
kind: InferenceService
metadata:
  name: hist-feature-fetcher
spec:
  predictor:
    containers:
      - name: kserve-container
        image: rachitchauhan885/hist-feature-fetcher:v1
        args:
          - --model_name
          - hist-feature-fetcher
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
---
kind: InferenceService
metadata:
  name: latest-feature-fetcher
spec:
  predictor:
    containers:
      - name: kserve-container
        image: rachitchauhan885/latest-feature-fetcher:v1
        args:
          - --model_name
          - latest-feature-fetcher
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
---
kind: InferenceService
metadata:
  name: qbo-in
  namespace: rachit
spec:
  predictor:
    containers:
      - name: kserve-container
        image: rachitchauhan885/qbo-in:v1
        args:
          - --model_name
          - qbo-in
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
---
kind: InferenceService
metadata:
  name: qbo-us
  namespace: rachit
spec:
  predictor:
    containers:
      - name: kserve-container
        image: rachitchauhan885/qbo-us:v1
        args:
          - --model_name
          - qbo-us
        resources:
          requests:
            cpu: 100m
            memory: 256Mi
---
kind: InferenceGraph
metadata:
  name: all-in-one
spec:
  nodes:
    root:
      routerType: Switch
      routes:
      - service: qbo-in
        condition: "{.input.country_code == \"IN\"}"
      - service: qbo-us
        condition: "{.input.country_code == \"US\"}"
      nextRoutes:
      - nodeName: feature-fetcher
        data: $response
    feature-fetcher:
      routerType: Ensemble
      routes:
      - service: hist-feature-fetcher
      - service: latest-feature-fetcher
      nextRoutes:
      - nodeName: txn-categorizer
        data: $response
    txn-categorizer:
      routerType: Single
      routes:
      - service: custom-model
```

The final output on calling all-in-one-graph is:
```
{
  "feature-fetcher": {
    "txn-categorizer": {
      "predictions": {
        "http://hist-feature-fetcher.rachit.svc.cluster.local/v1/models/hist-feature-fetcher:predict": {
          "historical_features": {
            "last_tax_year_status": "Single",
            "predictions": "From IN model"
          }
        },
        "http://latest-feature-fetcher.rachit.svc.cluster.local/v1/models/latest-feature-fetcher:predict": {
          "latest_features": {
            "predictions": "From IN model",
            "this_tax_year_status": "Married-Joint"
          }
        }
      }
    }
  }
}
```

But it should be this:
```{
  "txn-categorizer": {
    "predictions": {
      "feature-fetcher": {
        "http://hist-feature-fetcher.rachit.svc.cluster.local/v1/models/hist-feature-fetcher:predict": {
          "historical_features": {
            "last_tax_year_status": "Single",
            "predictions": "From IN model"
          }
        },
        "http://latest-feature-fetcher.rachit.svc.cluster.local/v1/models/latest-feature-fetcher:predict": {
          "latest_features": {
            "predictions": "From IN model",
            "this_tax_year_status": "Married-Joint"
          }
        }
      }
    }
  }
}
```
The outer most keys are not correctly placed. Can you verify ?

njhill:
@yuzisun thanks for the latest changes, I really like the node/service unification and sequence addition. While reviewing them I did a bit of streamlining of the graph execution logic and added a bit more validation, see https://github.com/yuzisun/kserve/pull/8. I do have some further thoughts/comments, will add them soon.

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,253 @@
...
+	//
+	// - `Ensemble:` routes the request to multiple models and then merge the responses
+	//
+	// - `Switch:` routes the request to two or more services with specified routing rule
+	//
+	RouterType InferenceRouterType `json:"routerType"`
+
+	// Routes defines destinations for the current router node
+	// +optional
+	Routes []InferenceStep `json:"routes,omitempty"`
```
njhill: Should we rename this now?
```suggestion
	Steps []InferenceStep `json:"steps,omitempty"`
```

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,253 @@
...
+
+// +k8s:openapi-gen=true
+
+type InferenceTarget struct {
+	// The node name for routing as next step
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
+
+	// named reference for InferenceService
+	Service string `json:"service,omitempty"`
```
njhill: WDYT about renaming this to `ServiceName` for clarity/consistency?

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,253 @@
...
+
+	// InferenceService URL, mutually exclusive with Service
+	// +optional
+	ServiceUrl string `json:"serviceUrl,omitempty"`
```
njhill: per go style convention
```suggestion
	ServiceURL string `json:"serviceUrl,omitempty"`
```

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,253 @@
...
+	RouterType InferenceRouterType `json:"routerType"`
+
+	// Routes defines destinations for the current router node
+	// +optional
+	Routes []InferenceStep `json:"routes,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+
+type InferenceTarget struct {
```
njhill, line 194:
```suggestion
// Exactly one InferenceTarget field must be specified
type InferenceTarget struct {
```

njhill (commented):
A few renaming suggestions

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -0,0 +1,257 @@
...
+// +k8s:openapi-gen=true
+// Exactly one InferenceTarget field must be specified
+
+type InferenceTarget struct {
+	// The node name for routing as next step
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
+
+	// named reference for InferenceService
+	ServiceName string `json:"serviceName,omitempty"`
```
pvaneck, line 200: I noticed that in the example InferenceGraph YAML files and doc snippets,  `service` is used instead of `serviceName` as declared here and in the generated CRD. Should YAML files and docs be updated with `serviceName`?

pkg/apis/serving/v1alpha1/inference_graph_validation.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+			if target.ServiceURL != "" {
+				count += 1
+			}
+			if count == 0 {
+				return fmt.Errorf("Step %d (\"%s\") in node \"%s\" of InferenceGraph \"%s\" does not specify an inference target",
+					i, route.StepName, nodeName, ig.Name)
+			}
+			if count != 1 {
+				return fmt.Errorf("Step %d (\"%s\") in node \"%s\" of InferenceGraph \"%s\" specifies more than one of nodeName, service, serviceUrl",
```
pvaneck: s/service/serviceName/ ?

pvaneck (commented):
Can you rebase on master to get the new github action e2e workflow?

pkg/apis/serving/v1alpha1/inference_graph.go
yuzisun: updated

njhill (approved):
Thanks @yuzisun !

pvaneck:
Thanks! Good to finally have this.
/lgtm

cmd/router/main.go
```diff
@@ -37,32 +38,30 @@ import (
...
 	if err != nil {
 		log.Error(err, "error while reading the response")
-		return err
 	}
-	res <- body
-	return nil
+	return body, err
 }
 
 func pickupRoute(routes []v1alpha1.InferenceStep) *v1alpha1.InferenceStep {
```
rachitchauhan43, line 55: @yuzisun : Can you explain how this function is working ? How is it taking care of weighted distribution ?

PR #1916, opened 2021-11-16
Fix path to kserve.yaml in quick_install.sh script

iamahern:
The current quick install script is broken. It will install Istio, KNative, and Cert Manager, but the KServe install itself will fix.

The script is attempting to install KServe from a local filesystem location (`install/${KSERVE_VERSION}/${KSERVE_CONFIG}`). This path does not exist. It appears to want to refer to the release GH release file (`https://github.com/kserve/kserve/releases/download/${KSERVE_VERSION}/${KSERVE_CONFIG}`).

I could not find an existing issue.

To validate/reproduce:

1. Create a new  / empty Kind / Minikube install (on MacOS 4GB / 4CPU seems to do the trick).
2. `cd` into an empty directory.
3. (To reproduce) Run the quick install instructions: https://kserve.github.io/website/get_started/#install-the-kserve-quickstart-environment
4. (To validate) Run the patched script in the PR.

ALSO - the existing quick start needs to be patched to get new users up and running once the fix is confirmed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/quick_install.sh
+++ b/hack/quick_install.sh
@@ -103,6 +103,6 @@ KSERVE_CONFIG=kfserving.yaml
 if [ ${KSERVE_VERSION:3:1} -gt 6 ]; then KSERVE_CONFIG=kserve.yaml; fi
 
 # Retry inorder to handle that it may take a minute or so for the TLS assets required for the webhook to function to be provisioned
-for i in 1 2 3 4 5 ; do kubectl apply -f install/${KSERVE_VERSION}/${KSERVE_CONFIG} && break || sleep 15; done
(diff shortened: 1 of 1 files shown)


yuzisun:
@iamahern Did you install from the KServe root directory?

iamahern:
@yuzisun - No. I followed the instructions on the KServe website here:
https://kserve.github.io/website/get_started/

Based on the instructions, I should not need to download/run from the root of the code repo.

yuzisun:
> @yuzisun - No. I followed the instructions on the KServe website here: https://kserve.github.io/website/get_started/
> 
> Based on the instructions, I should not need to download/run from the root of the code repo.

Ah make sense! Thanks for the fix!

PR #1918, opened 2021-11-17
Return only JSON responses from KFServer

markwinter:

Currently KFServer returns a mix of HTML and JSON responses.

I have made it so only JSON responses are returned.

This should be nicer to program against.

**Release note**:

```release-note
KFServer returns only JSON responses
```


markwinter:
Fixing the tests

yuzisun:
Thanks @markwinter ! Nice fix!

/lgtm
/approve

PR #1925, opened 2021-11-25
Change to fix sdk replace does not wait for new isvc to be ready

### LINKED ISSUE #1739: Unexpected result for SDK replace method with watch=True
/kind bug

Used Python SDK to:
* Execute SDK `create` method to create isvc and wait until it's available
* Execute SDK `replace` method (with `watch=True`) to create new revision

In SDK `replace` method with `watch=True` I'd expect to wait for the new revision to become available. Instead watch immediately returns with `READY` equal to `True` because isvc was already available from the previous revision.

The wording in SDK documentation for any non-`create` method that accepts `watch=True` is a bit misleading. What would be the case for using `watch=True` in `replace`, `patch`, `promote` etc.?

I'd like to understand how in general one can (with SDK) `create` isvc and then `replace` it while checking if and when new revision becomes available. 

In other words, how can one use SDK to create pinned or canary rollout and observe new revision's progress.

andyi2it:
Made changes to wait for the new replaced version to be ready

Made changes to  wait for isvc to be ready when watch true is set.
Implemented using generation of new isvc and comparing the `latestreadyrevision`

Fixes #1739 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/api/watch.py
+++ b/python/kserve/kserve/api/watch.py
@@ -21,9 +21,10 @@ from ..constants import constants
 from ..utils import utils
 
 
-def isvc_watch(name=None, namespace=None, timeout_seconds=600):
+def isvc_watch(name=None, namespace=None, timeout_seconds=600, generation=0):
     """Watch the created or patched InferenceService in the specified namespace"""
 
+    print("isvc watch called")
     if namespace is None:
         namespace = utils.get_default_target_namespace()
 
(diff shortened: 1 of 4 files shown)


python/kserve/kserve/api/watch.py
```diff
@@ -21,7 +21,7 @@
 from ..utils import utils
 
 
-def isvc_watch(name=None, namespace=None, timeout_seconds=600):
+def isvc_watch(name=None, namespace=None, timeout_seconds=600, generation=0):
```
yuzisun, line 24: The generation field can be fetched from isvc metadata, so I think we do not need to pass `generation` here.

python/kserve/kserve/api/watch.py
```diff
@@ -46,24 +46,32 @@ def isvc_watch(name=None, namespace=None, timeout_seconds=600):
...
-                        status = condition.get('status', 'Unknown')
-                tbl(isvc_name, status, 100-traffic_percent, traffic_percent, url)
+                if constants.OBSERVED_GENERATION in isvc['status']:
+                    observed_generation = isvc['status'][constants.OBSERVED_GENERATION]
+                    for t in traffic:
+                        if t["latestRevision"]:
+                            traffic_percent = t["percent"]
+
+                    if generation != 0 and observed_generation != generation:
+                        time.sleep(2)
```
yuzisun: When `observed_generation` is different from `generation` I think we can remove the sleep,  mark the status `Unknown` and then continue.

PR #1926, opened 2021-11-26
Follow up work on ServingRuntime support - PR1 #1924

### LINKED ISSUE #1924: Follow up work for ServingRuntime support
/kind feature

- Add template variable support for enabling passing in isvc name to runtime container.
- Add ability to disable a Framework/ModelType, in a ServingRuntime and add associated logic?
- Add E2E tests using new format
- Eventually add default ClusterServingRuntimes as a part of the default installation (perhaps generate separate yaml file for 
default runtimes as including the runtimes in the main default install causes that race condition with the CRD not existing yet).

Suresh-Nakkeran:
fixes #1924 

Below are features added as part of this PR,

- Template variable support for enabling passing in isvc name to runtime container.
- More default ClusterServingRuntimes
- Logic in InferenceService mutator to convert old schema into the new schema.


config/runtimes/kserve-sklearnserver.yaml
```diff
@@ -10,13 +10,6 @@ spec:
     - name: kserve-container
       image: kserve-sklearnserver:replace
       args:
-        - --model_name=model
+        - --model_name={{.Name}}
         - --model_dir=/mnt/models
         - --http_port=8080
```
yuzisun, line 15: Can we keep the resource defaults? I think we can remove the current hard coded resource defaults logic in mutating webhook now.

config/runtimes/kserve-lgbserver.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-lgbserver
+spec:
+  supportedModelTypes:
+    - name: lightgbm
+      version: "0"
+  containers:
+    - name: kserve-container
+      image: kserve-lgbserver:replace
```
yuzisun, line 11: the version here should be a template variable also and replaced by the `runtimeVersion` field from the model spec.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -86,4 +89,235 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
 			}
 		}
 	}
+	isvc.setPredictorModelDefaults(config)
+}
+
+func (isvc *InferenceService) setPredictorModelDefaults(config *InferenceServicesConfig) {
```
yuzisun: looks like the `config` is not really used here.

config/runtimes/kserve-mlserver.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-mlserver
+spec:
+  supportedModelTypes:
+    - name: sklearn
+      version: "0"
+    - name: xgboost
+      version: "0"
```
yuzisun: xgboost should be version 1 according to https://kserve.github.io/website/modelserving/v1beta1/serving_runtime/

config/runtimes/kserve-torchserveserver-gpu.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-torchserveserver-gpu
+spec:
+  supportedModelTypes:
+    - name: pytorch
+      version: "0"
```
yuzisun: pytorch major version should be 1

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  supportedModelTypes:
+    - name: tensorrt
+      version: "0"
+    - name: tensorflow
+      version: "1"
+    - name: onnx
+      version: "0"
+    - name: pytorch
+      version: "0"
```
yuzisun: pytorch major version is 1

pkg/constants/constants.go
```diff
@@ -275,6 +275,46 @@ const (
...
 )
 
+// built-in runtime servers
+const (
+	SKLearnServer    = "kserve-sklearnserver"
+	MLServer         = "kserve-mlserver"
+	TFServing        = "kserve-tensorflow-serving"
+	TFServingGPU     = "kserve-tensorflow-serving-gpu"
+	XGBServer        = "kserve-xgbserver"
+	PYTorchServer    = "kserve-pytorchserver"
```
yuzisun: nit: `PyTorchServer`

config/runtimes/kserve-torchserveserver-gpu.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-torchserveserver-gpu
```
yuzisun: I think the gpu runtime can be consolidated, it is the image tag which is different.

for cpu image it is `0.5.0` while gpu image tag is `0.5.0-gpu`

config/runtimes/kserve-torchserveserver.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-torchserveserver
```
yuzisun: nit: we can just name it `kserve-torchserve`

config/runtimes/kserve-tensorflow-serving-gpu.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-tensorflow-serving-gpu
```
yuzisun: the gpu runtime can be consolidated, the image tag is different.

for cpu it is `tensorflow-serving:1.14` while gpu tag is `tensorflow-serving:1.14-gpu`

config/runtimes/kserve-tensorflow-serving-gpu.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  supportedModelTypes:
+    - name: tensorflow
+      version: "1"
```
yuzisun: We should also add tensorflow version "2" to the supported list.

config/runtimes/kserve-mlserver.yaml
Suresh-Nakkeran: Thanks, @yuzisun . I will update version by referring the link.

config/runtimes/kserve-sklearnserver.yaml
Suresh-Nakkeran: okay, @yuzisun. But we will keep the current hardcoded resource defaults logic in mutating webhook for now since it may be used in multi-model deployment.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -86,4 +89,224 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
...
+		PredictorExtensionSpec: isvc.Spec.Predictor.Tensorflow.PredictorExtensionSpec,
+		Runtime:                &runtime,
+	}
+	// remove tensorflow spec
+	isvc.Spec.Predictor.Tensorflow = nil
+}
+
+func (isvc *InferenceService) assignXGBoostRuntime() {
+	// skips if the storage uri is not specified
+	if isvc.Spec.Predictor.SKLearn.StorageURI == nil {
```
pvaneck: SKLearn -> XGBoost

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -86,4 +89,224 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
...
+		PredictorExtensionSpec: isvc.Spec.Predictor.LightGBM.PredictorExtensionSpec,
+		Runtime:                &runtime,
+	}
+	// remove lightgbm spec
+	isvc.Spec.Predictor.LightGBM = nil
+}
+
+func (isvc *InferenceService) assignPaddleRuntime() {
+	// skips if the storage uri is not specified
+	if isvc.Spec.Predictor.SKLearn.StorageURI == nil {
```
pvaneck: SKLearn -> Paddle

config/runtimes/kserve-lgbserver.yaml
pvaneck: I think this is probably fine since this is replaced by the image tags in kustomization.yaml to provide the default image tag. Would need additional logic elsewhere to replace the tag  if runtimeVersion is provided. Although a user could also just pass in the `image` value itself as an override.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -86,4 +89,224 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
 			}
 		}
 	}
+	isvc.setPredictorModelDefaults()
+}
+
+func (isvc *InferenceService) setPredictorModelDefaults() {
+	if isvc.Spec.Predictor.Model != nil {
```
pvaneck: We need to consider when a PATCH or update to an existing inference service happens. For example, if a user deployed an Isvc using the  `predictor.tensorflow` schema, the mutator will convert this to `predictor.model`. Then when a user updates the isvc also using the `predictor.tensorflow` schema (for example, when doing a canary rollout), at this point in the code, the isvc spec will have both `Predictor.Tensorflow`and `Predictor.Model` as non-nil. 

This ties in to the failing canary rollout e2e test where it is failing the `validateExactlyOneImplementation` call above with `Exactly one of [SKLearn, XGBoost, Tensorflow, PyTorch, Triton, ONNX, PMML, LightGBM, Paddle, Model, PodSpec] must be specified in PredictorSpec`

pvaneck (commented):
Hey, thanks for this! Did a quick passthrough and it seems the main thing is to get the e2e tests passing. Found some issues which may help with that.

pkg/apis/serving/v1beta1/inference_service_defaults.go
Suresh-Nakkeran: Thanks, @pvaneck . It's a copy paste error during clean up.

pkg/apis/serving/v1beta1/inference_service_defaults.go
Suresh-Nakkeran: @yuzisun your thoughts on how we can handle update/patch request?

config/runtimes/kustomization.yaml
```diff
@@ -1,14 +1,68 @@
...
+    newName: kserve/pytorchserver
+    newTag: latest
+
+  - name: kserve-pytorchserver-gpu
+    newName: kserve/pytorchserver
+    newTag: latest-gpu
+
+  - name: kserve-torchserve
+    newName: pytorch/torchserve-kfs
+    newTag: 0.5.0
```
pvaneck: Shouldn't the torchserve tags be 0.4.1 and 0.4.1-gpu?

pkg/apis/serving/v1beta1/inference_service_defaults.go
pvaneck: I think removing  the `isvc.Spec.Predictor.Model != nil` check here, and moving the call to `isvc.setPredictorModelDefaults()` in `DefaultInferenceService` to the beginning of that function should work.

Suresh-Nakkeran:
Yes @pvaneck . I am also thinking the same.

config/runtimes/kserve-tensorflow-serving-gpu.yaml
pvaneck: @yuzisun If we are using the `1.14` tensorflow-serving tags, doesn't that mean TF 2.x wouldn't be supported?

config/runtimes/kserve-tensorflow-serving-gpu.yaml
pvaneck: Right, seems like the image tag is the only difference. If we remove the `*-gpu.yaml` servingruntimes, do we then rely on the user to pass in the `*-gpu` tag through the runtimeVersion field?

config/runtimes/kustomization.yaml
```diff
@@ -1,14 +1,68 @@
...
+    newName: tensorflow/serving
+    newTag: 1.14.0-gpu
+
   - name: kserve-sklearnserver
     newName: kserve/sklearnserver
     newTag: latest
 
+  - name: mlserver
+    newName: docker.io/seldonio/mlserver
+    newTag: 0.2.1
```
pvaneck: Looks like MLServer was updated to `0.5.3` via https://github.com/kserve/kserve/commit/f90c669bdc0324aed5b3d527c53c7562858c2dcc . Can you rebase and update the version here?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -163,3 +168,30 @@ func GetServingRuntime(cl client.Client, name string, namespace string) (*v1alph
 	}
 	return nil, goerrors.New("No ServingRuntimes or ClusterServingRuntimes with the name: " + name)
 }
+
+// Replace placeholders in runtime container by values from inferenceservice metadata
+func ReplacePlaceholders(container *v1.Container, meta metav1.ObjectMeta) error {
```
pvaneck, line 173: Mind adding some unit tests for these util functions in utils_test.go?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -128,6 +128,15 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
 		container.Name = constants.InferenceServiceContainerName
 
+		// Replace placeholders in runtime container by values from inferenceservice metadata
+		err = isvcutils.ReplacePlaceholders(container, isvc.ObjectMeta)
+		if err != nil {
+			return errors.Wrapf(err, "failed to replace placeholders in serving runtime Container")
+		}
+
+		// Update image tag if GPU is enabled or runtime version is provided
+		isvcutils.UpdateImageTag(container, *isvc.Spec.Predictor.Model.RuntimeVersion)
```
pvaneck: This line is causing a `nil pointer dereference"` error. A check to make sure RuntimeVersion is not nil should fix it.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -163,3 +168,30 @@ func GetServingRuntime(cl client.Client, name string, namespace string) (*v1alph
...
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(buf.Bytes(), container)
+}
+
+// Update image tag if GPU is enabled or runtime version is provided
+func UpdateImageTag(container *v1.Container, runtimeVersion string) {
+	image := container.Image
+	if utils.IsGPUEnabled(container.Resources) &&
```
pvaneck: Hmm, I'm wondering if we should have this gpu block. It might have the unintended side effect of appending `-gpu` to image tags that might not need it since, in this case, any inferenceservice with a `resources.limits` object containing `nvidia.com/gpu` will automatically have the `-gpu` appended.

I'm thinking maybe just keep it simple here, and just rely on the user passing in the correct runtimeVersion or using a runtime that already specifies a gpu-enabled image. WDYT?

pkg/constants/constants.go
```diff
@@ -275,6 +275,46 @@ const (
 	ModelMeshDeployment DeploymentModeType = "ModelMesh"
 )
 
+// built-in runtime servers
+const (
+	SKLearnServer    = "kserve-sklearnserver"
+	MLServer         = "kserve-mlserver"
+	TFServing        = "kserve-tensorflow-serving"
+	TFServingGPU     = "kserve-tensorflow-serving-gpu"
```
pvaneck: The gpu consts should be removed now right?

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -87,3 +91,228 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
...
+func (isvc *InferenceService) assignTensorflowRuntime() {
+	// skips if the storage uri is not specified
+	if isvc.Spec.Predictor.Tensorflow.StorageURI == nil {
+		isvc.Spec.Predictor.Model = nil
+		return
+	}
+	// assign built-in runtime based on gpu config
+	var runtime = constants.TFServing
+	if utils.IsGPUEnabled(isvc.Spec.Predictor.Tensorflow.Resources) {
+		runtime = constants.TFServingGPU
```
pvaneck: Since the GPU runtimes were removed, looks like we don't need these checks anymore.

config/runtimes/kserve-mlserver.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-mlserver
+spec:
+  supportedModelTypes:
+    - name: sklearn
+      version: "2"
```
yuzisun: sklearn version should be "0"
xgboost version should be "1"
@adriangonz what are the other frameworks MLServer support?

config/runtimes/kserve-torchserve.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-torchserve
+spec:
+  supportedModelTypes:
+    - name: pytorch
+      version: "2"
```
yuzisun: pytorch version should be "1"

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  supportedModelTypes:
+    - name: tensorrt
+      version: "2"
+    - name: tensorflow
+      version: "2"
+    - name: onnx
+      version: "2"
+    - name: pytorch
+      version: "2"
```
yuzisun:
```suggestion
      version: "1"
```

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: kserve-tritonserver
+spec:
+  supportedModelTypes:
+    - name: tensorrt
+      version: "2"
+    - name: tensorflow
+      version: "2"
+    - name: onnx
+      version: "2"
```
yuzisun:
```suggestion
      version: "1"
```

pkg/controller/v1beta1/inferenceservice/utils/utils.go
yuzisun: Agreed, for example triton does not differentiate cpu or gpu image while tfserving, torchserve does. So I think the logic should be following:
1) if `runtimeVersion` is specified, user should pass in the right gpu image version like `tfserving:1.14-gpu`
2) if `runtimeVersion` is not specified, we can add an optional field for gpuImage on `ServingRuntime`?

config/runtimes/kserve-mlserver.yaml
adriangonz: We could add MLflow and LightGBM to that list. Although I'm assuming that'd require changes in the code as well?

Suresh-Nakkeran:
@pvaneck @yuzisun I have few questions.

1. In `pkg/controller/v1beta1/inferenceservice/components/predictor.go:line 105`, For the case no explicit runtime provided, we get the list of supported serving runtime specs which are not in same order every time since we are not sorting it, which causing issue when we deploy ISVC with new schema directly just framework name, version.
Do we need to address this issue?

2. Are we not allowed to filter servingruntime based on protocol version (V1 , V2)?

config/runtimes/kserve-mlserver.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-mlserver
+spec:
+  supportedModelTypes:
+    - name: sklearn
+      version: "0"
+    - name: xgboost
+      version: "1"
```
pvaneck, line 10: Lets add lightgbm with version "3" and mlflow with version "1" to the list

config/runtimes/kserve-lgbserver.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-lgbserver
+spec:
+  supportedModelTypes:
+    - name: lightgbm
+      version: "1"
```
pvaneck: Based on https://github.com/kserve/kserve/blob/master/python/lgbserver/setup.py#L36, the version should be "2"

config/runtimes/kserve-paddleserver.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-paddleserver
+spec:
+  supportedModelTypes:
+    - name: paddle
+      version: "1"
```
pvaneck: Based on https://github.com/kserve/kserve/blob/master/python/paddleserver/setup.py#L28, I believe version here should be "2"

config/runtimes/kserve-pmmlserver.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-pmmlserver
+spec:
+  supportedModelTypes:
+    - name: pmml
+      version: "1"
```
pvaneck: I think based on the version of jpmml-evaluator (https://github.com/kserve/kserve/blob/master/python/pmmlserver/setup.py#L35) , we should have two items in this supportedModelTypes list. One for pmml version "3" and one for pmml version "4".

config/runtimes/kserve-xgbserver.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-xgbserver
+spec:
+  supportedModelTypes:
+    - name: xgboost
+      version: "1"
```
pvaneck: Based on https://github.com/kserve/kserve/blob/master/python/xgbserver/setup.py#L36, version here should be "0".

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-tritonserver
+spec:
+  supportedModelTypes:
+    - name: tensorrt
+      version: "2"
```
pvaneck: I htink based on the tag `21.09-py3` we have for triton, tensorrt should be "8". And tensorflow version "1" is also supported.

pvaneck:
> 1. In `pkg/controller/v1beta1/inferenceservice/components/predictor.go:line 105`, For the case no explicit runtime provided, we get the list of supported serving runtime specs which are not in same order every time since we are not sorting it, which causing issue when we deploy ISVC with new schema directly just framework name, version.
>    Do we need to address this issue?
> 2. Are we not allowed to filter servingruntime based on protocol version (V1 , V2)?

1. Hmm, didn't realize the ordering was non-deterministic. I'm assuming you are getting a different serving runtime each time you deploy a specific model? If we want consistency in what SR is selected, then sorting would do the trick. Although, maybe a random selection could be considered a feature not a bug? @njhill How do you think this case should be handled?

2. I don't believe we have any mechanism for filtering v1 vs v2. The ServingRuntimes are agnostic to that detail. And I believe v1 will eventually be phased out.

pvaneck (commented):
Just some version comments for the ClusterServingRuntimes.

config/runtimes/kserve-tensorflow-serving-gpu.yaml
yuzisun: Version is not specified on the tfserving runtime right? The runtime template can work with both 1.x and 2.x unless 2.x has deprecated some of the runtime arguments used on the template. I think we probably only want to create new runtime if different versions require different configurations.

config/runtimes/kserve-mlserver.yaml
yuzisun: @adriangonz the purpose of introducing servingruntime is that we no longer need code change for adding new frameworks.. we only need to create new serving runtime CR and install along with kserve

Suresh-Nakkeran:
@pvaneck @yuzisun I updated the model supported versions. Can you please review and let me know if i missed any?

pvaneck:
Thanks @Suresh-Nakkeran. I think this is pretty much ready to merge. A new Developer Certificate of Origin (DCO) check was recently added. Can you follow the steps [here](https://github.com/kserve/kserve/pull/1926/checks?check_run_id=4461127469) to get that check passing?

config/runtimes/kserve-mlserver.yaml
```diff
@@ -0,0 +1,37 @@
...
+          value: "9000"
+        - name: "MODELS_DIR"
+          value: "/mnt/models"
+      resources:
+        requests:
+          cpu: "1"
+          memory: 2Gi
+        limits:
+          cpu: "1"
+          memory: 2Gi
```
yuzisun, line 37: @Suresh-Nakkeran What happens if user specified the resource requirement on inference service yaml, do we overwrite that?

yuzisun:
@Suresh-Nakkeran Awesome work! 

I added a [commit](https://github.com/kserve/kserve/pull/1926/commits/8d375950e60d5a1073512c534829690b28bfc9fb) to fix the protocol version.

/lgtm
/approve

PR #1933, opened 2021-12-02
Support grpc for transformer

### LINKED ISSUE #1796: transformer gRPC support
/kind feature

Currently the [default transformer `predict` handler](https://github.com/kubeflow/kfserving/blob/master/python/kfserving/kfserving/kfmodel.py#L138
) only supports REST call to the predictor, when predictor is using gRPC protocol user has to overwrite the `predict` handler to make the gRPC call with v2 protocol, so we'd like to provide a default implementation similar to the way we did for REST.

1: Add a protocol field on the `InferenceService` [predictor spec](https://github.com/kubeflow/kfserving/blob/master/pkg/apis/serving/v1beta1/predictor.go#L76)

2: We then pass the protocol field to the transformer, so if it is gRPC we can make a async gRPC call with `tritonclient`.
```python
import grpc

def __init__(self, name: str, predictor_host: str):
    channel_options = (("grpc.lb_policy_name", "round_robin"),)
    _channel = grpc.aio.insecure_channel(self.predictor_host, options=channel_options)
    self._client_stub = service_pb2_grpc.GRPCInferenceServiceStub(_channel)

async def predict(self, features: Dict) -> Dict:
    if self.protocol == "gRPC":
        request = _get_inference_request(
            model_name=self.model_name, inputs=inputs, outputs=outputs)
        async_result = await self._client_stub.ModelInfer(
            request=request, timeout=self.predict_timeout)
        response = InferResult(async_result)
```

3: We also want to generate the distributed trace for this transformer->predictor call with [opentelemetry tracing API](https://opentelemetry-python.readthedocs.io/en/latest/getting-started.html).

- Add e2e tests
- Add a sklearn gRPC example doc

xcjason:
Make gRPC predictor call work out of the box in the transformer.

Fixes #1796 

```
kind: InferenceService
metadata:
  name: torch-transformer
spec:
  predictor:
    triton:
      storageUri: gs://kfserving-examples/models/torchscript
      runtimeVersion: 20.10-py3
      env:
      - name: OMP_NUM_THREADS
        value: "1"
      ports:
      - name: h2c
        protocol: TCP
        containerPort: 9000
      args:
      - --log-verbose=1
  transformer:
    containers:
    - image: yuzisun/grpc-image-transformer:latest
      name: kfserving-container
      command:
      - "python"
      - "-m"
      - "model"
      args:
      - --model_name
      - cifar10
      - --protocol
      - grpc-v2
```
      

**Release note**:

```release-note

```


.github/workflows/python-test.yml
```diff
@@ -42,13 +42,19 @@ jobs:
...
         pip install -e ./xgbserver
+        pytest ./xgbserver
         pip install -e ./pytorchserver
+        pytest ./pytorchserver
         pip install -e ./pmmlserver
+        pytest ./pmmlserver
         pip install -e ./lgbserver
+        pytest ./lgbserver
         pip install -e ./paddleserver[test]
-        pytest
```
andyi2it, line 54: Any specific reason for not having a single `pytest` at the end?

docs/samples/v1beta1/triton/torchscript/torch_grpc_transformer.yaml
```diff
@@ -0,0 +1,31 @@
...
+        value: "1"
+      ports:
+      - name: h2c
+        protocol: TCP
+        containerPort: 9000
+      args:
+      - --log-verbose=1
+  transformer:
+    containers:
+    - image: yuzisun/grpc-image-transformer:latest
```
andyi2it: Should be changed to use `kserve` instead of `yuzisun` isn't it?

docs/samples/v1beta1/triton/torchscript/torch_grpc_transformer.yaml
```diff
@@ -0,0 +1,31 @@
...
+      name: kfserving-container
```
andyi2it: I suppose its `kserve-container`

python/custom_transformer/requirements.txt
```diff
@@ -0,0 +1,2 @@
+kserve
+torchvision
```
andyi2it, line 2: May not be required since we have `setup.py`?

.github/workflows/python-test.yml
yuzisun: We are running into issues where different sub projects have dependency conflicts, so testing them separately to make sure it is testing with the right requirement versions.

yuzisun:
@pvaneck Can you help take a look?

docs/samples/v1beta1/triton/torchscript/README.md
```diff
@@ -6,7 +6,6 @@ C++ is very often the language of choice, The following example will outline the
 to a serialized representation that can be loaded and executed purely from C++ like Triton Inference Server, with no dependency on Python.
 
 ## Setup
-1. Your ~/.kube/config should point to a cluster with [KFServing 0.5 installed](https://github.com/kubeflow/kfserving/#install-kfserving).
```
andyi2it, line 9: I suppose it was meant to be `kserve` 0.7 installed. with corresponding reference to the new repo.

docs/samples/v1beta1/triton/torchscript/image.json
```diff
@@ -0,0 +1,9 @@
...
...rkRaAh/evx0+MLf+KBQldz3fd9PkjhuNO7evfOH//p9sRqHre7x2dW1/eHhnZc9yW8Mh8v54uGjr9HZ86Val7ay3npZdPv738yK1rXmzPMIqqWxjvs1Kv7F8dXw3mtIcmoMQbfebJbLabv10ttv/eil79x9999/SylrNrO93f24kTKTt/p8cKhXgf/5gwejLXWi0ey3OzebjPvW0ScuOh5byWhZVYUhBhk/WgVTmzhRgVo5ZABsd9B98/WXfWEPD/b+8Z//5de/fW86Xo1WWFXHkph5aY5Px0Rp17mTdUMkjlKBfohUautWVvhC+pzmtNBCONT8aAm/+5//femg05dRKPig3x90Gjdv7BOnRpPZO79877MvHtaVMoYQB84q6zUsCE4CQ5mBwOeEOFopcEA59xmiq4whKBAYBaUpbEH+8bOjX73/0deT7da6k6dfX+tlvhBbxd/9z48/f3hRGM/yBgQp8ROIm+ABYbamUFlrra4NqYxzAIxBGMo0EIEQVEZWhNpRmaS83dmZL9xosbz/4LHVB4TInf4+Zd5Hn/zfe+9/UGNIuAcAhBBbK4cO0TrnrKOCc8oYYZIzxhhPkpgBgNPWARJBLPb7zaTR5JwxITxTyWeX6zp/9IOXbwfpYFXhn/7ySeWMNtrzfEQsioIQwiinlBBHPMYpcAKcemEQBJxzrc0mzy262mAz6/QGndjn5WbD0VjiAJmvCLva1p89uXi7cBu3OV9svDg2BavqOgwDLnhV1xQYUCY4d8AdAeH5W22VyYMgcM7VBvNKxWkn3ekro548fizQAkFHHDImEHwr4mdXm3fe/f2j08uTi0leayRO+JJJGSZxI20SSrU2da2cI4wxrQ1jlBJXFtsi31Li0qzV6w+ms/nx8fHp0RNiLW+laVVt8lJJFhiDILz//ujLk4uLVa7n29IoEkWxQfQ8j0vpB5YB40JaAgYdReectVorrQLf77TbWWegHNSSl55ELvKq5HVVekBqqwWThhEHAEF8ejEBzox2xmBVVXmeA4DneZEUQeADoPS9IIyVMtP5HInhArJG1Gul/X5rmdeb5WK7Wqat1nQy5XVZeYyGnKAuKSNIEB0iYUY5Z6lzzjmHiACwWCzmumzEUTNrNRj4xLdYc2qZx+qq9jjl1JpiZYp6u5yhVr4nKsb+CkyFkScvikzRAAAAAElFTkSuQmCC"
```
pvaneck: nit: the spacing seems to be off here with b64 being underindented

python/custom_transformer/model.py
```diff
@@ -0,0 +1,83 @@
...
+
+parser = argparse.ArgumentParser(parents=[kserve.kfserver.parser])
+parser.add_argument(
+    "--predictor_host", help="The URL for the model predict function", required=True
+)
+parser.add_argument(
+    "--protocol", help="The protocol for the predictor", required=True
+)
+parser.add_argument(
+        "--model_name",
```
pvaneck: nit: inconsistent indentation compared to the other `add_argument`s above.

docs/samples/v1beta1/triton/torchscript/image.json
```diff
@@ -0,0 +1,9 @@
+{
+    "instances":[
```
pvaneck: I suppose at some point we want the user entry point to be v2 format, right?

pvaneck (commented):
Overall looks good. Tried it out locally and was able to run the Isvc and perform inference. Left some comments inline.

docs/samples/v1beta1/triton/torchscript/image.json
yuzisun: I think for transformer we should support v2 protocol as well, however a lot of time data may come in out of your control so transformer also serves as an adapter layer to translate between user input protocol and kserve inference protocol, if user has control for the original data source then they should align with the kserve v2 protocol.

PR #1934, opened 2021-12-06
Update CloudEvent Handling in KFServer

### LINKED ISSUE #1671: Cloud events responses need to be configurable for custom predictors and transformers
/kind feature

Cloud events responses need to be configurable for custom predictors and transformers. Currently it returns the same Ce-Type and Ce-Source, which means that the transformer/predictor will get those events again except this time without the proper payload, at any case they need to be configurable so the proper consumer can pick them up. 

I already have an example how this should be done but I am unsure how to allow for configuration view the custom spec. Should this be env variables? 

https://github.com/pwc-labs/kfserving/blob/ce-time-fix/python/kfserving/kfserving/handlers/http.py#L85

Also note that: RFC 3339 is not compatible with ISO. I corrected this as well. 

```python
self.set_header('ce-time', datetime.utcnow().replace(tzinfo=pytz.utc).trftime('%Y-%m-%dT%H:%M:%SZ'))
````

### LINKED ISSUE #1923: KServe CloudEvents infinite-loop using Knative Eventing Broker
/kind bug

We're using Knative eventing `Broker` and `Trigger` with a filter for triggering the `PredictHandler` handler in a custom model server using KServe python library.
Currently, KServe python library sends the response CloudEvent with the same attributes from the original CloudEvent and there is no option to filter the response CloudEvent to avoid the **infinite-loop** of CloudEvents being sent to KServe from the `Broker`.

We expect to at least change the response CloudEvent `source` field to KServe instance name or some env variable.

Line of problematic code: https://github.com/kserve/kserve/blob/464308defb7245110a3dbf255629d7f418bb9947/python/kserve/kserve/handlers/http.py#L99

**Environment:**

- Istio Version: 1.11.4
- Knative Version: 1.0
- KFServing Version: 0.7
- Kubeflow version: N/A
- Kfdef: N/A
- Minikube version: N/A
- Kubernetes version: (use `kubectl version`): 1.21.2
- OS (e.g. from `/etc/os-release`): Amazon Linux 2

### PR DESCRIPTION
**Pull request by markwinter**

This PR updates the handling of CloudEvent requests to the KFServer.

There were two main issues:
- The response CE attributes were just a copy of the request CE attributes. This is incorrect for a couple of reasons:
  - Each event should have a unique ID
  - The source and type fields are inherently incorrect
  - Not usable with Knative Eventing `Trigger` as the response will match the request filter in the `Trigger`
- A CloudEvent was only returned as a response if the request was a binary CE. If you sent a structured CE then the server would respond with normal a KFServer JSON response.
  - Now a structured CE request will get a structured CE response

I've also made it so that the CE `source` and `type` fields can be changed using environment variables `CE_SOURCE` `CE_TYPE`

Users can also optionally merge CE fields if they set the environment variable `CE_MERGE` to `"true"`

Fixes #1923
Fixed #1671

**Release note**:

```release-note
Update CloudEvent handling
  - Structured CloudEvent requests now get a structured CloudEvent as a response
  - Correctly set CloudEvent id, time, source, type fields
  - Source and type fields can be overwritten by env variables `CE_SOURCE` `CE_TYPE`
  - Optionally merge original CloudEvent request attributes with response attributes with env variable `CE_MERGE`
```

### DIFF
--- a/python/kserve/kserve/handlers/http.py
+++ b/python/kserve/kserve/handlers/http.py
@@ -10,19 +10,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from typing import Any
 
 import tornado.web
 import json
-import pytz
 import cloudevents.exceptions as ce
-from cloudevents.http import CloudEvent, from_http, is_binary, is_structured, to_binary, to_structured
+from cloudevents.http import from_http
 from cloudevents.sdk.converters.util import has_binary_headers
 from http import HTTPStatus
 from kserve.kfmodel_repository import KFModelRepository
 from kserve.kfmodel import ModelType
-from datetime import datetime
+from kserve.utils.utils import is_structured_cloudevent, create_response_cloudevent
 
 from ray.serve.api import RayServeHandle
 
@@ -65,20 +63,30 @@ class HTTPHandler(BaseHandler):
 
 
 class PredictHandler(HTTPHandler):
+    def get_binary_cloudevent(self) -> str:
+        try:
+            # Use default unmarshaller if contenttype is set in header
+            if "ce-contenttype" in self.request.headers:
+                event = from_http(self.request.headers, self.request.body)
+            else:
+                event = from_http(self.request.headers, self.request.body, lambda x: x)
+
+            return event
+        except (ce.MissingRequiredFields, ce.InvalidRequiredFields, ce.InvalidStructuredJSON,
+                ce.InvalidHeadersFormat, ce.DataMarshallerError, ce.DataUnmarshallerError) as e:
+            raise tornado.web.HTTPError(
+                status_code=HTTPStatus.BAD_REQUEST,
+                reason="Cloud Event Exceptions: %s" % e
+            )
+
     async def post(self, name: str):
+        is_cloudevent = False
+        is_binary_cloudevent = False
+
         if has_binary_headers(self.request.headers):
-            try:
-                # Use default unmarshaller if contenttype is set in header
-                if "ce-contenttype" in self.request.headers:
-                    body = from_http(self.request.headers, self.request.body)
-                else:
-                    body = from_http(self.request.headers, self.request.body, lambda x: x)
-            except (ce.MissingRequiredFields, ce.InvalidRequiredFields, ce.InvalidStructuredJSON,
-                    ce.InvalidHeadersFormat, ce.DataMarshallerError, ce.DataUnmarshallerError) as e:
-                raise tornado.web.HTTPError(
-                    status_code=HTTPStatus.BAD_REQUEST,
-                    reason="Cloud Event Exceptions: %s" % e
-                )
+            is_cloudevent = True
+            is_binary_cloudevent = True
+            body = self.get_binary_cloudevent()
         else:
             try:
                 body = json.loads(self.request.body)
@@ -87,6 +95,10 @@ class PredictHandler(HTTPHandler):
                     status_code=HTTPStatus.BAD_REQUEST,
                     reason="Unrecognized request format: %s" % e
                 )
+
+            if is_structured_cloudevent(body):
+                is_cloudevent = True
+
         # call model locally or remote model workers
         model = self.get_model(name)
         if not isinstance(model, RayServeHandle):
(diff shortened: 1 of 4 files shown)


python/kserve/kserve/handlers/http.py
```diff
@@ -65,20 +66,68 @@ def get_model(self, name: str):
 
 
 class PredictHandler(HTTPHandler):
+    def create_cloudevent(self, model_name: str, body: Union[Dict, CloudEvent], response: Dict,
+                          binary_event=False) -> Dict:
+        ce_attributes = {}
+
+        if os.getenv("CE_MERGE", "false") == "true":
```
shahar3012: Maybe use the `bool()` function for comparison rather than comparing strings?

python/kserve/kserve/handlers/http.py
```diff
@@ -65,20 +66,68 @@ def get_model(self, name: str):
...
+
+        for k, v in event_headers.items():
+            if k == "ce-time":
+                # Set the time with the timezone data
+                self.set_header('ce-time', datetime.utcnow().replace(tzinfo=pytz.utc).
+                                strftime('%Y-%m-%dT%H:%M:%S.%f%z'))
+            else:
+                self.set_header(k, v)
+
+        self.set_header("Content-Type", "application/cloudevents+json")
```
shahar3012: We should set the `Content-Type` header according to the Content Mode - binary or structured:

- For **binary** content mode we use `Content-Type: application/json; charset=utf-8` - the body contains only the data payload.
- For **structured** content mode we use `Content-Type: application/cloudevents+json; charset=UTF-8` - the body contains both the data payload and the CloudEvents attributes.

shahar3012 (changes requested):

python/kserve/kserve/handlers/http.py
markwinter: Thanks. I was trying to find info like this before 👍

python/kserve/kserve/handlers/http.py
markwinter: Do you mean like `bool(str)` ? This returns True for any non empty string, and False for empty string.

I've added `.lower()` to help match more cases

python/kserve/kserve/handlers/http.py
```diff
@@ -65,20 +64,63 @@ def get_model(self, name: str):
 
 
 class PredictHandler(HTTPHandler):
+    def create_cloudevent(self, model_name: str, body: Union[Dict, CloudEvent], response: Dict,
```
yuzisun: Can we move this function to utils? suggest renaming to `create_response_cloudevent` as it is specific to response.

python/kserve/kserve/kfmodel.py
```diff
@@ -110,38 +111,25 @@ def load(self) -> bool:
 
     async def preprocess(self, request: Dict) -> Union[Dict, ModelInferRequest]:
```
yuzisun: so request type should be really `Union[Dict, CloudEvent]` here?

python/kserve/kserve/kfmodel.py
```diff
@@ -110,38 +111,25 @@ def load(self) -> bool:
...
-                    attributes = request._attributes
-                    if "content-type" in attributes:
-                        if attributes["content-type"] == "application/cloudevents+json" or \
-                           attributes["content-type"] == "application/json":
-                            raise tornado.web.HTTPError(
-                                status_code=HTTPStatus.BAD_REQUEST,
-                                reason="Unrecognized request format: %s" % e
-                            )
+            try:
+                request = json.loads(request.data.decode('UTF-8'))
```
yuzisun: I think this will fail the decoding for avro or protobuf case while the original code allows it

markwinter: Oh I see. It's a best effort try to decode it otherwise leave the request alone.

I was confused because the `DummyAvroCEModel` in `tests/test_server` overrides the `preprocess` to handle the Avro case.

I'll update

markwinter:
@yuzisun 
Applied your suggestions, thanks

python/kserve/kserve/kfmodel.py
```diff
@@ -108,40 +110,35 @@ def load(self) -> bool:
         self.ready = True
         return self.ready
 
-    async def preprocess(self, request: Dict) -> Union[Dict, ModelInferRequest]:
+    async def preprocess(self, request: Dict) -> Union[Dict, CloudEvent, ModelInferRequest]:
```
yuzisun: I think if it is `CloudEvent` with protobuf/avro encoding, user should overwrite preprocess to decode with the corresponding data schema as KServe dose not know how to decode the message, so the return type should still be a `Dict`. The avro test case is a bit misleading, should move [`_parserequest`](https://github.com/kserve/kserve/blob/master/python/kserve/test/test_server.py#L136) into the `preprocess` handler instead of `predict`. Here we can change the request type to `Union[Dict, CloudEvent]` and leave return type as is.

markwinter: If users must overwrite preprocess for avro/protobuf, then do we still need this part? As we are assuming only CloudEvent JSON UTF-8 in default preprocess

https://github.com/kserve/kserve/blob/b02eb376271a6473ee04b449a64942e635372cc0/python/kserve/kserve/kfmodel.py#L129-L135

yuzisun: In your code there is a slight behavior change, it now returns `request`(CloudEvent) instead of `request.data`(Payload) for the binary cloudevent case with binary payload.

So we actually have something like following in `preprocess` to deal with different payload format. Welcome any suggestion for a better way doing this.
```python
def preprocess(self, inputs: Union[Dict, CloudEvent]) -> Dict:
        # We call preprocess form super class to return raw input. Its follows below interface.
        # if inputs is python dict, then output will be python dict
        # if inputs is CloudEvent containing msg in json format, then output will be python dict
        # if inputs is CloudEvent containing msg in raw bytes based format,
        # then output will be raw bytes
        inputs = super().preprocess(inputs)

        # if input is json/python dict, no decoding is required.
        # for raw bytes, msg will be decoded according to its schema

        if isinstance(inputs, bytes):            
            # Attempt parsing bytes as AVRO format
            parsed_avro_schema = avro.schema.parse(self.AVRO_SCHEMA)
            reader = avro.io.DatumReader(parsed_avro_schema)
            bytes_reader = io.BytesIO(inputs)
            decoder = avro.io.BinaryDecoder(bytes_reader)
            inputs = reader.read(decoder) 
```

markwinter: Ah I see now thanks 👍 

For now I will keep it the same so the CloudEvent changes can be merged. But I might revisit this later to see if I can make the process simpler/clearer.

yuzisun:
Thanks @markwinter !! Awesome work always.

/approve

markwinter:
@yuzisun 
Thanks!

PR #1948, opened 2021-12-15
added autoSelect field in framework

### LINKED ISSUE #1924: Follow up work for ServingRuntime support

Suresh-Nakkeran:

Fixes #1924 

1. Add `AutoSelect` field so only designated serving runtime can be automatically selected otherwise user needs to specify the the runtime name to pick it. 

**Release note**:

```release-note

```


pvaneck (commented):
Hey, thanks. I will take a closer look tomorrow, but for now, make sure to run the generation commands to regenerate some of the code/yaml. I think this is `make generate` and `make manifests`. This is needed anytime API changes are made.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -135,11 +135,14 @@ func (m *ModelSpec) getModelTypeLabel() string {
...
 
-func getServingRuntimeSupportedModelTypeLabelSet(supportedModelTypes []v1alpha1.Framework) stringSet {
+func (m *ModelSpec) getServingRuntimeSupportedModelTypeLabelSet(supportedModelTypes []v1alpha1.Framework) stringSet {
 	set := make(stringSet, 2*len(supportedModelTypes)+1)
 
 	// model type labels
 	for _, t := range supportedModelTypes {
+		if m.Runtime == nil && t.AutoPlace == nil || *t.AutoPlace == false {
+			continue
+		}
```
pvaneck: The boolean check  at the end can cause a nil pointer exception. Suggestion:
```
// If runtime isn't explicitly set, only add labels for modelTypes where AutoPlace is true.
if m.Runtime != nil || (t.AutoPlace != nil && *t.AutoPlace) {
```

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -30,6 +30,9 @@ type Framework struct {
 	// Can be "major", "major.minor" or "major.minor.patch".
 	// +optional
 	Version *string `json:"version,omitempty"`
+	// Set to true to allow use of this framework for auto placement
```
pvaneck:
```suggestion
	// Set to true to allow the ServingRuntime to be used for automatic model placement if
	// this framework is specified with no explicit runtime. 
```

config/crd/serving.kserve.io_inferenceservices.yaml
```diff
@@ -4943,6 +4943,8 @@ spec:
                           type: array
                         framework:
                           properties:
+                            autoPlace:
```
pvaneck: Hmm, seems that one side effect of using the same struct for both the Predictor ModelSpec and ServingRuntime Spec is that this gets added to the InferenceService schema where it doesn't really make sense to have it. I'm not entirely sure of the best way to handle this. Perhaps two different structs? (i.e.  Rename Framework in `servingruntime_types.go`  to `ModelType` to keep with the current SR naming conventions, then in `predictor_model.go`, create a `Framework` struct with just the Name and Version fields.)

pvaneck (commented):
Hey, thanks for generating the additional code/files. I left some more comments inline.

@njhill Would be useful to get your feedback on having this field in lieu of a `disabled` type of field for controlling automatic placement. I suppose one difference is that having a `disabled` type of field implies that a SR will support automatic placement for all listed modelTypes unless `disabled` is true for each specific modelType. On the otherhand, with `autoPlace`, a SR's modelType would need this explicitly set to true to be used in automatic placement consideration. What do you think is preferred?

Also open to alternative naming suggestions. /cc @yuzisun

Suresh-Nakkeran:
@pvaneck @yuzisun I incorporated all review comments. can you please review?

hack/update-openapigen.sh
```diff
@@ -30,6 +30,7 @@ go run k8s.io/kube-openapi/cmd/openapi-gen \
 
 # Hack, the name is required in openAPI specification even if set "+optional" for v1.Container in PredictorExtensionSpec.
 sed -i'.bak' -e 's/Required: \[\]string{\"name\"},//g' $OPENAPI_SPEC_FILE && rm -rf $OPENAPI_SPEC_FILE.bak
```
yuzisun: this line can be removed.

config/crd/serving.kserve.io_inferenceservices.yaml
yuzisun: @pvaneck I think it is better to keep the names consistent between SR and InferenceService, I prefer `Framework` as `ModelType` in ML world often means binary/multiclass classifier or regression type of model.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -135,14 +146,17 @@ func (m *ModelSpec) getModelTypeLabel() string {
...
+func (m *ModelSpec) getServingRuntimeSupportedModelTypeLabelSet(supportedModelTypes []v1alpha1.ModelType) stringSet {
 	set := make(stringSet, 2*len(supportedModelTypes)+1)
 
 	// model type labels
 	for _, t := range supportedModelTypes {
-		set.add("mt:" + t.Name)
-		if t.Version != nil {
-			set.add("mt:" + t.Name + ":" + *t.Version)
+		// If runtime isn't explicitly set, only add labels for modelTypes where AutoPlace is true.
+		if m.Runtime != nil || (t.AutoPlace != nil && *t.AutoPlace) {
```
njhill: I don't think it makes sense to have the `m.Runtime` check here since AFAICS this code path already assumes that `m.Runtime == nil`.

njhill (commented):
Thanks @Suresh-Nakkeran this looks generally good to me, I have a couple of thoughts on the name/behaviour though:
- I'm wondering what the best default setting should be for `autoPlace`, we could consider changing it to something like `disableAutoPlacement` which I think would be a better "out of the box" experience, otherwise choosing just the type/framework won't work initially. This is a bit like the `disabled` thing we had previously discussed but of course won't disable the runtime if specified explicitly.
- I think it would be more accurate to use the word "select/selection" rather then "place" in this context (i.e. could rename to `autoSelect` or `disableAutoSelect`)

config/crd/serving.kserve.io_inferenceservices.yaml
njhill: I guess we should then rename the field in SR to be something like `supportedModelFrameworks`?

Just throwing it out there but how about `ModelFormat` / `supportedModelFormats`?

njhill:
@pvaneck kind of related, I was looking at the [`GetSupportingRuntimes`](https://github.com/kserve/kserve/pull/1948/files#diff-4422c60e7b70f271c7cd09c3c28b90ea119be418ec984e088a295bf429a3b891R101) func and see that it's only called right now with `isMMs` as `false`. I think this was discussed but are we still planning to add a field to the InferenceService somewhere to allow explicitly indication of this? If it's too late to change the CRD I guess an annotation could serve that purpose for now.

Importantly though, I think the behaviour of the func above should be adjusted to take a `*bool` (and spec field optional) and in the "unspecified" case it could choose either kind of runtime. I.e. a given SR must be either MMS or non-MMS but an InferenceService doesn't _have_ to be pinned to either kind. This should hopefully work as long as we make the selection logic (ordering) deterministic and consistent between kserve and mm controllers.

pkg/apis/serving/v1beta1/predictor_model.go
pvaneck: I believe this check is for https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L95 where a runtime is explicitly specified, and the given framework/version in the predictor is checked against the supportingModelTypes in the SR. Perhaps it may make sense to remove that check altogether, and just use the ServingRuntime regardless if a user explicitly specifies it? WDYT?

config/crd/serving.kserve.io_inferenceservices.yaml
pvaneck: I am fine with either `supporedModelFrameworks` or `supportedModelFormats`. Do note, that this will require a small followup PR in ModelMesh-Serving to adjust the SR spec there as well. (Eventually, we'll consolidate to a single source)

pvaneck:
Thanks, @njhill.

> I think it would be more accurate to use the word "select/selection" rather then "place" in this context (i.e. could rename to autoSelect or disableAutoSelect)

I am also somewhat leaning towards `disableAutoSelect` as well so that frameworks in a runtime are enabled by default as opposed to disabled by default. A user won't have to remember to put `autoPlace: true` on every supported framework in the list if they make a custom SR.

> I think this was discussed but are we still planning to add a field to the InferenceService somewhere to allow explicitly indication of this? If it's too late to change the CRD I guess an annotation could serve that purpose for now.

The `"serving.kserve.io/deploymentMode": "ModelMesh"` annotation is currently the only way to denote a model should be MMS, in which case, the kserve-controller won't even process the predictor, leaving it the the MM-Serving controller. I'm assuming this will eventually be a first class field or at least something like `isMMS`. But yea, I think allowing placement in either type runtime could be useful (though not in this PR, ofc). Would this be for cases where `SomeFramework` is only supported by a MM runtime, and no SMS runtimes, and a user deploys an ISvc with no explicit MMS indication?

yuzisun:
> Thanks, @njhill.
> 
> > I think it would be more accurate to use the word "select/selection" rather then "place" in this context (i.e. could rename to autoSelect or disableAutoSelect)
> 
> I am also somewhat leaning towards `disableAutoSelect` as well so that frameworks in a runtime are enabled by default as opposed to disabled by default. A user won't have to remember to put `autoPlace: true` on every supported framework in the list if they make a custom SR.


I think disableAutoSelect might be more difficult to work with for user as almost every framework we now have at least two SRs supporting it.

- auto placement can only pick one SR so it is better to be explicit, user can decide which SR they want to use for a given framework and then set that to be true. It is also easier to review which SR is chosen for auto replacement for the framework.
- With disableAutoSelect they have to go over the list of supported SRs and disable all other ones except the one they want auto placement which can be error prone.

config/crd/serving.kserve.io_inferenceservices.yaml
yuzisun: Now I think `ModelFormat`/`supportedModelFormats` might make more sense as we have `onnx`, `pmml`, `tensorrt` on the list..

pkg/apis/serving/v1beta1/predictor_model.go
njhill: @pvaneck ah yes I missed this, you're right.. sorry @Suresh-Nakkeran

njhill:
> I think disableAutoSelect might be more difficult to work with for user as almost every framework we now have at least two SRs supporting it.

@yuzisun as per our slack exchange I'm fine with this... so @Suresh-Nakkeran's latest `autoSelect` changes LGTM.

pvaneck:
I'm fine with `autoSelect` as well. Just the failing Docker publisher job needs to be re-run.
/lgtm

PR #1951, opened 2021-12-17
Rename KF prefixed classes

markwinter:
**What this PR does / why we need it**:

1. Renames the following:
  - `KFModel` -> `Model`
  - `KFServer` -> `ModelServer`
  - `KFModelRepository` -> `ModelRepository`
  - `kserve.api.kf_serving_client` -> `kserve.api.kserve_client`
  - `kserve.api.kf_serving_watch` -> `kserve.api.watch`

Also updated some comments that said KFServing still

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1947

**Release note**:

```release-note
KF-prefixed components have been renamed
  - `KFModel` -> `Model`
  - `KFServer` -> `ModelServer`
  - `KFModelRepository` -> `ModelRepository`
  - `kserve.api.kf_serving_client` -> `kserve.api.kserve_client`
  - `kserve.api.kf_serving_watch` -> `.kserve.api.watch`
```

PR #1954, opened 2021-12-18
sklearnserver/xgbserver version upgrade

yuzisun:
**What this PR does / why we need it**:
upgrade sklearn to 1.0.1 and xgboost version to 1.5.0

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1791 

**Special notes for your reviewer**:
other improvements include:
1. look for models from model repository only if the model file is missing from the parent directory
2. updated test model artifacts with sklearn 1.0.1 and xgboost 1.5.0 version
3. Support predict_proba for sklearnserver

**Release note**:

```release-note
- Upgrade scikit-learn to 1.0.1 for sklearnserver
- Update xgboost to 1.5.0 for xgbserver
- Update alibi to 0.6.2
```

PR #1955, opened 2021-12-20
Remove h5py installation step in python test workflow

markwinter:

h5py 2.9 is not compatible with the tensorflow version required by alibiexplainer. Should just use the version pulled in by alibiexplainer

https://github.com/kserve/kserve/pull/1951#issuecomment-997393658


yuzisun:
thanks! removed in [my PR](https://github.com/kserve/kserve/pull/1954) as well.

/lgtm
/approve

PR #1964, opened 2021-12-27
Update transformer e2e tests 

### LINKED ISSUE #1667: Transformer not working with Tensorflow tensors in column format
/kind bug

Following this guide: https://github.com/kubeflow/kfserving/tree/master/docs/samples/v1beta1/transformer/torchserve_image_transformer

I created a transformer converting base64 to data with tensors in column format to be sent to a predictor running tensorflow  serving 2.5.1.

Here's a part of the final data I send - the result of `preprocess` of `ImageTransformer(kfserving.KFModel)`
```json
{
   "inputs": {
      "input_image": [
         [
              ...data
         ]
    }
}
```

But I get an error because `inputs` is not a list. I believe it is from this line here https://github.com/kubeflow/kfserving/blob/master/python/kfserving/kfserving/handlers/http.py#L45

However, tensorflow serving does not allow it being a list, but expects a dict/map (as far as I understand) https://www.tensorflow.org/tfx/serving/api_rest#specifying_input_tensors_in_column_format

Not to get an error that inputs has to be a list.

The request succeeds when I remove the validation check. 

I'm pretty new to this, so I might be misunderstanding something. But I think the data should be allowed to be a dict for tensorflow serving.

yuzisun:
Update transformer e2e test to use image built from CI pipeline.

Fixes #1667 

1. Fix the tfserving input validation, `inputs` is a `Dict` in tfserving v1 protocol while it is a `List` in v2 kserve inference protocol.
2. Fix the issue for e2e transformer test that it was not testing using the image built from CI and instead it used the image with latest tag which can potentially cause issue when several PRs simultaneously running CI.

**Release note**:

```release-note

```


yuzisun:
@Suresh-Nakkeran Please take a look, this PR fixes the transformer e2e test that always uses the image built from CI. Also added printing events in the post e2e test step for easier debugging.

Suresh-Nakkeran:
\lgtm

PR #1971, opened 2022-01-02
Fixes Azure blob download as class is no longer hashable

laozc:
Signed-off-by: Zhongcheng Lao <Zhongcheng.Lao@microsoft.com>

This fixes an issue in storage-initializer v0.7.0 (azsdk-python-storage-blob/12.8.1)
which prevents Azure Blob to be downloaded as model files.

```code
  File "/storage-initializer/scripts/initializer-entrypoint", line 14, in <module>
    kserve.Storage.download(src_uri, dest_path)
  File "/usr/local/lib/python3.7/site-packages/kserve/storage.py", line 71, in download
    Storage._download_blob(uri, out_dir)
  File "/usr/local/lib/python3.7/site-packages/kserve/storage.py", line 208, in _download_blob
    for blob in set(blobs):
TypeError: unhashable type: 'BlobProperties'
```

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -221,7 +221,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
                 else:
                     blobs += container_client.list_blobs(name_starts_with=item.name,
                                                          include=['snapshots'])
-        for blob in set(blobs):
+        for blob in blobs:
             dest_path = os.path.join(out_dir, blob.name.replace(prefix, "", 1).lstrip("/"))
             Path(os.path.dirname(dest_path)).mkdir(parents=True, exist_ok=True)
             logging.info("Downloading: %s to %s", blob.name, dest_path)


python/kserve/kserve/storage.py
```diff
@@ -221,7 +221,7 @@ def _download_blob(uri, out_dir: str):  # pylint: disable=too-many-locals
                 else:
                     blobs += container_client.list_blobs(name_starts_with=item.name,
                                                          include=['snapshots'])
-        for blob in set(blobs):
+        for blob in blobs:
```
yuzisun, line 224: @maganaluis Can you help review here? not sure if we had `set` there for any reason.

maganaluis: @yuzisun It should be okay to remove it. Looking back to the upgrade to 12.X.X there were a lot of changes to the mock unit-tests, I don't think there should be duplicates keys coming from the client. So this must have been to handle those unit-tests.

yuzisun:
Thanks @laozc!!

/lgtm
/approve

laozc:
Thanks. Glad to contribute to kserve.

PR #1973, opened 2022-01-04
Organise handlers into separate files

markwinter:

Currently handlers are split between `model_server.py` and `handlers/http.py`.

This PR moves the handlers to separate files under the `handlers` directory. I think this makes things a bit more organised


python/kserve/kserve/handlers/explain.py
```diff
@@ -0,0 +1,27 @@
+import json
```
yuzisun, line 15: Can you help add the license on top?

python/kserve/kserve/handlers/base.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2021 The KServe Authors.
```
yuzisun: 2021 -> 2022

python/kserve/kserve/handlers/liveness.py
```diff
@@ -0,0 +1,6 @@
+from kserve.handlers.base import BaseHandler
+
+
+class LivenessHandler(BaseHandler):  # pylint:disable=too-few-public-methods
```
yuzisun: this can be merged to the `health.py`?

python/kserve/kserve/handlers/list.py
```diff
@@ -0,0 +1,10 @@
+from kserve.handlers.base import BaseHandler
+from kserve.model_repository import ModelRepository
+
+
+class ListHandler(BaseHandler):
```
yuzisun: I think `ListHandler` and `LoadHandler` can be merged into `model_management.py`

markwinter:
@yuzisun Thanks for reviewing. I've applied your suggestions.

yuzisun:
Thanks @markwinter for the nice refactoring!

/lgtm
/approve

PR #1984, opened 2022-01-11
fix: allow to set worker count more than 1

### LINKED ISSUE #1759: Setting workers more than 1
If I set `workers=2` when starting KFServer, it always kills the 2nd worker with the following error message

```
  File "/usr/lib/python3.8/asyncio/selector_events.py", line 261, in _add_reader
    key = self._selector.get_key(fd)
  File "/usr/lib/python3.8/selectors.py", line 192, in get_key
    raise KeyError("{!r} is not registered".format(fileobj)) from None
KeyError: '6 is not registered'

[I 210811 10:50:57 kfserver:151] Registering model: test
[I 210811 10:50:57 kfserver:121] Setting asyncio max_workers as 12
[I 210811 10:50:57 kfserver:128] Listening on port 8080
[I 210811 10:50:57 kfserver:130] Will fork 2 workers
[I 210811 10:50:57 process:123] Starting 2 processes
[E 210811 10:50:57 __main__:51] fail to load model test from dir /home/swapnesh/test/test_files/. exception type <class 'FileExistsError'>, exception msg: [Errno 17] File exists
[I 210811 10:50:58 process:163] child 1 (pid 5829) exited normally
```

Why can't I set more than 1 worker when max is 12?

Suresh-Nakkeran:
according to https://github.com/tornadoweb/tornado/issues/2426#issuecomment-400895086, the async io initialization code should be done after the "fork".

fixes #1759

### DIFF
--- a/python/kserve/kserve/model_server.py
+++ b/python/kserve/kserve/model_server.py
@@ -117,10 +117,6 @@ class ModelServer:
             # formula as suggest in https://bugs.python.org/issue35279
             self.max_asyncio_workers = min(32, utils.cpu_count()+4)
 
-        logging.info(f"Setting asyncio max_workers as {self.max_asyncio_workers}")
-        asyncio.get_event_loop().set_default_executor(
-            concurrent.futures.ThreadPoolExecutor(max_workers=self.max_asyncio_workers))
-
         self._http_server = tornado.httpserver.HTTPServer(
             self.create_application(), max_buffer_size=self.max_buffer_size)
 
@@ -128,6 +124,10 @@ class ModelServer:
         self._http_server.bind(self.http_port)
         logging.info("Will fork %d workers", self.workers)
         self._http_server.start(self.workers)
+
+        logging.info(f"Setting max asyncio worker threads as {self.max_asyncio_workers}")
+        asyncio.get_event_loop().set_default_executor(
+            concurrent.futures.ThreadPoolExecutor(max_workers=self.max_asyncio_workers))
 
         # Need to start the IOLoop after workers have been started
         # https://github.com/tornadoweb/tornado/issues/2426


python/kserve/kserve/model_server.py
```diff
@@ -117,9 +117,10 @@ def start(self, models: Union[List[Model], Dict[str, Deployment]], nest_asyncio:
             # formula as suggest in https://bugs.python.org/issue35279
             self.max_asyncio_workers = min(32, utils.cpu_count()+4)
 
-        logging.info(f"Setting asyncio max_workers as {self.max_asyncio_workers}")
-        asyncio.get_event_loop().set_default_executor(
-            concurrent.futures.ThreadPoolExecutor(max_workers=self.max_asyncio_workers))
+        if self.workers == 1:
```
yuzisun: @Suresh-Nakkeran just saw this [comment](https://github.com/tornadoweb/tornado/issues/2426#issuecomment-400895086), can you try removing worker check and moving the asyncio initialization code after the "fork"(line131) see if multi-processing works .

Suresh-Nakkeran: you're right @yuzisun . multi-processing is working by moving initialization code after starting http server.
image

python/kserve/kserve/model_server.py
```diff
@@ -129,6 +125,10 @@ def start(self, models: Union[List[Model], Dict[str, Deployment]], nest_asyncio:
         logging.info("Will fork %d workers", self.workers)
         self._http_server.start(self.workers)
 
+        logging.info(f"Setting max asyncio worker threads as {self.max_asyncio_workers}")
+        asyncio.get_event_loop().set_default_executor(
+            concurrent.futures.ThreadPoolExecutor(max_workers=self.max_asyncio_workers))
```
yuzisun, line 130: @sukumargaonkar Is it ok to move this call after starting the http server?

PR #1996, opened 2022-01-17
Never ever use the default serviceaccount, but kserve-controller-manager

juliusvonkohout:
This is a simple security issue. Someone abused the default service account instead of using the proper kserve-controller-manager service account. This is very ugly and a security risks if there are any other pods , because they use the default service account by default.

**Release note**:

```release-note
NONE
```


yuzisun:
Thanks @juliusvonkohout !

/lgtm
/approve

PR #1997, opened 2022-01-18
Fix security issues in kserve manifest and update sklearn example storage uri

### LINKED ISSUE #1995: Leftover in the manifests
/kind bug

There is a service https://github.com/kserve/kserve/blob/2208f100af02c05031e1910516f31ae3ece87aaf/manifests/charts/templates/service.yaml#L26 that lists port 443 for the kserve-controller-manager, but according to the deployment nothing at 443 is served https://github.com/kserve/kserve/blob/2208f100af02c05031e1910516f31ae3ece87aaf/manifests/charts/templates/deployment.yaml. This would also be a severe security issue because 443 prevents run as non root pods.

yuzisun:

Fixes #1995 

1. Use non default service account for kserve controller
2. Add security context for kserve controller
3. Fix metric service

**Release note**:

```release-note

```


juliusvonkohout:
@yuzisun amazing response time and thank you for your effort.

PR #1998, opened 2022-01-18
Check deployment mode in isvc webhook

pvaneck:
InferenceServices with deployment mode of ModelMesh should not have the runtime assigning mutator run on it. This causes
an sms runtime to be assigned to the predictor when it shouldn't be if targeting deployment on modelmesh.


chinhuang007 (approved):
Looks good! Thanks for the fix.

PR #1999, opened 2022-01-19
Add autoSelect labels to more ServingRuntime formats

pvaneck:
Some formats like onnx did not have an autoSelect-able runtime assigned, so using an InferenceService with `modelFormat: onnx` would show `no runtime found to support predictor with model type...`. This enables `autoSelect` on formats I thought needed them.

### DIFF
--- a/config/runtimes/kserve-tritonserver.yaml
+++ b/config/runtimes/kserve-tritonserver.yaml
@@ -12,10 +12,12 @@ spec:
       version: "2"
     - name: onnx
       version: "1"
+      autoSelect: true
     - name: pytorch
       version: "1"
     - name: triton
       version: "2"
+      autoSelect: true
   containers:
     - name: kserve-container
       image: kserve-tritonserver:replace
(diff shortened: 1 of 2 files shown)


njhill (approved):

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -12,10 +12,12 @@ spec:
       version: "2"
     - name: onnx
       version: "1"
+      autoSelect: true
     - name: pytorch
       version: "1"
     - name: triton
       version: "2"
+      autoSelect: true
```
yuzisun, line 20: Technically triton isn’t a model format, I guess it is here for backwards compatibility?

pvaneck: Right, in case someone does `modelFormat.name: triton`.  I'm okay with removing this `triton` entry, but i think `triton` still needs to be in this list to support when the old triton key is specified in the predictor.

Issue #1965, opened 2021-12-28
Chore: Cleanup `InferenceService` configmap for ML framework related fields


yuzisun:
/kind feature

**Describe the solution you'd like**
Cleanup `InferenceService` configmap for framework related fields which are now replaced by `ClusterServingRuntime`.

e.g
```
"image": "tensorflow/serving",
"defaultImageVersion": "2.6.2",
"defaultGpuImageVersion": "2.6.2-gpu",
"supportedFrameworks": [
     "tensorflow"
],
```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #2000, opened 2022-01-19
multi-model serving with servingruntime

Suresh-Nakkeran:

#1965

### DIFF
--- a/config/runtimes/kserve-mlserver.yaml
+++ b/config/runtimes/kserve-mlserver.yaml
@@ -19,10 +19,6 @@ spec:
       env:
         - name: "MLSERVER_MODEL_IMPLEMENTATION"
           value: "{{.Labels.modelClass}}"
-        - name: "MLSERVER_MODEL_NAME"
-          value: "{{.Name}}"
-        - name: "MLSERVER_MODEL_URI"
-          value: "/mnt/models"
         - name: "MLSERVER_HTTP_PORT"
           value: "8080"
         - name: "MLSERVER_GRPC_PORT"
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -100,6 +100,12 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return fmt.Errorf("specified runtime %s does not support specified framework/version", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
+			// TODO: adding mlserver specific environment variables, will be removed after mlserver issue fix
+			// Issue: https://github.com/SeldonIO/MLServer/issues/465
+			if *isvc.Spec.Predictor.Model.Runtime == "kserve-mlserver" && isvc.Spec.Predictor.Model.GetStorageUri() != nil {
+				addMlServerEnvs(isvc)
+			}
+
 			sRuntime = *r
 		} else {
 			runtimes, err := isvc.Spec.Predictor.Model.GetSupportingRuntimes(p.client, isvc.Namespace, false)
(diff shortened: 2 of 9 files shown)


yuzisun:
@Suresh-Nakkeran Let's chat about this, the `multipleModel` flag is for Modelmesh

config/runtimes/kserve-mlserver.yaml
```diff
@@ -19,10 +19,6 @@ spec:
       env:
         - name: "MLSERVER_MODEL_IMPLEMENTATION"
           value: "{{.Labels.modelClass}}"
-        - name: "MLSERVER_MODEL_NAME"
-          value: "{{.Name}}"
-        - name: "MLSERVER_MODEL_URI"
```
yuzisun, line 24: `MLServerModelsDirEnv` should be there by default, see [here](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/predictor_sklearn.go#L127)

test/e2e/predictor/test_sklearn.py
```diff
@@ -139,6 +140,8 @@ def test_sklearn_v2_runtime_kserve():
             runtime="kserve-mlserver",
             storage_uri="gs://seldon-models/sklearn/mms/lr_model",
             protocol_version="v2",
+            env=[V1EnvVar(name="MLSERVER_MODEL_NAME", value="isvc-sklearn-v2-runtime"),
```
yuzisun: I thought these are not needed on spec as you inject in the mutator.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -48,12 +48,8 @@ func IsMemoryResourceAvailable(isvc *v1beta1api.InferenceService, totalReqMemory
 
 	container := isvc.Spec.Predictor.GetImplementation().GetContainer(isvc.ObjectMeta, isvc.Spec.Predictor.GetExtensions(), isvcConfig)
 
-	if constants.InferenceServiceContainerName == container.Name {
-		predictorMemoryLimit := container.Resources.Limits.Memory()
-		return predictorMemoryLimit.Cmp(totalReqMemory) >= 0
-	}
-
-	return false
+	predictorMemoryLimit := container.Resources.Limits.Memory()
```
yuzisun, line 51: Any reason to remove the check here?

Suresh-Nakkeran: @yuzisun In ServingRuntime implementation, the container name would be nil since it is set while resolving runtime spec.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -100,6 +100,20 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return fmt.Errorf("specified runtime %s does not support specified framework/version", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
+			// TODO: adding mlserver specific environment variables, will be removed after mlserver issue fix
+			if *isvc.Spec.Predictor.Model.Runtime == "kserve-mlserver" && isvc.Spec.Predictor.Model.GetStorageUri() != nil {
+				isvc.Spec.Predictor.Model.Env = append(isvc.Spec.Predictor.Model.Env,
+					v1.EnvVar{
+						Name:  constants.MLServerModelNameEnv,
+						Value: isvc.Name,
+					},
```
yuzisun: I think the `MLServerModelNameEnv` should be still in the mutator and only add if user does not specify the env variable, so later we can simply remove this block

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -100,6 +100,20 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return fmt.Errorf("specified runtime %s does not support specified framework/version", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
+			// TODO: adding mlserver specific environment variables, will be removed after mlserver issue fix
+			if *isvc.Spec.Predictor.Model.Runtime == "kserve-mlserver" && isvc.Spec.Predictor.Model.GetStorageUri() != nil {
```
yuzisun, line 104: add a nil check for `runtime` to be safe in case nothing is matched.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -100,6 +100,20 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return fmt.Errorf("specified runtime %s does not support specified framework/version", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
+			// TODO: adding mlserver specific environment variables, will be removed after mlserver issue fix
```
yuzisun: Let's create an issue on MLServer and link it here

PR #2004, opened 2022-01-23
support custom lib for sklearn model

hwk42:
in some scenarios,we need add custom logic like transformer when training a model,but joblib/pickle not support to persist the module ,which run into this 'No module named xxx '


yuzisun:
I think the better way is to build a custom image with these dependencies based on the sklearnserver and the images are scanned for security vulnerability, allow running some arbitrary python code here is a bit risky.

PR #2005, opened 2022-01-23
Use non root user for python server images

### LINKED ISSUE #1982: SecurityContext in inferenceservice.yaml?
/kind bug

Deployed KServe in an enterprise environment with strict security constraints. Tried to create an Inference Service directly from Jupyter Notebooks. The service is initializing but it fails because the container (storage initializer) tries to run as root.

image

Create the service and get predictions.

Is it possible to apply a SecurityContext to each docker image mentioned in the inferenceservice.yaml config map?

**Environment:**

- Istio Version: 1.9.6
- Knative Version: 0.22.1
- KFServing Version: 0.6.1
- Kubeflow version: 1.7.0
- Enterprise cluster

yuzisun:

Fixes #1982 

1. Use first non root user in python server images

**Release note**:

```release-note

```

### DIFF
--- a/python/aiffairness.Dockerfile
+++ b/python/aiffairness.Dockerfile
@@ -1,6 +1,11 @@
 FROM python:3.7
 
-COPY . .
+COPY kserve kserve
+COPY aiffairness aiffairness
+COPY third_party third_party
+ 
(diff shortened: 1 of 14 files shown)


pvaneck (commented):
Thanks, @yuzisun.

/lgtm

PR #2049, opened 2022-02-18
Add ingress class name configuration

### LINKED ISSUE #2048: Add ingressClassName configuration
/kind feature

When using RawDeployment and there are several ingress controller in the cluster, kserve administrator should be able to configure which ingress to use. Currently it's only possible by adding  `kubernetes.io/ingress.class` annotation to all inference service. However, users might miss adding this annotation which leads to inference service not working as expected or creating external ingress (it happens on GKE)
The annotation is also being [deprecated by Kubernetes](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation). 

The better approach is to add  ingress class name in the ingress config and use it to set  `ingressClassName` field during ingress specification creation.

pradithya:
To allow selecting specific ingress controller when using RawDeployment

Fixes #2048

**Type of changes**

- [x] Run the controller with and without ingress class name in cluster and create inference service

- Logs
Ingress generated when ingress class name is set to "istio"
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
    queue.sidecar.serving.knative.dev/resourcePercentage: "20"
  creationTimestamp: "2022-02-18T09:48:52Z"
  generation: 1
  name: xgboost-sample-413
  namespace: test
  resourceVersion: "10446008"
  uid: 79862c9c-e97f-4da9-9d58-e39a0b5481b5
spec:
  ingressClassName: istio
  rules:
  - host: xgboost-sample-413-test.example.com
    http:
      paths:
      - backend:
          service:
            name: xgboost-sample-413-predictor-default
            port:
              number: 80
        path: /
        pathType: Prefix
  - host: xgboost-sample-413-predictor-default-test.example.com
    http:
      paths:
      - backend:
          service:
            name: xgboost-sample-413-predictor-default
            port:
              number: 80
        path: /
        pathType: Prefix
```

**Checklist**:

**Release note**:

```release-note
Add `ingressClassName` configuration to allow selecting specific ingress controller for RawDeployment
```

### DIFF
--- a/pkg/apis/serving/v1beta1/configmap.go
+++ b/pkg/apis/serving/v1beta1/configmap.go
@@ -114,11 +114,12 @@ type InferenceServicesConfig struct {
 
 // +kubebuilder:object:generate=false
 type IngressConfig struct {
-	IngressGateway          string `json:"ingressGateway,omitempty"`
-	IngressServiceName      string `json:"ingressService,omitempty"`
-	LocalGateway            string `json:"localGateway,omitempty"`
-	LocalGatewayServiceName string `json:"localGatewayService,omitempty"`
-	IngressDomain           string `json:"ingressDomain,omitempty"`
+	IngressGateway          string  `json:"ingressGateway,omitempty"`
+	IngressServiceName      string  `json:"ingressService,omitempty"`
+	LocalGateway            string  `json:"localGateway,omitempty"`
+	LocalGatewayServiceName string  `json:"localGatewayService,omitempty"`
+	IngressDomain           string  `json:"ingressDomain,omitempty"`
+	IngressClassName        *string `json:"ingressClassName,omitempty"`
 }
 
 // +kubebuilder:object:generate=false
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
@@ -190,7 +190,8 @@ func createRawIngress(scheme *runtime.Scheme, isvc *v1beta1api.InferenceService,
 			Annotations: isvc.Annotations,
 		},
 		Spec: netv1.IngressSpec{
-			Rules: rules,
+			IngressClassName: ingressConfig.IngressClassName,
+			Rules:            rules,
 		},
 	}
 	if err := controllerutil.SetControllerReference(isvc, ingress, scheme); err != nil {
--- a/test/e2e/predictor/test_raw_deployment.py
+++ b/test/e2e/predictor/test_raw_deployment.py
@@ -35,7 +35,6 @@ def test_raw_deployment_kserve():
     service_name = "raw-sklearn"
     annotations = dict()
     annotations['serving.kserve.io/deploymentMode'] = 'RawDeployment'
-    annotations['kubernetes.io/ingress.class'] = 'istio'
 
     predictor = V1beta1PredictorSpec(
         min_replicas=1,
@@ -70,7 +69,6 @@ def test_raw_deployment_runtime_kserve():
     service_name = "raw-sklearn-runtime"
     annotations = dict()
     annotations['serving.kserve.io/deploymentMode'] = 'RawDeployment'
-    annotations['kubernetes.io/ingress.class'] = 'istio'
 
     predictor = V1beta1PredictorSpec(
         min_replicas=1,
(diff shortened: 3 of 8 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
```diff
@@ -190,7 +190,8 @@ func createRawIngress(scheme *runtime.Scheme, isvc *v1beta1api.InferenceService,
 			Annotations: isvc.Annotations,
 		},
 		Spec: netv1.IngressSpec{
-			Rules: rules,
+			IngressClassName: ingressConfig.IngressClassName,
```
yuzisun, line 193: what happens when `IngressClassName` is nil?

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -114,11 +114,12 @@ type InferenceServicesConfig struct {
...
-	IngressServiceName      string `json:"ingressService,omitempty"`
-	LocalGateway            string `json:"localGateway,omitempty"`
-	LocalGatewayServiceName string `json:"localGatewayService,omitempty"`
-	IngressDomain           string `json:"ingressDomain,omitempty"`
+	IngressGateway          string  `json:"ingressGateway,omitempty"`
+	IngressServiceName      string  `json:"ingressService,omitempty"`
+	LocalGateway            string  `json:"localGateway,omitempty"`
+	LocalGatewayServiceName string  `json:"localGatewayService,omitempty"`
+	IngressDomain           string  `json:"ingressDomain,omitempty"`
+	IngressClassName        *string `json:"ingressClassName,omitempty"`
```
yuzisun, line 122: We can default the `IngressClassName` to `istio` and add a comment that only for `RawDeployment` we allow other ingress option currently.

Let's set the default to the [config](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L161)  and the [test config](https://github.com/kserve/kserve/blob/master/config/overlays/test/configmap/inferenceservice.yaml#L161), add a test case without using the [ingress class annotation](https://github.com/kserve/kserve/blob/master/test/e2e/predictor/test_raw_deployment.py#L38).

yuzisun (commented):
Thanks @pradithya for the contribution!

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
pradithya: If it's `nil` all IngressClassName will be empty and will use default ingress in the cluster (https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class) . In GKE case it will use create Google Ingress. But essentially it will behave the same as existing behaviour. 

If inference service owner want to use specific ingress then they'll have to add ingress class annotation (kubernetes.io/ingress.class)

pkg/apis/serving/v1beta1/configmap.go
pradithya: I updated `istio` as default ingress class name. We'll need to add the istio `ingressClass` creation (https://istio.io/latest/docs/tasks/traffic-management/ingress/kubernetes-ingress/#specifying-ingressclass) in the documentation since it's not created by default.

pradithya:
PR for doc update https://github.com/kserve/website/pull/113

pkg/apis/serving/v1beta1/configmap.go
pradithya: Updated the test too

pradithya:
The last e2e fails are unrelated to the PR
```
FAILED predictor/test_multi_model_serving.py::test_mms_sklearn_kserve[v2-gs://seldon-models/sklearn/mms/lr_model]
FAILED predictor/test_xgboost.py::test_xgboost_v2_runtime_kserve - requests.e...
```

pradithya:
@yuzisun Ready to be reviewed again

test/e2e/predictor/test_raw_deployment.py
```diff
@@ -35,7 +35,6 @@ def test_raw_deployment_kserve():
     service_name = "raw-sklearn"
     annotations = dict()
     annotations['serving.kserve.io/deploymentMode'] = 'RawDeployment'
-    annotations['kubernetes.io/ingress.class'] = 'istio'
```
pradithya, line 38: Removing this annotation from test as it causes some errors:
```
annotations.kubernetes.io/ingress.class: Invalid value: \"istio\": can not be set when the class field is also set
```

Somehow the cluster or istio rejects when both `ingressClassName` and `annotations.kubernetes.io/ingress.class` are being set.

yuzisun:
Thanks @pradithya !

/lgtm
/approve

Issue #2065, opened 2022-02-26
Support auto selection based on protocol version


yuzisun:
/kind feature

**Describe the solution you'd like**
Currently for example `sklearn` model, user has to explicitly specify the serving runtime name to choose the runtime that can support v2 protocol because `kserve-sklearnserver` is marked for auto selection.

**Anything else you would like to add:**
One option is to add `protocolVersion` to the `SupportedModelFormat`

```
// +k8s:openapi-gen=true
type SupportedModelFormat struct {
	// Name of the model format.
	// +required
	Name string `json:"name"`
	// Version of the model format.
	// Used in validating that a predictor is supported by a runtime.
	// Can be "major", "major.minor" or "major.minor.patch".
	// +optional
	Version *string `json:"version,omitempty"`
        // Supported protocol version
        // v1
        // grpc-v1
        // grpc-v2
        // v2
        ProtocolVersion string `json:"protocolVersion,omitempty"`
	// Set to true to allow the ServingRuntime to be used for automatic model placement if
	// this model format is specified with no explicit runtime.
	// +optional
	AutoSelect *bool `json:"autoSelect,omitempty"`
}
```

njhill:
This was implemented in the main kserve controller in https://github.com/kserve/kserve/pull/2118.

PR #2070, opened 2022-03-02
Added protocolversion support in servingruntime

Suresh-Nakkeran:
Issue #2065 

This PR is to add support for protocolversion in servingruntime.

### DIFF
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -6,9 +6,11 @@ spec:
   supportedModelFormats:
     - name: tensorflow
       version: "1"
+      protocolVersion: v1
       autoSelect: true
     - name: tensorflow
       version: "2"
+      protocolVersion: v1
       autoSelect: true
   containers:
     - name: kserve-container
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -6,6 +6,12 @@ spec:
   supportedModelFormats:
     - name: tensorflow
       version: "1"
+      autoSelect: true
+    - name: tensorflow
+      version: "2"
+      autoSelect: true
+    - name: tensorflow
+      version: "1"
       protocolVersion: v1
       autoSelect: true
     - name: tensorflow
--- a/pkg/apis/serving/v1beta1/inference_service_defaults.go
+++ b/pkg/apis/serving/v1beta1/inference_service_defaults.go
@@ -250,6 +250,21 @@ func (isvc *InferenceService) assignPaddleRuntime() {
 	isvc.Spec.Predictor.Paddle = nil
 }
 
+func (isvc *InferenceService) SetRuntimeDefaults() {
+	// add mlserver specific default values
+	if *isvc.Spec.Predictor.Model.Runtime == constants.MLServer {
+		isvc.SetMlServerDefaults()
+	}
+	// add torchserve specific default values
+	if *isvc.Spec.Predictor.Model.Runtime == constants.TorchServe {
+		isvc.SetTorchServeDefaults()
+	}
+	// add triton specific default values
+	if *isvc.Spec.Predictor.Model.Runtime == constants.TritonServer {
+		isvc.SetTritonDefaults()
+	}
+}
+
 func (isvc *InferenceService) SetMlServerDefaults() {
 	// set environment variables based on storage uri
 	if isvc.Spec.Predictor.Model.StorageURI == nil {
@@ -301,6 +316,11 @@ func (isvc *InferenceService) SetTorchServeDefaults() {
 }
 
 func (isvc *InferenceService) SetTritonDefaults() {
+	// set 'v2' as default protocol version for triton server
+	if isvc.Spec.Predictor.Model.ProtocolVersion == nil {
+		protocolV2 := constants.ProtocolV2
+		isvc.Spec.Predictor.Model.ProtocolVersion = &protocolV2
+	}
 	// set model-control-model arg to 'explicit' if storage uri is nil
 	if isvc.Spec.Predictor.Model.StorageURI == nil {
 		isvc.Spec.Predictor.Model.Args = append(isvc.Spec.Predictor.Model.Args,
--- a/pkg/apis/serving/v1beta1/inference_service_defaults.go
+++ b/pkg/apis/serving/v1beta1/inference_service_defaults.go
@@ -134,15 +134,9 @@ func (isvc *InferenceService) assignSKLearnRuntime() {
 		defaultProtocol := constants.ProtocolV1
 		isvc.Spec.Predictor.SKLearn.ProtocolVersion = &defaultProtocol
 	}
-	var runtime = constants.SKLearnServer
-	if isvc.Spec.Predictor.SKLearn.ProtocolVersion != nil &&
-		constants.ProtocolV2 == *isvc.Spec.Predictor.SKLearn.ProtocolVersion {
-		runtime = constants.MLServer
-	}
 	isvc.Spec.Predictor.Model = &ModelSpec{
 		ModelFormat:            ModelFormat{Name: constants.SupportedModelSKLearn},
 		PredictorExtensionSpec: isvc.Spec.Predictor.SKLearn.PredictorExtensionSpec,
-		Runtime:                &runtime,
 	}
 	// remove sklearn spec
 	isvc.Spec.Predictor.SKLearn = nil
@@ -154,11 +148,9 @@ func (isvc *InferenceService) assignTensorflowRuntime() {
 		defaultProtocol := constants.ProtocolV1
 		isvc.Spec.Predictor.Tensorflow.ProtocolVersion = &defaultProtocol
 	}
-	var runtime = constants.TFServing
 	isvc.Spec.Predictor.Model = &ModelSpec{
 		ModelFormat:            ModelFormat{Name: constants.SupportedModelTensorflow},
 		PredictorExtensionSpec: isvc.Spec.Predictor.Tensorflow.PredictorExtensionSpec,
-		Runtime:                &runtime,
 	}
 	// remove tensorflow spec
 	isvc.Spec.Predictor.Tensorflow = nil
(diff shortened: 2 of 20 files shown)


config/runtimes/kserve-tensorflow-serving.yaml
```diff
@@ -6,9 +6,19 @@ spec:
...
       autoSelect: true
     - name: tensorflow
       version: "2"
+      protocolVersion: v1
+      autoSelect: true
+    - name: tensorflow
+      version: "1"
+      protocolVersion: v2
+      autoSelect: true
+    - name: tensorflow
```
yuzisun, line 17: Tfserving runtime does not support v2 protocol.

test/e2e/predictor/test_triton.py
```diff
@@ -32,6 +32,7 @@ def test_triton():
         min_replicas=1,
         triton=V1beta1TritonSpec(
             storage_uri='gs://kfserving-examples/models/torchscript',
+            protocol_version="v2",
```
yuzisun: - To maintain backwards compatibility we can default the protocol version to v2 if the runtime is set to triton. 
- if protocol version is specified while runtime is not, we can do the auto selection based on protocol and model format, e.g for tensorflow if v1 protocol is specified it assigns the tfserving runtime, when v2 is specified it assigns the triton runtime.

config/runtimes/kserve-tensorflow-serving.yaml
Suresh-Nakkeran: @yuzisun updated!

test/e2e/predictor/test_triton.py
Suresh-Nakkeran: @yuzisun updated!

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -152,32 +138,6 @@ func (isvc *InferenceService) assignSKLearnRuntime() {
 	if isvc.Spec.Predictor.SKLearn.ProtocolVersion != nil &&
 		constants.ProtocolV2 == *isvc.Spec.Predictor.SKLearn.ProtocolVersion {
 		runtime = constants.MLServer
```
yuzisun: Is this check still needed as we do auto selection based on `protocolVersion` now?

yuzisun:
@pvaneck Can you help review this as well?

test/e2e/predictor/test_triton.py
```diff
@@ -85,6 +85,7 @@ def test_triton_runtime():
                 name="pytorch",
             ),
             runtime="kserve-tritonserver",
+            protocol_version="v2",
```
yuzisun: since tritonserver runtime is specified, do you still need to set the `protocol_version`?

Suresh-Nakkeran: yes @yuzisun.. even though we provide runtime, we validate model format against runtime's supported model format.. the protocol version should be added in order to satisfy.. At the end of validation, we are adding defaults..

pkg/constants/constants.go
```diff
@@ -191,8 +191,10 @@ const (
 
 // InferenceService protocol enums
 const (
-	ProtocolV1 InferenceServiceProtocol = "v1"
-	ProtocolV2 InferenceServiceProtocol = "v2"
+	ProtocolV1     InferenceServiceProtocol = "v1"
+	ProtocolV2     InferenceServiceProtocol = "v2"
+	ProtocolGRPCV1 InferenceServiceProtocol = "grpc-v1"
+	ProtocolGRPCV2 InferenceServiceProtocol = "grpc-v2"
```
pvaneck: Do we need these grpc-specific protocols or is just v1 and v2 enough? I don't see them being used.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -113,7 +114,23 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+		if *isvc.Spec.Predictor.Model.Runtime == constants.MLServer {
+			isvc.SetMlServerDefaults()
+		}
+		// add torchserve specific default values
+		if *isvc.Spec.Predictor.Model.Runtime == constants.TorchServe {
+			isvc.SetTorchServeDefaults()
+		}
+		// add triton specific default values
+		if *isvc.Spec.Predictor.Model.Runtime == constants.TritonServer {
+			isvc.SetTritonDefaults()
```
pvaneck: Should we be setting the defaults for these runtimes in all cases? I know previously these were for maintaining backwards compatibility with the old schema, but I'm not sure about adding things to the container spec that aren't defined in the runtime CR itself.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -113,7 +114,23 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return fmt.Errorf("no runtime found to support predictor with model type: %v", isvc.Spec.Predictor.Model.ModelFormat)
 			}
 			// Get first supporting runtime.
-			sRuntime = runtimes[0]
+			for rtName, rtSpec := range runtimes {
+				sRuntime = rtSpec
+				isvc.Spec.Predictor.Model.Runtime = proto.String(rtName)
```
pvaneck: Perhaps just use `&rtName` instead of `proto.String`?

test/e2e/predictor/test_triton.py
pvaneck: Hmm, I've been leaning more and more towards removing the validation when a user explicitly specifies a runtime name. Just trust that the user is deploying a compatible model and proceed with the deployment. WDYT?

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -155,9 +159,9 @@ func (m *ModelSpec) getServingRuntimeSupportedModelFormatLabelSet(supportedModel
 	for _, t := range supportedModelFormats {
 		// If runtime isn't explicitly set, only add labels for modelFormats where AutoSelect is true.
 		if m.Runtime != nil || (t.AutoSelect != nil && *t.AutoSelect) {
-			set.add("mt:" + t.Name)
```
pvaneck, line 158: What should be the behavior if there is an entry in a ServingRuntimes supportedModelFormat list that doesn't specify protocolVersion? Can we have an Isvc match a runtime with such an entry? Since right now there is always a protocolversion in the modelFormatLabel

pvaneck (commented):
Thanks for implementing protocolVersion support. I left some comments inline.

pkg/apis/serving/v1beta1/predictor_model.go
Suresh-Nakkeran: @pvaneck ProtocolVersion should be present in all entries of supportedModelFormat in ServingRuntimes. The default ProtocolVersion would be added with ISVC config if the user is not specified.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: @pvaneck Previously, Mlserver and Triton servers do not have a supportedModelFormat with autoSelect, If user provide runtime, the defaults are applied in mutator itself. With this changes, those servers also can be chosen based on ProtocolVersion. So we are adding defaults once the runtime has been selected.

pkg/constants/constants.go
Suresh-Nakkeran: @pvaneck I hope, trochserve is going to implement grpc based v1, v2 endpoints. In order to differentiate added 2 more entries in ProtocolVersion.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: updated. Thanks!

pkg/constants/constants.go
yuzisun: I think the v1/v2 here really include both REST and gRPC protocol, the only issue is that torchserve has not implemented v2 gRPC yet.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -210,90 +162,47 @@ func (isvc *InferenceService) assignXGBoostRuntime() {
...
 	isvc.Spec.Predictor.Model = &ModelSpec{
 		ModelFormat:            ModelFormat{Name: constants.SupportedModelXGBoost},
 		PredictorExtensionSpec: isvc.Spec.Predictor.XGBoost.PredictorExtensionSpec,
-		Runtime:                &runtime,
 	}
 	// remove xgboost spec
 	isvc.Spec.Predictor.XGBoost = nil
 }
 
 func (isvc *InferenceService) assignPyTorchRuntime() {
```
yuzisun, line 173: Do we need to default the protocolVersion for pytorch here?

pkg/apis/serving/v1beta1/predictor_model.go
yuzisun: I think there could be use case where a custom serving runtime is not following v1/v2 protocol, we might need to account for this.

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -141,11 +141,15 @@ func (m *ModelSpec) RuntimeSupportsModel(srSpec *v1alpha1.ServingRuntimeSpec) bo
 }
 
 func (m *ModelSpec) getModelFormatLabel() string {
+	pv := constants.ProtocolV1
```
yuzisun: maybe default to ProtocolUnknown if protocol version is not specified

pkg/constants/constants.go
Suresh-Nakkeran: @yuzisun @pvaneck shall I remove "grpc-v1" and "grpc-v2" entries ?

pkg/apis/serving/v1beta1/predictor_model.go
njhill: Yes we definitely need to support this case.

njhill:
@yuzisun @Suresh-Nakkeran @pvaneck Hey sorry for being so late to this one, I completely missed it before.

I'm not sure that protocolVersion belongs in the supported model format struct since that pertains to the serialized form of the model to load. It's really orthogonal to that and related to the API that's exposed, i.e. it doesn't make sense to me for it to be different per model.

Wouldn't it make more sense as a separate top-level field in the ServingRuntime? e.g. something like `supportedProtocolVersions` with a list of strings? We already reflect whether a particular runtime exposes grpc and/or REST via the `grpcDataEndpoint` and `httpDataEndpoint` fields so I'm not sure we need separate protocol versions for those.

In either case it may take some more thought as to how we can generalize the modelmesh side to support this too.

We could also consider automatic transcoding between v1 and v2 so that a predictor specifying one can be supported by a runtime that only supports the other (not for this PR of course!)

pvaneck:
Thanks for your input, @njhill.

From what I gather, the main motivation for supporting protocolversion is so that InferenceServices like the following would automatically map to the MLServer runtime as opposed to the `kserve-skleanserver` runtime.

```yaml
kind: "InferenceService"
metadata:
  name: "sklearn-irisv2"
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      protocolVersion: v2
      storageUri: "gs://seldon-models/sklearn/mms/lr_model"
```

I believe this is mainly for backwards compatibility(?), so probably worth having the supported protocol versions be top-level as it reduces the number of combinations that need to be added to the supportedModelFormats list. This list looks to get be quite unwieldy if a runtime supports multiple formats along with multiple versions already.

At the top level, I think an *optional* `supportedProtocolVersions` list of strings makes sense. Doesn't seem like the protocol would be different for different formats in a serving runtime. In most cases, this list would only contain a single entry (either v1 or v2). But looks like in the torchserve case, it would have both v1 and v2.

When it comes to ServingRuntime selection for a particular model format, if `protocolVersion` is set in the predictor spec, select the runtime with a compatible supportedProtocolVersion and modelFormat with autoSelect as true. This means we'd have multiple `autoSelect: true` for sklearn (kserve-mlserver and kserver-sklearnserver). In order to maintain the backwards compatible behavior, if a `protocolVersion` is not specified, default to the ServingRuntime that supports `v1` as we do now (potentially v2 could be default as well). 

Current typical behavior could be maintained where a user doesn't specify protocolVersion in the predictor spec and the ServingRuntime doesn't have a `supportedProtocolVersions` list.

Does this seem sensible?

njhill:
Thanks @pvaneck, yes that sounds like what I had in mind. @yuzisun @Suresh-Nakkeran any thoughts?

yuzisun:
> Thanks @pvaneck, yes that sounds like what I had in mind. @yuzisun @Suresh-Nakkeran any thoughts?

I think originally @Suresh-Nakkeran attempted adding it to the top level, forgot the reason why we moved it down to `SupportedModelFormat`, suresh is going to revisit this and get back to us.

yuzisun:
> > Thanks @pvaneck, yes that sounds like what I had in mind. @yuzisun @Suresh-Nakkeran any thoughts?
> 
> I think originally @Suresh-Nakkeran attempted adding it to the top level, forgot the reason why we moved it down to `SupportedModelFormat`, suresh is going to revisit this and get back to us.

@Suresh-Nakkeran The reason to move down to `SupportedModelFormat` is that previously we only allow marking one of the ml framework to be auto selected, now in order to incorporate with criteria like protocol version at top level we will need to mark more ml frameworks to be auto selected. is that right ? If that's the case I think this might be desirable as we can get a list of candidates and select the top one with some ordering criteria.

njhill:
@Suresh-Nakkeran regarding how to select a runtime during reconcilation when more than one "match" with autoselect, another idea would be to order by the runtime CR's `Revision` rather than name. That should be a bit more "stable" and not cause InferenceServices to move between runtimes if for example a new runtime was created with autoselect=true.

yuzisun:
closing this and move over to https://github.com/kserve/kserve/pull/2118

PR #2072, opened 2022-03-03
fix: configmap url in sklearn example

ittus:

Fix error URL in doc

**Type of changes**

- Document

### DIFF
--- a/docs/samples/v1beta1/sklearn/v1/README.md
+++ b/docs/samples/v1beta1/sklearn/v1/README.md
@@ -93,7 +93,7 @@ you saved from your training environment, however you can build your own SKLearn
 ).
 
 To use your SKLearnServer image:
-- Add the image to the KServe [configmap](../../../config/configmap/inferenceservice.yaml)
+- Add the image to the KServe [configmap](../../../../../config/configmap/inferenceservice.yaml)
 ```yaml
         "sklearn": {
             "image": "<your-dockerhub-id>/kfserving/sklearnserver",
--- a/docs/samples/v1beta1/sklearn/v1/README.md
+++ b/docs/samples/v1beta1/sklearn/v1/README.md
@@ -93,7 +93,7 @@ you saved from your training environment, however you can build your own SKLearn
 ).
 
 To use your SKLearnServer image:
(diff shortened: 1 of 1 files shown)


docs/samples/v1beta1/sklearn/v1/README.md
```diff
@@ -93,7 +93,7 @@ you saved from your training environment, however you can build your own SKLearn
 ).
 
 To use your SKLearnServer image:
-- Add the image to the KServe [configmap](../../../config/configmap/inferenceservice.yaml)
+- Add the image to the KServe [configmap](../../../../../config/configmap/inferenceservice.yaml)
```
yuzisun: Maybe better to just change to absolute path `https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml`

ittus: @yuzisun  Thanks for your comment. Can you explain more the benefit of absolute path vs relative path?
I think with relative path, when we change the repo's name or move to a different account, we don't need to change the path.

pvaneck: I suppose if we were to add this to the website repo, it would make sense to use absolute paths. @yuzisun, I noticed that this doc [isn't on the website](https://github.com/kserve/website/tree/main/docs/modelserving/v1beta1/sklearn), should this be updated and added there, or was the omission intentional?

yuzisun: I think it might be still worth adding v1 example as a lot of users found it is much easier for sklearn models, originally I thought we are going to align with v2 protocol.

ittus: @yuzisun 
Thank you. I changed to absolute path.

yuzisun:
@ittus Can you help sign the commits following https://github.com/kserve/kserve/pull/2072/checks?check_run_id=5623982399 ?

ittus:
@yuzisun Thank you. I signed the commit.

yuzisun:
thanks @ittus !
/lgtm
/approve

PR #2074, opened 2022-03-04
fix: replace image tag

### LINKED ISSUE #2068: Cannot use colon with proxy port in predictor image name in inferenceservice-config
We changed the configuration of the predictor images in the ```inferenceservice-config``` ConfigMap (https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml) to use a proxy and a proxy port.

So e.g. ```"image": "ourproxy:8888/tensorflow/serving"```
instead of  ```"image": "tensorflow/serving"```

However using a colon in the predictor image name doesn't work because of ```strings.Split(image, ":")[0]```
in https://github.com/kserve/kserve/blob/922a9b7e8a9a86b5ae65faf4ce863927873fd456/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L189

Can this please be changed to split the image name and the runtimeversion using the last colon in this string?
Or is there another way to configure a proxy for pulling these predictor images?

ittus:
Fixes #2068 

**Type of changes**

- Logs

**Checklist**:

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"html/template"
+	"regexp"
 	"strings"
 
 	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -186,7 +187,12 @@ func ReplacePlaceholders(container *v1.Container, meta metav1.ObjectMeta) error
 func UpdateImageTag(container *v1.Container, runtimeVersion *string, isvcConfig *v1beta1.InferenceServicesConfig) {
 	image := container.Image
 	if runtimeVersion != nil && len(strings.Split(image, ":")) > 0 {
```
yuzisun: I think now the `len(strings.Split(image, ":")) > 0` check is redundant and can be removed.

yuzisun:
@ittus Thanks for the fix! a small comment otherwise it looks good.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
ittus: Thanks. I removed it in https://github.com/kserve/kserve/pull/2074/commits/cc1397c3dc15990ef1c757513c42140ba32ac01e

yuzisun:
Thanks @ittus for your contribution!!

/lgtm
/approve

PR #2084, opened 2022-03-12
feat: Add ModelMesh related status fields to isvc

### LINKED ISSUE #2042: InferenceService Status restructuring
Currently, ModelMesh has its own set of status fields that it populates in the InferenceService status since we [allow unknown fields](https://github.com/kserve/kserve/pull/1855/files) there. Similarly, KServe has its own status fields that are predominantly based on knative fields. 

The goal here is to converge on a common set of status fields that would work with the various different deployment modes.

As an overview, here are the current status fields:

### [InferenceServiceStatus](https://github.com/kserve/kserve/blob/7347e2fa644f2fed842759644fa5a995ca42a756/pkg/apis/serving/v1beta1/inference_service_status.go#L28)
```
InferenceServiceStatus
  ├──── duckv1.Status `json:",inline"`
  │        │    // The 'Generation' of the Service that was last processed by the controller.
  │        ├─── ObservedGeneration int64 `json:"observedGeneration,omitempty"`
  │        │
  │        │    // Conditions the latest available observations of a resource's current state.
  │        |    // Condition defines a readiness condition for a Knative resource.
  │        ├─── Conditions Conditions `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
  │        | 
  │        |     // Additional Status fields for the Resource to save some additional State as well as convey more information to the user.
  │        ├─── Annotations map[string]string `json:"annotations,omitempty"`
  │
  ├──── Address *duckv1.Addressable `json:"address,omitempty"`
  ├──── URL *apis.URL `json:"url,omitempty"`
  ├──── Components map[ComponentType]ComponentStatusSpec `json:"components,omitempty"`
```
`ComponentTypes` are `predictor`, `explainer`, `transformer`. Which map to a [ComponentStatusSpec](https://github.com/kserve/kserve/blob/7347e2fa644f2fed842759644fa5a995ca42a756/pkg/apis/serving/v1beta1/inference_service_status.go#L48).

An example of an actual status can be found here: https://pastebin.com/wkCrZyxk

### [ModelMesh Predictor Status](https://github.com/kserve/modelmesh-serving/blob/5f0e5bbf900d18bf11dd90cc971c1eb7a319c61a/apis/serving/common/common_types.go#L113)

```
PredictorStatus
  ├──── Available bool `json:"available"`
  │
  │     // One of 'UpToDate', 'InProgress', 'BlockedByFailedLoad', or 'InvalidSpec'
  ├──── TransitionStatus TransitionStatus `json:"transitionStatus"
  │
  │     // High level state string: Pending, Standby, Loading, Loaded, FailedToLoad
  ├──── ActiveModelState ModelState `json:"activeModelState"
  ├──── TargetModelState ModelState `json:"targetModelState"`
  │
  │     // Details of last failure, when load of target model is failed or blocked
  ├──── LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
  │
  │     // Addressable endpoint for the deployed trained model. This will be "static" and will not change when the model is mutated
  ├──── HTTPEndpoint string `json:"httpEndpoint"`
  ├──── GrpcEndpoint string `json:"grpcEndpoint"`
  │
  │     // How many copies of this predictor's models failed to load recently
  ├──── FailedCopies int `json:"failedCopies"`
```
An example of an actual predictor status can be found here:  https://pastebin.com/wBM3WgFW

___

**Action item**: Decide how we can restructure the fields in a way that make sense for all the deployment modes. Do we still rely on the Knative types?

pvaneck:
ModelMesh needs specific fields to convey predictor status information. This commit introduces a new ModelStatus object to contain these fields.

Note: This is tentative for now and is more of an amalgamation of the ModelMesh predictor status spec with the InferenceService status spec as discussed in #2042. This will at least make the status more usable with ModelMesh without relying on unknown fields being preserved by Kubernetes (via `x-kubernetes-preserve-unknown-fields`).

ModelMesh does have the notion of the REST endpoint and gRPC endpoint being different, so a new `grpcURL` field was introduced to the component status field.

Open for discussion and iteration.

Closes: #2042 

**Release note**:
```release-note
Add additional model-related status fields to InferenceService CRD
```


pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+	TransitionStatus TransitionStatus `json:"transitionStatus"`
+
+	// High level state string: Pending, Standby, Loading, Loaded, FailedToLoad
+	// +kubebuilder:default=Pending
+	ActiveModelState ModelState `json:"activeModelState"`
+	// +kubebuilder:default=""
+	TargetModelState ModelState `json:"targetModelState"`
+
+	// Details of last failure, when load of target model is failed or blocked
+	//+optional
```
yuzisun:
```suggestion
	// +optional
```

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
 	IngressReady apis.ConditionType = "IngressReady"
 )
 
+type ModelStatus struct {
+	// Whether the available predictor endpoint reflects the current Spec or is in transition
+	// +kubebuilder:default=UpToDate
+	TransitionStatus TransitionStatus `json:"transitionStatus"`
```
yuzisun, line 121: @Suresh-Nakkeran I think for single model `TransitionStatus` can be set to following state:
- `UpToDate` when `latestCreatedRevision` == `lastestReadyRevision`
- `InProgress` when `latestCreatedRevision` is still in progress
- `BlockedByFailedLoad` when `latestCreatedRevision` fails

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+// TransitionStatus Enum values
+const (
+	// Predictor is up-to-date (reflects current spec)
+	UpToDate TransitionStatus = "UpToDate"
+	// Waiting for target model to reach state of active model
+	InProgress TransitionStatus = "InProgress"
+	// Target model failed to load
+	BlockedByFailedLoad TransitionStatus = "BlockedByFailedLoad"
+	// TBD
+	InvalidSpec TransitionStatus = "InvalidSpec"
```
yuzisun, line 166: What are the cases for invalid spec?

Suresh-Nakkeran: @yuzisun Can we update InvalidSpec status If NoMatching/Disabled/Invalid Runtime provided?

pvaneck: In MM, this is for when the predictor's currently active model configuration was not transitioned to match its current spec because the current spec is invalid. This would typically be coupled with an error message in the lastFailureInfo object.

pvaneck:
I see that @Suresh-Nakkeran has overlapping work done in #2088. I'll close this in favor of that PR.

yuzisun:
> I see that @Suresh-Nakkeran has overlapping work done in #2088. I'll close this in favor of that PR.

@pvaneck we can get your api in first and it is good to separate the two PRs, he was working on to populate the status for single model case.

pvaneck:
> @pvaneck we can get your api in first and it is good to separate the two PRs, he was working on to populate the status for single model case.

Ah, okay. Reopened and pushed some small adjustments. I noticed that `Available` was being used in Suresh's PR. Should we maintain that in ModelStatus? Since I know we also have the PredictoryReady condition

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+type ModelStatus struct {
+	// Whether the available predictor endpoint reflects the current Spec or is in transition
+	// +kubebuilder:default=UpToDate
+	TransitionStatus TransitionStatus `json:"transitionStatus"`
+
+	// High level state string: Pending, Standby, Loading, Loaded, FailedToLoad
+	// +kubebuilder:default=Pending
+	ActiveModelState ModelState `json:"activeModelState"`
+	// +kubebuilder:default=""
+	TargetModelState ModelState `json:"targetModelState"`
```
njhill: @Suresh-Nakkeran @yuzisun I wonder whether these could also be set to appropriate values based on the corresponding revision states. `TargetModelState` is only set to non-empty when the target model (revision) != active model (i.e. transition status != `UpToDate`)

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+
+	// Details of last failure, when load of target model is failed or blocked
+	// +optional
+	LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
```
njhill: Maybe even something could be added to this for when one of the revisions fails?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+	Location string `json:"location,omitempty"`
+	// High level class of failure
+	//+optional
+	Reason FailureReason `json:"reason,omitempty"`
+	// Detailed error message
+	//+optional
+	Message string `json:"message,omitempty"`
+	// Internal ID of model, tied to specific Spec contents
+	//+optional
+	ModelId string `json:"modelId,omitempty"`
```
njhill: This would be equivalent to revision name.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+	// +kubebuilder:default=""
+	TargetModelState ModelState `json:"targetModelState"`
+
+	// Details of last failure, when load of target model is failed or blocked
+	// +optional
+	LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
+
+	// How many copies of this predictor's models failed to load recently
+	// +kubebuilder:default=0
+	FailedCopies int `json:"failedCopies"`
```
njhill, line 147: The intention was to have additional fields here `loadedCopies` and `loadingCopies`. Unfortunately it's not easy to distinguish between loaded and loading because we don't record that transition explicitly in etcd - but it will become possible with some internal improvements to model-mesh that we'd like to do at some point.

It would probably be good however to add a `totalCopies` field, that one would be easy to populate and probably still useful to have in the status.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -65,6 +68,9 @@ type ComponentStatusSpec struct {
 	// It generally has the form http[s]://{route-name}.{route-namespace}.{cluster-level-suffix}
 	// +optional
 	URL *apis.URL `json:"url,omitempty"`
+	// gRPC endpoint of the component if different from URL.
+	// +optional
+	GrpcURL *apis.URL `json:"grpcUrl,omitempty"`
```
njhill, line 77: So the thinking is we only fill this in if different from `URL`?

This seems good but I guess what's missing is a way to tell which of REST/gRPC are exposed in all cases. I.e. when `GrpcURL` isn't present, does `URL` indicate REST or gRPC or both?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +111,96 @@ const (
...
+	TargetModelState ModelState `json:"targetModelState"`
+
+	// Details of last failure, when load of target model is failed or blocked
+	// +optional
+	LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
+
+	// How many copies of this predictor's models failed to load recently
+	// +kubebuilder:default=0
+	FailedCopies int `json:"failedCopies"`
+}
```
njhill, line 151: I found that the modelmesh predictors' status output from `kubectl describe` is kind of jumbled because there's no way to control the order of printed fields AFAIK.

So I was thinking that it would be good to group/nest some of the fields within other structs, specifically:
- Active and target model states
- Model copy counts assuming that we add another per the comment above

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -42,6 +43,8 @@ type InferenceServiceStatus struct {
 	URL *apis.URL `json:"url,omitempty"`
 	// Statuses for the components of the InferenceService
 	Components map[ComponentType]ComponentStatusSpec `json:"components,omitempty"`
+	// Model related statuses
+	ModelStatus ModelStatus `json:"modelStatus,omitempty"`
```
njhill, line 47: General thought regarding potential terminology confusion...

In modelmesh "model" generally refers a concrete/immutable model behind a particular predictor, kind of equivalent to a revision in knative. That's why we have individual model states within the struct, but at this level it should really be called `predictorStatus` I think? Thoughts?

njhill (commented):
Thanks alot for this @pvaneck.

pkg/apis/serving/v1beta1/inference_service_status.go
njhill: Maybe related, @yuzisun @Suresh-Nakkeran any thoughts on @pvaneck's suggestion here https://github.com/kserve/kserve/issues/2042#issuecomment-1053868785

pkg/apis/serving/v1beta1/inference_service_status.go
pvaneck: Sure, I think this would be useful. In SMS, this can correspond to the number of replicas, though would need to check how to populate this for both the Knative and raw deployment cases.

pkg/apis/serving/v1beta1/inference_service_status.go
pvaneck: Hmm, yea. I Haven't quite settled on how to handle this. Leaning towards my original approach of having both `HttpURL` and `GrpcURL`. And if they are the same, then it'll just be populated twice in both fields.

pkg/apis/serving/v1beta1/inference_service_status.go
pvaneck: Any recommendation for the json tag naming for these structs?

njhill: For the first one maybe `modelState` and second one `modelReplicas`?

pvaneck:
Grouped modelState and modelCopy fields into structs which should tidy up the prints when described. Also, an additional `restUrl` field was added to the ComponentStatusSpec.

pkg/apis/serving/v1beta1/inference_service_status.go
yuzisun: As discussed in the community meeting, we may want to differentiate model status and predictor service status as when a new model load fails the predictor service status can still be ready serving the previous model version.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +115,108 @@ const (
 	IngressReady apis.ConditionType = "IngressReady"
 )
 
+type ModelStatus struct {
+	// Whether the available predictor endpoints reflect the current Spec or is in transition
+	// +kubebuilder:default=UpToDate
+	TransitionStatus TransitionStatus `json:"transitionStatus"`
+
+	// State information of the predictor's model.
```
yuzisun, line 123: // +optional

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +115,108 @@ const (
...
+	ModelState *ModelStateContainer `json:"modelState,omitempty"`
+
+	// Details of last failure, when load of target model is failed or blocked
+	// +optional
+	LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
+
+	// Model replica information of the predictor's model.
```
yuzisun: // +optional

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +115,108 @@ const (
...
+	ActiveModelState ModelState `json:"activeModelState"`
+	// +kubebuilder:default=""
+	TargetModelState ModelState `json:"targetModelState"`
+}
+
+type ModelCopies struct {
+	// How many copies of this predictor's models failed to load recently
+	// +kubebuilder:default=0
+	FailedCopies int `json:"failedCopies"`
+	// Total number copies of this predictor's models that are currently loaded
```
yuzisun, line 148: // +optional

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -105,6 +115,108 @@ const (
...
+
+	// Details of last failure, when load of target model is failed or blocked
+	// +optional
+	LastFailureInfo *FailureInfo `json:"lastFailureInfo,omitempty"`
+
+	// Model replica information of the predictor's model.
+	ModelCopies *ModelCopies `json:"modelCopies,omitempty"`
+}
+
+type ModelStateContainer struct {
```
yuzisun: should `TransitionStatus` and `lastFailureInfo` be inside `ModelStateContainer`? seems like they are all related to the model state.

pkg/apis/serving/v1beta1/inference_service_status.go
yuzisun: shall we call it `ModelRevisionName` ?

pkg/apis/serving/v1beta1/inference_service_status.go
pvaneck: Potentially, although the intent of the ModelStateContainer is to specifically keep the state fields (i.e. `activeModelState` and `targetModelState`)  next to each other if a user describes the resource. Adding the other fields would leave us in the same boat. Also, I'm wondering if modelStatus and modelStates are too similar. Would just `states` and `copies` be more succinct?

```yaml
modelStatus:
    transitionStatus: UpToDate
    states:
        activeModelState: Loaded
        targetModelState: 
    copies:
        failedCopies: 0
        totalCopies: 2
```

yuzisun: Also rename `ModelStateContainer` to `ModelTransitionState` ?

pvaneck:
Updated again. Let me know if there any other name change suggestions, especially for the json keys.

yuzisun:
> Updated again. Let me know if there any other name change suggestions, especially for the json keys.

lgtm, @njhill do you have more comments?

pkg/apis/serving/v1beta1/inference_service_status.go
njhill: I'm not sure about this one, these reflect states of individual models/revisions rather than the state of the transition (which is what the separate `transitionStatus` field indicates). In the most common case, it would just reflect the state of the the one model backing the predictor (loaded / not loaded / loading / failed).

How about `ModelRevisionStates`?

njhill:
Thanks @pvaneck, looks great! @yuzisun @pvaneck I just have one more comment, added [above](https://github.com/kserve/kserve/pull/2084#discussion_r835658507).

pkg/apis/serving/v1beta1/inference_service_status.go
pvaneck: Yea, that makes sense. Updated.

pvaneck:
I think this should be good to go now. 👍

PR #2088, opened 2022-03-14
Added logic to update ModelStatus

Suresh-Nakkeran:

As part of this PR, ModelStatus field is added in InferenceServiceStatus struct which will be used for updating single model deployment status and ModelMesh predictor status.

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -411,11 +411,10 @@ func (ss *InferenceServiceStatus) UpdateModelRevisionStates(modelState ModelStat
 		ss.ModelStatus.TransitionStatus = InProgress
 	} else if modelState == Loaded {
 		ss.ModelStatus.TransitionStatus = UpToDate
-	} else if modelState == FailedToLoad {
-		ss.ModelStatus.TransitionStatus = BlockedByFailedLoad
-	} else if modelState == Loaded {
 		ss.ModelStatus.ModelCopies = &ModelCopies{TotalCopies: totalCopies}
 		ss.ModelStatus.ModelRevisionStates.ActiveModelState = Loaded
+	} else if modelState == FailedToLoad {
+		ss.ModelStatus.TransitionStatus = BlockedByFailedLoad
 	}
 	if info != nil {
 		ss.SetModelFailureInfo(info)
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -400,11 +400,11 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
 	}
 }
 
-func (ss *InferenceServiceStatus) UpdateModelRevisionStates(modelState ModelState, info *FailureInfo) {
+func (ss *InferenceServiceStatus) UpdateModelRevisionStates(modelState ModelState, totalCopies int, info *FailureInfo) {
 	if ss.ModelStatus.ModelRevisionStates == nil {
-		ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{ActiveModelState: modelState}
+		ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{TargetModelState: modelState}
 	} else {
-		ss.ModelStatus.ModelRevisionStates.ActiveModelState = modelState
+		ss.ModelStatus.ModelRevisionStates.TargetModelState = modelState
 	}
 	// Update transition status, failure info based on new model state
 	if modelState == Pending || modelState == Loading {
@@ -413,6 +413,9 @@ func (ss *InferenceServiceStatus) UpdateModelRevisionStates(modelState ModelStat
 		ss.ModelStatus.TransitionStatus = UpToDate
 	} else if modelState == FailedToLoad {
 		ss.ModelStatus.TransitionStatus = BlockedByFailedLoad
+	} else if modelState == Loaded {
+		ss.ModelStatus.ModelCopies = &ModelCopies{TotalCopies: totalCopies}
+		ss.ModelStatus.ModelRevisionStates.ActiveModelState = Loaded
 	}
 	if info != nil {
 		ss.SetModelFailureInfo(info)
@@ -424,9 +427,9 @@ func (ss *InferenceServiceStatus) UpdateModelTransitionStatus(status TransitionS
 	// Update model state to 'FailedToLoad' in case of invalid spec provided
 	if ss.ModelStatus.TransitionStatus == InvalidSpec {
 		if ss.ModelStatus.ModelRevisionStates == nil {
-			ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{ActiveModelState: FailedToLoad}
+			ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{TargetModelState: FailedToLoad}
 		} else {
-			ss.ModelStatus.ModelRevisionStates.ActiveModelState = FailedToLoad
+			ss.ModelStatus.ModelRevisionStates.TargetModelState = FailedToLoad
 		}
 	}
 	if info != nil {
@@ -444,17 +447,19 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
 
 func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeplyment bool) {
 	// Check at least one pod is running for the latest revision of inferenceservice
-	if len(podList.Items) == 0 {
+	totalCopies := len(podList.Items)
+	if totalCopies == 0 {
+		ss.UpdateModelRevisionStates(Pending, totalCopies, nil)
 		return
 	}
 	// Update model state to 'Loaded' if inferenceservice status is ready.
 	// For serverless deployment, the latest created revision and the latest ready revision should be equal
 	if ss.IsReady() {
 		if rawDeplyment {
-			ss.UpdateModelRevisionStates(Loaded, nil)
+			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
 			return
 		} else if statusSpec.LatestCreatedRevision == statusSpec.LatestReadyRevision {
-			ss.UpdateModelRevisionStates(Loaded, nil)
+			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
 			return
 		}
 	}
@@ -464,18 +469,18 @@ func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatu
 	for _, cs := range podList.Items[0].Status.InitContainerStatuses {
 		if cs.Name == constants.StorageInitializerContainerName {
 			if cs.State.Running != nil {
-				ss.UpdateModelRevisionStates(Loading, nil)
+				ss.UpdateModelRevisionStates(Loading, totalCopies, nil)
 				return
 			} else if cs.State.Terminated != nil &&
 				cs.State.Terminated.Reason == constants.StateReasonError {
-				ss.UpdateModelRevisionStates(FailedToLoad, &FailureInfo{
+				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:  ModelLoadFailed,
 					Message: cs.State.Terminated.Message,
 				})
 				return
 			} else if cs.State.Waiting != nil &&
 				cs.State.Waiting.Reason == constants.StateReasonCrashLoopBackOff {
-				ss.UpdateModelRevisionStates(FailedToLoad, &FailureInfo{
+				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:  ModelLoadFailed,
 					Message: cs.LastTerminationState.Terminated.Message,
 				})
@@ -489,18 +494,18 @@ func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatu
 		if cs.Name == constants.InferenceServiceContainerName {
 			if cs.State.Terminated != nil &&
 				cs.State.Terminated.Reason == constants.StateReasonError {
-				ss.UpdateModelRevisionStates(FailedToLoad, &FailureInfo{
+				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:  ModelLoadFailed,
 					Message: cs.State.Terminated.Message,
 				})
 			} else if cs.State.Waiting != nil &&
 				cs.State.Waiting.Reason == constants.StateReasonCrashLoopBackOff {
-				ss.UpdateModelRevisionStates(FailedToLoad, &FailureInfo{
+				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:  ModelLoadFailed,
 					Message: cs.LastTerminationState.Terminated.Message,
 				})
 			} else {
-				ss.UpdateModelRevisionStates(Pending, nil)
+				ss.UpdateModelRevisionStates(Pending, totalCopies, nil)
 			}
 		}
 	}
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -188,7 +188,7 @@ const (
 )
 
 // FailureReason enum
-// +kubebuilder:validation:Enum=ModelLoadFailed;RuntimeUnhealthy;NoSupportingRuntime;RuntimeNotRecognized;InvalidPredictorSpec
+// +kubebuilder:validation:Enum=ModelLoadFailed;RuntimeUnhealthy;RuntimeDisabled;NoSupportingRuntime;RuntimeNotRecognized;InvalidPredictorSpec
 type FailureReason string
 
 // FailureReason enum values
@@ -423,7 +423,14 @@ func (ss *InferenceServiceStatus) UpdateModelTransitionStatus(status TransitionS
 	ss.ModelStatus.TransitionStatus = status
 	// Update model state to 'FailedToLoad' in case of invalid spec provided
 	if ss.ModelStatus.TransitionStatus == InvalidSpec {
-		ss.UpdateModelRevisionStates(FailedToLoad, info)
+		if ss.ModelStatus.ModelRevisionStates == nil {
+			ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{ActiveModelState: FailedToLoad}
+		} else {
+			ss.ModelStatus.ModelRevisionStates.ActiveModelState = FailedToLoad
+		}
+	}
+	if info != nil {
+		ss.SetModelFailureInfo(info)
 	}
 }
 
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -17,6 +17,9 @@ limitations under the License.
 package v1beta1
 
 import (
+	"reflect"
+
+	"github.com/kserve/kserve/pkg/constants"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -194,6 +197,8 @@ const (
 	ModelLoadFailed FailureReason = "ModelLoadFailed"
 	// Corresponding ServingRuntime containers failed to start or are unhealthy
 	RuntimeUnhealthy FailureReason = "RuntimeUnhealthy"
+	// The ServingRuntime is disabled
+	RuntimeDisabled FailureReason = "RuntimeDisabled"
 	// There are no ServingRuntime which support the specified model type
 	NoSupportingRuntime FailureReason = "NoSupportingRuntime"
 	// There is no ServingRuntime defined with the specified runtime name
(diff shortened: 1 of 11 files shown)


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -223,6 +267,37 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "fails to list inference service pods by label")
+	}
+	if len(podList.Items) > 0 {
+		isvc.Status.PropagateModelStatus(statusSpec, podList, rawDeployment)
+	}
 	return nil
 }
+
+func (p *Predictor) listPodsByLabel(namespace string, labelKey string, labelVal string) (*v1.PodList, error) {
```
yuzisun: Can move this function to the utils file

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -223,6 +267,37 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 		}
 		isvc.Status.PropagateStatus(v1beta1.PredictorComponent, status)
 	}
-
+	statusSpec, _ := isvc.Status.Components[v1beta1.PredictorComponent]
+	if rawDeployment {
+		podLabelValue = "isvc." + isvc.ObjectMeta.Name + "-predictor-default"
```
yuzisun: Let's use the util function [DefaultPredictorServiceName](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L360)

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -268,3 +379,100 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+	// update model status based on kserve container state
+	for _, cs := range podList.Items[0].Status.ContainerStatuses {
+		if cs.Name == constants.InferenceServiceContainerName {
+			if cs.State.Terminated != nil &&
+				cs.State.Terminated.Reason == constants.StateReasonError {
+				ss.UpdateModelStatus(FailedToLoad, &FailureInfo{
+					Reason:  ModelLoadFailed,
+					Message: cs.State.Terminated.Message,
+				})
+				return
```
yuzisun: nit: we do not need return here

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -268,3 +379,100 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+			} else if cs.State.Waiting != nil &&
+				cs.State.Waiting.Reason == constants.StateReasonCrashLoopBackOff {
+				ss.UpdateModelStatus(FailedToLoad, &FailureInfo{
+					Reason:  ModelLoadFailed,
+					Message: cs.LastTerminationState.Terminated.Message,
+				})
+				return
```
yuzisun: nit: we do not need return here

yuzisun:
@Suresh-Nakkeran Please help rebase paul's status API PR

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: @yuzisun Updated, Thanks!

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: @yuzisun Updated, Thanks!

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -383,3 +386,98 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+	}
+	// update model failure info
+	if info != nil {
+		ss.SetModelFailureInfo(info)
+	}
+}
+
+func (ss *InferenceServiceStatus) UpdateModelTransitionStatus(status TransitionStatus, info *FailureInfo) {
+	ss.ModelStatus.TransitionStatus = status
+	// update model state
```
alexagriffith: nit: can we update this comment to something more helpful? The function name `UpdateModelRevisionStates` is clear enough to know its a model state update in my opinion. If we want a comment here, maybe something like
```
// if the model transition status is invalid, we need to update the model revision state and/because .... which will ...
```
?

alexagriffith: kind of similar to the comment on line 427. That comment is helpful and adds good context.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -223,6 +265,19 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+	if rawDeployment {
+		podLabelValue = constants.GetRawServiceLabel(constants.DefaultPredictorServiceName(isvc.ObjectMeta.Name))
+	} else {
+		podLabelValue = statusSpec.LatestCreatedRevision
+	}
+	podList, err := isvcutils.ListPodsByLabel(p.client, isvc.ObjectMeta.Namespace, podLabelKey, podLabelValue)
+	if err != nil {
+		return errors.Wrapf(err, "fails to list inferenceservice pods by label")
+	}
+	if len(podList.Items) > 0 {
```
yuzisun: nit: move the > 0 check into `PropagateModelStatus`

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+					Reason:  v1beta1.RuntimeNotRecognized,
+					Message: "Waiting for runtime to become available",
+				})
 				return err
 			}
 
 			if r.IsDisabled() {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+					Reason:  v1beta1.RuntimeUnhealthy,
```
yuzisun: We might need to define the reason for Disabled runtime instead of using `RuntimeUnhealthy`

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+					Message: "Specified runtime is disabled",
+				})
 				return fmt.Errorf("specified runtime %s is disabled", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
 			// Verify that the selected runtime supports the specified framework.
 			if !isvc.Spec.Predictor.Model.RuntimeSupportsModel(r) {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+					Reason:  v1beta1.NoSupportingRuntime,
+					Message: "Specified runtime does not support specified framework/version",
```
yuzisun, line 116:
```suggestion
					Message: "No runtime found for supporting the specified model format",
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: @yuzisun Added a new reason `RuntimeDisabled`.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: @yuzisun updated, Thanks!

pkg/apis/serving/v1beta1/inference_service_status.go
Suresh-Nakkeran: @yuzisun updated, Thanks!

pkg/apis/serving/v1beta1/inference_service_status.go
Suresh-Nakkeran: @yuzisun updated, Thanks!

pkg/apis/serving/v1beta1/inference_service_status.go
Suresh-Nakkeran: @alexagriffith updated, Thanks!

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -1229,6 +1241,12 @@ var _ = Describe("v1beta1 inference service controller", func() {
 						},
 					},
 				},
+				ModelStatus: v1beta1.ModelStatus{
+					TransitionStatus: v1beta1.InProgress,
+					ModelRevisionStates: &v1beta1.ModelRevisionStates{
+						ActiveModelState: v1beta1.Pending,
```
yuzisun: Can we add a test cases for the error state? We can mock it by creating a pod with storage initializer error state

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -383,3 +388,102 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+	if info != nil {
+		ss.SetModelFailureInfo(info)
+	}
+}
+
+func (ss *InferenceServiceStatus) UpdateModelTransitionStatus(status TransitionStatus, info *FailureInfo) {
+	ss.ModelStatus.TransitionStatus = status
+	// Update model state to 'FailedToLoad' in case of invalid spec provided
+	if ss.ModelStatus.TransitionStatus == InvalidSpec {
+		ss.UpdateModelRevisionStates(FailedToLoad, info)
```
njhill: This will overwrite `TransitionStatus` to be `BlockedByFailedLoad` which I don't think is what we want here?

I'm not sure it makes sense to update the model state here at all actually since an invalid spec doesn't affect the active model state, just means we can't fully reconcile.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 		if isvc.Spec.Predictor.Model.Runtime != nil {
 			r, err := isvcutils.GetServingRuntime(p.client, *isvc.Spec.Predictor.Model.Runtime, isvc.Namespace)
 			if err != nil {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
```
njhill, line 96: Better for `InvalidSpec` here to be `BlockedByFailedLoad`?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+					Reason:  v1beta1.RuntimeNotRecognized,
+					Message: "Waiting for runtime to become available",
+				})
 				return err
 			}
 
 			if r.IsDisabled() {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
```
njhill, line 104: Better for `InvalidSpec` here to be `BlockedByFailedLoad`?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+					Reason:  v1beta1.RuntimeDisabled,
+					Message: "Specified runtime is disabled",
+				})
 				return fmt.Errorf("specified runtime %s is disabled", *isvc.Spec.Predictor.Model.Runtime)
 			}
 
 			// Verify that the selected runtime supports the specified framework.
 			if !isvc.Spec.Predictor.Model.RuntimeSupportsModel(r) {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
```
njhill, line 122: Better for `InvalidSpec` here to be `BlockedByFailedLoad`?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -110,22 +125,38 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 				return err
 			}
 			if len(runtimes) == 0 {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
```
njhill, line 136: Better for `InvalidSpec` here to be `BlockedByFailedLoad`?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -91,15 +94,27 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 		if isvc.Spec.Predictor.Model.Runtime != nil {
 			r, err := isvcutils.GetServingRuntime(p.client, *isvc.Spec.Predictor.Model.Runtime, isvc.Namespace)
 			if err != nil {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+					Reason:  v1beta1.RuntimeNotRecognized,
+					Message: "Waiting for runtime to become available",
+				})
 				return err
```
njhill: For all of these new `UpdateModelTransitionStatus` updates I think we need to return `nil` rather than `err` (but take are to log errors if they aren't returned).

This is because returning an error here will cause `InferenceServiceReconciler.Reconcile` to abort and not actually write the updated Status back to the apiserver. I think we want to continue to reconcile other components even if this fails (and the ingress?).

Alternatively maybe the logic in `InferenceServiceReconciler.Reconcile` could be changed to log but not abort when individual component reconciliations fail?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -223,6 +265,17 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+	if rawDeployment {
+		podLabelValue = constants.GetRawServiceLabel(constants.DefaultPredictorServiceName(isvc.ObjectMeta.Name))
+	} else {
+		podLabelValue = statusSpec.LatestCreatedRevision
+	}
+	podList, err := isvcutils.ListPodsByLabel(p.client, isvc.ObjectMeta.Namespace, podLabelKey, podLabelValue)
+	if err != nil {
+		return errors.Wrapf(err, "fails to list inferenceservice pods by label")
+	}
+	isvc.Status.PropagateModelStatus(statusSpec, podList, rawDeployment)
```
njhill, line 300: _Ideally_ here we would separately check the pods for the "target" revision/replicaset and active/current revision/replicaset, and update the targetModelState and activeModelState fields respectively (unless the service/deployment is up-to-date in which case only active applies and targetModelState can be set to empty).

I know this might be some nontrivial rework though so will leave it to you and @yuzisun to decide whether it's worth doing at this point.

We also probably want to do this part even if the other prediction reconciliation above fails (though this isn't really a comment above this PR, the logic was already this way).

In other words, it would be good if the status checking + update part was separate to the spec processing part (latter being what actually creates the knative service or raw deployment). One of these two things failing shouldn't prevent the other from running. It's still of value for the status to reflect what's going on even if there's problems with the current spec.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -383,3 +388,102 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+		return false
+	}
+	ss.ModelStatus.LastFailureInfo = info
+	return true
+}
+
+func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeplyment bool) {
+	// Check at least one pod is running for the latest revision of inferenceservice
+	if len(podList.Items) == 0 {
+		return
```
njhill, line 452: Wouldn't this also translate to a particular model state. probably `Standby` or `Loading`?

pkg/apis/serving/v1beta1/inference_service_status.go
yuzisun: It does not affect the active model, but invalid spec would result in model deployment failure and we should surface this error back to the user right?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
yuzisun: I think the ingress shouldn't be created if the component reconciliation fails, though maybe status should be always updated even when reconciliation fails and we can use `defer` pattern to update the status.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
yuzisun: I think that's good point! I did not realize that `targetModelState` is not actually populated and we are only setting `activeModelState`. `targetModelState` is equivalent to `lastestCreatedRevision` and `activeModelState` is equivalent to `lastestReadyRevision` in knative mode.

Regarding the spec processing I think there are still dependencies for the resource reconciliations, for example if the deployment reconciliation fails then the controller should not continue to create the ingress resource, otherwise user may not get unexpected error when they call the service. I agree the status checking + update should be separate and happen regardless of the reconciliation result.

pkg/apis/serving/v1beta1/inference_service_status.go
njhill: Yes it will still be reflected in the transitionStatus of `InvalidSpec` and the `failureInfo`. I thought the distinction made sense since it isn't a model loading failure per se, the load isn't even attempted because the spec is invalid.

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -1167,4 +1167,275 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
+
+			var failureInfo = v1beta1.FailureInfo{
+				Reason:  v1beta1.RuntimeDisabled,
+				Message: "Specified runtime is disabled",
+			}
+			Expect(inferenceService.Status.ModelStatus.TransitionStatus).To(Equal(v1beta1.InvalidSpec))
```
yuzisun, line 1350: I think the `eventually` block only waits until inference service exists, will need to try until status object is added.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -394,3 +399,109 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
 		conditionSet.Manage(ss).ClearCondition(conditionType)
 	}
 }
+
+func (ss *InferenceServiceStatus) UpdateModelRevisionStates(modelState ModelState, info *FailureInfo) {
+	if ss.ModelStatus.ModelRevisionStates == nil {
+		ss.ModelStatus.ModelRevisionStates = &ModelRevisionStates{ActiveModelState: modelState}
+	} else {
+		ss.ModelStatus.ModelRevisionStates.ActiveModelState = modelState
```
yuzisun: here should be `TargetModelState` as we are passing the status from `latestCreatedRevision`. `ActiveModelState` is equivalent to `latestReadyRevision` and most of time should be `loaded` when it is not empty, when `latestReadyRevison` is empty and there is only one revision, only `TargetModelState` should be populated. Things get a bit tricky for the raw deployment case, as it is doing rolling upgrade when model is updated, it is hard to differentiate between `ActiveModel` and `TargetModel`.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -394,3 +399,109 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+			return
+		} else if statusSpec.LatestCreatedRevision == statusSpec.LatestReadyRevision {
+			ss.UpdateModelRevisionStates(Loaded, nil)
+			return
+		}
+	}
+	// Update model state to 'Loading' if storage initializer is running.
+	// If the storage initializer is terminated due to error or crashloopbackoff, update model
+	// state to 'ModelLoadFailed' with failure info.
+	for _, cs := range podList.Items[0].Status.InitContainerStatuses {
```
yuzisun, line 468: I think we need to sort the pods in certain way here, otherwise we may get a different pod each time status gets reconciled?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -171,6 +172,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if err := reconciler.Reconcile(isvc); err != nil {
 			r.Log.Error(err, "Failed to reconcile", "reconciler", reflect.ValueOf(reconciler), "Name", isvc.Name)
 			r.Recorder.Eventf(isvc, v1.EventTypeWarning, "InternalError", err.Error())
+			r.updateStatus(isvc)
```
yuzisun: Can we use the defer pattern so that status is always updated even if reconciliation fails.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -394,3 +399,109 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+					Message: cs.State.Terminated.Message,
+				})
+			} else if cs.State.Waiting != nil &&
+				cs.State.Waiting.Reason == constants.StateReasonCrashLoopBackOff {
+				ss.UpdateModelRevisionStates(FailedToLoad, &FailureInfo{
+					Reason:  ModelLoadFailed,
+					Message: cs.LastTerminationState.Terminated.Message,
+				})
+			} else {
+				ss.UpdateModelRevisionStates(Pending, nil)
```
yuzisun: What happens when container is in ready state?

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -1167,4 +1167,275 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			// Create the InferenceService object and expect the Reconcile and knative service to be created
+			Expect(k8sClient.Create(context.TODO(), isvc)).NotTo(gomega.HaveOccurred())
+			defer k8sClient.Delete(context.TODO(), isvc)
+
+			inferenceService := &v1beta1.InferenceService{}
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, predictorServiceKey, inferenceService)
+				if err != nil {
+					return false
+				}
```
yuzisun, line 1339: need to add the following poll logic to make sure model status is populated before asserting
```go
if inferenceService.Status.ModelStatus.LastFailureInfo == nil {
    return false
}
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -205,45 +223,79 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
...
+	var podLabelValue string
+
 	// Here we allow switch between knative and vanilla deployment
 	if isvcutils.GetDeploymentMode(annotations, deployConfig) == constants.RawDeployment {
+		rawDeployment = true
+		podLabelKey = constants.RawDeploymentAppLabel
 		r, err := raw.NewRawKubeReconciler(p.client, p.scheme, objectMeta, &isvc.Spec.Predictor.ComponentExtensionSpec,
 			&podSpec)
 		if err != nil {
+			modelStatus = setModelStatusError(v1beta1.PredictorReconcileFailed, "Fails to create NewRawKubeReconciler for predictor")
```
yuzisun: I think these are internal reconciliation errors which are not suitable to expose to users. We can generate a kube event with `InternalError` type.

yuzisun: actually we are already generating the event at the top level, so I prefer removing this code.
```go
if err := reconciler.Reconcile(isvc); err != nil {
	r.Log.Error(err, "Failed to reconcile", "reconciler", reflect.ValueOf(reconciler), "Name", isvc.Name)
	r.Recorder.Eventf(isvc, v1.EventTypeWarning, "InternalError", err.Error())
}
```

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -394,3 +399,114 @@ func (ss *InferenceServiceStatus) ClearCondition(conditionType apis.ConditionTyp
...
+		ss.ModelStatus.ModelRevisionStates.TargetModelState = modelState
+	}
+	// Update transition status, failure info based on new model state
+	if modelState == Pending || modelState == Loading {
+		ss.ModelStatus.TransitionStatus = InProgress
+	} else if modelState == Loaded {
+		ss.ModelStatus.TransitionStatus = UpToDate
+	} else if modelState == FailedToLoad {
+		ss.ModelStatus.TransitionStatus = BlockedByFailedLoad
+	} else if modelState == Loaded {
```
yuzisun: This condition is the same as line 412

yuzisun, line 416: This condition is the same as line 412

pkg/controller/v1beta1/inferenceservice/components/predictor.go
yuzisun: @njhill I think here it should be `InvalidSpec` as it has not got to the point to load the model yet?

pkg/apis/serving/v1beta1/inference_service_status.go
yuzisun: @Suresh-Nakkeran I agree here that we can remove the code updating to `FailedToLoad`, the model loading has not started yet.

Suresh-Nakkeran: okay @yuzisun

PR #2097, opened 2022-03-17
Fix canary rollout falling back to prev rolledout version

### LINKED ISSUE #1968: When kserve releases a new version of the model, the current version is 9 and the new version is 10. Why does version 8 start ？
/kind bug

Deploy a model which version in knative is 9. Before that version 8 has been deployed successfully.
When I deploy version 10 with canary percent 1% 。As a result pods in version 8 and version 10 has been created. 

Pods in version 10 are created only.

**Environment:**

- Istio Version: 1.11.4
- Knative Version: 1.0.0
- KFServing Version:
- Kubeflow version:
- Minikube version:
- Kubernetes version: (use `kubectl version`): v1.21.3
- OS (e.g. from `/etc/os-release`): centos 7

wenyangchou:

Fixes #1968

```
When canaryTrafficPercent is specified to roll out the new version from the current stable version, the previous rolled out version gets incorrectly spawn up for a brief time due to the condition of latestCreated == latestReady. We need additional check for "latestReady != latestCreated" so that the last rolled out stays on the current stable version. 
```
It seems that the problem has been solved. But the reality is that before the release of the new revision, the `createKnativeService` function would run serveral times wth the old configuration. We might as well think differently, `if else` here as the note says 

`This is to handle case when the latest ready revision is rolled out with 100% and then rolled back
	// so here we need to get the revision that is previously rolled out with 100%` . 

When the latest ready revision is rolled out with 100%, could we rolle out a new revision with the same as previous revision configuration instead of roll back?

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
@@ -19,8 +19,6 @@ package knative
 import (
 	"context"
 	"fmt"
-	"strings"
-
 	"github.com/golang/protobuf/proto"
 	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
 	"github.com/kserve/kserve/pkg/constants"
@@ -88,17 +86,8 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 	}
 	lastRolledoutRevision := componentStatus.LatestRolledoutRevision
 
-	// This is to handle case when the first revision rolled out.
-	if strings.HasSuffix(componentStatus.LatestRolledoutRevision, "-00001") {
-		lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
-	}
-
 	// Log component status and canary traffic percent
-	if componentExtension.CanaryTrafficPercent != nil {
-		log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestReadyRevision", componentStatus.LatestReadyRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision, "PreviousRolledoutRevision", componentStatus.PreviousRolledoutRevision, "CanaryTrafficPercent", *componentExtension.CanaryTrafficPercent)
-	} else {
-		log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestReadyRevision", componentStatus.LatestReadyRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision, "PreviousRolledoutRevision", componentStatus.PreviousRolledoutRevision, "CanaryTrafficPercent", "nil")
-	}
+        log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestReadyRevision", componentStatus.LatestReadyRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision, "PreviousRolledoutRevision", componentStatus.PreviousRolledoutRevision, "CanaryTrafficPercent", componentExtension.CanaryTrafficPercent)
 
 	trafficTargets := []knservingv1.TrafficTarget{}
 	// Split traffic when canary traffic percent is specified
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
@@ -19,6 +19,7 @@ package knative
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
@@ -86,13 +87,19 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 		annotations[autoscaling.ClassAnnotationKey] = autoscaling.KPA
 	}
 	lastRolledoutRevision := componentStatus.LatestRolledoutRevision
-	// This is to handle case when the latest ready revision is rolled out with 100% and then rolled back
-	// so here we need to get the revision that is previously rolled out with 100%
-	if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
-		componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
-		componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent < 100 {
+
+	// This is to handle case when the first revision rolled out.
+	if strings.HasSuffix(componentStatus.LatestRolledoutRevision, "-00001") {
 		lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
 	}
+
+	// Log component status and canary traffic percent
+	if componentExtension.CanaryTrafficPercent != nil {
+		log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestReadyRevision", componentStatus.LatestReadyRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision, "PreviousRolledoutRevision", componentStatus.PreviousRolledoutRevision, "CanaryTrafficPercent", *componentExtension.CanaryTrafficPercent)
+	} else {
+		log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestReadyRevision", componentStatus.LatestReadyRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision, "PreviousRolledoutRevision", componentStatus.PreviousRolledoutRevision, "CanaryTrafficPercent", "nil")
+	}
+
 	trafficTargets := []knservingv1.TrafficTarget{}
 	// Split traffic when canary traffic percent is specified
 	if componentExtension.CanaryTrafficPercent != nil && lastRolledoutRevision != "" {
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -335,7 +335,7 @@ func (ss *InferenceServiceStatus) PropagateStatus(component ComponentType, servi
 					if traffic.Percent != nil && *traffic.Percent < 100 {
 						// check the possibility that the traffic is split over the same revision
 						if val, ok := revisionTraffic[traffic.RevisionName]; ok {
-							if val == 100 {
+							if val == 100 && statusSpec.PreviousRolledoutRevision != "" {
 								statusSpec.LatestRolledoutRevision = statusSpec.PreviousRolledoutRevision
 							}
 						}
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -313,6 +313,12 @@ func (ss *InferenceServiceStatus) PropagateStatus(component ComponentType, servi
 		ss.Components[component] = ComponentStatusSpec{}
 	}
 	statusSpec.LatestCreatedRevision = serviceStatus.LatestCreatedRevisionName
+	revisionTraffic := map[string]int64{}
+	for _, traffic := range serviceStatus.Traffic {
+		if traffic.Percent != nil {
+			revisionTraffic[traffic.RevisionName] += *traffic.Percent
+		}
+	}
 	for _, traffic := range serviceStatus.Traffic {
 		if traffic.RevisionName == serviceStatus.LatestReadyRevisionName && traffic.LatestRevision != nil &&
 			*traffic.LatestRevision {
@@ -327,7 +333,12 @@ func (ss *InferenceServiceStatus) PropagateStatus(component ComponentType, servi
 				// so here we need to rollback the LatestRolledoutRevision to PreviousRolledoutRevision
 				if serviceStatus.LatestReadyRevisionName == serviceStatus.LatestCreatedRevisionName {
 					if traffic.Percent != nil && *traffic.Percent < 100 {
-						statusSpec.LatestRolledoutRevision = statusSpec.PreviousRolledoutRevision
+						// check the possibility that the traffic is split over the same revision
+						if val, ok := revisionTraffic[traffic.RevisionName]; ok {
+							if val == 100 {
+								statusSpec.LatestRolledoutRevision = statusSpec.PreviousRolledoutRevision
+							}
+						}
 					}
 				}
 			}


wenyangchou:
[PR2040](https://github.com/kserve/kserve/pull/2040) does not solve the problem actually.

yuzisun:
@wenyangchou Can you help sign the commit following https://github.com/kserve/kserve/pull/2097/checks?check_run_id=5586553936

yuzisun:
> [PR2040](https://github.com/kserve/kserve/pull/2040) does not solve the problem actually.

@wenyangchou How would you deal with the case when the current revision(2) is rolled out with 100% and then rolled back, when you fix the current version and do a canary rollout with revision (3) it should split the traffic between revision 3 and previously rolled out revision (1) right?

wenyangchou:
> **What this PR does / why we need it**:
> 
> **Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_: Fixes #1968
> 
> **Special notes for your reviewer**:
> 
> ```
> When canaryTrafficPercent is specified to roll out the new version from the current stable version, the previous rolled out version gets incorrectly spawn up for a brief time due to the condition of latestCreated == latestReady. We need additional check for "latestReady != latestCreated" so that the last rolled out stays on the current stable version. 
> ```
> 
> It seems that the problem has been solved. But the reality is that before the release of the new revision, the `createKnativeService` function would run serveral times wth the old configuration. We might as well think differently, `if else` here as the note says
> 
> `This is to handle case when the latest ready revision is rolled out with 100% and then rolled back // so here we need to get the revision that is previously rolled out with 100%` .
> 
> When the latest ready revision is rolled out with 100%, could we rolle out a new revision with the same as previous revision configuration instead of roll back?

I did another detailed test.Take my current model as the example.

The current revision is 02.I roll out revision 03, the `compentStatus` change record is as follows.

|action or event|LastestRolloutRevision|LatestReadyRevision|LatestCreatedRevision|PreviousRolloutRevision|CanaryTrafficPercent|
|-------------------------|--------------|--------------|--------------|--------------|--------------|
|trigger isvc canary 1% roll out revision03|2|2|2|1|1|
|isvc actual create canary 1% roll out revision03|2|2|3|1|1|
|revision03 is ready with 1% percent|2|3|3|1|1|
|revision03 is ready with 100% percent|3|3|3|2|100|
|revision03 roll back when revision03 roll out with only 1%|2|3|3|1|0|
|revision03 roll back when revision03 roll out with only 100%|3|3|3|2|0|

wenyangchou:
> https://github.com/kserve/kserve/pull/2097/checks?check_run_id=5586553936

done

wenyangchou:
> How would you deal with the case when the current revision(2) is rolled out with 100% and then rolled back, when you fix the current version and do a canary rollout with revision (3) it should split the traffic between revision 3 and previously rolled out revision (1) right?

You are right, I have reorganized here.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -90,9 +90,12 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 	// so here we need to get the revision that is previously rolled out with 100%
 	if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
 		componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
-		componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent < 100 {
+		((componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 100) ||
+			(componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 0)) {
```
yuzisun: Not sure if I understand the change, we want to only rollback to previous revision when canaryPercent is < 100% right?

wenyangchou: Exactly. But the process there is a bit chaotic. Let's take my situation above as an example. 

**When we roll out revision 03 with canary percent < 100%, it  will have 2 stages.**

- create revision 03 but revision 03 is not ready. 
- create revision 03 and revision 03 is ready.
In these two case, `lastestRolloutRevision` is 02. As in the second and third rows of the table above which below `trigger isvc canary 1% roll out revision03`
```
lastRolledoutRevision := componentStatus.LatestRolledoutRevision
// This is to handle case when the latest ready revision is rolled out with 100% and then rolled back
// so here we need to get the revision that is previously rolled out with 100%
if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
((componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 100) ||
	(componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 0)) {
lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
}
```
*latestRevision* is 03 and *preRevision* is 02 ,this if branch missed.

**When we roll out revision 03 with canary percent 100%**

As in the forth row of the table above. In this case we don't need rollback to previous, the if branch hit.

Then case when rolled out with 100% and then rolled back is similar to this.

It seems that the previous logic just happens to work, but it doesn't match the actual state, the BUG is caused by another suitation that when I roll out a new revision, a old componentStatus was wrongly triggered.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -90,9 +90,12 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
...
 		lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
 	}
+
+	log.Info(fmt.Sprintf("LatestRolledoutRevision:%s,LatestReadyRevision:%s,LatestCreatedRevision:%s,PreviousRolledoutRevision:%s,CanaryTrafficPercent:%d", componentStatus.LatestRolledoutRevision, componentStatus.LatestReadyRevision, componentStatus.LatestCreatedRevision, componentStatus.PreviousRolledoutRevision, *componentExtension.CanaryTrafficPercent))
```
yuzisun: CanaryTrafficPercent can nil here

wenyangchou: Done.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
yuzisun: @wenyangchou Thanks for the detailed explanation! I think you are right here, we can totally remove the condition. It incorrectly sets the traffic to `PreviousRolledoutRevision` before `LatestCreatedRevision` is updated in status when canarying a new revision. The correct state is set when [propagating the state](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go), so removing this condition is fine and it just won't rollback the traffic immediately in the first reconciliation but once `latestRolledoutRevision` in status is rolled back the next reconciliation applies correctly.

wenyangchou: This can not totally remove the condition for the time being. Because in the scenario of roll out **the first revision** with traffic percent 1%, when `LatestRolledoutRevision` is not empty that means **revision 01** is ready. In this case the if condition should be hint, otherwise the `PreviousRolledoutRevision` would be the same as `LatestRolledoutRevision` instead of empty. As a result, the state of the isvc directly jumps repeatedly between true and false.

All in all, the if condition still needs to be retained.

yuzisun: I am not sure if I follow here, the first revision should be always fully rolled out when there is only one revision. Also when `componentExtension.CanaryTrafficPercent == 100`, `lastRolledoutRevision` variable is not used anyway for setting the traffic.

wenyangchou: Yeath, there are still some conditions missing. I'm adding.

yuzisun: When you rollout a revision with `CanaryTrafficPercent` 0 you will hit the same problem as before where you can incorrectly bring up the previously rolled out revision.

wenyangchou: The if condition has been updated to 

```
if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
((componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 100) ||
	(componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent == 0) ||
	componentStatus.PreviousRolledoutRevision == "") {
lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
}
```
When I rollout a revision with CanaryTrafficPercent 0, it would not hit the if condition. For the `LastestRolloutRevision ` is not equals to the `LatestReadyRevision`. Only when the new revision has been rollout with traffic percent 100 can the `LastestRolloutRevision` be updated.

wenyangchou:
```
componentStatus.PreviousRolledoutRevision == ""
```
another condition has been added.

yuzisun: it is possible as the reconciliation can happen before `LastestCreatedRevision ` and `LatestReadyRevision` in status are updated with the new revision, so you can still hit the if condition.

yuzisun: The case I am talking about is when 
`LastestRolloutRevision`: 2, `LastestRolloutRevision`: 2, `LastestCreatedRevision`: 2
and then you update the isvc for some changes with 0 `CanaryTrafficPercent`, it creates a new revision but you can hit the if condition as they are still the same before status is updated with `LastestCreatedRevision`: 3.

wenyangchou: Emmm, I probably know what you said. After testing, it does.I will think about it.

wenyangchou: There seems to be no good way to solve this special case. Do you have any other good suggestions.

wenyangchou: These states are asynchronous and somewhat chaotic.

yuzisun: I think removing the condition would solve it, for your special case I think we can probably add a validation that it does not allow setting canary traffic for the first revision(don't think there is a good reason to do so as well).

yuzisun: see the events below for rolling out a revision with 0% then 100% and then rollback with 0%.
|action|LastestRollout|LatestReady|LatestCreated|PreviousRollout|Canary|Traffic split|
|-------------------------|--------------|--------------|--------------|--------------|--------------|-------|
|(user)trigger isvc canary 0% roll out revision03|2|2|2|1|0|100(r2)|
|revision03 is created|2|2|3|1|0|100(r2)|
|revision03 is ready |2|3|3|1|0|100(r2)|
|(user)update revision03 with 100% percent|2|3|3|1|100|100(r2)|
|[LastestRollout is updated to LatestReady](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go#L205)|3|3|3|2|100|100(r3)|
|(user)revision03 roll back to 0%|3|3|3|2|0|100(r3)|
|[LastestRollout is updated to PreviousRollout](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go#L213)|2|3|3|2|0|100(r2)|

yuzisun: I think the if condition check is causing the chaos, we should be totally based on the state on the status and `LastestRolloutRevision` is the revision for which the traffic percent should be applied on if the canary percent is < 100.

wenyangchou: > I think removing the condition would solve it, for your special case I think we can probably add a validation that it does not allow setting canary traffic for the first revision(don't think there is a good reason to do so as well).

Well, there seems to be no better solution.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -86,13 +86,22 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
...
+	// And the canary traffic percent of the first revision should always be 100.
+	if componentStatus.PreviousRolledoutRevision == "" {
 		lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
+		fullCanaryTrafficPercent := int64(100)
+		componentExtension.CanaryTrafficPercent = &fullCanaryTrafficPercent
 	}
+
+	// Log component status and canary traffic percent
+	if componentExtension.CanaryTrafficPercent != nil {
+		log.Info(fmt.Sprintf("LatestRolledoutRevision:%s,LatestReadyRevision:%s,LatestCreatedRevision:%s,PreviousRolledoutRevision:%s,CanaryTrafficPercent:%d", componentStatus.LatestRolledoutRevision, componentStatus.LatestReadyRevision, componentStatus.LatestCreatedRevision, componentStatus.PreviousRolledoutRevision, *componentExtension.CanaryTrafficPercent))
```
yuzisun: Here we should be stick with the structured logging used across the repo, so better to be consistent.
```golang
log.Info("revision status:", "LatestRolledoutRevision", componentStatus.LatestRolledoutRevision, "LatestCreatedRevision", componentStatus.LatestCreatedRevision)
```
For more details please see https://pkg.go.dev/github.com/go-logr/logr

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -86,13 +86,22 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
...
 	lastRolledoutRevision := componentStatus.LatestRolledoutRevision
-	// This is to handle case when the latest ready revision is rolled out with 100% and then rolled back
-	// so here we need to get the revision that is previously rolled out with 100%
-	if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
-		componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
-		componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent < 100 {
+
+	// This is to handle case when the first revision rolled out.
+	// And the canary traffic percent of the first revision should always be 100.
+	if componentStatus.PreviousRolledoutRevision == "" {
```
yuzisun: I still do not quite understand the condition here, can you elaborate more?  Initially `PreviousRolledoutRevision` is always empty and is only populated when you rolled out a second revision.  Seems the logic here should be
```golang
if componentStatus.LastestRolledoutRevision == "" && componentExtension.CanaryTrafficPercent  != nil {
     fullCanaryTrafficPercent := int64(100)
     componentExtension.CanaryTrafficPercent = &fullCanaryTrafficPercent
}
```

wenyangchou: Right. When isvc rollout for the first time, there is no `PreviousRolledoutRevision`. At this time, the `PreviousRolledoutRevision` should always be empty and the `CanaryTrafficPercent` should be set to 100 as well，because setting to less than 100 makes no sense at all.  If not, it will lead to the situation I discussed above.

> This can not totally remove the condition for the time being. Because in the scenario of roll out **the first revision** with traffic percent 1%, when `LatestRolledoutRevision` is not empty that means **revision 01** is ready. In this case the if condition should be hint, otherwise the `PreviousRolledoutRevision` would be the same as `LatestRolledoutRevision` instead of empty. As a result, the state of the isvc directly jumps repeatedly between true and false.
> 
> All in all, the if condition still needs to be retained.

yuzisun: I think here we should check `LatestRolledoutRevision` instead of `PreviousRolledoutRevision` as `PreviousRolledoutRevision == ""` can be in the rollout of second revision rollout where  `CanaryTrafficPercent` can be allowed.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -86,13 +86,22 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
...
 		lastRolledoutRevision = componentStatus.PreviousRolledoutRevision
```
yuzisun: this line can be removed.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
wenyangchou: Would check `LatestCreatedRevision` be better? As the second revision is rolled out, `LatestRolledoutRevision` may not have been updated.

yuzisun: hmm though it is also possible `PreviousRolledoutRevision` is still empty when you rollout second revision where `CanaryTrafficPercent` should be allowed. `PreviousRolledoutRevision` is only updated when you fully rollout the second revision.

yuzisun: I think checking `LatestRolledoutRevision` would be fine, we can document or return an error that you can only apply `CanaryTrafficPercent` until `LatestRolledoutRevision` is updated.

wenyangchou: If check `LatestRolledoutRevision`, when I roll out revision 02.
```
if [condition] {
     fullCanaryTrafficPercent := int64(100)
     componentExtension.CanaryTrafficPercent = &fullCanaryTrafficPercent
}
``` 
The if condition will be hit, it result in the direct full roll out of the revision 02. Is that so, or do I have a problem understanding?

wenyangchou: `LatestCreatedRevision` has been updated to 02 and `LatestRolledoutRevision` is 01, the if condition hit and the traffic percent will be 100.

wenyangchou: Maybe you are right, I need to test it again.

wenyangchou: After testing, you're really right.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -86,13 +87,19 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
...
 	}
 	lastRolledoutRevision := componentStatus.LatestRolledoutRevision
-	// This is to handle case when the latest ready revision is rolled out with 100% and then rolled back
-	// so here we need to get the revision that is previously rolled out with 100%
-	if componentStatus.LatestRolledoutRevision == componentStatus.LatestReadyRevision &&
-		componentStatus.LatestReadyRevision == componentStatus.LatestCreatedRevision &&
-		componentExtension.CanaryTrafficPercent != nil && *componentExtension.CanaryTrafficPercent < 100 {
+
+	// This is to handle case when the first revision rolled out.
+	if strings.HasSuffix(componentStatus.LatestRolledoutRevision, "-00001") {
```
yuzisun: this might not be a reliable check as user can choose to override the revision name, I think it is fine to check if `LatestRolledoutRevision` is not empty.

also why do we still need `lastRolledoutRevision = componentStatus.PreviousRolledoutRevision` ?

wenyangchou: `lastRolledoutRevision = componentStatus.PreviousRolledoutRevision` is actually for set `lastRolledoutRevision` to empty. When I roll out the first revision with 1 traffic percent, The revision status change log is as follows:

```
{"level":"info","ts":1647955245.7690232,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955245.9323924,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955246.1248095,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955246.2239192,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955246.240706,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
```

```
{"level":"info","ts":1647955295.2115061,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.2362065,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.3836544,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.4536629,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.5241714,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.5614073,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.8245306,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.924045,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.9450102,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955295.9623299,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.0252357,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.1252217,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.1559985,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.2245576,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.2484438,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.3239365,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.3661742,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.7657032,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.8372052,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.9283204,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955296.9545665,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.1925607,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.2462313,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.4247804,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.5241365,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.7251525,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.749457,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.933567,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955297.964174,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.2083428,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.2352567,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.2509632,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.3649652,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.3926592,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.5239766,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.63417,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.7239954,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955298.7527847,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.1775756,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.252463,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.3240964,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.3448806,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.372835,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.4322815,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.4523559,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.6674411,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.7398708,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.823951,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.8468132,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955299.8683274,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.0341926,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.0640078,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.139318,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.185193,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.23705,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.4156194,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.448103,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.5240514,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.624298,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.655463,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.8565493,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.8808355,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.8999777,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.9406445,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955300.9859104,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955301.0343618,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
{"level":"info","ts":1647955301.0574412,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":0}
```
When isvc status is ready, the status starts to change constantly
```
NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.amh-group.com   False   100    0        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   70s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.amh-group.com   False          100                              test-pmml-predictor-default-00001   71s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.amh-group.com   False   100    0        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   73s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.amh-group.com   True    100    0        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   75s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.amh-group.com   True 
``` 
If I remove the if condition and not set `lastRolledoutRevision` to empty, how can I solve this problem?

wenyangchou: > this might not be a reliable check as user can choose to override the revision name, I think it is fine to check if `LatestRolledoutRevision` is not empty.
> 
> also why do we still need `lastRolledoutRevision = componentStatus.PreviousRolledoutRevision` ?

Can we overwrite the revision name by isvc? If not I think we do not have to consider this condition,  at least not particularly necessary at the moment.

yuzisun: I thought we talked about that the canaryTrafficPercent 1 does not really make sense for the first revision rollout, we can override that to 100 and only allowing it for the second revision rollout. Any problem you see from the following logic?

```golang
if componentStatus.LatestRolledoutRevision == "" && canaryTrafficPercent != nil {
    canaryTrafficPercent = 100
}
```

also I would not rely on the revision name to check the first revision, knative can change that and in fact in older release it is a generated name,  also knative allows overriding that and we just did not expose it in kserve.

wenyangchou:
```
if componentStatus.LatestRolledoutRevision == "" && canaryTrafficPercent != nil {
    canaryTrafficPercent = 100
}
```
With the logic, isvc changes as follows:
```
NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True    99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   62s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True    99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   63s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True           100                              test-pmml-predictor-default-00001   64s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   False   99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   65s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True    99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   67s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   False          100                              test-pmml-predictor-default-00001   68s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True    99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   70s

NAME        URL                                                          READY   PREV   LATEST   PREVROLLEDOUTREVISION               LATESTREADYREVISION                 AGE
test-pmml   http://test-pmml.phantom.dev-predict-phantom.wenyangchou.com   True    99     1        test-pmml-predictor-default-00001   test-pmml-predictor-default-00001   72s
```
revision status changes :
```
{"level":"info","ts":1647998486.6303008,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998486.7241223,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998486.9247065,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998486.9527392,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998487.024337,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":100}
{"level":"info","ts":1647998487.0524247,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":100}
{"level":"info","ts":1647998487.078394,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":100}
{"level":"info","ts":1647998487.2460775,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":100}
{"level":"info","ts":1647998487.2860572,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":100}
{"level":"info","ts":1647998487.3398628,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998487.5762186,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998487.638925,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998487.8716,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
{"level":"info","ts":1647998487.9417357,"logger":"KsvcReconciler","msg":"revision status:","LatestRolledoutRevision":"test-pmml-predictor-default-00001","LatestReadyRevision":"test-pmml-predictor-default-00001","LatestCreatedRevision":"test-pmml-predictor-default-00001","PreviousRolledoutRevision":"","CanaryTrafficPercent":1}
```
If I do not set the previous revision, the latest revision is the same as the previous revision and compete with each other because the isvc will trigger asynchronously.

yuzisun: I can see the issue now, initially [`LatestRolledoutRevision` was empty](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go#L118) so in the reconcile step it updates traffic split to 100 for the revision 1, then `LatestRolledoutRevision` was updated to revision 1 in status propagation, the next reconcile step updates traffic split to 1/99 for revision 1 because [`LatestRolledoutRevision` is not empty and `canaryTrafficPercent` is set](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go#L98), finally in the status propagation it [rolls back `LatestRolledoutRevision` to `PreviousRolledoutRevision`](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go#L215) as the latest revision split is < 100 which results in infinite loop.

yuzisun: the problem with your check is that it prevents doing canary rollout for the second revision, so let's think what would be the proper fix.

wenyangchou: I think really solve this problem, we need to add a status to mark whether the current revision is before or after `updating knative`. The revision can only be used for judgment after `updating knative`.

yuzisun: We could store the `generation` field in the isvc status,  every time isvc status propagation is invoked from ksvc status update the controller updates the isvc status `generation` field.

yuzisun: @wenyangchou The fix should work for all the cases without spurious revision spawn up, let's get in this first and we can discuss with your idea if that can improve the current logic

wenyangchou: There are still some preblems. When I roll out the first revision with 100 traffic percent, and then roll out the first revision with 0 traffic percent, the isvc status would always be false.

yuzisun: @wenyangchou It should be fixed for your case now with https://github.com/kserve/kserve/pull/2097/commits/8eb4ccd85dd364df2753933420d07975792ad0f9

wenyangchou: approve

wenyangchou: approve

wenyangchou:
approve

wenyangchou:
approve

yuzisun:
@wenyangchou thanks for the great collaboration and thorough testing to make the canary rollout feature robust! Really appreciate it!

/lgtm
/approve

PR #2101, opened 2022-03-17
Fix cluster local label InferenceService

TimKleinloog:

This PR fixes the advanced cluster local example for recent (recommended) Knative versions. The label to create a local InferenceService has changed to `networking.knative.dev/visibility: cluster-local`, see https://knative.dev/docs/serving/services/private-services/.

**Type of changes**

I created the cluster_local.yaml example with:
* KServe: 0.7.0, 
* Istio 1.12.2, 
* Knative 0.23.1

Using the label in this will create a private service, the original not.

### DIFF
--- a/docs/samples/v1beta1/advanced/cluster_local.yaml
+++ b/docs/samples/v1beta1/advanced/cluster_local.yaml
@@ -3,7 +3,7 @@ kind: "InferenceService"
 metadata:
   name: "sklearn-iris-local"
   labels:
-    serving.knative.dev/visibility: cluster-local
+    networking.knative.dev/visibility: cluster-local
 spec:
   predictor:
     sklearn:


pvaneck:
Thanks for the label fix! Can you sign off your commit to pass the DCO check? Some Instructions [here](https://github.com/kserve/kserve/pull/2101/checks?check_run_id=5592848710).

TimKleinloog:
Sure, sign-off added. Happy to contribute!

yuzisun:
Also find the deprecated annotation name [here](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L172), probably should change this too.

yuzisun:
> Sure, sign-off added. Happy to contribute!



> Also find the deprecated annotation name [here](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L172), probably should change this too.

@TimKleinloog Are you able to verify if  just changing the label name on isvc would work as I find there are a couple of places in the code using the old label name?

TimKleinloog:
@yuzisun just changing the label on the isvc did change the URL for me from:
```
NAME            URL                                        READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                     AGE
lightgbm-iris   http://lightgbm-iris.default.example.com   True           100                              lightgbm-iris-predictor-default-00002   18h
```
to
```
NAME            URL                                              READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION                     AGE
lightgbm-iris   http://lightgbm-iris.default.svc.cluster.local   True           100                              lightgbm-iris-predictor-default-00001   18h
```
I will dive a bit deeper if I experience any issues related to the other labels. Do you expect scenarios where this will be the case?
I think we need to change them anyway and could be logical to include them in this PR.

PR #2118, opened 2022-03-28
added protocolversion in servingruntime spec

Suresh-Nakkeran:

This PR is to add ProtocolVersion in ServingRuntimeSpec top-level.


Suresh-Nakkeran:
@yuzisun @pvaneck @njhill This is a new PR, I have kept the protocol version in servingruntime spec top-level. And also removed `grpc-v1`, `grpc-v2`. Can you please review?

config/runtimes/kserve-torchserve.yaml
```diff
@@ -7,6 +7,7 @@ spec:
     - name: pytorch
       version: "1"
       autoSelect: true
+  protocolVersion: ""
```
yuzisun: Should this be a list as torchserve supports both v1 and v2

pkg/apis/serving/v1beta1/predictor_model_test.go
```diff
@@ -36,6 +36,7 @@ func TestGetSupportingRuntimes(t *testing.T) {
 	sklearnRuntime := "sklearn-runtime"
 	pmmlRuntime := "pmml-runtime"
 	mlserverRuntime := "mlserver-runtime"
+	clusterPrefix := "cluster-"
```
yuzisun: what's this `clusterPrefix` for ?

Suresh-Nakkeran: @yuzisun Previously we receive servingruntimespecs as list, now we receive it as map..It's just to differentiate servingruntimes and clusterservingruntimes names...

config/runtimes/kserve-torchserve.yaml
Suresh-Nakkeran: @yuzisun I changed it to accept list of protocol versions.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -86,6 +87,10 @@ type ServingRuntimeSpec struct {
 	// +optional
 	Disabled *bool `json:"disabled,omitempty"`
 
+	// Supported protocol versions (i.e. v1, v2 or empty for unknown protocol version).
+	// +optional
+	ProtocolVersions []constants.InferenceServiceProtocol `json:"protocolVersions"`
```
pvaneck: Should we add omitempty in the json struct tag?

njhill:
Thanks a lot @Suresh-Nakkeran I completely missed that you'd created a new PR for this. I added a comment to the other PR which I'll copy here:

Regarding how to select a runtime during reconcilation when more than one "match" with autoselect, another idea would be to order by the runtime CR's `Revision` rather than name. That should be a bit more "stable" and not cause InferenceServices to move between runtimes if for example a new runtime was created with autoselect=true.

pkg/constants/constants.go
```diff
@@ -191,8 +191,9 @@ const (
 
 // InferenceService protocol enums
 const (
-	ProtocolV1 InferenceServiceProtocol = "v1"
-	ProtocolV2 InferenceServiceProtocol = "v2"
+	ProtocolV1      InferenceServiceProtocol = "v1"
+	ProtocolV2      InferenceServiceProtocol = "v2"
+	ProtocolUnknown InferenceServiceProtocol = ""
```
pvaneck: Do we explicitly need a ProtocolUnknown const? or would the absence of entries in the list be enough?

pkg/apis/serving/v1beta1/predictor_model.go
```diff
@@ -100,7 +100,12 @@ func (ss stringSet) contains(s string) bool {
 // GetSupportingRuntimes Get a list of ServingRuntimeSpecs that correspond to ServingRuntimes and ClusterServingRuntimes that
 // support the given model. If the `isMMS` argument is true, this function will only return ServingRuntimes that are
 // ModelMesh compatible, otherwise only single-model serving compatible runtimes will be returned.
-func (m *ModelSpec) GetSupportingRuntimes(cl client.Client, namespace string, isMMS bool) ([]v1alpha1.ServingRuntimeSpec, error) {
+func (m *ModelSpec) GetSupportingRuntimes(cl client.Client, namespace string, isMMS bool) (map[string]v1alpha1.ServingRuntimeSpec, error) {
+
+	modelProtcolVersion := constants.ProtocolV1
```
pvaneck: I think we should still support the cases where a user:
  1. has a ServingRuntime that omits the `protocolVersions` list 
  2. deploys an InferenceService without specifying a `protocolVersion`

So, pretty much the current behavior we have now. Defaulting the modelProtocolVersion here to ProtocolV1 here will prevent the matching of the model to the intended ServingRuntime in these cases.

Thenmaybe when you have the final map of supported runtimes, first sort by protocol (to ensure v1 is at the top of the list), then by name or whatever we decide) 

@yuzisun @njhill WDYT?

pvaneck (commented):
Thanks, @Suresh-Nakkeran. This is looking good. Left some comments inline.

pvaneck:
> Regarding how to select a runtime during reconcilation when more than one "match" with autoselect, another idea would be to order by the runtime CR's Revision rather than name. That should be a bit more "stable" and not cause InferenceServices to move between runtimes if for example a new runtime was created with autoselect=true.

Is `Revision` a metadata field of the CR? I don't see that field.

Suresh-Nakkeran:
> Thanks a lot @Suresh-Nakkeran I completely missed that you'd created a new PR for this. I added a comment to the other PR which I'll copy here:
> 
> Regarding how to select a runtime during reconcilation when more than one "match" with autoselect, another idea would be to order by the runtime CR's `Revision` rather than name. That should be a bit more "stable" and not cause InferenceServices to move between runtimes if for example a new runtime was created with autoselect=true.

pkg/apis/serving/v1alpha1/servingruntime_types.go
Suresh-Nakkeran: yes @pvaneck, I missed to add. Thanks!

pkg/constants/constants.go
Suresh-Nakkeran: I think, we can remove the protocolUnkown. I have not used it elsewhere.

pkg/apis/serving/v1beta1/predictor_model.go
Suresh-Nakkeran: yes @pvaneck . I agree.

njhill:
> Is `Revision` a metadata field of the CR? I don't see that field.

@pvaneck @Suresh-Nakkeran oops sorry, I meant `ResourceVersion` (in the [`ObjectMeta`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L171)).

pkg/apis/serving/v1beta1/predictor_model.go
njhill: I agree too :)

yuzisun:
> > Is `Revision` a metadata field of the CR? I don't see that field.
> 
> @pvaneck @Suresh-Nakkeran oops sorry, I meant `ResourceVersion` (in the [`ObjectMeta`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L171)).

@njhill seems like `resourceVersion` is for a given resource, not sure how the value is compared with a different runtime resource and if the value is time related. Maybe we can use the `creationTimestamp` field?

config/runtimes/kserve-tritonserver.yaml
```diff
@@ -18,6 +21,8 @@ spec:
     - name: triton
```
yuzisun, line 21: I guess pytorch is not auto selected because torchserve also support v2 protocol right?

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -384,15 +276,32 @@ func (isvc *InferenceService) setMlServerDefaults() {
...
+	// set 'v2' as default protocol version for triton server
+	if isvc.Spec.Predictor.Model.ProtocolVersion == nil {
+		protocolV2 := constants.ProtocolV2
+		isvc.Spec.Predictor.Model.ProtocolVersion = &protocolV2
+	}
+	// set model-control-model arg to 'explicit' if storage uri is nil
+	if isvc.Spec.Predictor.Model.StorageURI == nil {
+		isvc.Spec.Predictor.Model.Args = append(isvc.Spec.Predictor.Model.Args,
+			fmt.Sprintf("%s=%s", "--model-control-mode", "explicit"))
+	}
```
yuzisun, line 306: wondering if this can be moved to the triton serving runtime itself instead of explicit code here.

pkg/apis/serving/v1beta1/predictor_model_test.go
yuzisun: ok can you help leave a comment or rename it to `clusterServingRuntimePrefix`?

config/runtimes/kserve-tritonserver.yaml
Suresh-Nakkeran: yes @yuzisun

pkg/apis/serving/v1beta1/inference_service_defaults.go
Suresh-Nakkeran: @yuzisun If we add `--model-control-mode=explicit` arg for single model serving(storageUri != nil), the model is not registered.

njhill:
> > > Is `Revision` a metadata field of the CR? I don't see that field.
> > 
> > 
> > @pvaneck @Suresh-Nakkeran oops sorry, I meant `ResourceVersion` (in the [`ObjectMeta`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L171)).
> 
> @njhill seems like `resourceVersion` is for a given resource, not sure how the value is compared with a different runtime resource and if the value is time related. Maybe we can use the `creationTimestamp` field?

@yuzisun actually it's global, at least for all resources of a given type (in practice I think it's just the etcd last-modified revision).

But yes `creationTimestamp` is probably better. I was thinking that `resourceVersion` would give better stability e.g. if a `ServingRuntime` was modified to add in a supported model format already covered by a different one. But a similar situation could happen if `resourceVersion` so may as well go with the less-ambiguous option.

pkg/apis/serving/v1beta1/inference_service_defaults.go
njhill: @Suresh-Nakkeran @yuzisun I'm also wondering if we could handle most of this other stuff in the serving runtimes. Basically having some other flag in the ServingRuntime to indicate that it's a "legacy style" mutli-model server, similar to the existing `multiModel` field. Then have a separate ServingRuntime defined for each kind of usage. So in theory you could have three Triton ServingRuntimes defined:
- Single model server
- "Legacy" multi model server
- Modelmesh multi model server

Then if storageUri isn't set we select only runtimes with this flag set. I think this would mean the MLServer-specific logic above in `SetMlServerDefaults()` could also be moved to the runtime definition.

WDYT?

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -384,15 +276,32 @@ func (isvc *InferenceService) setMlServerDefaults() {
...
 	if isvc.ObjectMeta.Labels == nil {
 		isvc.ObjectMeta.Labels = map[string]string{constants.ServiceEnvelope: constants.ServiceEnvelopeKServe}
 	} else {
 		isvc.ObjectMeta.Labels[constants.ServiceEnvelope] = constants.ServiceEnvelopeKServe
 	}
-	if isvc.Spec.Predictor.Model.ProtocolVersion != nil &&
-		constants.ProtocolV2 == *isvc.Spec.Predictor.Model.ProtocolVersion {
+	if constants.ProtocolV2 == *isvc.Spec.Predictor.Model.ProtocolVersion {
 		isvc.ObjectMeta.Labels[constants.ServiceEnvelope] = constants.ServiceEnvelopeKServeV2
 	}
```
njhill, line 293: How about if we had separate SRs defined for TorchServe v1 and v2 protocols?

pkg/apis/serving/v1beta1/inference_service_defaults.go
yuzisun: For MLServer it requires adding the framework class label based on the model format, if we move to the runtime definition then it requires splitting into multiple similar looking runtimes which would be hard to maintain. Maybe the runtime should be like helm template that allows conditions.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -198,6 +203,12 @@ type ClusterServingRuntimeList struct {
 	Items           []ClusterServingRuntime `json:"items"`
 }
 
+// SupportedRuntime is the schema for supported runtime result of automaic selection
```
pvaneck: nit: s/automaic/automatic/

config/runtimes/kserve-mlserver.yaml
```diff
@@ -6,13 +6,18 @@ spec:
...
       version: "1"
+      autoSelect: true
     - name: lightgbm
       version: "3"
+      autoSelect: true
     - name: mlflow
       version: "1"
       autoSelect: true
+  protocolVersions:
+    - v2
```
yuzisun, line 20:
```yaml
- v2
- grpc-v2
```

yuzisun: for serverless mode it can only be either rest or grpc, but for raw deployment you can actually do both rest and grpc in the same serving runtime.

pvaneck (commented):
Can you also run `make generate` for the Open API generation?

config/runtimes/kserve-torchserve.yaml
```diff
@@ -7,6 +7,9 @@ spec:
     - name: pytorch
       version: "1"
       autoSelect: true
+  protocolVersions:
+    - v1
```
yuzisun, line 11: it also supports grpc-v1 and `grpc-v2` has not implemented yet.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -210,3 +221,15 @@ func (srSpec *ServingRuntimeSpec) IsDisabled() bool {
...
+func (srSpec *ServingRuntimeSpec) IsProtocolVersionSupported(modelProtocolVersion constants.InferenceServiceProtocol) bool {
+	if len(modelProtocolVersion) == 0 || srSpec.ProtocolVersions == nil || len(srSpec.ProtocolVersions) == 0 {
+		return true
+	}
+	for _, srProtocolVersion := range srSpec.ProtocolVersions {
+		if srProtocolVersion == modelProtocolVersion {
+			return true
+		}
+	}
+	return false
```
yuzisun, line 236: Do we need to add a protocol field on inference service spec as it will never match with `grpc-v1` or `grpc-v2`?

config/runtimes/kserve-torchserve.yaml
Suresh-Nakkeran: @yuzisun updated, Thanks!

pkg/apis/serving/v1alpha1/servingruntime_types.go
Suresh-Nakkeran: @pvaneck updated, Thanks!

pkg/apis/serving/v1alpha1/servingruntime_types.go
Suresh-Nakkeran: @yuzisun I added grpc-v1 and grpc-v2 in constants.. It should work.

pkg/apis/serving/v1beta1/inference_service_defaults.go
yuzisun: Created https://github.com/kserve/kserve/issues/2141 for follow up.

yuzisun:
Thanks @Suresh-Nakkeran !

/lgtm
/approve

PR #2124, opened 2022-03-31
Add default container annotation

haoxins:

Will be a bit more easier for debugging.

**Release note**:

```release-note
Add default container annotation into the Pod
```


haoxins:
now, the DCO is passed :)

pvaneck:
Thanks!
/lgtm

PR #2136, opened 2022-04-07
Enable transformers to work with ModelMesh

chinhuang007:
The PR enables transformers to work with ModelMesh predictors in the same InferenceService instance

It addresses the [issue](https://github.com/kserve/modelmesh-serving/issues/127). The design doc is available [here](https://docs.google.com/document/d/18KfONdaaLIaYw1FLj0xxQ-o35mVi5LsjjWjS_2PnZxs/edit?usp=sharing)

**Type of changes**
- [x ] New feature (non-breaking change which adds functionality)

Plan to add or update transformer examples with ModelMesh predictor as a separate PR

**Checklist**:
- [x ] Has code been commented, particularly in hard-to-understand areas?

**Release note**:

```release-note
Transformers are now enabled to work with ModelMesh predictors in the same InferenceService instance.
```


pkg/apis/serving/v1beta1/transformer_custom.go
```diff
@@ -72,6 +73,23 @@ func (c *CustomTransformer) GetContainer(metadata metav1.ObjectMeta, extensions
 	container := &c.Containers[0]
 	argumentPredictorHost := fmt.Sprintf("%s.%s", constants.DefaultPredictorServiceName(metadata.Name), metadata.Namespace)
 
+	deploymentMode, ok := metadata.Annotations[constants.DeploymentMode]
+	logger, _ := pkglogging.NewLogger("", "INFO")
+	if ok && (deploymentMode == string(constants.ModelMeshDeployment)) {
+		// Get predictor host and protocol from annotations in modelmesh deployment mode
+		argumentPredictorHost = metadata.Annotations["predictor-host"]
+		argumentPredictorProtocol := metadata.Annotations["predictor-protocol"]
```
yuzisun, line 86: Should we add a `protocol` field on the `InferenceService` component spec instead of using the annotation?

chinhuang007:
The use of annotation is intended to minimize spec and interface changes. We could certainly make a bigger spec change by adding a new field and then have it populated as needed.

pkg/apis/serving/v1beta1/transformer_custom.go
pvaneck: @yuzisun I don't believe these annotations are meant for the user to use, but only as an internal intermediary for passing in needed values from the transformer Reconcile function.

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -79,6 +87,32 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) error {
...
+		}
+
+		// add predictor host and protocol to metadata
+		predictorURL, err := url.Parse(isvc.Status.Components["predictor"].URL.String())
+		if err != nil {
+			return fmt.Errorf("unable to parse predictor URL: %v", err)
+		}
+
+		isvc.ObjectMeta.Annotations["predictor-host"] = predictorURL.Host
+		if predictorURL.Scheme == "http" {
```
pvaneck: nit: add the grpc check as the first conditional check since in most cases, the URL from ModelMesh will be gRPC based.

Also, for completeness, check for `https` as a potential scheme as well.

pkg/constants/constants.go
```diff
@@ -109,6 +109,13 @@ var (
 	DefaultMinReplicas  = 1
 )
 
+// Predictor Protocol Constants
+var (
```
pvaneck: Heads up that https://github.com/kserve/kserve/pull/2118/files#diff-36043890c52c8201a8bc84238c219be45ce07bf172d89f693c7c54ffe70d046e is defining some protocol constants and will soon be merged, so you can rebase once that is in.

pkg/apis/serving/v1beta1/transformer_custom.go
```diff
@@ -72,6 +73,23 @@ func (c *CustomTransformer) GetContainer(metadata metav1.ObjectMeta, extensions
 	container := &c.Containers[0]
 	argumentPredictorHost := fmt.Sprintf("%s.%s", constants.DefaultPredictorServiceName(metadata.Name), metadata.Namespace)
 
+	deploymentMode, ok := metadata.Annotations[constants.DeploymentMode]
+	logger, _ := pkglogging.NewLogger("", "INFO")
```
pvaneck: I think we should use the logging library that is typically used in KServe:

```go
import (
    logf "sigs.k8s.io/controller-runtime/pkg/log"
)

var log = logf.Log.WithName("CustomTransformerReconciler")
```

Another [example](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_validation.go#L40).

pvaneck (commented):
Thanks, @chinhuang007. Left some comments.

pkg/apis/serving/v1beta1/transformer_custom.go
chinhuang007: Thanks for pointing out the common KServe logger. Just updated, please take a look.

pkg/apis/serving/v1beta1/transformer_custom.go
```diff
@@ -72,6 +78,23 @@ func (c *CustomTransformer) GetContainer(metadata metav1.ObjectMeta, extensions
...
+		argumentPredictorHost = metadata.Annotations["predictor-host"]
+		argumentPredictorProtocol := metadata.Annotations["predictor-protocol"]
+
+		// Set predictor protocol if not provided in container arguments
+		if !utils.IncludesArg(container.Args, "--protocol") {
+			customTransformerLogger.Info("Set predictor protocol based on ModelMesh predictor URL", "protocol", argumentPredictorProtocol)
+			container.Args = append(container.Args, []string{
+				"--protocol",
+				argumentPredictorProtocol,
+			}...)
```
njhill:
```suggestion
			container.Args = append(container.Args, "--protocol", argumentPredictorProtocol)
```

pkg/apis/serving/v1beta1/transformer_custom.go
njhill: @chinhuang007 I think it would be simpler to just add an additional `predictorURL` arg to this `GetContainer` function? Then no intermediary annotations would be needed.

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -79,6 +87,32 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) error {
 		}),
 		Annotations: annotations,
 	}
+
+	// Need to wait for predictor URL in modelmesh deployment mode
+	if isvcutils.GetDeploymentMode(annotations, deployConfig) == constants.ModelMeshDeployment {
+		// check if predictor URL is populated
+		if isvc.Status.Components["predictor"].URL == nil {
+			// exit transformer reconcile with an error when predictor URL not populated
+			return fmt.Errorf("Predictor URL from ModelMesh is not ready")
```
njhill: I think we would want to differentiate this from actual errors since it just means "not ready yet". Instead of returning an error from the top-level reconcile func it would be better to return e.g. `ctrl.Result{RequeueAfter: 2 * time.Second}` (and could log an info message here).

I guess an additional return value might need to be added to this func to facilitate that.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -223,7 +230,7 @@ func (r *InferenceServiceReconciler) updateStatus(desiredService *v1beta1api.Inf
 		// This is important because the copy we loaded from the informer's
```
njhill, line 234: Not sure whether the DeepEqual comparison above should be reduced in scope to not look at the parts of the status that the modelmesh controller "owns" in the case that this is a modelmesh deployment?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -223,7 +230,7 @@ func (r *InferenceServiceReconciler) updateStatus(desiredService *v1beta1api.Inf
...
 		// cache may be stale and we don't want to overwrite a prior update
 		// to status with this stale state.
-	} else if err := r.Status().Update(context.TODO(), desiredService); err != nil {
+	} else if err := r.Status().Patch(context.TODO(), desiredService, client.Merge, &client.PatchOptions{}); err != nil {
```
njhill: Could add a comment explaining why we're doing a patch rather than update here?

njhill (commented):
Thanks @chinhuang007 looks good, I added a few comments inline.

pkg/apis/serving/v1beta1/transformer_custom.go
chinhuang007: @njhill GetContainer is an interface, also implemented in other places. The goal is to minimize interface change therefore use of intermediary annotations.

pkg/controller/v1beta1/inferenceservice/components/transformer.go
chinhuang007: @njhill This requires interface changes as well. I will be happy to make such changes if necessary.

pkg/controller/v1beta1/inferenceservice/controller.go
chinhuang007: This is needed to prevent from overriding status when ModelMesh and KServe were not using the same InferenceService status spec. After ModelMesh changed to use the type definitions from KServe, Patch is no longer needed. Will change back to Update.

pkg/apis/serving/v1beta1/transformer_custom.go
chinhuang007: @yuzisun @njhill @pvaneck The current use of intermediary annotations is aimed to minimize interface changes. Adding a new arg to GetContainer, part of ComponentImplementation interface, would require changes in 30+ files, including predictor, explainer, and a bunch of examples. Not a difficult task, I just wanted to discuss whether it is warranted before making an interface change.

If we decide a new arg should be added, I'd like to make sure the type is correct since the impact is broad. I am currently thinking InferenceServiceStatus might work?

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -79,6 +88,33 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) error {
...
+			p.Log.Info("Transformer reconciliation is waiting for predictor URL to be populated")
+			return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
+		}
+
+		// add predictor host and protocol to metadata
+		predictorURL, err := url.Parse(isvc.Status.Components["predictor"].URL.String())
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("unable to parse predictor URL: %v", err)
+		}
+
```
njhill: can simplify:
```suggestion
		predictorURL := (*url.URL)(isvc.Status.Components["predictor"].URL)
		if predictorURL == nil {
			// tansformer reconcile will retry every 3 second until predictor URL is populated
			p.Log.Info("Transformer reconciliation is waiting for predictor URL to be populated")
			return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
		}

		// add predictor host and protocol to metadata
```

njhill (commented):
Thanks @chinhuang007 updates look good! I just made one more small simplification suggestion.

pkg/controller/v1beta1/inferenceservice/components/transformer.go
njhill: Thanks @chinhuang007, looks like you already made the change.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -168,7 +175,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		reconcilers = append(reconcilers, components.NewExplainer(r.Client, r.Scheme, isvcConfig))
 	}
 	for _, reconciler := range reconcilers {
-		if err := reconciler.Reconcile(isvc); err != nil {
+		if _, err := reconciler.Reconcile(isvc); err != nil {
```
pvaneck: Question: is the Result with the `RequeueAfter` propagated out? Seems like we would just ignore the returned Requeue Result here.

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -79,6 +88,34 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) error {
 		}),
 		Annotations: annotations,
 	}
+
+	// Need to wait for predictor URL in modelmesh deployment mode
+	if isvcutils.GetDeploymentMode(annotations, deployConfig) == constants.ModelMeshDeployment {
+		// check if predictor URL is populated
+		predictorURL := (*url.URL)(isvc.Status.Components["predictor"].URL)
+		if predictorURL == nil {
+			// tansformer reconcile will retry every 3 second until predictor URL is populated
```
pvaneck: nit: transformer typo

pkg/controller/v1beta1/inferenceservice/components/transformer.go
chinhuang007: @njhill Good point! Changed as suggested.

pkg/controller/v1beta1/inferenceservice/controller.go
chinhuang007: Good catch! Add code to propagate requeue result to isvc reconcile.

pvaneck (commented):
/lgtm

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -79,6 +88,29 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) error {
...
+		// check if predictor URL is populated
+		predictorURL := (*url.URL)(isvc.Status.Components["predictor"].URL)
+		if predictorURL == nil {
+			// transformer reconcile will retry every 3 second until predictor URL is populated
+			p.Log.Info("Transformer reconciliation is waiting for predictor URL to be populated")
+			return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
+		}
+
+		// add predictor host and protocol to metadata
+		isvc.ObjectMeta.Annotations["predictor-host"] = predictorURL.Host
```
yuzisun, line 103: @chinhuang007 Can you help move this to a constant and rename following the internal annotation naming convention
https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L80

njhill (approved):

nishank-lily:
Can you also update this document with an example of a transformer working with model mesh?
https://kserve.github.io/website/0.9/modelserving/mms/modelmesh/overview/#kserve-integration
This still says transformers do not yet work with ModelMesh.

chinhuang007:
Sure, will update the doc.

PR #2147, opened 2022-04-18
Add Volumes to ServingRuntimePodSpec; allow other built-in ServerTypes

njhill:

- Added `Volumes` to `ServingRuntimePodSpec`
- Remove enum constraint to allow other built-in `serverType`s in `ServingRuntime` `builtInAdapter` (so that CRD change isn't required if more built-in server types are added)
- Added unit test to cover merging volumes
- Improved existing environment variable merging (replace rather than add if same name is encountered)

**Type of changes**

Unit test included

**Checklist**:

Signed-off-by: Nick Hill <nickhill@us.ibm.com>


pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -112,14 +103,57 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
...
 
+// MergeRuntimeVolumes Merge the predictor Volumes list with the runtime Volumes list, allowing users
+// to override/augment runtime volume settings from the predictor spec.
+func MergeRuntimeVolumes(runtimeVolumes []v1.Volume, predictorVolumes []v1.Volume) []v1.Volume {
+	if len(runtimeVolumes) == 0 {
+		return predictorVolumes
+	} else if len(predictorVolumes) == 0 {
+		return runtimeVolumes
+	}
+	vols := append(([]v1.Volume)(nil), runtimeVolumes...)
```
tjohnson31415: If you want to ensure a single allocation, can extend the slice with capacity in case each predictor entry is added:
```suggestion
	vols := make([]v1.Volume, len(runtimeVolumes), len(runtimeVolumes)+len(predictorVolumes))
	copy(vols, runtimeVolumes)
```

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -112,14 +103,57 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
...
+
+// MergeEnvs Merge the predictor Env list with the runtime Env list, allowing users
+// to override/augment environment variables from the predictor spec.
+func MergeEnvs(runtimeEnv []v1.EnvVar, predictorEnv []v1.EnvVar) []v1.EnvVar {
+	if len(runtimeEnv) == 0 {
+		return predictorEnv
+	} else if len(predictorEnv) == 0 {
+		return runtimeEnv
+	}
+	env := append(([]v1.EnvVar)(nil), runtimeEnv...)
```
tjohnson31415: If you want to ensure a single allocation, can extend the slice with capacity in case each predictor entry is added:
```suggestion
	env := make([]v1.Volume, len(runtimeEnv), len(runtimeEnv)+len(predictorEnv))
	copy(env, runtimeEnv)
```

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -131,10 +138,9 @@ type ServingRuntimeStatus struct {
 
 // ServerType constant for specifying the runtime name
 // +k8s:openapi-gen=true
-// +kubebuilder:validation:Enum=triton;mlserver
 type ServerType string
 
-// ServerType Enum
+// Built-in ServerTypes (others may be supported)
```
tjohnson31415, line 143: Should update the godoc comment above `ServerType` in the `BuiltInAdapter` struct below to indicate that it can have other values.

pkg/apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -131,10 +138,9 @@ type ServingRuntimeStatus struct {
...
 const (
 	// Model server is Triton
 	Triton ServerType = "triton"
```
tjohnson31415, line 146: If there are values used elsewhere, should they be added to this list?
In particular, I'm thinking about the addition of the `"ovms"` server type that would be added in https://github.com/kserve/modelmesh-serving/pull/141.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -155,6 +155,7 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) error {
 		podSpec.Containers = []v1.Container{
 			*container,
 		}
+		podSpec.Volumes = isvcutils.MergeRuntimeVolumes(sRuntime.Volumes, isvc.Spec.Predictor.Volumes)
```
pvaneck: Would it make sense to have this call happen within `MergePodSpec` since this is a PodSpec field (similar to how `MergeEnvs` is called from within `MergeRuntimeContainers`)?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -77,9 +77,7 @@ func GetDeploymentMode(annotations map[string]string, deployConfig *v1beta1api.D
 func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *v1.Container) (*v1.Container, error) {
 	// Default container configuration from the runtime.
 	coreContainer := v1.Container{
```
pvaneck: Not necessarily part of this PR, but just now noticing that since the runtimeContainer type is now v1.Container (instead of the subset of fields via v1alpha1.Container), if a user uses a field not listed here in a ServingRuntime, it would be dropped.

Perhaps can just update here to something like `coreContainer := runtimeContainer.DeepCopy()` or  maybe even use `runtimeContainer` directly?

pvaneck (commented):
Thanks, @njhill, for the spec update! Left some comments inline.

pvaneck:
Oh, and don't forget to run `make generate`. (Forgot about this, too :blush:)

pkg/controller/v1beta1/inferenceservice/utils/utils.go
njhill: I guess I wasn't too concerned about the allocation here given how infrequently this will run and likelihood that the slices are small, so opted to keep concise.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
njhill: Same answer :)

pkg/apis/serving/v1alpha1/servingruntime_types.go
njhill: The list isn't meant to necessarily be exhaustive. I wasn't sure about adding `ovms` to the interface since it's still kind of beta, but I suppose this won't show up in the actual CRD so no harm to add it here.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
njhill: Sure, I think I only added it here to match how `MergeRuntimeContainers` was done. I could move that in there too?

njhill: Hmm I guess the situation is a bit more complicated since there's a `Containers` list in the inline predictor PodSpec and a separate single `Container` in the `PredictorExtensionSpec`. I assume we're only considering the latter for consistency with how the other types of non-custom predictor work but there isn't really any why you shouldn't be able to add/override multiple containers when using a `ServingRuntime`. The logic right now kind of assumes that there's only one.

Perhaps we should at least look for one in the SR with the same name and merge that - and if the name is different we could just add it rather than merging with the first one in the list?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
njhill: I updated this in the latest commit.

pvaneck (approved):
/lgtm

njhill:
@pvaneck I updated this to do a proper strategic merge patch, could you take another quick look?

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
```diff
@@ -1223,8 +1223,6 @@ func TestMergeRuntimeContainers(t *testing.T) {
 				Name:  "kserve-container",
 				Image: "default-image",
 				Args: []string{
-					"--foo=bar",
```
pvaneck: Looks like arg entries are just overwritten with the new strategic merge which I think is why the e2e test is failing [this](https://github.com/kserve/kserve/blob/master/test/e2e/batcher/test_batcher_custom_port.py#L48) test. Here, an arg is provided with the assumption that the other default args provided by the SR are also passed in (i.e.  `--model_name={{.Name}}` and `--model_dir=/mnt/models`). Should we try to maintain this behavior

pvaneck (commented):
@njhill Thanks, good to know about this strategicpatch package. Left a comment inline about a change in behavior.

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
njhill: Thanks @pvaneck, this is a tricky one. I actually wasn't sure how best to handle args so thought we may as well follow the Kubernetes merge rules for that field (replace rather than combine). 

The trouble is that it's kind of situation dependent... i.e. the syntax/semantics of a particular model server's args and the reason that they are being set in the SR. Even in the case of the referenced test failure, the goal is to set a different http port and it will end up adding two such parameters rather than replacing the [existing one](https://github.com/kserve/kserve/blob/89d56de823db57fa40d2458fd0d8178e7a578f28/manifests/charts/templates/clusterservingruntimes.yaml#L127).

But I can see that here that out of the two options, appending is more likely to be the desired behaviour, so I'll make an update to restore that.

pvaneck (approved):
Thanks, @njhill.

/lgtm

yuzisun:
Thanks @njhill !

/approve

PR #2159, opened 2022-04-27
Update Infer proto docs to mention BFloat16 type

rmccorm4:

Adds mention of [BFloat16 (BF16)](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) type to the Infer API, similarly handled as FP16. Triton Inference Server is adding `TYPE_BF16` support and we want to keep parity with the KServe docs: https://github.com/triton-inference-server/common/pull/49


deadeyegoodwin:
@yuzisun This looks good to me.

PR #2181, opened 2022-05-14
Move annotation keys to constants

chinhuang007:
Move annotation keys to constants so that we don't have
strings in different code segments. Just a tiny cosmetic
change, no impact to functionality.

Signed-off-by: Chin Huang <chhuang@us.ibm.com>

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -95,6 +95,8 @@ var (
 	AgentModelConfigVolumeNameAnnotationKey          = InferenceServiceInternalAnnotationsPrefix + "/configVolumeName"
 	AgentModelConfigMountPathAnnotationKey           = InferenceServiceInternalAnnotationsPrefix + "/configMountPath"
 	AgentModelDirAnnotationKey                       = InferenceServiceInternalAnnotationsPrefix + "/modelDir"
+	PredictorHostAnnotationKey                       = InferenceServiceInternalAnnotationsPrefix + "/predictor-host"
+	PredictorProtocolAnnotationKey                   = InferenceServiceInternalAnnotationsPrefix + "/predictor-protocol"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/constants/constants.go
```diff
@@ -95,6 +95,8 @@ var (
 	AgentModelConfigVolumeNameAnnotationKey          = InferenceServiceInternalAnnotationsPrefix + "/configVolumeName"
 	AgentModelConfigMountPathAnnotationKey           = InferenceServiceInternalAnnotationsPrefix + "/configMountPath"
 	AgentModelDirAnnotationKey                       = InferenceServiceInternalAnnotationsPrefix + "/modelDir"
+	PredictorHostAnnotationKey                       = "predictor-host"
```
yuzisun: Can you add the `InferenceServiceInternalAnnotationsPrefix` prefix ?

pkg/controller/v1beta1/inferenceservice/components/transformer.go
```diff
@@ -100,12 +100,12 @@ func (p *Transformer) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, er
 		}
 
 		// add predictor host and protocol to metadata
-		isvc.ObjectMeta.Annotations["predictor-host"] = predictorURL.Host
+		isvc.ObjectMeta.Annotations[constants.PredictorHostAnnotationKey] = predictorURL.Host
```
yuzisun, line 103: @chinhuang007 It is also needed to be changed here 
https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/transformer_custom.go#L85

yuzisun:
Thanks @chinhuang007 !
/lgtm
/approve

PR #2206, opened 2022-05-29
chore: Update E2E tests to use GH Actions

### LINKED ISSUE #2087: kserve e2e testing and optional-test-infra deprecation
We are using Kubeflow's optional-test-infra for the e2e tests.

The Kubeflow test team plans to deprecate that infrastructure, reference kubeflow/testing/issues/993

We've discussed in the past some options for the e2e tests, like stay in AWS and use Tekton, or move to Github Actions.

One of the reasons to stay in AWS was to have a test environment which is close to what people are using in production, essentially the ephemeral EKS cluster with the ALB capabilities.

On the other hand, looking at knative/serving's [e2e tests](https://github.com/knative/serving/blob/main/.github/workflows/kind-e2e.yaml) we are inspired by the powerful combination of different setups using Github Actions matrix and KinD.

Let's have a small discussion about it and eventually make a decision to move on to a future-proof option. Personally I like the later option with Github actions and KinD. @animeshsingh @pvaneck @yuzisun @chinhuang007

pvaneck:

We are no longer able to run end-to-end tests on AWS infra anymore, so this PR shifts the tests to be able to run on GitHub Actions runners. Some changes included are:

- A GitHub Actions workflow was added for e2e tests. This workflow will setup minikube, and call scripts to install needed dependencies and run the tests.
- KServe config `test` overlay image tags were updated to remove the dependency on ECR.
- Overlays for istio and knative were added to the test directory for adjusting the resource requests for all Istio/Knative containers. Reduction is needed to work on the constrained GH Actions environments.
- Resources deployed in tests also had their resources adjusted to accommodate the new test environment. 
- Some test scripts for unit testing and linting were removed as they are obsolete or unused.
- Old test scripts mean for AWS infra were moved to `test/scripts/aws` while new test scripts for GH Actions were moved to `test/scripts/gh-actions`.

**Note:**
-  Currently only the images for the controller and storage initializer are being built and used in the tests, while runtime/agent/explainer/image-transformer images are still being pulled from Docker Hub. The rest are commented out in `build-images.sh`, but building every single image will add another 20-30 minutes to workflow. Will maybe want to implement selective building and tag replacement based on modified code.
- Pytest parallelism was removed as this wasn't working on the constrained environment. So each test is run serially for now. 
- Tests that were disabled were:
  - explainer tests (need to figure out correct resource settings for these as was getting OOM)
  - paddle tests (was hanging, maybe some issue pulling the model file)
  - triton tests (the triton image is pretty big and was hitting disk issues)

Fixes #2087


pvaneck:
@yuzisun Can we remove the kubeflow-kserve-presubmit required check?

yuzisun:
> @yuzisun Can we remove the kubeflow-kserve-presubmit required check?

@pvaneck Thanks for the awesome work!! 

Currently only the images for the controller and storage initializer are being built and used in the tests, while runtime/agent/explainer/image-transformer images are still being pulled from Docker Hub. The rest are commented out in build-images.sh, but building every single image will add another 20-30 minutes to workflow. Will maybe want to implement selective building and tag replacement based on modified code.
>> Is it possible to parallelize the image builds like we did in the argo workflow? 

Pytest parallelism was removed as this wasn't working on the constrained environment. So each test is run serially for now.
Tests that were disabled were:
>> maybe we can create multiple test clusters and partition the e2e tests running on them, for example raw deployment and serverless tests can run on different clusters.

explainer tests (need to figure out correct resource settings for these as was getting OOM)
paddle tests (was hanging, maybe some issue pulling the model file)
triton tests (the triton image is pretty big and was hitting disk issues)

>> For triton tests when I run locally I had to increase the docker disk size, not sure if this is possible in github action environment.

This is a great start! @andyi2it @Suresh-Nakkeran can help the above to enhance and speed up the CI time.

test/scripts/gh-actions/build-images.sh
```diff
@@ -0,0 +1,73 @@
...
+# sed -i -e "s/latest/${GITHUB_SHA}/g" config/overlays/test/configmap/inferenceservice.yaml
+
+# Update agent tag
+# sed -i -e "s/kserve\/agent:latest/kserve\/agent:${GITHUB_SHA}/g" config/overlays/test/configmap/inferenceservice.yaml
+
+# Update storage init tag
+sed -i -e "s/kserve\/storage-initializer:latest/kserve\/storage-initializer:${GITHUB_SHA}/g" config/overlays/test/configmap/inferenceservice.yaml
+
+# Update runtimes
+# sed -i -e "s/latest/${GITHUB_SHA}/g" config/overlays/test/runtimes/kustomization.yaml
```
yuzisun: I think we still need this to test the model server images built from PR?

test/e2e/predictor/test_triton.py
```diff
@@ -34,20 +34,20 @@ def test_triton():
...
             ),
         )
     )
     transformer = V1beta1TransformerSpec(
         min_replicas=1,
         containers=[V1Container(
-                      image='809251082950.dkr.ecr.us-west-2.amazonaws.com/kserve/image-transformer:'
-                            + os.environ.get("PULL_BASE_SHA"),
+                      image='kserve/image-transformer:'
+                            + 'latest',  # os.environ.get("GITHUB_SHA"),
```
yuzisun: here should use `GITHUB_SHA` if image-transformer image is built right?

test/scripts/gh-actions/setup-deps.sh
```diff
@@ -0,0 +1,86 @@
...
+mkdir istio_tmp
+pushd istio_tmp >/dev/null
+  curl -L https://istio.io/downloadIstio | ISTIO_VERSION=${ISTIO_VERSION} sh -
+  cd istio-${ISTIO_VERSION}
+  export PATH=$PWD/bin:$PATH
+  istioctl manifest generate > ${SCRIPT_DIR}/../../overlays/istio/generated-manifest.yaml
+popd
+
+kubectl create ns istio-system
+for i in 1 2 3 ; do kustomize build test/overlays/istio | kubectl apply -f - && break || sleep 15; done
```
yuzisun, line 50: this is fine, we can also use istio helm chart to customize the resources which might be easier.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.3'
+      - name: Setup Minikube
+        run: |
+          wget --no-verbose https://github.com/kubernetes/minikube/releases/download/v1.21.0/minikube-linux-amd64
+          sudo cp minikube-linux-amd64 /usr/local/bin/minikube
+          sudo chmod 755 /usr/local/bin/minikube
+          sudo apt-get install -y conntrack socat
+          minikube start --driver=none --kubernetes-version v1.21.0
```
yuzisun, line 27: how does minikube exposes the load balancer?

test/scripts/gh-actions/build-images.sh
```diff
@@ -0,0 +1,73 @@
...
+  # docker build -t ${LGB_IMG} -f lgb.Dockerfile .
+  # docker build -t ${PMML_IMG} -f pmml.Dockerfile .
+  # docker build -t ${PADDLE_IMG} -f paddle.Dockerfile .
+
+  # docker build -t ${ALIBI_IMG} -f alibiexplainer.Dockerfile .
+  # docker build -t ${AIX_IMG} -f aixexplainer.Dockerfile .
+
+  # docker build -t ${IMAGE_TRANSFORMER_IMG} -f custom_transformer.Dockerfile .
+
+  docker build -t ${STORAGE_INIT_IMG} -f storage-initializer.Dockerfile .
```
yuzisun, line 57: how does kserve pulls the locally built image without pushing to registry?

yuzisun: oh I see https://minikube.sigs.k8s.io/docs/drivers/none/, so minikube is running inside the github action VM.

test/scripts/gh-actions/build-images.sh
```diff
@@ -0,0 +1,73 @@
...
+popd
+
+# Update KServe configurations to use the correct tag.
+# sed -i -e "s/latest/${GITHUB_SHA}/g" config/overlays/test/configmap/inferenceservice.yaml
+
+# Update agent tag
+# sed -i -e "s/kserve\/agent:latest/kserve\/agent:${GITHUB_SHA}/g" config/overlays/test/configmap/inferenceservice.yaml
```
yuzisun, line 65: any reason to comment this out?

test/e2e/predictor/test_triton.py
pvaneck: Right, just commented out here for now since we are not currently building it. But will renable it.

test/scripts/gh-actions/build-images.sh
pvaneck: Only reason was to speed up my testing of the workflow since building all the images does add a quite a bit of time. I can re-enable all the image builds/tag replacements to see how it goes. 

I think to get the images built in parallel, we can use parallel jobs within the workflow. But we'd need to push to some external container registry (similar how to ecr was used previously) since all the environments won't be shared.

andyi2it: How about github's container registry? That way, we can build the images in parallel and add wait hooks for the e2e jobs.
We can probably run the e2e as 2 different workflows as well... It may reasonably reduce the time...

May be I can do that once this is merged...

.github/workflows/e2e-test.yml
johnugeorge: Kind cluster is lightweight when compared to minikube. This can be explored as well.

pvaneck: Yea, kind is also a decent option I haven't yet explored in the context of GH Actions, though with minikube's bare-metal (none) driver, the VM/container creation aspect of minikube is eliminated which lends itself to be pretty lightweight itself.

Also, not certain on the load balancer aspect, however it does appear that the tests are able to successfully perform inference on the deployed Isvcs using the istio-ingressgateway:

```
Name:                     istio-ingressgateway
Namespace:                istio-system
Labels:                   app=istio-ingressgateway
                          install.operator.istio.io/owning-resource=unknown
                          istio=ingressgateway
                          istio.io/rev=default
                          operator.istio.io/component=IngressGateways
                          release=istio
Annotations:              <none>
Selector:                 app=istio-ingressgateway,istio=ingressgateway
Type:                     LoadBalancer
IP Family Policy:         SingleStack
IP Families:              IPv4
IP:                       10.108.148.193
IPs:                      10.108.148.193
Port:                     status-port  15021/TCP
TargetPort:               15021/TCP
NodePort:                 status-port  32358/TCP
Endpoints:                172.17.0.4:15021
Port:                     http2  80/TCP
TargetPort:               8080/TCP
NodePort:                 http2  32531/TCP
Endpoints:                172.17.0.4:8080
Port:                     https  443/TCP
TargetPort:               8443/TCP
NodePort:                 https  32[511](https://github.com/kserve/kserve/runs/6683774002?check_suite_focus=true#step:6:512)/TCP
Endpoints:                172.17.0.4:8443
Session Affinity:         None
External Traffic Policy:  Cluster
Events:                   <none>
```

yuzisun:
Thanks @pvaneck !!

/lgtm
/approve

PR #2208, opened 2022-05-31
[WIP] Coerce trained categorical columns on input of lgbserver

johnpaulett:

Aims to workaround microsoft/LightGBM#5244: when a LightGBM model is trained with categorical columns (`dtype == 'category'`), the `Booster.predict()` must also get the same columns as a `category` dtype. Unfortunately at present, kserve simply takes the input JSON data and converts it to a DataFrame (e.g. so string inputs will remain an `object` dtype, not `category`.  When a client tries to call the :predict endpoint, they will get the error
```
{'error': 'train and valid dataset categorical_feature do not match.'}
```

While I personally think this fix ultimately belongs inside the LightGBM's `Booster.predict()`, since it in theory is aware of the categorical columns it was trained on, I am opening this PR in kserve to a) solicit feedback from the kserve developers, since some conversation on microsoft/LightGBM#5244 thought it did not belong in LightGBM, b) as a workaround that I am using until any fix is upstreamed by LightGBM.

I have also investigated Seldon's MLServer, with it's `content_type`, but the [pandas codec](https://github.com/SeldonIO/MLServer/blob/master/mlserver/codecs/pandas.py) appears to just wrap the [numpy dtypes](https://github.com/SeldonIO/MLServer/blob/master/mlserver/codecs/numpy.py#L11) (of which `category` is not one).

**Type of changes**

Manual testing. I plan to contribute a custom categorical model to the example_models, but wanted to float this general PR idea first to the kserve development team

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/lgbserver/lgbserver/model.py
+++ b/python/lgbserver/lgbserver/model.py
@@ -16,7 +16,7 @@ import kserve
 import lightgbm as lgb
 from lightgbm import Booster
 import os
-from typing import Dict
+from typing import Dict, List, Optional
 import pandas as pd
 from kserve.model import ModelMissingError, InferenceError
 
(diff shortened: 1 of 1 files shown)


python/lgbserver/lgbserver/model.py
```diff
@@ -42,16 +42,48 @@ def load(self) -> bool:
...
         self.ready = True
         return self.ready
 
+    def _load_categorical(
+        self, booster: Booster, model_file: str
+    ) -> Optional[List[str]]:
+        # LightGBM does not currently force trained categorical columns to
+        # categorical during predict, so pull the `categorical_feature` from the
+        # saved model
+        # https://github.com/microsoft/LightGBM/issues/5244
```
yuzisun, line 55: there is another implementation, not sure if it is better
https://github.com/microsoft/LightGBM/issues/5244#issuecomment-1270739681

yuzisun:
Closing staled PR

PR #2223, opened 2022-06-07
update graph sample

Iamlovingit:
Signed-off-by: iamlovingit <bitfrog@163.com>

Update `inference graph` sample:
**Change log:**
* update diagrams of `sequence node` and `graph.png`
* update `swtich.yaml`
* update `docs/samples/graph/README.md`


docs/samples/graph/README.md
```diff
@@ -100,27 +104,48 @@ kubectl apply -f switch.yaml
...
+            "strval": "kserve"
+        },
+        {
+            "name": "red",
+            "intval":1,
+            "strval": "1red-server1"
+        }]
+}
+```
+From the `switch conditions` and `input data`, the matched condition is `"instances.#(strval%*red-server*)"`, so we can get the `response` from `single-3`. BTW, if there are multi matched `conditions`, `inference graph` will pickup the first matched service in order, and if the is no `conditon` matched, `inference graph` will return the `input data` directly.
```
yuzisun:
```suggestion
By evaluating the `switch conditions` with `input data`, the condition `"instances.#(strval%*red-server*)"` is matched, so you get the `response` from target service `single-3`. If there are multiple matched `conditions`, `inference graph` pickups the first matched condition in order and routes to the corresponding target service, and if there is no `condition` matched, `inference graph` returns the `input data` back directly. 
```

docs/samples/graph/switch.yaml
```diff
@@ -10,21 +10,35 @@ spec:
...
+---
+apiVersion: "serving.kserve.io/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "single-3"
+spec:
+  predictor:
+    minReplicas: 1
+    containers:
+    - image: iamlovingit/bgtest:latest
```
yuzisun, line 37: Can you also commit the example code so we can recreate the image if needed

Iamlovingit: Sure

docs/samples/graph/bgtest/bgtest/main.go
```diff
@@ -0,0 +1,41 @@
...
+	Instances []Instance `json:"instances"`
+}
+
+func main() {
+	defaultURL := fmt.Sprintf("/v1/models/%s:predict", os.Getenv("target"))
+	r := gin.Default()
+	r.POST("/splitter", handle)
+	r.POST("/switch", handle)
+	r.POST("/single", handle)
+	r.POST("/ensemble", handle)
```
yuzisun, line 29: seems like they are all using the same handle, should we just use the default url which is following the protocol ?

PR #2229, opened 2022-06-08
manifest for release v0.9.0-rc0

Suresh-Nakkeran:

**Release note**:
```
manifest for release v0.9.0-rc0

```


yuzisun:
@Suresh-Nakkeran I think the [versions](https://github.com/kserve/kserve/blob/master/charts/kserve/values.yaml) here needs to be updated to `0.9.0-rc0`, we can actually parameterize it with the release value defined [here](https://github.com/kserve/kserve/blob/master/charts/kserve/Chart.yaml#L3) using `{.Chart.Version}`.

hack/generate-install.sh
```diff
@@ -38,6 +38,7 @@ RELEASES=(
     "v0.7.0"
     "v0.8.0-rc0"
     "v0.8.0"
+    "v0.9.0-rc0"
```
yuzisun, line 41: We added new inference graph crd so need to copy that to the chart

```
cp config/crd/serving.kserve.io_inferencegraphs.yaml charts/kserve/crds/serving.kserve.io_inferencegraphs.yaml
```

PR #2264, opened 2022-06-22
update md for human-readable

drinktee:
update some urls in required_api.md for human-readable.


yuzisun:
thanks @drinktee ! it was fixed in https://github.com/kserve/kserve/pull/2286/files

PR #2285, opened 2022-06-29
chore: Disable storage spec type validation

pvaneck:
Due to differences in storage provider support between the KServe storage initializer and ModelMesh puller, this commit
disables the storage spec type validation. With this enabled, Isvcs on ModelMesh trying to use the new storage spec with azure or gcs providers will be blocked by KServe's validation webhook. This check could potentially be re-enabled once gs/azure storage spec support is provided in the storage initializer.

Overall, I think we should be fine without this validation.

### DIFF
--- a/pkg/apis/serving/v1beta1/predictor.go
+++ b/pkg/apis/serving/v1beta1/predictor.go
@@ -139,7 +139,9 @@ func (s *PredictorSpec) GetExtensions() *ComponentExtensionSpec {
 func (p *PredictorExtensionSpec) Validate() error {
 	return utils.FirstNonNilError([]error{
 		validateStorageURI(p.GetStorageUri()),
-		validateStorageSpec(p.GetStorageSpec(), p.GetStorageUri()),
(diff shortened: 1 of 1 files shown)


Tomcli:
@pvaneck Are you able to add a feature flag to disable the validation for modelmesh development/between releases? This way we can validate the spec if the users intended to use it with KServe with itself.

yuzisun:
> @pvaneck Are you able to add a feature flag to disable the validation for modelmesh development/between releases? This way we can validate the spec if the users intended to use it with KServe with itself.

@Tomcli how much work is left to support existing storage options with the new storage spec for kserve ?

Tomcli:
> > @pvaneck Are you able to add a feature flag to disable the validation for modelmesh development/between releases? This way we can validate the spec if the users intended to use it with KServe with itself.
> 
> @Tomcli how much work is left to support existing storage options with the new storage spec for kserve ?

I was thinking to wait until the modelmesh pullman package to support all storage first so we can simply migrate with the pullman package. This way we don't have to create temporary mapping from the storageSpec to the python storage init-container code.

PR #2315, opened 2022-07-05
chore: Update ModelMesh helm charts

pvaneck:
These updates coincide with MM 0.9 release.

```release-note
Helm charts updated for ModelMesh 0.9
```

### DIFF
--- a/charts/kserve/templates/configmap.yaml
+++ b/charts/kserve/templates/configmap.yaml
@@ -38,7 +38,7 @@ data:
       limits:
         cpu: "2"
         memory: "512Mi"
-    serviceAccountName: modelmesh
+    serviceAccountName: ""
     metrics:
       enabled: true
 kind: ConfigMap
@@ -203,4 +203,4 @@ data:
 kind: ConfigMap
 metadata:
   name: inferenceservice-config
-  namespace: {{ .Release.Namespace }} 
+  namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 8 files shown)


charts/kserve/templates/configmap.yaml
```diff
@@ -38,7 +38,7 @@ data:
       limits:
         cpu: "2"
```
yuzisun, line 39: Any reason why the default cpu limit for puller and mm sidecar need 2-3 CPU, this adds quite a lot overhead for each modelmesh replica.

charts/kserve/templates/configmap.yaml
```diff
@@ -38,7 +38,7 @@ data:
...
         memory: "512Mi"
-    serviceAccountName: modelmesh
+    serviceAccountName: ""
```
yuzisun, line 41: is this service account used by modelmesh controller, any reason to be empty?

charts/kserve/templates/configmap.yaml
pvaneck: Not sure. This was carried over from the defaults used internally. This can certainly be lowered, but this should be lowered in the modelmesh-serving repo first.

charts/kserve/templates/configmap.yaml
pvaneck: This was the service account used by runtime pods and was mainly for internal usage for passing in private registry  secrets to the runtime. Not needed by default.

yuzisun:
Thanks @pvaneck !

/lgtm
/approve

PR #2342, opened 2022-07-18
user defined model name overwrite issue fix

### LINKED ISSUE #2332: Kserve defaulting causing duplicates of environment variable
/kind bug

Create example xgboost isvc and enable gRPC
```
kind: "InferenceService"
metadata:
  name: "xgboost-iris"
spec:
  predictor:
    xgboost:
      protocolVersion: "v2"
      storageUri: "gs://kfserving-examples/models/xgboost/iris"
      ports:
      - containerPort: 9000
        name: h2c
        protocol: TCP
```

The pod spec has duplicated environment variable
```
    Environment:
      MLSERVER_MODEL_NAME:            xgboost-iris
      MLSERVER_MODEL_URI:             /mnt/models
      MLSERVER_MODEL_NAME:            xgboost-iris
      MLSERVER_MODEL_URI:             /mnt/models
```

Additionally, attempt to override the defaults leads to duplicated environment variable with different values

```
kind: "InferenceService"
metadata:
  name: "xgboost-iris"
spec:
  predictor:
    xgboost:
      protocolVersion: "v2"
      storageUri: "gs://kfserving-examples/models/xgboost/iris"
      ports:
      - containerPort: 9000
        name: h2c
        protocol: TCP
      env:
      - name: MLSERVER_MODEL_NAME
        value: my-model
```

The pod spec:
```
    Environment:
      MLSERVER_MODEL_NAME:            my-model
      MLSERVER_MODEL_NAME:            xgboost-iris
      MLSERVER_MODEL_URI:                /mnt/models
```

- Defaulting should not duplicate environment variable and should prioritise user's defined environment variable

I believe it's because the defaulter always append `.Env` without checking the presence of existing environment variable. (https://github.com/kserve/kserve/blob/a6ed8e4b006e27433de2336e0e8b7cec11137dc1/pkg/apis/serving/v1beta1/inference_service_defaults.go#L264)

**Environment:**

- Kserve: 0.8.0

Suresh-Nakkeran:

- This PR is to address the user defined envs were overwritten by the default model server env issue. If the user provided custom model name as env variable that would also reflected in the inference service url.
- Fix IsMMSPredictor function as it incorrectly marks custom model server as MMS

fixes #2332 #2568 

**Type of changes**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -378,21 +378,14 @@ func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService, disableIs
 	if url, err := apis.ParseURL(serviceUrl); err == nil {
 		isvc.Status.URL = url
 		path := ""
+		modelName := isvcutils.GetModelName(isvc)
 		if isvc.Spec.Transformer != nil {
 			// As of now transformer only supports protocol V1
-			path = constants.PredictPath(isvc.Name, constants.ProtocolV1)
+			path = constants.PredictPath(modelName, constants.ProtocolV1)
 		} else if !isvcutils.IsMMSPredictor(&isvc.Spec.Predictor) {
 
 			protocol := isvc.Spec.Predictor.GetImplementation().GetProtocol()
-			modelName := isvc.Name
-			// Set model name from env variable for mlserver
-			if isvc.Spec.Predictor.Model != nil {
-				for _, env := range isvc.Spec.Predictor.Model.Env {
-					if env.Name == constants.MLServerModelNameEnv {
-						modelName = env.Value
-					}
-				}
-			}
+
 			if protocol == constants.ProtocolV1 {
 				path = constants.PredictPath(modelName, constants.ProtocolV1)
 			} else if protocol == constants.ProtocolV2 {
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -384,10 +384,19 @@ func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService, disableIs
 		} else if !isvcutils.IsMMSPredictor(&isvc.Spec.Predictor) {
 
 			protocol := isvc.Spec.Predictor.GetImplementation().GetProtocol()
+			modelName := isvc.Name
+			// Set model name from env variable for mlserver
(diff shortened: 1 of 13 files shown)


pkg/utils/utils.go
```diff
@@ -111,3 +111,19 @@ func IsPrefixSupported(input string, prefixes []string) bool {
...
 }
+
+func AppendEnvVarIfNotExists(slice []v1.EnvVar, elems ...v1.EnvVar) []v1.EnvVar {
+	for _, item := range elems {
+		appendEnvVarIfNotExists(slice, item)
+	}
+	return slice
+}
+
+func appendEnvVarIfNotExists(slice []v1.EnvVar, elem v1.EnvVar) []v1.EnvVar {
```
yuzisun: seems like it is cleaner to combine the two functions?

Suresh-Nakkeran: updated!

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -373,10 +373,19 @@ func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService) error {
 		}
 		if !isvcutils.IsMMSPredictor(&isvc.Spec.Predictor, isvcConfig) {
 			protocol := isvc.Spec.Predictor.GetImplementation().GetProtocol()
+			modelName := isvc.Name
+			// Set model name from env variable for mlserver
+			if isvc.Spec.Predictor.Model != nil {
```
yuzisun: Hmm actually each model server has ways to overwrite the model name, not sure how to handle this systematically or maybe we can treat this field as default model url

pradithya: To add additional context to https://github.com/kserve/kserve/issues/2332. I'd like to have different value for isvc.name and the mlserver model name. So I don't think we should add ability to override isvc.name using env. 

The reason to have different value is for having different host and the model_name value in the request payload when calling a model. As of now, it must always be the same. 
E.g if I have isvc.name = "my-model". It will generate host `my-model.<namespace>.<base_url>` and the model payload should contains "model_name" = "my-model" field.  In my case I would like to keep the expected "model_name" in the payload to be the same for several isvc by overriding `MODEL_NAME` env var so that I could introduce some traffic shaping in front of these models.

yuzisun: @pradithya here it is overriding the model name in the url path not the host, the host is still using the isvc name, I think that’s what you expected.

pradithya: Yes correct, that's my intention.

PR #2348, opened 2022-07-21
fix: HA deployment issue

Suresh-Nakkeran:

This PR fixes the issue in the HA deployment that leader election was not enabled on kserve controller manager.

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -47,3 +47,4 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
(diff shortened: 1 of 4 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -47,3 +47,4 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
+        - "--leader-elect"
```
yuzisun, line 50: Should be —leader-elect=true ?

Suresh-Nakkeran: @yuzisun It's boolean, true is optional.

Issue #2257, opened 2022-06-19
Is path-based serving supported?


kandrio:
/kind feature

Hello everyone!

I see that the current way to serve `KServe` models is host-based. That is, the URL of a model looks something like this:
- `http://<model-name>-<namespace>.<domain>`.

This way, one needs to create a DNS record and a certificate for the sub-domain `<model-name>-<namespace>` where the model is hosted.

So, my question is: _is there an alternative way? A **path-based way**?_.

In that case, one should not bother about DNS records and certificates. The URL of the model could look something like this:
- `http://<domain>/<namespace>/<model-name>`

I couldn't find any issues with a similar question. Is something like this currently supported?

If not, then what are the difficulties of such a feature?

padrian2s:
in standalone mode (serverless), the routes are coming from knative,  

https://knative.dev/docs/serving/services/private-services/

yes, the wildcard root-based certificates are not easy to get.

yuzisun:
@kandrio There is no difficulties and I created [the issue](https://github.com/kserve/kserve/issues/824) 2 years ago but did not find person signing for this contribution.

kandrio:
@yuzisun Thanks for pointing me to the issue. It was really helpful and it got me in sync with the proposal.

I believe that supporting path-based serving would be a great feature for KServe. And I'm interested in taking up the implementation.

Let me expose an initial draft that I have in mind.

## Design

At the end of the day, the model should be exposed under a URL that looks like this:
- `status.url`: `<Scheme>://<IngressDomain>/<Namespace>/<Name>`

Here's a list of steps that I think would make sense for  supporting path-based serving:
1. Extend the `inferenceservice-config` `ConfigMap` with a `urlTemplate` field.
    * This field essentially will be a template for the `inferenceservice.status.url` field.
    * It will allow the user to enable path-based serving by setting: `"urlTemplate": "{{ .Scheme }}://{{ .IngressDomain }}/{{ .Namespace }}/{{ .Name }}"`
    * This will have default value: `""`. In that case, the KServe controller works as before.
2. Extend the KServe controller to add an extra rule in the top-level `VirtualService` that it creates for each model.
    * The KServe Controller will add the extra rule in the case that the `"urlTemplate"` is path-based.
    * This extra rule will route all requests that reach the public `Gateway`, with a `/<namespace>/<model-name>` prefix (path-based requests), to the local `Gateway`.
[...]

Issue #824, opened 2020-05-14
Make KFServing accessible from GCP IAP Ingress


yuzisun:
/kind feature

**Describe the solution you'd like**
This is a follow up for https://github.com/kubeflow/kfserving/issues/760

in GCP/IAP setup, KFServing inference service is protected behind IAP and in order to access from outside, we need additional virtual service rule to work with path based routing because currently KFServing only supports host based routing. So we need to add additional rule in KFServing top level virtual service to rewrite url path and convert to host based routing. 

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: kfserving-kubeflow-yuzi-dan-kfserving-test
  namespace: kubeflow-yuzi-dan
spec:
  gateways:
  - kubeflow/kubeflow-gateway
  hosts:
  - '*'
  http:
  - match:
    - uri:
        prefix: /kfserving/kubeflow-yuzi-dan/sklearn-iris
    rewrite:
        uri: /v1/models/sklearn-iris:predict
    route:
    - destination:
        host: cluster-local-gateway.istio-system.svc.cluster.local
      headers:
        request:
          set:
            Host: sklearn-iris-predictor-default.kubeflow-yuzi-dan.svc.cluster.local
      weight: 100
    timeout: 300s
```

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

kaivalya97:
@yuzisun I'd like to work on this. Can you explain a bit more and suggest a starting point?

owennewo:
@kaivalya97 - I'm in a similar position to you.  I'd also like to help.  It might be easier to discuss no kubeflow kfserving slack channel.

yuzisun:
@kaivalya97 @owennewo Currently KFServing creates a top level virtual service for routing so we could add additional path based routing rules or a separate virtual service based on a configuration. You can join our WG meeting to discuss this and take a look at the code [here](https://github.com/kubeflow/kfserving/blob/master/pkg/controller/inferenceservice/resources/istio/virtualservice.go#L182) to get some ideas.

yuzisun:
@kaivalya97 here is doc for GCP/IAP https://github.com/kubeflow/kfserving/tree/master/docs/samples/gcp-iap and explained what changes are needed, let me know if you need more information.

jal06:
@yuzisun , I have 2 questions about the gcp-iap sample 

[...]

PR #2202, opened 2022-05-27
Allow InferenceService url scheme to be configurable

markwinter:
**What this PR does / why we need it**:

This allows the URL scheme to be configurable in the `inferenceservice-config` configmap.

We have a wildcard certificate in the istio gateway to allow https traffic. We would like this to be reflected in the isvc url so users know they can use https.

**Type of changes**
Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
add `urlScheme` to the inference service configmap which by default is set to http
```

PR #2357, opened 2022-07-27
Feature enable ingress for path based routing

kandrio:

This PR introduces path-based serving. It is related to the following issues:
- https://github.com/kserve/kserve/issues/824
- https://github.com/kserve/kserve/issues/2257

**Type of changes**

I added tests in the source code for every change I introduced.

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/serving/v1beta1/configmap.go
+++ b/pkg/apis/serving/v1beta1/configmap.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"text/template"
 
 	"github.com/kserve/kserve/pkg/constants"
 	v1 "k8s.io/api/core/v1"
@@ -74,6 +75,7 @@ type IngressConfig struct {
 	DomainTemplate          string  `json:"domainTemplate,omitempty"`
 	UrlScheme               string  `json:"urlScheme,omitempty"`
 	DisableIstioVirtualHost bool    `json:"disableIstioVirtualHost,omitempty"`
+	PathTemplate            string  `json:"pathTemplate,omitempty"`
 }
 
 // +kubebuilder:object:generate=false
@@ -113,6 +115,19 @@ func NewIngressConfig(cli client.Client) (*IngressConfig, error) {
 
 		if ingressConfig.IngressGateway == "" || ingressConfig.IngressServiceName == "" {
 			return nil, fmt.Errorf("invalid ingress config - ingressGateway and ingressService are required")
+		}
+		if ingressConfig.PathTemplate != "" {
+			// TODO: ensure that the generated path is valid, that is:
+			// * both Name and Namespace are used to avoid collisions
+			// * starts with a /
+			// For now simply check that this is a valid template.
+			_, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
+			if err != nil {
+				return nil, fmt.Errorf("Invalid ingress config, unable to parse pathTemplate: %v", err)
+			}
+			if ingressConfig.IngressDomain == "" {
+				return nil, fmt.Errorf("Invalid ingress config - igressDomain is required if pathTemplate is given")
+			}
 		}
 	}
 
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -89,7 +89,9 @@ func getServiceHost(isvc *v1beta1.InferenceService) string {
 	}
 }
 
-func getServiceUrl(isvc *v1beta1.InferenceService, urlScheme string, disableIstioVirtualHost bool) string {
+func getServiceUrl(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig) string {
+	urlScheme := config.UrlScheme
+	disableIstioVirtualHost := config.DisableIstioVirtualHost
 	if isvc.Status.Components == nil {
 		return ""
 	}
@@ -331,9 +333,10 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
 	return desiredIngress
 }
 
-func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService, disableIstioVirtualHost bool) error {
+func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService) error {
 	serviceHost := getServiceHost(isvc)
-	serviceUrl := getServiceUrl(isvc, ir.ingressConfig.UrlScheme, disableIstioVirtualHost)
+	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
+	disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost
 	if serviceHost == "" || serviceUrl == "" {
 		return nil
 	}
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -90,6 +90,34 @@ func getServiceHost(isvc *v1beta1.InferenceService) string {
 }
 
 func getServiceUrl(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig) string {
+
+	url := getHostBasedServiceUrl(isvc, config)
+	if url == "" {
+		return ""
+	}
+	if config.PathTemplate == "" {
+		return url
+	} else {
+		return getPathBasedServiceUrl(isvc, config)
+	}
+
+}
+
+func getPathBasedServiceUrl(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig) string {
+	path, err := GenerateUrlPath(isvc.Name, isvc.Namespace, config)
+	if err != nil {
+		log.Error(err, "Failed to generate URL path from pathTemplate")
+		return ""
+	}
+	url := &apis.URL{}
+	url.Scheme = config.UrlScheme
+	url.Host = config.IngressDomain
+	url.Path = path
+
+	return url.String()
+}
+
+func getHostBasedServiceUrl(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig) string {
 	urlScheme := config.UrlScheme
 	disableIstioVirtualHost := config.DisableIstioVirtualHost
 	if isvc.Status.Components == nil {
@@ -312,6 +340,62 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
 	if !isInternal {
 		hosts = append(hosts, serviceHost)
 		gateways = append(gateways, config.IngressGateway)
+	}
+	if config.PathTemplate != "" {
+		path, err := GenerateUrlPath(isvc.Name, isvc.Namespace, config)
+		if err != nil {
+			log.Error(err, "Failed to generate URL from pathTemplate")
+			return nil
+		}
+		url := &apis.URL{}
+		url.Path = strings.TrimSuffix(path, "/") // remove trailing "/" if present
+		url.Host = config.IngressDomain
+		// In this case, we have a path-based URL so we add a path-based rule
+		httpRoutes = append(httpRoutes, &istiov1alpha3.HTTPRoute{
+			Match: []*istiov1alpha3.HTTPMatchRequest{
+				{
+					Uri: &istiov1alpha3.StringMatch{
+						MatchType: &istiov1alpha3.StringMatch_Prefix{
+							Prefix: url.Path + "/",
+						},
+					},
+					Authority: &istiov1alpha3.StringMatch{
+						MatchType: &istiov1alpha3.StringMatch_Regex{
+							Regex: constants.HostRegExp(url.Host),
+						},
+					},
+					Gateways: []string{config.IngressGateway},
+				},
+				{
+					Uri: &istiov1alpha3.StringMatch{
+						MatchType: &istiov1alpha3.StringMatch_Exact{
+							Exact: url.Path,
+						},
+					},
+					Authority: &istiov1alpha3.StringMatch{
+						MatchType: &istiov1alpha3.StringMatch_Regex{
+							Regex: constants.HostRegExp(url.Host),
+						},
+					},
+					Gateways: []string{config.IngressGateway},
+				},
+			},
+			Rewrite: &istiov1alpha3.HTTPRewrite{
+				Uri: "/",
+			},
+			Route: []*istiov1alpha3.HTTPRouteDestination{
+				createHTTPRouteDestination(backend, isvc.Namespace, config.LocalGatewayServiceName),
+			},
+			Headers: &istiov1alpha3.Headers{
+				Request: &istiov1alpha3.Headers_HeaderOperations{
+					Set: map[string]string{
+						"Host": network.GetServiceHostname(backend, isvc.Namespace),
+					},
+				},
+			},
+		})
+		// Include ingressDomain to the domains (both internal and external) derived by KNative
+		hosts = append(hosts, url.Host)
 	}
 
 	annotations := utils.Filter(isvc.Annotations, func(key string) bool {
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -160,6 +160,16 @@ jobs:
           ./test/scripts/gh-actions/run-e2e-tests.sh "fast or pmml or graph"
 
           kubectl get pods -n kserve
+      - name: Patch inferenceservice config
+        run : |
+          kubectl patch configmaps -n kserve inferenceservice-config --patch-file config/overlays/test/configmap/inferenceservice-ingress.yaml
+          kubectl describe configmaps -n kserve inferenceservice-config
+      - name: Run E2E tests with path-based routing
+        timeout-minutes: 40
+        run: |
+          ./test/scripts/gh-actions/run-e2e-tests.sh fast
+
+          kubectl get pods -n kserve
       - name: Check system status
         if: always()
         run: |
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -180,7 +180,7 @@ func main() {
 		Scheme: mgr.GetScheme(),
 		Recorder: eventBroadcaster.NewRecorder(
 			mgr.GetScheme(), v1.EventSource{Component: "v1beta1Controllers"}),
-	}).SetupWithManager(mgr, deployConfig, ingressConfig.DisableIstioVirtualHost); err != nil {
+	}).SetupWithManager(mgr, deployConfig, ingressConfig); err != nil {
 		setupLog.Error(err, "unable to create controller", "v1beta1Controller", "InferenceService")
 		os.Exit(1)
 	}
--- a/config/overlays/test/configmap/inferenceservice-ingress.yaml
+++ b/config/overlays/test/configmap/inferenceservice-ingress.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: inferenceservice-config
+  namespace: kserve
+data:
+  ingress: |-
+    {
+        "ingressGateway" : "knative-serving/knative-ingress-gateway",
+        "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
+        "localGateway": "knative-serving/knative-local-gateway",
+        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
+        "ingressClassName" : "istio",
+        "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
+        "urlScheme": "http",
+        "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}",
+        "ingressDomain": "mydomain.com"
+    }
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -205,7 +205,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	} else {
 		reconciler := ingress.NewIngressReconciler(r.Client, r.Scheme, ingressConfig)
 		r.Log.Info("Reconciling ingress for inference service", "isvc", isvc.Name)
-		if err := reconciler.Reconcile(isvc, ingressConfig.DisableIstioVirtualHost); err != nil {
+		if err := reconciler.Reconcile(isvc); err != nil {
 			return reconcile.Result{}, errors.Wrapf(err, "fails to reconcile ingress")
 		}
 	}
@@ -274,13 +274,13 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1api.InferenceServiceStatus, deplo
 	return equality.Semantic.DeepEqual(s1, s2)
 }
 
-func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployConfig *v1beta1api.DeployConfig, disableIstioVirtualHost bool) error {
+func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployConfig *v1beta1api.DeployConfig, ingressConfig *v1beta1api.IngressConfig) error {
 	if deployConfig.DefaultDeploymentMode == string(constants.RawDeployment) {
 		return ctrl.NewControllerManagedBy(mgr).
 			For(&v1beta1api.InferenceService{}).
 			Owns(&appsv1.Deployment{}).
 			Complete(r)
-	} else if disableIstioVirtualHost == false {
+	} else if ingressConfig.DisableIstioVirtualHost == false {
 		return ctrl.NewControllerManagedBy(mgr).
 			For(&v1beta1api.InferenceService{}).
 			Owns(&knservingv1.Service{}).
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler_test.go
@@ -685,3 +685,96 @@ func TestGetServiceUrl(t *testing.T) {
 		})
 	}
 }
+
+func TestGetServiceUrlPathBased(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+	serviceName := "my-model"
+	namespace := "test"
+	isvcAnnotations := map[string]string{"test": "test", "kubectl.kubernetes.io/last-applied-configuration": "test"}
+	labels := map[string]string{"test": "test"}
+	predictorUrl, _ := url.Parse("http://my-model-predictor-default.example.com")
+	ingressConfig := &v1beta1.IngressConfig{
+		UrlScheme:               "http",
+		IngressDomain:		 "my-domain.com",
+		PathTemplate:            "/serving/{{ .Namespace }}/{{ .Name }}",
+		DisableIstioVirtualHost: false,
+	}
+
+	cases := map[string]struct {
+		isvc    *v1beta1.InferenceService
+		matcher gomegaTypes.GomegaMatcher
+	}{
+		"component is empty": {
+			isvc: &v1beta1.InferenceService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        serviceName,
+					Namespace:   namespace,
+					Annotations: isvcAnnotations,
+					Labels:      labels,
+				},
+				Spec: v1beta1.InferenceServiceSpec{
+					Predictor: v1beta1.PredictorSpec{},
+				},
+			},
+			matcher: gomega.Equal(""),
+		},
+		"predictor url is empty": {
+			isvc: &v1beta1.InferenceService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        serviceName,
+					Namespace:   namespace,
+					Annotations: isvcAnnotations,
+					Labels:      labels,
+				},
+				Spec: v1beta1.InferenceServiceSpec{
+					Predictor: v1beta1.PredictorSpec{
+						SKLearn: &v1beta1.SKLearnSpec{},
+					},
+				},
+				Status: v1beta1.InferenceServiceStatus{
+					Status:  duckv1.Status{},
+					Address: nil,
+					URL:     nil,
+					Components: map[v1beta1.ComponentType]v1beta1.ComponentStatusSpec{
+						v1beta1.PredictorComponent: v1beta1.ComponentStatusSpec{},
+					},
+					ModelStatus: v1beta1.ModelStatus{},
+				},
+			},
+			matcher: gomega.Equal(""),
+		},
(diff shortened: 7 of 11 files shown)


yuzisun:
@kandrio Can you help rebase the code from master?

dimara:
@yuzisun I will continue the work that @kandrio started. I will do the rebase ASAP and ping here once done! Thanks!

dimara:
@yuzisun I rebased of the original PR and resolved the conflicts. Are there any other blockers?

dimara:
@yuzisun any new on this? cc @adriangonz @Iamlovingit

lizzzcai:
Hi @dimara, thanks for your follow-up work on this PR. I added my use case under the previous [issue](https://github.com/kserve/kserve/issues/2257#issuecomment-1223614792). It looks you have a separated `UrlTemplate` for the path-based url, have you consider reuse the existing `DomainTemplate`? (by working with `pathTemplate` to generate the Url).

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/url.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"text/template"
+
+	"net/url"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"knative.dev/pkg/apis"
+)
+
+type UrlTemplateValues struct {
```
lizzzcai: Is it ok to support values from `annotations` and `labels` as well, similar to [DomainTemplate](https://github.com/kserve/kserve/blob/39ec2d19de1d856a9527b800fda8fb380bc9441f/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go#L30-L36).

dimara:
Hi @lizzzcai! First of all thanks for the feedback.

> It looks you have a separated UrlTemplate for the path-based url

Well I rebased the initial PR without making any important changes. I think your comment came after the PR was submitted. My bad I guess for not taking it into account before doing the rebase. Sorry for that.

> have you consider reuse the existing DomainTemplate? (by working with pathTemplate to generate the Url).

Isn't ``DomainTemplate`` used only in *raw* deployment, i.e., without KNative? Note that with the current implementation, the host-based routing that KNative sets up (based on its own config-network and config-domain) still works and as such we can reach the underlying KNative Services directly. With `urlTemplate` KServe will configure an extra/alternative "route" without messing up the existing configuration.

So based on the above, do we need to change the initial implementation or keep it as is? The alternative would be to make KServe take into account all the `domainTemplate`, `urlScheme`, `ingressDomain`, `pathTemplate` settings regardless the deployment type, right?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/url.go
dimara: ACK! Let's decide whether we go with urlTemplate and if we do then I will add a fixup on top. Thanks!

dimara:
@lizzzcai @yuzisun any news on this? How can we move forward?

lizzzcai:
Hi @dimara , sorry for the late reply. I don't use the raw deployment mode but from what I see the `DomainTemplate` uses the `ingressDomain` to define its domain. 

In stead of using `ingressDomain` directly in this PR, maybe the user can define a `DomainTemplate: {{ .IngressDomain }}`, which same as the domain in knative config-domain, and use it with the `pathTemplate` to generate the full path based url `{{.DomainTemplate}}/{{.pathTemplate}}`. 

In this case, we are able to generalize the implementation on both raw deployment mode and serverless mode, and support path-based routing in raw deployment mode in the future as well.

However, from what I see the current host is getting from [knative](https://github.com/kserve/kserve/blob/3c93df286a053ea4d913201a31d972aa9058330f/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go#L314) directly. So setting value to `DomainTemplate` can be `optional` (set to knative servicehost if it is empty), as you only check if UrlTempalte (pathTemplate) [exist](https://github.com/kserve/kserve/blob/3c93df286a053ea4d913201a31d972aa9058330f/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go#L323) to enable path-based routing.

what do you think?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -311,6 +320,88 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
 		hosts = append(hosts, serviceHost)
 		gateways = append(gateways, config.IngressGateway)
 	}
+	if config.UrlTemplate != "" {
```
yuzisun: Can we add the test cases for the generated vs matching rules?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -125,6 +125,15 @@ func getServiceUrl(isvc *v1beta1.InferenceService, urlScheme string, disableIsti
 	}
 }
 
+func contains(s []string, e string) bool {
```
yuzisun: We can move this to the utils file

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -311,6 +320,88 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
...
+				Headers: &istiov1alpha3.Headers{
+					Request: &istiov1alpha3.Headers_HeaderOperations{
+						Set: map[string]string{
+							"Host": network.GetServiceHostname(backend, isvc.Namespace),
+						},
+					},
+				},
+			})
+		} else if !contains(hosts, url.Host) {
+			// In this case, we have a host-based URL, so we add a host-based rule
```
yuzisun: I am confused about this case, in which scenario the `url.Host` is not in the `hosts` list?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/url_test.go
```diff
@@ -0,0 +1,119 @@
...
+				ingressConfig: &v1beta1.IngressConfig{
+					IngressDomain: "my.domain",
+				},
+			},
+			want: "",
+		},
+		{
+			name: "path based url template",
+			args: args{
+				name:      "model",
```
yuzisun: is this the service name or model name? There could be multiple models in a given InferenceService

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -311,6 +320,88 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
...
+						Authority: &istiov1alpha3.StringMatch{
+							MatchType: &istiov1alpha3.StringMatch_Regex{
+								Regex: constants.HostRegExp(url.Host),
+							},
+						},
+						Gateways: []string{config.IngressGateway},
+					},
+				},
+				Rewrite: &istiov1alpha3.HTTPRewrite{
+					Uri: "/",
```
yuzisun, line 362: I am not sure how this is going to work, for example if the `UrlTemplate` is `"https://{{ .IngressDomain }}/{{ .Namespace }}/{{ .Name }}` and the user is calling with the following url `http://istio-ingressgateway.istio-system/user/isvc/v2/models/cifar/infer`, `url.path` would return `user/isvc` and rewrite to `/`.

yuzisun: Does this only replace the matched prefix portion?

yuzisun:
Thanks @kandrio for the great work and apologize for the late review.

yuzisun:
@kandrio @lizzzcai Great discussions! My two cents is that we probably should not set `DomainTemplate` and `IngressDomain` in isvc configuration for serverless mode as they are only supposed to be used with `RawDeployment`, otherwise we may get into inconsistent state between Knative and KServe. In serverless mode we take the host directly from Knative so for path based routing we only need the `PathTemplate` part really, we can use this to decide whether to inject path based routing rules or not. On the other hand `RawDeployment` is implemented with Kubernetes ingress resource which does not support url rewrite, we might need to consider using [Gateway API](https://gateway-api.sigs.k8s.io/) in future to normalize the implementation between `Serverless` and `RawDeployment` mode.

dimara:
> However, from what I see the current host is getting from knative directly.

Yes. So `domainTemplate` etc. are currently ignored in the default deployment type (serverless / with KNative). 

> So setting value to DomainTemplate can be optional.

We must think of existing deploymens here, since `domainTemplate` currently has a default value `"{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}"`. And users may not have set this at all. So it is not clear when to use the KNative domain or use the one generated based on KServe configuration.

What we can do is have `pathTemplate` optional (empty by default, same as we do in this PR), and if this is set then it will use the `domainTemplate`, `ingressDomain, and `urlScheme`. So if a user wants to use path-based serving, which is not the default, they will have to specify something like:
```
        "ingressDomain"  : "mydomain.com",
        "domainTemplate": "{{ .IngressDomain }}",
        "urlScheme": "https",
        "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}"
```
If `pathTemplate` is not set, then the serverless deployment type will continue to work as it does now, i.e., obtain the URL from Knative.

Another option would be to have a new separate boolean config, e.g., `generateURL` or something, that will be `false` by default. And only if the user sets this then all the above values will be taken into account. This way all existing options can be used for both raw and KNative deployments.

I think both options above are not that intrusive and do not need specific upgrade notes. @lizzzcai what do you think?

lizzzcai:
Hi @dimara , thanks for your update.

> What we can do is have pathTemplate optional (empty by default, same as we do in this PR), and if this is set then it will use the domainTemplate, ingressDomain, and urlScheme`. 

from my understanding of the previous discussion,  `domainTemplate` will not be used in serverless mode even it is set.

I agreed with @yuzisun  idea. Just add `PathTemplate` should be enough and it will only work in serverless mode (need to document that it will not work for raw deployment mode), and inject it if the value is set. `PathTemplate` should support value from annotations and labels as well.

I agreed with the Gateway API idea as well, which can be implemented in the Raw deployment first to support path-based routing if there is any requirement.

dimara:
>  In serverless mode we take the host directly from Knative so for path based routing we only need the PathTemplate part really, we can use this to decide whether to inject path based routing rules or not. 

@yuzisun @lizzzcai I believe we cannot use the host of KNative since this is specific to each isvc (includes both name and namespace). The idea is to keep KNative routing as is and have KServe to add support for path-based routing on top under a single domain. That is, have https://mydomain.com/serving/myns/myisvc point to KNative's internal http://myisvc.myns.svc.cluster.local. I understand configuring KNative's `domainTemplate` and `domain` is orthogonal. This will expose the underlying KNative Services.

For example, with current implementation, if we have KServe with:
```
urlTemplate: https://mydomain.com/serving/{{ .Namespace }}/{{ .Name }}
```
and KNative with default values for config-domain and domainTemplate, KServe will create a toplevel VirutalService like:
```
metadata:
  name: sklearn-iris
  namespace: kubeflow-user
spec:
  gateways:
  - knative-serving/knative-serving-cluster-gateway
  - knative-serving/knative-serving-gateway
  hosts:
  - sklearn-iris.kubeflow-user.svc.cluster.local
  - sklearn-iris-kubeflow-user.example.com
  - mydomain.com
  http:
  ...
   - headers:
      request:
        set:
          Host: sklearn-iris-predictor-default.kubeflow-user.svc.cluster.local
    match:
    - authority:
        regex: ^mydomain\.com(?::\d{1,5})?$
      gateways:
      - knative-serving/knative-serving-gateway
      uri:
        prefix: /serving/kubeflow-user/sklearn-iris/
    - authority:
        regex: ^mydomain\.com(?::\d{1,5})?$
      gateways:
      - knative-serving/knative-serving-gateway
      uri:
        exact: /serving/kubeflow-user/sklearn-iris
    rewrite:
      uri: /
    route:
    - destination:
        host: knative-serving-cluster-ingressgateway.knative-serving.svc.cluster.local
        port:
          number: 80
      weight: 100
```

This means that the user will be able to hit the ISVC with `https://mydomain.com/serving/kubeflow-user/sklearn-iris/v1/models/sklearn-iris:predict` or by port-forwarding knative-serving-ingressgateway and setting the `Host` header to `sklearn-iris-kubeflow-user.example.com`.


> My two cents is that we probably should not set DomainTemplate and IngressDomain in isvc configuration for serverless mode as they are only supposed to be used with RawDeployment
 
This is what argued before as well and that is why we went for a new option.
 
> otherwise we may get into inconsistent state between Knative and KServe.

Well with current implementation KNative routing continues to work as is. All we do is add an **extra** route on top to support path-based routing. As I said above, my understanding is that we cannot use the host of the underlying KNative service since this includes both name and namespace. @yuzisun @lizzzcai do I miss something here?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
dimara: You mean extend `ingress_reconciler_test.go` to test createIngress() for path-based routing scenarios?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
dimara: Well this is a corner case where we have set a host-based url template:
```
urlTemplate: https://{{ .Name }}-{{ .Namespace}}.serving.mydomain.com/
```
This is more or less the case where we configure KServe to expose the ISVC and let KNative with default values (example.com domain).

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/url_test.go
dimara: This is service (isvc) name. 
> There could be multiple models in a given InferenceService

ACK. I believe the value here, `model`, was taken from the existing `domain_test.go`. But if you want we can change it to something else.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
dimara: Yes. For example:
```
/serving/myns/myisvc/v1/models/myisvc:predict
```
will become
```
/v1/models/myisvc:predict
```

Note that we end up having two rules; one with prefix one with exact match:
```
    - ...
      uri:
        prefix: /serving/kubeflow-user/sklearn-iris/
    - ...
      uri:
        exact: /serving/kubeflow-user/sklearn-iris
    rewrite:
      uri: /
```

yuzisun:
@kandrio I think only configuring the domain in kserve and leaving the knative with default domain name will not work as top level of virtual service routes to the vs created by the knative, so the domain needs to be synced between knative and kserve in order for mydomain.com to work which is why we should configure the domain in knative instead of kserve for serverless mode.

yuzisun:
I guess what you are saying is that we can route to the knative provisioned virtual service with in cluster domain so we do not actually need to use the external domain configuration in Knative right ?

dimara:
> so the domain needs to be synced between knative and kserve in order for mydomain.com to work which is why we should configure the domain in knative instead of kserve for serverless mode.

The problem is that you cannot configure KNative to use a single domain (path-based routing). It uses host-based routing for all the VirutalServices it creates.

> route to the knative provisioned virtual service with in cluster domain so we do not actually need to use the external domain configuration in Knative right ?

Right. As described in https://github.com/kserve/kserve/pull/2357#issuecomment-1315129192, `https://mydomain/serving/kubeflow-user/sklearn-iris` will be routed to `knative-serving-cluster-ingressgateway.knative-serving.svc.cluster.local` using Host `sklearn-iris-predictor-default.kubeflow-user.svc.cluster.local`.
Still, you can configure KNative config-domain and domainTemplate, it if you want to have the underlying KSVC (predictor, transformer) be directly accessible. But these will always use host-based routing.

So based on the above, should we proceed with the existing implementation (`urlTemplate`) or use a `pathTemplate` as described in https://github.com/kserve/kserve/pull/2357#issuecomment-1313279838. @yuzisun @lizzzcai what do you think?

lizzzcai:
> The problem is that you cannot configure KNative to use a single domain (path-based routing). It uses host-based routing for all the VirutalServices it creates.
Hi @dimara , if knative and kserve vs you created above in this PR using the same domain, for example:
```
# domain: "{{ .Name }}-{{ .Namespace }}.example.com"
sklearn-iris-default.example.com # knative
sklearn-iris-default.example.com/default/sk-learn-iris # kserve path-based routing, assume that sharing the same domain with knative.
```
In this case, if user send a request to `sklearn-iris-default.example.com/default/sk-learn-iris/v1/model/<name>:predict`, the request will be routed to knative endpoint rather than kserve path based endpoint based on the matching logic?

If this is the case, I will still prefer to use `pathTemplate` with the `domainTemplate` for path-based routing, future we can support in raw deployment as well. `urlTemplate` will make `domainTemplate` a bit redundant.

dimara:
Let me try explain a bit more. Let's assume we create a `sklearn-iris` ISVC in the kubeflow-user namespace. KServe will create:
* `sklearn-iris-predictor-default` ksvc (KNative Service)
* `sklearn-iris` Istio VirtualService
* `sklearn-iris` Service

Until now, the KServe VirtualService was rewriting the request to reach the underlying KNative VirtualService. Specifically:
1. KServe looks at the status of the KNative service that reported a url like http://sklearn-iris-predictor-default-kubeflow-user.example.com
2. removes the `predictor-default` (see https://github.com/kserve/kserve/blob/faffcc2f4e026902012fbf00c28a0b16c5a9a29e/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go#L87)
3. uses the result in the `hosts` of the VirtualService
4. sets the `Host` header to the **internal** host of the KNative Service, that is sklearn-iris-predictor-default.kubeflow-user.svc.cluster.local
5. re-routes the request to the KNative cluster ingressgateway.

This PR changes the Istio VirtualService to support path-based routing, i.e., use **per-isvc separate prefix** under a **single** domain. Using the KNative domain does not help since this is different per isvc. The path-based support comes on-top, that is host-based rules based on KNative Services are also added and host-based routing continues to work.

> I will still prefer to use pathTemplate with the domainTemplate for path-based routing, 

I am OK with that, but as we mentioned above, currently `domainTemplate` is not supposed to work with serverless mode (KNative) but only in raw deployments. If we go with your suggestion we have to go with one of the options described in  https://github.com/kserve/kserve/pull/2357#issuecomment-1313279838 and as a result all existing options will be used for both raw and KNative deployments.

@lizzzcai @yuzisun what do you think? How can we move forward?

yuzisun:
> I am OK with that, but as we mentioned above, currently `domainTemplate` is not supposed to work with serverless mode (KNative) but only in raw deployments. If we go with your suggestion we have to go with one of the options described in [#2357 (comment)](https://github.com/kserve/kserve/pull/2357#issuecomment-1313279838) and as a result all existing options will be used for both raw and KNative deployments.
> 
> @lizzzcai @yuzisun what do you think? How can we move forward?

I think `domain` and `domainTemplate` can still be configured in Knative and if `pathTemplate` exists then we can extract the domain from the predictor host to construct the url. In this way we can keep consistent that in serverless mode `domain` and `domainTemplate` is configured with Knative and in raw deployment it is configured with KServe.

dimara:
> I think domain and domainTemplate can still be configured in Knative 

@yuzisun KNative uses host-based routing. So the domainTemplate will be in the format of `{{ .Name }}-{{ .Namespace }}.example.com` which mean a **different domain per-isvc**. So we cannot have a single domain and use path-based routing as this PR aims for.

If we have a `domainTemplate: example.com` in KNative nothing will work since KNative will end up creating VirtualServices with **conflicting** hosts. For example if I create two isvc, `sklearn-iris` and `sklearn-iris-2`, I will get two KNative services that will create two VirtualServices:
* sklearn-iris-predictor-default-ingress
* sklearn-iris-2-predictor-default-ingress

The first will have:
```
  hosts:
  - example.com
  - sklearn-iris-predictor-default.kubeflow-user
  - sklearn-iris-predictor-default.kubeflow-user.svc.
  - sklearn-iris-predictor-default.kubeflow-user.svc.cluster.local
```
The second will have:
```
  - example.com
  - sklearn-iris-2-predictor-default.kubeflow-user
  - sklearn-iris-2-predictor-default.kubeflow-user.svc.
  - sklearn-iris-2-predictor-default.kubeflow-user.svc.cluster.local
```
And both will have a rule that will match:
```
  - match:
    - authority:
        prefix: example.com
      gateways:
      - knative-serving/knative-serving-gateway
```
but with different route:
```
    - destination:
        host: sklearn-iris-predictor-default-00002.kubeflow-user.svc.cluster.local
        port:
          number: 80
```

All KNative services will report the same URL and if you hit them you will not know where Istio will route you to. I tried that and actually the KSVC do not get ready:
```
$ kubectl get ksvc -A
NAMESPACE       NAME                               URL                  LATESTCREATED                            LATESTREADY                              READY     REASON
kubeflow-user   sklearn-iris-2-predictor-default   http://example.com   sklearn-iris-2-predictor-default-00001   sklearn-iris-2-predictor-default-00001   True      
kubeflow-user   sklearn-iris-predictor-default     http://example.com   sklearn-iris-predictor-default-00001     sklearn-iris-predictor-default-00001     Unknown   Uninitialized
```
And the `net-istio-controller` was complaining:
```
{"severity":"ERROR","timestamp":"2022-11-25T03:01:19.082636575Z","logger":"net-istio-controller.istio-ingress-controller","caller":"status/status.go:403","message":"Probing of http://example.com:80 failed, IP: 172.17.0.38:8080, ready: false, error: unexpected hash: want \"3008c6929ee505f2969ae52d7dee73e9b87aba272e3f48db84c107f43ec3c868\", got \"2f018fb32ed5c854768fd0c6bc5d00fdbaf75e3d665c030567025ce92de18863\
```
These hashes are the K-Network-Hash added by Istio VS. For my KSVC I got:
* sklearn-iris-predictor-default: 3008c6929ee505f2969ae52d7dee73e9b87aba272e3f48db84c107f43ec3c868
* sklearn-iris-2-predictor-default: 2f018fb32ed5c854768fd0c6bc5d00fdbaf75e3d665c030567025ce92de18863

In other words, KNative tries to probe sklearn-iris-predictor-default and got as a response the hash ofr sklearn-iris-2-predictor-default. Once I restored `domainTemplate: '{{.Name}}-{{.Namespace}}.{{.Domain}}'` both KSVC became ready.

So if we want path-based routing under a **single** domain, we have to do that in KServe on top of KNative's host-based routing. This is what this PR does. Am I missing something here?

yuzisun:
@kandrio I do not think we need to modify the `domainTemplate` in Knative, we can extract the domain from the predictor host in inference service status. For example KServe can get the domain `example.com` from `sklearn-iris-user1.default.example.com` using the `domainTemplate` as the parser, all we need to know is the domain. In knative the domain is configured [here ](https://github.com/knative/serving/blob/main/config/core/configmaps/domain.yaml), alternatively we can get this directly by reading the knative domain configmap.

dimara:
@yuzisun first of all sorry for the late reply. Let me go through your comments/suggestions.

> For example KServe can get the domain example.com from sklearn-iris-user1.default.example.com using the domainTemplate as the parser, all we need to know is the domain. 

Oh. Now I get what you meant. I got confused when you said:

>  we can extract the domain from the predictor host to construct the url.

I thought with "extract" you meant take the predictor host as we currently do.

So to summarize your suggestion is to:
1. Introduce a `pathTemplate` config in inferenceservice-config of KServe.
2. This will be used in serverless mode for now.
3. If `pathTemplate` is set, the controller will obtain/compute the domain from KNative and use it to configure the toplevel VirtualService

My objection regarding (3) is that how Knative Services are exposed is kinda orthogonal to how KServe exposes their Inference service via the **top-level** VirtualService. KNative may use a completely different domain (or keep the dummy example.com) or choose not to expose its Services at all (if for example having a wildcard certificate and DNS record for the KNative URLs may not be an option). Plus, making KServe depend 100% on KNative would not allow us to have the two modes (raw, serverless) share the same expose logic in the future.

Leaving my objection aside, I believe there are some implementation issues if we try to use the domain from KNative. Specifically, you suggested to have the controller inspect the KNative configuration:
> we can get this directly by reading the knative domain configmap.

This is bit problematic since you will have to allow KServe controller to read confimaps in other namespaces. Plus, KNative may decide to change the structure of these configmaps and this will break KServe. Apart from that, this configmap is kinda complicated since it allows for specifying different domains based on selectors so parsing it may not be that trivial.

You also suggested:
>  get the domain from the host using the domainTemplate as the parser

While reading domainTemplate from config-network is easier (but with the same RBAC and compatibility issues mentioned above), To use the Go Template in a reverse manner (see 
https://github.com/golang/go/issues/52473), you have to do some tricks like rendering the template without the domain and compare the result with the status.url of the KNative service.


That said, I think we should decouple from how KNative exposes their Services. This PR supports exposing the InferenceService in a manner that the underlying KNative services don't need to be exposed at all. So I would either go with the initial implementation (with urlTemplate) or with what described in https://github.com/kserve/kserve/pull/2357#issuecomment-1313279838 (with pathTemplate). @yuzisun @lizzzcai What do you think?

dimara:
Note that we already have support for setting the `urlScheme` in serverless deployments (see https://github.com/kserve/kserve/pull/2202). This is in-line with our suggestion where we have KServe decide on the final InferenceService URL, regardless the deployment mode. Specifically, we can use the `ingressDomain` along with `pathTemplate` and `urlScheme` to construct the final URL (as suggested in https://github.com/kserve/kserve/pull/2357#issuecomment-1313279838).

yuzisun:
> My objection regarding (3) is that how Knative Services are exposed is kinda orthogonal to how KServe exposes their Inference service via the **top-level** VirtualService. KNative may use a completely different domain (or keep the dummy example.com) or choose not to expose its Services at all (if for example having a wildcard certificate and DNS record for the KNative URLs may not be an option). Plus, making KServe depend 100% on KNative would not allow us to have the two modes (raw, serverless) share the same expose logic in the future.

I am ok with configuring `ingressDomain` in KServe for path based routing, but I think we need to carefully document the configurations as currently in serverless mode the domain has to be configured in Knative when you use host based urls.

yuzisun:
@kandrio thanks for the awesome work! to wrap up the PR I think we still need to add an e2e test for the path based routing.

dimara:
@yuzisun just found time to finally switch to this. I will ping here once I have the PR ready with all the changes we discussed above. 
> we still need to add an e2e test for the path based routing.

OK. I'll try to add one.

dimara:
@yuzisun I am done with the refactor. Here is a summary of what I ended up implementing:

* introduce `pathTemplate` in `ingressConfig`
* if `pathTemplate` is given require `ingressDomain`
* Add `path.go` (similar to `domain.go`) that will be responsible for rendering / validating `pathTemplate` given an isvc. Add unittests for this.
* Have `getServiceUrl()` take into account `pathTemplate`. If this is set then the `isvc.status.url` will end up having a path-based URL.
* Extend `test-fast` e2e test, to also test path-based routing.
* if `pathTemplate` is set then the virtualservice will obtain a new rule for the path (as it did in the initial PR) and a host for `ingressDomain`.
* the new version seems to be simpler and more straightforward than then initial implementation since we don't have to cover all corner cases with a user provided urlTemplate, e.g., one that may be host-based.

I noticed that the `test-fast` and `test-helm` of the e2e tests failed: https://github.com/kserve/kserve/actions/runs/3947181022/

But what failed seems totally irrelevant. For example, the minikube-setup step failed with:
```
Error from server (InternalError): Internal error occurred: failed calling webhook "config.webhook.serving.knative.dev": failed to call webhook: Post "https://webhook.knative-serving.svc:443/config-validation?timeout=10s": dial tcp 10.108.221.58:443: connect: connection refused
783
```
I understand that `kubectl patch cm config-deployment -n knative-serving` failed because the webhook refused the connection even though we wait for this to become ready.  This seems to be a race. For example the same step succeeded in `test-slow` job. Maybe related with istio sidecar running in this node. Can we re-run this?

Any idea why the test-helm job failed? It seems port-forward to the predictor pod failed:
```
E0118 10:02:59.708333   11169 portforward.go:400] an error occurred forwarding 8008 -> 8008: error forwarding port 8008 to pod 3b7d1ad3ce8eccdfc8e66f1db4599ba5a9ded6c9b4527f7393ee8b08013743be, uid : exit status 1: 2023/01/18 10:02:59 socat[13582] E connect(5, AF=2 127.0.0.1:8008, 16): Connection refused
```
This seems to be irrelevant with this PR, right? Since it hits the endpoint directly from inside the predictor. Can we ignore this?

@yuzisun Assuming the e2e tests pass, is there any other blocker for merging this PR? cc @lizzzcai

yuzisun:
Thanks @kandrio for the update! I will go through the PR again, we are making a few fixes for the flaky tests, meanwhile I have rerun the tests for your PR.

dimara:
> Thanks @kandrio for the update!

Just for the record I am @dimara :) @kandrio was the one that initially submitted the PR but he had to drop so we continued his work.

> we are making a few fixes for the flaky tests

ACK.
> meanwhile I have rerun the tests for your PR.

OK. Let me know if you have any comments about the revised PR, then. Thanks!

dimara:
@yuzisun did you have the time to take a look?  I see that you re-submitted the tests and everything (including the test-fast that this PR extends) worked. Let me know if there are any blockers for merging this. Thanks!

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -114,6 +116,19 @@ func NewIngressConfig(cli client.Client) (*IngressConfig, error) {
...
+			// * both Name and Namespace are used to avoid collisions
+			// * starts with a /
+			// For now simply check that this is a valid template.
+			_, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
+			if err != nil {
+				return nil, fmt.Errorf("Invalid ingress config, unable to parse pathTemplate: %v", err)
+			}
+			if ingressConfig.IngressDomain == "" {
+				return nil, fmt.Errorf("Invalid ingress config - igressDomain is required if pathTemplate is given")
+			}
```
yuzisun, line 130: The `inferenceservice-config`  has the default value `example.com` so this check does not take effect , let's add a comment in the configmap that `IngressDomain` is used in case of `RawDeployment ` and path based routing for `Serverless` mode. https://github.com/kserve/kserve/blob/b15cd9e43771aa90ae8f2e0094c4333197af6bd1/config/configmap/inferenceservice.yaml#L62

yuzisun:
@dimara Thanks for the awesome work to implement this long wanted feature!! Sorry for slow reviews.
I have created doc follow up issues
- https://github.com/kserve/kserve/issues/2678
- https://github.com/kserve/website/issues/225

/lgtm
/approve

pkg/apis/serving/v1beta1/configmap.go
dimara: @yuzisun isn't the default value assigned a few lines below if this is not given?

https://github.com/kserve/kserve/blob/c587ef1769f33f6134ba16b6ef1e7f569eb3eda4/pkg/apis/serving/v1beta1/configmap.go#L138

zamonia500:
Does this PR cover gRPC inference too?

PR #2360, opened 2022-07-29
Pass request headers to predict method (#2284)

### LINKED ISSUE #2284: Pass on the headers you receive for Kserve inference request and in Graph request
/kind feature

In the custom model's predict method, when you have body as dict available, similarly it would be great to have headers available for custom model to access

andyi2it:
* Pass inference request headers to model's predict method

Fixes #2284 

**Type of changes**

**Checklist**:

**Release note**:

```release-note
breaking change: the `predict`, `preprocess`, `explain` API needs to add the `headers` argument
```

### DIFF
--- a/python/aixexplainer/aixserver/model.py
+++ b/python/aixexplainer/aixserver/model.py
@@ -11,15 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import asyncio
-import logging
 from typing import Dict
 
+import asyncio
+import logging
+import kserve
 import numpy as np
 from aix360.algorithms.lime import LimeImageExplainer
 from lime.wrappers.scikit_image import SegmentationAlgorithm
-
-import kserve
 
 
 class AIXModel(kserve.Model):  # pylint:disable=c-extension-no-member
--- a/python/aixexplainer/aixserver/model.py
+++ b/python/aixexplainer/aixserver/model.py
@@ -11,14 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Dict
-
 import asyncio
 import logging
-import kserve
+from typing import Dict
+
 import numpy as np
 from aix360.algorithms.lime import LimeImageExplainer
 from lime.wrappers.scikit_image import SegmentationAlgorithm
+
+import kserve
 
 
 class AIXModel(kserve.Model):  # pylint:disable=c-extension-no-member
@@ -48,23 +49,23 @@ class AIXModel(kserve.Model):  # pylint:disable=c-extension-no-member
         resp = loop.run_until_complete(self.predict(scoring_data))
         return np.array(resp["predictions"])
 
-    def explain(self, request: Dict) -> Dict:
-        instances = request["instances"]
+    def explain(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
+        instances = payload["instances"]
         try:
-            top_labels = (int(request["top_labels"])
-                          if "top_labels" in request else
+            top_labels = (int(payload["top_labels"])
+                          if "top_labels" in payload else
                           self.top_labels)
-            segmentation_alg = (request["segmentation_alg"]
-                                if "segmentation_alg" in request else
+            segmentation_alg = (payload["segmentation_alg"]
+                                if "segmentation_alg" in payload else
                                 self.segmentation_alg)
-            num_samples = (int(request["num_samples"])
-                           if "num_samples" in request else
+            num_samples = (int(payload["num_samples"])
+                           if "num_samples" in payload else
                            self.num_samples)
-            positive_only = ((request["positive_only"].lower() == "true") | (request["positive_only"].lower() == "t")
-                             if "positive_only" in request else
+            positive_only = ((payload["positive_only"].lower() == "true") | (payload["positive_only"].lower() == "t")
+                             if "positive_only" in payload else
                              self.positive_only)
-            min_weight = (float(request['min_weight'])
-                          if "min_weight" in request else
+            min_weight = (float(payload['min_weight'])
+                          if "min_weight" in payload else
                           self.min_weight)
         except Exception as err:
             raise Exception("Failed to specify parameters: %s", (err,))
--- a/python/alibiexplainer/alibiexplainer/anchor_images.py
+++ b/python/alibiexplainer/alibiexplainer/anchor_images.py
@@ -11,24 +11,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import kserve
 import logging
-import numpy as np
+from typing import Callable, List, Optional, Dict
+
 import alibi
+import numpy as np
 from alibi.api.interfaces import Explanation
 from alibi.utils.wrappers import ArgmaxTransformer
+
+import kserve
 from alibiexplainer.explainer_wrapper import ExplainerWrapper
-from typing import Callable, List, Optional
 
 logging.basicConfig(level=kserve.constants.KSERVE_LOGLEVEL)
 
 
 class AnchorImages(ExplainerWrapper):
     def __init__(
-        self,
-        predict_fn: Callable,
-        explainer: Optional[alibi.explainers.AnchorImage],
-        **kwargs
+            self,
+            predict_fn: Callable,
+            explainer: Optional[alibi.explainers.AnchorImage],
+            **kwargs
     ):
         if explainer is None:
             raise Exception("Anchor images requires a built explainer")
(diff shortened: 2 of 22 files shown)


python/kserve/kserve/model.py
```diff
@@ -174,16 +174,16 @@ def postprocess(self, response: Union[Dict, ModelInferResponse]) -> Dict:
             return response.get_response(as_json=True)
         return response
 
-    async def _http_predict(self, request: Dict) -> Dict:
+    async def _http_predict(self, request: Dict, headers: Dict[str, str] = None) -> Dict:
         predict_url = PREDICTOR_URL_FORMAT.format(self.predictor_host, self.name)
         if self.protocol == PredictorProtocol.REST_V2.value:
             predict_url = PREDICTOR_V2_URL_FORMAT.format(self.predictor_host, self.name)
-        json_header = {'Content-Type': 'application/json'}
+        # json_header = {'Content-Type': 'application/json'}
```
luranhe: Should this be the default header used when none is provided, to preserve backward compatibility?

python/aixexplainer/aixserver/model.py
```diff
@@ -11,15 +11,16 @@
...
 import asyncio
 import logging
-import kserve
+from typing import Dict
+
 import numpy as np
 from aix360.algorithms.lime import LimeImageExplainer
 from lime.wrappers.scikit_image import SegmentationAlgorithm
 
+import kserve
```
yuzisun: Do we need to import kserve here?

python/kserve/kserve/model.py
```diff
@@ -130,36 +131,36 @@ def load(self) -> bool:
         self.ready = True
         return self.ready
 
-    async def preprocess(self, request: Union[Dict, CloudEvent]) -> Union[Dict, ModelInferRequest]:
+    async def preprocess(self, payload: Union[Dict, CloudEvent]) -> Union[Dict, ModelInferRequest]:
```
yuzisun: I think we also need the request header for preprocess

rachitchauhan43:
@andyi2it : Can you review the comments and make necessary changes ? It would be a useful feature for us to use in case of Inference Graphs.

python/sklearnserver/sklearnserver/model.py
```diff
@@ -46,8 +47,8 @@ def load(self) -> bool:
         self.ready = True
         return self.ready
 
-    def predict(self, request: Dict) -> Dict:
-        instances = request["instances"]
+    def predict(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
```
yuzisun, line 49: changing this is fine, but just want to make sure it can still work without adding the headers to be backwards compatible.

python/kserve/kserve/model.py
```diff
@@ -174,47 +177,55 @@ def postprocess(self, response: Union[Dict, ModelInferResponse]) -> Dict:
...
         return async_result
 
-    async def predict(self, request: Union[Dict, ModelInferRequest]) -> Union[Dict, ModelInferResponse]:
+    async def predict(self, payload: Union[Dict, ModelInferRequest],
+                      headers: Dict[str, str] = None) -> Union[Dict, ModelInferResponse]:
         """
         The predict handler can be overridden to implement the model inference.
         The default implementation makes a call to the predictor if predictor_host is specified
-        :param request: Dict|ModelInferRequest passed from preprocess handler
+        :param headers: Dict|ModelInferenceRequest headers
```
yuzisun:
```suggestion
        :param headers: Dict
```

python/kserve/kserve/model.py
```diff
@@ -174,47 +177,55 @@ def postprocess(self, response: Union[Dict, ModelInferResponse]) -> Dict:
...
-            return await self._http_predict(request)
+            return await self._http_predict(payload, headers)
 
-    async def explain(self, request: Dict) -> Dict:
+    async def explain(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
         """
         The explain handler can be overridden to implement the model explanation.
         The default implementation makes an call to the explainer if explainer_host is specified
-        :param request: Dict passed from preprocess handler
+        :param headers: Dict|ModelInferRequest headers
```
yuzisun:
```suggestion
        :param headers: Dict
```

python/kserve/kserve/model.py
yuzisun: Also need to fix the order of the arguments

python/alibiexplainer/alibiexplainer/anchor_images.py
```diff
@@ -11,32 +11,34 @@
...
 logging.basicConfig(level=kserve.constants.KSERVE_LOGLEVEL)
 
 
 class AnchorImages(ExplainerWrapper):
     def __init__(
-        self,
-        predict_fn: Callable,
-        explainer: Optional[alibi.explainers.AnchorImage],
-        **kwargs
+            self,
```
yuzisun: is the formatting correct here?

python/kserve/kserve/model.py
```diff
@@ -130,36 +131,38 @@ def load(self) -> bool:
...
 
-    async def preprocess(self, request: Union[Dict, CloudEvent]) -> Union[Dict, ModelInferRequest]:
+    async def preprocess(self, payload: Union[Dict, CloudEvent], headers: Dict[str, str] = None) -> Union[
+            Dict, ModelInferRequest]:
         """
         The preprocess handler can be overridden for data or feature transformation.
         The default implementation decodes to Dict if it is a binary CloudEvent
         or gets the data field from a structured CloudEvent.
-        :param request: Dict|CloudEvent|ModelInferRequest
+        :param headers: Dict|ModelInferRequest headers
```
yuzisun:
```suggestion
        :param headers: Dict
```

python/kserve/kserve/model.py
```diff
@@ -174,47 +177,55 @@ def postprocess(self, response: Union[Dict, ModelInferResponse]) -> Dict:
...
+        if not ('Content-Type' in headers):
+            headers['Content-Type'] = 'application/json'
+        headers.pop("Host", None)
         response = await self._http_client.fetch(
             predict_url,
             method='POST',
             request_timeout=self.timeout,
-            headers=json_header,
-            body=json.dumps(request)
+            headers=headers,
```
yuzisun: The headers are from the original request, I think we need to explicitly to copy over the selected headers if needed for the new request.

andyi2it: But what headers should we copy over unless we configure allowed headers somewhere beforehand? Also, wouldn't someone need trace id to track any requests in their system for error logging and debugging until we have distributed tracing properly implemented?
For example any wrong inputs, etc can be traced easily if the trace id goes across transformer, predictor, explainer.

yuzisun: Headers like `X-RequestId` and TraceId should be copied over as they are the same on the request chain, however request specific headers like `Content-Length`, `SpanId`, `Host` should not be propagated. I think explicitly adding the headers required for propagation is easier to reason about than copying by default which can be hard to debug if some header is passed along unexpectedly like the `Host` issue here.

python/kserve/kserve/model.py
```diff
@@ -174,47 +176,57 @@ def postprocess(self, response: Union[Dict, ModelInferResponse]) -> Dict:
...
             predict_url = PREDICTOR_V2_URL_FORMAT.format(self.predictor_host, self.name)
-        json_header = {'Content-Type': 'application/json'}
+
+        # Adjusting headers. Inject content type if not exist.
+        # Also, removing host, as the header is the one passed to transformer and contains transformer's host
+        if headers is None:
+            headers = {}
+        if not ('Content-Type' in headers):
+            headers['Content-Type'] = 'application/json'
+        headers.pop("Host", None)
```
yuzisun:
```suggestion
         predict_headers = {'Content-Type': 'application/json'}
        if "X-Request-Id" in headers:
            predict_headers["X-Request-Id"] = headers["X-Request-Id"]
```

python/kserve/kserve/model.py
```diff
@@ -130,36 +130,38 @@ def load(self) -> bool:
...
 
-    async def preprocess(self, request: Union[Dict, CloudEvent]) -> Union[Dict, ModelInferRequest]:
+    async def preprocess(self, payload: Union[Dict, CloudEvent], headers: Dict[str, str] = None) -> Union[
+            Dict, ModelInferRequest]:
         """
         The preprocess handler can be overridden for data or feature transformation.
         The default implementation decodes to Dict if it is a binary CloudEvent
         or gets the data field from a structured CloudEvent.
-        :param request: Dict|CloudEvent|ModelInferRequest
+        :param headers: Dict
```
yuzisun: Need to fix the param order

yuzisun:
Thanks @andyi2it !

/lgtm
/approve

python/kserve/kserve/model.py
rachitchauhan43: @yuzisun  @andyi2it : We should have some way of configuring the headers that we wish to be propagated. Because I believe most of the companies have their own trace-id in their ecosystem and they would want that to be present along the chain.
By just retaining over the 2 headers  ['X-Request-Id', 'X-B3-Traceid'], we are assuming that no additional headers can be there. 

Isn't it a blocker for adoption by the community? 

For a long term solution, I believe we can have a feature to support configuring headers that users want to propagate.
For short term solution , are we open to extend this [if section](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model.py#L187-L191) to at least include more generic headers like `Authorization` ?

PR #2364, opened 2022-08-01
aix-explainer: update versions to fix CVEs

MessKon:
Fixing critical CVEs in latest kserve/alx-explainer version
Comparing this to latest stable and `v0.9.0` `kserve/aix-explainer` images:
image

image

image

Not sure there's an open issue for this

**Type of changes**

`kserve/alx-explainer` image builds ok with no issues

- Logs

1. This PR only changes image versions

**Release note**:
```release-note
fixing vulnerabilities (CVEs) in kserve/aix-explainer

```


python/aixexplainer/aixserver/model.py
```diff
@@ -17,7 +17,7 @@
 import logging
 import kserve
 import numpy as np
-from aix360.algorithms.lime import LimeImageExplainer
+from aix360.algorithms.lime.lime_wrapper import LimeImageExplainer
```
Tomcli: you don't need `.lime_wrapper` since lime __init__.py exposed `LimeImageExplainer` as a public package.

MessKon: thank you Tommy. I just went for this as the error on the e2e test was hinting that the `aix360.algorithms.lime` module cannot be found. I'll revert the change

python/aixexplainer/setup.py
```diff
@@ -33,7 +33,7 @@
     install_requires=[
         "kserve>=0.7.0",
         "argparse >= 1.4.0",
-        "aix360 >= 0.1.0",
+        "aix360 >= 0.2.0",
```
MessKon, line 36: needed so that we make sure we avoid version `0.1.0` which does not include the `lime` module used [here](https://github.com/kserve/kserve/blob/36888daa9b8139706dd7e226caaac1e2bd4dc567/python/aixexplainer/aixserver/model.py#L20).

python/aixexplainer.Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM python:3.7
+FROM python:3.7-slim
 
 COPY third_party third_party
 
 COPY kserve kserve
 RUN pip install --no-cache-dir --upgrade pip && pip install --no-cache-dir -e ./kserve
 
+RUN apt update && apt install -y build-essential
```
MessKon, line 8: `python:3.7-slim` does not include gcc which is used fro installing few pip dependencies in `aixexplainer`.
we install  `build-essential` and remove it after, as no longer needed

python/aixexplainer.Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM python:3.7
+FROM python:3.7-slim
```
MessKon, line 1: ideally we'd want to use `python:3.8-slim` which:
a. includes only 5 critical CVEs (according to `grype`)
b. includes `gcc` and so there's no need for `apt get` and `apt remove` in this file.

however, `python:3.8` supports `tensorflow` >= 2.2 whilst `aix360` versions 0.2.0 and 0.2.1 [demand tensorflow 1.14](https://github.com/Trusted-AI/AIX360/blob/1d29fe2046bfc2e738a7ba86977b86e8cedbaed8/setup.py#L26).
 in fact the extra critical CVEs reported in `python:3.7-slim` all come from `tensorflow`.
image

For now, there's no other option for fixing CVEs, unless `aix360` update their dependency

Tomcli: Let me open an issue to the aix360 community to update the tensorflow version.

MessKon: thank you Tommy! 
dependabot seems to have already [raised a PR](https://github.com/Trusted-AI/AIX360/pull/150) for it.

Issue #2003, opened 2022-01-21
Storage initializer goes into CrashLoopBackOff, can't copy model from KMS enabled S3 bucket


revolutionisme:
/kind bug

**What steps did you take and what happened:**
Created an inference service using the following:

```
apiVersion: "serving.kserve.io/v1beta1"
kind: "InferenceService"
metadata:
  name: abalone-xgboost
  namespace: kubeflow
spec:
  predictor:
    serviceAccountName: pipeline-runner
    xgboost:
      protocolVersion: "v2"
      storageUri: "s3://bucket-name/abalone-xgboost/model.tar.gz"
```

The pipeline-runner service account consists of a role (IRSA) and is configured as follows:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/kfp-example-pod-role
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{},"labels":{"application-crd-id":"kubeflow-pipelines"},"name":"pipeline-runner","namespace":"kubeflow"}}
  creationTimestamp: "2021-10-01T13:38:27Z"
  labels:
    application-crd-id: kubeflow-pipelines
  name: pipeline-runner
  namespace: kubeflow
  ownerReferences:
  - apiVersion: app.k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: false
    kind: Application
    name: pipeline
    uid: 072e21eb-eaa7-4b3b-a353-4b2342e23477
  resourceVersion: "712344"
  selfLink: /api/v1/namespaces/kubeflow/serviceaccounts/pipeline-runner
  uid: d123b35b-9232-4379-b133-aee3245e34ff2
secrets:
- name: pipeline-runner-token-bn2hf
```

The role `kfp-example-pod-role` has full permisisons to read S3 buckets and objects and even KMS permission.

[A clear and concise description of what the bug is.]
When I apply the inference service defined above, the deployment is in 0/1 state, with the pods in `CrashLoopBackOff` state. When I check the logs of `storage-initializer` inside the pod, the following log is what I get.

```
[...]

PR #2373, opened 2022-08-06
AWS IRSA S3 Support

### LINKED ISSUE #2113: IAM Role for Service Account(IRSA) support for models stored in S3
/kind feature

[Currently](https://kserve.github.io/website/modelserving/storage/s3/s3/), AWS users need to create a secret with long term/static IAM credentials for downloading models stored in S3. Security best practice is to use [IAM role for service account(IRSA)](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) which enables automatic credential rotation and fine grained access control

Support annotating IAM role in a service account which can be used by storage initializer to download models from S3.

It will be super helpful to know the customers/users interested in this feature for appropriate prioritization. Please upvote or comment on the issue if you are interested in this feature

related: https://github.com/awslabs/kubeflow-manifests/issues/44

matty-rose:
At the moment to use AWS IRSA functionality for storage initializer, dummy kubernetes secrets need to be created to get kserve to correctly parse s3 annotations and configure storage init container correctly with environment variables (see https://github.com/kserve/kserve/issues/2003#issuecomment-1053530385) 

This PR allows configuration of S3 storage init three ways now:
1. Configure S3 options in global `inferenceservice` configmap so multiple k8s service accounts/IRSA can be configured without needing to repeat annotations
2. Configure S3 options on individual k8s service accounts/IRSA (which will be used instead of global options where specified)
3. Configure S3 options on k8s secret attached to k8s service account to preserve static credential functionality (AWS access key + secret key)

Fixes #2113 

**Type of changes**

For both tests I followed the [developer guide](https://kserve.github.io/website/master/developer/developer/), using `istioctl` 1.14.3 to deploy Istio, deploying Knative serving 1.6, and then running the `make deploy-dev` command on this branch.

- [x] Test A: Deploy S3 Example using only service account (no secret) and with all possible s3 annotations configured
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: s3
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/s3access
    serving.kserve.io/s3-endpoint: s3.amazonaws.com # replace with your s3 endpoint e.g minio-service.kubeflow:9000
    serving.kserve.io/s3-usehttps: "1" # by default 1, if testing with minio you can set to 0
    serving.kserve.io/s3-region: "us-east-2"
    serving.kserve.io/s3-useanoncredential: "false" # omitting this is the same as false, if true will ignore provided credential and use anonymous credentials
    serving.kserve.io/s3-verifyssl: "0"
    serving.kserve.io/s3-usevirtualbucket: "false"
    serving.kserve.io/s3-cabundle: "test"
---
kind: "InferenceService"
metadata:
  name: "mnist-s3"
spec:
  predictor:
    serviceAccountName: s3
    model:
      modelFormat:
        name: tensorflow
      storageUri: "s3://kserve-examples/mnist"
```
Describing pod with `kubectl describe pod -n default -l serving.kserve.io/inferenceservice=mnist-s3` gives the following configuration on storage-init container
image

- [x] Test B: Deploy S3 Example using only service account (no secret) configuring all values inside `inferenceservice` configmap
```yaml
  credentials: |-
    {
       "gcs": {
           "gcsCredentialFileName": "gcloud-application-credentials.json"
       },
       "s3": {
           "s3AccessKeyIDName": "AWS_ACCESS_KEY_ID",
           "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY",
           "s3Endpoint": "s3.ap-southeast-2.amazonaws.com",
           "s3UseHttps": "0",
           "s3Region": "ap-southeast-2",
           "s3VerifySSL": "0",
           "s3UseVirtualBucket": "true",
           "s3UseAnonymousCredential": "true",
           "s3CABundle": "test"
       }
    }
```

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: s3-configmap
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/s3access
---
kind: "InferenceService"
metadata:
  name: "mnist-s3-configmap"
spec:
  predictor:
    serviceAccountName: s3-configmap
    model:
      modelFormat:
        name: tensorflow
      storageUri: "s3://kserve-examples/mnist"
```
Describing pod with `kubectl describe pod -n default -l serving.kserve.io/inferenceservice=mnist-s3-configmap` gives following configuration on storage init container
image

**Checklist**:

- [ ] Have you made corresponding changes to the documentation?: I will make a separate PR to the kserve website repo updating documentation

**Release note**:

```release-note
Adds AWS S3 configuration + IRSA support to inferenceservice configmap and k8s svcaccount
```

### DIFF
--- a/config/configmap/inferenceservice.yaml
+++ b/config/configmap/inferenceservice.yaml
@@ -143,6 +143,14 @@ data:
         "cpuLimit": "1",
         "storageSpecSecretName": "storage-config"
     }
+  # ====================================== CREDENTIALS ======================================
+  # For a quick reference about AWS ENV variables:
+  # AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
+  # Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables
+  #
+  # The `s3AccessKeyIDName` and `s3SecretAccessKeyName` fields are only used from this configmap when static credentials (IAM User Access Key Secret)
+  # are used as the authentication method for AWS S3. 
+  # The rest of the fields are used in both authentication methods (IAM Role for Service Account & IAM User Access Key Secret) if a non-empty value is provided.
   credentials: |-
     {
        "gcs": {
@@ -150,7 +158,14 @@ data:
        },
        "s3": {
            "s3AccessKeyIDName": "AWS_ACCESS_KEY_ID",
-           "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY"
+           "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY",
+           "s3Endpoint": "",
+           "s3UseHttps": "",
+           "s3Region": "",
+           "s3VerifySSL": "",
+           "s3UseVirtualBucket": "",
+           "s3UseAnonymousCredential": "",
+           "s3CABundle": ""
        }
     }
   ingress: |-
--- a/pkg/credentials/s3/s3_service_account.go
+++ b/pkg/credentials/s3/s3_service_account.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2022 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package s3
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+/*
+For a quick reference about AWS ENV variables:
+AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
+Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables
+*/
+
+func BuildServiceAccountEnvs(serviceAccount *v1.ServiceAccount, s3Config *S3Config) []v1.EnvVar {
+	envs := []v1.EnvVar{}
+
+	envs = append(envs, BuildS3EnvVars(serviceAccount.Annotations, s3Config)...)
+
+	return envs
+}
--- a/pkg/credentials/s3/s3_service_account_test.go
+++ b/pkg/credentials/s3/s3_service_account_test.go
@@ -1,0 +1,159 @@
+/*
+Copyright 2022 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package s3
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestS3ServiceAccount(t *testing.T) {
+	scenarios := map[string]struct {
+		config         S3Config
+		serviceAccount *v1.ServiceAccount
+		expected       []v1.EnvVar
+	}{
+		"NoConfig": {
+			serviceAccount: &v1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "s3-service-account",
+				},
+			},
+			expected: []v1.EnvVar{},
+		},
+
+		"S3Endpoint": {
+			serviceAccount: &v1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "s3-service-account",
+					Annotations: map[string]string{
+						InferenceServiceS3SecretEndpointAnnotation: "s3.aws.com",
+					},
+				},
+			},
+			expected: []v1.EnvVar{
+				{
+					Name:  S3Endpoint,
+					Value: "s3.aws.com",
+				},
+				{
+					Name:  AWSEndpointUrl,
+					Value: "https://s3.aws.com",
+				},
+			},
+		},
+
+		"S3HttpsOverrideEnvs": {
+			serviceAccount: &v1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "s3-service-account",
+					Annotations: map[string]string{
+						InferenceServiceS3SecretEndpointAnnotation: "s3.aws.com",
+						InferenceServiceS3SecretHttpsAnnotation:    "0",
+						InferenceServiceS3SecretSSLAnnotation:      "0",
+					},
+				},
+			},
+			expected: []v1.EnvVar{
+				{
+					Name:  S3UseHttps,
+					Value: "0",
+				},
+				{
+					Name:  S3Endpoint,
+					Value: "s3.aws.com",
+				},
+				{
+					Name:  AWSEndpointUrl,
+					Value: "http://s3.aws.com",
+				},
+				{
+					Name:  S3VerifySSL,
+					Value: "0",
+				},
+			},
+		},
+
+		"S3EnvsWithAnonymousCredentials": {
+			serviceAccount: &v1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "s3-service-account",
+					Annotations: map[string]string{
+						InferenceServiceS3SecretEndpointAnnotation: "s3.aws.com",
+						InferenceServiceS3UseAnonymousCredential:   "true",
+					},
+				},
+			},
+			expected: []v1.EnvVar{
+				{
+					Name:  S3Endpoint,
+					Value: "s3.aws.com",
+				},
+				{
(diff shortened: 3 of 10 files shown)


yuzisun:
Thanks @matty-rose !! This is great!

/cc @surajkota

surajkota:
Thanks for the PR, I will support reviewing this

matty-rose:
Hi @surajkota @yuzisun I think this is ready for review now

config/configmap/inferenceservice.yaml
```diff
@@ -150,7 +150,14 @@ data:
        },
        "s3": {
            "s3AccessKeyIDName": "AWS_ACCESS_KEY_ID",
-           "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY"
+           "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY",
+           "s3Endpoint": "s3.amazonaws.com",
+           "s3UseHttps": "1",
+           "s3Region": "",
```
matty-rose, line 164: I was unsure about having empty values here since I couldn't see any other place in the configmap where this was done, but it makes sense to me to have them here so that users know all the possible configuration options without needing to look through the source code

pkg/credentials/s3/utils.go
```diff
@@ -0,0 +1,105 @@
+package s3
+
+import v1 "k8s.io/api/core/v1"
+
+func BuildS3EnvVars(annotations map[string]string, s3Config *S3Config) []v1.EnvVar {
+	envs := []v1.EnvVar{}
+
+	if s3Endpoint, ok := annotations[InferenceServiceS3SecretEndpointAnnotation]; ok {
```
matty-rose, line 24: this s3 endpoint if/else logic block is unchanged, just moved files from `s3_secret.go` to here since I use it in both `s3_secret.go` and `s3_service_account.go`

pkg/credentials/s3/s3_service_account.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2022 The KServe Authors.
```

pkg/credentials/s3/s3_service_account.go
```diff
@@ -0,0 +1,35 @@
...
+package s3
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+/*
+For a quick reference about AWS ENV variables:
+AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
+Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables
```
yuzisun, line 26: Thanks for adding the reference!

pkg/credentials/s3/s3_service_account_test.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2022 The KServe Authors.
```

pkg/credentials/s3/utils.go
```diff
@@ -0,0 +1,105 @@
+package s3
```
yuzisun, line 17: Let's add the license header on top

pkg/credentials/s3/utils_test.go
```diff
@@ -0,0 +1,84 @@
+package s3
```
yuzisun, line 17: Add the license header

yuzisun:
Thanks @matty-rose for the awesome work with detailed descriptions!! The PR looks great!

Would appreciate if you can help update the website doc for the IAM Role support [here](https://kserve.github.io/website/master/modelserving/storage/s3/s3/#create-s3-secret)

config/configmap/inferenceservice.yaml
yuzisun: Make sense, maybe add a comment with the AWS/S3 reference you had in the other place.

pkg/credentials/s3/utils_test.go
matty-rose: done

pkg/credentials/s3/utils.go
matty-rose: done

config/configmap/inferenceservice.yaml
matty-rose: have added the quick links above the credentials entry in configmap

matty-rose:
@yuzisun have opened a website PR here: https://github.com/kserve/website/pull/178

yuzisun:
@surajkota Do you have comments here?

surajkota:
Apologies, this slipped my list because of a customer request. Will take a look tomorrow, I hope that is okay

config/configmap/inferenceservice.yaml
```diff
@@ -143,14 +143,24 @@ data:
...
+  # For a quick reference about AWS ENV variables:
+  # AWS Cli: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
+  # Boto: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables
   credentials: |-
     {
        "gcs": {
            "gcsCredentialFileName": "gcloud-application-credentials.json"
        },
        "s3": {
            "s3AccessKeyIDName": "AWS_ACCESS_KEY_ID",
```
surajkota, line 160: few questions:
1. there are 2 variables used to construct the service endpoint URL: `s3Endpoint` and `s3UseHttps`. Depending on the value of useHttps the final AWS_ENDPOINT_URL is set. @yuzisun Is there a particular reason to do it this way? Just for my information
2. how is s3VerifySSL used?
3. In _Test B: Deploy S3 Example using only service account (no secret) configuring all values inside inferenceservice configmap_, does the user still need to keep "s3AccessKeyIDName": "AWS_ACCESS_KEY_ID" and "s3SecretAccessKeyName": "AWS_SECRET_ACCESS_KEY", in configmap? It can be a little confusing

surajkota (commented):
Thanks for the patience, great work @matty-rose. I have left a few questions on the PR. And I have a one more:

What is the difference between the S3 download code in the following locations:
1. https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L147 -this is used by the storage initializer
2. https://github.com/kserve/kserve/blob/master/pkg/agent/storage/utils.go#L124-L153 - when does this code get executed?

yuzisun:
> Thanks for the patience, great work @matty-rose. I have left a few questions on the PR. And I have a one more:
> 
> What is the difference between the S3 download code in the following locations:
> 
> 1. https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L147 -this is used by the storage initializer
> 2. https://github.com/kserve/kserve/blob/master/pkg/agent/storage/utils.go#L124-L153 - when does this code get executed?

For 1) It is used for single model serving storage initializer where we inject the init container to download models before starting the model server.
For 2) this is the multi model implementation where you have a long running model agent injected as a sidecar to dynamically update models without tearing down pods.

For historical reasons they are implemented differently, we are planning to merge these implementation with ModelMesh [model puller](https://github.com/kserve/modelmesh-runtime-adapter/tree/main/model-serving-puller).

config/configmap/inferenceservice.yaml
yuzisun: for 1) I think this is mostly for backwards compatibility and integration with Kubeflow environment, for example [tfserving was using S3_ENDPOINT](https://github.com/kubeflow/examples/blob/master/code_search/kubeflow/vendor/kubeflow/tf-serving@dddba26893846041b7cee25ee6121e02a04bd503/prototypes/tf-serving-aws.jsonnet#L51), I am not sure if this is still the case currently.
for 2) Same as other setting, when `s3VerifySSL` is set to false it then does not verify the server side certificates which is mostly for testing purpose.
for 3) I do not think these are needed if we are using service account without secret, @matty-rose maybe we can add a comment in the configmap to state which of the settings are only needed for static credentials?

matty-rose: yep for 3. those fields are not required in the configmap if not using static credentials, have added a comment to make this clearer

surajkota: 1. One change I would recommend based on #2377 is to remove the default value for all the fields except access and secret key name, especially s3Endpoint. This PR lgtm once this change is made.
1. Thanks for adding the comment in this file. I think this should go to the website PR as well and also add information about default values used by each or atleast some of these like https, endpoint, region in website. I will try to review that one today.
1. I could not find any reference for `S3_VERIFY_SSL` variable in boto3 or aws go sdk docs. So, I am not sure how its being used. Dont want to block the PR for this though

surajkota (changes requested):
thanks for the responses, One change request, after that this PR lgtm

config/configmap/inferenceservice.yaml
matty-rose: I have made the change removing the endpoint/use https default values and will try find some time this week to make further updates to the website PR to describe the use of these better. Thanks for your review @surajkota

yuzisun: @surajkota `S3_VERIFY_SSL`/`S3_ENDPOINT`/`S3_USE_HTTPS` were originally inherited from tensorflow/tfserving(see https://docs.w3cub.com/tensorflow~guide/deploy/s3). I think we may no longer need these as we always use storage initializer to download models to a local volume and tfserving runtime never directly use these environment variables to get models from s3 in kserve. I created an [issue](https://github.com/kserve/kserve/issues/2397) to track this to clean up these variables.

surajkota: Ah, got it now

surajkota (approved):
Great work and explanation! Once we take it to the website, I think it would make users happy

config/configmap/inferenceservice.yaml
surajkota: Approved the PR, maybe we should remove `s3VerifySSL` from this configmap since it will not be used or not talk about it on the website for now

yuzisun:
Thanks @matty-rose for the great work and @surajkota for the detailed review! 

/lgtm
/approve

PR #2377, opened 2022-08-09
storage-initializer: Let boto3 decide the endpoint for S3

dimara:
Until now KServe was using the global HTTP endpoint for S3 by default
unless the user explicitly provides it with the corresponding
annotation. If one wants to use IAM roles for service accounts instead
of Secrets with HMAC credentials, storage-initializer will fail in case:

* the bucket is configured with SSE
* if the bucket allows HTTPS requests only
* of an airgapped environment

Specifically, if the bucket is configured with SSE it fails with:

  botocore.exceptions.ClientError: An error occurred (InvalidArgument) when calling the GetObject operation: Requests specifying Server Side Encryption with AWS KMS managed keys must be made over a secure connection.

If the bucket policy allows HTTPS requests only it fails with:

  botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied

In case of an airgapped environment, S3 is only available via the
regional VPC endpoint so requests to the global endpoint will hang.

Handle the above cases by letting boto3 decide the S3 endpoint based
on the environment.

In case of IRSA, AWS will set AWS_REGION and AWS_DEFAULT_REGION and as
such boto3 will use the regional endpoint. boto3 will use HTTPS by
default which is inline with the default value of s3-usehttps
annotation. Finally, if the bucket is in different region boto3 will be
automatically redirected there.

This change should not break existing deployments that don't use IRSA
but don't set s3-endpoint either; boto3 will default using the global
HTTPS endpoint, https://s3.amazonaws.com.

Handle also the case where endpoint_url is unset or empty, where boto3
fails with:

  ValueError: Invalid endpoint:

With this commit, one can use service accounts with the
`eks.amazonaws.com/role-arn` and `eks.amazonaws.com/sts-regional-endpoints`
annotations directly without the need of an attached secret.

Refs #2003
Refs #2113

Signed-off-by: Dimitris Aragiorgis <dimara@arrikto.com>

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -151,9 +151,13 @@ class Storage(object):  # pylint: disable=too-few-public-methods
         #    if awsAnonymousCredential env var true, passed in via config
         # 2. Environment variables
         # 3. ~/.aws/config file
-        s3 = boto3.resource('s3',
-                            endpoint_url=os.getenv("AWS_ENDPOINT_URL", "http://s3.amazonaws.com"),
-                            config=Storage.get_S3_config())
+        kwargs = {
+            "config": Storage.get_S3_config()
+        }
+        endpoint_url = os.getenv("AWS_ENDPOINT_URL")
+        if endpoint_url:
+            kwargs.update({"endpoint_url": endpoint_url})
+        s3 = boto3.resource("s3", **kwargs)
         parsed = urlparse(uri, scheme='s3')
         bucket_name = parsed.netloc
         bucket_path = parsed.path.lstrip('/')


python/kserve/kserve/storage.py
```diff
@@ -151,8 +151,9 @@ def _download_s3(uri, temp_dir: str):
         #    if awsAnonymousCredential env var true, passed in via config
         # 2. Environment variables
         # 3. ~/.aws/config file
+        endpoint_url = os.getenv("AWS_ENDPOINT_URL") or None
         s3 = boto3.resource('s3',
-                            endpoint_url=os.getenv("AWS_ENDPOINT_URL", "http://s3.amazonaws.com"),
+                            endpoint_url=endpoint_url,
```
andrewroblox, line 156: what happens when `endpoint_url` is set to None here?

yuzisun: what happens when `endpoint_url` is set to None here?

dimara: As I mention in the commit message, boto3 will first try to use the global HTTPS endpoint, that is `https://s3.amazonaws.com` and then it will get redirected to the region of the bucket.

I tried the same snippet with storage-initializer with debug logging enabled:
```
boto3.set_stream_logger('', logging.DEBUG)
```
And without AWS_REGION, AWS_DEFAULT_REGION, AWS_ENDPOINT_URL set.

Indeed the first request was to the global endpoint:
```2022-08-16 06:58:43,456 botocore.endpoint [DEBUG] Making request for OperationModel(name=ListObjects) with params: {'url_path': '/MYBUCKET', 'query_string': {'prefix': 'serving/models/sklearn/1.0/model', 'encoding-type': 'url'}, 'method': 'GET', 'headers': {'User-Agent': 'Boto3/1.18.32 Python/3.7.11 Linux/5.4.181-99.354.amzn2.x86_64 Botocore/1.21.32 Resource'}, 'body': b'', 'url': 'https://s3.amazonaws.com/MYBUCKET?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x7f46bf4b8290>, 'has_streaming_input': False, 'auth_type': None, 'encoding_type_auto_set': True, 'signing': {'bucket': 'MYBUCKET'}}}
...
2022-08-16 06:58:44,608 botocore.endpoint [DEBUG] Sending http request: <AWSPreparedRequest stream_output=False, method=GET, url=https://MYBUCKET.s3.amazonaws.com/?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url,...
...
2022-08-16 06:58:43,457 botocore.utils [DEBUG] URI updated to: https://MYBUCKET.s3.amazonaws.com/?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url
...
2022-08-16 06:58:44,608 urllib3.connectionpool [DEBUG] Starting new HTTPS connection (1): MYBUCKET.s3.amazonaws.com:443
2022-08-16 06:58:44,722 urllib3.connectionpool [DEBUG] https://MYBUCKET.s3.amazonaws.com:443 "GET /?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url HTTP/1.1" 400 None

```

Then it got redirected to the region of the bucket:
```
2022-08-16 06:58:44,723 botocore.parsers [DEBUG] Response headers: {'x-amz-bucket-region': 'eu-west-1',...
2022-08-16 06:58:44,726 botocore.utils [DEBUG] S3 client configured for region us-east-1 but the bucket MYBUCKET is in region eu-west-1; Please configure the proper region to avoid multiple unnecessary redirects and signing attempts.
...
2022-08-16 06:58:44,726 botocore.utils [DEBUG] URI updated to: https://MYBUCKET.s3.eu-west-1.amazonaws.com/?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url
...
2022-08-16 06:58:44,968 urllib3.connectionpool [DEBUG] https://MYBUCKET.s3.eu-west-1.amazonaws.com:443 "GET /?prefix=serving%2Fmodels%2Fsklearn%2F1.0%2Fmodel&encoding-type=url HTTP/1.1" 200 None
```

nishank-lily:
This is a very important change. Please merge it. I am also stuck due to this. I am getting this error `botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied`, but when I remove the `endpoint_url` in the boto3.resource(), the code works.

yuzisun:
> This is a very important change. Please merge it. I am also stuck due to this. I am getting this error `botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied`, but when I remove the `endpoint_url` in the boto3.resource(), the code works.

I guess my question is that, do we need to remove `endpoint_url` in the boto3.resource() or can also set `endpoint_url` to `None` as did in this PR.

python/kserve/kserve/storage.py
surajkota: @dimara I think the line `endpoint_url = os.getenv("AWS_ENDPOINT_URL") or None` is a bit confusing to read because it seems like we are trying to set the endpoint to `None`. What do you think about this instead?

```
kwargs = {
  "service_name": "s3",
  "config": Storage.get_S3_config()
}

endpoint_url = os.getenv("AWS_ENDPOINT_URL") or None
if endpoint_url:
  kwargs["AWS_ENDPOINT_URL"] = endpoint_url

s3 = boto3.resource(**kwargs)

**Review by surajkota — CHANGES_REQUESTED** [review]
Thanks for the PR @dimara. Nice work and PR description!

File: python/kserve/kserve/storage.py (hunk 1 above)
Comment (nishank-lily) [reply python/kserve/kserve/storage.py]: I am not sure whether we need `AWS_ENDPOINT_URL` or not. Because when I tested, the boto3 automatically figured out the endpoint URL. I just did `s3 = boto3.resource('s3')` and it worked. 
I think we don't need to do this `endpoint_url = os.getenv("AWS_ENDPOINT_URL") or None` as `os.getenv` automatically returns None if the env variable does not exist. so only this is enough `endpoint_url = os.getenv("AWS_ENDPOINT_URL")`

Comment (surajkota) [reply python/kserve/kserve/storage.py]: Afaik, it is needed for users who are using private VPC and use privatelink to access S3 in commercial regions

Comment (yuzisun) [reply python/kserve/kserve/storage.py]: s3 is not just for AWS, it can also be on prem S3 API compatible storage which needs this environment to override with private links.

Comment (yuzisun) [reply python/kserve/kserve/storage.py]: ye I think `endpoint_url = os.getenv("AWS_ENDPOINT_URL")` is good enough.

Comment (surajkota) [reply python/kserve/kserve/storage.py]: @dimara lets address this comment and merge this PR whenever you get a chance?

Comment (dimara) [reply python/kserve/kserve/storage.py]: @surajkota @nishank-lily @yuzisun thanks for the feedback. Let me comment on the above.

I used `os.getenv("AWS_ENDPOINT_URL") or None` to cover the case where the AWS_ENDPOINT_URL is an empty string. I understand this is possible if `endpoint_url` is not set in the storage secret json because of this line:
https://github.com/kserve/kserve/blob/3994c637d3ecedea3c64be969961af3f1fa9a3cc/python/kserve/kserve/storage.py#L115

Passing `endpoint_url=None` to `boto3.resource()` is the default value based on the method's prototype (see https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.resource).

I see that you suggest to pass `endpoint_url` kwarg only if it is set (and not empty).  So I guess what you want is:
```
endpoint_url = os.getenv("AWS_ENDPOINT_URL")
if endpoint_url:
  kwargs["endpoint_url"] = endpoint_url
```

So, do you want me to make this change?

yuzisun: I think to be consistent we should update line 115 to default to None as well, do you think it is cleaner this way?

dimara: @yuzisun line 115 is about setting an environment variable. IMO the equivalent of Python's `None` is either set AWS_ENDPOINT_URL envvar to empty string (as the code currently does) or not set at all. So I wouldn't make any change there. So the question is whether we want to pass `endpoint_url` to boto3 *only* if the AWS_ENDPOINT_URL is set and not empty (see snippet above).

dimara: @yuzisun @surajkota I updated the PR with the changes you suggested above, that is avoid having the confusing `os.getenv("AWS_ENDPOINT_URL") or None`. Are you OK with them? Can we merge this now?

surajkota (approved):

nishank-lily:
Can someone tell me where to set `AWS_ENDPOINT_URL` in the kserve `InferenceService` yaml file for the current version?

surajkota:
@nishank-lily https://kserve.github.io/website/0.8/modelserving/storage/s3/s3/#create-s3-secret

nishank-lily:
I tried this, and it didn't work. In the end I had to disable https on S3 for development. Here's what I did.

> annotations:
     serving.kubeflow.org/s3-endpoint: s3.us-west-2.amazonaws.com # replace with your s3 endpoint
     serving.kubeflow.org/s3-usehttps: "1" # by default 1, for testing with minio you need to set to 0
     serving.kubeflow.org/s3-region: "us-west-2" # replace with the region the bucket is created in
     serving.kubeflow.org/s3-useanoncredential: "false" # omitting this is the same as false, if true will ignore credential provided and use anonymous credentials
type: Opaque
stringData:
  AWS_DEFAULT_REGION: us-west-2
  AWS_REGION: us-west-2
  AWS_ENDPOINT_URL: https://s3.us-west-2.amazonaws.com

This setting is not updating the value of `AWS_ENDPOINT_URL` env var.

surajkota:
@nishank-lily You will also need to specify empty values for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` keys in your secret. See https://github.com/kserve/kserve/issues/2003#issuecomment-1053530385. Please let us know if this works

This has been fixed in https://github.com/kserve/kserve/pull/2373 and should be available in next release.

yuzisun:
> I tried this, and it didn't work. In the end I had to disable https on S3 for development. Here's what I did.
> 
> > annotations:
> > serving.kubeflow.org/s3-endpoint: s3.us-west-2.amazonaws.com # replace with your s3 endpoint
> > serving.kubeflow.org/s3-usehttps: "1" # by default 1, for testing with minio you need to set to 0
> > serving.kubeflow.org/s3-region: "us-west-2" # replace with the region the bucket is created in
> > serving.kubeflow.org/s3-useanoncredential: "false" # omitting this is the same as false, if true will ignore credential provided and use anonymous credentials
> > type: Opaque
> > stringData:
> > AWS_DEFAULT_REGION: us-west-2
> > AWS_REGION: us-west-2
> > AWS_ENDPOINT_URL: https://s3.us-west-2.amazonaws.com
> 
> This setting is not updating the value of `AWS_ENDPOINT_URL` env var.

Your api group name is wrong, it should be serving.kserve.io.

yuzisun:
Thanks @dimara !!

/lgtm
/approve

PR #2380, opened 2022-08-11
Allows to make Istio and VirtualServices optional

### LINKED ISSUE #1336: Is it possible to make top level KFServing virtual service optional?
/kind feature
/kind question

- As KFServing users, we would like to use KFServing without Istio. For example, KNative supports other Ingress like Contour, Kourier, Ambassador, Gloo, etc but KFServing always requires Istio.
- The top level KFServing's virtual service needs Istio but the virtualservice could be optional? We cannot access the URL in `isvc`, but it still can work with the URL in `ksvc`. The URL in `isvc` is necessary?

- Especially this Ingress reconciler codes https://github.com/kubeflow/kfserving/blob/ec9e8ca569ea173de3717a41130e37cfb756ac8d/pkg/controller/v1beta1/inferenceservice/controller.go#L138-L147 should be optional. For example, if `ingressGateway` and `ingressService` are not configured, the controller could skip the reconciler and then KFServing controller does not touch VirtualService and Gateway at all.

Suresh-Nakkeran:
Signed-off-by: Suresh-Nakkeran <suresh.n@ideas2it.com>

This PR allows followining:
- Istio is optional for Serverless mode and other [networking layer Knative supported](https://knative.dev/docs/install/yaml-install/serving/install-serving-with-yaml/#install-a-networking-layer) can be integrated with KServe such as Contour. 
- KServe Serverless installation can fully run in mesh mode without knative local gateway.

Fixes #1336 

**Type of changes**

- [X] When `DisableIstioVirtualHost` is true, KServe does not create the top level virtual service thus Istio is no longer required for serverless mode.
- [X] When `DisableIstioVirtualHost` is true, KServe works with [Contour](https://projectcontour.io).
- [X] When `DisableIstioVirtualHost` is false, KServe works like before.

1. The Service url in status returns the InferenceService component ksvc url instead of the url with virtual host(isvc name) as top level vs is now disabled, e.g when transformer is added to the InferenceService, the url returns the transformer ksvc url otherwise it returns the predictor ksvc url.

**Checklist**:

**Release note**:

```release-note
`DisableIstioVirtualHost` is introduced in the configmap to disable top level Istio virtual service, Istio is no longer required for serverless mode.
```

### DIFF
--- a/config/configmap/inferenceservice.yaml
+++ b/config/configmap/inferenceservice.yaml
@@ -177,7 +177,8 @@ data:
         "ingressDomain"  : "example.com",
         "ingressClassName" : "istio",
         "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
-        "urlScheme": "http"
+        "urlScheme": "http",
+        "disableIstioVirtualHost": false
     }
   logger: |-
     {
--- a/pkg/controller/v1beta1/inferenceservice/controller_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller_test.go
@@ -1653,4 +1653,113 @@ var _ = Describe("v1beta1 inference service controller", func() {
 		})
 	})
 
+	Context("When creating inference service with predictor and without top level istio virtual service", func() {
+		It("Should have knative service created", func() {
+			By("By creating a new InferenceService")
+			// Create configmap
+			copiedConfigs := make(map[string]string)
+			for key, value := range configs {
+				if key == "ingress" {
+					copiedConfigs[key] = `{
+						"disableIstioVirtualHost": true,
+						"ingressGateway": "knative-serving/knative-ingress-gateway",
+						"ingressService": "test-destination",
+						"localGateway": "knative-serving/knative-local-gateway",
+						"localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local"
+					}`
+				} else {
+					copiedConfigs[key] = value
+				}
+			}
+			var configMap = &v1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      constants.InferenceServiceConfigMapName,
+					Namespace: constants.KServeNamespace,
+				},
+				Data: copiedConfigs,
+			}
+			Expect(k8sClient.Create(context.TODO(), configMap)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(context.TODO(), configMap)
+			serviceName := "foo-disable-istio"
+			var expectedRequest = reconcile.Request{NamespacedName: types.NamespacedName{Name: serviceName, Namespace: "default"}}
+			var serviceKey = expectedRequest.NamespacedName
+			var storageUri = "s3://test/mnist/export"
+			ctx := context.Background()
+			isvc := &v1beta1.InferenceService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      serviceKey.Name,
+					Namespace: serviceKey.Namespace,
+				},
+				Spec: v1beta1.InferenceServiceSpec{
+					Predictor: v1beta1.PredictorSpec{
+						ComponentExtensionSpec: v1beta1.ComponentExtensionSpec{
+							MinReplicas: v1beta1.GetIntReference(1),
+							MaxReplicas: 3,
+						},
+						Tensorflow: &v1beta1.TFServingSpec{
+							PredictorExtensionSpec: v1beta1.PredictorExtensionSpec{
+								StorageURI:     &storageUri,
+								RuntimeVersion: proto.String("1.14.0"),
+								Container: v1.Container{
+									Name:      "kfs",
+									Resources: defaultResource,
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, isvc)).Should(Succeed())
+			defer k8sClient.Delete(ctx, isvc)
+			inferenceService := &v1beta1.InferenceService{}
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, serviceKey, inferenceService)
+				if err != nil {
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
+
+			actualService := &knservingv1.Service{}
+			predictorServiceKey := types.NamespacedName{Name: constants.DefaultPredictorServiceName(serviceKey.Name),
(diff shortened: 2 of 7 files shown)


config/configmap/inferenceservice.yaml
```diff
@@ -199,5 +199,6 @@ data:
     }
   deploy: |-
     {
-      "defaultDeploymentMode": "Serverless"
+      "defaultDeploymentMode": "Serverless",
+      "istioDisabled": true
```
yuzisun: Let's keep this flag in the ingress config, I think the better name is `disableIstioVirtualHost`.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -102,7 +102,16 @@ func getServiceUrl(isvc *v1beta1.InferenceService, urlScheme string) string {
...
 			url.Scheme = urlScheme
-			return strings.Replace(url.String(), fmt.Sprintf("-%s-default", string(constants.Transformer)), "", 1)
+			if !isIstioDisabled {
+				return strings.Replace(url.String(), fmt.Sprintf("-%s-default", string(constants.Transformer)), "", 1)
+			} else {
+				return strings.Replace(
+					url.String(),
+					fmt.Sprintf("-%s-default", string(constants.Transformer)),
+					fmt.Sprintf("-%s-default", string(constants.Predictor)),
+					1,
```
yuzisun: we can directly return the transformer url here

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -323,46 +336,48 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
...
 
-func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService) error {
+func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService, isIstioDisabled bool) error {
 	serviceHost := getServiceHost(isvc)
-	serviceUrl := getServiceUrl(isvc, ir.ingressConfig.UrlScheme)
+	serviceUrl := getServiceUrl(isvc, ir.ingressConfig.UrlScheme, isIstioDisabled)
 	if serviceHost == "" || serviceUrl == "" {
 		return nil
 	}
 	//Create ingress
```
yuzisun:
```suggestion
	// When Istio virtual host is disabled, we return the underlying component url.
	// When Istio virtual host is enabled. we return the url using inference service virtual host name and redirect to the corresponding transformer, predictor or explainer url.
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -1653,4 +1653,108 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			actualIsvc := &v1beta1.InferenceService{}
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, expectedRequest.NamespacedName, actualIsvc)
+				if err != nil {
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
+
+			Expect(actualIsvc.Status.URL).To(gomega.Equal(&apis.URL{
```
yuzisun, line 1754: There is another internal URL field `actualIsvc.Status.Address.URL` which we need to check, I think that might be still returning the incorrect url using top level isvc host instead of predictor host.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -379,9 +390,13 @@ func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService) error {
 				path = constants.PredictPath(isvc.Name, constants.ProtocolV2)
 			}
 		}
+		hostPrefix := isvc.Name
+		if disableIstioVirtualHost == true {
+			hostPrefix = fmt.Sprintf("%s-%s-default", hostPrefix, string(constants.Predictor))
```
yuzisun: For transformer case it uses the transformer ksvc host name

Suresh-Nakkeran: @yuzisun updated!

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -401,3 +413,13 @@ func routeSemanticEquals(desired, existing *v1alpha3.VirtualService) bool {
 		equality.Semantic.DeepEqual(desired.ObjectMeta.Labels, existing.ObjectMeta.Labels) &&
 		equality.Semantic.DeepEqual(desired.ObjectMeta.Annotations, existing.ObjectMeta.Annotations)
 }
+
+func getHostPrefix(isvc *v1beta1.InferenceService, disableIstioVirtualHost bool) string {
+	if disableIstioVirtualHost == true {
+		if isvc.Spec.Transformer != nil {
+			return fmt.Sprintf("%s-%s-default", isvc.Name, string(constants.Transformer))
```
yuzisun: nit: let's use the [default transformer name utility function](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L435)

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -401,3 +413,13 @@ func routeSemanticEquals(desired, existing *v1alpha3.VirtualService) bool {
...
+		}
+		return fmt.Sprintf("%s-%s-default", isvc.Name, string(constants.Predictor))
```
yuzisun: Same here use the [default predictor name utility](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L411) so we can get rid of the default name later on easier.

rachitchauhan43:
@Suresh-Nakkeran : 1qq, say we go with disabling the top level virtual service and now deploy an ISVC with both transformer and a predictor.
1. Now when i make a call to my ISVC, I will have to use transformer’s url  right ?
2. Will Kserve still support routing request from transformer to predictor or not ?

Is it possible to add a diagram of what was there earlier with Top Level VS and what can now be achieved with disabling top level VS ?

PR #2386, opened 2022-08-17
Alexagriffith/implement prometheus metrics python

alexagriffith:

Our users have come to us with a lot of questions regarding latencies in the `kserve-container`. As of now, we have no way to know the latencies of each step in the transformer/predictor. Having observability into the latency of each step will let us quickly identify which step is the bottleneck when request latencies are high. At the moment, there is no easy way to figure this out. 

This PR adds prometheus histogram latency metrics for each step (pre/post processing, explain, predict) in the python SDK. The metrics are exposed on /metrics. Additionally, we will log the latency of each step per request id. If you choose not to use prometheus metrics, evaluating the logs is another option. 

Fixes #

**Type of changes**

- [x] Test documented in transformer sample https://github.com/kserve/kserve/pull/2386/files#diff-a8d78136ac31d86dd82d5b2d24410ca6eb48265aae92c5ea63e360d56dfdf58aR146 
I used the transformer to curl localhost:8080/metrics and see the prometheus metrics output. 

- Logs

**Checklist**:

**Release note**:

```release-note

```


python/kserve/requirements.txt
```diff
@@ -25,3 +25,4 @@ cffi==1.15.0
 idna==3.3
 certifi==2021.10.8
 tritonclient==2.18.0
+prometheus-client
```
alexagriffith: todo: specify version

python/kserve/kserve/model.py
```diff
@@ -25,13 +28,20 @@
...
 
 PREDICTOR_URL_FORMAT = "http://{0}/v1/models/{1}:predict"
 EXPLAINER_URL_FORMAT = "http://{0}/v1/models/{1}:explain"
 PREDICTOR_V2_URL_FORMAT = "http://{0}/v2/models/{1}/infer"
 EXPLAINER_V2_URL_FORMAT = "http://{0}/v2/models/{1}/explain"
 
+PRE_HIST_TIME = Histogram('request_preprocessing_seconds', 'pre-processing request latency')
+POST_HIST_TIME = Histogram('request_postprocessing_seconds', 'post-processing request latency')
+PREDICT_HIST_TIME = Histogram('request_predict_processing_seconds', 'prediction request latency')
+EXPLAIN_HIST_TIME = Histogram('request_explain_processing_seconds', 'explain request latency')
```
alexagriffith, line 43: we have four histogram metrics, one for each step in __call__

python/kserve/kserve/model.py
```diff
@@ -60,6 +70,10 @@ def __str__(self):
         return self.reason
 
 
+def get_latency_ms(start, end):
+    return round((end - start) * 1000, 9)
```
alexagriffith, line 74: queue-proxy rounds to 9 decimals, so I set this number to do the same.

python/kserve/kserve/model.py
```diff
@@ -75,19 +89,43 @@ def __init__(self, name: str):
         self._http_client_instance = None
         self._grpc_client_stub = None
 
-    async def __call__(self, body, model_type: ModelType = ModelType.PREDICTOR):
-        request = await self.preprocess(body) if inspect.iscoroutinefunction(self.preprocess) \
-            else self.preprocess(body)
+    async def __call__(self, body, model_type: ModelType = ModelType.PREDICTOR, headers={}):
+        request_id = headers.get("X-RequestID")
```
alexagriffith, line 93: there is a PR that will pass in headers here that is IP https://github.com/kserve/kserve/pull/2360

one question - if there is no header (like, if you aren't using queue proxy) do we want to create a request id or leave it as none?

python/kserve/kserve/model.py
```diff
@@ -75,19 +89,43 @@ def __init__(self, name: str):
...
+
+        # latency vars
+        preprocess_ms = 0
+        explain_ms = 0
+        predict_ms = 0
+        postprocess_ms = 0
+
+        with PRE_HIST_TIME.time():
```
alexagriffith, line 101: each step uses prometheus's `.time()` to record a histogram, and python's time func for the log at the end that spits out just the latencies for each step per request id.

python/kserve/kserve/model_server.py
```diff
@@ -50,6 +53,11 @@
 
 tornado.log.enable_pretty_logging()
 
+class MetricsHandler(RequestHandler):
```
alexagriffith, line 56: took this from https://github.com/globocom/tornado-prometheus/blob/master/tornado_prometheus/__init__.py 

we only needed the metrics handler, when we switch to fastapi, this should be super easy to change.

python/kserve/kserve/model_server.py
```diff
@@ -68,6 +76,7 @@ def __init__(self, http_port: int = args.http_port,
 
     def create_application(self):
         return tornado.web.Application([
+            (r"/metrics", MetricsHandler),
```
alexagriffith, line 79: expose metrics on 8080/metrics

python/kserve/requirements.txt
```diff
@@ -21,3 +21,4 @@ ray[serve]==2.0.0
 grpcio>=1.34.0
 tritonclient==2.18.0
 protobuf~=3.19.0
+prometheus-client>=0.13.1
```
alexagriffith, line 24: 0.14 is the latest version, but ray[serve] requires <0.14, and this is the latest version that is <0.14

python/kserve/test/test_inference_service_client.py
```diff
@@ -22,7 +22,7 @@
 from kserve import V1beta1InferenceService
 from kserve import KServeClient
 
-kserve_client = KServeClient(config_file='./kserve/test/kubeconfig')
+kserve_client = KServeClient(config_file='./test/kubeconfig')
```
alexagriffith, line 25: q about this - if you run from the Makefile, you need the path I changed it to. The Makefile command wont work as is with this path, so it is broken. Should we change it?

python/kserve/kserve/model.py
```diff
@@ -75,19 +89,43 @@ def __init__(self, name: str):
...
             raise NotImplementedError
-        response = self.postprocess(response)
+
+        with POST_HIST_TIME.time():
+            start = time.time()
+            response = self.postprocess(response)
+            postprocess_ms = get_latency_ms(start, time.time())
+
+        logging.info(f"requestId: {request_id}, preprocess_ms: {preprocess_ms}, explain_ms: {explain_ms}, predict_ms: {predict_ms}, postprocess_ms: {postprocess_ms}")
+
```
alexagriffith, line 128: note on the code: calling `start = time.time()` seems redundant here - Ive put some thought into it but not sure if the alternatives are cleaner. afaik, using a decorator would mean we have to log each step's time, so instead of one single log at the end of __call__ we would have one for each step, probably not great to expand the logs that much. 

im open to suggestions!

docs/samples/v1beta1/transformer/torchserve_image_transformer/image_transformer/__main__.py
```diff
@@ -18,22 +18,21 @@
 from .transformer_model_repository import TransformerModelRepository
 
 DEFAULT_MODEL_NAME = "model"
+CONFIG_PATH = "/mnt/models/config/config.properties"
 
 parser = argparse.ArgumentParser(parents=[kserve.model_server.parser])
 parser.add_argument(
     "--predictor_host", help="The URL for the model predict function", required=True
 )
+parser.add_argument("--config_path", help="The path to the config file containing the list of model names", required=False, default=CONFIG_PATH)
```
alexagriffith, line 27: allow for us to easily use a local config file to run transformer locally to test

docs/samples/v1beta1/transformer/torchserve_image_transformer/local_config.properties
```diff
@@ -0,0 +1 @@
+model_snapshot={"name":"startup.cfg","modelCount":1,"models":{"mnist":{"1.0":{"defaultVersion":true,"marName":"mnist.mar","minWorkers":1,"maxWorkers":5,"batchSize":1,"responseTimeout":120}}}}
```
alexagriffith, line 1: took this from documentation https://kserve.github.io/website/0.8/modelserving/v1beta1/torchserve/#create-the-torchserve-inferenceservice

alexagriffith: also ran /v1/models and "mnist" is the model we print out

python/kserve/README.md
```diff
@@ -54,6 +54,12 @@ It supports the following storage providers:
     * `https://<some_url>.com/model.joblib`
     * `http://<some_url>.com/model.joblib`
 
+### Metrics
```
alexagriffith, line 57: should I add more here?

docs/samples/v1beta1/transformer/torchserve_image_transformer/README.md
```diff
@@ -142,3 +142,25 @@ Handling connection for 8080
 * Connection #0 to host localhost left intact
 {"predictions": [2]}
 ```
+
```
alexagriffith, line 145: step by step how I tested

alexagriffith (commented):
Can we add label donotmerge to this PR? I need to wait until the headers PR is merged.

python/kserve/test/test_inference_service_client.py
yuzisun: looks like his is breaking the test, probably better to change the Makefile

docs/samples/v1beta1/transformer/torchserve_image_transformer/README.md
```diff
@@ -142,3 +142,25 @@ Handling connection for 8080
...
+## Testing with a local transformer & predictor in Kubernetes (developer guide)
```
yuzisun, line 146: I think this is super useful! We should move this to a more general local developer guide for custom transformer/predictor.

python/kserve/kserve/model.py
yuzisun, line 93: according to the logging format `"X-RequestID": "{{index .Request.Header "X-Request-Id"}}"`, the header name is actually `X-Request-Id`

python/kserve/test/test_inference_service_client.py
alexagriffith: fixed in next PR

docs/samples/v1beta1/transformer/torchserve_image_transformer/README.md
alexagriffith: in kerve/website?

PR #2390, opened 2022-08-18
Enable explainers to work with ModelMesh

### LINKED ISSUE #2388: Enable explainers to work with ModelMesh
/kind feature

I would like the existing KServe explainers to work with Modelmesh.

The explainers currently work with predictors deployed by KServe, not the ones deployed by ModelMesh.

chinhuang007:
Currently explainers don't work when the inference service deploymentMode is ModelMesh. This PR enables KServe explainers to work with ModelMesh.

Fixes #2388

**Type of changes**

The alibi cifar10 explainer example is updated with ModelMesh annotation and used to validate the changes.

1. The controllers are changed to wait for predictor status and provide predictor protocol for explainers to adjust payloads and innovation
2. Ingress reconciler is updated to bypass predictor route when deploymentMode is ModelMesh
3. Alibi explainer is enhanced with logic to handle additional predictor protocols, REST v2 and gRPC
4. AIX, AIF, ART are unchanged since these don't use framework based servingtimes for predictors, not suitable for ModelMesh
5. The default triton server in ModelMesh is not configured to support batch size > 1. Manual configuration (max-batch-size>=100) is needed to make the alibi cifar10 explainer to work. 

**Release note**:
```
KServe explainers are now enabled to work with ModelMesh. Please note the Triton server in ModelMesh has to be configured with max-batch-size=100 or more for the Alibi cifar10 explainer to work.
```

### DIFF
--- a/pkg/apis/serving/v1beta1/explainer_alibi.go
+++ b/pkg/apis/serving/v1beta1/explainer_alibi.go
@@ -68,31 +68,24 @@ func (s *AlibiExplainerSpec) GetResourceRequirements() *v1.ResourceRequirements 
 }
 
 func (s *AlibiExplainerSpec) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
-	alibiExplainerLogger.Info("=====hello 0 alibi=====", "metadata.Name", metadata.Name)
 	var args = []string{
 		constants.ArgumentModelName, metadata.Name,
 		constants.ArgumentHttpPort, constants.InferenceServiceDefaultHttpPort,
 	}
 
 	argumentPredictorHost := fmt.Sprintf("%s.%s", constants.DefaultPredictorServiceName(metadata.Name), metadata.Namespace)
-	//argumentPredictorPort := constants.InferenceServiceDefaultHttpPort
 	deploymentMode, ok := metadata.Annotations[constants.DeploymentMode]
-	alibiExplainerLogger.Info("=====hello 1.0 alibi=====", "argumentPredictorHost", argumentPredictorHost)
 
 	if ok && (deploymentMode == string(constants.ModelMeshDeployment)) {
 		// Get predictor host and protocol from annotations in modelmesh deployment mode
-		alibiExplainerLogger.Info("=====hello 1.1 alibi=====", "argumentPredictorHost", metadata.Annotations[constants.PredictorHostAnnotationKey])
 		argumentPredictorHost = metadata.Annotations[constants.PredictorHostAnnotationKey]
 		argumentPredictorProtocol := metadata.Annotations[constants.PredictorProtocolAnnotationKey]
-		alibiExplainerLogger.Info("=====hello 1.2 alibi=====", "argumentPredictorHost", argumentPredictorHost)
-		alibiExplainerLogger.Info("=====hello 2.1 alibi=====", "argumentPredictorProtocol", argumentPredictorProtocol)
 
 		// Set predictor protocol if not provided in container arguments
 		if !utils.IncludesArg(args, "--protocol") {
 			alibiExplainerLogger.Info("Set predictor protocol based on ModelMesh predictor URL", "protocol", argumentPredictorProtocol)
 			args = append(args, "--protocol", argumentPredictorProtocol)
 		}
-		//argumentPredictorPort = argumentPredictorProtocol
 	}
 
 	if !utils.IncludesArg(s.Container.Args, constants.ArgumentPredictorHost) {
@@ -125,7 +118,6 @@ func (s *AlibiExplainerSpec) GetContainer(metadata metav1.ObjectMeta, extensions
 	}
 	s.Name = constants.InferenceServiceContainerName
 	s.Args = append(args, s.Args...)
-	alibiExplainerLogger.Info("=====hello 3.1 alibi=====", "s.Args", s.Args)
 	return &s.Container
 }
 
--- a/pkg/apis/serving/v1beta1/explainer_alibi.go
+++ b/pkg/apis/serving/v1beta1/explainer_alibi.go
@@ -85,13 +85,13 @@ func (s *AlibiExplainerSpec) GetContainer(metadata metav1.ObjectMeta, extensions
 		argumentPredictorHost = metadata.Annotations[constants.PredictorHostAnnotationKey]
 		argumentPredictorProtocol := metadata.Annotations[constants.PredictorProtocolAnnotationKey]
 		alibiExplainerLogger.Info("=====hello 1.2 alibi=====", "argumentPredictorHost", argumentPredictorHost)
-		alibiExplainerLogger.Info("=====hello 2 alibi=====", "argumentPredictorProtocol", argumentPredictorProtocol)
+		alibiExplainerLogger.Info("=====hello 2.1 alibi=====", "argumentPredictorProtocol", argumentPredictorProtocol)
 
 		// Set predictor protocol if not provided in container arguments
-		//if !utils.IncludesArg(args, "--protocol") {
-		//	customExplainerLogger.Info("Set predictor protocol based on ModelMesh predictor URL", "protocol", argumentPredictorProtocol)
-		//	args = append(args, "--protocol", argumentPredictorProtocol)
-		//}
+		if !utils.IncludesArg(args, "--protocol") {
+			alibiExplainerLogger.Info("Set predictor protocol based on ModelMesh predictor URL", "protocol", argumentPredictorProtocol)
+			args = append(args, "--protocol", argumentPredictorProtocol)
+		}
 		//argumentPredictorPort = argumentPredictorProtocol
 	}
 
@@ -125,6 +125,7 @@ func (s *AlibiExplainerSpec) GetContainer(metadata metav1.ObjectMeta, extensions
 	}
 	s.Name = constants.InferenceServiceContainerName
 	s.Args = append(args, s.Args...)
+	alibiExplainerLogger.Info("=====hello 3.1 alibi=====", "s.Args", s.Args)
 	return &s.Container
 }
 
--- a/pkg/apis/serving/v1beta1/explainer_alibi.go
+++ b/pkg/apis/serving/v1beta1/explainer_alibi.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kserve/kserve/pkg/utils"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // AlibiExplainerType is the explanation method
@@ -54,6 +55,11 @@ type AlibiExplainerSpec struct {
 	ExplainerExtensionSpec `json:",inline"`
 }
 
+var (
+	// logger for the alibi explainer
+	alibiExplainerLogger = logf.Log.WithName("inferenceservice-v1beta1-alibi-explainer")
+)
+
 var _ ComponentImplementation = &AlibiExplainerSpec{}
 
 func (s *AlibiExplainerSpec) GetResourceRequirements() *v1.ResourceRequirements {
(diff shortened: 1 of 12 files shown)


chinhuang007:
@yuzisun Wondering if you can take a look?

yuzisun:
> @yuzisun Wondering if you can take a look?

Sorry I will be back from vacation tomorrow, will get this done this week.

pkg/apis/serving/v1beta1/explainer_alibi.go
```diff
@@ -66,9 +72,24 @@ func (s *AlibiExplainerSpec) GetContainer(metadata metav1.ObjectMeta, extensions
 		constants.ArgumentModelName, metadata.Name,
 		constants.ArgumentHttpPort, constants.InferenceServiceDefaultHttpPort,
 	}
+
+	argumentPredictorHost := fmt.Sprintf("%s.%s", constants.DefaultPredictorServiceName(metadata.Name), metadata.Namespace)
+	deploymentMode, ok := metadata.Annotations[constants.DeploymentMode]
```
yuzisun, line 77: We also need to check the global config `deploymentMode` so can use the util function [GetDeploymentMode here](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L68)

pkg/apis/serving/v1beta1/explainer_custom.go
```diff
@@ -69,6 +75,22 @@ func (c *CustomExplainer) GetStorageSpec() *StorageSpec {
 // GetContainer transforms the resource into a container spec
 func (c *CustomExplainer) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig) *v1.Container {
 	container := &c.Containers[0]
+
+	argumentPredictorHost := constants.PredictorURL(metadata, false)
+	deploymentMode, ok := metadata.Annotations[constants.DeploymentMode]
+
+	if ok && (deploymentMode == string(constants.ModelMeshDeployment)) {
```
yuzisun, line 82: same here to call [GetDeploymentMode](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L68)

pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -79,6 +87,29 @@ func (e *Explainer) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 		}),
 		Annotations: annotations,
 	}
+
+	// Need to wait for predictor URL in ModelMesh deployment mode
+	if isvcutils.GetDeploymentMode(annotations, deployConfig) == constants.ModelMeshDeployment {
+		// check if predictor URL is populated
+		predictorURL := (*url.URL)(isvc.Status.Components["predictor"].URL)
```
yuzisun: Why do we need the cast for `URL`? I think it is just a typedef

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -203,6 +204,8 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			return reconcile.Result{}, errors.Wrapf(err, "fails to reconcile ingress")
 		}
 	} else {
+		// }
+		// if deploymentMode == constants.Serverless || (deploymentMode == constants.ModelMeshDeployment && isvc.Spec.Explainer == nil) {
```
yuzisun: Can be removed now?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,6 +79,16 @@ func getServiceHost(isvc *v1beta1.InferenceService) string {
...
+	if isvc.Spec.Explainer != nil && deploymentMode == constants.ModelMeshDeployment {
+		if explainerStatus, ok := isvc.Status.Components[v1beta1.ExplainerComponent]; !ok {
+			return ""
+		} else if explainerStatus.URL == nil {
+			return ""
+		} else {
+			return strings.Replace(explainerStatus.URL.Host, fmt.Sprintf("-%s-default", string(constants.Explainer)), "",
+				1)
+		}
+	}
```
yuzisun, line 90: I do not think we need this block as the `explain` verb matches the routing rule to explainer ksvc with the isvc host.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -278,21 +292,23 @@ func createIngress(isvc *v1beta1.InferenceService, config *v1beta1.IngressConfig
...
-		Match: createHTTPMatchRequest("", serviceHost,
-			network.GetServiceHostname(isvc.Name, isvc.Namespace), isInternal, config),
-		Route: []*istiov1alpha3.HTTPRouteDestination{
-			createHTTPRouteDestination(backend, isvc.Namespace, config.LocalGatewayServiceName),
-		},
-		Headers: &istiov1alpha3.Headers{
-			Request: &istiov1alpha3.Headers_HeaderOperations{
-				Set: map[string]string{
-					"Host": network.GetServiceHostname(backend, isvc.Namespace),
+	// Add predict route, only if deployment mode is not ModelMesh
```
yuzisun, line 302: This is the internal route which is still needed to route to explainer I think, it is not modelmesh specific as the request always hits explainer first which is orchestrated by kserve

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -379,12 +405,14 @@ func (ir *IngressReconciler) Reconcile(isvc *v1beta1.InferenceService) error {
 				path = constants.PredictPath(isvc.Name, constants.ProtocolV2)
 			}
 		}
-		isvc.Status.Address = &duckv1.Addressable{
-			URL: &apis.URL{
-				Host:   network.GetServiceHostname(isvc.Name, isvc.Namespace),
-				Scheme: "http",
-				Path:   path,
-			},
+		if deploymentMode != constants.ModelMeshDeployment || isvc.Spec.Transformer != nil {
```
yuzisun, line 420: Why this is specific to ModelMesh?

python/alibiexplainer/alibiexplainer/explainer.py
```diff
@@ -71,9 +86,50 @@ def _predict_fn(self, arr: Union[np.ndarray, List]) -> np.ndarray:
                 instances.append(req_data.tolist())
             else:
                 instances.append(req_data)
+        shape = np.shape(instances)
         loop = asyncio.get_running_loop()  # type: ignore
-        resp = loop.run_until_complete(self.predict({"instances": instances}))
-        return np.array(resp["predictions"])
+
+        # Prepare the request beased on the predictor protocol
+        if self.protocol == PredictorProtocol.GRPC_V2.value:
```
yuzisun: Can we add an e2e test ?

chinhuang007:
@yuzisun Thanks for the review. Sorry, being tied with other things. Will get back to your comments shortly.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
chinhuang007: This block is needed because in the case of explainer with ModelMesh predictor we don't want to return the ModelMesh serving url host, something like "modelmesh-serving.namespace:8033". The returned string will be added to hosts for the desiredIngress and fail the ingress creation. The logic could be implemented in other places. But it won't work without the logic of using explainer's url host instead predictor's when the explainer has ModelMesh predictor.

pkg/apis/serving/v1beta1/explainer_custom.go
chinhuang007: GetContainer is part of serving component interface, not acting as a controller/reconciler. The serving components do not have the controller client or DeployConfig which is required by the global GetDeploymentMode function. I am not sure whether it is necessary to use the utility function, which seems designed for controllers, in serving components.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
chinhuang007: The condition is to filter out the predict route when predictor is managed by ModelMesh. Since currently ModelMesh doesn't work with ingress, we won't be able to orchestrate explainers to work with ModelMesh predictors automatically in KServe ingress. We might want to think about whether a better solution exists.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
chinhuang007: It is because the field is updated by ModelMesh already. We don't want to overwrite it here.

python/alibiexplainer/alibiexplainer/explainer.py
chinhuang007: e2e is possible but it requires ModelMesh installation and an issue to be addressed, https://github.com/kserve/modelmesh-serving/issues/254

chinhuang007: Please note the Alibi explainer output values might/will be different between requests due to some randomness in it's logic. e2e would prove functionally it works. Unfortunately I am not the Alibi expert to interpret the results though.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -188,6 +188,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 	//Reconcile ingress
+	//Skip if deployment mode is modelmesh and explainer exists
```
yuzisun: staled comment?

python/alibiexplainer/alibiexplainer/explainer.py
```diff
@@ -71,9 +86,50 @@ def _predict_fn(self, arr: Union[np.ndarray, List]) -> np.ndarray:
...
+            data = np.array(instances, dtype=np.float32).flatten()
```
yuzisun: is this always going to be float32 type? we might need to map the np array type to v2 tensor type

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
yuzisun: We can route with the modelmesh internal in cluster service right?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
yuzisun: We should really support ModelMesh ingress then as if only predictor presents the ingress should still work.

chinhuang007: I agree. That itself is a big work item. Unfortunately I can't address it in this PR.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
chinhuang007: The default for modelmesh service is grpc. It seems a bit complicated to make it a generic route in ingress to modelmesh, involving getting modelmesh config for service name, port number, payload conversion etc. I would suggest a separate discuss and PR to tackle it.

python/alibiexplainer/alibiexplainer/explainer.py
chinhuang007: ModelMesh has never been part of KServe e2e. It would take some thoughts, alignment, and work to get it right.  So for this PR I added a notebook for the Alibi cifar10 example with instructions for end to end execution of running the Alibi explainer with predictor hosted in ModelMesh.

python/alibiexplainer/alibiexplainer/explainer.py
chinhuang007: For images, it is float32. Other types don't seem to work. I don't know about tabular and text cases so I added a conditional check for this PR.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
yuzisun: Can we add a @TODO here so we remember to address this in future?

spolti:
HI all, is this issue still relevant?

spolti:
Is this issue still valid?

yuzisun:
> Is this issue still valid?

As model mesh is in maintenance mode, I think we want to minimize the kserve dependency on MM.

PR #2391, opened 2022-08-21
Update python/kserve dependency and art example

yuzisun:
- Update python requirement to 3.7.
- Do not pin kserve version
- Update adversarial example: delete the sklearnserver code living in the example which should directly use the sklearnserver image and load the model from cloud storage with `StorageUri` field.
-  Update art e2e test case.

**Release note**:

```release-note

```


chinhuang007 (approved):
Looks good, thanks, @yuzisun !

PR #2396, opened 2022-08-24
Propagating IG headers to it's nodes.

### LINKED ISSUE #2289: Inference graphs should pass on the headers to all the steps in the flow.
/kind feature

Use case:  Transformers (pre and post-processors) of inference services(ISVC) need to call other internal services to fetch some data. For making those calls, headers passed on to inference graph endpoint will be needed to be passed on to every step in inference graph, i.e to every ISVC, to allow them call those internal services.
Example: AuthZ and tracing related headers will be needed.

rachitchauhan43:
It propagates IG headers to it's nodes.

PR to propagate headers to predict method is already there in progress #2360 . This PR can be merged independently of #2360 as headers can be propagated from IG to it's nodes and when #2360 is merged then headers will start propagating to nodes as well.

Fixes #2289 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```


rachitchauhan43:
Context: Passing down headers to IG's nodes was a 2 level problem as explained below:
1. First level problem was, IG was not passing down the headers to it's nodes. 
Will be solved by this [PR](https://github.com/kserve/kserve/pull/2396)
2. Second level problem was, ISVCs were not passing the down the headers to their methods like pre-process/predict/post-process. 
Will be solved by #2360 
 
Screen Shot 2022-08-24 at 8 37 03 AM

andyi2it:
@rachitchauhan43 
Thanks for making these changes. Will update here.

rachitchauhan43:
> @rachitchauhan43 Thanks for making these changes. Will update here.

@andyi2it / @yuzisun  / @Iamlovingit : Are we good to approve and merge this ?

andyi2it (approved):

cmd/router/main.go
```diff
@@ -38,8 +38,13 @@ import (
 
 var log = logf.Log.WithName("InferenceGraphRouter")
 
-func callService(serviceUrl string, input []byte) ([]byte, error) {
-	resp, err := http.Post(serviceUrl, "application/json", bytes.NewBuffer(input))
+func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, error) {
+	client := http.Client{}
```
yuzisun: Should this http client be cached in some way?  we probably need to optimize here to make sure connections are reused by url.

cmd/router/main.go
```diff
@@ -38,8 +38,13 @@ import (
...
+	req, err := http.NewRequest("POST", serviceUrl, bytes.NewBuffer(input))
+	req.Header = headers // propagating the headers to steps i.e. to ISVCs
```
yuzisun: Does it always make sense to propagate all the headers to the services or we need to filter for certain headers.

rachitchauhan43: I don't see a need to why would we want to filters the headers to be propagated. 
Do you have a use case in mind ?

cmd/router/main.go
rachitchauhan43: @yuzisun : Are you talking about reusing underlying TCP connections for making further HTTP requests ?
If that is the concern then I did a quick test and observed that when we are using 
`body, err := ioutil.ReadAll(resp.Body)` at line [53](https://github.com/kserve/kserve/pull/2396/files#diff-d8ccd6fde43fcb90d60f6319f818662f9e4340295fc9bbde755fd93d98da33c1R53) then multiple TCP connections are not being opened and same TCP connection is being used for new http requests. See image below:
connection_reused


Where as when I am commenting out line 53 then I am seeing multiple TCP connections being put into TIME_WAIT state. See attached image:
connection_not_reused

So, if that is the concern then I guess we are good. You can check my gist here that I used to test this : https://gist.github.com/rachitchauhan43/8b7fc6a24074d0d0c0d9095f4e583a07

Also, referred official go docs to verify that we are correctly using `http` client from package.
See here: https://go.dev/src/net/http/response.go#L59

cmd/router/main.go
yuzisun: Like the span Id used for distributed tracing which should not be passed to the next step. I guess we need to explicitly support distributed tracing in graph orchestrator so it can propagate headers correctly.

cmd/router/main.go
yuzisun: Thanks for the detailed answer, I am actually referring to reusing the http connection across different requests, currently we are creating the client for each request. From the net http doc

Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

rachitchauhan43: Ok. Will take care of that.

cmd/router/main.go
rachitchauhan43: Ok. Since router (or better to say IG) doesn't support distributed tracing at this time point in time, so we have few approaches that we can proceed with:

1. Since router (or IG) doesn’t support distributed tracing at this time, so we go ahead with propagating all the headers and then the work for enabling distributed tracing (which will be a separate PR) should take care of remove tracing related headers and keep the rest as it is. Basically keeping things backward compatible.
Passing all the headers at this point in time anyways won't break distributed tracing since router is not instrumented for distributed tracing anyways. Is this understanding Right ? 

2. Second approach is, we do pass on only selective headers. I checked how ApolloGraphQL does it for it's router (see [here](https://www.apollographql.com/docs/router/configuration/header-propagation/))
This work might be scrapped later on once router code is instrumented for distributed tracing.

@yuzisun Let me know your thoughts on both the approaches and which one can be picked up.

cmd/router/main.go
rachitchauhan43: @yuzisun : Since, I am not changing any config of client at this time and only dealing with updating `Request` object so reverted back to using global http client provided by net/http package.

rachitchauhan43:
@njhill : As discussed in last bi-weekly, can you review this PR ? It will be very useful feature for us to consume in Inference Graphs.

cmd/router/main.go
yuzisun: We actually run into some issue with propagating headers between transformer and predictor in the other related PR as it accidentally copied over the `HOST` header from the previous request which messes up the calls, I think it is still better to explicitly copy over the headers which are the same across the request chain instead of copying everything which can be very hard to debug if some header is passed along unexpectedly.

rachitchauhan43: @yuzisun : I have made passing headers configurable using `router` section of `inferenceservice-config` config map.
This is how headers to propagated can be mentioned in router's config.
```
{
    "image" : "docker.io/rachitchauhan885/router-20acce65bc84b6e5b28ae59ce41635d3@sha256:729d6c170f28355075dd609dd2e101e8ce519342e406982b906364107703660e",
    "memoryRequest": "100Mi",
    "memoryLimit": "500Mi",
    "cpuRequest": "100m",
    "cpuLimit": "100m",
    "headers": {
      "propagate":[
        "Authorization"
      ]
    }
}
```

cmd/router/main.go
```diff
@@ -183,7 +194,8 @@ func graphHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 var (
-	jsonGraph = flag.String("graph-json", "", "serialized json graph def")
+	jsonGraph          = flag.String("graph-json", "", "serialized json graph def")
+	headersToPropagate = strings.Split(os.Getenv("PROPAGATE_HEADERS"), ",")
```
yuzisun: Why environment variable instead of args?

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -152,6 +153,12 @@ func createKnativeService(componentMeta metav1.ObjectMeta, graph *v1alpha1api.In
 										"--graph-json",
 										string(bytes),
 									},
+									Env: []v1.EnvVar{
+										{
+											Name:  "PROPAGATE_HEADERS",
+											Value: strings.Join(config.Headers["propagate"], ","),
```
yuzisun: What happens when `propagate` does not exist ?

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -66,11 +66,12 @@ const (
...
-	CpuRequest    string `json:"cpuRequest"`
-	CpuLimit      string `json:"cpuLimit"`
-	MemoryRequest string `json:"memoryRequest"`
-	MemoryLimit   string `json:"memoryLimit"`
+	Image         string              `json:"image"`
+	CpuRequest    string              `json:"cpuRequest"`
+	CpuLimit      string              `json:"cpuLimit"`
+	MemoryRequest string              `json:"memoryRequest"`
+	MemoryLimit   string              `json:"memoryLimit"`
+	Headers       map[string][]string `json:"headers"`
```
yuzisun: Why it is a map of strings ?

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
rachitchauhan43: Value of "PROPAGATE_HEADERS" is empty string in that case.

pkg/controller/v1alpha1/inferencegraph/controller.go
rachitchauhan43: Today, I added the feature so that user can specify which headers to propagate using "propagate" list.
Now, if in future someone wants to add section like "remove" to remove certain headers or "transform" to update certain headers before passing down to nodes then this design makes things extensible.

cmd/router/main.go
rachitchauhan43: No particular preference here. I just followed what we discussed over slack chat.

pkg/controller/v1alpha1/inferencegraph/controller.go
rachitchauhan43: I tried to learn from how [others](https://www.apollographql.com/docs/router/configuration/header-propagation/) are doing it and made it even simpler to start with.

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
yuzisun: I think it is a bit confusing here, if  `propagate` does not exist then we should not inject this env variable

pkg/controller/v1alpha1/inferencegraph/controller.go
yuzisun: Can we add some comments here ？

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
rachitchauhan43: Addressed. Adding this env variable only when it is added in router's config.

pkg/controller/v1alpha1/inferencegraph/controller.go
rachitchauhan43: Added.

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -170,6 +171,18 @@ func createKnativeService(componentMeta metav1.ObjectMeta, graph *v1alpha1api.In
 			},
 		},
 	}
+
+	// Only adding this env variable "PROPAGATE_HEADERS" if router's headers config has the key "propagate"
+	value, exists := config.Headers["propagate"]
+	if exists {
+		service.Spec.ConfigurationSpec.Template.Spec.PodSpec.Containers[0].Env = []v1.EnvVar{
+			{
+				Name:  "PROPAGATE_HEADERS",
```
yuzisun: Can we define this as a constant as it is used in multiple places

cmd/router/main.go
```diff
@@ -38,8 +39,19 @@ import (
...
-	resp, err := http.Post(serviceUrl, "application/json", bytes.NewBuffer(input))
+func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, error) {
+	req, err := http.NewRequest("POST", serviceUrl, bytes.NewBuffer(input))
+	for _, h := range headersToPropagate {
+		_, exists := headers[h]
+		if exists {
+			for _, v := range headers[h] {
+				req.Header.Add(h, v)
+			}
+		}
```
yuzisun: this block of code can be simplified to following
```go
if values, ok := headers[h]; ok {
    for _, v := range values {
         req.Header.Add(h, v)
    }
}
```

yuzisun (commented):
Thanks @rachitchauhan43 ! a few last small comments then we are good to merge!

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
rachitchauhan43: Addressed

cmd/router/main.go
rachitchauhan43: Done

yuzisun:
Really nice work on this !! Thanks @rachitchauhan43 !

/lgtm
/approve

PR #2418, opened 2022-09-09
Retrieve SAS token

### LINKED ISSUE #2392: Azure Blob Storage to support shared access signatures(SAS)
/kind feature

1. Able to access Azure Storage resources using shared access signatures([SAS](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview)), as our user only have `container_uri` and `SAS token` as credentials.

1. [example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_authentication.py#L39) of using `SAS token` from azure.
2. KServe has `AZURE_STORAGE_ACCESS_KEY` [environment variable](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L465-L466), but it is only used for [Azure File Share](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L394). To support it, just reuse it to pass the `SAS token` similar to the their [example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_authentication.py#L39).

tjandy98:

Fixes #2392

**Type of changes**

- [ x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -345,9 +345,9 @@ class Storage(object):  # pylint: disable=too-few-public-methods
                      account_name,
                      container_name,
                      prefix)
-        token = Storage._get_azure_storage_token()
+        token = Storage._get_azure_storage_token() or Storage._get_azure_storage_access_key()
         if token is None:
-            logging.warning("Azure credentials not found, retrying anonymous access")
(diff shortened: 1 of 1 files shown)


yuzisun:
Thanks @tjandy98 !

/lgtm
/approve

PR #2448, opened 2022-10-03
Update the regex for endpoints to support models with `.` in them

### LINKED ISSUE #2447: Kserve endpoint does not work if modelname contains `.`
/kind bug

I deployed a model where I wanted to include the version of the model in the model name. This version was `1.0.1` so my model was `mycoolmodel-v1.0.1`. However this made reaching the endpoint for predict not work due to https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model_server.py#L100 not including `.` in the regex.

If `.` is not supported, it should have at least errored rather than looking like it deployed successfully. Ideally endpoints should work fine with models that include `.` in their name.

I think this could be fixed by just adding `\.` to the regex https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model_server.py#L100 so 
`([a-zA-Z0-9_-]+)` -> `([a-zA-Z0-9_-\.]+)`

**Environment:**

- Istio Version:
- Knative Version:
- KFServing Version: 0.9.0
- Kubeflow version:
- Minikube version:

f4hy:
Models with `.` in their name are not accessible  but no error is displayed. This should make such models accessible.

Fixes #2447 

**Type of changes**


yuzisun:
@f4hy I am closing this PR as it is no longer a limitation in 0.10 after we switched to FastAPI, but thanks for the PR!

https://github.com/kserve/kserve/blob/aed4d5b2740bd54912d92b7127318e47dfcf57eb/python/kserve/kserve/protocol/rest/server.py#L106

PR #2425, opened 2022-09-19
add prometheus metrics for kserve model server

alexagriffith:
**What this PR does / why we need it**:

Our users have come to us with a lot of questions regarding latencies in the `kserve-container`. As of now, we have no way to know the latencies of each step in the transformer/predictor. Having observability into the latency of each step will let us quickly identify which step is the bottleneck when request latencies are high. At the moment, there is no easy way to figure this out. 

This PR adds prometheus histogram latency metrics for each step (pre/post processing, explain, predict) in the python SDK. The metrics are exposed on /metrics. Additionally, we will log the latency of each step per request id. If you choose not to use prometheus metrics, evaluating the logs is another option. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Type of changes**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [x] Test documented in transformer sample https://github.com/kserve/kserve/pull/2425/files#diff-a8d78136ac31d86dd82d5b2d24410ca6eb48265aae92c5ea63e360d56dfdf58aR148 
I used the transformer to curl localhost:8080/metrics and see the prometheus metrics output. 

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [x] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

PR #2459, opened 2022-10-05
adding prometheus port configs for aggregating metrics in queue proxy

alexagriffith:
Signed-off-by: alexagriffith <agriffith50@bloomberg.net>

This PR follows https://github.com/kserve/kserve/pull/2425 where prometheus metrics were added in python sdk. 
This PR is the first of two that will allow the queue proxy to scrape metrics from the kserve-container. The reason this is needed is because of the limitations of knative only being able to scrape one port per pod. The end goal is to extend queue-proxy so that it scrapes metrics from kserve-container and aggregates them/exposes them on one port. A PoC for this already exists and. I am using it to test (this will be next PR). 

This PR's goal is to pass the kserve-container prometheus port into the queue-proxy so that queue-proxy can use it to scrape metrics and aggregate them. To do this, a new mutator is added. 

**Type of changes**

- [x] Unit test
- [x] Deployed each model and checked pod yaml and configs 

- Logs

**Checklist**:

**Release note**:

```release-note

```

### DIFF
--- a/config/runtimes/kserve-mlserver.yaml
+++ b/config/runtimes/kserve-mlserver.yaml
@@ -4,7 +4,9 @@ metadata:
   name: kserve-mlserver
 spec:
   annotations:
+    # mlserver version 1.1.0 uses port 8082 as default instead of 8080.
     prometheus.kserve.io/port: '8080'
+    prometheus.kserve.io/path: "/metrics"
   supportedModelFormats:
     - name: sklearn
       version: "0"
--- a/config/runtimes/kserve-mlserver.yaml
+++ b/config/runtimes/kserve-mlserver.yaml
@@ -3,6 +3,8 @@ kind: ClusterServingRuntime
 metadata:
   name: kserve-mlserver
 spec:
+  annotations:
+    prometheus.kserve.io/port: '8080'
   supportedModelFormats:
     - name: sklearn
       version: "0"
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -4,7 +4,7 @@ metadata:
   name: kserve-tensorflow-serving
 spec:
   annotations:
-    prometheus.kserve.io/port: '8501'
+    prometheus.kserve.io/port: '8080'
     prometheus.kserve.io/path: "/metrics"
   supportedModelFormats:
     - name: tensorflow
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -4,7 +4,7 @@ metadata:
   name: kserve-tensorflow-serving
 spec:
   annotations:
-    prometheus.kserve.io/port: '9153'
+    prometheus.kserve.io/port: '8080'
   supportedModelFormats:
     - name: tensorflow
       version: "1"
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -3,6 +3,8 @@ kind: ClusterServingRuntime
 metadata:
   name: kserve-tensorflow-serving
 spec:
+  annotations:
+    prometheus.kserve.io/port: '9153'
   supportedModelFormats:
     - name: tensorflow
       version: "1"
--- a/config/runtimes/kserve-tensorflow-serving.yaml
+++ b/config/runtimes/kserve-tensorflow-serving.yaml
@@ -4,7 +4,8 @@ metadata:
   name: kserve-tensorflow-serving
 spec:
   annotations:
-    prometheus.kserve.io/port: '8080'
+    prometheus.kserve.io/port: '8501'
+    prometheus.kserve.io/path: "/metrics"
   supportedModelFormats:
     - name: tensorflow
       version: "1"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -78,8 +78,13 @@ var (
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/maxScale"
 	RollOutDurationAnnotationKey                = KnativeServingAPIGroupName + "/rollout-duration"
 	EnableMetricAggregation                     = KServeAPIGroupName + "/enable-metric-aggregation"
-	KserveContainerPrometheusPort               = "prometheus.kserve.io/port"
-	SetPrometheusAggregateAnnotation            = KServeAPIGroupName + "/set-prometheus-aggregate-annotation"
+	SetPrometheusAggregateAnnotation            = KServeAPIGroupName + "/enable-prometheus-aggregate-scraping"
+	KserveContainerPrometheusPortKey            = "prometheus.kserve.io/port"
+	KServeContainerPrometheusPathKey            = "prometheus.kserve.io/path"
+	PrometheusPortAnnotationKey                 = "prometheus.io/port"
+	PrometheusPathAnnotationKey                 = "prometheus.io/path"
+	DefaultPrometheusPath                       = "/metrics"
+	QueueProxyAggregatePrometheusMetricsPort    = "9088"
 )
 
 // InferenceService Internal Annotations
@@ -179,11 +184,12 @@ const (
 
 // InferenceService Environment Variables
 const (
-	CustomSpecStorageUriEnvVarKey              = "STORAGE_URI"
-	CustomSpecProtocolEnvVarKey                = "PROTOCOL"
-	CustomSpecMultiModelServerEnvVarKey        = "MULTI_MODEL_SERVER"
-	KServeContainerPrometheusPortEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_PORT"
-	QueueProxyAggregatePrometheusPortEnvVarKey = "AGGREGATE_PROMETHEUS_PORT"
+	CustomSpecStorageUriEnvVarKey                     = "STORAGE_URI"
+	CustomSpecProtocolEnvVarKey                       = "PROTOCOL"
+	CustomSpecMultiModelServerEnvVarKey               = "MULTI_MODEL_SERVER"
+	KServeContainerPrometheusMetricsPortEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_METRICS_PORT"
+	KServeContainerPrometheusMetricsPathEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_METRICS_PATH"
+	QueueProxyAggregatePrometheusMetricsPortEnvVarKey = "AGGREGATE_PROMETHEUS_METRICS_PORT"
 )
 
 type InferenceServiceComponent string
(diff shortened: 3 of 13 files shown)


config/runtimes/kserve-tensorflow-serving.yaml
```diff
@@ -3,6 +3,8 @@ kind: ClusterServingRuntime
 metadata:
   name: kserve-tensorflow-serving
 spec:
+  annotations:
```
alexagriffith, line 6: the only one I haven't figured out - docs are a little confusing. seems we should use --rest_api_port (8080 for us) https://www.tensorflow.org/tfx/serving/serving_config  but when I hit that with /metrics or /monitoring/prometheus/metrics , I get a `404`

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/kserve/kserve/pkg/constants"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	defaultPrometheusPort          = "8080"
+	PrometheusPortAnnotationKey    = "prometheus.io/port"
+	PrometheusPathAnnotationKey    = "prometheus.io/path"
+	PrometheusPathAnnotationValue  = "/metrics"
+	QueueProxyAggregateMetricsPort = "9088"
```
alexagriffith: I can move this to `constants` , if we think that would be a better place for them?

alexagriffith (commented):
I need to add documentation about how to use 

```
annotations:
    serving.kserve.io/enable-metric-aggregation: "true"
    serving.kserve.io/set-prometheus-aggregate-annotation: "true"
```    

but with this PR, aggregation won't actually happen yet. It will in the next PR with the queue-proxy extension. Should I wait to add this documentation in the next PR? Also where is best to place this documentation?

config/runtimes/kserve-mlserver.yaml
```diff
@@ -3,6 +3,8 @@ kind: ClusterServingRuntime
 metadata:
   name: kserve-mlserver
 spec:
+  annotations:
+    prometheus.kserve.io/port: '8080'
```
yuzisun, line 8: Does MLServer use port 8080 too ?

pkg/webhook/admission/pod/metrics_aggregate_injector_test.go
```diff
@@ -0,0 +1,250 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2022 The KServe Authors.
```

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2022 The KServe Authors.
```

pkg/constants/constants.go
```diff
@@ -176,9 +179,11 @@ const (
 
 // InferenceService Environment Variables
 const (
-	CustomSpecStorageUriEnvVarKey       = "STORAGE_URI"
-	CustomSpecProtocolEnvVarKey         = "PROTOCOL"
-	CustomSpecMultiModelServerEnvVarKey = "MULTI_MODEL_SERVER"
+	CustomSpecStorageUriEnvVarKey              = "STORAGE_URI"
+	CustomSpecProtocolEnvVarKey                = "PROTOCOL"
+	CustomSpecMultiModelServerEnvVarKey        = "MULTI_MODEL_SERVER"
+	KServeContainerPrometheusPortEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_PORT"
```
yuzisun:
```suggestion
	KServeContainerPrometheusPortEnvVarKey     = "KSERVE_CONTAINER_METRICS_PORT"
```

yuzisun: Prometheus is one way of exposing metrics, might want to be agnostics of underlying technology here.

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -0,0 +1,61 @@
...
+	PrometheusPathAnnotationValue  = "/metrics"
+	QueueProxyAggregateMetricsPort = "9088"
+)
+
+// InjectMetricsAggregator looks for the annotations to enable aggregate kserve-container and queue-proxy metrics and
+// if specified, sets port-related EnvVars in queue-proxy and the aggregate prometheus annotation.
+func InjectMetricsAggregator(pod *v1.Pod) error {
+	for i, container := range pod.Spec.Containers {
+		if container.Name == "queue-proxy" {
+			if enableMetricAgg, ok := pod.ObjectMeta.Annotations[constants.EnableMetricAggregation]; ok && enableMetricAgg == "true" {
```
yuzisun, line 31: I think we want a global configuration here in the [inference service configmap](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml) instead of adding this annotation to each inference service. The annotation can be still used to override for the global value.

yuzisun: see https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging

pkg/constants/constants.go
alexagriffith: I think there are other metrics we can export though, do we want to be specific that this is just used for prometheus scraping?

pkg/webhook/admission/pod/metrics_aggregate_injector.go
alexagriffith: the `enbableMetricAgg` annotation? I think here it checks if the annotation exists and is set to "true". I dont think we add the annotation to each inference service here right? I can add to configmag if needed, but just wondering how adding the enable metric agg annotation will help with this? we will still have this if statement right?

alexagriffith: the prometheus docs link to the prometheus annotation, do you mean that we should add that one to configmap?

config/runtimes/kserve-mlserver.yaml
alexagriffith: good catch, I think I lumped it with the python ones. I think its 8082, checking.

alexagriffith: mlserver docs seems to suggest its using port 8082 https://mlserver.readthedocs.io/en/latest/reference/settings.html

but I tested it and 8082 doesn't emit prometheus metrics, but 8080/metrics does.

alexagriffith: tested using https://github.com/kserve/kserve/tree/master/docs/samples/v1beta1/xgboost and confirmed mlserver was used because it set 8082 as the env var

pkg/webhook/admission/pod/metrics_aggregate_injector.go
alexagriffith: right now default is false (or no annotation set), are you suggesting we change it to true? the only issue with that I think is the custom queue-proxy image. maybe we want to leave default as not present / false for now.

pkg/constants/constants.go
yuzisun: I think the name is a bit confusing, the port is for the metrics endpoint and prometheus is really the metrics format.

pkg/webhook/admission/pod/metrics_aggregate_injector.go
yuzisun: We should add to the configmap but still default to false, when user configures to be true then we enable the metrics merging in the mutator and set the env on queue proxy. Without adding to configmap user needs to add the `EnableMetricAggregation` annotation to every inference service yaml if we want to turn on this feature across board right?

When configmap is set to false, the if condition can be still kept to override the configmap default to enable the metrics merging if you want to turn on selective inference service.

pkg/constants/constants.go
```diff
@@ -77,6 +77,9 @@ var (
 	MinScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/minScale"
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/maxScale"
 	RollOutDurationAnnotationKey                = KnativeServingAPIGroupName + "/rollout-duration"
+	EnableMetricAggregation                     = KServeAPIGroupName + "/enable-metric-aggregation"
+	KserveContainerPrometheusPort               = "prometheus.kserve.io/port"
+	SetPrometheusAggregateAnnotation            = KServeAPIGroupName + "/set-prometheus-aggregate-annotation"
```
yuzisun: I think we can name it as `enable-prometheus-scraping`, wdyt?

pkg/webhook/admission/pod/metrics_aggregate_injector.go
yuzisun: We can add this in a separate PR though.

alexagriffith: So right now if the annotation is not present or it is set to false explicitly (not likely currently) it will not set the aggregation annotations. So at this point the annotation is only added to yaml if the inference service specifies it.

I see what you are saying, I can set a default to false, and every pod yaml will have this annotation where at the moment some pod yaml may not have it at all.

alexagriffith: Or *it won't set the aggregation env vars/ annotation

pkg/constants/constants.go
alexagriffith: Enable Prometheus scraping may be confusing because there is an annotation Prometheus.io/scrape: true that can be set and this is separate from that

yuzisun: Though this annotation does not sound like something for user to configure and worry about, we might want to avoid extra yaml UX burden for people.

config/runtimes/kserve-tensorflow-serving.yaml
yuzisun: Did you try 8501? See https://stackoverflow.com/questions/52975301/prometheus-for-tensorflow-serving

config/runtimes/kserve-mlserver.yaml
yuzisun: From mlserver maintainer, the longer version is that 1.0.0 exposed the metrics endpoint in the same HTTP server as the REST inference endpoints. But on 1.1.0 we decided to move that to a separate server (which runs on port 8082 by default).

yuzisun: From mlserver maintainer, the longer version is that 1.0.0 exposed the metrics endpoint in the same HTTP server as the REST inference endpoints. But on 1.1.0 we decided to move that to a separate server (which runs on port 8082 by default).

alexagriffith: Okay, so we are using earlier version?

config/runtimes/kserve-tensorflow-serving.yaml
alexagriffith: yes

pkg/webhook/admission/pod/metrics_aggregate_injector.go
alexagriffith: after discussing online: in the next PR this will be set in config map at a global level to give infra engineers the ability to set a global default for their services. since it doesn't actually scrape metrics until the next PR, it will be added then.

config/runtimes/kserve-mlserver.yaml
alexagriffith: added a comment in the runtime YAML about this.

pkg/constants/constants.go
alexagriffith: changing to `enable-prometheus-aggregate-scraping`

pkg/constants/constants.go
alexagriffith: changed

PR #2476, opened 2022-10-19
removed arch dependency for multiarc support

pranavpandit1:
These changes are targeted to remove the arch-specific dependency.
Also, these changes are verified for Intel and ppc64le.


yuzisun:
Thanks @pranavpandit1 for fixing this!! 

/lgtm
/approve

Issue #2465, opened 2022-10-07
merge model server metrics into queue-proxy 


alexagriffith:
/kind feature

**Describe the solution you'd like**
The ability to expose one port in queue-proxy that emits prometheus metrics from both queue-proxy and the kserve-container. 
To do this, we have added prometheus latency metrics to the python SDK model servers, and configured default settings/annotations for each of the serving runtimes. The queue-proxy extension can take in these configs, scrape the appropriate port and expose the metrics on a new queue-proxy port. 

See more about why prometheus can only scrape one port per pod in the issue https://github.com/knative/serving/issues/13299

alexagriffith:
adding prometheus metrics in python model server https://github.com/kserve/kserve/pull/2425

alexagriffith:
adding configs for queue-proxy scraping https://github.com/kserve/kserve/pull/2459

alexagriffith:
tensorflow issue that was realized from work in #2459 https://github.com/tensorflow/serving/issues/2067

PR #2478, opened 2022-10-19
Agriffith96/add queue proxy ext

alexagriffith:
**What qpext does** 
This directory handles extending the Knative queue-proxy sidecar container.

The qpext creates a new port in the queue-proxy container that scrapes Prometheus metrics from both `queue-proxy` and `kserve-container`. 
When the new aggregate metrics endpoint is hit, the response will contain metrics from both `queue-proxy` and `kserve-container`. 
The qpext adds the functionality to emit metrics from both containers on a single endpoint. 

**Why qpext is needed**
If an InferenceService uses Knative, then it has at least two containers in one pod, `queue-proxy` and `kserve-container`. A limitation of using Prometheus is that it supports scraping only one endpoint in the pod. 
When there are multiple containers in a pod that emit Prometheus metrics, this becomes an issue (see [Prometheus for multiple port annotations issue #3756](https://github.com/prometheus/prometheus/issues/3756) for the 
full discussion on this topic). In an attempt to make an easy-to-use solution, the queue-proxy is extended to handle this use case. 

see also: [KServe Issue #2645](https://github.com/kserve/kserve/issues/2465), 

Fixes #2645

**Type of changes**

- [x] unit tests 
- [x] e2e test
- [x]  ran in kind cluster

**Checklist**:


config/overlays/test/qpext_image_patch.yaml
```diff
@@ -0,0 +1,2 @@
+data:
```
alexagriffith: if we want this to be default, we will need to add it to the kustomize.yaml as a patch. not sure we want that though. we probably want it to be configurable in some way 🤔

yuzisun: We probably do not want this to be default at least for the initial version, also this Kustomize layout is only patching kserve not knative. What you need to patch is actually the knative configmap.

alexagriffith: yeah my b - we would want something similar if we wanted default - since we dont though its fine as is. in the e2e test it applies the patch in the knative namespace

pkg/webhook/admission/pod/mutator.go
```diff
@@ -76,6 +77,11 @@ func (mutator *Mutator) Handle(ctx context.Context, req admission.Request) admis
 	return admission.PatchResponseFromRaw(req.AdmissionRequest.Object.Raw, patch)
 }
 
+//TODO: should we use this? if so, replace panic.
+func UnmarshallError(configMapKey string, err error) error {
+	return fmt.Errorf("Unable to unmarshall %v json string due to %e ", configMapKey, err)
```
dangermike:
```suggestion
	return fmt.Errorf("Unable to unmarshall %v json string due to %w ", configMapKey, err)
```

https://goplay.tools/snippet/P4i5r2FHu-3

alexagriffith: woops good catch

qpext/cmd/qpext/main.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return 0 * time.Second, err
+	}
+
+	return time.Duration(timeoutSeconds * 1e9), nil
+}
+
+func applyHeaders(into http.Header, from http.Header, keys ...string) {
+	for _, key := range keys {
+		val := from.Get(key)
```
dangermike, line 76: Does this function need to support multi-valued headers? Also, if a key is present in both `from` and `into`, should you replace or merge?

dangermike: https://goplay.tools/snippet/jwB_ebVFZwS

qpext/cmd/qpext/main.go
```diff
@@ -0,0 +1,255 @@
...
+	return expfmt.FmtText
+}
+
+// scrapeAndWriteAgentMetrics gathers a slice of prometheus metric families and encodes the metrics.
+func scrapeAndWriteAgentMetrics(w io.Writer) error {
+	mfs, err := promRegistry.Gather()
+	enc := expfmt.NewEncoder(w, expfmt.FmtText)
+	if err != nil {
+		return err
+	}
```
dangermike: Check the error before doing more stuff
```suggestion
	if err != nil {
		return err
	}
	enc := expfmt.NewEncoder(w, expfmt.FmtText)
```

charts/kserve/templates/configmap.yaml
```diff
@@ -200,6 +200,11 @@ data:
   transformers: |-
     {
     }
+  metricsAggregator: |-
```
alexagriffith, line 203: @yuzisun hey do you happen to know if I really need this duplicated like four times? I just looked for where the config/configmap/inferenceservice.yaml values were and found them in this charts directory, not sure if I need it though bc when I tested it worked but maybe im missing something?

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -17,46 +17,91 @@ limitations under the License.
...
+	EnableMetricAggregation           string `json:"enableMetricAggregation"`
+	EnablePrometheusAggregateScraping string `json:"enablePrometheusAggregateScraping"`
+}
+
+func newMetricsAggregator(configMap *v1.ConfigMap) (*MetricsAggregator, error) {
+	ma := &MetricsAggregator{}
+
+	if maConfigVal, ok := configMap.Data[MetricsAggregatorConfigMapKeyName]; ok {
+		err := json.Unmarshal([]byte(maConfigVal), &ma)
+		//TODO: should I also panic here? seems weird that the other injectors do.
```
alexagriffith: I notice we panic in the other injectors after this unmarshal line of code but I don't get why because we can return err also. we could create a common error type and use that - so I guess my q is - is there a reason we want to panic after unmarshal instead of return err ? if so, should we remove the return err from each of these methods?

pkg/webhook/admission/pod/mutator.go
```diff
@@ -76,6 +77,11 @@ func (mutator *Mutator) Handle(ctx context.Context, req admission.Request) admis
 	return admission.PatchResponseFromRaw(req.AdmissionRequest.Object.Raw, patch)
 }
 
+//TODO: should we use this? if so, replace panic.
```
alexagriffith: related to prev comment

qpext/logger.go
```diff
@@ -0,0 +1,48 @@
...
+limitations under the License.
+*/
+
+package logger
+
+import (
+	"go.uber.org/zap"
+)
+
+type Logger struct {
```
dangermike: not sure why you need this struct, but you can avoid all the func definitions below by using this instead:

```go
type Logger = zap.Logger
```

And then your initialize becomes:

```go
func InitializeLogger() *Logger {
	zaplogger, _ := zap.NewProduction()
	return zaplogger
}
```

At that point, the `Info`, `Debug`, Error`, and `Fatal` functions defined below are no longer necessary

alexagriffith: thaaaaank you! I knew it should be easier

charts/kserve/templates/configmap.yaml
yuzisun: We need this for helm chart installation, KServe currently supports two installation methods: helm and kustomize.

test/qpext/test_qpext.py
```diff
@@ -0,0 +1,98 @@
...
+        namespace=KSERVE_TEST_NAMESPACE,
+        version="v2",
+    )
+    # temporary sleep until this is fixed https://github.com/kserve/kserve/issues/604
+    time.sleep(10)
+    cluster_ip = get_cluster_ip()
+    host = urlparse(isvc["status"]["url"]).netloc
+    headers = {"Host": host}
+
+    url = f"http://{cluster_ip}:{METRICS_AGG_PORT}/{METRICS_PATH}"
```
yuzisun: I think this will not work as `cluster_ip` is for the ingress gateway which does not listen on metrics agg port.

pkg/webhook/admission/pod/metrics_aggregate_injector.go
yuzisun: if we do not panic on bad configuration and continue we may leave the controller in a bad state depending on some unexpected configurations. We'd rather let the program crash and not continue.

alexagriffith: ok that is fine, we probably shouldn't have an err as a return param in the func then?

test/qpext/test_qpext.py
alexagriffith: yeah you are right, I just copy pasted what the other tests did without realizing. I can get the IP from the pod. Not sure if local host_ip will work though. I dont think we can port-forward with how the test is set up

.github/workflows/e2e-test.yml
```diff
@@ -103,7 +103,55 @@ jobs:
...
+          ./test/scripts/gh-actions/build-qpext-image.sh
+          docker image ls
+          cat ./config/overlays/test/configmap/inferenceservice.yaml
+      - name: Upload qpext image
+        uses: ishworkh/docker-image-artifact-upload@v1
+        with:
+          image: "kserve/qpext:${{ github.sha }}"
+      - name: Download qpext image
+          uses: ishworkh/docker-image-artifact-download@v1
+          with:
```
yuzisun: Wrong indentation here looks
https://github.com/kserve/kserve/actions/runs/3286615384

alexagriffith: grrr YAML - will fix

qpext/cmd/qpext/main.go
alexagriffith: good point - took me a second to make sure this wouldnt be an issue but it looks like http.NewRequestWithContext() a couple lines up creates a header map but doesn't put anything into it. So I think that we are always filling an empty header with the headers from the original request.

pkg/webhook/admission/pod/metrics_aggregate_injector.go
alexagriffith: I actually forgot to add a validation check here that can return err. so I will need it in the func

charts/kserve/templates/configmap.yaml
alexagriffith: ok got it, hope I have it in the right places lol

config/overlays/test/qpext_image_patch.yaml
```diff
@@ -0,0 +1,2 @@
...
+  queueSidecarImage: kserve/qpext:latest
```
yuzisun: We can remove this file now

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -17,46 +17,90 @@ limitations under the License.
...
+	EnablePrometheusScraping string `json:"enablePrometheusScraping"`
+}
+
+func newMetricsAggregator(configMap *v1.ConfigMap) (*MetricsAggregator, error) {
+	ma := &MetricsAggregator{}
+
+	if maConfigVal, ok := configMap.Data[MetricsAggregatorConfigMapKeyName]; ok {
+		err := json.Unmarshal([]byte(maConfigVal), &ma)
+		if err != nil {
+			panic(fmt.Errorf("Unable to unmarshall %v json string due to %v ", MetricsAggregatorConfigMapKeyName, err))
```
yuzisun, line 42:
```suggestion
			panic(fmt.Errorf("Unable to unmarshall %s json string due to %v ", MetricsAggregatorConfigMapKeyName, err))
```

python/kserve/requirements.txt
```diff
@@ -21,5 +21,6 @@ grpcio>=1.34.0
 tritonclient==2.18.0
 protobuf~=3.19.0
 prometheus-client>=0.13.1
+portforward
```
yuzisun: this looks like a bit risky dependency, can we only install it in the e2e test?

qpext/README.md
```diff
@@ -0,0 +1,166 @@
...
+|------------------------------------------|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| AGGREGATE_PROMETHEUS_METRICS_PORT        | 9088     | The metrics aggregation port in queue-proxy that is added in the qpext.                                                                                                         | 
+| KSERVE_CONTAINER_PROMETHEUS_METRICS_PORT | 8080     | The default metrics port for the `kserve-container`. If present, the default ClusterServingRuntime overrides this value with each runtime's default prometheus port.            |
+| KSERVE_CONTAINER_PROMETHEUS_METRICS_PATH | /metrics | The default metrics path for the `kserve-container`. If present, the default ClusterServingRuntime annotation overrides this value with each runtime's default prometheus path. |   
+
+To implement this feature, configure the InferenceService YAML annotations.
```
yuzisun: Do we support the case in which user just need to enable the aggregation in the configmap so that they do not need to add the annotation to the inference yaml?

config/overlays/test/qpext_image_patch.yaml
alexagriffith: I use this in the e2e test

python/kserve/requirements.txt
alexagriffith: yes was testing it

qpext/README.md
alexagriffith: yes but what is the best way to instruct people to change the config map? right now default is false. 

for docs we need to tell kserve maintainers the best way to implement that change and then tell users how to check defaults (just check configmap). the latter is straight forward, does the former require a patch? not sure if that is the recommended way or not

pkg/webhook/admission/pod/metrics_aggregate_injector.go
```diff
@@ -17,46 +17,90 @@ limitations under the License.
...
+	//Only set metric configs if the required annotations are set
+	enableMetricAggregation, ok := pod.ObjectMeta.Annotations[constants.EnableMetricAggregation]
+	if !ok {
+		if pod.ObjectMeta.Annotations == nil {
+			pod.ObjectMeta.Annotations = make(map[string]string)
+		}
+		pod.ObjectMeta.Annotations[constants.EnableMetricAggregation] = ma.EnableMetricAggregation
+	} else if enableMetricAggregation == "true" {
+		setMetricAggregationEnvVars(pod)
+	}
```
yuzisun: I think we also need to set the env vars when the annotation does not present but is enabled in the configmap
```go
if pod.ObjectMeta.Annotations[constants.EnableMetricAggregation] == "true" {
 	setMetricAggregationEnvVars(pod)
}

```

alexagriffith: this code does that! see the lines right above

alexagriffith: ah wait, I need to remove the "else" and just have "if"

alexagriffith: good point

alexagriffith: updated

config/overlays/test/qpext_image_patch.yaml
alexagriffith: changed it to using a the patch command with json instead of a file

python/kserve/requirements.txt
alexagriffith: moved

qpext/README.md
alexagriffith: updated - lmk what you think

qpext/README.md
```diff
@@ -0,0 +1,197 @@
...
+spec:
+  predictor:
+    sklearn:
+      protocolVersion: v2
+      storageUri: "gs://seldon-models/sklearn/iris"
+```
+The default port for sklearn runtime is `8080`, and the default path is `/metrics`. 
+By setting the annotations in the InferenceService YAML, the default runtime configurations are overridden.
+
+**KServe Developer's Note:** If the qpext is implemented in the cluster and you wish to set the default annotation values to `true`,
```
alexagriffith, line 87: @yuzisun added some more documentation here about the defaults

python/kserve/requirements.txt
```diff
@@ -17,9 +17,9 @@ avro>=1.11.0
 boto3==1.21.0
 psutil>=5.9.0
 ray[serve]==2.0.0
-grpcio>=1.34.0
+grpcio>=1.34.0Ï
```
yuzisun:
```suggestion
grpcio>=1.34.0
```

qpext/README.md
yuzisun: For people who installs kserve with helm chart they can change the helm variables you defined, with yaml installation the kserve admin would need to patch or kustomize the configmap before installing it.

pkg/webhook/admission/pod/metrics_aggregate_injector.go
yuzisun: I think you would need to use `pod.ObjectMeta.Annotations[constants.EnableMetricAggregation]` in the if condition as `enableMetricAggregation` is only set when the annotation exists, when annotation is absent you add the annotation with the value from configmap.

qpext/README.md
alexagriffith: Ok I think I added that info here- do you think I should be more explicit ?

pkg/webhook/admission/pod/metrics_aggregate_injector.go
alexagriffith: fixed

yuzisun:
Awesome work @alexagriffith !!

/lgtm
/approve

hack/image_patch_dev.sh
```diff
@@ -77,4 +77,9 @@ data:
         "cpuRequest": "100m",
         "cpuLimit": "100m"
     }
+    metricsAggregator: |-
```
rachitchauhan43, line 80: @alexagriffith @yuzisun : This `metricsAggregator` is improperly intended and hence it is getting inserted as part of router config and causing json marshalling errors in inference graph controller. 
Fixing it in https://github.com/kserve/kserve/pull/2539

PR #2508, opened 2022-11-04
Remove "default" suffix from generated component name - updated

### LINKED ISSUE #1397: Reason for still using "default" in revision name?

Suresh-Nakkeran:
This PR is to remove the suffix '-default' in inference service components.

Fixes #1397 

**Type of changes**

- I tested by using an existing predictor service and then updating the controller with this change to verify if the service is still working as expected

- I tested by using the existing service with both predictor and transformer and then updating the controller with this change to verify if the service is still working as expected

**Checklist**:

**Release note**:

```release-note
Remove the suffix '-default' in inference service component names for predictor, transformer and explainer ksvc or service. 
The change is only for new inference services and for existing services the behavior is unchanged.
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -51,14 +51,17 @@ type Predictor struct {
 	scheme                 *runtime.Scheme
 	inferenceServiceConfig *v1beta1.InferenceServicesConfig
 	credentialBuilder      *credentials.CredentialBuilder
+	deploymentMode         constants.DeploymentModeType
 	Log                    logr.Logger
 }
 
-func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceServiceConfig *v1beta1.InferenceServicesConfig) Component {
+func NewPredictor(client client.Client, scheme *runtime.Scheme, inferenceServiceConfig *v1beta1.InferenceServicesConfig,
+	deploymentMode constants.DeploymentModeType) Component {
 	return &Predictor{
 		client:                 client,
 		scheme:                 scheme,
 		inferenceServiceConfig: inferenceServiceConfig,
+		deploymentMode:         deploymentMode,
 		Log:                    ctrl.Log.WithName("PredictorReconciler"),
 	}
 }
@@ -252,6 +255,14 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	if err == nil {
 		predictorName = constants.DefaultPredictorServiceName(isvc.Name)
 	}
+
+	if p.deploymentMode == constants.RawDeployment {
+		existing := &v1.Service{}
+		err = p.client.Get(context.TODO(), types.NamespacedName{Name: constants.DefaultPredictorServiceName(isvc.Name), Namespace: isvc.Namespace}, existing)
+		if err == nil {
+			predictorName = constants.DefaultPredictorServiceName(isvc.Name)
+		}
+	}
 	// Labels and annotations from isvc will overwrite labels and annotations from ServingRuntimePodSpec
 	objectMeta := metav1.ObjectMeta{
 		Name:      predictorName,
@@ -264,18 +275,12 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	}
 
 	p.Log.Info("Resolved container", "container", container, "podSpec", podSpec)
-
-	deployConfig, err := v1beta1.NewDeployConfig(p.client)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-
 	var rawDeployment bool
 	var podLabelKey string
 	var podLabelValue string
 
 	// Here we allow switch between knative and vanilla deployment
-	if isvcutils.GetDeploymentMode(annotations, deployConfig) == constants.RawDeployment {
+	if p.deploymentMode == constants.RawDeployment {
 		rawDeployment = true
 		podLabelKey = constants.RawDeploymentAppLabel
 		r, err := raw.NewRawKubeReconciler(p.client, p.scheme, objectMeta, &isvc.Spec.Predictor.ComponentExtensionSpec,
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -19,6 +19,8 @@ package components
 import (
 	"fmt"
 
+	"context"
+
 	"github.com/go-logr/logr"
 	"github.com/kserve/kserve/pkg/constants"
 	"github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/reconcilers/knative"
@@ -31,6 +33,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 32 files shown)


config/configmap/inferenceservice.yaml
```diff
@@ -61,7 +61,7 @@ data:
         "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
         "ingressDomain"  : "example.com",
         "ingressClassName" : "istio",
-        "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
+        "domainTemplate": "{{ .Name }}.{{ .IngressDomain }}",
```
yuzisun: Any reason why namespace is removed here?

Peilun-Li:
Thanks for the work! Not an expert yet in the detailed controller logic so to confirm my understanding:
1. Existing isvc with "-default" suffix would still work with the new controller
2. With the new controller, new revision deployments to existing isvc with "-default" suffix would not automatically remove the "-default" suffix. To remove the "-default" suffix in the existing isvc, we'll need to first delete the isvc, then redeploy it again. (If that's the case, it means we'll have a short service downtime during the delete-redeploy. Is there any easy way we can automatically trim down the "-default" suffix upon new revision deployments? Could be fine though to admit that downtime if it's complicated to implement an automatic trim down)

I may be wrong here so definitely correct me!

Peilun-Li:
Follow up on this one: looks like automatically remove `default` suffix for existing isvc can be hard and I think the benefit of this PR would outperform the minor migration downtime. So LGTM overall!

yuzisun:
@Suresh-Nakkeran Can you help rebase this from master?

yuzisun:
> Follow up on this one: looks like automatically remove `default` suffix for existing isvc can be hard and I think the benefit of this PR would outperform the minor migration downtime. So LGTM overall!

Implementing the auto migration can be tricky and cause downtimes, so this won't help for existing inference services in the cluster but you can always recreate the isvc if you are ok with the downtimes, and for new inference services it should work as expected.

yuzisun:
@Suresh-Nakkeran Can you rebase from master?

Suresh-Nakkeran:
> @Suresh-Nakkeran Can you rebase from master?

sure

rachitchauhan43:
@yuzisun/ @Suresh-Nakkeran  : Won't this be a breaking change for inference graphs ? For inference graphs, we use serviceURLs like:

```apiVersion: serving.kserve.io/v1alpha1
kind: InferenceGraph
metadata:
  name: model-switch
spec:
  nodes:
    root:
      routerType: Switch
      steps:
      - condition: '[@this].#(country_code==IN)'
        serviceUrl: http://qbo-in-predictor-default.NS.svc.cluster.local/v1/models/qbo-in:predict
      - condition: '[@this].#(country_code==US)'
        serviceUrl: http://qbo-us-predictor-default.NS.svc.cluster.local/v1/models/qbo-us:predict
```

So, if we don't have top level VS disabled and still using serviceURLS like this then this will break all the deployed Inference Graphs , right ?

yuzisun:
> @yuzisun/ @Suresh-Nakkeran : Won't this be a breaking change for inference graphs ? For inference graphs, we use serviceURLs like:
> 
> ```
> kind: InferenceGraph
> metadata:
>   name: model-switch
> spec:
>   nodes:
>     root:
>       routerType: Switch
>       steps:
>       - condition: '[@this].#(country_code==IN)'
>         serviceUrl: http://qbo-in-predictor-default.NS.svc.cluster.local/v1/models/qbo-in:predict
>       - condition: '[@this].#(country_code==US)'
>         serviceUrl: http://qbo-us-predictor-default.NS.svc.cluster.local/v1/models/qbo-us:predict
> ```
> 
> So, if we don't have top level VS disabled and still using serviceURLS like this then this will break all the deployed Inference Graphs , right ?

It won't break the existing services with `default` suffix as we have checks for that and we only remove the suffix for newly created inference services, but ye for newly created ones the serviceUrl needs to be updated without `-default`

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
```diff
@@ -100,11 +100,11 @@ func generateMetadata(isvc *v1beta1api.InferenceService,
 	var name string
 	switch componentType {
 	case constants.Transformer:
-		name = constants.DefaultTransformerServiceName(isvc.Name)
+		name = constants.TransformerServiceName(isvc.Name)
```
yuzisun: @Suresh-Nakkeran will this change break the existing inference service deployed in raw kubernetes mode?

Suresh-Nakkeran: Yes @yuzisun .. I tested. It will break the existing inference service if it is deployed in raw kubernetes mode.

Suresh-Nakkeran: I am working on the fixes, will update the PR

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -298,7 +313,7 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	}
 	statusSpec, _ := isvc.Status.Components[v1beta1.PredictorComponent]
 	if rawDeployment {
-		podLabelValue = constants.GetRawServiceLabel(constants.DefaultPredictorServiceName(isvc.ObjectMeta.Name))
+		podLabelValue = constants.GetRawServiceLabel(constants.PredictorServiceName(isvc.ObjectMeta.Name))
```
yuzisun: Should here use `predictorName` ?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
```diff
@@ -183,43 +184,53 @@ func createRawIngress(scheme *runtime.Scheme, isvc *v1beta1api.InferenceService,
...
+			rules = append(rules, generateRule(host, predictorName, "/", constants.CommonDefaultHttpPort))
 		}
-		rules = append(rules, generateRule(explainerHost, constants.DefaultExplainerServiceName(isvc.Name), "/", constants.CommonDefaultHttpPort))
+		rules = append(rules, generateRule(explainerHost, explainerName, "/", constants.CommonDefaultHttpPort))
 	} else {
-		host, err := generateIngressHost(ingressConfig, isvc, string(constants.Predictor), true)
+		err := client.Get(context.TODO(), types.NamespacedName{Name: constants.DefaultPredictorServiceName(isvc.Name), Namespace: isvc.Namespace}, existing)
+		if err == nil {
+			predictorName = constants.DefaultPredictorServiceName(isvc.Name)
+		}
```
yuzisun, line 212: Is this needed ? seems `predictorName` can be used directly

Suresh-Nakkeran: This is required to check the predictor service already exists with the default suffix.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: Yes @yuzisun . You're correct! Let me update.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
Suresh-Nakkeran: @yuzisun updated!

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Suresh-Nakkeran: updated!

PR #2524, opened 2022-11-12
Added support to accept the request body/payload in any format (not just json)

### LINKED ISSUE #2400: custom predictor request body content-type is limited
/kind feature

The content-type of the request body in the kserve custom predictor seems to be limited to json
If I make a request in a non-Json format, there is an error, but I think it is because of the code below
kserve/handlers/predict.py
```python
  async def post(self, name: str):
        is_cloudevent = False
        is_binary_cloudevent = False

        if has_binary_headers(self.request.headers):
            is_cloudevent = True
            is_binary_cloudevent = True
            body = self.get_binary_cloudevent()
        else:
            try:
                body = json.loads(self.request.body)
            except json.decoder.JSONDecodeError as e:
                raise tornado.web.HTTPError(
                    status_code=HTTPStatus.BAD_REQUEST,
                    reason="Unrecognized request format: %s" % e
                )
                
```

If base64 encoding is performed to make the same torch sensor into json, the size will be increased as shown below
  - before base64 encoding
      - 350K
  - after base64 encoding
      - 466K

I would like to ask if there is a way to request a custom predictor in a format other than json
After base64 encoding, the request body size is increased, and that part is a bottleneck

SatishReddyBethi:
Currently, KServe only accepts http requests with body in `.json` format. This is alright if the model takes input as an `int`, `float` or `string`. However, if the model's input is an image (such as `.jpg` or `.png`), the current KServe API forces the client to convert the images from binary data to `base64` strings and wrap them as `json` before sending a request. From a developer's perspective, this takes away the freedom to decide a desired input format for their custom transformers.

Advantages:
- On the client side, it removes the additional steps of converting images/files to `base64` strings and wrapping it as `json` before sending a request to the server (*Note: if the developer wants to send images in `json`, this still allows that.*)
- The memory footprint of each request is much smaller because we are sending the binary data directly instead of converting the data to a `base64` string
- It makes testing endpoints much easier through API platforms such as Postman
- It lets the client send any file directly from a file system to the custom transformers as input data without throwing `Unrecognized request format` errors
- It lets the developer have control over the decisions on how the input data looks like and how to process that data through their custom transformers

Fixes #2400 

**Type of changes**

- [x] Test:
1. Create a transformer with a preprocessor that takes input as a `bytes`
2. Run a KServe server with the above transformer
3. Send a request to the server with body/payload as an image/file 
  - `curl -X POST http://localhost:8080/v1/models/<model-name>:predict -H 'Content-Type: image/png' --data-binary '@image.png'`
4. Verify that the image/file data stream is received on your transformer's preprocessor

**Release note**:

```
- Added support to accept the request body/payload in any format (not just json)
- Now, we are required to set `content-type` header to `application/json` for the server to recognize and decode the `json` type payload.
```

### DIFF
--- a/python/kserve/test/test_dataplane.py
+++ b/python/kserve/test/test_dataplane.py
@@ -120,7 +120,8 @@ class TestDataPlane:
 
     async def test_infer(self, dataplane_with_model):
         body = b'{"instances":[[1,2]]}'
-        resp = await dataplane_with_model.infer(self.MODEL_NAME, body)
+        headers = {"content-type": "application/json"}
+        resp = await dataplane_with_model.infer(self.MODEL_NAME, body, headers)
         assert resp == (
             {"predictions": [[1, 2]]},  # body
             {}  # headers
--- a/python/kserve/test/test_dataplane.py
+++ b/python/kserve/test/test_dataplane.py
@@ -120,8 +120,7 @@ class TestDataPlane:
 
     async def test_infer(self, dataplane_with_model):
         body = b'{"instances":[[1,2]]}'
-        headers = {"content-type": "application/json"}
-        resp = await dataplane_with_model.infer(self.MODEL_NAME, body, headers)
+        resp = await dataplane_with_model.infer(self.MODEL_NAME, body)
         assert resp == (
             {"predictions": [[1, 2]]},  # body
             {}  # headers
--- a/docs/samples/explanation/aif/germancredit/simulate_predicts.py
+++ b/docs/samples/explanation/aif/germancredit/simulate_predicts.py
@@ -7,8 +7,7 @@ if len(sys.argv) < 3:
     raise Exception("No endpoint specified. ")
 endpoint = sys.argv[1]
 headers = {
-    "Host": sys.argv[2],
-    "content-type": "application/json"
+    'Host': sys.argv[2]
 }
 
 with open('input.json') as file:
--- a/docs/samples/explanation/aif/germancredit/simulate_predicts.py
+++ b/docs/samples/explanation/aif/germancredit/simulate_predicts.py
@@ -7,7 +7,8 @@ if len(sys.argv) < 3:
     raise Exception("No endpoint specified. ")
 endpoint = sys.argv[1]
 headers = {
-    'Host': sys.argv[2]
+    "Host": sys.argv[2],
+    "content-type": "application/json"
 }
 
 with open('input.json') as file:
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -32,7 +32,7 @@ from .grpc import grpc_predict_v2_pb2 as pb
 import time
 import logging
 
-JSON_HEADERS = ["application/json"]
+JSON_HEADERS = ["application/json", "application/cloudevents+json", "application/ld+json"]
 
 
 class DataPlane:
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -209,12 +209,21 @@ class DataPlane:
 
     def decode(self, body, headers) -> Tuple[Union[Dict, InferRequest, CloudEvent], Dict]:
         t1 = time.time()
-        decoded_body = body
         attributes = {}
         if isinstance(body, InferRequest):
-            return decoded_body, attributes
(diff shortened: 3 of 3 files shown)


yuzisun:
@SatishReddyBethi Thanks for the PR! Can you help sign off the commit?

SatishReddyBethi:
@yuzisun Done. Thank you for mentioning.

python/kserve/kserve/handlers/dataplane.py
```diff
@@ -27,6 +27,7 @@
 from kserve.utils.utils import is_structured_cloudevent, create_response_cloudevent
 from ..constants import constants
 
+JSON_HEADERS = ["application/json"]
```
yuzisun: nit from lint:
```
./python/kserve/kserve/handlers/dataplane.py:32:1: E302 expected 2 blank lines, found 1
```

python/kserve/kserve/model.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, name: str):
         self._grpc_client_stub = None
         self.enable_latency_logging = False
 
-    async def __call__(self, body: Union[Dict, CloudEvent, ModelInferRequest],
+    async def __call__(self, body: Union[bytes, Dict, CloudEvent, ModelInferRequest],
```
yuzisun: Also need to update the type hint for the `preprocess` handler.

SatishReddyBethi: Fixed

python/kserve/kserve/handlers/dataplane.py
SatishReddyBethi: Add two blank lines after `JSON_HEADERS`

python/kserve/kserve/handlers/dataplane.py
```diff
@@ -233,7 +235,7 @@ async def infer(
             is_cloudevent = True
             is_binary_cloudevent = True
             body = self.get_binary_cloudevent(body, headers)
-        else:
+        elif headers["content-type"] in JSON_HEADERS or headers["Content-Type"] in JSON_HEADERS:
```
yuzisun: @SatishReddyBethi see the test failures here https://github.com/kserve/kserve/actions/runs/3502614095/jobs/5950435196, we need to add None check for `headers`

SatishReddyBethi: @yuzisun I have added the check. Can you rerun the test?

python/kserve/test/test_dataplane.py
```diff
@@ -118,7 +118,8 @@ async def test_model_metadata(self, dataplane_with_model):
 
     async def test_infer(self, dataplane_with_model):
         body = b'{"instances":[[1,2]]}'
-        resp = await dataplane_with_model.infer(self.MODEL_NAME, body)
+        headers = {"content-type": "application/json"}
```
yuzisun: @SatishReddyBethi This is a slight behavior change, let’s document in the release note that now we require application/json header to decode the json bytes.

yuzisun:
@SatishReddyBethi Can you help rebase from master ? There is a recent numpy 1.24 breaking change.

SatishReddyBethi:
@yuzisun My branch is now up-to-date with master.

python/kserve/test/test_dataplane.py
SatishReddyBethi: Updated the release notes

yuzisun:
Sorry @SatishReddyBethi, there are quite some recent refactoring which caused the conflicts, can you possibly rebase and resolve it?

SatishReddyBethi:
@yuzisun Merged my branch with master and resolved all conflicts

nhn-juhyeon:
is that work? i met `TypeError: Type is not JSON serializable: bytes` in https://github.com/kserve/kserve/blob/v0.11.2/python/kserve/kserve/model.py#L244

javierdlrm:
@yuzisun @SatishReddyBethi These [changes](https://github.com/kserve/kserve/commit/feea1354a09322eda477c8ee615b324df969e83e#diff-237f21ff5311ba6433c472e165b0dd873a997efaa61d1a33e02c1620f2fc041cR35) do not take into account the charset parameter in the content-type header.

Only the default charset UTF-8 is supported, by excluding the parameter entirely from the header. Other options such as "application/json; charset=UTF-8" or "application/json; charset=UTF-16" do not work.

PR #2542, opened 2022-11-18
[WIP - Do Not Merge]Initial Set of changes for step level dependency in Inference Graphs

### LINKED ISSUE #2484: Inference Graph’s response code should be influenced by node’s response code
/kind feature

Users should be able to define every node as either a hard dependency or a soft dependency in an Inference Graph

**Current situation:**

IG assumes that every node in the graph is a `soft dependency` including the last node.
By `soft dependency`, I mean IG's router doesn't halt IG's execution even if any node returns non-200. It continues executing the graph silently.

For example:
If we have an IG with 3 nodes  => { N1,N2,N3} in a `Sequence`, and now if the middle node, N2, returns non-200, IG will continue to call N3 with N2’s erroneous response body.
So, the user would have to make sure that the N3 node can handle that unexpected and erroneous response body from N2 properly.
Even if N3 (i.e. last node) handles it properly and returns 400 due to bad/unexpected request body, IG will still return 200 since the last node is also a soft dependency. And this behavior is not right.

So, basically IG’s response can not be influenced in any way and is always 200. 

Only exception is, IG will return 500 if something went wrong when IG router called its [node](https://github.com/kserve/kserve/blob/master/cmd/router/main.go#L53-L63). 
For example: Some http connection error that disrupted the connection.

To make sure we **do not release a breaking change**, we can rollout this feature keeping every node, including last one, as a soft dependency by default.

Although I feel last should be always a hard dependency. If the last node returns non-200, IG should never return 200. But for the sake of keeping changes non-breaking, we can keep all nodes configurable.

**Initial idea to implement:**
We can add a field in IG spec to mark the node as a `hard` or `soft` dependency to decide whether to stop IG execution right away or not.

- For nodes mentioned as `hard` dependency, IG's execution will stop if that node errors out
- For nodes mentioned as `soft` dependency, IG's execution will not stop and continue

rachitchauhan43:
Signed-off-by: rachitchauhan43 <rachitchauhan43@gmail.com>

Fixes #2484

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```


rachitchauhan43:
Closing in favor of https://github.com/kserve/kserve/pull/3039

PR #2551, opened 2022-11-20
Fix S3 Storage Initialiser downloading additional files when given file path does not include file extension

### LINKED ISSUE #2473: S3 storage initializer doesn't work as expected for file without extension
/kind bug

I created the following files in my s3 bucket:
```
test/churn-pickle
test/churn-pickle-logs
test/churn-pickle-report
```
`churn-pickle`, `churn-pickle-logs` and `churn-pickle-report` are `file` without `extension`.

I use storage initializer to download `s3://<bucket>/test/churn-pickle` to `/mnt/models`. However, I got the following result:

```
❯ ls /mnt/models
total 24
-rw-r--r--  1 i543026  staff    12B 14 Oct 19:25 -logs
-rw-r--r--  1 i543026  staff    12B 14 Oct 19:25 -report
-rw-r--r--  1 i543026  staff    12B 14 Oct 19:25 churn-pickle
```

1. if the given path is a `file`, only the file is downloaded. In this case, `s3://<bucket>/test/churn-pickle` will be downloaded to `/mnt/models/churn-pickle`.
2. If the given path is a `folder`, all the files under the folder will be downloaded. 
3. If the given path is a `prefix` (not a folder or file), nothing should be downloaded.

I assume that the storage initializer can only download from folders and files, not from `prefix`.

The issue was probably caused by:
1. objects are filter by the [prefix](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L167).
2. replace logic [here](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage.py#L167).

**Environment:**

- Istio Version:
- Knative Version:
- KFServing Version: 0.9
- Kubeflow version:
- Minikube version:

tjandy98:

Fixes #2473

**Type of changes**

### DIFF
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -181,7 +181,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             target_key = (
                 obj.key.rsplit("/", 1)[-1]
                 if bucket_path == obj.key
-                else obj.key.replace(bucket_path.rsplit("/",1)[0], "", 1).lstrip("/")
+                else obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
             )
             target = f"{temp_dir}/{target_key}"
             if not os.path.exists(os.path.dirname(target)):
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -180,7 +180,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             # (without any subpaths).
             if bucket_path == obj.key:
                 target_key = obj.key.rsplit("/", 1)[-1]
-            elif bucket_path.rsplit("/", 1)[-1] in obj.key.rsplit("/", 1)[-1]:
+            elif obj.key.rsplit("/", 1)[-1].startswith(bucket_path.rsplit("/", 1)[-1]):
                 target_key = obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
             else:
                 target_key = obj.key.replace(bucket_path, "").lstrip("/")
--- a/python/kserve/kserve/storage.py
+++ b/python/kserve/kserve/storage.py
@@ -181,7 +181,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             target_key = (
                 obj.key.rsplit("/", 1)[-1]
                 if bucket_path == obj.key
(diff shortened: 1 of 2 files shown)


python/kserve/kserve/storage.py
```diff
@@ -181,7 +181,7 @@ def _download_s3(uri, temp_dir: str):
             target_key = (
                 obj.key.rsplit("/", 1)[-1]
                 if bucket_path == obj.key
-                else obj.key.replace(bucket_path, "", 1).lstrip("/")
+                else obj.key.replace(bucket_path.rsplit("/",1)[0], "", 1).lstrip("/")
```
yuzisun: Can you help add a test ? Thanks !

tjandy98: @yuzisun I have added it, thank you.

yuzisun: Thanks ! there is a small lint issue and please also help update the PR with a meaningful title as that goes into the release notes.

python/kserve/kserve/storage.py
```diff
@@ -181,7 +181,7 @@ def _download_s3(uri, temp_dir: str):
             target_key = (
                 obj.key.rsplit("/", 1)[-1]
                 if bucket_path == obj.key
-                else obj.key.replace(bucket_path, "", 1).lstrip("/")
+                else obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
```
yuzisun: @tjandy98 would this break the case `s3://test-bucket/a` that bucket path `test-bucket/a` becomes `test-bucket` after `rsplit` and object key `test-bucket/a/b/model.bin` is replaced with `dir/a/b/model.bin` which is supposed to be `dir/b/model.bin?

yuzisun: >>> s = "test-bucket/a/b/model.bin"
>>> bucket_path = "test-bucket/a"
>>> s.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
'a/b/model.bin'
>>> s.replace(bucket_path, "", 1).lstrip("/")
'b/model.bin'

tjandy98: @yuzisun You are right, thanks for providing an example! I have made some changes and added a test based on this.

yuzisun:
fixed in https://github.com/kserve/kserve/pull/3088

PR #2569, opened 2022-11-28
Handle scenario where annotations are not given in raw deployment mode

### LINKED ISSUE #2560: Cannot run InferenceService in RawDeployment mode without specifying some annotations
/kind bug

Configure kserve 0.9 in RawDeployment mode. Run an InferenceService with `metadata: {annotations: null}`.

The mutating web hook fails admission because https://github.com/kserve/kserve/blob/8fdd82fe109967efe94b7a012f07f9d75ab2eb60/pkg/apis/serving/v1beta1/inference_service_defaults.go#L86 doesn't check for nil annotations map before trying to setup a new annotation.

Kserve to be able to deal with an InferenceService in RawDeployment mode without annotations.

From 
```yaml
kind: "InferenceService"
metadata:
  generateName: inference-service-v2-
  labels:
    inference-service-test.cambridgeconsultants.com: "true"
  annotations:
    hello: ok
spec:
  predictor:
    triton:
      storageUri: "gs://kfserving-examples/models/torchscript"
      runtimeVersion: 20.10-py3
      env:
      - name: OMP_NUM_THREADS
        value: "1"
```

The workaround is to simply add a dummy annotation. Unfortunately `annotations: {}` didn't work. 🤷

**Environment:**

- KServe Version: 0.9
- Cloud Environment: EKS
- OS (e.g. from `/etc/os-release`): MacOs

andyi2it:
Signed-off-by: Andrews Arokiam <andrews.arokiam@ideas2it.com>

Adds a nil check for annotation and creates one if it is not provided.
Adds a test case for the above change and improves the overall coverage

Fixes #2560

**Type of changes**

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_defaults.go
+++ b/pkg/apis/serving/v1beta1/inference_service_defaults.go
@@ -86,6 +86,9 @@ func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesC
 	if !ok && deployConfig != nil {
 		if deployConfig.DefaultDeploymentMode == string(constants.ModelMeshDeployment) ||
 			deployConfig.DefaultDeploymentMode == string(constants.RawDeployment) {
+			if isvc.ObjectMeta.Annotations == nil {
+				isvc.ObjectMeta.Annotations = map[string]string{}
+			}
 			isvc.ObjectMeta.Annotations[constants.DeploymentMode] = deployConfig.DefaultDeploymentMode
 		}
 	}
--- a/pkg/apis/serving/v1beta1/inference_service_defaults_test.go
+++ b/pkg/apis/serving/v1beta1/inference_service_defaults_test.go
@@ -31,63 +31,290 @@ import (
 
 func TestInferenceServiceDefaults(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
-	config := &InferenceServicesConfig{
-		Explainers: ExplainersConfig{
-			AlibiExplainer: ExplainerConfig{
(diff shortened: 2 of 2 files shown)


pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -80,6 +80,9 @@ func (isvc *InferenceService) Default() {
 }
 
 func (isvc *InferenceService) DefaultInferenceService(config *InferenceServicesConfig, deployConfig *DeployConfig) {
+	if isvc.ObjectMeta.Annotations == nil {
+		isvc.ObjectMeta.Annotations = map[string]string{}
```
yuzisun: only initialize the annotation map when we add some annotations, can move this into the if block. This way we can still keep the nil semantics that there is no annotations.

pkg/apis/serving/v1beta1/inference_service_defaults_test.go
```diff
@@ -31,63 +31,290 @@ import (
 
 func TestInferenceServiceDefaults(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
-	config := &InferenceServicesConfig{
-		Explainers: ExplainersConfig{
-			AlibiExplainer: ExplainerConfig{
-				ContainerImage:      "alibi",
-				DefaultImageVersion: "v0.4.0",
+	scenarios := map[string]struct {
```
yuzisun, line 34: Thanks for adding the tests!

PR #2588, opened 2022-12-08
Added github action to test kserve helm

### LINKED ISSUE #2570: Add e2e test for KServe helm chart
/kind feature

- Create a separate github action to install KServe with Istio helm chart and KServe helm chart
- Add e2e test for ModelMesh and classic inference service

Suresh-Nakkeran:

As part of this PR, a github action has been added to make sure the inference e2e tests are working fine with kserve helm chart installation. 
Fixes #2570


test/e2e/common/utils.py
```diff
@@ -232,3 +231,19 @@ def predict_grpc(service_name, payload, version=constants.KSERVE_V1BETA1_VERSION
         options=(('grpc.ssl_target_name_override', host),))
     stub = grpc_predict_v2_pb2_grpc.GRPCInferenceServiceStub(channel)
     return stub.ModelInfer(pb.ModelInferRequest(model_name=model_name, inputs=payload))
+
+
+def predict_modelmesh(service_name, input_json, model_name=None):
+    with open(input_json) as json_file:
+        data = json.load(json_file)
+        cluster_ip = get_cluster_ip("modelmesh-serving", "default")
```
yuzisun: this is internal service which does not have cluster ip, we might need to do port forward here.

test/scripts/gh-actions/setup-kserve-helm.sh
```diff
@@ -0,0 +1,51 @@
...
+cat ./charts/kserve-resources/values.yaml
+
+make deploy-helm
+
+echo "Updating modelmesh default replicas count..."
+kubectl patch servingruntimes mlserver-0.x --type='merge' -p '{"spec":{"replicas":1}}'
+
+echo "Waiting for KServe started ..."
+kubectl wait --for=condition=Ready pods --all --timeout=180s
+kubectl get events -A
```
yuzisun: Can we use helm `--wait` ? see https://helm.sh/docs/helm/helm_install/#options

Suresh-Nakkeran: okay Dan

test/e2e/common/utils.py
```diff
@@ -233,3 +233,15 @@ def predict_grpc(service_name, payload, version=constants.KSERVE_V1BETA1_VERSION
...
     return stub.ModelInfer(pb.ModelInferRequest(model_name=model_name, inputs=payload))
+
+
+def predict_modelmesh(service_name, input_json, pod_name, model_name=None):
+    with open(input_json) as json_file:
+        data = json.load(json_file)
+
+        if model_name is None:
+            model_name = service_name
+        with portforward.forward("default", pod_name, 8008, 8008):
```
yuzisun, line 244: I think we should port-forward the service instead ? pod works too but it may not necessarily test the routing and setup. 
https://github.com/kserve/modelmesh-serving/tree/main/docs/predictors#using-the-deployed-model

test/e2e/helm/test_kserve_sklearn.py
```diff
@@ -0,0 +1,72 @@
+#
```
yuzisun, line 2:
```suggestion
# Copyright 2022 The KServe Authors.
#
```

test/e2e/common/utils.py
Suresh-Nakkeran: @yuzisun In model-mesh e2e test we have updated modelmesh controller replica size to 1, so it would be having only one pod. The forward method in portforward module expects namespace, pod name. I see other examples outside also uses the pod name not service name. your thoughts?

yuzisun:
Great work! Thanks @Suresh-Nakkeran !!

/lgtm
/approve

PR #2589, opened 2022-12-08
adding prom tags and tests

alexagriffith:

upon building a grafana dashboard from the prometheus metrics, I noticed we don't have the tags we add on from knative/queueproxy. if someone is using these metrics, these tags will most likely be wanted to be able to run queries using the service name and configuration and revision names. these tags make the dashboards much easier to make and much more useful. Here I add the `SERVING_SERVICE`, `SERVING_CONFIGURATION`, and `SERVING_REVISION` environment values from queue-proxy to the prom metric labels. this part is done in qpext because only if in serverless mode we should add them.

model_name is a common label used in other serving runtimes (triton/torchserve). we should be consistent with them and add this as well. it also adds value for building dashboards/querying metrics. 

**Type of changes**

- [ x] New feature (non-breaking change which adds functionality)

- [ x] Test A added in unit tests
- [x ] Test B visualized in grafana testing a pod on kubernetes 

- Logs

**Checklist**:


python/kserve/kserve/metrics.py
```diff
@@ -0,0 +1,20 @@
+from prometheus_client import Histogram
+import os
+
+PROM_KEYS = {"K_SERVICE": "service_name", "K_CONFIGURATION": "configuration_name",
```
sukumargaonkar: how are these env vars inserted?
by kserve controller?

alexagriffith: yes, if serverless, kserve-container will have them

python/kserve/test/test_metrics.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# coding: utf-8
+
+import unittest
```
yuzisun: KServe primarily use `pytest` for the unit testing, so let's be consistent with the other tests.

python/kserve/test/test_metrics.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
# Copyright 2022 The KServe Authors.
```

python/kserve/kserve/metrics.py
yuzisun: for raw deployment mode, we do not have configuration and revision but we probably still need service and deployment name.

python/kserve/test/test_metrics.py
alexagriffith: okay I actually took this from one of the other tests in the file

alexagriffith: maybe it was one randomly using unittest

python/kserve/kserve/metrics.py
alexagriffith: with this code, we will still add them? or are you suggesting a change?

python/kserve/test/test_metrics.py
alexagriffith: I updated it to remove that use case of unittest but I added a mock, seems safer than actually editing an env var.

qpext/cmd/qpext/main.go
```diff
@@ -88,16 +93,25 @@ func negotiateMetricsFormat(contentType string) expfmt.Format {
 	return expfmt.FmtText
 }
 
-// scrapeAndWriteAgentMetrics gathers a slice of prometheus metric families and encodes the metrics.
-func scrapeAndWriteAgentMetrics(w io.Writer) error {
-	mfs, err := promRegistry.Gather()
-	if err != nil {
-		return err
```
alexagriffith, line 95: istio has agent, we don't ... so this basically did nothing but the code is useful for what we need to do now

qpext/cmd/qpext/main.go
```diff
@@ -37,6 +38,8 @@ import "knative.dev/serving/pkg/queue/sharedmain"
 
 var (
 	promRegistry *prometheus.Registry
+	EnvVars      = []string{"SERVING_SERVICE", "SERVING_CONFIGURATION", "SERVING_REVISION"}
+	LabelKeys    = []string{"service_name", "configuration_name", "revision_name"}
```
alexagriffith: slices instead of maps to keep order of label insertion - otherwise testing is a pain

qpext/cmd/qpext/main.go
```diff
@@ -192,6 +239,7 @@ func (sc *ScrapeConfigurations) handleStats(w http.ResponseWriter, r *http.Reque
 	w.Header().Set("Content-Type", string(format))
 
 	// Write out the metrics
+	//TODO: do we need this?
```
alexagriffith: we originally took inspo from istio - but when I tested, this doesn't do anything? its an empty []

python/kserve/kserve/metrics.py
```diff
@@ -0,0 +1,12 @@
...
+import os
+
+PROM_LABELS = ['model_name']
+PRE_HIST_TIME = Histogram('request_preprocess_seconds', 'pre-process request latency', PROM_LABELS)
+POST_HIST_TIME = Histogram('request_postprocess_seconds', 'post-process request latency', PROM_LABELS)
+PREDICT_HIST_TIME = Histogram('request_predict_seconds', 'predict request latency', PROM_LABELS)
+EXPLAIN_HIST_TIME = Histogram('request_explain_seconds', 'explain request latency', PROM_LABELS)
+
+
+def get_labels(model_name):
```
alexagriffith, line 10: we can add more in future if we want, for now this is just cleaner to have so we can insert like **prom_labels in models.py

qpext/cmd/qpext/main.go
```diff
@@ -171,51 +271,48 @@ func (sc *ScrapeConfigurations) handleStats(w http.ResponseWriter, r *http.Reque
...
-		// Without app metrics format use a default
-		format = expfmt.FmtText
 	}
 
+	// Since we convert the scraped metrics to text, set the format as text even if
+	// the content type is originally open metrics.
+	format := expfmt.FmtText
 	w.Header().Set("Content-Type", string(format))
 
-	// Write out the metrics
```
alexagriffith, line 194: removed this, we aren't using it

qpext/cmd/qpext/main.go
```diff
@@ -156,8 +247,17 @@ func (sc *ScrapeConfigurations) handleStats(w http.ResponseWriter, r *http.Reque
 	var queueProxyCancel, appCancel context.CancelFunc
 
 	defer func() {
+		if queueProxy != nil {
```
alexagriffith, line 245: after the pod has been up for a while, I will see quite a few context cancellation errors and im not sure why - if you see anything with the way im using cancel/close let me know. so far, i dont see why its happening

alexagriffith: actually I think this is just happened with the current master version of qpext - will monitor again to confirm. I think these changes /should/ fix that.

qpext/cmd/qpext/main.go
```diff
@@ -80,24 +81,109 @@ func applyHeaders(into http.Header, from http.Header, keys ...string) {
...
+		var newMetric []*io_prometheus_client.Metric
+		for _, metric := range mf.Metric {
+			newMetric = append(newMetric, &io_prometheus_client.Metric{
+				Label: metric.Label,
+				Counter: &io_prometheus_client.Counter{
+					Value: metric.Untyped.Value,
+				},
+				TimestampMs: metric.TimestampMs,
+			})
+		}
```
yuzisun, line 123: This block and the block from line 134-143 can be extracted out of the if condition and shared between the two if checks.

qpext/cmd/qpext/main.go
```diff
@@ -134,6 +219,7 @@ func scrape(url string, header http.Header, logger *zap.Logger) (io.ReadCloser,
 		return nil, cancel, "", fmt.Errorf("error scraping %s: %v", url, err)
 	}
 	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
```
yuzisun, line 222: hmm seems like we are not closing the `resp.Body` for the normal case after done reading the body?

qpext/cmd/qpext/main.go
```diff
@@ -116,7 +202,6 @@ func scrape(url string, header http.Header, logger *zap.Logger) (io.ReadCloser,
 			logger.Error("Failed to parse timeout header", zap.Error(err), zap.String("timeout", timeoutString))
 		} else {
 			ctx, cancel = context.WithTimeout(ctx, timeout)
-			defer cancel()
```
yuzisun, line 119: I think we still need this, failing to call the cancel func leaks the child resources.

yuzisun: oh actually just realized that it is called outside of the `scrape` func

alexagriffith: yeah I think this was causing ctx cancel issues

qpext/cmd/qpext/main.go
alexagriffith: yeah I was analyzing how istio does it https://github.com/istio/istio/blob/c8a7356b99ccbd51ef728148845c0cabfffcc3a6/pilot/cmd/pilot-agent/status/server.go#L637 seems like they just call close explicitly when http status != OK

qpext/cmd/qpext/main.go
alexagriffith: the types are different one is counter one is gauge

alexagriffith: but I can make a helper method out of it Ithink

qpext/cmd/qpext/main.go
alexagriffith: I think resp body closes when it is done reading to Reader

qpext/cmd/qpext/main.go
alexagriffith: discussed - probably not saving much code/readability with a helper method here

PR #2600, opened 2022-12-15
Removing GOARCH dependency for multiarc support.

pranavpandit1:
These changes are targeted for multiple architectures compatible base images


pranavpandit1:
False PR, changes already upstreamed

PR #2616, opened 2022-12-22
Update to v0.10.0-rc1 release manifests

yuzisun:
Signed-off-by: Dan Sun <dsun20@bloomberg.net>

We have decided to delay the v0.10.0 release to first week of Jan and publish as rc1 instead, thus removing the v0.10.0 manifests generated in https://github.com/kserve/kserve/pull/2615


hack/generate-install.sh
```diff
@@ -65,8 +66,8 @@ kustomize build config/runtimes | sed s/:latest/:$TAG/ >> $RUNTIMES_INSTALL_PATH
 yq -i 'select(.metadata.name == "inferenceservice-config").data.ingress |= (fromjson | .ingressGateway = "kubeflow/kubeflow-gateway" | tojson)' $KUBEFLOW_INSTALL_PATH
 
 # Copy CRD files to charts crds directory
-cp config/crd/serving.kserve.io_clusterservingruntimes.yaml charts/kserve-crd/serving.kserve.io_clusterservingruntimes.yaml
-cp config/crd/serving.kserve.io_inferenceservices.yaml charts/kserve-crd/serving.kserve.io_inferenceservices.yaml
-cp config/crd/serving.kserve.io_trainedmodels.yaml charts/kserve-crd/serving.kserve.io_trainedmodels.yaml
-cp config/crd/serving.kserve.io_inferencegraphs.yaml charts/kserve-crd/serving.kserve.io_inferencegraphs.yaml
-cp config/crd/serving.kserve.io_servingruntimes.yaml charts/kserve-crd/serving.kserve.io_servingruntimes.yaml
+cp config/crd/serving.kserve.io_clusterservingruntimes.yaml charts/kserve-crd/templates/serving.kserve.io_clusterservingruntimes.yaml
```
yuzisun, line 69: fix the script that it should copy to `templates` directory instead for `kserve-crd` chart.

PR #2626, opened 2023-01-01
fix: Incorrect URL when making predictions via curl

terrytangyuan:
This fixes the following error:
```
curl: (3) URL using bad/illegal format or missing URL
```

Signed-off-by: Yuan Tang <terrytangyuan@gmail.com>


yuzisun:
thanks @terrytangyuan !

/lgtm
/approve

PR #2629, opened 2023-01-03
Add Unified Inference Type and refactor REST/gRPC server code

yuzisun:
* Add unified inference type for REST/gRPC protocol so that in custom model user can use the same type to program the `preprocess`,`predict`,`postprocess` handlers
* Change to use relative imports
* Refactor decode/encode functions
* Refactor and separate out `rest` and `grpc` server code

Fixes #

**Type of changes**

- [X] Tested the v2 REST/gRPC custom example
- [X] Tested the v2 REST/gRPC custom transformer example

- Logs

**Checklist**:

- [X] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
For CloudEvent case the response ce-source is updated to kserve.io.inference.$model
```


python/kserve/kserve/protocol/grpc/server.py
```diff
@@ -12,35 +12,34 @@
...
+from .interceptors import LoggingInterceptor
+from .servicer import InferenceServicer
+from kserve.protocol.dataplane import DataPlane
+from kserve.protocol.model_repository_extension import ModelRepositoryExtension
 
 from grpc import aio
 
 
-MAX_GRPC_MESSAGE_LENGTH = 8388608
+MAX_GRPC_MESSAGE_LENGTH = -1
```
yuzisun: Can we  keep it unlimited, what's the downside?

python/kserve/kserve/model.py
```diff
@@ -218,23 +222,36 @@ async def preprocess(self, payload: Union[Dict, CloudEvent, ModelInferRequest],
...
         """
         if headers:
-            if "grpc" in headers.get("user-agent", "") and isinstance(response, ModelInferResponse):
-                return response
-            if "application/json" in headers.get("content-type", "") and isinstance(response, ModelInferResponse):
-                return convert_grpc_response_to_dict(response)
+            if "grpc" in headers.get("user-agent", ""):
+                if isinstance(response, ModelInferResponse):
+                    return response
+                elif isinstance(response, InferResponse):
```
yuzisun, line 244: predict handler can return the `InferResponse` and the postprocess default handler converts to gRPC or REST based on the request type.

python/kserve/kserve/protocol/grpc/servicer.py
```diff
@@ -76,20 +77,28 @@ async def ModelMetadata(
...
     async def ModelInfer(
         self, request: pb.ModelInferRequest, context: ServicerContext
     ) -> pb.ModelInferResponse:
         headers = to_headers(context)
-        response_body, _ = await self._data_plane.infer(body=request, headers=headers, model_name=request.model_name)
+
+        infer_inputs = [InferInput(name=input_tensor.name, shape=input_tensor.shape,
+                                   datatype=input_tensor.datatype, data=input_tensor.contents)
+                        for input_tensor in request.inputs]
+        infer_request = InferRequest(model_name=request.model_name, infer_inputs=infer_inputs,
```
yuzisun: Construct the `InferRequest` from gRPC `ModelInferRequest` and pass to the model handlers, so user can program against the same type as REST, the framework handles the REST/gRPC decoding and encoding.

python/kserve/kserve/protocol/rest/v1_endpoints.py
```diff
@@ -70,10 +70,9 @@ async def predict(self, model_name: str, request: Request) -> Union[Response, Di
 
         if not isinstance(response, dict):
             return Response(content=response, headers=response_headers)
-
         return response
 
-    async def explain(self, model_name: str, request: Request) -> Dict:
+    async def explain(self, model_name: str, request: Request) -> Union[Response, Dict]:
```
yuzisun, line 75: align with predict handler with the same interface.

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -0,0 +1,568 @@
...
+        raw_output_contents = []
+        for infer_output in self.outputs:
+            if isinstance(infer_output.data, numpy.ndarray):
+                infer_output.set_data_from_numpy(infer_output.data, binary_data=True)
+            infer_output_dict = {
+                "name": infer_output.name,
+                "shape": infer_output.shape,
+                "datatype": infer_output.datatype,
+            }
+            if infer_output._raw_data is not None:
```
yuzisun, line 530: Do not set data field if the raw data is not None

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -0,0 +1,568 @@
...
+
+                """
+        infer_outputs = []
+        for infer_output in self.outputs:
+            infer_output_dict = {
+                "name": infer_output.name,
+                "shape": infer_output.shape,
+                "datatype": infer_output.datatype
+            }
+            if isinstance(infer_output.data, numpy.ndarray):
```
yuzisun, line 501: for np array set the parameter `content-type` to numpy array so that the server side can decode it to np array ?

python/kserve/kserve/errors.py
```diff
@@ -25,11 +25,16 @@ def __str__(self):
 
 
 class InferenceError(RuntimeError):
-    def __init__(self, reason):
+    def __init__(self, reason, status=None, debug_details=None):
         self.reason = reason
+        self.status = status
+        self.debug_details = debug_details
```
alexagriffith, line 31: 👍

python/kserve/kserve/inference_client.py
```diff
@@ -0,0 +1,103 @@
...
+        self._channel.close()
+
+    def infer(self,
+              infer_request: InferRequest,
+              client_timeout=None,
+              headers=None):
+        if headers is not None:
+            metadata = headers.items()
+        else:
+            metadata = ()
```
alexagriffith:
```suggestion
metadata = headers.items() if headers is not None else tuple()
```

alexagriffith: nit but a little cleaner imo - whatever u think

python/kserve/kserve/inference_client.py
```diff
@@ -0,0 +1,103 @@
...
+        if self._verbose:
+            print("infer, metadata {}\n{}".format(metadata, request))
+
+        try:
+            response = self._client_stub.ModelInfer(
+                request=request,
+                metadata=metadata,
+                timeout=client_timeout)
+            if self._verbose:
+                print(response)
```
alexagriffith: just curious why do we print instead of log here?

python/kserve/kserve/model.py
```diff
@@ -183,18 +187,18 @@ def get_output_types(self) -> List[Dict]:
         # return [{ "name": "", "datatype": "INT32", "shape": [1,5], }]
         return []
 
-    async def preprocess(self, payload: Union[Dict, CloudEvent, ModelInferRequest],
-                         headers: Dict[str, str] = None) -> Union[Dict, ModelInferRequest]:
+    async def preprocess(self, payload: Union[Dict, CloudEvent, InferRequest],
```
alexagriffith, line 190: 👍 I like the rename

python/kserve/kserve/model.py
```diff
@@ -218,23 +222,36 @@ async def preprocess(self, payload: Union[Dict, CloudEvent, ModelInferRequest],
 
         return response
 
-    def postprocess(self, response: Union[Dict, ModelInferResponse], headers: Dict[str, str] = None) -> Dict:
-        """The postprocess handler can be overridden for inference response transformation
+    def postprocess(self, response: Union[Dict, InferResponse, ModelInferResponse], headers: Dict[str, str] = None) \
+            -> Union[Dict, ModelInferResponse]:
+        """The postprocess handler can be overridden for inference response transformation.
+        The default implementation converts the v2 infer response types to gRPC or REST.
+        For gRPC request it converts InferResponse to gRPC message or directly returns ModelInferResponse from
```
alexagriffith, line 229: do you mean to have InferRequest here instead of ModelInferRequest since the name change?

alexagriffith: oh wait I see there are two types now.. I see.. Im thinking the naming is a little confusing but right now I can't think of something better so I will keep thinking

python/kserve/kserve/model_server.py
```diff
@@ -212,71 +120,31 @@ def start(self, models: Union[List[Model], Dict[str, Deployment]]) -> None:
...
+            self.max_asyncio_workers = min(32, utils.cpu_count() + 4)
         logging.info(f"Setting max asyncio worker threads as {self.max_asyncio_workers}")
         asyncio.get_event_loop().set_default_executor(
             concurrent.futures.ThreadPoolExecutor(max_workers=self.max_asyncio_workers))
 
         async def serve():
             serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             serversocket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-            serversocket.bind((cfg.host, cfg.port))
+            serversocket.bind(('0.0.0.0', self.http_port))
```
alexagriffith, line 133: is it ok to convert this to being hard coded?

alexagriffith: not sure if this was for testing so just want to check

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -200,12 +205,49 @@ def model_ready(self, model_name: str) -> bool:
...
+        if headers and has_binary_headers(headers):
+            body = self.get_binary_cloudevent(body, headers)
+        else:
+            if type(body) is bytes:
+                try:
+                    body = orjson.loads(body)
+                except orjson.JSONDecodeError as e:
+                    raise InvalidInput(f"Unrecognized request format: {e}")
+        t2 = time.time()
+        logging.info(f"decode takes {round((t2 - t1) * 1000, 9)}")
```
alexagriffith:
```suggestion
        logging.info(f"decoded request in {round((t2 - t1) * 1000, 9)} ms")
```

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -200,12 +205,49 @@ def model_ready(self, model_name: str) -> bool:
...
+    def encode(self, model_name, body, response, headers) -> Tuple[Dict, Dict[str, str]]:
+        response_headers = {}
+        # if we received a cloudevent, then also return a cloudevent
+        is_cloudevent = False
+        is_binary_cloudevent = False
+        if headers:
+            if has_binary_headers(headers):
+                is_cloudevent = True
+                is_binary_cloudevent = True
+            if "content-type" in headers and "application/cloudevents+json" == headers["content-type"]:
```
alexagriffith:
```suggestion
            if headers.get("content-type", "") == "application/cloudevents+json":
```

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -251,41 +278,34 @@ async def infer(
...
             Dict: Explanation result.
 
         Raises:
             InvalidInput: An error when the body bytes can't be decoded as JSON.
         """
-        try:
-            body = orjson.loads(body)
-        except orjson.JSONDecodeError as e:
-            raise InvalidInput(f"Unrecognized request format: {e}")
+        body = self.decode(body, headers)
```
alexagriffith, line 301: 👍

python/kserve/kserve/protocol/grpc/server.py
alexagriffith: i guess it just is user problem to handle but otherwise we limit them. i think they can override it in their own app anyways if they want

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -0,0 +1,544 @@
...
+                                self._data.append(
+                                    str(obj.item(), encoding='utf-8'))
+                except UnicodeDecodeError:
+                    raise_error(
+                        f'Failed to encode "{obj.item()}" using UTF-8. Please use binary_data=True, if'
+                        ' you want to pass a byte array.')
+            else:
+                self._data = [val.item() for val in input_tensor.flatten()]
+        else:
+            self._data = None
```
alexagriffith, line 177:
```suggestion
            self._data = None
            self._raw_data = b''
```

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -0,0 +1,544 @@
...
+            if self._datatype == "BYTES":
+                serialized_output = serialize_byte_tensor(input_tensor)
+                if serialized_output.size > 0:
+                    self._raw_data = serialized_output.item()
+                else:
+                    self._raw_data = b''
```
alexagriffith, line 183:
```suggestion
```

alexagriffith: eh actually just reread this and idk, you could remove this else like I suggested but its fine either way

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+def get_content(datatype: str, data: InferTensorContents):
+    if datatype == "BOOL":
+        return list(data.bool_contents)
+    elif datatype == "UINT8" or datatype == "UINT16" or datatype == "UINT32":
+        return list(data.uint_contents)
+    elif datatype == "UINT64":
+        return list(data.uint64_contents)
+    elif datatype == "INT8" or datatype == "INT16" or datatype == "INT32":
```
alexagriffith:
```suggestion
    elif datatype in ["INT8", "INT16", "INT32"]:
```

alexagriffith: can we do same thing for line 192 ?

python/kserve/kserve/protocol/rest/server.py
```diff
@@ -0,0 +1,177 @@
...
+        self.cfg = uvicorn.Config(
+            app=app,
+            host="0.0.0.0",
+            port=http_port,
+            log_config={
+                "version": 1,
+                "formatters": {
+                    "default": {
+                        "()": "uvicorn.logging.DefaultFormatter",
+                        "datefmt": "%Y-%m-%d %H:%M:%S",
```
alexagriffith: instead of hard coding this twice save it as a constant? not rlly a big deal if we dont just maybe better consistency if we want to keep them the same

python/kserve/kserve/utils/numpy_codec.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+
+def to_np_dtype(dtype):
```
alexagriffith, line 18: what would make this cleaner is if you had a map of np to dtype and you could loop through the map checking the type and converting it. that would be less repetitive than all the elifs

for example 

datatype_to_np = {"BOOL": bool, "INT8": np.int8}

 def to_np_dtype(dtype):
   for datatype, np in datatype_to_np.items():
     if dtype == datatype:
         return np 
         
         
 and u can use same map for the other way 
 def from_np_dtype(np_dtype):
   for datatype, np in datatype_to_np.items():
     if np_datatype == np:
          return datatype

alexagriffith (commented):
finished review- left some comments

alexagriffith (approved):
overall lgtm just some small comments and suggestions

python/kserve/kserve/model_server.py
yuzisun: the host is not really configurable, by default it binds to localhost which is only accessible to process running on the same node, `0.0.0.0` tells to be reachable from a public IP of the container. Since you don’t know which ip address the container will be assigned within the subnetwork, 0.0.0.0 says to fastapi to accept incoming connections no matter the network adapter or the ip address.

alexagriffith (approved):
/lgtm

PR #2640, opened 2023-01-10
fix light gbm model format

alexagriffith:

This fixes a bug where the lightgbm runtime YAML version is out of date. 

lightgbm version defined as 3.3.2 https://github.com/kserve/kserve/blob/master/python/lgbserver/setup.py#L41, but runtime YAML was not updated and only supports major version 2. 

We should use 3.3.2 as source of truth since this is what we use to build the image. 

**Type of changes**


sukumargaonkar:
looks good to me

PR #2655, opened 2023-01-19
Implement v2/open inference endpoints for OOTB python model servers

### LINKED ISSUE #2756: v2 protocol request "id" not received in predict method payload
/kind bug

The "id" field sent in the input request of v2 protocol is not received inside the predict method payload.

Steps to Reproduce: 

- Start SKlearn server locally - https://github.com/kserve/kserve/tree/master/python/sklearnserver by running `python __main__.py`
- Make infer request - `curl -v -H "Content-Type: application/json" -d @./input.json http://localhost:8080/v2/models/model/infer`
- The payload received at this line - https://github.com/kserve/kserve/blob/6afc78ec7139e6a86a8df49ff3396ab7eeeeaf0d/python/sklearnserver/sklearnserver/model.py#L50 does not have `id`

Sample input json:
```json
{
    "id": "d3b15cad-50a2-4eaf-80ce-8b0a428bd298",
    {
    "inputs": [
      {
        "name": "input-0",
        "shape": [2, 4],
        "datatype": "FP32",
        "data": [
          [6.8, 2.8, 4.8, 1.4],
          [6.0, 3.4, 4.5, 1.6]
        ]
      }
    ]
}
```
Note: I used sklearn example just to test the above scenario.

Suresh-Nakkeran:

Fixes #2756 

**Type of changes**

- [X]  Added unit/e2e tests for sklearn, xgboost, lightgbm, paddle, pmml
- Logs

**Checklist**:

**Release note**:
NONE
```release-note
Implement v2 endpoints for kserve sklearnserver, xgbserver, lightgbmserver, pmmlserver, paddleserver
```


alexagriffith (commented):
can you rebase? seems like a lot of dans changes are in here

python/kserve/test/test_server.py
```diff
@@ -73,7 +77,17 @@ def load(self):
...
+            infer_input = request.to_rest()
+            response_id = generate_uuid()
+            result = infer_input["inputs"][0]["data"]
+            infer_output = InferOutput(name="output-0", shape=list(
+                infer_input["inputs"][0]["shape"]), datatype=infer_input["inputs"][0]["datatype"], data=result)
+            infer_response = InferResponse(model_name=self.name, infer_outputs=[
+                                           infer_output], response_id=response_id)
+            return infer_response.to_rest()
+        else:
+            return {"predictions": request["instances"]}
```
alexagriffith: this code is repeated in the below section (starting with line 107)- can we move this code to a helper method to avoid repeating it?

python/kserve/test/test_server.py
```diff
@@ -227,6 +251,16 @@ def test_predict(self, http_server_client):
...
 
+    def test_infer(self, http_server_client):
+        input_data = b'{"inputs": [{"name": "input-0","shape": [1, 2],"datatype": "FP32","data": [[1,2]]}]}'
+        resp = http_server_client.post('/v2/models/TestModel/infer',
+                                       data=input_data)
+
+        result = json.loads(resp.content)
+        assert resp.status_code == 200
+        assert result["outputs"][0]["data"] == [[1, 2]]
+        assert resp.headers['content-type'] == "application/json"
```
alexagriffith: can we move this logic to helper method also? seems like we repeat this starting at line 322

python/lgbserver/lgbserver/model.py
```diff
@@ -51,14 +55,39 @@ def load(self) -> bool:
...
-            inputs = pd.concat(dfs, axis=0)
+            if isinstance(payload, Dict):
+                dfs = []
+                for input in payload['inputs']:
+                    dfs.append(pd.DataFrame(
+                        input, columns=self._booster.feature_name()))
+                inputs = pd.concat(dfs, axis=0)
+
+                result = self._booster.predict(inputs)
+                return {"predictions": result.tolist()}
```
alexagriffith, line 67: @yuzisun should we keep this like this? can we replace with inferrequest here to avoid hard coding key/dict?

python/paddleserver/paddleserver/model.py
```diff
@@ -58,17 +63,41 @@ def get_model_files(ext: str) -> str:
...
-        except Exception as e:
-            raise Exception("Failed to predict %s" % e) from e
+    def predict(self, payload: Union[Dict, InferRequest], headers: Dict[str, str] = None) -> Union[Dict, InferResponse]:
+
+        if isinstance(payload, Dict) and "instances" in payload:
+            instances = payload["instances"]
+            try:
+                inputs = np.array(instances, dtype='float32')
+            except Exception as e:
+                raise Exception("Failed to initialize NumPy array from inputs:%s, %s"
```
alexagriffith: can we use Exception(f"Failed to initialize NumPy array from inputs:{e},{instances}") ? formatting here is kind of funny

alexagriffith:
> can you rebase? seems like a lot of dans changes are in here

actually I think u just need to compare this branch to dans instead of master since his isn't merged yet

test/e2e/predictor/test_lightgbm.py
```diff
@@ -139,3 +139,40 @@ def test_lightgbm_v2_runtime_kserve():
         3.0601420299625077e-06]
 
     kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
+
+
+@pytest.mark.fast
+def test_lightgbm_v2_kserve():
+    service_name = "isvc-lightgbm-v2-kserve"
+
+    predictor = V1beta1PredictorSpec(
```
alexagriffith, line 149: do u need to set v2 here? how is this different from the method above `test_lightgbm_v2_runtime_kserve` ?

alexagriffith: should `protocol_version="v2",` be in spec?

test/e2e/predictor/test_paddle.py
```diff
@@ -119,3 +119,48 @@ def test_paddle_runtime():
...
+
+@pytest.mark.slow
+def test_paddle_v2_kserve():
+    predictor = V1beta1PredictorSpec(
+        min_replicas=1,
+        model=V1beta1ModelSpec(
+            model_format=V1beta1ModelFormat(
+                name="paddle",
+            ),
+            runtime="kserve-paddleserver",
```
alexagriffith, line 127: I think we need the protocol: v2 in spec?

alexagriffith (changes requested):
for each serving runtime that this PR makes v2 available, please remember to add v2 in the runtime YAMLs

for example, here for lgbserver
https://github.com/kserve/kserve/blob/master/config/runtimes/kserve-lgbserver.yaml#L13

alexagriffith:
> for each serving runtime that this PR makes v2 available, please remember to add v2 in the runtime YAMLs
> 
> for example, here for lgbserver https://github.com/kserve/kserve/blob/master/config/runtimes/kserve-lgbserver.yaml#L13

and just as a note so we don't forget: this will require a website update in docs/modelserving/servingruntimes.md

Suresh-Nakkeran:
> can you rebase? seems like a lot of dans changes are in here

I will rebase the master once Dan's code is merged.

python/lgbserver/lgbserver/model.py
```diff
@@ -51,14 +55,39 @@ def load(self) -> bool:
...
+
+            elif isinstance(payload, InferRequest):
+                dfs = []
+                infer_input = payload.to_rest()
```
yuzisun: we can access the inputs directly here without serializing to rest dict

python/kserve/kserve/constants/constants.py
```diff
@@ -71,3 +71,22 @@
 
 # K8S metadata key constants
 GENERATION = 'generation'
+
+# model input datatypes and numpy datatypes mapping
+DATATYPE_TO_NUMPY = {
```
yuzisun: I think we already have the conversion util defined here?
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/utils/numpy_codec.py#L18

python/kserve/kserve/utils/utils.py
```diff
@@ -133,3 +134,31 @@ def to_headers(context: ServicerContext) -> Dict[str, str]:
...
 
     return headers
+
+
+def get_predict_input(payload: Union[Dict, InferRequest]):
+    if isinstance(payload, Dict):
+        return payload["inputs"] if "inputs" in payload else payload["instances"]
+    elif isinstance(payload, InferRequest):
+        input = payload.inputs[0]
+        if input.datatype == "MIXED":
```
yuzisun: For mixed we need to construct the data frame from multiple inputs.

python/kserve/requirements.txt
```diff
@@ -23,5 +23,6 @@ protobuf>=3.19.0
 prometheus-client>=0.13.1
 orjson>=3.8.0
 httpx>=0.23.0
-fastapi>=0.85.0,<=0.88.0
+fastapi>=0.89.1
+uvicorn>=0.16.0
```
yuzisun: We can remove the uvicorn requirement, it is already specified by fastapi

python/kserve/test/test_server.py
```diff
@@ -73,7 +77,12 @@ def load(self):
         self.ready = True
 
     async def predict(self, request, headers=None):
-        return {"predictions": request["instances"]}
+        if isinstance(request, InferRequest):
+            inputs = get_predict_input(request)
+            infer_response = get_predict_response(request, inputs, self.name)
+            return infer_response.to_rest()
```
yuzisun: `to_rest` should be called in `postprocess` handler automatically if the request includes the `application/json` content type header, so I think we just need to return `InferResponse` here.
https://github.com/kserve/kserve/blob/eb794b5ab35f63564a04df02546c653b4cafd3aa/python/kserve/kserve/model.py#L250

test/e2e/predictor/test_autoscaling.py
```diff
@@ -160,7 +160,7 @@ def test_sklearn_kserve_cpu():
     kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
 
 
-@pytest.mark.fast
+@pytest.mark.slow
```
yuzisun: We can separate raw deployment tests into its own test suite

test/e2e/predictor/test_lightgbm.py
yuzisun: maybe `protocol_version` should be an array as kserve python serving runtime now supports both v1 and v2.

test/e2e/predictor/test_lightgbm.py
```diff
@@ -125,12 +122,115 @@ def test_lightgbm_v2_runtime_kserve():
...
     kserve_client.create(isvc)
     kserve_client.wait_isvc_ready(service_name, namespace=KSERVE_TEST_NAMESPACE)
 
-    res = predict(service_name, "./data/iris_input_v2.json", protocol_version="v2")
-    assert res["outputs"][0]["data"] == [
+    json_file = open("./data/iris_input_v2_grpc.json")
+    payload = json.load(json_file)["inputs"]
+
+    response = predict_grpc(service_name=service_name, payload=payload, model_name=model_name)
+    fields = response.outputs[0].contents.ListFields()
```
yuzisun: I think you should be able to directly access the float content field instead of using general `ListFields` functions

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -138,108 +138,109 @@ def test_mms_sklearn_kserve(protocol_version: str, storage_uri: str):
...
-        assert responses[1]["predictions"] == [1, 1]
-    elif protocol_version == "v2":
-        assert responses[0]["outputs"][0]["data"] == [1.0, 1.0]
-        assert responses[1]["outputs"][0]["data"] == [1.0, 1.0]
-
-    # Clean up inference service and trained models
-    for model_name in model_names:
-        kserve_client.delete_trained_model(model_name, KSERVE_TEST_NAMESPACE)
-    kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
+# NOTE: Temporarily commented due to fail run while PR action deployment
```
yuzisun: we can add back now?

test/e2e/predictor/test_autoscaling.py
Suresh-Nakkeran: Yes @yuzisun .. I added new e2e test group called raw and added all raw deployment tests.

python/kserve/test/test_server.py
Suresh-Nakkeran: updated!

python/kserve/requirements.txt
Suresh-Nakkeran: removed!

python/kserve/kserve/constants/constants.py
Suresh-Nakkeran: Yeah. removed!

python/kserve/kserve/utils/utils.py
yuzisun: @Suresh-Nakkeran we should get the all the inputs and return pandas framework if data types are different

yuzisun: Let's remove the check on line 144 as it is in a separate PR

python/kserve/kserve/utils/utils.py
```diff
@@ -133,3 +134,31 @@ def to_headers(context: ServicerContext) -> Dict[str, str]:
         headers[metadatum.key] = metadatum.value
 
     return headers
+
+
+def get_predict_input(payload: Union[Dict, InferRequest]):
```
yuzisun, line 139: Add return type which is`Union[Dict, np.ndarray]`

python/pmmlserver/pmmlserver/model.py
```diff
@@ -57,10 +59,26 @@ def load(self) -> bool:
...
+                result = [self.evaluator.evaluate(
+                    dict(zip(self.input_fields, instance))) for instance in instances]
+                return {"predictions": result}
+            elif isinstance(payload, InferRequest):
+                result = [self.evaluator.evaluate(
+                    dict(zip(self.input_fields, instance))) for instance in payload.inputs[0].data]
+                infer_output = InferOutput(
+                    name="output-0",
+                    shape=[len(result)],
+                    datatype="MIXED",
```
yuzisun: datatype should be converted from result

python/kserve/kserve/utils/utils.py
Suresh-Nakkeran: @yuzisun return type is `Union[List, np.ndarray]` correct?

python/pmmlserver/pmmlserver/model.py
Suresh-Nakkeran: @yuzisun How can we get the type from the result? I hope the result will be a list.

python/kserve/kserve/utils/utils.py
yuzisun: you are right

python/pmmlserver/pmmlserver/model.py
yuzisun: here should be the data type of the values in the list, does `evaluate` return the numpy array?

python/paddleserver/paddleserver/model.py
```diff
@@ -59,17 +62,17 @@ def get_model_files(ext: str) -> str:
         self.ready = True
         return self.ready
 
-    def predict(self, payload: Dict, headers: Dict[str, str] = None) -> Dict:
-        instances = payload["instances"]
+    def predict(self, payload: Union[Dict, InferRequest], headers: Dict[str, str] = None) -> Union[Dict, InferResponse]:
         try:
-            inputs = np.array(instances, dtype='float32')
+            instance = get_predict_input(payload)
+            np_array_input = np.array(instance, dtype='float32')
```
yuzisun, line 69: for v2 protocol the instance is already np array which is the return type from `get_predict_input`

python/pmmlserver/pmmlserver/model.py
yuzisun: added a commit to convert the response to pandas and then multiple infer outputs

test/e2e/predictor/test_lightgbm.py
yuzisun: Created https://github.com/kserve/kserve/issues/2784 to track this issue, if we specify v2 then kserve currently selects the mlserver runtime

test/e2e/predictor/test_paddle.py
yuzisun: Created https://github.com/kserve/kserve/issues/2784 to track this issue, if we specify v2 then kserve currently selects the mlserver runtime

yuzisun:
> > for each serving runtime that this PR makes v2 available, please remember to add v2 in the runtime YAMLs
> > for example, here for lgbserver https://github.com/kserve/kserve/blob/master/config/runtimes/kserve-lgbserver.yaml#L13
> 
> and just as a note so we don't forget: this will require a website update in docs/modelserving/servingruntimes.md

Created https://github.com/kserve/website/issues/241

python/lgbserver/lgbserver/model.py
yuzisun: will address in https://github.com/kserve/kserve/pull/2748

PR #2670, opened 2023-02-02
Add mirroring graph node type

### LINKED ISSUE #2240: Add a new inference routing type: Mirroring
/kind feature
We are facing a new use case: we want to add a testing branch to an online service(main branch), this testing branch will get some of online queries, but we don't want the testing branch response. So far, inference graph has no type of routing node can slove this, so we want to add an new routing type: Mirroring. A mirroring routing node need 2 fields: `serviceName` and `mirroring percent`.

```
                           +--------------+
      <---------Res--------+              |
Client          100%       |              |
      ----+-----Req-------->     Main     |
          |                |              |
          |                |              |
          |                +--------------+
          |
          |
          |                +--------------+
          |                |              |
          |  20%           |              |
           +-----Req-------->   Testing    |
                           |              |
          <-------X--------+              |
                           +--------------+
```

Iamlovingit:
Signed-off-by: iamlovingit <freecode666@gmail.com>

We are facing a new use case: we want to add a testing branch to an online service(main branch), this testing branch will get some of online queries, but we don't want the testing branch response. So far, inference graph has no type of routing node can slove this, so we want to add an new routing type: Mirroring. A mirroring routing node need 2 fields: serviceName and mirroring percent.

Fixes #2240 

**Type of changes**

## Design
```
                           +--------------+
      <---------Res--------+              |
Client          100%       |              |
      ----+-----Req-------->     Main     |
          |                |              |
          |                |              |
          |                +--------------+
          |
          |
          |                +--------------+
          |                |              |
          |  20%           |              |
           +-----Req-------->   Testing   |
                           |              |
          <-------X--------+              |
                           +--------------+
```

### Example yaml
```yaml
kind: InferenceGraph
metadata:
  name: model-mirroring
spec:
  nodes:
    root:
      routerType: Mirroring
      steps:
      - serviceUrl: http://single-1.default.${domain}/switch # Main route
      - serviceUrl: http://single-2.default.${domain}/switch # Mirroring route
        weight: 100
      - serviceUrl: http://single-3.default.${domain}/switch # Mirroring route
        weight: 20
```

#### Default weight
```yaml
kind: InferenceGraph
metadata:
  name: model-mirroring
spec:
  nodes:
    root:
      routerType: Mirroring
      steps:
      - serviceUrl: http://single-1.default.${domain}/switch # Main route
      - serviceUrl: http://single-2.default.${domain}/switch # Mirroring route, default weight 100
      - serviceUrl: http://single-3.default.${domain}/switch # Mirroring route
        weight: 20
```
#### Wrong node type
```yaml
kind: InferenceGraph
metadata:
  name: model-mirroring
spec:
  nodes:
    root:
      routerType: WrongType
      steps:
      - serviceUrl: http://single-1.default.${domain}/switch # Main route
      - serviceUrl: http://single-2.default.${domain}/switch # Mirroring route, default weight 100
      - serviceUrl: http://single-3.default.${domain}/switch # Mirroring route
        weight: 20
```
#### Wrong mirroring weight
```yaml
kind: InferenceGraph
metadata:
  name: model-mirroring
spec:
  nodes:
    root:
      routerType: Mirroring
      steps:
      - serviceUrl: http://single-1.default.${domain}/switch # Main route
      - serviceUrl: http://single-2.default.${domain}/switch # Mirroring route
         weight: 101
      - serviceUrl: http://single-3.default.${domain}/switch # Mirroring route
        weight: -20
```

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -20,13 +20,14 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
 	"io"
 	"net/http"
 	"os"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/kserve/kserve/pkg/constants"
 
@@ -53,6 +54,9 @@ func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, 
 			}
 		}
 	}
+
+	//TODO in "Binary Data Type" case, graph will make something error.
+	//TODO Content-Type get value from request's header or response's header.
 	req.Header.Add("Content-Type", "application/json")
 	resp, err := http.DefaultClient.Do(req)
 
@@ -71,7 +75,7 @@ func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, 
 func pickupRoute(routes []v1alpha1.InferenceStep) *v1alpha1.InferenceStep {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	//generate num [0,100)
-	point := r.Intn(99)
+	point := r.Intn(100)
 	end := 0
 	for _, route := range routes {
 		end += int(*route.Weight)
@@ -142,6 +146,37 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
 			return nil, 404, err
 		}
 		return handleSplitterORSwitchNode(route, graph, input, headers)
+	}
+	if currentNode.RouterType == v1alpha1.Mirroring {
+		//* check the step len, if step is nil, return input directly
+		if currentNode.Steps == nil {
+			return input, 200, nil
+		}
+
+		if len(currentNode.Steps) > 1 { //* exist mirroring type services.
+
+			//* create goroutines for mirror service, and ignore response
+			for i := range currentNode.Steps[1:] { //* ignore the main service (Steps[0])
+				step := currentNode.Steps[i]
+				//* get mirroring service weight.
+				var weight int64
+				if step.Weight != nil {
+					weight = *step.Weight
+				} else {
+					weight = 100
+				}
+				//* generate rand num [0,100)
+				r := rand.New(rand.NewSource(time.Now().UnixNano()))
+				point := r.Intn(100)
+
+				if point < int(weight) {
+					go callService(step.ServiceURL, input, headers)
+				}
+			}
+		}
+		//* As definition mentioned, the first one of step's is main service, others are mirror service.
+		//* handle main service
+		return executeStep(&currentNode.Steps[0], graph, input, headers)
 	}
 	if currentNode.RouterType == v1alpha1.Ensemble {
 		ensembleRes := make([]chan EnsembleStepOutput, len(currentNode.Steps))
--- a/pkg/apis/serving/v1alpha1/inference_graph_validation.go
+++ b/pkg/apis/serving/v1alpha1/inference_graph_validation.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
 	"fmt"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -37,6 +38,10 @@ const (
 	WeightNotProvidedError = "InferenceGraph[%s] Node[%s] Route[%s] missing the 'Weight'"
 	// InvalidWeightError defines the error message for sum of traffic weight is not 100
 	InvalidWeightError = "InferenceGraph[%s] Node[%s] splitter node: the sum of traffic weights for all routing targets should be 100"
+	// InvalidMirroringWeightError defines the error message for sum of traffic weight is not 100
+	InvalidMirroringWeightError = "InferenceGraph[%s] Node[%s] mirroring node: the weight shoule be in (1,100]"
+	// InvalidMirroringSpecError defines the error message for sum of traffic weight is not 100
+	InvalidMirroringSpecError = "InferenceGraph[%s] Node[%s] mirroring node: needs at least 2 steps"
 	// DuplicateStepNameError defines the error message for more than one step contains same name
 	DuplicateStepNameError = "Node \"%s\" of InferenceGraph \"%s\" contains more than one step with name \"%s\""
 	// TargetNotProvidedError defines the error message for inference graph target not specified
@@ -84,6 +89,15 @@ func (ig *InferenceGraph) ValidateCreate() (admission.Warnings, error) {
 	if err := validateInferenceGraphSplitterWeight(ig); err != nil {
 		return nil, err
 	}
+
+	if err := validateInferenceGraphMirroringSpec(ig); err != nil {
+		return nil, err
+	}
+
+	if err := validateInferenceGraphMirroringWeight(ig); err != nil {
+		return nil, err
+	}
+
 	return nil, nil
 }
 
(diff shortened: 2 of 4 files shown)


rachitchauhan43:
@Iamlovingit : Is it possible to have same traffic flowing through both nodes ? We have a use case where we want to make sure that all requests are routed to both the nodes. Basically 100% of traffic both the nodes. We can simulate it using ensemble but then we would have to makes sure that testing node response is not included in the ensemble o/p.

So, question is can I set weight to 100% for multiple steps in the node ?

cmd/router/main.go
```diff
@@ -109,6 +110,30 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
+		//* create goroutines for mirror service, and ignore response
+		for i := range currentNode.Steps {
+			step := currentNode.Steps[i]
+			var weight int64
+			if step.Weight != nil {
+				weight = *step.Weight
+			} else {
+				weight = 100
+			}
+			r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
rachitchauhan43: Can you put some inline comments on how it is working ?

Iamlovingit:
> 



> @Iamlovingit : Is it possible to have same traffic flowing through both nodes ? We have a use case where we want to make sure that all requests are routed to both the nodes. Basically 100% of traffic both the nodes. We can simulate it using ensemble but then we would have to makes sure that testing node response is not included in the ensemble o/p.
> 
> So, question is can I set weight to 100% for multiple steps in the node ?

@rachitchauhan43  This is supported, user can set 100% to every mirror step in the node. Example:
```yaml
kind: InferenceGraph
metadata:
  name: model-mirroring
spec:
  nodes:
    root:
      routerType: Mirroring
      steps:
      - serviceUrl: http://single-1.default.${domain}/switch # Main route
      - serviceUrl: http://single-2.default.${domain}/switch # 100% mirroring route
        weight: 100
      - serviceUrl: http://single-3.default.${domain}/switch # Anohter 100% mirroring route
        weight: 100
```

rachitchauhan43:
> > 
> 
> > @Iamlovingit : Is it possible to have same traffic flowing through both nodes ? We have a use case where we want to make sure that all requests are routed to both the nodes. Basically 100% of traffic both the nodes. We can simulate it using ensemble but then we would have to makes sure that testing node response is not included in the ensemble o/p.
> > So, question is can I set weight to 100% for multiple steps in the node ?
> 
> @rachitchauhan43 This is supported, user can set 100% to every mirror step in the node. Example:
> 
> ```yaml
> apiVersion: "serving.kserve.io/v1alpha1"
> kind: InferenceGraph
> metadata:
>   name: model-mirroring
> spec:
>   nodes:
>     root:
>       routerType: Mirroring
>       steps:
>       - serviceUrl: http://single-1.default.${domain}/switch # Main route
>       - serviceUrl: http://single-2.default.${domain}/switch # 100% mirroring route
>         weight: 100
>       - serviceUrl: http://single-3.default.${domain}/switch # Anohter 100% mirroring route
>         weight: 100
> ```

This will be awesome as we are looking for a way to carry out Shadow mode testing where we want to stream 100% traffic to a newer version but don't want responses from that newer version to go back to customers.

rachitchauhan43:
@Iamlovingit : Can you fix the DCO check ? Also, are you planning to have to get it out for kserve-0.11 ?

cmd/router/main.go
Iamlovingit: update

Iamlovingit:
@rachitchauhan43 @yuzisun please help review this PR

rachitchauhan43:
Is it possible to add e2e tests in here for this https://github.com/kserve/kserve/tree/master/test/e2e/graph to cover for basic scenarios like such graph is able to come up in every regression?

cmd/router/main.go
```diff
@@ -109,6 +113,37 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
 		}
 		return executeStep(route, graph, input, headers)
 	}
+	if currentNode.RouterType == v1alpha1.Mirroring {
+		//* check the step len, if step is nil, return input directly
+		if currentNode.Steps == nil {
+			return input, nil
+		}
```
yuzisun: I think there should be some validation on the inference graph yaml that the steps should be more than two for mirroring node type

Iamlovingit:
> Is it possible to add e2e tests in here for this https://github.com/kserve/kserve/tree/master/test/e2e/graph to cover for basic scenarios like such graph is able to come up in every regression?

Good idea. can you suggest some senarios? I am not sure choose which model?

cmd/router/main.go
Iamlovingit: add a validation for mirroring spec

rachitchauhan43:
> > Is it possible to add e2e tests in here for this https://github.com/kserve/kserve/tree/master/test/e2e/graph to cover for basic scenarios like such graph is able to come up in every regression?
> 
> Good idea. can you suggest some senarios? I am not sure choose which model?

@Iamlovingit : I guess testing that it actually made calls to the other node from where response is not expected will be tricky.  I don't have an idea at the moment. 

Go ahead and please rebase it.

pkg/apis/serving/v1alpha1/inference_graph_validation.go
```diff
@@ -37,6 +38,10 @@ const (
 	WeightNotProvidedError = "InferenceGraph[%s] Node[%s] Route[%s] missing the 'Weight'"
 	// InvalidWeightError defines the error message for sum of traffic weight is not 100
 	InvalidWeightError = "InferenceGraph[%s] Node[%s] splitter node: the sum of traffic weights for all routing targets should be 100"
+	// InvalidMirroringWeightError defines the error message for sum of traffic weight is not 100
```
rachitchauhan43, line 41:
```
// InvalidMirroringWeightError defines the error message for sum of traffic weight is not 100
```

This is no longer true , right ? Every step can be  100% as we discussed in comments.

cmd/router/main.go
```diff
@@ -53,6 +54,9 @@ func callService(serviceUrl string, input []byte, headers http.Header) ([]byte,
 			}
 		}
 	}
+
+	//TODO in "Binary Data Type" case, graph will make something error.
+	//TODO Content-Type get value from request's header or response's header.
```
rachitchauhan43, line 59: 👍

cmd/router/main.go
```diff
@@ -143,6 +147,37 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
+					weight = *step.Weight
+				} else {
+					weight = 100
+				}
+				//* generate rand num [0,100)
+				r := rand.New(rand.NewSource(time.Now().UnixNano()))
+				point := r.Intn(100)
+
+				if point < int(weight) {
+					go callService(step.ServiceURL, input, headers)
```
rachitchauhan43, line 173: @Iamlovingit : Since we are making fire and forget call, should we make callService accept a timeout setting param so that go routine doesn't hang waiting for response from upstream? See [here](https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779) . If we do not set any timeout, then it will wait infinitely OR will rely on upstream's timeout.

With this code , it can turn out to be a **network leak** and will lead to **connection hangs** ultimately.
Possible solutions:
1. Either, we can expose another field in spec for mirroring node for specifying timeout that can be passed to callService method and then http client can be initiated with that timeout value, OR
2. We can see how to do proper fire-and-forget calls using golang routines

Iamlovingit:
@rachitchauhan43 @yuzisun please review this PR when you guys have time. :>

rachitchauhan43 (changes requested):

namirinz:
Any progress? @Iamlovingit  @rachitchauhan43 @yuzisun

PR #2679, opened 2023-02-06
Allowing setting minReplicas for IG router.

### LINKED ISSUE #2456: Allow setting min and max replicas for Inference Graph router.
/kind feature

Why can’t I set min replicas for the IG’s router ? 
Today, this is not possible because of this https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go#L128

Whereas in case of ISVC we let minReplicas override ksvc’s min-scale as implemented [here](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go#L74-L78).

rachitchauhan43:
Also added some logging for IG router to log step execution time.

Signed-off-by: rachitchauhan43 <rachitchauhan43@gmail.com>

- Fixes #2456 -- Now IG router respect minReplicas annotation
- Also, adds some [more logging](https://github.com/rachitchauhan43/kserve/blob/ig_configurable/cmd/router/main.go#L44) to IG router to log step execution time and not just node level execution time
- Also, making sure that slice [headersToPropagate](https://github.com/rachitchauhan43/kserve/blob/ig_configurable/cmd/router/main.go#L206-L209) stays of length 0 if it is not initialized. Otherwise, with the current implementation it was getting initialized to a slice of 1 with empty string and un-necessarily getting iterating [here](https://github.com/kserve/kserve/blob/master/cmd/router/main.go#L45) and wasting cpu cycles.
- Initializing logger in main_test.go in [init() method](https://github.com/rachitchauhan43/kserve/blob/ig_configurable/cmd/router/main_test.go#L17-L19) to make sure we can see logs when tests run

Fixes #2456 

**Type of changes**

Created an IG with following YAML:
```
kind: InferenceGraph
metadata:
  annotations:
    admiral.io/ignore: "true"
    autoscaling.knative.dev/min-scale: "2"
  name: usual-flow
  namespace: kserve-test
spec:
  nodes:
    root:
      routerType: Sequence
      steps:
      - data: $response
        name: root-node-model-sim-step
        serviceUrl: http://model-sim-predictor-default.kserve-test.svc.cluster.local/v1/models/model-sim:predict
```
And now, i am able to see 2 replicas of IG router coming up as `autoscaling.knative.dev/min-scale: "2"` is being respected now.

- Logs

**Checklist**:

**Release note**:

```release-note

```


cmd/router/main.go
```diff
@@ -41,6 +41,8 @@ import (
 var log = logf.Log.WithName("InferenceGraphRouter")
 
 func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, error) {
+	defer timeTrack(time.Now(), "step", serviceUrl)
+	log.Info("Entering callService", "url", serviceUrl)
```
yuzisun, line 45: Do we need this at INFO level? the message does not convey a lot info, we can log more useful information like request id header etc.

cmd/router/main.go
```diff
@@ -90,13 +92,13 @@ func pickupRouteByCondition(input []byte, routes []v1alpha1.InferenceStep) *v1al
 	return nil
 }
 
-func timeTrack(start time.Time, name string) {
+func timeTrack(start time.Time, nodeOrStep string, name string) {
```
yuzisun, line 95: Where did we use this for timing the step?

yuzisun:
> /assign @Iamlovingit @yuzisun

lgtm mostly, a few small comments

cmd/router/main.go
rachitchauhan43: [Here](https://github.com/kserve/kserve/pull/2679/files#diff-d8ccd6fde43fcb90d60f6319f818662f9e4340295fc9bbde755fd93d98da33c1R44). So, `callService` method is essentially getting called for every step. That is how it is printing.
Screenshot 2023-02-13 at 9 27 07 AM

cmd/router/main.go
rachitchauhan43: What I have seen so far is that it's helpful for tracing a lot. I don't see any downside of having it.

yuzisun:
Thanks @rachitchauhan43 !
/lgtm
/approve

PR #2710, opened 2023-02-21
Allow IG spec to specify(override) headers to propagate

### LINKED ISSUE #2693: Make "propagate headers" configurable per Inference Graph instead of global
/kind feature

Today, the headers that we can pass from IG router to it's nodes are configurable only at global level.
This was done as part of this https://github.com/kserve/kserve/issues/2289

We want to make sure that this feature is available at per IG level too instead of only being available at global level so that IG owner can decide which all headers to propagate.

navneetraju:

Allows the IG spec to specify(override) the headers to be propagated to the inference services, allowing IG owners to decide instead of only being globally available

Fixes #2693

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```


rachitchauhan43:
@navneetraju : Can you sign off the commit? You can see instructions here https://github.com/kserve/kserve/pull/2710/checks?check_run_id=11482085190

rachitchauhan43:
@navneetraju : Also, since you have changed the spec, this must have changes for generated code and CRDs too ? Can you check-in those as well?

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -175,12 +175,24 @@ func createKnativeService(componentMeta metav1.ObjectMeta, graph *v1alpha1api.In
...
 
 	// Only adding this env variable "PROPAGATE_HEADERS" if router's headers config has the key "propagate"
-	value, exists := config.Headers["propagate"]
-	if exists {
+	configHeaders, configHeadersExists := config.Headers["propagate"]
+	specHeaders, specHeadersExists := graph.Spec.Headers["propagate"]
+
+	// Use spec headers over router config headers
+	if specHeadersExists {
+		log.Info("Found spec header")
```
rachitchauhan43: Can we log the headers here ? 
Like log.Info("Found spec header", "specHeaders", specHeaders)

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -175,12 +175,24 @@ func createKnativeService(componentMeta metav1.ObjectMeta, graph *v1alpha1api.In
...
+		service.Spec.ConfigurationSpec.Template.Spec.PodSpec.Containers[0].Env = []v1.EnvVar{
+			{
+				Name:  constants.RouterHeadersPropagateEnvVar,
+				Value: strings.Join(specHeaders, ","),
+			},
+		}
+	} else if configHeadersExists {
+		log.Info("Found config header")
```
rachitchauhan43: Here as well.

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
rachitchauhan43:
```suggestion
		log.Info("Found headers in the spec")
```

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
rachitchauhan43:
```suggestion
		log.Info("Found headers in global router config")
```

rachitchauhan43 (changes requested):

navneetraju:
IG yaml tested with:
```
---
kind: InferenceService
metadata:
  name: postprocessor
spec:
  predictor:
    minReplicas: 1
    containers:
      - name: kserve-container
        image: docker.****.com/****/ig-**-post-processing-node/service/ig-**-post-processing-node:0.2.3
        imagePullPolicy: Always
        resources:
          limits:
            cpu: '2'
            memory: '1Gi'
          requests:
            cpu: '500m'
            memory: '512Mi'
        args:
          - --model_name
          - postprocessor
---
kind: InferenceGraph
metadata:
  name: inferencegraph***
spec:
  headers:
    propagate:
      - Offeringid
      - Appid
  nodes:
    root:
      routerType: Sequence
      steps:
        - serviceUrl: http://postprocessor.default.svc.cluster.local/v1/models/postprocessor:predict
```

Logs in isvc:
Screenshot 2023-02-22 at 9 48 35 PM

rachitchauhan43 (approved):

rachitchauhan43:
@navneetraju : Can you also add screenshots from kserve controller logs on how it logs lines from reconciler ?

navneetraju:
Here is a sample controller log:

Screenshot 2023-02-23 at 9 43 29 AM

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -42,6 +42,19 @@ type InferenceGraph struct {
...
+		 "propagate": [
+			"Custom-Header1",
+			"Custom-Header2"
+		  ]
+		}
+		Note: Making Headers, a map of strings, gives the flexibility to extend it in the future to support adding more
+		operations on headers. For example: Similar to "propagate" operation, one can add "transform" operation if they
+		want to transform headers keys or values before passing down to nodes.
+	*/
+	Headers map[string][]string `json:"headers,omitempty"`
```
yuzisun, line 58: Not sure if this field is at the right level, do we envision that there are other configurations which need to be overridden on the `InferenceGraph` spec?

yuzisun:
Looks we do not have an integration test for inference graph reconciliation like we do have for inference service, as we are making more changes we should have the tests in place.

pkg/apis/serving/v1alpha1/inference_graph.go
navneetraju: @yuzisun Yes, there is another PR for adding the affinity spec as well resource spec. https://github.com/kserve/kserve/pull/2711

navneetraju:
Seems like there is an [e2e test ](https://github.com/kserve/kserve/blob/master/test/e2e/graph/test_inference_graph.py) for inference graph, this should be testing the reconciliation. This test can be enhanced if needed

yuzisun:
> Seems like there is an [e2e test ](https://github.com/kserve/kserve/blob/master/test/e2e/graph/test_inference_graph.py) for inference graph, this should be testing the reconciliation. This test can be enhanced if needed

right, but would be nice to run `make test` to test the reconciliation locally without needing the whole setup to run the e2e test.

pkg/apis/serving/v1alpha1/inference_graph.go
rachitchauhan43: So, we have enabled headers as an optional filed in the spec  in #2396 , but it was only controlled globally. This PR is to let it be overridden if specified at spec level.

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -42,6 +42,19 @@ type InferenceGraph struct {
...
+		headers: {
+		 "propagate": [
+			"Custom-Header1",
+			"Custom-Header2"
+		  ]
+		}
+		Note: Making Headers, a map of strings, gives the flexibility to extend it in the future to support adding more
+		operations on headers. For example: Similar to "propagate" operation, one can add "transform" operation if they
+		want to transform headers keys or values before passing down to nodes.
+	*/
```
rachitchauhan43, line 56:
```suggestion
	*/
	// +optional
```

pkg/apis/serving/v1alpha1/inference_graph.go
yuzisun: Affinity and resources are standard pod and container fields which exist on inference service as well, I am a bit concerned about that this can open the door for introducing a set of random configuration fields on inference graph spec.

yuzisun: For configuration overrides, we generally use annotations such as `DeploymentMode` in kserve and knative autoscaling annotations.

You can see the knative examples here with global config and annotation overrides
https://knative.dev/docs/serving/autoscaling/scale-to-zero/#scale-to-zero-last-pod-retention-period

rachitchauhan43: @navneetraju : Do you want try getting these overrides from annotations as Dan suggested ? Since these are not standard fields for a spec, so does make sense that it will encourage others in future to add more non-standard fields to the spec.

rachitchauhan43:
> Seems like there is an [e2e test ](https://github.com/kserve/kserve/blob/master/test/e2e/graph/test_inference_graph.py) for inference graph, this should be testing the reconciliation. This test can be enhanced if needed

@navneetraju : These are the tests that you need to extend to test reconciliation e2e. 
https://github.com/rachitchauhan43/kserve/blob/master/pkg/controller/v1alpha1/inferencegraph/controller_test.go

But you can also extend the other test suite as well over [here](https://github.com/rachitchauhan43/kserve/tree/master/test/e2e/graph)

pkg/apis/serving/v1alpha1/inference_graph.go
navneetraju: Sure, i think it makes sense. My only concern is that we are treating the headers as an object with sub-objects such as propagate, since we want to extensible where we may want to add a block headers config for example. Do we want to represent such objects as an annotation or do we restrict this to propagate only, where give an annotation such as:
```
kind: InferenceGraph
metadata:
  name: inferencegraph***
spec:
  metadata:
    annotations:
      propagate-headers: "header_a,header_b"
  nodes:
    root:
      routerType: Sequence
      steps:
        - serviceUrl: http://postprocessor.default.svc.cluster.local/v1/models/postprocessor:predict
```

rachitchauhan43: @navneetraju : I guess we can do an annotation like this:

`serving.kserve.io/inferencegraph-headers: {"propagate":["Custom-Header1","Custom-Header2"]}`

This annotation will accommodate all rules for propagation/blocking.

@sivanantha321 / @yuzisun : Your thoughts ? Are you guys OK with this approach of overriding this propagate headers option per IG ?

rachitchauhan43 (changes requested):

pkg/apis/serving/v1alpha1/inference_graph.go
sivanantha321: @rachitchauhan43 What if we introduce new features as you suggested [here](https://github.com/kserve/kserve/blob/430da16356129f0c0f5864d8b29e002a84ca5d55/pkg/controller/v1alpha1/inferencegraph/controller.go#L83). Would those be also accommodated by this annotation or should we introduce annotation just for propagating headers ?

rachitchauhan43: @sivanantha321 :  I think let's just keep one annotation that tells what to do with headers. So, if we add a functinality say `drop` certain headers, then it would look like this:
```
serving.kserve.io/inferencegraph-headers: {"propagate":["Custom-Header1","Custom-Header2"],"drop":["Custom-Header3"]}
```

rachitchauhan43: @navneetraju : Are you still interested i reviving this PR ? If not , please let us know. We will take over then.

navneetraju:
Have been slightly out of sync with this. If we are good with this format: ```serving.kserve.io/inferencegraph-headers: {"propagate":["Custom-Header1","Custom-Header2"]}```, I can proceed with the changes.

yuzisun:
> Have been slightly out of sync with this. If we are good with this format: `serving.kserve.io/inferencegraph-headers: {"propagate":["Custom-Header1","Custom-Header2"]}`, I can proceed with the changes.

@navneetraju sgtm

seeking-dev:
> Have been slightly out of sync with this. If we are good with this format: `serving.kserve.io/inferencegraph-headers: {"propagate":["Custom-Header1","Custom-Header2"]}`, I can proceed with the changes.

Hi @navneetraju, Do we have any update on it?

seeking-dev:
Looking forward to it

PR #2726, opened 2023-03-02
Fix knative installation failure in e2e test

### LINKED ISSUE #2730: Knative installation keeps failing in e2e tests
/kind bug

The e2e tests are failing every now and then while running the knative installation step, more specifically while patching the configmap. A solution has to be provided so that the installation completes successfully using some kind of retry mechanism.

All e2e tests to run without any issues.

**Environment:**

e2e environment

andyi2it:
Add a sleep.
Add a simple retry mechanism in case of failure.

* Fixes issue with knative installation/patch during e2e*

Fixes #2730 #2731

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -324,6 +324,9 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.17.3'
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
       - uses: ./.github/actions/minikube-setup
       - uses: ./.github/actions/base-download
       - name: Build queue proxy extension image
--- a/test/scripts/gh-actions/setup-deps.sh
+++ b/test/scripts/gh-actions/setup-deps.sh
@@ -21,7 +21,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )";
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]:-$0}")" &>/dev/null && pwd 2>/dev/null)"
 
 ISTIO_VERSION="1.12.0"
 KNATIVE_VERSION="knative-v1.4.0"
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -324,6 +324,9 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.17.3'
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
rachitchauhan43, line 329: @andyi2it : Curious to know how did you figure out from the [logs](https://github.com/kserve/kserve/actions/runs/4315581912/jobs/7531712384) that it was python version issue.

Would you mind sharing how you figured this out?

andyi2it: @rachitchauhan43 
Its not from the e2e logs alone.
I ran it locally, and compared the differences. My local had an earlier version of python and it was working.
So, I tried pinning the version. 
However, the most accurate detail is that, the portforward library was the one that was breaking the e2e.
The newer versions of python uses the newer portforward library by default and hence the error.

Given that the root cause is identified, we can remove pinning python to 3.9. 
However, I think we should pin python to the max version we support. Will create a separate issue for that.

rachitchauhan43: @andyi2it : This is really awesome work. Thanks for this fix 👍

test/scripts/gh-actions/setup-deps.sh
```diff
@@ -71,11 +71,16 @@ for i in 1 2 3 ; do kubectl apply -k test/overlays/knative && break || sleep 15;
...
-echo "Add knative hpa..."
+# echo "Add knative hpa..."
 # kubectl apply -f https://github.com/knative/serving/releases/download/knative-v1.0.0/serving-hpa.yaml
 
 # Skip tag resolution for certain domains
-kubectl patch cm config-deployment --patch '{"data":{"registries-skipping-tag-resolving":"nvcr.io,index.docker.io"}}' -n knative-serving
+# sleep to avoid knative webhook timeout error
+sleep 5
+# Retry if configmap patch fails
+for i in 1 2 3; do
```
rachitchauhan43, line 81: @andyi2it : This is to address which test failure ?

rachitchauhan43 (approved):

test/scripts/gh-actions/setup-deps.sh
andyi2it: This one is for #2730.
This PR fixes both #2730 and #2731

PR #2762, opened 2023-03-20
fix: model_server log level

shahar3012:
**What this PR does / why we need it**:
Fixed a bug to support KSERVE_LOGLEVEL to allow logging both DEBUG and INFO lines

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: 
Fixes #2761

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [X] Specify the env `KSERVE_LOGLEVEL` to `DEBUG` which then prints the debugging logs
```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: inference-service
spec:
  predictor:
    containers:
      - image: inferenceservice-image-tag
        env:
          - name: KSERVE_LOGLEVEL
            value: 'DEBUG'
```
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
Introduce KSERVE_LOGLEVEL env variable to allow configuring the logging level
```

PR #2782, opened 2023-03-31
python: allow to override the UvicornServer's default log config

### LINKED ISSUE #2778: Customize the Uvicorn's access log format on Kserve 0.10+
/kind feature

Hi folks! As far as I can see the Uvicorn's logging config in 0.10+ for FastAPI is hardcoded in the `UvicornServer`, and I didn't find a good way to change the access log format to something custom. For example, I'd like to use a json/ECS format to be able to ingest my model server's access logs, index them and query the results via Kibana (for example, to inspect traffic patterns etc..).

Maybe we could allow to pass a custom configuration from the `ModelServer`'s class parameters?

None

elukey:
It is useful to be able to change Uvicorn's log settings, for example to be able to change the access log's format.

The Uvicorn's Config class provides a "configure_logging" function that is smart about the log_config parameter - if it is a string, and not a dictionary, it is interpreted as a file containing logging directives and that needs to be loaded from disk. This is very convenient for the KServe use case: one can either use the default config, hardcoded in a dictionary, or anything provided as yaml or json file.

This change fixes also the default UvicornServer's config.

Sadly the current Uvicorn implementation doesn't support
changing the access log format, but they suggest to use
asgi-logger:

https://github.com/encode/uvicorn/pull/947#issuecomment-1283006367

It seems the only way forward to allow a decent customization
of a vital sign like the access log.

The main issue is that `asgi-logger` requires Python 3.8, so the dependency is added
as extra to Poetry's config.

Fixes: #2778

**Type of changes**

I tried to use the following snippet:

```
import kserve
from typing import Dict
import argparse
import logging

logger = logging.getLogger()

class CustomPredictor(kserve.Model):
    def __init__(self, name: str):
        super().__init__(name)
        self.name = name
        self.load()

    def load(self):
        self.ready = True

    def predict(self, request: Dict, headers: Dict) -> Dict:
        return {"predictions": True}

if __name__ == "__main__":
    model = CustomPredictor("test")
    kserve.ModelServer(workers=1, enable_latency_logging=False, log_config_file="/tmp/test1.yaml").start([model])
```
I saved in `/tmp/test1.yaml` a logging config, and ran the above script. I was able to observe access logs following the configuration that I passed in the file.

Moreover, if I run the `ModelServer` class without `log_config_file`, I see access logs as expected (as opposed to now that I don't see them, IIUC due to the wrong default config hardcoded in the `UvicornServer`'s class).

I also tried the new `access_log_format` parameter, that works as expected (namely I am able to use something like `'%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'` to customize the access logs:

```
127.0.0.1:48276 - - [01/Apr/2023:12:02:07 +0000] "GET /v1/ HTTP/1.1" 404 - "-" "curl/7.74.0"
```

For the moment the support is only for the asgi-logger's default, namely stdout, but we can surely add an option to log to disk or similar (but not sure if KServe users really need it).

- Logs

**Checklist**:

No idea if unit tests are needed in this case, lemme know if you fell differently.
Not sure if any are needed.

**Release note**:

```
Fix the default Uvicorn's log configuration and allow to override it using a custom json or yaml file. Add an experimental parameter to the model server class to support gunicorn's access_log_format config.
```

### DIFF
--- a/python/kserve/kserve/model_server.py
+++ b/python/kserve/kserve/model_server.py
@@ -55,6 +55,10 @@ parser.add_argument("--enable_docs_url", default=False, type=lambda x: bool(strt
                     help="Enable docs url '/docs' to display Swagger UI.")
 parser.add_argument("--enable_latency_logging", default=True, type=lambda x: bool(strtobool(x)),
                     help="Output a log per request with latency metrics.")
+parser.add_argument("--log_config_file", default=None, type=str,
+                    help="File path containing UvicornServer's log config. Needs to be a yaml or json file.")
+parser.add_argument("--access_log_format", default=None, type=str,
+                    help="Format to set for the access log (provided by asgi-logger).")
 
 args, _ = parser.parse_known_args()
 
@@ -76,6 +80,8 @@ class ModelServer:
         enable_grpc (bool): Whether to turn on grpc server. Default: ``True``
         enable_docs_url (bool): Whether to turn on ``/docs`` Swagger UI. Default: ``False``.
         enable_latency_logging (bool): Whether to log latency metric. Default: ``True``.
+        log_config_file (dict): File path containing UvicornServer's log config. Default: ``None``.
+        access_log_format (string): Format to set for the access log (provided by asgi-logger). Default: ``None``
     """
 
     def __init__(self, http_port: int = args.http_port,
@@ -86,7 +92,9 @@ class ModelServer:
                  registered_models: ModelRepository = ModelRepository(),
                  enable_grpc: bool = args.enable_grpc,
                  enable_docs_url: bool = args.enable_docs_url,
-                 enable_latency_logging: bool = args.enable_latency_logging):
+                 enable_latency_logging: bool = args.enable_latency_logging,
+                 log_config_file: str = args.log_config_file,
+                 access_log_format: str = args.access_log_format):
         self.registered_models = registered_models
         self.http_port = http_port
         self.grpc_port = grpc_port
@@ -100,6 +108,8 @@ class ModelServer:
         self.model_repository_extension = ModelRepositoryExtension(
             model_registry=self.registered_models)
         self._grpc_server = GRPCServer(grpc_port, self.dataplane, self.model_repository_extension)
+        self.log_config_file = log_config_file
+        self.access_log_format = access_log_format
 
     def start(self, models: Union[List[Model], Dict[str, Deployment]]) -> None:
         if isinstance(models, list):
@@ -143,7 +153,9 @@ class ModelServer:
                     sig, lambda s=sig: asyncio.create_task(self.stop(sig=s))
                 )
             self._rest_server = UvicornServer(self.http_port, [serversocket],
-                                              self.dataplane, self.model_repository_extension, self.enable_docs_url)
+                                              self.dataplane, self.model_repository_extension,
+                                              self.enable_docs_url, log_config_file=self.log_config_file,
+                                              access_log_format=self.access_log_format)
             if self.workers == 1:
                 await self._rest_server.run()
             else:
@@ -153,7 +165,8 @@ class ModelServer:
                 # https://github.com/tiangolo/fastapi/issues/1586
                 multiprocessing.set_start_method('fork')
                 server = UvicornServer(self.http_port, [serversocket],
-                                       self.dataplane, self.model_repository_extension, self.enable_docs_url)
+                                       self.dataplane, self.model_repository_extension,
+                                       self.enable_docs_url, custom_log_config=self.log_config)
                 for _ in range(self.workers):
                     p = Process(target=server.run_sync)
                     p.start()
--- a/python/kserve/kserve/protocol/rest/server.py
+++ b/python/kserve/kserve/protocol/rest/server.py
@@ -126,7 +126,8 @@ class RESTServer:
 
 class UvicornServer:
     def __init__(self, http_port: int, sockets: List[socket.socket],
-                 data_plane: DataPlane, model_repository_extension, enable_docs_url):
+                 data_plane: DataPlane, model_repository_extension, enable_docs_url,
+                 log_config_file: str = None, access_log_format: str = None):
         super().__init__()
         self.sockets = sockets
         rest_server = RESTServer(data_plane, model_repository_extension, enable_docs_url)
@@ -136,46 +137,71 @@ class UvicornServer:
             client=PrintTimings(),
             metric_namer=StarletteScopeToName(prefix="kserve.io", starlette_app=app)
         )
+        log_config = {
+            "version": 1,
+            "formatters": {
+                "default": {
+                    "()": "uvicorn.logging.DefaultFormatter",
+                    "datefmt": DATE_FMT,
+                    "fmt": "%(asctime)s.%(msecs)03d %(name)s %(levelprefix)s %(message)s",
+                    "use_colors": None,
+                },
+                "access": {
+                    "()": "uvicorn.logging.AccessFormatter",
+                    "datefmt": DATE_FMT,
+                    "fmt": '%(asctime)s.%(msecs)03d %(name)s %(levelprefix)s %(client_addr)s %(process)s - '
+                           '"%(request_line)s" %(status_code)s',
+                    # noqa: E501
+                },
+            },
+            "handlers": {
+                "default": {
+                    "formatter": "default",
+                    "class": "logging.StreamHandler",
+                    "stream": "ext://sys.stderr",
+                },
+                "access": {
+                    "formatter": "access",
+                    "class": "logging.StreamHandler",
+                    "stream": "ext://sys.stdout",
+                },
+            },
+            "loggers": {
+                "uvicorn": {"handlers": ["default"], "level": "INFO"},
+                "uvicorn.error": {"level": "INFO"},
+                "uvicorn.access": {"handlers": ["access"], "level": "INFO", "propagate": False},
+            },
+        }
+
+        # If the log_config value is a string ending up with ".json"
+        # or ".yaml", it is interpreted as file path and the log configuration
+        # is loaded from disk.
+        if log_config_file:
+            log_config = log_config_file
+
         self.cfg = uvicorn.Config(
             app=app,
             host="0.0.0.0",
+            log_config=log_config,
             port=http_port,
-            log_config={
-                "version": 1,
-                "formatters": {
-                    "default": {
-                        "()": "uvicorn.logging.DefaultFormatter",
-                        "datefmt": DATE_FMT,
-                        "fmt": "%(asctime)s.%(msecs)03d %(name)s %(levelprefix)s %(message)s",
-                        "use_colors": None,
-                    },
-                    "access": {
-                        "()": "uvicorn.logging.AccessFormatter",
-                        "datefmt": DATE_FMT,
-                        "fmt": '%(asctime)s.%(msecs)03d %(name)s %(levelprefix)s %(client_addr)s %(process)s - '
-                               '"%(request_line)s" %(status_code)s',
-                        # noqa: E501
-                    },
-                },
-                "rest": {
-                    "default": {
-                        "formatter": "default",
-                        "class": "logging.StreamHandler",
-                        "stream": "ext://sys.stderr",
-                    },
-                    "access": {
-                        "formatter": "access",
-                        "class": "logging.StreamHandler",
-                        "stream": "ext://sys.stdout",
-                    },
-                },
-                "loggers": {
-                    "uvicorn": {"rest": ["default"], "level": "INFO"},
-                    "uvicorn.error": {"level": "INFO"},
-                    "uvicorn.access": {"rest": ["access"], "level": "INFO", "propagate": False},
-                },
-            }
         )
+
+        # More context in https://github.com/encode/uvicorn/pull/947
+        # At the time of writing the ASGI specs are not clear when it comes
+        # to change the access log format, and hence the Uvicorn upstream devs
+        # chose to create a custom middleware for this.
+        # The allowed log format is specified in https://github.com/Kludex/asgi-logger#usage
+        if access_log_format:
+            from asgi_logger import AccessLoggerMiddleware
+            # As indicated by the asgi-logger docs, we need to clear/unset
+            # any setting for uvicorn.access to avoid log duplicates.
+            logging.getLogger("uvicorn.access").handlers = []
+            app.add_middleware(
+                AccessLoggerMiddleware, format=access_log_format)
+            # The asgi-logger settings don't set propagate to False,
+            # so we get duplicates if we don't set it explicitly.
+            logging.getLogger("access").propagate = False
+
         self.server = _NoSignalUvicornServer(config=self.cfg)
 
     def run_sync(self):
--- a/python/kserve/poetry.lock
+++ b/python/kserve/poetry.lock
@@ -204,6 +204,21 @@ test = ["contextlib2", "coverage[toml] (>=4.5)", "hypothesis (>=4.0)", "mock (>=
 trio = ["trio (>=0.16,<0.22)"]
 
 [[package]]
+name = "asgi-logger"
+version = "0.1.0"
+description = "Middleware based uvicorn access logger! :tada:"
+category = "main"
+optional = true
+python-versions = ">=3.8,<4.0"
+files = [
+    {file = "asgi-logger-0.1.0.tar.gz", hash = "sha256:b289f530bf49d14320242a567580ffc8ad053ba5667eece27d25a97822d3a0aa"},
+    {file = "asgi_logger-0.1.0-py3-none-any.whl", hash = "sha256:f2d1252cfc48b4a806d1810f0308ba2979fa73aca06ecc809903a8ebc5a6ef23"},
+]
+
+[package.dependencies]
+asgiref = ">=3.4.1,<4.0.0"
+
+[[package]]
 name = "asgiref"
 version = "3.6.0"
 description = "ASGI specs, helper code, and adapters"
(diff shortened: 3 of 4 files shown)


elukey:
Hi folks! Still need to figure out how to test this change, but early feedback is really good :)

elukey:
Added a simple test case to the PR's description, should be ready for a review :)

elukey:
lovely, `asgi-logger` requires python 3.8+, and of course some tests in CI depend on 3.7. Not sure what's the best forward, lemme know :)

elukey:
Some tests are still failing, IIUC due to the aix360 package that depends on tensorflow==1.14 that in turn is not compatible with Python 3.8 afaics..

elukey:
Summary for the reviewers (based on my understanding of the code):

* It seems that the `UvicornServer`'s default logging config is not working, so Uvicorn's access logs are not available in Kserve pods. The issue should be fixable simply replacing "rest" with "handlers" in the logging config.
* Uvicorn's config seems to support passing the path of a yaml or json file (instead of a dict) to change the logging configuration. I added the code to support it, that seems working nicely.
* The main issue of the above is that I cannot change the access log format, for example to dump some request headers like User Agent etc.. From https://github.com/encode/uvicorn/pull/947#issuecomment-1283006367 the way forward seems to use `asgi-logger` as Uvicorn Middleware, but for the moment I think it supports only printing to stdout (should be fine for most of the use cases in theory). The main drawback is that `asgi-logger` runs only on python3.8+, and this breaks CI. When I tried to upgrade all the docker images to python3.8-slim I found a problem with the `aix360` package, that supports on `tensorflow==1.14` and that in turn runs only on Python up to 3.7 (lovely).

If you want I can try to split the pull request into two steps:
* Fix/Customize access logging (so users will see something, but they will not be able to configure the access log format like they do with gunicorn for example).
* Work on the asgi-logger etc.. to figure out how to best add it to Kserve.

Lemme know :)

elukey:
Created https://github.com/Trusted-AI/AIX360/issues/172 to see if we can get a bump in the aix360's version.
Python 3.7 is going to be EOLed in a couple of months time, so if AIX360 is not getting updated I'd think about excluding it from CI.

rachitchauhan43:
@elukey : What about model_server logging_format ? Can you include changes to allow formatting that as well ?
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model_server.py#L61 

This PR has no changes https://github.com/kserve/kserve/pull/2762 and has been closed. If, we do not see the traction on this one, it would be great to get the changes corresponding to influence the logger format too via this PR itself.

elukey:
> @elukey : What about model_server logging_format ? Can you include changes to allow formatting that as well ? https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model_server.py#L61
> 
> This PR has no changes #2762 and has been closed. If, we do not see the traction on this one, it would be great to get the changes corresponding to influence the logger format too via this PR itself.

Hi! I think that we should keep the two things separated, my PR is related to Uvicorn's logging, meanwhile what you point out is related to the model server's logging. It is equally important but my PR is already big and I'd like to avoid adding things, unless the upstream devs suggests so. Could you open a PR like #2762 with the required changes?

elukey:
@yuzisun @theofpa Hi! If you have time could you give me some feedback about this change? Thanks a lot in advance :)

yuzisun:
> @yuzisun @theofpa Hi! If you have time could you give me some feedback about this change? Thanks a lot in advance :)

Sorry @elukey you may need to resolve the conflict as we have now migrated to poetry for python dependency resolution.

elukey:
> > @yuzisun @theofpa Hi! If you have time could you give me some feedback about this change? Thanks a lot in advance :)
> 
> Sorry @elukey you may need to resolve the conflict as we have now migrated to poetry for python dependency resolution.

Thanks a lot for the feedback, I have pushed another version of the pull request, but the `asgi-logger` problem remains, namely that it requires python 3.8.. One possible workaround is to add an extra (IIRC it is called in that way) for `kserve` to install it only if requested (something like `pip install kserve[extra-logging]` or similar). Do you have any preference? The access log customization could be an "experimental" feature and CI could keep working with Python 3.7 for the time being.

elukey:
Added the `asgi-logger` dependency as extra to poetry's config, this should avoid to move to python3.8 in all docker images :)

elukey:
@yuzisun should be ready for a review :)

elukey:
Forgot to run `poetry lock`, this is the result:

```
The current project's Python requirement (>=3.7.1,<3.11) is not compatible with some of the required packages Python requirement:
  - asgi-logger requires Python >=3.8,<4.0, so it will not be satisfied for Python >=3.7.1,<3.8

Because asgi-logger (0.1.0) requires Python >=3.8,<4.0
 and no versions of asgi-logger match >0.1.0,<0.2.0, asgi-logger is forbidden.
So, because kserve depends on asgi-logger (^0.1.0), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    
    For asgi-logger, a possible solution would be to set the `python` property to ">=3.8,<3.11"
```

elukey:
Found a workaround and re-tested everything, lemme know your thoughts :)

elukey:
@yuzisun Hi! Do you think that the PR could be reviewed? Let me know if it is missing anything major, it would be great if it was included in 0.11 (to get access log working again asap :).

python/kserve/kserve/model_server.py
```diff
@@ -76,6 +80,7 @@ class ModelServer:
         enable_grpc (bool): Whether to turn on grpc server. Default: ``True``
         enable_docs_url (bool): Whether to turn on ``/docs`` Swagger UI. Default: ``False``.
         enable_latency_logging (bool): Whether to log latency metric. Default: ``True``.
+        log_config_file (dict): File path containing UvicornServer's log config. Default: ``None``.
```
yuzisun, line 83: missing `access_log_format`

python/kserve/kserve/protocol/rest/server.py
```diff
@@ -136,46 +137,67 @@ def __init__(self, http_port: int, sockets: List[socket.socket],
...
+
+        # More context in https://github.com/encode/uvicorn/pull/947
+        # At the time of writing the ASGI specs are not clear when it comes
+        # to change the access log format, and hence the Uvicorn upstream devs
+        # chose to create a custom middleware for this.
+        # The allowed log format is specified in https://github.com/Kludex/asgi-logger#usage
+        if access_log_format:
+            from asgi_logger import AccessLoggerMiddleware
+            logging.getLogger("uvicorn.access").handlers = []
+            app.add_middleware(
```
yuzisun, line 199: How this is different from the `uvicorn.logging.AccessFormatter` config?

python/kserve/kserve/model_server.py
elukey: Added :)

python/kserve/kserve/protocol/rest/server.py
elukey: My understanding from
* https://github.com/encode/uvicorn/blob/master/uvicorn/logging.py#L102
* https://github.com/encode/uvicorn/pull/947#issuecomment-1283006367
is that the access log formatter provided by uvicorn is not as powerful as other ones (like Gunicorn's). For example, in my use case I'd need to log request headers like User Agent to then be able to collect them in logstash and have a nice aggregation in Kibana, but other people may have different use cases. With the default access log formatter we get a limited format to use, meanwhile with `asgi-logger` we have way more possibilities available. It is sad that uvicorn didn't really decide what to do yet, but I suppose this is due to ASGI specs (see what's written in the github issue above). Note that if one wants to use and override `uvicorn.logging.AccessFormatter` they can via `log_config_file`.

elukey: The idea of the patch is to have several type of access logs:
* By default, the one provided by the hardcoded config in the `UvicornServer` class, currently broken due to the missing `handlers` key in the python log config (so we don't have any log).
* A custom, Uvicorn based config provided via external yaml/json file, to customize the Uvicorn's `AccessFormatter` and the other ones (for example, emitting Json etc..).
* A custom, `asgi-logger` based logging format only for stdout (so limited but effective for the k8s use case imho) to get all the configs offered by Gunicorn (see https://docs.gunicorn.org/en/stable/settings.html#access-log-format).

yuzisun: @elukey thanks for the explanation. We also have starlette metrics logging for printing the time spent on the endpoints, do you think we still need it with the uvicorn access logging?

elukey: @yuzisun I'd keep it for the moment, so we can figure out with the community what's needed and what not. I didn't check but starlette shouldn't have its own access log right? Everything should go through uvicorn, but the precise relationship between the two is not 100% clear to me (if you want to shed some clarity I'd be very happy). If you want we can create a documentation page somewhere with all the details discussed in here, so folks in the community can refer to it and report use cases etc..

yuzisun: Starlette currently logs the cpu/wall time for each invocation of the rest endpoints like following
```
2023-04-23 20:26:26.788 6562 root INFO [timing():48] kserve.io.v1.model.custom-model 0.0009667873382568359, ['http_status:404', 'http_method:GET', 'time:wall']
2023-04-23 20:26:26.788 6562 root INFO [timing():48] kserve.io.v1.model.custom-model 0.0005819999999996384, ['http_status:404', 'http_method:GET', 'time:cpu']
```

So with all the access logging enabled, user would see three distinct log lines from `uvicorn access logging`, `asgi access logging` and `starlette timer logging` right?

elukey: @yuzisun: only two sources: uvicorn's access log (since `asgi-logger` overrides it, if used) and the startlette timer logging.

yuzisun: ok, let's document all the logging arguments [here](https://kserve.github.io/website/0.10/modelserving/v1beta1/custom/custom_model/#arguments)?

elukey: @yuzisun sure makes sense! Is it something that I can change in the KServe's repo or is it a sort-of-wiki/CMS that needs access to? (to understand if I have to modify the PR or not).

elukey: I see https://github.com/kserve/website/blob/5d86aa0ead60b6e35c80f46e42866ab25e38aa1d/docs/modelserving/v1beta1/custom/custom_model/README.md, I'll file a separate pull request with the new info :)

yuzisun: let’s create an issue in the website repo and submit the PR there

elukey: @yuzisun done! https://github.com/kserve/website/pull/255

elukey: mmm something must have changed, I tested the PR again and I see more logging than needed. Will do some research and report back.
Edit: it is the `"propagate": False` field in the default logging config that I removed, without it I see the new logs but often duplicated, meanwhile with it I don't see any access log at all.

yuzisun: I do not see this field on the diff, where do you see that ?

elukey: @yuzisun it is in the line:
`"uvicorn.access": {"rest": ["access"], "level": "INFO", "propagate": False`

I left out the `propagate` bit and I saw duplicated messages, and I noticed that the code to initialize the `AccessLoggerMiddleware` was not correct. Now I see the output as expected (namely as described above).

elukey: Output without access log format (so default):

```
2023-04-25 16:48:37.559 5441 root INFO [__call__():128] requestId: N.A., preprocess_ms: 0.04029274, explain_ms: 0, predict_ms: 0.028371811, postprocess_ms: 0.020742416
2023-04-25 16:48:37.560 uvicorn.access INFO:     127.0.0.1:49250 5441 - "POST /v1/models/test%3Apredict HTTP/1.1" 200 OK
2023-04-25 16:48:37.561 5441 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0045812129974365234, ['http_status:200', 'http_method:POST', 'time:wall']
2023-04-25 16:48:37.562 5441 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0044759999999997024, ['http_status:200', 'http_method:POST', 'time:cpu']
```

With access log format (User agent added etc..):

```
2023-04-25 16:46:57.905 5395 root INFO [__call__():128] requestId: N.A., preprocess_ms: 0.021219254, explain_ms: 0, predict_ms: 0.01001358, postprocess_ms: 0.004529953
2023-04-25 16:46:57.906 5395 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0013918876647949219, ['http_status:200', 'http_method:POST', 'time:wall']
2023-04-25 16:46:57.906 5395 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0013870000000002491, ['http_status:200', 'http_method:POST', 'time:cpu']
127.0.0.1:41344 - - [25/Apr/2023:16:46:57 +0000] "POST /v1/models/test:predict HTTP/1.1" 200 - "-" "curl/7.74.0"
2023-04-25 16:46:57.906 5395 access INFO [log():67] 127.0.0.1:41344 - - [25/Apr/2023:16:46:57 +0000] "POST /v1/models/test:predict HTTP/1.1" 200 - "-" "curl/7.74.0"
```

So now it seems working fine :) @yuzisun the code should be ready for another review :)

elukey: And also third option, custom config file (note the `Custom-cfg-file` prefix in the access log formatter):

```
version: 1
disable_existing_loggers: False
formatters:
  default:
    (): 'uvicorn.logging.DefaultFormatter'
    fmt: '%(asctime)s %(levelprefix)-9s %(name)s -: %(message)s'
  access:
    (): 'uvicorn.logging.AccessFormatter'
    fmt: 'Custom-cfg-file - %(asctime)s %(levelprefix)-9s %(name)s -: %(client_addr)s - "%(request_line)s" %(status_code)s'
handlers:
  default:
    class: logging.StreamHandler
    formatter: default
    stream: ext://sys.stderr
  access:
    class: logging.StreamHandler
    formatter: access
    stream: ext://sys.stdout
loggers:
  uvicorn:
    level: INFO
    handlers:
      - default
  uvicorn.error:
    level: INFO
  uvicorn.access:
    level: INFO
    propagate: False
    handlers:
      - access
```

and logs:

```
2023-04-25 16:55:54.004 5476 root INFO [__call__():128] requestId: N.A., preprocess_ms: 0.015735626, explain_ms: 0, predict_ms: 0.00667572, 
postprocess_ms: 0.003099442                                                                                                                 
Custom-cfg-file - 2023-04-25 16:55:54,005 INFO:     uvicorn.access -: 127.0.0.1:35892 - "POST /v1/models/test%3Apredict HTTP/1.1" 200 OK    
2023-04-25 16:55:54.005 5476 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0012352466583251953, ['http_statu
s:200', 'http_method:POST', 'time:wall']                                                                                                    
2023-04-25 16:55:54.005 5476 root INFO [timing():48] kserve.io.kserve.protocol.rest.v1_endpoints.predict 0.0012319999999998998, ['http_statu
s:200', 'http_method:POST', 'time:cpu'] 
```

elukey: Almost! Sigh :(
When using asgi-logger I still see a little duplication:

```
127.0.0.1:41344 - - [25/Apr/2023:16:46:57 +0000] "POST /v1/models/test:predict HTTP/1.1" 200 - "-" "curl/7.74.0"
2023-04-25 16:46:57.906 5395 access INFO [log():67] 127.0.0.1:41344 - - [25/Apr/2023:16:46:57 +0000] "POST /v1/models/test:predict HTTP/1.1" 200 - "-" "curl/7.74.0"
```

In theory it shouldn't happen, maybe it is due how to asgi-logger sets the logger + propagation. Will investigate :)

elukey: Ok should be fixed now, I added an extra setting for asgi-logger (to avoid propagation).

elukey: Asked in https://github.com/Kludex/asgi-logger/issues/6 some guidance about what I have written, maybe there is another/better way.

elukey: @yuzisun the change should be ready for a review, I checked and the output looks good now (no more repetitions).

python/kserve/kserve/model_server.py
```diff
@@ -55,6 +55,10 @@
                     help="Enable docs url '/docs' to display Swagger UI.")
 parser.add_argument("--enable_latency_logging", default=True, type=lambda x: bool(strtobool(x)),
                     help="Output a log per request with latency metrics.")
+parser.add_argument("--log_config_file", default=None, type=str,
```
yuzisun, line 58: @elukey Do you have an example how to pass this on the inference service yaml? I guess you need to mount this log file as ConfigMap

elukey: @yuzisun Yes exactly something similar, it just need a file to be present on the pod, it shouldn't be too difficult. I can try to come up with an example, lemme know if it is needed :)

yuzisun:
Looks goods! Thanks @elukey !

/lgtm
/approve

PR #2853, opened 2023-04-25
Fixes for Integrating KServe with Openshift

skonto:

This adds the following fixes for running KServe on Openshift:
- Adds the capability to inject custom annotations to the ksvc required by the [Openshift Serverless (deploys Knative Serving) and Service Mesh Integration](https://docs.openshift.com/container-platform/4.8/serverless/admin_guide/serverless-ossm-setup.html#serverless-ossm-setup_serverless-ossm-setup).
The annotation for example ` serving.knative.openshift.io/enablePassthrough: “true”` needs to be added at the ksvc level for inference services and graphs. This will not allow to add arbitrary annotations, thus should not change current semantics.
- Adds the ability to run the storage injector with user id 1337 in order to solve the issue when [Istio cni and dns proxy are enabled](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers).
Right now the security context is copied from the user container but we cant use the same user id there for many reasons eg. [is not allowed](https://istio.io/latest/docs/reference/config/analysis/ist0144/), security etc. The latter applies if we use pod security context or some other workaround.  
The uid can be set to an arbitrary value using the annotation: `serving.kserve.io/storage-initializer-uid: "..."`.
Note that on Openshift that uid is not 1337 but project_id_range+1, so we need this to be configurable.
Note: In a future PR we should be able to setup the Knative Serving init-containers feature flag and avoid the custom injection done or provide some template to allow the user to configure that part.

**Type of changes**

Tested on OCP 4.12 with the following inference service:
```
oc apply -n test -f - <<EOF
kind: InferenceService
metadata:
  name: sklearn-from-uri
  annotations:
    sidecar.istio.io/inject: "true"
    sidecar.istio.io/rewriteAppHTTPProbers: "true"
    serving.knative.openshift.io/enablePassthrough: “true”
spec:
  predictor:
    sklearn:
      storageUri: https://github.com/skonto/serverless-guides/blob/main/kserve/kfserving-examples/models/sklearn/1.0/model/model.joblib?raw=true
EOF
```
The manager deployment was started with the right env var. 

No

**Release note**:

```release-note
 Openshift integration improvements.
```

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -31,12 +31,13 @@ import (
 
 // KServe Constants
 var (
-	KServeName                     = "kserve"
-	KServeAPIGroupName             = "serving.kserve.io"
-	KnativeAutoscalingAPIGroupName = "autoscaling.knative.dev"
-	KnativeServingAPIGroupName     = "serving.knative.dev"
-	KServeNamespace                = getEnvOrDefault("POD_NAMESPACE", "kserve")
-	KServeDefaultVersion           = "v0.5.0"
+	KServeName                       = "kserve"
+	KServeAPIGroupName               = "serving.kserve.io"
+	KnativeAutoscalingAPIGroupName   = "autoscaling.knative.dev"
+	KnativeServingAPIGroupNamePrefix = "serving.knative"
+	KnativeServingAPIGroupName       = KnativeServingAPIGroupNamePrefix + ".dev"
+	KServeNamespace                  = getEnvOrDefault("POD_NAMESPACE", "kserve")
+	KServeDefaultVersion             = "v0.5.0"
 )
 
 // InferenceService Constants
@@ -131,8 +132,10 @@ var (
 
 // Controller Constants
 var (
-	ControllerLabelName = KServeName + "-controller-manager"
-	DefaultMinReplicas  = 1
+	ControllerLabelName              = KServeName + "-controller-manager"
+	DefaultMinReplicas               = 1
+	IstioCniDnsProxyEnabledEnvVarKey = "ISTIO_CNI_DNS_PROXY_ENABLED"
+	IstioSidecarUID                  = int64(1337)
 )
 
 type AutoscalerClassType string
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -84,6 +84,7 @@ var (
 	MinScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/min-scale"
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/max-scale"
 	RollOutDurationAnnotationKey                = KnativeServingAPIGroupName + "/rollout-duration"
+	KnativeOpenshiftEnablePassthroughKey        = "serving.knative.openshift.io/enablePassthrough"
 	EnableMetricAggregation                     = KServeAPIGroupName + "/enable-metric-aggregation"
 	SetPrometheusAnnotation                     = KServeAPIGroupName + "/enable-prometheus-scraping"
 	KserveContainerPrometheusPortKey            = "prometheus.kserve.io/port"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -132,10 +132,9 @@ var (
 
 // Controller Constants
 var (
-	ControllerLabelName              = KServeName + "-controller-manager"
-	DefaultMinReplicas               = 1
-	IstioCniDnsProxyEnabledEnvVarKey = "ISTIO_CNI_DNS_PROXY_ENABLED"
-	IstioSidecarUID                  = int64(1337)
+	ControllerLabelName          = KServeName + "-controller-manager"
+	DefaultMinReplicas           = 1
+	IstioSidecarUIDAnnotationKey = KServeAPIGroupName + "/storage-initializer-uid"
 )
 
 type AutoscalerClassType string
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
@@ -19,8 +19,6 @@ package knative
 import (
 	"context"
 	"fmt"
-	"strings"
-
 	"github.com/golang/protobuf/proto"
 	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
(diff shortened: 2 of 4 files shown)


skonto:
@yuzisun @alexagriffith  gentle ping.

skonto:
TODO: support scenario where OCP SM uses a uid other than 1337 to run the Istio side car.

ReToCode:
> TODO: support scenario where OCP SM uses a uid other than 1337 to run the Istio side car.

We could make it configurable? The default in Istio is 1337, but I think users could overwrite it.

skonto:
> We could make it configurable? The default in Istio is 1337, but I think users could overwrite it.

Yes that is my idea too given that OCP requires a project based id that varies.

skonto:
I have updated the PR. I am using an annotation to pass the uid for the storage initializer.
This can be refactored if a more generic solution is added for adjusting that container's properties.
The annotations on Openshift can be used as follows:

```
kubectl apply -n test -f - <<EOF
kind: InferenceService
metadata:
  name: sklearn-from-uri
  annotations:
    sidecar.istio.io/inject: "true"
    sidecar.istio.io/rewriteAppHTTPProbers: "true"
    serving.knative.openshift.io/enablePassthrough: "true"
    serving.kserve.io/storage-initializer-uid: "1000720001"
spec:
  predictor:
    sklearn:
      storageUri: https://github.com/skonto/serverless-guides/blob/main/kserve/kfserving-examples/models/sklearn/1.0/model/model.joblib?raw=true
EOF
```

skonto:
@yuzisun pls review, this is meant to make the Openshift guide added lately to work out of the box with Istio.

skonto:
@alexagriffith @yuzisun gentle ping.

yuzisun:
@alexagriffith and I am not too familiar with OpenShift, @ckadner can you help review this PR?

ckadner:
@skonto -- I have not deployed KServe in while, and not on OpenShift. Can you point me to how you deployed and configured your cluster? You can find me on Slack `@ckadner` (both on Kubeflow and the IBM workspaces), if that makes it easier.

yuzisun:
@skonto Can you help sign off the commit following https://github.com/kserve/kserve/pull/2853/checks?check_run_id=13400347618  ?

skonto:
@yuzisun I think its ready now.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -108,6 +109,14 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 		}
 	}
 
+	// Allow custom annotations for ksvcs that start with serving.knative but not part of serving.knative.dev group name.
+	for aKey, _ := range annotations {
+		if !strings.HasPrefix(aKey, constants.KnativeServingAPIGroupName) && strings.HasPrefix(aKey, constants.KnativeServingAPIGroupNamePrefix) {
```
yuzisun: Is it better to just check the full api group name `serving.knative.openshift.io`?

skonto: @yuzisun I thought about that but I didnt want to add vendor specific annotations, trying to keep it neutral for future use as well. Do you want me to change that?

yuzisun: @skonto I think this is no longer needed as we now allow component level labels and annotations, see https://github.com/kserve/kserve/pull/2925

skonto: @yuzisun unofrtunately that is per component. The requirement is to set this at the ksvc level, because the OCP Serverless reconciler expects that annotation (`serving.knative.openshift.io/enablePassthrough: "true"`) to be at that level.
For example here is the [diff](https://gist.github.com/skonto/be34e95512337c41a909c49b6ed4007b) between a kserve ksvc (does not work) and a regular ksvc that works as expected (I tried this using the new feature).

yuzisun: ok, l'd prefer using a more explicit list, we can add `serving.knative.openshift.io/enablePassthrough` to `managedKsvcAnnotations` ?

skonto: Yes that is possible. I will update it.

skonto: @yuzisun done.

ReToCode:
@yuzisun , @alexagriffith would you mind taking another look? We'd like to do some additional patching in opendatahub-io/kserve which is based on these changes.

pkg/constants/constants.go
```diff
@@ -131,8 +133,9 @@ var (
 
 // Controller Constants
 var (
-	ControllerLabelName = KServeName + "-controller-manager"
-	DefaultMinReplicas  = 1
+	ControllerLabelName          = KServeName + "-controller-manager"
+	DefaultMinReplicas           = 1
+	IstioSidecarUIDAnnotationKey = KServeAPIGroupName + "/storage-initializer-uid"
```
yuzisun, line 138: Are we planning to make this configurable [here](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L52) also without having to add this annotation to every inference service?

yuzisun:
> @yuzisun , @alexagriffith would you mind taking another look? We'd like to do some additional patching in opendatahub-io/kserve which is based on these changes.

Looks good, do we need more documentation updates?

pkg/constants/constants.go
ReToCode: The "issue" is, that this is specific to the namespace where the service is deployed. At the moment there is no better way than to specify it per InferenceService. We're working with the istio folks to hopefully get rid of the underlying issue.

ReToCode:
> Looks good, do we need more documentation updates?

Yes we should definitely update https://github.com/kserve/kserve/blob/master/docs/OPENSHIFT_GUIDE.md#prerequisites. Is it ok if I do a followup PR for this?

yuzisun:
Thanks @skonto @ReToCode ! We are including this to the upcoming 0.11 release 

/lgtm
/approve

PR #2869, opened 2023-05-02
Implement dynamic versioning using poetry plugin

### LINKED ISSUE #2800: Allow dynamic version extraction for poetry
/kind feature
This is to follow up the PR for migrating to use poetry, currently the version needs to be hard coded in the poetry package file but there are many packages need to be updated during the release, so we need a way to extract the version dynamically like the VERSION file we use previously.

reference: https://pypi.org/project/poetry-version-plugin/

andyi2it:
This PR adds provision to install a new poetry plugin from local, thereby dynamically setting the version from a VERSION file.
The plugin should be installed along with poetry using `pip install`.
The plugin would read the version from the file and dynamically use the version from it for the corresponding packages. 

Fixes #2800 

**Type of changes**

**Release note**:

```release-note
NONE
```


.github/workflows/python-test.yml
```diff
@@ -37,12 +37,17 @@ jobs:
...
+        env:
+          POETRY_VERSION: 1.4.0
+        run: |
+          pip install poetry==$POETRY_VERSION
+          poetry config virtualenvs.create true
+          poetry config virtualenvs.in-project true
+          poetry config installer.parallel true
+      - name: Install poetry version plugin
+        run: |
+          pip install -e python/plugin/poetry-version-plugin
```
yuzisun, line 49: Can we install the plug-in with poetry plugin
$ poetry plugin add poetry-version-plugin

yuzisun:
Discussed with @andyi2it, `poetry plugin` is deprecated and replaced with `poetry self add`, however this requires the plugin to be published, for now we have modified the plugin code so that it reads from the file instead of the way that plugin does.

PR #2881, opened 2023-05-05
Move cloud event decode logic from preprocess to decode method

sivanantha321:

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from importlib import metadata
 from typing import Dict, Union, Tuple, Optional
 
 import cloudevents.exceptions as ce
 import orjson
-import pkg_resources
 from cloudevents.http import CloudEvent, from_http
 from cloudevents.sdk.converters.util import has_binary_headers
 from ray.serve.api import RayServeHandle
@@ -43,7 +43,7 @@ class DataPlane:
 
         # Dynamically fetching version of the installed 'kserve' distribution. The assumption is
         # that 'kserve' will already be installed by the time this class is instantiated.
-        self._server_version = pkg_resources.get_distribution("kserve").version
+        self._server_version = metadata.version("kserve")
 
     @property
     def model_registry(self):
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -248,7 +248,7 @@ class DataPlane:
                 del attributes["data"]
         return decoded_body, attributes
 
-    def encode(self, model_name, body, response, headers, attributes: Dict) -> Tuple[Dict, Dict[str, str]]:
+    def encode(self, model_name, response, headers, req_attributes: Dict) -> Tuple[Dict, Dict[str, str]]:
         response_headers = {}
         # if we received a cloudevent, then also return a cloudevent
         is_cloudevent = False
@@ -260,7 +260,7 @@ class DataPlane:
             if headers.get("content-type", "") == "application/cloudevents+json":
                 is_cloudevent = True
         if is_cloudevent:
-            response_headers, response = create_response_cloudevent(model_name, body, response, attributes,
+            response_headers, response = create_response_cloudevent(model_name, response, req_attributes,
                                                                     is_binary_cloudevent)
 
             if is_binary_cloudevent:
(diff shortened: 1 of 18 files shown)


python/kserve/test/test_server.py
```diff
@@ -507,7 +507,8 @@ def test_predict_ce_avro_binary(self, http_server_client):
         writer.write(msg, encoder)
         data = bytes_writer.getvalue()
 
-        event = dummy_cloud_event(data, set_contenttype=True)
+        event = dummy_cloud_event(data)
+        event._attributes["content-type"] = "application/avro"
```
yuzisun: This should be set in the dummy_cloud_event and pass in the content_type as argument

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -217,10 +216,28 @@ def decode(self, body, headers) -> Union[Dict, InferRequest]:
...
+                # If decoding or parsing failed, check if it was supposed to be JSON UTF-8
+                if "content-type" in payload._attributes and \
+                        (payload._attributes["content-type"] == "application/cloudevents+json" or
+                         payload._attributes["content-type"] == "application/json"):
+                    raise InvalidInput(f"Failed to decode or parse binary json cloudevent: {e}")
+
+        elif isinstance(payload, dict):
+            if is_structured_cloudevent(payload):
+                response = payload["data"]
+        return response
```
yuzisun: I think this block of code can be merged into the `decode` function.
```python
if headers and has_binary_headers(headers):
    payload = self.get_binary_cloudevent(body, headers)
    try:
        body = orjson.loads(payload.data.decode('UTF-8'))
    except (orjson.JSONDecodeError, UnicodeDecodeError) as e:
         # If decoding or parsing failed, check if it was supposed to be JSON UTF-8
         if "content-type" in payload._attributes and \
                        (payload._attributes["content-type"] == "application/cloudevents+json" or
                         payload._attributes["content-type"] == "application/json"):
             raise InvalidInput(f"Failed to decode or parse binary json cloudevent: {e}")
elif isinstance(body, dict) and is_structured_cloudevent(payload):
    body = body["data"]
elif type(body) is bytes:
    try:
        body = orjson.loads(body)
    except orjson.JSONDecodeError as e:
        raise InvalidInput(f"Unrecognized request format: {e}")
```

sivanantha321: [create_response_cloudevent](https://github.com/kserve/kserve/blob/9bd6375957a82db022491daac49e64fdde604ac0/python/kserve/kserve/utils/utils.py#L93) requires cloudevent `attributes` field. This function is used in [encode method](https://github.com/kserve/kserve/blob/9bd6375957a82db022491daac49e64fdde604ac0/python/kserve/kserve/protocol/dataplane.py#L236). if we add this logic to decode method `attributes` field will not be available because, the returned body only contains cloudevent `data`.

yuzisun: I see that’s the binary cloudevent case, can we return a tuple with the attributes and then pass along to the encode function ?

python/kserve/test/test_server.py
sivanantha321: done

python/kserve/kserve/protocol/dataplane.py
sivanantha321: done

python/kserve/kserve/utils/utils.py
```diff
@@ -94,18 +94,17 @@ def is_structured_cloudevent(body: Dict) -> bool:
            and "data" in body
 
 
-def create_response_cloudevent(model_name: str, body: Union[Dict, CloudEvent], response: Dict,
+def create_response_cloudevent(model_name: str, body: Union[Dict, CloudEvent], response: Dict, attributes: Dict,
```
yuzisun: body arg is no longer needed looks

sivanantha321: removed!

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -31,7 +31,7 @@ class InferInput:
     _datatype: str
     _parameters: Dict
 
-    def __init__(self, name, shape, datatype, data=None, parameters={}):
+    def __init__(self, name, shape, datatype, data=None, parameters=None):
```
sivanantha321, line 34: We should not use mutable objects in arguments as it can introduce [unintended side effects](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

Issue #2676, opened 2023-02-05
KServe 0.11.0 release tracking


rachitchauhan43:
/kind feature

**Describe the solution you'd like**
KServe 0.11.0 release tracking:
RC0 release Date: Will be released in the week of `Friday, April 28, 2023 - Thursday, May 4, 2023`

Release Date: Will be released in the week of `Friday, June 9, 2023 | Thursday, June 15,   2023`

## KServe Model Serving:
See [project dashboard](https://github.com/orgs/kserve/projects/6/views/1?sortedBy%5Bdirection%5D=asc&sortedBy%5BcolumnId%5D=Status) to see what all features/bugs are covered as part 0.11 and their current status 

## Website:

# **TIMELINE**
| Week | Start                     | End                         | When | Event (Actual timelines)                                                                                |
| ---- |---------------------------| --------------------------- | ---- | ------------------------------------------------------------------------------------------------------- |
| 0    | Friday, January 27, 2023  | Thursday, February 2, 2023  |      | Gap week between releases. 0.10.0 release happened on Sunday, January 22, 2023                          |
| 1    | Friday, February 3, 2023  | Thursday, February 9, 2023  |      | Development                                                                                             |
| 2    | Friday, February 10, 2023 | Thursday, February 16, 2023 |      | Development                                                                                             |
| 3    | Friday, February 17, 2023 | Thursday, February 23, 2023 |      | Development                                                                                             |
| 4    | Friday, February 24, 2023 | Thursday, March 2, 2023     |      | Development                                                                                             |
| 5    | Friday, March 3, 2023     | Thursday, March 9, 2023     |      | Development                                                                                             |
| 6    | Friday, March 10, 2023    | Thursday, March 16, 2023    |      | Development                                                                                             |
| 7    | Friday, March 17, 2023    | Thursday, March 23, 2023    |      | Development                                                                                             |
| 8    | Friday, March 24, 2023    | Thursday, March 30, 2023    |      | Development                                                                                             |
| 9    | Friday, March 31, 2023    | Thursday, April 6, 2023     |      | Development                                                                                             |
| 10   | Friday, April 7, 2023     | Thursday, April 13, 2023    |      | Development                                                                                             |
| 11   | Friday, April 14, 2023    | Thursday, April 20, 2023    |      | Development +<br>Start the prep i.e.<br>Announce/Reminder about upcoming feature freeze date in 2 weeks |
[...]

PR #2893, opened 2023-05-11
Created a new RC release version for version 0.11.0

andyi2it:
Creates a new pre-release for KServe

https://github.com/kserve/kserve/issues/2676

**Type of changes**

**Release note**:

```release-note
NONE
```


python/VERSION
```diff
@@ -1 +1 @@
-0.10.1
\ No newline at end of file
+0.11.0-rc0
```
yuzisun: The python version convention is 0.11.0rc0

charts/kserve-resources/values.yaml
```diff
@@ -1,5 +1,5 @@
 kserve:
-  version: &defaultVersion v0.10.1
+  version: &defaultVersion v0.11.0-rc0
   modelmeshVersion: &defaultModelMeshVersion v0.10.0
```
ckadner, line 3: FYI: we will update the ModelMesh version here after the first KServe RC (@rafvasq)

ckadner: @rafvasq

PR #2917, opened 2023-05-19
Checked model readiness before attempting inference.

### LINKED ISSUE #2775: KServe REST Server doesn't check model readiness before attempting inference.
/kind feature

As of now, it's possible to perform inference over a `Model` when the model is not in the ready state.
Neither the V1, nor V2 REST endpoints check whether `Model.ready` has been set before attempting inference:
- https://github.com/kserve/kserve/blob/236c1dcff60a4932b4db28c6cec9ce6b1dc470f0/python/kserve/kserve/protocol/rest/v1_endpoints.py#L55
- https://github.com/kserve/kserve/blob/236c1dcff60a4932b4db28c6cec9ce6b1dc470f0/python/kserve/kserve/protocol/rest/v2_endpoints.py#L130

I understand that it's expected that the client checks the separate readiness endpoint before sending inference requests, and that the InferenceService itself should not be marked as ready in Kubernetes until the readiness endpoint returns 200, but I think the inference endpoints should raise `ModelNotReady` themselves when appropriate.

Is there a principled reason for not including this that I'm not aware of? Happy to make a PR to fix this if it's deemed necessary. 

This caused an issue in our deployed environment, where one client which checked the readiness endpoint was failing on a deployed inference service, but another client which did not check the readiness was returning successful predictions, which caused a lot of confusion until we realised we'd forgotten to set the `.ready` flag on our `Model`.

andyi2it:
Adds a model ready check before serving the request.

Fixes #2775 

**Type of changes**

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/alibiexplainer/alibiexplainer/explainer.py
+++ b/python/alibiexplainer/alibiexplainer/explainer.py
@@ -64,9 +64,6 @@ class AlibiExplainer(kserve.Model):
         else:
             raise NotImplementedError
 
-    def load(self):
-        self.ready = True
-
     def _predict_fn(self, arr: Union[np.ndarray, List]) -> np.ndarray:
         instances = []
         for req_data in arr:
--- a/python/alibiexplainer/alibiexplainer/explainer.py
+++ b/python/alibiexplainer/alibiexplainer/explainer.py
@@ -64,8 +64,8 @@ class AlibiExplainer(kserve.Model):
         else:
             raise NotImplementedError
 
-    def load(self) -> bool:
-        pass
+    def load(self):
+        self.ready = True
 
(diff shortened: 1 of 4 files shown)


python/alibiexplainer/alibiexplainer/explainer.py
```diff
@@ -64,8 +64,8 @@ def __init__(  # pylint:disable=too-many-arguments
         else:
             raise NotImplementedError
 
-    def load(self) -> bool:
-        pass
+    def load(self):
+        self.ready = True
```
yuzisun: We can remove this as it is already defined in base https://github.com/kserve/kserve/blob/b59dde1739a29dd3d6484f6e4ea2039d85f17617/python/kserve/kserve/model.py#L173

Issue #2851, opened 2023-04-25
KServe RestServer Disabling Other Customized Module Level Loggings


Yucao42:
/kind bug

**What steps did you take and what happened:**
We have used our own module level loggings but no log from our code will show up. Our logs are likely to be suppressed by this part of the [code](https://github.com/kserve/kserve/blob/389659a0c5067d2086cf960dcb1149d676254025/python/kserve/kserve/protocol/rest/server.py#L141-L176) where `disable_existing_loggers=False` is not included. Since this value is defaulted to be `True` according to the [document](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema) of logging . Omitting this will disable every other logger but just keep the kserve *rest server ones*.

**What did you expect to happen:** Our customized module level loggings to show up.

**What's the InferenceService yaml:**
[To help us debug please run `kubectl get isvc $name -n $namespace -oyaml` and paste the output]

**Anything else you would like to add:**
A one line change of adding  `disable_existing_loggers=False` between [code](https://github.com/kserve/kserve/blob/389659a0c5067d2086cf960dcb1149d676254025/python/kserve/kserve/protocol/rest/server.py#L141-L176) might be able to fix this. Also we might want to use module level logger in `model_server` [here](https://github.com/kserve/kserve/blob/389659a0c5067d2086cf960dcb1149d676254025/python/kserve/kserve/model_server.py#L63) according to [logging library documentation which suggests not using root logger](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) just like the used uvicorn server [logger](https://github.com/encode/uvicorn/blob/master/uvicorn/server.py#L34). Thank you!

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Cloud Environment:[k8s_istio/istio_dex/gcp_basic_auth/gcp_iap/aws/aws_cognito/ibm]
- Minikube/Kind version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

yuzisun:
Thanks @Yucao42 for submitting the issue!

yuzisun:
Thanks @Yucao42 for submitting the issue!

amitchnick:
[...]

PR #2927, opened 2023-05-22
python: remove last traces of log_config_file (#2851)

elukey:
After a chat with @yuzisun in:

https://github.com/kserve/website/pull/266

it was suggested to remove `log_config_file` in favor of a more standardized `log_config`. To keep argparse and the model server's signatures inline, this change adds the possibility to accept as input a dictionary "serialized" as string, unpacking it with the json library. I am not sure what users will use this functionality, but let's discuss if we want it or not as suggested by @yuzisun.

### DIFF
--- a/python/kserve/kserve/model_server.py
+++ b/python/kserve/kserve/model_server.py
@@ -15,6 +15,7 @@
 import argparse
 import asyncio
 import concurrent.futures
+import json
 import multiprocessing
 import signal
 import socket
@@ -56,8 +57,10 @@ parser.add_argument("--enable_latency_logging", default=True, type=lambda x: boo
                     help="Output a log per request with latency metrics.")
 parser.add_argument("--configure_logging", default=True, type=lambda x: bool(strtobool(x)),
                     help="Whether to configure KServe and Uvicorn logging")
-parser.add_argument("--log_config_file", default=None, type=str,
-                    help="File path containing UvicornServer's log config. Needs to be a yaml or json file.")
+parser.add_argument("--log_config", default=None, type=str,
+                    help=("File path or Dictionary (serialized as string) containing "
+                          "UvicornServer's log config. "
+                          "The file path needs to point to a yaml or json file."))
 parser.add_argument("--access_log_format", default=None, type=str,
                     help="Format to set for the access log (provided by asgi-logger).")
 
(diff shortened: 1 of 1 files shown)


elukey:
@yuzisun @cmaddalozzo opened this pull request to see if we want to have a more consistent signature, but as written in the commit I am not 100% sure about how many people will need/use/pass a dict via argparse.

I will update https://github.com/kserve/website/pull/266 once we decide how to proceed, thanks!

python/kserve/kserve/model_server.py
```diff
@@ -56,8 +57,10 @@
                     help="Output a log per request with latency metrics.")
 parser.add_argument("--configure_logging", default=True, type=lambda x: bool(strtobool(x)),
                     help="Whether to configure KServe and Uvicorn logging")
-parser.add_argument("--log_config_file", default=None, type=str,
-                    help="File path containing UvicornServer's log config. Needs to be a yaml or json file.")
+parser.add_argument("--log_config", default=None, type=str,
```
yuzisun, line 60: Did you try setting the type to dict so you do not need to parse the json in your code ?

elukey: @yuzisun I haven't tried since in theory we shouldn't be able to use a file path right? Or will it be possible?

elukey: I am also trying to test type=dict in argparse but I don't find a good way to pass the dict via command line (the str format seems not ok when type is dict). Does it work for you?

yuzisun: you are right, this does not work

cmaddalozzo: I don't know if it makes sense to change the name of this arg since it's very difficult to pass a dict as an argument from the command line. If instantiating the model server programmatically it is possible to pass the argument as dictionary. That's why we've named the function parameter `log_config` (as it can can be a file path or a dictionary), while keeping the CLI argument as `log_config_file`. Does this make sense?

python/kserve/kserve/model_server.py
```diff
@@ -114,11 +117,18 @@ def __init__(self, http_port: int = args.http_port,
         # Logs can be passed as a path to a file or a dictConfig.
         # We rely on Uvicorn to configure the loggers for us.
         if configure_logging:
-            self.log_config = log_config if log_config is not None else KSERVE_LOG_CONFIG
+            # log_config could be a dictionary serialized as string,
```
cmaddalozzo, line 120: I'd suggest if we want to enable the user to provide a full logging config without needing to create a log config file we should do it via an environment variable. I don't think it's a great practice to pass really large strings as command line arguments. Maybe we can look for a `KSERVE_LOG_CONFIG` environment variable and expect that to be a JSON formatted string containing the dict config.

python/kserve/kserve/model_server.py
elukey: @yuzisun any preference? I am happy to leave things as they are :)

yuzisun: Ok sounds good ! Thanks for the PR though

yuzisun:
@elukey We can close this?

PR #2977, opened 2023-06-08
fix IngressReady condition

### LINKED ISSUE #2733: Improve InferenceService status
**Is your feature request related to a problem? If so, please describe.**

https://github.com/kserve/modelmesh-serving/blob/main/docs/predictors/inferenceservice-cr.md

> conditions - Various condition entries. Pertinent entries are:
>
> PredictorReady: predictor readiness condition. Status is true when the predictor's endpoints are ready to serve inferencing requests. Note that this does not necessarily mean requests will respond immediately, the corresponding model may or may not be loaded in memory. In the case that it isn't there may be some delay before the response comes back.
>
> Ready: aggregated condition of all conditions.

The “Ready” status is confusing. The model can be callable when the `PredictorReady` is False. What are the `conditions` here? What does `PredictorReady` mean? Is Predictor CR still used or is it just a legacy way of calling things? It would be nice if there is a way distinguish if the model has some non-retriable error like the wrong file path, or if it’s just temporarily not loaded due to lack of usage and can be available when called. 

**Describe your proposed solution**

Talking with @yuzisun, he suggests that
> currently it is really reflecting the “model deployment status”  as your active model is still callable but the target model deployment may fail.

Rename the current Status to "model deployment status", and have something to indicate the current model's status. Right now I think it's just the activeModelState. 

**Describe alternatives you have considered**
Since
> Upon creation, the active model status of an InferenceService will always transition to Loaded state (unless the loading fails), but later if unused, it is possible that the active model status ends up in a Standby state which means the model is still available to serve requests but the first request could incur a loading delay. 

Use activeModelState in combination with lastFailureInfo, if a model has been loaded before, it shouldn't have any lastFailureInfo. 

**Additional context**

tessapham:

Fixes the IngressReady condition to not automatically switch to False when service is being deployed, which in turn fixes the Ready status being set to False prematurely.

This PR is the first step to fix https://github.com/kserve/kserve/issues/2733

**Type of changes**

Tested manually by creating an inference service and watching its Ready status and spec updates. Also tested edge cases: isvc with failed transformer, isvc with failed predictor.

Validation: The Ready status stays Unknown while deployment is in progress.

**Release note**:

```release-note
NONE.
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -278,9 +278,13 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
 	}
 
 	if !isvc.Status.IsConditionReady(v1beta1.PredictorReady) {
+		status := corev1.ConditionFalse
+		if isvc.Status.IsConditionUnknown(v1beta1.PredictorReady) {
+			status = corev1.ConditionUnknown
+		}
 		isvc.Status.SetCondition(v1beta1.IngressReady, &apis.Condition{
 			Type:   v1beta1.IngressReady,
-			Status: corev1.ConditionFalse,
+			Status: status,
 			Reason: "Predictor ingress not created",
 		})
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -277,7 +277,7 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
 		return nil
 	}
 
-	if !isvc.Status.IsConditionReady(v1beta1.PredictorReady) {
+	if isvc.Status.IsConditionFalse(v1beta1.PredictorReady) {
```
yuzisun: I think we need to check `IsConditionUnknown` as well to return early, the ingress should not be created until all of the components are in ready state.
https://github.com/kserve/kserve/actions/runs/5215454086/jobs/9413002639?pr=2977

yuzisun:
Thanks @tessapham ! Looks good!

/lgtm
/approve

PR #3008, opened 2023-06-23
Add RoutesReady and LastDeploymentReady status conditions

### LINKED ISSUE #2733: Improve InferenceService status

tessapham:

Fixes https://github.com/kserve/kserve/issues/2733

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -116,8 +116,8 @@ const (
 	ExplainerReady apis.ConditionType = "ExplainerReady"
 	// IngressReady is set when Ingress is created
 	IngressReady apis.ConditionType = "IngressReady"
-	// ServiceReady is set when network configurations for all components have completed.
-	ServiceReady apis.ConditionType = "ServiceReady"
+	// RoutesReady is set when network configurations for all components have completed.
+	RoutesReady apis.ConditionType = "RoutesReady"
 )
 
 type ModelStatus struct {
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -272,7 +272,7 @@ func (ss *InferenceServiceStatus) InitializeConditions() {
 	conditionSet.Manage(ss).InitializeConditions()
 }
 
-// IsReady returns if the service is ready to serve the requested configuration.
+// IsReady returns the readiness for the latest revision.
 func (ss *InferenceServiceStatus) IsReady() bool {
 	return conditionSet.Manage(ss).IsHappy()
 }
@@ -346,7 +346,7 @@ func (ss *InferenceServiceStatus) PropagateServiceReadyStatus(component Componen
 	statusSpec := ss.Components[component]
 	serviceConditionType := serviceConditionsMap[component]
 	totalTraffic := int64(0)
-	for _, traffic := range statusSpec.Traffic{
+	for _, traffic := range statusSpec.Traffic {
 		if traffic.Percent != nil {
 			totalTraffic += *traffic.Percent
 		}
@@ -411,7 +411,7 @@ func (ss *InferenceServiceStatus) PropagateStatus(component ComponentType, servi
 			*traffic.LatestRevision {
 			if statusSpec.LatestRolledoutRevision != serviceStatus.LatestReadyRevisionName {
 				if traffic.Percent != nil && *traffic.Percent == 100 {
-					// track the last revision that's rolled out
+					// track the last revision that's fully rolled out
 					statusSpec.PreviousRolledoutRevision = statusSpec.LatestRolledoutRevision
 					statusSpec.LatestRolledoutRevision = serviceStatus.LatestReadyRevisionName
 				}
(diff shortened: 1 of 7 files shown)


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -339,17 +339,16 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
...
 	}
-	statusSpec, _ := isvc.Status.Components[v1beta1.PredictorComponent]
+	statusSpec := isvc.Status.Components[v1beta1.PredictorComponent]
 	if rawDeployment {
 		podLabelValue = constants.GetRawServiceLabel(predictorName)
 	} else {
 		podLabelValue = statusSpec.LatestCreatedRevision
 	}
-	podList, err := isvcutils.ListPodsByLabel(p.client, isvc.ObjectMeta.Namespace, podLabelKey, podLabelValue)
+	predictorPod, err := isvcutils.ListPodsByLabel(p.client, isvc.ObjectMeta.Namespace, podLabelKey, podLabelValue)
```
yuzisun: should be `predictorPods` ? it is the pod list for the latest revision

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -116,6 +116,8 @@ const (
 	ExplainerReady apis.ConditionType = "ExplainerReady"
 	// IngressReady is set when Ingress is created
 	IngressReady apis.ConditionType = "IngressReady"
+	// ServiceReady is set when network configurations for all components have completed.
+	ServiceReady apis.ConditionType = "ServiceReady"
```
yuzisun: Should we rename this to `RoutesReady` ? now I think the aggregation of `ConfigurationReady` conditions is actually the deployment status, the `Ready` condition is really the further aggregation of the two.

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -220,6 +221,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 									"key3":                                "val3FromPredictor",
 								},
 								Annotations: map[string]string{
+									"serving.kserve.io/deploymentMode":                         "Serverless",
```
yuzisun, line 224: I guess we need to add this here because in the test we did not set the `DefaultDeploymentMode` from config right?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
tessapham: It actually only gets 1 pod for the latest revision: https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go#L486.

pkg/controller/v1beta1/inferenceservice/controller_test.go
tessapham: @yuzisun In this test the inference services weren't created with a `deployConfig` as they weren't created using [`SetupWithManager()`](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/controller.go#L277), so I added this annotation as a workaround. Without the deployment mode, `RoutesReady` status wouldn't be set.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -326,6 +331,29 @@ func getDeploymentCondition(deployment *appsv1.Deployment, conditionType appsv1.
...
+		Type:   conditionType,
+		Status: v1.ConditionTrue,
+	}
+	for _, component := range componentList {
+		if !ss.IsConditionReady(conditionsMap[component]) {
+			crossComponentCondition.Status = v1.ConditionFalse
+			if ss.IsConditionUnknown(conditionsMap[component]) { // include check for nil condition
+				crossComponentCondition.Status = v1.ConditionUnknown
+			}
+			crossComponentCondition.Reason = string(conditionsMap[component]) + " not ready"
```
yuzisun, line 351: I think we should aggregate the reasons from individual conditions so we can show the deployment error message

pkg/controller/v1beta1/inferenceservice/components/predictor.go
yuzisun: though here is still returning a list of pods and in status propagation method it only fetches the first pod.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -230,7 +242,7 @@ func (r *InferenceServiceReconciler) updateStatus(desiredService *v1beta1api.Inf
 	if err := r.Get(context.TODO(), namespacedName, existingService); err != nil {
 		return err
 	}
-	wasReady := inferenceServiceReadiness(existingService.Status)
+	wasReady := isvcLatestRevisionReadiness(existingService.Status)
```
yuzisun: We probably need to change back here as the ready condition also covers the routesReady which may not necessarily mean for the latest revision as we found out.

PR #3034, opened 2023-07-16
Support secret name overrides with annotation

### LINKED ISSUE #284: option to load credentials directly from a secret for s3
Currently to download from a private S3 bucket [you create both a secret and a service account that you link it to](https://github.com/kubeflow/kfserving/blob/master/docs/samples/s3/s3_secret.yaml). You then set the [serviceAccountName on the KFService ](https://github.com/kubeflow/kfserving/blob/master/docs/samples/s3/tensorflow_s3.yaml#L7) and it is [used to add env vars based on the secret to the initContainer that downloads the model](https://github.com/kubeflow/kfserving/blob/master/pkg/controller/kfservice/resources/credentials/service_account_credentials.go#L94).

It might be easier for s3 users to just create a secret containing entries intended as environment variables and link that directly by having a 'envSecretRefName' entry in the CRD. This could be used in the implementation to use 'envFrom' to apply the secret's values.

It seems the[ original idea for credentials was to use a Secret ](https://github.com/kubeflow/kfserving/issues/36)and this morphed into a ServiceAccount as a 'first pass'. Presumably there's no in principle objection to also supporting a direct secret for s3?

yuzisun:

This is to allow overriding the secret names via annotation and kserve by default gets the secret references from the service account.

Fixes #284 

**Type of changes**

- [X]  Add the `serving.kserve.io/storageSecretName` annotation to the inference service yaml and KServe gets the secret reference from annotation instead of service account.

- Logs

**Checklist**:

- [X] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
user can now override the storage secret name by adding the annotation `serving.kserve.io/storageSecretName`
```


pkg/credentials/service_account_credentials_test.go
```diff
@@ -167,7 +167,7 @@ func TestS3CredentialBuilder(t *testing.T) {
 		g.Expect(c.Create(context.TODO(), existingServiceAccount)).NotTo(gomega.HaveOccurred())
 		g.Expect(c.Create(context.TODO(), existingS3Secret)).NotTo(gomega.HaveOccurred())
 
-		err := builder.CreateSecretVolumeAndEnv(scenario.serviceAccount.Namespace, scenario.serviceAccount.Name,
+		err := builder.CreateSecretVolumeAndEnv(scenario.serviceAccount.Namespace, nil, scenario.serviceAccount.Name,
```
greenmoon55: Do we have a test case with both `annotations` and `storageSecretName` to make sure `annotations` take precedence?

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -97,6 +97,7 @@ data:
      # Example
```
greenmoon55, line 93: Can you add a comment to line 88 above that `storageSecretNameAnnotation` takes precedence over`storageSpecSecretName`?

config/configmap/inferenceservice.yaml
```diff
@@ -123,6 +124,10 @@ data:
      # The rest of the fields are used in both authentication methods (IAM Role for Service Account & IAM User Access Key Secret) if a non-empty value is provided.
      credentials: |-
        {
+          # KServe by default gets the secret reference from service account, the configured annotation allows overriding with
```
greenmoon55: nit: `KServe by default gets the secret reference from service account` if you use storage uri instead of storage spec.

pkg/credentials/service_account_credentials.go
```diff
@@ -203,56 +200,94 @@ func (c *CredentialBuilder) CreateSecretVolumeAndEnv(namespace string, serviceAc
...
+			}
 		}
+	} else {
+		for _, secretRef := range serviceAccount.Secrets {
+			err := c.mountSecretCredential(secretRef.Name, namespace, container, volumes)
+			if err != nil {
+				continue
+			}
+		}
+	}
```
greenmoon55: Can we return early in Line 209 to avoid the duplication here?

greenmoon55 (changes requested):
Overall LGTM, some minor comments.

hack/quick_install.sh
```diff
@@ -30,7 +30,7 @@ while getopts ":hsr" option; do
    esac
 done
 
-export ISTIO_VERSION=1.16.2
```
greenmoon55, line 33: Was this intentional?

pkg/credentials/service_account_credentials.go
```diff
@@ -55,38 +56,51 @@ var (
...
+	StorageSecretNameAnnotation string        `json:"storageSecretNameAnnotation,omitempty"`
 }
 
 type CredentialBuilder struct {
 	client client.Client
 	config CredentialConfig
 }
 
-var log = logf.Log.WithName("CredentialBulder")
+var log = logf.Log.WithName("CredentialBuilder")
```
greenmoon55, line 70: haha good catch!

pkg/credentials/service_account_credentials_test.go
greenmoon55: It would be nice to have a new test case because the default config comes with both `annotations` and `storageSecretName`.

hack/quick_install.sh
yuzisun: yes, bumped up as this is the recommended version for kserve 0.11

pkg/credentials/service_account_credentials_test.go
```diff
@@ -36,6 +36,8 @@ import (
 var configMap = &v1.ConfigMap{
 	Data: map[string]string{
 		"credentials": `{
+            "storageSecretNameAnnotation": "serving.kserve.io/storageSecretName",
```
greenmoon55, line 39: nit: indentation?

greenmoon55 (approved):
Don't forget to add the release note!

PR #3039, opened 2023-07-19
Inference Graph error response handling

### LINKED ISSUE #2484: Inference Graph’s response code should be influenced by node’s response code

rachitchauhan43:

Fixes #2484 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Add dependency field(Soft/Hard) to the inference graph steps.
```


config/crd/serving.kserve.io_clusterservingruntimes.yaml
```diff
@@ -1,9 +1,11 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.12.0
+    controller-gen.kubebuilder.io/version: v0.4.0
```
sivanantha321: Hey @rachitchauhan43, I noticed you are using old controller-gen version.  The controller-gen version is upgraded to 0.12.0. You should delete the stale binary file inside the bin folder.

rachitchauhan43: Thanks for pointing out that. Fixed it.

rachitchauhan43:
@yuzisun / @Iamlovingit : Can i have some reviews on this ?

cmd/router/main.go
```diff
@@ -56,15 +57,16 @@ func callService(serviceUrl string, input []byte, headers http.Header) ([]byte,
...
+		log.Error(err, "An error has occurred while calling service", "service", serviceUrl)
+		return nil, 500, err
 	}
 	defer resp.Body.Close()
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		log.Error(err, "error while reading the response")
 	}
-	return body, err
+	log.Info("Exiting callService", "serviceUrl", serviceUrl)
```
yuzisun: Should this be at debug level ?

cmd/router/main.go
```diff
@@ -98,33 +100,72 @@ func timeTrack(start time.Time, nodeOrStep string, name string) {
...
 		if route == nil {
-			return input, nil //TODO maybe should fail in this case?
+			errorMessage := "None of the routes matched with the switch condition"
+			log.Error(err, errorMessage)
+			/*
+				TODO Should it be 500 or 404 when none of the routes match.
+				For backward compatibility 500 is needed but 404 also looks fine since the router got bad input and didn't know what to do.
+				See what reviewers say.
+			*/
+			return nil, 500, errors.New(errorMessage)
```
yuzisun: I agree 404 makes more sense.

cmd/router/main.go
```diff
@@ -133,69 +174,120 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
 }
 
+func prepareErrorResponse(err error, errorMessage string) []byte {
+	igRoutingErr := &InferenceGraphRoutingError{
+		errorMessage,
+		fmt.Sprintf("%v", err),
+	}
+	errorResponseBytes, err := json.Marshal(igRoutingErr)
+	if err != nil {
+		log.Error(errors.New("error happened while marshalling"), "marshalling error")
```
yuzisun: Log the actual error

cmd/router/main_test.go
```diff
@@ -82,13 +82,14 @@ func TestSimpleModelChainer(t *testing.T) {
...
-	res, err := routeStep("root", graphSpec, jsonBytes, headers)
+	res, statusCode, err := routeStep("root", graphSpec, jsonBytes, headers)
 	var response map[string]interface{}
 	err = json.Unmarshal(res, &response)
 	expectedResponse := map[string]interface{}{
 		"predictions": "2",
 	}
 	fmt.Printf("final response:%v\n", response)
+	fmt.Println(statusCode)
 	assert.Equal(t, expectedResponse, response)
```
yuzisun: Can be combined to the previous log line

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -254,6 +268,10 @@ type InferenceStep struct {
 	// routing based on the condition
 	// +optional
 	Condition string `json:"condition,omitempty"`
+
+	// to decide whether a step is a hard or a soft dependency in the Inference Graph
+	// +optional
+	Dependency InferenceStepDependencyType `json:"dependency,omitempty"`
```
yuzisun, line 274: What’s the default value ?

python/test_resources/graph/error_404_isvc/requirements.txt
```diff
@@ -0,0 +1,2 @@
+fastapi
+kserve==0.10.1
```
yuzisun: Let’s use 0.11

test/e2e/graph/test-resources/ig_test_ensemble_scenario_7.yaml
```diff
@@ -0,0 +1,14 @@
+# Since both the steps are sift deps, so output will be combined output of both steps
```
yuzisun:
```suggestion
# Since both the steps are soft deps, so output will be combined output of both steps
```

test/e2e/graph/test-resources/ig_test_ensemble_scenario_8.yaml
```diff
@@ -0,0 +1,17 @@
+# Since both the steps are sift deps, so output will be combined output of both steps
```
yuzisun: Need to update the comment here for the hard dep case

cmd/router/main.go
```diff
@@ -98,33 +100,72 @@ func timeTrack(start time.Time, nodeOrStep string, name string) {
...
+			log.Error(err, errorMessage)
+			/*
+				TODO Should it be 500 or 404 when none of the routes match.
+				For backward compatibility 500 is needed but 404 also looks fine since the router got bad input and didn't know what to do.
+				See what reviewers say.
+			*/
+			return nil, 500, errors.New(errorMessage)
 		}
-		return executeStep(route, graph, input, headers)
+		stepType := "serviceUrl"
```
yuzisun, line 124: What's the significance of stepType? is it mostly for logging and debugging purpose?

cmd/router/main.go
```diff
@@ -133,69 +174,120 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
 func graphHandler(w http.ResponseWriter, req *http.Request) {
 	inputBytes, _ := io.ReadAll(req.Body)
-	if response, err := routeStep(v1alpha1.GraphRootNodeName, *inferenceGraph, inputBytes, req.Header); err != nil {
+	if response, statusCode, err := routeStep(v1alpha1.GraphRootNodeName, *inferenceGraph, inputBytes, req.Header); err != nil {
 		log.Error(err, "failed to process request")
-		w.WriteHeader(500) //TODO status code tbd
-		w.Write([]byte(fmt.Sprintf("Failed to process request: %v", err)))
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(statusCode)
+		//w.WriteHeader(http.StatusInternalServerError)
```
yuzisun: Remove the unused line

cmd/router/main.go
```diff
@@ -133,69 +174,120 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
+		ensembleStepOutput := EnsembleStepOutput{}
 		for i, resultChan := range ensembleRes {
 			key := currentNode.Steps[i].StepName
 			if key == "" {
 				key = strconv.Itoa(i) // Use index if no step name
 			}
 			select {
-			case response[key] = <-resultChan:
+			case ensembleStepOutput = <-resultChan:
+				if ensembleStepOutput.StepStatusCode != 200 && currentNode.Steps[i].Dependency == v1alpha1.Hard {
```
yuzisun: Here we can use the `isSuccessful` util to be consistent with other places

cmd/router/main.go
rachitchauhan43: @yuzisun : I can do that. But 1 quick thing, is it the only to do debug logging with controller-runtime's log package ?
```
log.V(-1).Info("Exiting callService", "serviceUrl", serviceUrl)
```

rachitchauhan43: I have entry log at info so will keep this at info for now. Or i can remove it if you see any issues with it.
Another thing is , I will fix the logging by moving uber-zap logger that is more flexible in a separate PR.

cmd/router/main.go
rachitchauhan43: Done

cmd/router/main.go
rachitchauhan43: DONE

cmd/router/main.go
rachitchauhan43: Done

cmd/router/main_test.go
rachitchauhan43: Not needed. Was added for debugging. Removing it.

cmd/router/main.go
rachitchauhan43: Done

python/test_resources/graph/error_404_isvc/requirements.txt
rachitchauhan43: Removing requirements.txt as it is not needed

cmd/router/main.go
rachitchauhan43: Yes

cmd/router/main.go
yuzisun: we can update to a more useful logging with the time spent on the service call

yuzisun: actually we already have `defer timeTracker` to log the info so we can remove this log line

cmd/router/utils.go
```diff
@@ -0,0 +1,12 @@
+package main
```
yuzisun: nit on the filename `router_error.go` ?

rachitchauhan43: Done

yuzisun:
@rachitchauhan43 There seems to be resource issues possibly after changing test parallelism to 2? 

```
isvc-sklearn-runtime-predictor-00001" failed with message: 0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod...
```

https://github.com/kserve/kserve/actions/runs/6097857892/job/16547655921?pr=3039

rachitchauhan43:
> @rachitchauhan43 There seems to be resource issues possibly after changing test parallelism to 2?
> 
> ```
> isvc-sklearn-runtime-predictor-00001" failed with message: 0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod...
> ```
> 
> https://github.com/kserve/kserve/actions/runs/6097857892/job/16547655921?pr=3039

Ok. Removed parallelism and bumping timeout to 120 mins to avoid job timeout issue.

yuzisun:
Awesome work @rachitchauhan43 !

/lgtm

yuzisun:
@rachitchauhan43 Created a doc issue for you https://github.com/kserve/website/issues/289

Issue #3041, opened 2023-07-21
Make images usable in OpenShift


rhuss:
/kind feature

**Describe the solution you'd like**

In OpenShift, containers are run under a randomly selected user ID. The current images do not play nicely with this; they expect to run with a fixed user id of 1000. I.e., the file access modes cannot deal with this, so the images need to be recreated for OpenShift. See https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html for more information about the criteria for running container images on OpenShift.

I.e., from https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images

>  By default, OpenShift Container Platform runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability, thereby achieving escalated permissions on the host node.
> 
> For an image to support running as an arbitrary user, directories and files written to by processes in the image must be owned by the root group and read/writable by that group. Files to be executed must also have group execute permissions.
> 
> Adding the following to your Dockerfile sets the directory and file permissions to allow users in the root group to access them in the built image:

```
RUN chgrp -R 0 /some/directory && \
    chmod -R g=u /some/directory
``` 

> Because the container user is always a member of the root group, the container user can read and write these files.

**Anything else you would like to add:**

Adapting the Dockerfiles accordingly is not harmful and fully compatible with the current usage when the container is running with the UID 1000

PR #3038, opened 2023-07-19
Add ARM64 arch support for serving runtimes

sivanantha321:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2989

**Type of changes**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
Added ARM64 arch support for serving runtimes
```

PR #3042, opened 2023-07-21
sklearnserver Docker image: Enable running under arbitrary UID to enable OpenShift

rhuss:

Set file's group to the root group and give the group the same permissions as the users. This is the recommendation of OpenShift (https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html#use-uid_create-images) to run under arbitrary user ids (but with those users also being member of the "root" group).

This change does is non-intrusive and will work on any platform.

Related to #3041, could be expanded to all KServe images, like in #3038 for a different purpose

**Type of changes**

Run `docker run -it --entrypoint bash rhuss/sklearnserver` and then

```
kserve@ed9703137c27:/$ id
uid=1000(kserve) gid=1000(kserve) groups=1000(kserve)
kserve@ed9703137c27:/$ cd /

kserve@ed9703137c27:/$ ls -ld kserve/ sklearnserver/ prod_venv/
drwxrwxr-x 1 kserve root 4096 Jul 18 06:00 kserve/
drwxrwxr-x 1 kserve root 4096 Jul 21 13:24 prod_venv/
drwxrwxr-x 1 kserve root 4096 Jul 21 12:59 sklearnserver/

kserve@ed9703137c27:/$ ls -l kserve
total 280
-rw-rw-r-- 1 kserve root    273 Jul 18 06:00 Makefile
-rw-rw-r-- 1 kserve root     36 Jul 18 06:00 OWNERS
-rw-rw-r-- 1 kserve root   6521 Jul 18 06:00 README.md
drwxrwxr-x 1 kserve root   4096 Jul 18 06:00 docs
drwxrwxr-x 1 kserve root   4096 Jul 18 06:00 kserve
-rw-rw-r-- 1 kserve root 246881 Jul 18 06:00 poetry.lock
-rw-rw-r-- 1 kserve root   2855 Jul 18 06:00 pyproject.toml
-rw-rw-r-- 1 kserve root     28 Jul 18 06:00 setup.cfg
drwxrwxr-x 1 kserve root   4096 Jul 18 06:00 test
```

**Checklist**:

**Release note**:

```release-note
Enable images to run in OpenShift
```

### DIFF
--- a/python/sklearn.Dockerfile
+++ b/python/sklearn.Dockerfile
@@ -48,7 +48,7 @@ COPY --from=builder --chown=1000:0 kserve kserve/
 COPY --from=builder --chown=1000:0 sklearnserver sklearnserver/
 
 # Give users in the root group the same permissions as the users
-# to allow running the image wit arbitrary UIDs
+# to allow running the image with arbitrary UIDs
 RUN chmod -R g=u "$VIRTUAL_ENV" kserve sklearnserver
 
 USER 1000
--- a/python/sklearn.Dockerfile
+++ b/python/sklearn.Dockerfile
@@ -42,9 +42,14 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN useradd kserve -m -u 1000 -d /home/kserve
 
-COPY --from=builder --chown=kserve:kserve $VIRTUAL_ENV $VIRTUAL_ENV
-COPY --from=builder kserve kserve
-COPY --from=builder sklearnserver sklearnserver
+# Make all files also owned by the root group
+COPY --from=builder --chown=1000:0 $VIRTUAL_ENV $VIRTUAL_ENV
+COPY --from=builder --chown=1000:0 kserve kserve/
+COPY --from=builder --chown=1000:0 sklearnserver sklearnserver/
+
+# Give users in the root group the same permissions as the users
+# to allow running the image wit arbitrary UIDs
+RUN chmod -R g=u "$VIRTUAL_ENV" kserve sklearnserver
 
 USER 1000
 ENTRYPOINT ["python", "-m", "sklearnserver"]


python/sklearn.Dockerfile
```diff
@@ -39,9 +39,14 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN useradd kserve -m -u 1000 -d /home/kserve
```
ckadner, line 40: do we also have to add the `kserve` user to group `0`?

```Dockerfile
RUN useradd kserve -m -u 1000 -g 0 -d /home/kserve
```

python/sklearn.Dockerfile
```diff
@@ -39,9 +39,14 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
...
 
-COPY --from=builder --chown=kserve:kserve $VIRTUAL_ENV $VIRTUAL_ENV
-COPY --from=builder kserve kserve
-COPY --from=builder sklearnserver sklearnserver
+# Make all files also owned by the root group
```
ckadner, line 45: do we also have to add the `kserve` user to group `0` above?

```Dockerfile
RUN useradd kserve -m -u 1000 -g 0 -d /home/kserve
```

python/sklearn.Dockerfile
```diff
@@ -39,9 +39,14 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
...
+COPY --from=builder --chown=1000:0 $VIRTUAL_ENV $VIRTUAL_ENV
+COPY --from=builder --chown=1000:0 kserve kserve/
+COPY --from=builder --chown=1000:0 sklearnserver sklearnserver/
+
+# Give users in the root group the same permissions as the users
+# to allow running the image wit arbitrary UIDs
```
ckadner: nit: "with" instead of "wit"

ckadner (changes requested):

python/sklearn.Dockerfile
rhuss: I don't think so. The group association is not relevant when running under the user "kserve" since, for this, all the user-permission flags are the relevant ones (as before).

For OpenShift, when the process is _not_ running as "kserve" but with an arbitrary UID that belongs to the root group, the group-permission flags are relevant. 

In the Dockerfile, we set the group-permission flags to the same values as the user-permission flags (that `chmod -R g=u`), so both cases (running under "kserve" and running under any user belonging to group 0) are equivalent.

python/sklearn.Dockerfile
rhuss:
```suggestion
# to allow running the image with arbitrary UIDs
```

PR #3060, opened 2023-08-01
Introduce Storage container CRD

### LINKED ISSUE #2727: Question: Custom storage initializer container
/kind feature

A question rather than a feature request. My team uses a dedicated model registry with an S3 storage backend. it comes with versioning and downloading utilities. We would like our storage initializer container to 
- check if the model required in the manifest is present in the locally mounted pod volume
  - if yes: great!
  - if not: download it from the model registry and put it there

Then the serving container would start up and access the artifact in the mounted pod volume much like the current off the shelve runtimes.

My question is: how would i best implement that?

My current approach would be to use the custom predictor and its pod spec to define all of the above resources much like with any other K8s deployment manifest. I.e. list the custom storage initializer image as one of the init containers etc.

Is there a better way, maybe using custom serving runtimes or something? and is there a place in the repo where i would see the manifests (i.e. the runtimes that you are providing out of the box, like sklearn etc) that are implementing the current storage initializer behaviour (the one that simply downloads the file specified in the `storageUri` field of the off the shelve predictor manifest)?

### LINKED ISSUE #2885: Enable overriding the storage initializer image for specific protocol(s)
/kind feature

Currently KServe only supports [a single storage initializer configuration](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L18). We would like administrators to be able to specify which container images handle which schemes/protocols.  This would enable cluster administrators to add support for new, non-standard, or unsupported schemes without needing to override the default image.

We can update the config to look as follows:

```yaml
  storageInitializer: |-
    {
        "memoryRequest": "100Mi",
        "memoryLimit": "1Gi",
        "cpuRequest": "100m",
        "cpuLimit": "1",
        "storageSpecSecretName": "storage-config",
        "enableDirectPvcVolumeMount": false,
        "defaultImage": "kserve/storage-initializer:latest"
        "imageOverrides": [
            {
               "image": "example/ftp-storage-initializer:latest",
               "schemes": ["ftp"]
            }
        ]
    }
``` 

The pod injector will inspect the `storageUri` and set the storage initializer image based on the URI scheme. If no override is found for the detected scheme the `defaultImage` will be used.

**Limitations**

1. This implementation would not allow administrators to provide per-image configuration like requests/limits.
2. The storage initializer currently inspects the full URI to determine which strategy to use for downloading. This implementation only allows images to be selected based on scheme, not the entire URI.

greenmoon55:

Currently KServe comes with a [default storage initializer image](https://github.com/kserve/kserve/blob/61c9bd334ae9766ffd1f2bf020764bf453cab54c/python/kserve/kserve/storage/storage.py#L59) for downloading models. It supports common storage protocols like S3, GCS, Azure. The cluster admin can specify a custom storage initializer image [in the config map](https://github.com/kserve/kserve/blob/61c9bd334ae9766ffd1f2bf020764bf453cab54c/config/configmap/inferenceservice.yaml#L57) if they want to support a custom storage protocol, but they cannot specify different storage initializer images for different protocols.

We would like to introduce a storage container CRD to allow developers to use different storage initializer images based on the storage uri in the InferenceService spec.

If there is a match based on storageUri, container spec from Storage container CR will be used. Otherwise we keep the current behavior, i.e. use image/resource defined in the configMap.

Please see the [design doc](https://docs.google.com/document/d/1UcHFCQcTcDBcZVd8MTEgSXXKr7GMpzHpjuJW8gEwyFU/edit?usp=sharing
) for more details.

Fixes #2885, fixes #2727

**Type of changes**

Added two unit test files.
Existing E2E tests pass.
Manually tested 2 scenarios.

1. Apply the default CRD, make sure that demo sklearn isvc is ready after downloading the models.
2. Modify the default CRD to use an invalid image. `k describe pod -n kserve-test sklearn-iris-predictor-00001-deployment-86469dc9ff-kz5z7` shows the invalid image is being used.

- Logs

**Checklist**:

Will add documentation in a separate PR.

**Release note**:

```release-note
Developers can install a Storage container CR to use custom storage initializer images based on storage uri.

```


greenmoon55:
~~hmmm, not sure why "files changed" tab is showing more changes than I actually did. I'll try to squash my commits locally.~~
Done

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the stoarge initializer init container, and the protocols it supports.
+// +k8s:openapi-gen=true
+type StorageContainerSpec struct {
+	// +optional
+	Disabled *bool `json:"disabled,omitempty"`
+
+	StorageContainer corev1.Container `json:"storageContainer" validate:"required"`
```
greenmoon55: Is it better to rename it from `StorageContainer` to `Container`?

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the stoarge initializer init container, and the protocols it supports.
+// +k8s:openapi-gen=true
+type StorageContainerSpec struct {
+	// +optional
+	Disabled *bool `json:"disabled,omitempty"`
```
yuzisun: What does disable mean, fallback to the configmap?

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+	StorageContainer corev1.Container `json:"storageContainer" validate:"required"`
+
+	SupportedPrefixes []string `json:"supportedPrefixes,omitempty"`
+	SupportedRegexes  []string `json:"supportedRegexes,omitempty"`
```
yuzisun: not sure if these two fields fit here as the struct is named `StorageContainerSpec`

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"regexp"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the stoarge initializer init container, and the protocols it supports.
```
yuzisun:
```suggestion
// StorageContainerSpec defines the container spec for the storage initializer init container, and the protocols it supports.
```

pkg/apis/serving/v1alpha1/storage_container_types.go
yuzisun: prefer `container` as the CR name already codifies storage.

pkg/apis/serving/v1alpha1/storage_container_types.go
yuzisun: better to be consistent with how serving runtime does this with `SupportedModelFormats`

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -313,6 +333,8 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		); err != nil {
 			return err
 		}
+		// initContainer.Args[0] is set up in CreateStorageSpecSecretEnvs
+		srcURI = initContainer.Args[0]
```
yuzisun, line 343: this only applies to storageSpec ?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -340,6 +373,36 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 	return nil
 }
 
+// Use values from crd container spec from most fields. Use default values from initContainer for some fields.
+func mergeContainerSpecs(initContainer *v1.Container, crdSpec *v1.Container) (*v1.Container, error) {
```
yuzisun: can we do strategy merge with the two container spec? I'd expect we do not need to explicitly enumerate all the fields and override here

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -340,6 +373,36 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+		newSpec.Resources.Requests[v1.ResourceMemory] = *initContainer.Resources.Requests.Memory()
+	}
+	if newSpec.Resources.Limits == nil {
+		newSpec.Resources.Limits = make(v1.ResourceList)
+	}
+	if newSpec.Resources.Limits.Cpu().IsZero() {
+		newSpec.Resources.Limits[v1.ResourceCPU] = *initContainer.Resources.Limits.Cpu()
+	}
+	if newSpec.Resources.Limits.Memory().IsZero() {
+		newSpec.Resources.Limits[v1.ResourceMemory] = *initContainer.Resources.Limits.Memory()
```
yuzisun: We should respect the values from crd container spec

config/rbac/role.yaml
```diff
@@ -208,6 +208,27 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
yuzisun, line 211: Also need this in the helm chart which is why helm test is failing I guess

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: Yeah, the src URI is generated in `CreateStorageSpecSecretEnvs` for storageSpec, otherwise the storage uri is provided in the isvc spec.

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: I believe this is like the strategy merge, here we set some fields from default values if they are empty in the CRD.

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: Yeah, the `if` here checks if the values are set in the crd or not, if not we use default values. This could be a validation on the CRD itself.

pkg/apis/serving/v1alpha1/storage_container_types.go
greenmoon55: Here we have two options.
1. Current implementation
```
SupportedPrefixes:
  "https"
  "ftp"
SupportedRegexes:
  "a*b"
  "b*a"
```
2. Like `SupportedModelFormats`
```
SupportedURIFormats:
  - name: s3
    type: prefix
    str: "s3://"
  - name: azure
    type: regex
    str: "a*b"
```
I don't have a strong opinion but I prefer option 1 because it is simpler. I'm okay with option 2 to keep it consistent with `SupportedModelFormats`.

For your first comment, how about renaming this spec from `StorageContainerSpec` to `StorageInitializerSpec`?

pkg/apis/serving/v1alpha1/storage_container_types.go
greenmoon55: If there isn't another CR that supports this url, then we will fallback to the configmap. I think it is the same as serving runtime.

config/rbac/role.yaml
greenmoon55: Let me check it!

charts/kserve-resources/templates/clusterstoragecontainer.yaml
```diff
@@ -0,0 +1,26 @@
...
+      limits:
+        memory: 1Gi
+        cpu: "1"
+  supportedPrefixes:
+    - 'gs://'
+    - 's3://'
+    - 'hdfs://'
+    - 'webhdfs://'
+    - 'pvc://'
+    - 'file://'
```
yuzisun: this is not supported, only for local testing not for cloud environment

pkg/apis/serving/v1alpha1/storage_container_types.go
yuzisun: I am ok with `StorageContainerSpec` as it is not necessarily need to be an init container.  I like the `SupportedURIFormats` idea which can be either prefix or regex.

charts/kserve-resources/templates/clusterstoragecontainer.yaml
greenmoon55: Do you mean "file://" is not supported?

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: Updated it to use strategy merge.

pkg/webhook/admission/pod/suite_test.go
```diff
@@ -32,7 +34,12 @@ var c client.Client
 
 func TestMain(m *testing.M) {
 	t := pkgtest.SetupEnvTest()
-	var err error
+
+	err := v1alpha1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		klog.Error(err, "Failed to add v1alpha1 to scheme")
+	}
```
greenmoon55, line 41: I'll create a separate PR to move these lines inside `pkgtest.SetupEnvTest()`

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -80,6 +83,28 @@ func getStorageInitializerConfigs(configMap *v1.ConfigMap) (*StorageInitializerC
...
 }
 
+func getContainerSpecForStorageUri(storageUri string, client client.Client) (*v1.Container, error) {
+	storageContainers := &v1alpha1.ClusterStorageContainerList{}
+	if err := client.List(context.TODO(), storageContainers); err != nil {
+		return nil, err
+	}
+
+	for _, sc := range storageContainers.Items {
+		if sc.IsDisabled() {
```
greenmoon55, line 93: skip checking if there is a match if disabled.
We can introduce a priority field in the future.

pkg/apis/serving/v1alpha1/storage_container_types_test.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2023 The KServe Authors.
```

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -326,6 +354,18 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		}
 	}
 
+	// Update container spec from a storage container CR if there is a match
+	storageContainerSpec, err := getContainerSpecForStorageUri(srcURI, mi.client)
+	if err != nil {
+		return err
+	}
+	if storageContainerSpec != nil {
```
yuzisun, line 364: What happens if `storageContainerSpec` is nil and nothing is matched?

yuzisun:
@greenmoon55 Seems like the commit is not signed

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: I'm thinking we can use config map for now, which can be removed in future versions. I added a comment here.

greenmoon55:
> @greenmoon55 Seems like the commit is not signed

Updated yesterday!

yuzisun:
Awesome work! Thanks @greenmoon55!

/lgtm
/approve

PR #3069, opened 2023-08-07
document status conditions for RoutesReady and LatestDeploymentReady

tessapham:

Documents new conditions for `InferenceServiceStatus`.


python/kserve/requirements.txt
```diff
@@ -0,0 +1,6 @@
+certifi >= 14.05.14
+future; python_version<="2.7"
+six >= 1.10
+python_dateutil >= 2.5.3
+setuptools >= 21.0.0
+urllib3 >= 1.15.1
```
tessapham: Not sure if this file and `python/kserve/setup.py` should be included as they weren't in the repo before.

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -34,7 +34,8 @@ type InferenceServiceStatus struct {
 	// - PredictorReady: predictor readiness condition; <br/>
 	// - TransformerReady: transformer readiness condition; <br/>
 	// - ExplainerReady: explainer readiness condition; <br/>
-	// - RoutesReady: aggregated routing condition; <br/>
+	// - RoutesReady: aggregated routing condition, i.e. endpoint readiness condition; <br/>
+	// - LatestDeploymentReady: aggregated configuration condition, i.e. latest deployment readiness condition; <br/>
```
yuzisun: Need to add a note that these two conditions are only applicable for serverless mode.

yuzisun: Actually the doc is not generated on CRD, we need a PR to the website to generate the [API doc](https://kserve.github.io/website/0.10/reference/api/) using the [script](https://github.com/kserve/website/blob/main/docs/hack/gen-api-reference-docs.sh).

go.sum
```diff
@@ -17,79 +17,235 @@ cloud.google.com/go v0.62.0/go.mod h1:jmCYTdRCQuc1PHIIJ/maLInMho30T/Y0M4hTdTShOY
 cloud.google.com/go v0.65.0/go.mod h1:O5N8zS7uWy9vkA9vayVHs65eM1ubvY4h553ofrNHObY=
 cloud.google.com/go v0.110.2 h1:sdFPBr6xG9/wkBbfhmUz/JmZC7X6LavQgcrVINrKiVA=
 cloud.google.com/go v0.110.2/go.mod h1:k04UEeEtb6ZBRTv3dZz4CeJC3jKGxyhl0sAiVVquxiw=
+cloud.google.com/go/accessapproval v1.6.0/go.mod h1:R0EiYnwV5fsRFiKZkPHr6mwyk2wxUJ30nL4j2pcFY2E=
```
yuzisun: Are these expected changes?

tessapham: These are the changes rendered after I ran `make manifests` and `make generate`.

yuzisun: hmm just tried out, that did not happen to me

tessapham: I reran `go mod tidy` and these changes went away.

tessapham: These changes were actually generated after I ran `make manifests`, and then running `go mod tidy` after than cleaned them up.

PR #3086, opened 2023-08-18
Allow disabling creation of the HPA in raw deployment mode

### LINKED ISSUE #2131: Allow disabling creation of the HPA in raw deployment mode
/kind feature

Allow disabling creation of the HPA in raw deployment mode.

I would like to be able to control the scaling of the deployment created by my KServe inference service with an external tool ([KEDA](https://keda.sh/)). In order to support this use case I need a way to disable KServe's creation of the HPA. I was thinking adding a new autoscaler class called `none` which disables the creation of any resources by the autoscaler reconciler.

andyi2it:

Fixes #2131 

**Type of changes**

**Checklist**:

```bash
cat <<EOF| kubectl apply -f -
kind: "InferenceService"
metadata:
  annotations:
    "serving.kserve.io/autoscalerClass": "external"
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: "sklearn-irisv2"
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      runtime: kserve-mlserver
      storageUri: "gs://seldon-models/sklearn/mms/lr_model"
EOF
```

**Release note**:

```release-note
Disable creating HPA component with the `serving.kserve.io/autoscalerClass: external` annotation.
```

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -139,7 +139,7 @@ func validateInferenceServiceAutoscaler(isvc *InferenceService) error {
 					} else {
 						return nil
 					}
-				case constants.AutoscalerClassNone:
+				case constants.AutoscalerClassExternal:
 					return nil
 				default:
 					return fmt.Errorf("unknown autoscaler class [%s]", class)
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -139,6 +139,8 @@ func validateInferenceServiceAutoscaler(isvc *InferenceService) error {
 					} else {
 						return nil
 					}
+				case constants.AutoscalerClassNone:
+					return nil
 				default:
 					return fmt.Errorf("unknown autoscaler class [%s]", class)
 				}
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -154,8 +154,8 @@ var (
 
 // Autoscaler Class
 var (
-	AutoscalerClassHPA  AutoscalerClassType = "hpa"
-	AutoscalerClassNone AutoscalerClassType = "none"
+	AutoscalerClassHPA      AutoscalerClassType = "hpa"
+	AutoscalerClassExternal AutoscalerClassType = "external"
 )
 
 // Autoscaler Metrics
@@ -171,7 +171,7 @@ var (
 // Autoscaler Class Allowed List
 var AutoscalerAllowedClassList = []AutoscalerClassType{
 	AutoscalerClassHPA,
-	AutoscalerClassNone,
+	AutoscalerClassExternal,
 }
 
 // Autoscaler Metrics Allowed List
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -154,7 +154,8 @@ var (
 
 // Autoscaler Class
 var (
-	AutoscalerClassHPA AutoscalerClassType = "hpa"
+	AutoscalerClassHPA  AutoscalerClassType = "hpa"
+	AutoscalerClassNone AutoscalerClassType = "none"
 )
 
 // Autoscaler Metrics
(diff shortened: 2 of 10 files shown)


yuzisun:
@andyi2it Actually if we disable the HPA,  how can user modify the replica of the isvc?

andyi2it:
@yuzisun 
The issue pertains to the use case where they use external load balancers like KEDA. In such case they can provide autoscaler class as None and manage the scaling there.

yuzisun:
> @yuzisun The issue pertains to the use case where they use external load balancers like KEDA. In such case they can provide autoscaler class as None and manage the scaling there.

Wondering if we should have a more native integration with KEDA, people may think that once set to none they can control the replica but in fact there is no way to set the replica number

lizzzcai:
> > @yuzisun The issue pertains to the use case where they use external load balancers like KEDA. In such case they can provide autoscaler class as None and manage the scaling there.
> 
> Wondering if we should have a more native integration with KEDA, people may think that once set to none they can control the replica but in fact there is no way to set the replica number

Didn't notice that when autoscaler class is `None`, the user can not set the replicas here. Will be good if can let user set it like using the `MinReplicas` or set both Min Max to the same value. 

For native integration with KEDA, it will be good to see it in the future in raw deployment mode as KEDA supports [HTTP](https://github.com/kedacore/http-add-on) as well.

yuzisun:
> > > @yuzisun The issue pertains to the use case where they use external load balancers like KEDA. In such case they can provide autoscaler class as None and manage the scaling there.
> > 
> > 
> > Wondering if we should have a more native integration with KEDA, people may think that once set to none they can control the replica but in fact there is no way to set the replica number
> 
> Didn't notice that when autoscaler class is `None`, the user can not set the replicas here. Will be good if can let user set it like using the `MinReplicas` or set both Min Max to the same value.
> 
> For native integration with KEDA, it will be good to see it in the future in raw deployment mode as KEDA supports [HTTP](https://github.com/kedacore/http-add-on) as well.


@lizzzcai @andyi2it Currently [deployment replica field is ignored ](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go#L100)as it is not allowed setting it when using HPA, the question is that there is no way for kserve to know if people are using external autoscaler like KEDA or not and when it is expected to set the replica field.

lizzzcai:
> @lizzzcai @andyi2it Currently [deployment replica field is ignored ](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go#L100)as it is not allowed setting it when using HPA, the question is that there is no way for kserve to know if people are using external autoscaler like KEDA or not and when it is expected to set the replica field.

I am not so familiar with raw deployment here, not sure if we can use the new annotation `"serving.kserve.io/autoscalerClass": "none"` as an indicator. If it is `none`, then the controller will not ignore the replicas.    However, this can be refined in a new issue as well if this PR is good enough to merge.

yuzisun:
> > @lizzzcai @andyi2it Currently [deployment replica field is ignored ](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go#L100)as it is not allowed setting it when using HPA, the question is that there is no way for kserve to know if people are using external autoscaler like KEDA or not and when it is expected to set the replica field.
> 
> I am not so familiar with raw deployment here, not sure if we can use the new annotation `"serving.kserve.io/autoscalerClass": "none"` as an indicator. If it is `none`, then the controller will not ignore the replicas. However, this can be refined in a new issue as well if this PR is good enough to merge.


I think none is confusing as kserve can’t decide if people are using it without autoscaler or using external autoscaler. So I think we need another option for “external” for keda and in this case the replica is managed by Keda.

rehevkor5:
Thanks for adding this. I'm interested in using it because KServe doesn't currently provide control over everything that K8s offers in the HPA spec, for example we can't use anything underneath [the `behavior` field](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior).

PR #3088, opened 2023-08-18
Storage initializer fix so that it downloads only specific file when provided uri is not a folder

### LINKED ISSUE #2473: S3 storage initializer doesn't work as expected for file without extension

andyi2it:

Fixes #2473 

**Type of changes**

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -191,11 +191,13 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             # If 'uri' is set to "s3://test-bucket/a/b/c/model.bin", then
             # the downloader will add to temp dir: model.bin
             # (without any subpaths).
-            target_key = (
-                obj.key.rsplit("/", 1)[-1]
-                if bucket_path == obj.key
-                else obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
-            )
+            if bucket_path == obj.key:
+                target_key = obj.key.rsplit("/", 1)[-1]
+            elif bucket_path.rsplit("/", 1)[-1] in obj.key.rsplit("/", 1)[-1]:
+                target_key = obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
+            else:
+                target_key = obj.key.replace(bucket_path, "").lstrip("/")
+
             target = f"{temp_dir}/{target_key}"
             if not os.path.exists(os.path.dirname(target)):
                 os.makedirs(os.path.dirname(target), exist_ok=True)
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -172,6 +172,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
         bucket_path = parsed.path.lstrip('/')
 
         count = 0
+        exact_obj_found = False
         bucket = s3.Bucket(bucket_name)
         for obj in bucket.objects.filter(Prefix=bucket_path):
             # Skip where boto3 lists the directory as an object
@@ -193,6 +194,7 @@ class Storage(object):  # pylint: disable=too-few-public-methods
             # (without any subpaths).
             if bucket_path == obj.key:
                 target_key = obj.key.rsplit("/", 1)[-1]
+                exact_obj_found = True
             elif obj.key.rsplit("/", 1)[-1].startswith(bucket_path.rsplit("/", 1)[-1]):
                 target_key = obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
             else:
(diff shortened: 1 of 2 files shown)


python/kserve/kserve/storage/storage.py
```diff
@@ -191,17 +192,24 @@ def _download_s3(uri, temp_dir: str):
...
-            target_key = (
-                obj.key.rsplit("/", 1)[-1]
-                if bucket_path == obj.key
-                else obj.key.replace(bucket_path, "", 1).lstrip("/")
-            )
+            if bucket_path == obj.key:
+                target_key = obj.key.rsplit("/", 1)[-1]
+                exact_obj_found = True
+            elif obj.key.rsplit("/", 1)[-1].startswith(bucket_path.rsplit("/", 1)[-1]):
+                target_key = obj.key.replace(bucket_path.rsplit("/", 1)[0], "", 1).lstrip("/")
```
yuzisun: @andyi2it Is this correct?  if `obj.key` is `churn-pickle-logs` and `bucket_path` is `churn-pickle`, then after replace it becomes `-logs`

yuzisun: Let's also add comment as the code is a bit convoluted here.

yuzisun: I think we need to define the expectation here when bucket is `s3://churn-pickle` and obj is `churn-pickle-logs/model.log`, do we still want to download all the files when there is a prefix matching or we do not match at all.

yuzisun: @lizzzcai Any thoughts?

lizzzcai: I feel like matching on a folder or exact file name rather than `prefix` will make things easier. But I am not sure if it will break any `existing user experience` here.


for bucket with following files:
```
test/churn-pickle
test/churn-pickle-logs
test/churn-pickle-report
```
1. When I download `s3://<bucket>/test/churn-pickle` to `/mnt/models`:
only one file is downloaded
```
/mnt/models/churn-pickle
```

2. when I download `s3://<bucket>/test` to `/mnt/models`:
All files under test will be downloaded.
```
/mnt/models/churn-pickle
/mnt/models/churn-pickle-logs
/mnt/models/churn-pickle-report
```

3. When I download `s3://<bucket>/test/churn` to `/mnt/models`:
nothing will be downloaded
```

```

not sure of any corner case here, and for cloud bucket storage, there is no easy way to know if a path is a file or folder.


@yuzisun @andyi2it any comments?

andyi2it: @yuzisun @lizzzcai 
Have refactored code to make it more readable.
With this PR, we won't download using prefixes only in case where an exact match is found.
So, for 3rd scenario you have, it should download
```
/mnt/models/churn-pickle
/mnt/models/churn-pickle-logs
/mnt/models/churn-pickle-report
```

This change should affect only the 1st scenario you have. When there is an exact match found, it will not use that as a prefix. This wouldn't or shouldn't break anything as without this PR, it would download the prefix matched files as below, which would not be usable in any case.

```
/mnt/models/-logs
/mnt/models/-report
```

lizzzcai: Hi @andyi2it,  for 1st scenario, there is a difference from the behaviour in ModelMesh [here](https://github.com/kserve/kserve/issues/2473#issuecomment-1715517612). In modelmesh, only `/mnt/models/churn-pickle` will be downloaded.

PR #2812, opened 2023-04-11
unpack archive files for hdfs

I321506:
**What this PR does / why we need it**:
Currently, zip/tar file downloaded from the HDFS is not uncompressed. Adding the changes to uncompress the zip/tar files to be consistent with S3, OSS etc.

**Type of changes**
Please delete options that are not relevant.

 - [x] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] This change requires a documentation update

**Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):**
Fixes #2813 

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:
"NONE".
```release-note

```

PR #3093, opened 2023-08-21
Unpack archive files for hdfs

### LINKED ISSUE #2813: kserve does not unpack downloaded zip/tar files from hdfs
/kind feature

Currently, zip/tar file downloaded from the HDFS is not uncompressed. Adding the changes to uncompress the zip/tar files to be consistent with S3, OSS etc.
So to be consistent with S3, OSS etc we need to unpack archive file incase of HDFS.

sivanantha321:
Thanks for the PR @I321506. This is rebase of the PR #2812 

Fixes #2813

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -15,26 +15,26 @@
 import base64
 import glob
 import gzip
+import json
 import logging
 import mimetypes
 import os
 import re
-import json
 import shutil
 import tarfile
 import tempfile
-from typing import Dict
 import zipfile
-from urllib.parse import urlparse
-import requests
 from pathlib import Path
+from typing import Dict
+from urllib.parse import urlparse
+
+import boto3
+import requests
 from azure.storage.blob import BlobServiceClient
 from azure.storage.blob._list_blobs_helper import BlobPrefix
 from azure.storage.fileshare import ShareServiceClient
-
-from botocore.client import Config
 from botocore import UNSIGNED
-import boto3
+from botocore.client import Config
 from google.auth import exceptions
 from google.cloud import storage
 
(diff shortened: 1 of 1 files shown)


yuzisun:
@sivanantha321 Have you verified this? not sure if we have any test with hdfs

python/kserve/kserve/storage/storage.py
```diff
@@ -339,17 +339,29 @@ def _download_hdfs(uri, out_dir: str):
...
         if status["type"] == "FILE":
             client.download(path, out_dir, n_threads=1)
+            count = count + 1
+            file_name = path.rsplit("/", 1)[-1]
+            dest_file_path = f"{out_dir}/{file_name}"
         else:
             files = client.list(path)
             for f in files:
                 client.download(f"{path}/{f}", out_dir, n_threads=int(config["N_THREADS"]))
+                count = count + 1
```
markwinter: count can be simplified to `count = len(files)` instead of counting each loop here.

And perhaps `count` could be renamed to `file_count` to make it a bit more clear, but I realise other parts of the code use `count` to do this as well

markwinter:
@yuzisun 
> not sure if we have any test with hdfs

there aren't any, unfortunately 😥
It would be good to add at least one, but at the time I wasn't sure how to mock hdfs storage

sivanantha321:
> @sivanantha321 Have you verified this? not sure if we have any test with hdfs

Tested it in my local machine. Working as expected :)

python/kserve/kserve/storage/storage.py
sivanantha321: Thanks for the suggestion. It is done :).

PR #3104, opened 2023-08-30
Adding script for setting up Telepresence for local development

rhuss:

This PR helps set up [Telepresence](https://www.telepresence.io/) for allowing local testing and debugging of `cmd/manager/main.go` directly from the IDE. Telepresence takes care for setting up a VPN that allows bi-directional communication, i.e. so that the local process can act as a webhook for the remote cluster. This allows for very quick _inner loop_ development with short turnaround times as you don't need to create container images and redeployments.

Usage:

```
# Install everything. This command is idempotent and can
# be called multiple times
hack/telepresence-setup.sh

# Remove everything from telepresence
hack/telepresence-setup.sh uninstall
```
 
This script has been tested for a KServe installation on Minikube, running on macOS 13.5

More information for using Telepresence for developing:

 * https://codefresh.io/blog/telepresence-2-local-development/
 * https://ttt.io/kubernetes-admission-controller

### DIFF
--- a/hack/telepresence-setup.sh
+++ b/hack/telepresence-setup.sh
@@ -30,7 +30,8 @@ fi
 
 if ! type telepresence >/dev/null 2>&1; then
   echo "Telepresence is not installed."
-  echo "Please install it, e.g. with \"brew install telepresence\""
+  echo "Please install it, e.g. with \"brew install telepresence\" for Mac"
+  echo "or download it from https://www.getambassador.io/docs/telepresence/latest/install for all OS"
   exit 1
 fi
 
--- a/hack/telepresence-setup.sh
+++ b/hack/telepresence-setup.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+# Script for setting up telepresence to run the KServe controller
+# locally, fully connected to the cluster (it works also for the
+# incoming direction when used as a webhook)
+#
+# Usage:
+#        # Install everything. This command is idempotent and can
+#        # be called multiple times
+#        ./telepresence-setup.sh
+#
+#        # Remove everything from telepresence
+#        ./telepresence-setup.sh uninstall
+#
+# More information for using telepresence for developing:
+#
+# * https://codefresh.io/blog/telepresence-2-local-development/
+# * https://ttt.io/kubernetes-admission-controller
+
+set -o pipefail
+set +e
+
+if [[ "$1" == "uninstall" ]]; then
+  telepresence helm uninstall
+  kubectl delete ns ambassador
+  telepresence quit -s
+  [ -d "${TMPDIR}/k8s-webhook-server" ] && rm -rf "${TMPDIR}/k8s-webhook-server"
+  exit 0
+fi
+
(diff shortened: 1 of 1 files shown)


hack/telepresence-setup.sh
```diff
@@ -0,0 +1,80 @@
...
+  telepresence helm uninstall
+  kubectl delete ns ambassador
+  telepresence quit -s
+  [ -d "${TMPDIR}/k8s-webhook-server" ] && rm -rf "${TMPDIR}/k8s-webhook-server"
+  exit 0
+fi
+
+if ! type telepresence >/dev/null 2>&1; then
+  echo "Telepresence is not installed."
+  echo "Please install it, e.g. with \"brew install telepresence\""
```
matzew: perhaps pointing to here: https://www.getambassador.io/docs/telepresence/latest/install

Since `brew` is a Mac specific tool?

rhuss: Good point, will update. Thanks!

hack/telepresence-setup.sh
```diff
@@ -0,0 +1,81 @@
...
+  kubectl delete ns ambassador
+  telepresence quit -s
+  [ -d "${TMPDIR}/k8s-webhook-server" ] && rm -rf "${TMPDIR}/k8s-webhook-server"
+  exit 0
+fi
+
+if ! type telepresence >/dev/null 2>&1; then
+  echo "Telepresence is not installed."
+  echo "Please install it, e.g. with \"brew install telepresence\" for Mac"
+  echo "or download it from https://www.getambassador.io/docs/telepresence/latest/install for Linux"
```
matzew: and windows? :smile:

rhuss: ok, removed Linux :-)

rhuss: but I doubt that this script will work for Windows anyways ....

matzew: but `Telepresence` offers downloads for that platform :-)

terrytangyuan (commented):
This would be a great addition. There's another alternative https://github.com/txn2/kubefwd but it didn't work well for me. 

For this PR, it would be great if we could add some CI/test that uses this Telepresence setup. It would be great to update contributing guide as well.

rhuss:
@terrytangyuan, do you mean a script that runs the CI tests locally, with the local code, against a cluster tunneled by Telepresence? That sounds like a great idea, so developing CI tests might become faster with faster turnarounds.

In the meantime, I think merging this PR is an excellent first step and harmless, as it just adds some help when developing operators and especially webhooks locally that are called from within the cluster. My primary motivation is to allow debugging by setting breakpoints directly from the IDE and quick turnaround times without building container images (just recompiling within the IDE)

terrytangyuan:
Yep that's exactly the idea.

I'll leave this to reviewers/approvers to comment.

rhuss:
If there is any interest, I could record a demo with a development workflow that involves this script. Just ping me in the case and I will add it to this issue.

terrytangyuan (approved):
/lgtm

PR #3110, opened 2023-09-03
Add support for modelcars

### LINKED ISSUE #3043: Enhancing KServe Model Fetching with oci: Schema and Sidecar Containers
/kind feature

I propose an enhancement for model fetching in KServe using an `oci:` schema for model location. Currently, an init container fetches the model data from a location specified by a URL within the `InferenceService`. This process, however, has several limitations, especially for large models. This proposal aims to reduce startup latency, minimize local disk storage waste, and offer easier optimization methods by fetching the model directly from an OCI image stored in node-local image storage. 

The proposed model setup involves:
1. Adopting an `oci:` URL schema as a reference to the model in an `InferenceService`. 
2. Using a sidecar container, instead of an init-container, added to the Knative Service to hold the actual model. This allows the model to be fully contained and distributed within an OCI image.
3. Configuring the overall Pod as `shareProcessNamespace: true` to ensure all sidecars can access each other's filesystem via `/proc/<pid>/root`. This way, the main container can directly access the model from the data-container.

This approach could provide several benefits, including:
- Reduced local disk space usage as pods running the same model on a node will access the same directory containing the model.
- Lower startup times due to the elimination of copying potentially large data.
- Ability to leverage known techniques to pre-fetch container images to nodes or use lazy-loading techniques (e.g., with the [stargz snapshotter for containerd](https://github.com/containerd/stargz-snapshotter)).
- Resolution of certain problems encountered with init-containers when using the [init-container with Istio's CNI plugin](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers).

If the community agrees, I would be keen to work on implementing a Proof of Concept (PoC) to measure the benefits of this approach. I appreciate any feedback and look forward to our discussion.

rhuss:

This PR supports an "oci://" schema for the storage URL, which allows accessing models direct from OCI image containers without copying the model data from, somewhere.

The benefits of this approach compared to the traditional `storageInitializer` methods are:

* Only pull the model data once per node, not every time a Pod starts. This greatly benefits cold start times, especially when dealing with large models. See some initial numbers below.
* Distributing images (also large images; Windows images used to be 9GB and larger) is a well-known business that any decent OCI registry can handle.
* There are techniques to preload images to nodes to reduce cold starts to be close to zero.
* In the future, additional optimizations for image loading could be implemented, benefiting from the layered nature of images. I am not sure if the current models do support layering (e.g., for foundational and extended models), but hypothetically, such models could benefit from lazy loading 

Design document for "modelcars" can be found in this [document](https://docs.google.com/document/d/1Bs4fnP8rhPMaoPoLSYVvuRq-z9vkGPQ0rKbmfH4I7js/edit?usp=sharing)

Here is a very raw [modelcar demo](https://www.youtube.com/watch?v=KzWH8v6CcR0) (please forgive me some errors :), but it should give you a good initial impression of  modelcars and how they work:

Fixes #3043 

**Type of changes**

This initial PR implements these features:

* Add new storageUri schema "oci" that points to a OCI image
* New injector that adds a so-called "modelcar" container to "kserve-container" as a sidecar
* Setup pod for sharing the process namespace (shareProcessNamespace = true)
* Add new feature flag "enableOciImageSource" to switch on support for "oci" URLs

------------------

Some initial smoke tests indicate that performance can be improved by a magnitude of order for large models.

The test has been run on Minikube on a MacBook M1 with Docker and a 6 CPU, 16 GB VM. Times are hand-stopped and start counting before the initial creation of the inference service:

| Model Name       | SI or OCI | First Run | Second Run |
|------------------|-----------|-----------|------------|
| SKLearn Iris (500 Bytes)   | SI        | 31s        | 31s        |
| SKLearn Iris (500 Bytes)   | OCI       | 12s        | 21s        |
| Bloom-56 (3.2 Gb)  | SI        | 300s       | 300s       |
| Bloom-56 (3.2 Gb)  | OCI       | 305s       | 13s        |

As you can see, the second run is more or less the same regardless of the model size (which is to be expected as the only time needed is to create the symlink, which goes immediately). The first run includes the pull times of the container image, which is not much slower than downloading it from an S3 bucket.

_Please describe the tests you ran to verify your changes and the relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration._

A handful of unit tests have been added. To test the functionality, perform the following steps:

* Run `make deploy-dev` from a checked version of this branch

* Enable the model car feature by setting the corresponding feature flag to true, e.g. with 

```
kubectl get cm inferenceservice-config -n kserve -o json | \ 
sed 's/\\"enableOciImageSource\\": false/\\"enableOciImageSource\\": true/' | \
kubectl apply -f -
```

* Restart controller pod (not sure if needed, just to be sure)

* Create an `InferenceService` with

```
kind: "InferenceService"
metadata:
  name: "sklearn-iris"
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      storageUri: "oci://rhuss/sklearn-sample-model"
```  
Note the `oci://` schema of the `storageUri`.
The Dockerfile for this image looks like this:

```
FROM busybox
COPY models /models
```

With the local `models/` directory holding the model:

```
ls -l models
.rw-r--r-- roland staff 5.3 KB Wed Aug 30 23:35:21 2023 model.joblib
```

* When running, check that the link has been set up properly:

```
pod=$(kubectl get pods -o name -l serving.kserve.io/inferenceservice=sklearn-iris)

# In the kserve-container /mnt/models is a symlink into the modelcar container:
kubectl exec $pod -c kserve-container -- ls -ld /mnt/models
lrwxrwxrwx 1 root root 20 Sep  3 15:00 /mnt/models -> /proc/32/root/models

# All process are shared, as wells as their root filesystem:
# (pid 32 is the id of the modelcar process, so appears above as part of the symlink)
kubectl exec $pod -c modelcar -- ps
PID   USER     TIME  COMMAND
    1 65535     0:00 /pause
    7 root      0:07 python -m sklearnserver --model_name=sklearn-iris --model_dir=/mnt/models --http_port=8080
   13 65532     0:01 /ko-app/queue
   32 root      0:00 tail -f /dev/null
   93 root      0:00 ps
```

A quick (and raw) code walkthrough can be found in this [10 minute video](https://www.youtube.com/watch?v=axegGpQ6nHs)

The overall architecture for this feature looks like in this diagram:

Modelcar_2

* A user creates a container image that only contains the model data. The only prerequisite for the base image is that it contains an `ln` and `tail` command. In order to even make the base images smaller, one could consider to move that into a small golang binary that could be compiled statically.

* The main entry point is a [StorageInitializerInjector](StorageInitializerInjector) which runs for an `oci://` storageUri. 

* The important trick here is to [set the symbolic link as part of the startup command](https://github.com/rhuss/kserve/blob/00e699d79baafa9fab944aa806670100803bbdc3/pkg/webhook/admission/pod/storage_initializer_injector.go#L499) of the modelcar container.

* Resources of the modelcar are minimal (only so much that an idling tail can run). Currently, this is set to 10MB but can be even smaller. The resource consumption of the modelcar sidecar is negligible (in contrast to e.g. a fluent sidecar with a fuse layer).

There are still the following limitations or things to take care of:

* The "kserve-container" must run with [UID 0](https://github.com/rhuss/kserve/blob/00e699d79baafa9fab944aa806670100803bbdc3/pkg/webhook/admission/pod/storage_initializer_injector.go#L160) to be able to follow the symbolic link created by the modelcar. It's not yet clear whether this is a general limitation or specific to Minikube on which I have tested this PR
* Since user containers start in parallel (in contrast to init-containers that start in sequence), there could be a rare situation that the kserve-container starts before the modelcar container could set the symbolic link in the shared volume. For that storage.py has been [extended](https://github.com/rhuss/kserve/blob/00e699d79baafa9fab944aa806670100803bbdc3/python/kserve/kserve/storage/storage.py#L63) with a retry mechanism when checking for the model. To enable this check, the environment variable "MODEL_INIT_MODE" needs to be set to "async". This means that runtime themselves have to assume that the model directory is not immediately available but maybe only after some (short) time. This never has happened in the experiments that I did, but this is not an academic question. Alternatively, a supervisord could be added before starting third-party runtimes that could perform this async wait.

* No support for `StorageSpec` yet (not sure if this is needed)

* No documentation has been added, as I wanted first to agree on whether adding this feature makes sense. Also, some pointers on where to add the documentation would be great. 

**Checklist**:

**Release note**:

```release-note
* Add support for modelcar contains for speeding up initialization of runtimes. This feature is provided as alpha and needs to be enabled by setting the configuration flag `enableOciImageSource` to `true` for the storage initializer configuration.
```

### DIFF
--- a/charts/kserve-resources/templates/configmap.yaml
+++ b/charts/kserve-resources/templates/configmap.yaml
@@ -64,7 +64,10 @@ data:
            "memoryLimit": "1Gi",
            "cpuRequest": "100m",
            "cpuLimit": "1",
-           "enableDirectPvcVolumeMount": false
+           "enableDirectPvcVolumeMount": false,
+           "enableModelcar": false,
+           "cpuModelcar": "10m",
+           "memoryModelcar": "15Mi"
        }
      storageInitializer: |-
        {
@@ -86,7 +89,23 @@ data:
            # enableDirectPvcVolumeMount controls whether users can mount pvc volumes directly.
            # if pvc volume is provided in storageuri then the pvc volume is directly mounted to /mnt/models in the user container.
            # rather than symlink it to a shared volume. For more info see https://github.com/kserve/kserve/issues/2737
-           "enableDirectPvcVolumeMount": false
+           "enableDirectPvcVolumeMount": false,
+
+           # enableModelcar enabled allows you to directly access an OCI container image by
+           # using a source URL with an "oci://" schema.
+           "enableModelcar": false,
+
+           # cpuModelcar is the cpu request and limit that is used for the passive modelcar container. It can be
+           # set very low, but should be allowed by any Kubernetes LimitRange that might apply.
+           "cpuModelcar": "10m",
+
+           # cpuModelcar is the memory request and limit that is used for the passive modelcar container. It can be
+           # set very low, but should be allowed by any Kubernetes LimitRange that might apply.
+           "memoryModelcar": "15Mi",
+
+           # uidModelcar is the UID under with which the modelcar process and the main container is running.
+           # Some Kubernetes clusters might require this to be root (0). If not set the user id is left untouched (default)
+           "uidModelcar": 10
        }
 
      # ====================================== CREDENTIALS ======================================
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -43,20 +43,29 @@ const (
 	StorageInitializerContainerImage        = "kserve/storage-initializer"
 	StorageInitializerContainerImageVersion = "latest"
 	PvcURIPrefix                            = "pvc://"
+	OciURIPrefix                            = "oci://"
 	PvcSourceMountName                      = "kserve-pvc-source"
 	PvcSourceMountPath                      = "/mnt/pvc"
 	CaBundleVolumeName                      = "cabundle-cert"
+	ModelcarContainerName                   = "modelcar"
+	ModelInitModeEnv                        = "MODEL_INIT_MODE"
+	CpuModelcarDefault                      = "10m"
+	MemoryModelcarDefault                   = "15Mi"
 )
 
 type StorageInitializerConfig struct {
 	Image                      string `json:"image"`
 	CpuRequest                 string `json:"cpuRequest"`
 	CpuLimit                   string `json:"cpuLimit"`
+	CpuModelcar                string `json:"cpuModelcar"`
 	MemoryRequest              string `json:"memoryRequest"`
 	MemoryLimit                string `json:"memoryLimit"`
 	CaBundleConfigMapName      string `json:"caBundleConfigMapName"`
 	CaBundleVolumeMountPath    string `json:"caBundleVolumeMountPath"`
+	MemoryModelcar             string `json:"memoryModelcar"`
 	EnableDirectPvcVolumeMount bool   `json:"enableDirectPvcVolumeMount"`
+	EnableOciImageSource       bool   `json:"enableModelcar"`
+	UidModelcar                *int64 `json:"uidModelcar"`
 }
 
 type StorageInitializerInjector struct {
@@ -110,6 +119,70 @@ func GetContainerSpecForStorageUri(storageUri string, client client.Client) (*v1
 	return nil, nil
 }
 
+// InjectModelcar injects a sidecar with the full model included to the Pod.
+// This so called "modelcar" is then directly accessed from the user container
+// via the proc filesystem (possible when `shareProcessNamespace` is enabled in the Pod spec)
+func (mi *StorageInitializerInjector) InjectModelcar(pod *v1.Pod) error {
+	srcURI, ok := pod.ObjectMeta.Annotations[constants.StorageInitializerSourceUriInternalAnnotationKey]
+	if !ok {
+		return nil
+	}
+
+	// Only inject modelcar if requested
+	if !strings.HasPrefix(srcURI, OciURIPrefix) {
+		return nil
+	}
+
+	// Add an emptyDir Volume to Pod
+	pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
+		Name: StorageInitializerVolumeName,
+		VolumeSource: v1.VolumeSource{
+			EmptyDir: &v1.EmptyDirVolumeSource{},
+		},
+	})
+
+	// Extract image reference for modelcar from URI
+	image := strings.TrimPrefix(srcURI, OciURIPrefix)
+
+	userContainer := getContainerWithName(pod, constants.InferenceServiceContainerName)
+	if userContainer == nil {
+		return fmt.Errorf("no container found with name %s", constants.InferenceServiceContainerName)
+	}
+
+	// Indicate to the runtime that it the model directory could be
+	// available a bit later only so that it should wait and retry when
+	// starting up
+	addOrReplaceEnv(userContainer, ModelInitModeEnv, "async")
+
+	// Mount volume initialized by the modelcar container to the user container
+	userContainer.VolumeMounts = append(userContainer.VolumeMounts, v1.VolumeMount{
+		Name:      StorageInitializerVolumeName,
+		MountPath: getParentDirectory(constants.DefaultModelLocalMountPath),
+		ReadOnly:  false,
+	})
+
+	// If configured, run as the given user. There might be certain installations
+	// of Kubernetes where sharing the filesystem via the process namespace only works
+	// when both containers are running as root
+	if mi.config.UidModelcar != nil {
+		userContainer.SecurityContext = &v1.SecurityContext{
+			RunAsUser: mi.config.UidModelcar,
+		}
+	}
+
+	// Create the modelcar that is used as a sidecar in Pod and add it to the end
+	// of the containers
+	modelContainer := mi.createModelContainer(image, constants.DefaultModelLocalMountPath)
+	pod.Spec.Containers = append(pod.Spec.Containers, *modelContainer)
+
+	// Enable process namespace sharing so that the modelcar's root filesystem
+	// can be reached by the user container
+	shareProcessNamespace := true
+	pod.Spec.ShareProcessNamespace = &shareProcessNamespace
+
+	return nil
+}
+
 // InjectStorageInitializer injects an init container to provision model data
 // for the serving container in a unified way across storage tech by injecting
 // a provisioning INIT container. This is a work around because KNative does not
@@ -126,6 +199,11 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		return nil
 	}
 
+	// Don't inject init-containers if a modelcar is used
+	if mi.config.EnableOciImageSource && strings.HasPrefix(srcURI, OciURIPrefix) {
+		return nil
+	}
+
 	// Don't inject if InitContainer already injected
 	for _, container := range pod.Spec.InitContainers {
 		if strings.Compare(container.Name, StorageInitializerContainerName) == 0 {
@@ -134,16 +212,8 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 	}
 
 	// Find the kserve-container (this is the model inference server) and transformer container
-	var userContainer *v1.Container
-	var transformerContainer *v1.Container
-	for idx, container := range pod.Spec.Containers {
-		if strings.Compare(container.Name, constants.InferenceServiceContainerName) == 0 {
-			userContainer = &pod.Spec.Containers[idx]
-		}
-		if container.Name == constants.TransformerContainerName {
-			transformerContainer = &pod.Spec.Containers[idx]
-		}
-	}
+	userContainer := getContainerWithName(pod, constants.InferenceServiceContainerName)
+	transformerContainer := getContainerWithName(pod, constants.TransformerContainerName)
 
 	if userContainer == nil {
 		return fmt.Errorf("Invalid configuration: cannot find container: %s", constants.InferenceServiceContainerName)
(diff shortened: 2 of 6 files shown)


pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -382,6 +449,83 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+				Name:      StorageInitializerVolumeName,
+				MountPath: getParentDirectory(modelPath),
+				ReadOnly:  false,
+			},
+		},
+		Args: []string{
+			"sh",
+			"-c",
+			// $$$$ gets escaped by YAML to $$, which is the current PID
+			fmt.Sprintf("ln -s /proc/$$$$/root/models %s && tail -f /dev/null", modelPath),
```
zregvart: If we're splitting hairs `tail` will invoke `read` periodically (every second IIRC), this doesn't impact performance

```suggestion
			fmt.Sprintf("ln -s /proc/$$$$/root/models %s && sleep infinity", modelPath),
```

rhuss: The reason for  using `tail` over `sleep` was mostly portability, also to keep the requirements for the base image small. 

From https://stackoverflow.com/a/45396600:

> Portability
> This is an important and valid concern:
> sleep infinity is a GNU coreutils extension not contemplated in 
 POSIX. GNU's implementation also supports a "fancy" syntax for time durations, like sleep 1h 5.2s while POSIX only allows a positive integer (e.g. sleep 0.5 is not allowed).
> Some compatible implementations: GNU coreutils, FreeBSD (at least from version 8.2?), Busybox (requires to be compiled with options FANCY_SLEEP and FLOAT_DURATION).
> The strtod behaviour is C and POSIX compatible (i.e. strtod("infinity", 0) is always valid in C99-conformant implementations, see §7.20.1.3).

Said that, probably the final solution should add its own command that does the linking & waiting (and signal handling) on its own (idealy leveraging the `pause()` system call), and then provide the base image from kserve.

rhuss:
Can anybody with enough karma please trigger the Github workflows. That would be great! (@yuzisun maybe?)

yuzisun:
> Can anybody with enough karma please trigger the Github workflows. That would be great! (@yuzisun maybe?)

Thanks @rhuss for the PR ! Any interest in presenting this on the community meeting next week ?

rhuss:
> Thanks @rhuss for the PR ! Any interest in presenting this on the community meeting next week ?

@yuzisun yes, sure! Where can I find the calendar for these  meetings ? Sorry, I checked kserve.io, but couldn't find it :(

taneem-ibrahim:
Hi @rhuss Thank you for this PR. I added you to the Sept 17th community call and added this PR on the agenda.

yuzisun:
> > Thanks @rhuss for the PR ! Any interest in presenting this on the community meeting next week ?
> 
> @yuzisun yes, sure! Where can I find the calendar for these meetings ? Sorry, I checked kserve.io, but couldn't find it :(

The meeting calendar is here, the next one is 9/13.
https://wiki.lfaidata.foundation/calendar/spacecalendar.action?spaceKey=kserve

rhuss:
Thanks, I will join. Looking forward to the call ...

yuzisun:
> Thanks, I will join. Looking forward to the call ...

@rhuss just to confirm if you can join the call tomorrow at 12pm EST time ?

rhuss:
@yuzisun yes, I will be there.

rhuss:
@yuzisun I've added some configuration to finetune the behavior (e.g. the UID under which the containers are running and the resources + comprehensive tests for this feature.

I think it would be ready for prime time now, and any review would be highly appreciated.

I'm planning to add some documentation with examples, but I'm unsure where I should put this. Any recommendations?

rhuss:
I overlooked that the process for adding a new feature involves the creation of a [design document](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c/edit), sorry for my ignorance. @yuzisun is this something I should still do, or is the description and demo video in this PR good enough ?

yuzisun:
> design document

Yes, I think you have most of the content and would just need to move to the design doc, thanks!

python/sklearnserver/sklearnserver/__main__.py
```diff
@@ -31,6 +32,7 @@
 args, _ = parser.parse_known_args()
 
 if __name__ == "__main__":
+    Storage.wait_for_model_data(args.model_dir)
```
yuzisun: I think if this is not the right place for waiting the model as it only applies to KServe implemented model runtimes, we need to make sure it works for all supported serving runtimes including triton, torchserve etc. So I think we need some other ways to ensure that the model data is already mounted and can be loaded into the main memory before model server container starts.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -382,6 +453,99 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+		}
+	}
+
+	container.Env = append(container.Env, v1.EnvVar{
+		Name:  envKey,
+		Value: envValue,
+	})
+}
+
+func (mi *StorageInitializerInjector) createModelContainer(image string, modelPath string) *v1.Container {
```
yuzisun, line 542: In KServe 0.11.1 [this PR](https://github.com/kserve/kserve/pull/3060) introduced `ClusterStorageContainer` which allows configuring custom storage container based on supported storage uri prefixes. For example you can customize a storage container using `oci://` prefix and in this way we no longer need to hard code the container spec here. We can also add an annotation to indicate injecting as sidecar or init container.

python/sklearnserver/sklearnserver/__main__.py
rhuss: Thanks, let revert the Python changes, as they are out of scope of this PR. One can consider this fixing dependency (e.g. that a model might not be immediately available during startup) in a latter PR.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: Let me dive into the ClusterStorageContainer feature, we might also add it with such a CR. As you say, we probably need to extend the mechanism to allow side-car based injection.

python/sklearnserver/sklearnserver/__main__.py
rhuss: Removed in latest commit.

rhuss:
@yuzisun I've added a design document now: [Enhancing KServe Model Fetching with Modelcars](https://docs.google.com/document/d/1Bs4fnP8rhPMaoPoLSYVvuRq-z9vkGPQ0rKbmfH4I7js/edit?usp=sharing), please feel free to distribute to any interested reviewer.

In the meantime, I'm looking deeper into `StorageContainer` and how this approach could be integrated with this.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: @yuzisun I've added a section to the [design document](https://docs.google.com/document/d/1Bs4fnP8rhPMaoPoLSYVvuRq-z9vkGPQ0rKbmfH4I7js/edit#bookmark=id.qb0sqvo6n1bq), that describes what changes would be needed in `ClusterStorageContainer` CRD to support model cars.

Please let me know if this is a potential way to go, I would then create an issue for these extensions and might be able to work on it, too.

rhuss:
I summarized the required changes to ClusterStorageContainer to support modelcar in #3192.

rhuss:
Sorry for the delay. I'm back on this PR, and I still think its a good idea to merge it. I rebased it to the latest changes (would be cool, if someone could trigger the tests). As mentioned in #3192, it's probably a not-so-good idea to add this to the `ClusterStorageContainer` because it would require some more flexibility for the `ClusterStorageContainer` to extract the container image to use for the initialization from the given storage URL. 

The whole feature is hidden behind a feature flag with a default to `false`, so it should have no impact if not used.

@yuzisun @taneem-ibrahim how should we proceed ? Is there anything I can do in addition to bring this PR forward ?

yuzisun:
> Sorry for the delay. I'm back on this PR, and I still think its a good idea to merge it. I rebased it to the latest changes (would be cool, if someone could trigger the tests). As mentioned in #3192, it's probably a not-so-good idea to add this to the `ClusterStorageContainer` because it would require some more flexibility for the `ClusterStorageContainer` to extract the container image to use for the initialization from the given storage URL.
> 
> The whole feature is hidden behind a feature flag with a default to `false`, so it should have impact if not used.
> 
> @yuzisun @taneem-ibrahim how should we proceed ? Is there anything I can do in addition to bring this PR forward ?

going through the PR now.

yuzisun:
I am going to test out this with a LLaMA model see how it works.

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -86,7 +89,23 @@ data:
...
+           # using a source URL with an "oci://" schema.
+           "enableModelcar": false,
+
+           # cpuModelcar is the cpu request and limit that is used for the passive modelcar container. It can be
+           # set very low, but should be allowed by any Kubernetes LimitRange that might apply.
+           "cpuModelcar": "10m",
+
+           # cpuModelcar is the memory request and limit that is used for the passive modelcar container. It can be
+           # set very low, but should be allowed by any Kubernetes LimitRange that might apply.
+           "memoryModelcar": "15Mi",
```
yuzisun, line 104: We should separate requests and limits, this number seems a bit too low and need to test with a reasonable large size model.

yuzisun:
> I am going to test out this with a LLaMA model see how it works.

tested, looks good.  made a small fix for the ConfigMap merge issue and allow oci prefix for the validation.

yuzisun:
Thanks @rhuss for this awesome feature!

/lgtm
/approve

PR #3130, opened 2023-09-19
Fix knative config-defaults values overrided by kserve

### LINKED ISSUE #3092: Is is possible to set a default value of InferenceService's timeout (Serverless mode)
/kind question

**What's the question**
I found that a pod created by KServe takes long time to be terminated (300 seconds).
As I understand, `terminationGracePeriodSeconds` in the pod config controls this behavior and we can configure it per InferenceService by `timeout` (Ref: https://github.com/kserve/kserve/issues/1546). 
However, I'd like to decrease this value for all ISVC in our platform, because they take too long time to evict their pods during K8S node operation.
Would it be possible to set a default value of InferenceService's timeout? 

Btw, the below shows my experiment with adjusting a default value at KNative Serving side. It doesn't work.

```
kind: InferenceService
metadata:
  name: pytorch-mnist
  namespace: default
spec:
  predictor:
    model:
      modelFormat:
        name: pytorch
      storageUri: gs://kfserving-examples/models/torchserve/image_classifier/v2

```

1. Set a default value of `revision-timeout-seconds` in KNative 
```
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app.kubernetes.io/component: controller
    app.kubernetes.io/name: knative-serving
  name: config-defaults
  namespace: knative-serving
data:
  revision-timeout-seconds: "120"
```
2. Restart KNative controller
3. Deploy the above KServe's InferenceService
4. `terminationGracePeriodSeconds` in InferenceService's pod is still set to `300`

**Environment:**

- Istio Version: v1.18.0
- Knative Version: v1.10.2
- KServe Version: v0.10.1
- Kubeflow version: N/A
- Cloud Environment: AWS
- Minikube/Kind version:
- Kubernetes version: 1.24

sivanantha321:

Fixes #3092

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
@@ -28,14 +28,15 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"knative.dev/pkg/kmp"
 	"knative.dev/serving/pkg/apis/autoscaling"
+	knserving "knative.dev/serving/pkg/apis/serving"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
-
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -177,9 +178,6 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 			},
 		},
 	}
-	//Call setDefaults on desired knative service here to avoid diffs generated because knative defaulter webhook is
-	//called when creating or updating the knative service
-	service.SetDefaults(context.TODO())
 	return service
 }
 
@@ -225,6 +223,19 @@ func (r *KsvcReconciler) Reconcile() (*knservingv1.ServiceStatus, error) {
 			log.Info("Creating knative service", "namespace", desired.Namespace, "name", desired.Name)
 			return &desired.Status, r.client.Create(context.TODO(), desired)
 		}
+		return nil, err
+	}
+	// Set ResourceVersion which is required for update operation.
+	desired.ResourceVersion = existing.ResourceVersion
+	// Add immutable annotations to avoid validation error during dry-run update.
+	desired.Annotations[knserving.CreatorAnnotation] = existing.Annotations[knserving.CreatorAnnotation]
+	desired.Annotations[knserving.UpdaterAnnotation] = existing.Annotations[knserving.UpdaterAnnotation]
+
+	// Do a dry-run update to avoid diffs generated because knative defaulter webhook is
+	// called when creating or updating the knative service. This will populate our local knative service object with any default values
+	// that are present on the remote version.
+	if err = r.client.Update(context.TODO(), desired, kclient.DryRunAll); err != nil {
+		log.Error(err, "Failed to perform dry-run create of knative service", "service", desired.Name)
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -179,7 +181,11 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 	}
 	//Call setDefaults on desired knative service here to avoid diffs generated because knative defaulter webhook is
 	//called when creating or updating the knative service
-	service.SetDefaults(context.TODO())
+	// Before calling setDefaults load the knative configmap into the context to avoid overriding the knative configmap values.
+	ctx := context.TODO()
+	knConfigStore := knservingconfig.NewStore(knlogging.FromContext(ctx).Named(constants.KnativeReconcilerName))
+	ctx = knservingconfig.ToContext(ctx, knConfigStore.Load())
+	service.SetDefaults(ctx)
```
yuzisun: I think probably not good idea to add too much Knative dependency here, calling dry-run could be a better option to get the expected object.

https://github.com/kserve/kserve/blob/1cc7e6c94eb848ba934eb75129ffba44db1fe2ed/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go#L103

sivanantha321: Done. Thanks for the suggestion :)

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -227,6 +225,19 @@ func (r *KsvcReconciler) Reconcile() (*knservingv1.ServiceStatus, error) {
 		}
 		return nil, err
 	}
+	// Set ResourceVersion which is required for update operation.
+	desired.ResourceVersion = existing.ResourceVersion
```
sivanantha321, line 228: @yuzisun From K8s docs: You can overwrite (update) an existing resource - for example, a ConfigMap - using an HTTP PUT. For a PUT request, it is the client's responsibility to specify the resourceVersion (taking this from the object being updated). Kubernetes uses that resourceVersion information so that the API server can detect lost updates and reject requests made by a client that is out of date with the cluster. In the event that the resource has changed (the resourceVersion the client provided is stale), the API server returns a 409 Conflict error response.

https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply

https://github.com/argoproj/argo-cd/issues/3657#issuecomment-722706739

https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/53#issuecomment-874186845

https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/53#issuecomment-874751106

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -227,6 +225,19 @@ func (r *KsvcReconciler) Reconcile() (*knservingv1.ServiceStatus, error) {
...
+	// Add immutable annotations to avoid validation error during dry-run update.
+	desired.Annotations[knserving.CreatorAnnotation] = existing.Annotations[knserving.CreatorAnnotation]
```
yuzisun, line 230: Is updatorAnnotation immutable ? The updator for desired service can be different from existing service.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -227,6 +225,19 @@ func (r *KsvcReconciler) Reconcile() (*knservingv1.ServiceStatus, error) {
...
+	desired.Annotations[knserving.UpdaterAnnotation] = existing.Annotations[knserving.UpdaterAnnotation]
```
yuzisun, line 231: Is updatorAnnotation immutable ? The updator for desired service can be different from existing service.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
sivanantha321: No updatorAnnotation is not immutable. Only the creatorAnnotation is immutable.

yuzisun: Right, my question is that the desired `updatedAnnotation` can be different from existing one, so we should not set the desired updator annotation to the existing updator annotation right?

yuzisun: Discussed with @sivanantha321 Knative has the check to prevent user setting the `updatorAnnotation`.

PR #3134, opened 2023-09-20
Fix validation for custom storageUri

### LINKED ISSUE #3121: Ability to add custom storage initializer prefixes
/kind feature

Right now I have a custom storage initializer logic that will look for a prefix that indicates it should use our internal model registry and client logic to download. That all seems to be working fine but then for the validating webhook it fails because the list is hardcoded to specific prefixes. Would it be possible to have additional ones via config and have the default storage initializer pod injection logic? 

https://github.com/kserve/kserve/blob/47442aeeac438c5697c8e6f4494bfbc6b31e7fd8/pkg/apis/serving/v1beta1/component.go#L49
This is the section that hardcodes the prefixes

greenmoon55:
We introduced storage container CRD in #3060 which allows users to use any custom storage uri and customize the container spec.
For more information, please take a look at the [design doc](https://docs.google.com/document/d/1UcHFCQcTcDBcZVd8MTEgSXXKr7GMpzHpjuJW8gEwyFU/edit?usp=sharing) and the [default StorageContainer CR](https://github.com/kserve/kserve/blob/823107299b8d93481c8d179ce2bbc726d0414448/charts/kserve-resources/templates/clusterstoragecontainer.yaml). 

However, we reject storage uris which are supported in a storage container CR because we have [another validation](https://github.com/kserve/kserve/blob/823107299b8d93481c8d179ce2bbc726d0414448/pkg/apis/serving/v1beta1/component.go#L166) in the validatingwebhook, which rejects uris not supported by the default storage initializer image.

This PR fixes the validation to allow custom uris that are supported by a storage container CR, and moves the validation logic from validatingwebhook to the controller.

#### Alternatives considered
1. Move the validation to storage initailizer injector (pod mutator) #3131. It was abandoned because the error only appears in the logs which is confusing to users.
2. Fetch storage container CRs in validating webhook. This seems promising but there are some issues.
    - Creating an isvc might take longer (not sure)
    - It could be a bigger change because we need to pass the kubebuilder client to many places (search [Validate()](https://github.com/search?q=repo%3Akserve%2Fkserve%20validate()&type=code)) including tests. 
    - It is a good idea to validate the isvc based on another CR (storage container) in the validating webhook?
   
#### Breaking change
Some [invalid Azure uris](https://github.com/kserve/kserve/pull/3134#pullrequestreview-1636589563) will pass the validation if they have the [default StorageContainer CR](https://github.com/kserve/kserve/blob/823107299b8d93481c8d179ce2bbc726d0414448/charts/kserve-resources/templates/clusterstoragecontainer.yaml#L23) installed because the last line allows all http(s) uris. 

Examples:
```
https://blob.core.windows.net/triton/simple_string/
https://foo.blob.core.windows.net/"
```

Fixes #3121

**Type of changes**

#### Test 1 Custom storage uri + custom StorageContainer CR
```yaml
kind: ClusterStorageContainer
metadata:
  name: abc
spec:
  container:
    name: storage-initializer
    image: kserve/storage-initializer:latest
    resources:
      requests:
        memory: 100Mi
        cpu: 100m
      limits:
        memory: 1Gi
        cpu: "1"
  supportedUriFormats:
    - prefix: abc://
```

```yaml
aiVersion: "serving.kserve.io/v1beta1"
kind: InferenceService
metadata: 
    name: sklearn-iris
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      storageUri: "abc://kfserving-examples/models/sklearn/1.0/model"
```
Inference service created and no error in the controller

#### Test 2 Custom storage uri + [default StorageContainer CR](https://github.com/kserve/kserve/blob/823107299b8d93481c8d179ce2bbc726d0414448/charts/kserve-resources/templates/clusterstoragecontainer.yaml) + custom StorageContainer CR

Same result as above.

#### Test 3 Custom storage uri + no StorageContainer CR
```
➜  ~ k get isvc
NAME           URL   READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION   AGE
sklearn-iris                                                                               25s
➜  ~ k describe isvc sklearn-iris
Name:         sklearn-iris
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  serving.kserve.io/v1beta1
Kind:         InferenceService
Metadata:
  Creation Timestamp:  2023-09-22T00:49:23Z
  Finalizers:
    inferenceservice.finalizers
  Generation:        1
  Resource Version:  167831
  UID:               fc9100e6-4816-45d1-a41a-9bf9da26bbf6
Spec:
  Predictor:
    Model:
      Model Format:
        Name:  sklearn
      Name:    
      Resources:
      Storage Uri:  abc://kfserving-examples/models/sklearn/1.0/model
Events:
  Type     Reason         Age                 From                Message
  ----     ------         ----                ----                -------
  Warning  InternalError  10s (x13 over 31s)  v1beta1Controllers  StorageURI not supported: storageUri, must be one of: [gs://, s3://, pvc://, file://, https://, http://, hdfs://, webhdfs://] or match https://{}.blob.core.windows.net/{}/{} or be an absolute or relative local path. StorageUri [abc://kfserving-examples/models/sklearn/1.0/model] is not supported.
```

#### Test 4 Supported uri + no StorageContainer CR
```yaml
kind: InferenceService
metadata: 
    name: sklearn-iris
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
```
```yaml
➜  ~ k get isvc
NAME           URL                                       READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION            AGE
sklearn-iris   http://sklearn-iris.default.example.com   True           100                              sklearn-iris-predictor-00001   59s
```
#### Test 5 Supported uri + default StorageContainer CR
Same result as above

- Logs

**Checklist**:

**Release note**:

```release-note
Fix validation for custom storageUri. Now users to use any custom storageUri defined in storage container CRs.
```


pkg/apis/serving/v1beta1/inference_service_validation_test.go
```diff
@@ -116,78 +116,6 @@ func TestInvalidAutoscalerHPAMetrics(t *testing.T) {
...
-	isvc.Spec.Predictor.Tensorflow.StorageURI = proto.String("https://kfserving.blob.core.windows.net/triton/")
-	g.Expect(isvc.ValidateCreate()).Should(gomega.Succeed())
-	isvc.Spec.Predictor.Tensorflow.StorageURI = proto.String("https://kfserving.blob.core.windows.net/triton")
-	g.Expect(isvc.ValidateCreate()).Should(gomega.Succeed())
-}
-
-func TestAzureBlobNoAccountFails(t *testing.T) {
-	g := gomega.NewGomegaWithT(t)
-	isvc := makeTestInferenceService()
-	isvc.Spec.Predictor.Tensorflow.StorageURI = proto.String("https://blob.core.windows.net/triton/simple_string/")
```
greenmoon55, line 164: This uri is actually supported in the new CR because "https" is a valid prefix. Just want to note that this is a behavior change.

pkg/apis/serving/v1beta1/inference_service_validation_test.go
```diff
@@ -116,78 +116,6 @@ func TestInvalidAutoscalerHPAMetrics(t *testing.T) {
...
-	g.Expect(isvc.ValidateCreate()).ShouldNot(gomega.Succeed())
-}
-
-func TestAzureBlobNoContainerFails(t *testing.T) {
-	g := gomega.NewGomegaWithT(t)
-	isvc := makeTestInferenceService()
-	isvc.Spec.Predictor.Tensorflow.StorageURI = proto.String("https://foo.blob.core.windows.net/")
```
greenmoon55, line 171: ditto

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
```diff
@@ -1719,3 +1721,85 @@ func TestGetPredictorEndpoint(t *testing.T) {
...
+		"webhdfs://",
+		"some/relative/path",
+		"/",
+		"foo",
+		"",
+	}
+	s := runtime.NewScheme()
+	err := v1alpha1.AddToScheme(s)
+	if err != nil {
+		klog.Error(err, "Failed to add v1alpha1 to scheme")
```
houshengbo: This is part of the test cases. If the error is not expected, we can just fail the tests here. No need to use klog.

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
```diff
@@ -1719,3 +1721,85 @@ func TestGetPredictorEndpoint(t *testing.T) {
...
+}
+
+func TestValidateStorageURIForCustomPrefix(t *testing.T) {
+	invalidUris := []string{
+		"custom://custom.com/model",
+	}
+	s := runtime.NewScheme()
+	err := v1alpha1.AddToScheme(s)
+	if err != nil {
+		klog.Error(err, "Failed to add v1alpha1 to scheme")
```
houshengbo: This is part of the test cases. If the error is not expected, we can just fail the tests here. No need to use klog.

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
```diff
@@ -1719,3 +1721,85 @@ func TestGetPredictorEndpoint(t *testing.T) {
...
+	storageContainerSpecs := &v1alpha1.ClusterStorageContainerList{
+		Items: []v1alpha1.ClusterStorageContainer{customSpec},
+	}
+	validUris := []string{
+		"custom://custom.com/model/",
+	}
+	s := runtime.NewScheme()
+	err := v1alpha1.AddToScheme(s)
+	if err != nil {
+		klog.Error(err, "Failed to add v1alpha1 to scheme")
```
houshengbo: This is part of the test cases. If the error is not expected, we can just fail the tests here. No need to use klog.

pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -72,6 +74,14 @@ func (e *Explainer) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	// StorageInitializer injector to mutate the underlying deployment to provision model data
 	if sourceURI := explainer.GetStorageUri(); sourceURI != nil {
 		annotations[constants.StorageInitializerSourceUriInternalAnnotationKey] = *sourceURI
+		err := isvcutils.ValidateStorageURI(sourceURI, e.client)
+		if err != nil {
+			isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+				Reason:  v1beta1.InvalidStorageURI,
+				Message: "StorageURI is not supported",
+			})
+			return ctrl.Result{}, fmt.Errorf("StorageURI not supported: %v", err)
```
yuzisun: model status is only meant for predictor, if the inference service has both explainer and predictor then it can override each other.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -318,3 +330,37 @@ func sortPodsByCreatedTimestampDesc(pods *v1.PodList) {
...
+func ValidateStorageURI(storageURI *string, client client.Client) error {
+	if storageURI == nil {
+		return nil
+	}
+
+	// StorageURI is defined in a storage container CR.
+	storageContainerSpec, err := pod.GetContainerSpecForStorageUri(*storageURI, client)
+	if err != nil {
+		return err
+	}
```
yuzisun, line 343: what happens if the storage container CR does not exist?

pkg/controller/v1beta1/inferenceservice/components/explainer.go
greenmoon55: removed

pkg/controller/v1beta1/inferenceservice/utils/utils.go
greenmoon55: Added comments. If no CR exists, it will be exactly the same as the current behavior.

yuzisun:
Thanks @greenmoon55 !

/lgtm
/approve

PR #3144, opened 2023-09-23
:warning: Deprecate old predictor schema

sivanantha321:
1. This PR deprecates the predictor old schema in favor of  new schema  which is introduced in `Kserve 0.8`  #1901.
2. After this PR users will get a warning message about the deprecation if they are using the old schema.
3. Upgrades Go dependencies.
4. It also addresses the breaking changes introduced by the `sigs.k8s.io/controller-runtime`.

For more info on the new predictor schema, please refer the [Kserve 0.8 release blog](https://kserve.github.io/website/0.11/blog/articles/2022-02-18-KServe-0.8-release/).

Fixes #

**Type of changes**

**Test 1**  Using Old Schema 
```yaml
kind: "InferenceService"
metadata:
  name: "sklearn-iris"
  labels:
    dummy: "to_trigger_update"
spec:
  predictor:
    sklearn:

```
**Result**
Screenshot from 2023-09-23 17-49-17

- Logs

**Checklist**:

**Release note**:

```release-note
Deprecate old predictor schema
```


pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -38,6 +38,10 @@ var (
 	}
 	// logger for the mutating webhook.
 	mutatorLogger = logf.Log.WithName("inferenceservice-v1beta1-mutating-webhook")
+
+	// deprecatedFieldName is used for checking whether the deprecated predictor old schema is used or not.
+	// If it is then a warning is thrown by the validating webhook.
+	deprecatedFieldName string
```
sivanantha321, line 44: @yuzisun Here, I am using this package level variable to find if the deprecated field is used on the validating webhook side. By default, the defaulting webhook is called before the validating webhook. It is not possible to detect if the user is using the deprecated field in the validating webhook because we are converting the old schema to new schema in the defaulting webhook. 

We can change order in which the webhook is called. But It might be a breaking change so it will not be the best solution. If you have any better solution please, let me know.

lizzzcai:
once this is merged, what is the plan to remove the old schema? We need to plan for migration if there is a plan to remove it.

sivanantha321:
> once this is merged, what is the plan to remove the old schema? We need to plan for migration if there is a plan to remove it.

We are going to deprecate it only when we graduate CRD from v1beta1 to v1. It won't be a breaking change as we will have a conversion webhook in place.

PR #3075, opened 2023-08-12
Allow ray >=2.7,<3

ddelange:
**What this PR does / why we need it**:

Loosen ray to allow 2.7:
- containing critical security fixes
- fixing python 3.11 support

follow up on https://github.com/kserve/kserve/pull/2993#issuecomment-1675249294

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3222

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

PR #3157, opened 2023-09-29
Changes to minimize exposure to security issues

israel-hdez:

Fixes some issues found in a security scan.

* python/kserve
  * Run `poetry update` over some packages.
  * Gets [a recent fix to the MSAL library](https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/3427c2577bb674196a59e58cbf5aa91ad92cba7d) to escape an unsafe string.
  * ~~Move away from [CVE-2023-33953](https://www.cve.org/CVERecord?id=CVE-2023-33953)~~
    * ~~NOTE: `ray` had to be updated to 2.6.0~~
  * Move away from [CVE-2023-37920](https://www.cve.org/CVERecord?id=CVE-2023-37920)
  * Move away from [CVE-2023-38325](https://www.cve.org/CVERecord?id=CVE-2023-38325)
  * [minor] Move away from [CVE-2023-37276](https://www.cve.org/CVERecord?id=CVE-2023-37276)

**Type of changes**

**Release note**:
```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/utils/utils.py
+++ b/python/kserve/kserve/utils/utils.py
@@ -16,7 +16,7 @@ import os
 import sys
 import uuid
 from kserve.protocol.grpc.grpc_predict_v2_pb2 import InferParameter
-from typing import Dict, Union
+from typing import Dict, Union, List
 
 from kserve.utils.numpy_codec import from_np_dtype
 import pandas as pd
@@ -149,6 +149,24 @@ def get_predict_input(payload: Union[Dict, InferRequest]) -> Union[np.ndarray, p
                 dfs.append(pd.DataFrame(input))
             inputs = pd.concat(dfs, axis=0)
             return inputs
+
+        # Handles the following input format
+        # {'inputs': [
+        #     [{'sepal_width_(cm)': 3.5},
+        #      {'petal_length_(cm)': 1.4},
+        #      {'petal_width_(cm)': 0.2},
+        #      {'sepal_length_(cm)': 5.1}]
+        # ]}
+        elif isinstance(instances[0], List) and len(instances[0]) != 0 and isinstance(instances[0][0], Dict):
+            data: Dict[str, List] = {}
+            for instance in instances:
+                for item in instance:
+                    for key, val in item.items():
+                        if key in data:
+                            data[key].append(val)
+                        else:
+                            data[key] = [val]
+            return pd.DataFrame(data)
         else:
             return np.array(instances)
 
--- a/python/lgbserver/lgbserver/test_model.py
+++ b/python/lgbserver/lgbserver/test_model.py
@@ -50,6 +50,17 @@ def test_model():
 
     response = model.predict({"instances": [request, request]})
     assert numpy.argmax(response["predictions"][0]) == 2
+
+    request = [
+        {'sepal_width_(cm)': 3.5}, {'petal_length_(cm)': 1.4},
+        {'petal_width_(cm)': 0.2}, {'sepal_length_(cm)': 5.1}
+    ]
+    response = model.predict({"inputs": [request, request]})
+    assert numpy.argmax(response["predictions"][0]) == 2
+
+    response = model.predict({"instances": [request, request]})
+    assert numpy.argmax(response["predictions"][0]) == 2
+
     # test v2 handler
     infer_input = InferInput(name="input-0", shape=[2, 4], datatype="FP32",
                              data=[[6.8, 2.8, 4.8, 1.6], [6.0, 3.4, 4.5, 1.6]])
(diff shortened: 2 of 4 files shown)


israel-hdez:
Ray seems to have an issue with Python 3.11 when the DummyModel of the tests is loaded (or invoked?):

```
2023-09-29 17:24:35,435 ERROR router.py:538 -- Unexpected error in fulfill_pending_requests.
  File "/home/ehernand/Projects/kserve/python/kserve/.venv/lib64/python3.11/site-packages/ray/serve/_private/router.py", line 529, in fulfil
l_pending_requests
    async for candidates in self.choose_two_replicas_with_backoff(
  File "/home/ehernand/Projects/kserve/python/kserve/.venv/lib64/python3.11/site-packages/ray/serve/_private/router.py", line 404, in choose
_two_replicas_with_backoff
    chosen_ids = random.sample(
                 ^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/random.py", line 439, in sample
    raise TypeError("Population must be a sequence.  "
TypeError: Population must be a sequence.  For dicts or sets, use sorted(d).
```

Since in python3.11, `random.sample` no longer is doing automatic conversion from dicts/sets to sequences and Ray is crashing because of this. So, it seems that Ray needs a fix.

That said, from [Ray's installation page](https://docs.ray.io/en/latest/ray-overview/installation.html) apparently support for python 3.11 is still experimental. So, I'm not sure if the failure in python3.11 should be a blocker for this PR.

yuzisun:
@israel-hdez I think we can separate the ray upgrade changes as there is an [open PR](https://github.com/kserve/kserve/pull/3075) already to upgrade to ray 2.7

israel-hdez:
@yuzisun Sorry for the delay.

I've updated the PR to leave out the ray upgrade, and in the description I stroked out the entry that no longer would be covered.

israel-hdez:
CI checks are passing.

cmd/agent/main.go
```diff
@@ -48,7 +48,7 @@ import (
 
 var (
 	port          = flag.String("port", "9081", "Agent port")
-	componentPort = flag.String("component-port", "8080", "Component port")
+	componentPort = flag.Int("component-port", 8080, "Component port")
```
ckadner: what is the benefit of changing the port flag to `Int`, if we need it as a `String` again in the `buildServer` function?

python/kserve/pyproject.toml
```diff
@@ -42,7 +42,7 @@ kubernetes = ">=23.3.0"
 python-dateutil = "^2.8.0"
 numpy = "^1.23.5"
 psutil = "^5.9.0"
-ray = { version = "~2.4.0", extras = ["serve"] }
+ray = {version = "~2.4.0", extras = ["serve"]}
```
ckadner: I would leave this line untouched since we are not upgrading ray in this PR

ckadner (commented):
thanks @israel-hdez -- just a few nits :-)

python/kserve/pyproject.toml
israel-hdez: I reverted.

cmd/agent/main.go
israel-hdez: To be honest, I don't know.

Being a string, a static code analysis tool reports that using `component-port` flag without sanitizing could facilitate Server-Side Request Forgery attacks. My understanding is that it would be possible to introduce special characters. For example, passing `8080/foo/bar`, buildServer would compute `127.0.0.1:8080/foo/bar` which looks more like an address.

I scanned the code several times and I'm not sure how would be the exploit, given that golang expects a `host:port` pattern and url paths are handled in another field. Most likely, I'm not seeing something that the tool has well identified.

In any case, IMHO, such exploit is very hard (if not impossible) to do, given that these are command-line args which kserve-controller binds and I don't think the user has too much control.

All-in-all, I'm doing the change just to avoid the warning in the code analysis tool. By changing to `Int` the `pflag` library does the filtering/sanitizing for us.

python/kserve/poetry.lock
```diff
@@ -2,98 +2,98 @@
 
 [[package]]
 name = "aiohttp"
-version = "3.8.4"
```
yuzisun: I think these changes should be a separate PR?

israel-hdez: I can either prepare another PR, or rearrange this PR into two commits.
Whatever works best for you.

yuzisun: Let's do separate PR

israel-hdez: OK, I did the split.

Now, this PR has only the python updates.
I opened another PR containing the updates of the go-code: #3245

israel-hdez: ...and after rebasing, looks like all security issues were fixed in the ray update PR. So, this PR is no longer needed. Only the other PR for the go-code is still relevant.

PR #3189, opened 2023-10-17
Verify codegen in CI

### LINKED ISSUE #205: Verify-codegen script as a part of the CI process
/kind feature

We are creating static go client code for the KFservice CRD. And we have scripts to verify and update these client files generated by the [kubernetes/code-generator]( https://github.com/kubernetes/code-generator) project from the APIs created using the kube-builder. Now the `hack/verify-codegen.sh` script could be a part of the CI process in this project to make sure that the CRD version, APIs and client files are consistent in any commit.
This issue is related to the pull request #177 

sivanantha321:

Fixes #205 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```


hack/update-codegen.sh
```diff
@@ -22,11 +22,14 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 CODEGEN_VERSION=$(cd "${KUBE_ROOT}" && grep 'k8s.io/code-generator' go.mod | awk '{print $2}')
 
 if [ -z "${GOPATH:-}" ]; then
-    export GOPATH=$(go env GOPATH)
+    GOPATH=$(go env GOPATH)
+    export GOPATH
```
yuzisun, line 26: Single line with export does not work?

hack/update-codegen.sh
```diff
@@ -22,11 +22,14 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
 fi
 CODEGEN_PKG="$GOPATH/pkg/mod/k8s.io/code-generator@${CODEGEN_VERSION}"
 
+# To avoid permission denied error
 chmod +x "${CODEGEN_PKG}/generate-groups.sh"
+chmod +x "${CODEGEN_PKG}/generate-internal-groups.sh"
```
yuzisun, line 32: what does internal group mean?

sivanantha321: That script is internally called by the `generate-groups.sh`. It is not something we need to worry about.

hack/update-codegen.sh
sivanantha321: Yes, It will work. It's just a best practice. For example if the go env command fails the script will fail here. But If we export in the same line the return code of the go command is masked by the export command which always succeeds. So the script won't fail. But it will eventually fail in some other line with a misleading error message.
https://github.com/koalaman/shellcheck/wiki/SC2155

PR #3199, opened 2023-10-25
RawDeployment support for Inference Graph

### LINKED ISSUE #2454: Inference Graph supports Raw Deployment
/kind feature
`isvc` supports both `Raw Deployment` and `Serverless`, but `IG` only supports `Serverless (Knative service)`, it is better to add supporting for `Raw Deployment`.

bmopuri:
This change is to bring RawDeployment to Inference graph. Currently Inference Services support raw deployment, this changes brings parity.

Fixes #2454 

**Type of changes**

**Checklist**:

**Release note**:

```release-note

```


pkg/constants/constants.go
```diff
@@ -73,6 +73,14 @@ const (
 	AgentModelDirArgName  = "--model-dir"
 )
 
+// InferenceService Annotations
```
rachitchauhan43:
```suggestion
// InferenceGraph Annotations
```

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -170,35 +172,42 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	for _, con := range ksvcStatus.Status.Conditions {
-		if con.Type == apis.ConditionReady {
-			if con.Status == "True" {
-				graph.Status.URL = ksvcStatus.URL
-			} else {
-				graph.Status.URL = nil
+		_, err = handleInferenceGraphRawDeployment(r.Client, r.Scheme, graph, routerConfig)
+
+		if err != nil {
+			return ctrl.Result{}, errors.Wrapf(err, "fails to reconcile inference graph raw")
```
rachitchauhan43:
```suggestion
			return ctrl.Result{}, errors.Wrapf(err, "fails to reconcile inference graph rawdeployment")
```

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -170,35 +172,42 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		//err = fmt.Errorf("RawDeployment mode is not supported for InferenceGraph")
```
rachitchauhan43: Let's remove this comment.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
```diff
@@ -0,0 +1,150 @@
...
+		Name:        name,
+		Namespace:   namespace,
+		Labels:      make(map[string]string),
+		Annotations: make(map[string]string),
+	}
+
+	return objectMeta
+}
+
+func constructGraphComponentExtensionSpec(annotations map[string]string) v1beta1.ComponentExtensionSpec {
```
rachitchauhan43: I feel we should not use this `ComponentExtensionSpec` from v1beta package as this is intended for ISVCs components (predictor/transformer) 
Can we create custom IGExtensionSpec struct, under v1alpha1 package, to keep ISVCs spec separate from IG spec ?

rachitchauhan43 (changes requested):

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
bmopuri: @rachitchauhan43 This is a great suggestion.  But I see a challenge here.  
ComponentExtensionSpec exists all the way from spec yaml for Preditor, Transformer etc. 

However if I introduce IGExtensionSpec, this is not really going into Inference graph spec.  Also I cannot reuse code in v1beta1.inferenceservice.reconcilers as is - good amount of refactoring is required. 

Currently I am simply using v1beta1.ComponentExtensionSpec to simply pass through the annotations mentioned in Inference graph.   
```
    serving.kserve.io/class: "hpa"
    serving.kserve.io/max-scale: "7"
    serving.kserve.io/metric: rps
    serving.kserve.io/min-scale: "1"
    serving.kserve.io/target: "40"
```

What I can do instead is, to use the ObjectMeta itself to propagate the HPA related annotations and change v1beta1.reconcilers to construct AutoScaler object using values from Annotations rather than ComponentExtensionSpec

pkg/constants/constants.go
```diff
@@ -73,6 +73,14 @@ const (
 	AgentModelDirArgName  = "--model-dir"
 )
 
+// InferenceGraph Annotations
+var (
+	InferenceGraphMinScaleAnnotationKey = KServeAPIGroupName + "/min-scale"
+	InferenceGraphMaxScaleAnnotationKey = KServeAPIGroupName + "/max-scale"
+	InferenceGraphTargetAnnotationKey   = KServeAPIGroupName + "/target"
+	InferenceGraphMetricsAnnotationKey  = KServeAPIGroupName + "/metrics"
```
yuzisun: We should be consistent with `InferenceService` using properly defined fields on spec instead of annotations.

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -168,37 +170,59 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			if con.Type == appsv1.DeploymentAvailable {
+				igAvailable = true
+				break
+			}
+		}
+		if !igAvailable {
+			//If Deployment resource not yet available, IG is not available as well. Reconcile again.
+			return reconcile.Result{Requeue: true}, errors.Wrapf(err,
+				"Failed to find inference graph deployment  %s", graph.Name)
+		}
```
yuzisun, line 190: The `Available` condition can mean that the deployment is in progress, I do not think this is the case we should retry the reconciliation.

bmopuri: @yuzisun   Thank you for your review.  In the raw deployment mode, the Deployment object is not available during the first reconciliation, momentarily DeploymentStatus returns Available status.   If we don't requeue at this place,  graph availability status and URL won't get updated until the next reconciliation phase.  

We do similar stuff when the inference service is not ready [code](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/inferencegraph/controller.go#L158)

bmopuri (commented):
@yuzisun  Your review is very well appreciated.  I have introduced Scaling fields in the InferenceGraphSpec rather than my previous approach of annotations. 

Please see if these changes look ok.

pkg/constants/constants.go
bmopuri: Agreed. Implemented accordintly

bmopuri:
@yuzisun   @rachitchauhan43   Pls let me know if we can take this change forward.  Looking forward to get this merged and start using raw deployment in our environments.

rachitchauhan43:
> @yuzisun @rachitchauhan43 Pls let me know if we can take this change forward. Looking forward to get this merged and start using raw deployment in our environments.

@bmopuri : LGTM. Can you rebase with master ?

rachitchauhan43 (approved):
LGTM

PR #3207, opened 2023-10-28
Update model.py

### LINKED ISSUE #3009: How to use set  REST_V2.value  in  `self.protocol` of class Model ?
/kind bug

How to use set  REST_V2.value  in  `self.protocol` of class Model ? 

image

https://github.com/kserve/kserve/blob/1314130e93375bbda6ba46703af07a52262d1399/python/kserve/kserve/model.py#L72C14-L72C22

I am currently using a custom container ( custom code deployed as a docker image) and currently its using `v1/models` apis .
I am facing this issue https://github.com/kserve/kserve/issues/3007 for v1 apis , hence I am migrating to `v2/models` apis.  When I searched the documentation , there was no way to pass  the protocol version for custom container. Also even Inside kserve python sdk there is no way to set protocol as  `REST_V2.value` in Model class since it's already hard-coded. 

My yaml 
```
kind: InferenceService
metadata:
  name: some-name
spec:
  predictor:
    containers:
      - name: some-name
        image: some-image:v1
        env:
          - name: PROTOCOL
            value: v2
          - name: MODEL_NAME
            value: some-name

            
```

My python code : 

```
import kserve
from typing import Dict
from kserve.utils.utils import generate_uuid
from kserve import InferRequest, InferResponse, InferOutput
import os

class AlexNetModel(kserve.Model):
    def __init__(self, name: str):
       super().__init__(name)
       self.name = name
       self.load()
       
    def load(self):
        self.ready = True

    def predict(self, payload: InferRequest, headers: Dict[str, str] = None) -> InferResponse:
       pass
  

if __name__ == "__main__":
    model = AlexNetModel("custom-model")
    kserve.ModelServer().start([model])


```

**Environment:**

- KServe Python sdk Version: kserve==0.10.1.1

DeeptanshuDas:

Fixes #3009 

**Type of changes**

**Checklist**:

**Release note**:

```release-note

```

### DIFF
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -69,7 +69,7 @@ class Model:
         """
         self.name = name
         self.ready = False
-        self.protocol = PredictorProtocol.REST_V1.value
+        self.protocol = PredictorProtocol.REST_V2.value
         self.predictor_host = None
         self.explainer_host = None
         # The timeout matches what is set in generated Istio resources.


yuzisun:
for custom predictor v1/v2 are both supported by default, you do not need to specifically set the version. The `self.protocol` field on model server is used for transformer to call out to predictor not by predictor itself

PR #3219, opened 2023-11-02
Add default clusterstoragecontainer cr into resources

### LINKED ISSUE #3132: Quick install script should install storageContainer CRD and the default CR
/kind feature

Quick install script should install storageContainer CRD and the default CR

homily707:

**Which issue(s) this PR fixes ** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3132

**Type of changes**

- [X] kustomize build ./config/default
- [X] ./hack/generate-install.sh

**Checklist**:

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ deploy: manifests
 	if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then ./hack/self-signed-ca.sh; fi;
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
 	sleep 2
-	${KUSTOMIZE} build config/runtimes | kubectl apply -f -
+	${KUSTOMIZE} build config/clusterresources | kubectl apply -f -
 
 
 deploy-dev: manifests
@@ -95,7 +95,7 @@ deploy-dev: manifests
 	# TODO: Add runtimes as part of default deployment
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
 	sleep 2
-	${KUSTOMIZE} build config/runtimes | kubectl apply -f -
+	${KUSTOMIZE} build config/clusterresources | kubectl apply -f -
 
 deploy-dev-sklearn: docker-push-sklearn kustomize
 	./hack/serving_runtime_image_patch.sh "kserve-sklearnserver.yaml" "${KO_DOCKER_REPO}/${SKLEARN_IMG}"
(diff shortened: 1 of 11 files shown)


homily707:
now only update Makefile in the pr. 'hack/generate-install.sh' still need update, I'm not sure, whether the storageContainer yaml should be appended to the 'install/<version>/kserve.yaml' or make a new file for it, but only one resource for one file?

yuzisun:
> now only update Makefile in the pr. 'hack/generate-install.sh' still need update, I'm not sure, whether the storageContainer yaml should be appended to the 'install//kserve.yaml' or make a new file for it, but only one resource for one file?


I think the storage container should be part of the kserve installation file, the goal is to replace the storage configuration in the `inferenceservice-config` configmap, so it should be part of the default installation.

Makefile
```diff
@@ -82,6 +82,7 @@ deploy: manifests
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
 	sleep 2
 	${KUSTOMIZE} build config/runtimes | kubectl apply -f -
+	${KUSTOMIZE} build config/storagecontainers | kubectl apply -f -
```
yuzisun: This should be part of the default overlay.

homily707: Does this mean add `config/storagecontainers` as part of  `config/default` resources?
I have tried that in the first version of this pr, but it didn't pass the test. due to this [issue](https://github.com/kubernetes-sigs/kustomize/issues/552). The namespace  configuration in `config/default` conflicts with `clusterstoragecontainers`. So I have to remove it from  `config/default`, and use it just like `config/runtimes`.

This also means we need to do things like this in `hack/generate-install.sh`:
the second sentence use append
```
kustomize build config/default | sed s/:latest/:$TAG/ > $INSTALL_PATH
kustomize build config/storagecontainers | sed s/:latest/:$TAG/ >> $INSTALL_PATH
```

I'm also considering another solution, to rename `config/runtimes` to `config/clusterresources`, and then we can add `clusterstorageruntimes` into `config/clusterresources`, but it may lead to more change, and I'm not sure if this is a good solution. Do you have any suggestion?

yuzisun: I see, seems like we do need to move to the cluster resources. Ideally we want to minimize the number of installation files needed, maybe we can merge the runtime and storage yamls in generate-install.sh.

homily707: updated, require approval to test

homily707:
hi～ I want to approval to run the test @yuzisun @joesepi @matzew

yuzisun:
Thanks @homily707 !!
/lgtm
/approve

PR #3230, opened 2023-11-08
Add support for tag and header based routing

### LINKED ISSUE #2428: Improved tag based routing experience
/kind feature

First thanks for the work on [tag based routing](https://github.com/kserve/kserve/tree/master/docs/samples/v1beta1/rollout#tag-based-routing). We are exploring adopting this feature as it could help provide better A/B testing experience in our system. However there are currently several pain points that make it less applicable:
1. Once enabling that feature, a `latest-` prefix will be added to the DNS label of the latest route/revision, in which case some service names with moderate length will not work anymore given there's a hard 63-character DNS label length restriction.
2. Internally we have the `Host` header reserved so it can't be used for this specific routing unfortunately :( 

We find [knative's tag header based routing](https://github.com/knative/docs/tree/main/code-samples/serving/tag-header-based-routing) feature to be helpful to unblock to some extent, which requires a revision tag to be set for each revision, and uses the revision tag `Knative-Serving-Tag` instead of the `Header` for routing. So when using both kserve's tag based routing (to add the `tag: latest` besides `tag: prev` to revisions) and knative's tag header based routing (to route based on `Knative-Serving-Tag` instead of `Header`), we could route for some basic examples though nonetheless there are still restrictions on the DNS label length (the `latest-` DNS prefix).

Some ideas to improve the story:
1. Currently we only add `tag: latest` when tag based routing feature is enabled. Can we instead have this `tag: latest` always added by default (just like how we did so for `tag: prev`), meanwhile without adding `latest-` prefix to the URL (only add `latest-` prefix to the URL when the tag based routing feature is enabled). With that we can use Knative tag header based routing without enabling kserve tag based routing and without introducing the extra DNS label length restriction.
2. Or if the above is hard to achieve, solving this PR https://github.com/kserve/kserve/pull/1681 may help unlock a bit the DNS label length restriction.

Thanks!

sivanantha321:

Fixes #2428

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```


sivanantha321:
Waiting for the opinion on how to route to a specific revision if both transformer and predictor are present

Issue #3171, opened 2023-10-05
S3 verifyssl/cabundle are not working


Jooho:
/kind bug

**What steps did you take and what happened:**
The purpose of this issue is to consolidate unresolved issues that were previously created for similar problems and work towards resolving them together.

- https://github.com/kserve/kserve/issues/1935
- https://github.com/kserve/kserve/issues/3084
- https://github.com/kserve/kserve/issues/1630

In summary, there are two major issues. 

Firstly, the problem persists where even when setting "s3-verifyssl" to false, it still attempts to perform SSL verification. 

Secondly, the "AWS_CA_BUNDLE" (serving.kserve.io/s3-cabundle) should specify the path to the certificate file, not the certificate content itself, but there seems to be no way to attach the CA bundle certificate file to the container.

I intend to address the first issue with a new pr. As for the second issue, I believe [this PR](The purpose of this issue is to consolidate unresolved issues that were previously created for similar problems and work towards resolving them together) will resolve it, but the problem is that this PR has been stagnant for approximately 5 months.

Considering the critical nature of our customers relying on self-signed certificates, I believe it is imperative to accelerate the resolution of this matter.

**What did you expect to happen:**
1. `serving.kserve.io/s3-verifyssl: '0'`  => Skip verification of SSL
2.  `serving.kserve.io/s3-cabundle: $PATH` option should work with a way to attach a secret to the container.

yuzisun:
@Jooho I think [this PR ](https://github.com/kserve/kserve/pull/2906) is to address the CA bundle mount

Jooho:
[The final pr](https://github.com/kserve/kserve/pull/3250) is merged so I will close this ticket.

nguyenthai0107:
Hello @Jooho .

im following basic [document](https://github.com/kserve/kserve/tree/master/docs/samples/cabundle).
Pretty sure  components are existed and correctly key e.g : cabundle.crt
But look like storage-initializer still cannot get cert 'cabundle.crt' for trust to call Self Signed Certificate Model Registry.
Could you please take a look .
![image](https://github.com/kserve/kserve/assets/69754985/5026c18e-fd3c-4a0c-b77e-c5e96ff2983f)

It look like error releate botocore 

```
      Exit Code:  1
      Started:    Wed, 10 Apr 2024 12:22:32 +0700
      Finished:   Wed, 10 Apr 2024 12:22:46 +0700
    Last State:   Terminated
      Reason:     Error
      Message:    
[...]

PR #2906, opened 2023-05-16
CA bundle mount options for storage initializer

danilopeixoto:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2897

**Type of changes**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevent result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [x] Test injected CA bundle secret volume
- [x] Test not injected CA bundle secret volume

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
CA bundle mount options for storage initializer
```

Issue #2897, opened 2023-05-13
Custom certificate bundle for storage initializer


danilopeixoto:
/kind feature

**Describe the solution you'd like**

The current version of KServe provides the `serving.kserve.io/s3-cabundle` annotation that allows administrators to specify the path to a certificate bundle for HTTPS validations during S3 resource initialization. However, I'm wondering if there is a way to mount a certificate bundle in the storage initializer container without having to rebuild the image.

Can KServe mount volume from secret into the container it generates? This would allow administrators to easily mount the necessary certificate bundle without the need to rebuild the container image.

**Anything else you would like to add:**

Example based on [Knative](https://knative.dev/docs/serving/tag-resolution/#procedure):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: controller
  namespace: knative-serving
spec:
  template:
    spec:
      containers:
        - name: controller
          volumeMounts:
            - name: custom-certs
              mountPath: /path/to/custom/certs
          env:
            - name: SSL_CERT_DIR
              value: /path/to/custom/certs
      volumes:
        - name: custom-certs
          secret:
            secretName: custom-certs
```

yuzisun:
@danilopeixoto this should be possible, we can probably add a configuration for the certificate secret name storage initializer can mount. Though this approach would require storing the ca bundle secret in every user namespace. The other option is to bake the ca bundle into the storage initializer container image but that also requires a custom container image build for every single kserve release.

danilopeixoto:
@yuzisun Thanks for outlining the options and consequences. Would it be feasible to include both options for administrators?

**Example**

```go
type StorageInitializerConfig struct {
	Image                      string `json:"image"`
	CpuRequest                 string `json:"cpuRequest"`
	CpuLimit                   string `json:"cpuLimit"`
	MemoryRequest              string `json:"memoryRequest"`
	MemoryLimit                string `json:"memoryLimit"`
	StorageSpecSecretName      string `json:"storageSpecSecretName"`
	CaBundleSecretName         string `json:"caBundleSecretName"`
	CaBundleVolumeMountPath    string `json:"caBundleVolumeMountPath"`
	EnableDirectPvcVolumeMount bool   `json:"enableDirectPvcVolumeMount"`
}
```

sivanantha321:
fixed by #3250 

/close

Issue #3154, opened 2023-09-27
Make ClusterServingRuntime optional


israel-hdez:
/kind feature

**Describe the solution you'd like**
There are some cases where is not desirable to install the `ClusterServingRuntime` CRD, and it is better to only have the namespace-scoped `ServingRuntime`.

The most notable case is in multi-tenant environments, where a single cluster is shared with several teams, users or customers. In such environments, namespaces are given to them and privileges are reduced to that single namespace. Thus, each user/team/customer is expected to create their own `ServingRuntime` in their namespaces.

Clearly, with reduced privileges, users would not be capable of creating `ClusterServingRuntimes`. However, by omitting creation/installation of the `ClusterServingRuntime` in the cluster, the risk of misconfigurations is fully removed.

The solution I would like is simple: provide a configuration flag in helm charts to give the option to omit installation of `ClusterServingRuntime` CRD. Clearly, this would require support in the controller codebase.

ckadner:
Similar to `--namespace-scope` mode in ModelMesh?

https://github.com/kserve/modelmesh-serving/blob/main/docs/install/README.md#cluster-scope-or-namespace-scope

ckadner:
Design template: https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU/edit

israel-hdez:
@ckadner Just to confirm, should I use the shorter RFC template you mentioned in the previous comment?
Or should I use [the longer one](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c/edit#heading=h.xw1gqgyqs5b)? (I remember in the call something was mentioned about the longer one).

ckadner:
@yuzisun mentioned in today's call, that the short template is mostly to get early feedback. The long-form is the required format in general. I don't mind either way. See PR #3155 to update the [feature request template](https://github.com/kserve/kserve/pull/3155/files#diff-148870715557a13127284f8aeaa28002ed6b034268af13c5d030ab59fd078220R17).

FYI, here is the one for ModelMesh going the other way, from single namespace to multi-namespace: https://docs.google.com/document/d/12Fscl9yEjDrOAqNEyRhA23MlqUYiHSDOP6XNi5lQYGA/edit#heading=h.6oqyeffyxm7q

israel-hdez:
[...]

PR #3250, opened 2023-11-15
feat: CA bundle mount options for storage initializer

Jooho:
This PR is a collaboration between @Jooho and @danilopeixoto based on the work done by @danilopeixoto in PR #2906 and some additional fixes contributed by @Jooho

It addresses #2897

Related issues listed in #3171

This PR is for self signed certificate for storage-initializer and this feature provides 2 ways
- Using global configuration (inferenceservice-config ConfigMap)
- Using storage-config Secret in a user namespace

If you set global configuration, the cabundle secret in a namespace where kserve controller is running will be copied to a user namespace in where ISVC object created.
However, if you want to set the cabundle secret for a specific infereceService, you can set an annotation 'serving.kserve.io/s3-cabundle-secret'

You can find the detail information from [here](https://github.com/Jooho/kserve/tree/kserve_ca_bundle_colab/docs/samples/cabundle)

Fixes #

**Type of changes**

**Notice**
The test model runtime becomes CrashLoopBackOff after running status but it is ok. The most important thing from this test is whether of not to pull a model from SSL Minio pod. So please ignore when the run pod failed at the end.

Test cluster: 
- kind v0.20.0
- Kubernetes 1.27.3
  ~~~
  Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean",   BuildDate:"2023-06-15T00:36:28Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/amd64"}
  ~~~
  
Test environment setup:
- Deploy minio with SSL

~~~
# Create demo folder
export DEMO_HOME=/tmp/demo
mkdir $DEMO_HOME
cd $DEMO_HOME

# Deploy kubernetes
kind create cluster
kubectl cluster-info --context kind-kind

# Deploy Minio with SSL
git clone git@github.com:Jooho/jhouse_openshift.git
cd jhouse_openshift/Minio/minio-tls-kserve/
source env.sh
./1.setup.sh 
./2.generate-cert.sh 
kubectl create ns minio
kubectl config set-context --current --namespace minio

sed 's+quay.io/opendatahub/modelmesh-minio-examples:caikit-flan-t5+kserve/modelmesh-minio-examples+g'  -i ${DEMO_HOME}/minio.yaml
./3.deploy-minio.sh 
cp /tmp/minio/minio_certs/root.crt /tmp/minio/minio_certs/cabundle.crt
~~~

- Deploy KServe 
~~~
cd /tmp/demo 

git clone git@github.com:Jooho/kserve.git
cd kserve

# Create the cabundle secret in kserve namespace
kubectl create ns kserve
kubectl config set-context --current --namespace kserve
kubectl create configmap cabundle --from-file=/tmp/minio/minio_certs/cabundle.crt -n kserve

# Deploy Cert Manager
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml

# Deploy KServer (RawDeployment version) - it should work with serverless version too
# Update controller/storage-initializer images built based on this PR
# Set caBundleSecretName in infereneservice-config cm

KSERVE_VERSION=0.11.1
wget https://github.com/kserve/kserve/releases/download/v${KSERVE_VERSION}/kserve.yaml
sed 's/Serverless/RawDeployment/g' -i ./kserve.yaml
sed 's+kserve/storage-initializer:v0.11.1+quay.io/jooholee/kserve-storage-initializer:cabundle-configmap+g' -i ./kserve.yaml
sed 's+kserve/kserve-controller:v0.11.1+quay.io/jooholee/kserve-controller:cabundle-configmap+g' -i ./kserve.yaml
kubectl apply -f ./kserve.yaml
~~~

**TEST**
- [X] 1. namespace scope test
- [X] 1-1. using json style storage-config
~~~
kubectl create ns kserve-demo
kubectl config set-context --current --namespace kserve-demo 
kustomize build ./config/runtimes/| sed 's/ClusterServingRuntime/ServingRuntime/g' |kubectl create -n kserve-demo -f -

kubectl create configmap local-cabundle --from-file=/tmp/minio/minio_certs/cabundle.crt -n kserve-demo

cat <<EOF|kubectl apply -f -
apiVersion: v1
stringData:
  localMinIO: |
    {
      "type": "s3",
      "access_key_id": "THEACCESSKEY",
      "secret_access_key": "THEPASSWORD",
      "endpoint_url": "https://minio.minio.svc:9000",
      "bucket": "modelmesh-example-models",
      "region": "us-south",
      "cabundle_configmap": "local-cabundle"
    }
kind: Secret
metadata:
  name: storage-config
type: Opaque   
EOF

cat<<EOF | kubectl create -f -
kind: InferenceService
metadata:
  annotations:
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: sklearn-iris-v2-rest
  namespace: kserve-demo
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      name: ""
      resources: {}
      runtime: kserve-sklearnserver
      storage:
        key: localMinIO
        path: sklearn/mnist-svm.joblib
EOF
~~~

- [X] 1-2. using annotation style storage-config
~~~
# Clean up previous test
kubectl delete isvc,secret --all --force

cat <<EOF|kubectl apply -f -
apiVersion: v1
data:
  AWS_ACCESS_KEY_ID: VEhFQUNDRVNTS0VZ
  AWS_SECRET_ACCESS_KEY: VEhFUEFTU1dPUkQ=
kind: Secret
metadata:
  annotations:
    serving.kserve.io/s3-endpoint: minio.minio.svc:9000
    serving.kserve.io/s3-region: us-east-2
    serving.kserve.io/s3-useanoncredential: "false"
    serving.kserve.io/s3-usehttps: "1"
    serving.kserve.io/s3-cabundle-configmap: "local-cabundle"
  name: storage-config
  namespace: kserve-demo
type: Opaque
EOF

cat<<EOF | kubectl create -f -
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
secrets:
- name: storage-config
EOF


cat<<EOF | kubectl create -f -
kind: InferenceService
metadata:
  annotations:
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: sklearn-iris-v2-rest
  namespace: kserve-demo
spec:
  predictor:
    serviceAccountName: sa
    model:
      modelFormat:
        name: sklearn
      name: ""
      resources: {}
      runtime: kserve-sklearnserver
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
EOF
~~~
- [X] 2. global scope test
- [X] 2-1. using inferenceservice-config configmap
~~~
kubectl delete secret,isvc --all --force
kubectl edit configmap inferenceservice-config -n kserve
  storageInitializer: |-
    {
        ...
        "caBundleConfigMapName": "cabundle", #<==
        "enableDirectPvcVolumeMount": false
    }
kubectl delete pod -l control-plane=kserve-controller-manager --force -n kserve

kubectl config set-context --current --namespace kserve-demo 

cat <<EOF|kubectl apply -f -
apiVersion: v1
stringData:
  localMinIO: |
    {
      "type": "s3",
      "access_key_id": "THEACCESSKEY",
      "secret_access_key": "THEPASSWORD",
      "endpoint_url": "https://minio.minio.svc:9000",
      "bucket": "modelmesh-example-models",
      "region": "us-south"
    }
kind: Secret
metadata:
  name: storage-config
type: Opaque   
EOF

cat<<EOF | kubectl create -f -
kind: InferenceService
metadata:
  annotations:
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: sklearn-iris-v2-rest
  namespace: kserve-demo
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      name: ""
      resources: {}
      runtime: kserve-sklearnserver
      storage:
        key: localMinIO
        path: sklearn/mnist-svm.joblib
EOF
~~~

- [X] 2-2. override the global cabundle by `cabundle_configmap` in StorageSpec
~~~
kubectl delete secret,isvc --all --force

cat <<EOF|kubectl apply -f -
apiVersion: v1
stringData:
  localMinIO: |
    {
      "type": "s3",
      "access_key_id": "THEACCESSKEY",
      "secret_access_key": "THEPASSWORD",
      "endpoint_url": "https://minio.minio.svc:9000",
      "bucket": "modelmesh-example-models",
      "region": "us-south",
      "cabundle_configmap": "local-cabundle"    #<==
    }
kind: Secret
metadata:
  name: storage-config
type: Opaque   
EOF

cat<<EOF | kubectl create -f -
kind: InferenceService
metadata:
  annotations:
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: sklearn-iris-v2-rest
  namespace: kserve-demo
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      name: ""
      resources: {}
      runtime: kserve-sklearnserver
      storage:
        key: localMinIO
        path: sklearn/mnist-svm.joblib
EOF
~~~
- [X] 2-3. override the global cabundle by annotation `serving.kserve.io/s3-cabundle-configmap` in storage-config
~~~
# Clean up previous test
kubectl delete isvc,secret --all --force

cat <<EOF|kubectl apply -f -
apiVersion: v1
data:
  AWS_ACCESS_KEY_ID: VEhFQUNDRVNTS0VZ
  AWS_SECRET_ACCESS_KEY: VEhFUEFTU1dPUkQ=
kind: Secret
metadata:
  annotations:
    serving.kserve.io/s3-endpoint: minio.minio.svc:9000
    serving.kserve.io/s3-region: us-east-2
    serving.kserve.io/s3-useanoncredential: "false"
    serving.kserve.io/s3-usehttps: "1"
    serving.kserve.io/s3-cabundle-configmap: "local-cabundle"   #<==
  name: storage-config
  namespace: kserve-demo
type: Opaque
EOF

cat<<EOF | kubectl create -f -
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
secrets:
- name: storage-config
EOF


cat<<EOF | kubectl create -f -
kind: InferenceService
metadata:
  annotations:
    "serving.kserve.io/deploymentMode": "RawDeployment"
  name: sklearn-iris-v2-rest
  namespace: kserve-demo
spec:
  predictor:
    serviceAccountName: sa
    model:
      modelFormat:
        name: sklearn
      name: ""
      resources: {}
      runtime: kserve-sklearnserver
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
EOF
~~~

- Logs

**Checklist**:

**Release note**:

```release-note
CA bundle mount options for storage initializer
```


ckadner:
@yuzisun -- I helped @Jooho to overcome some of the integration test hurdles and make sure we have proper co-authorship with @danilopeixoto. All the CI checks looks good now, but we need someone to properly review the actual code changes.

Could you please assign reviewers who has expertise with auth and certificate changes?

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundlesecret/cabundlesecret_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// Reconcile differences and update
+	diff, err := kmp.SafeDiff(desiredSecret.Data, existingSecret.Data)
+	if err != nil {
+		return fmt.Errorf("failed to diff cabundle secret: %w", err)
+	}
+	log.Info("Reconciling cabundle secret diff (-desired, +observed):", "diff", diff)
+	log.Info("Updating cabundle secret", "namespace", existingSecret.Namespace, "name", existingSecret.Name)
+	existingSecret.Data = desiredSecret.Data
+	err = c.client.Update(context.TODO(), existingSecret)
```
israel-hdez: I think that if the secret is not owned by the controller, the update should be skipped. We don't want to overwrite something that wasn't created by the controller.

python/kserve/kserve/storage/storage.py
```diff
@@ -170,6 +170,23 @@ def _download_s3(uri, temp_dir: str):
...
+        # If verify_ssl is true, then check there is custom ca bundle cert
+        if verify_ssl:
+            global_ca_bundle_secret = os.getenv("CA_BUNDLE_SECRET_NAME")
+            if global_ca_bundle_secret:
+                isvc_aws_ca_bundle_path = os.getenv("AWS_CA_BUNDLE")
+                if isvc_aws_ca_bundle_path:
+                    ca_bundle_full_path = isvc_aws_ca_bundle_path
+                else:
+                    global_ca_bundle_volume_mount_path = os.getenv("CA_BUNDLE_VOLUME_MOUNT_POINT")
+                    ca_bundle_full_path = global_ca_bundle_volume_mount_path + "/cabundle.crt"
```
israel-hdez, line 186: Should we support more than one bundle?

config/configmap/inferenceservice.yaml
```diff
@@ -63,7 +63,9 @@ data:
            "memoryLimit": "1Gi",
            "cpuRequest": "100m",
            "cpuLimit": "1",
-           "enableDirectPvcVolumeMount": false
+           "caBundleSecretName": "",
+           "caBundleVolumeMountPath": "/etc/ssl/custom-certs",
```
israel-hdez: Given most of these fields may go away, I'm no longer sure this is the right place for new stuff. Maybe new stuff should go only to the ClusterStorageContainer (see [doc](https://docs.google.com/document/d/1UcHFCQcTcDBcZVd8MTEgSXXKr7GMpzHpjuJW8gEwyFU/edit)). I'll let @yuzisun to comment about this.

In any case, minimally the mount path should be replicated in the default ClusterStorageContainer and whatever is in the CRD should have priority over the global config, since we don't know what's going to be the right place for any given container.

config/configmap/inferenceservice.yaml
```diff
@@ -82,6 +84,12 @@ data:
            # cpuLimit is the limits.cpu to set for the storage initializer init container.
            "cpuLimit": "1",
        
+           # caBundleSecretName is the Secret will be copied to a user namespace for the storage initializer init container.
+           "caBundleSecretName": "",
```
israel-hdez: I wonder if `ConfigMaps` should be used over `Secrets`.

Typically, CA Bundles do not contain private keys and these should not be need for verification purposes. So, I find more appropriate to use `ConfigMaps`. In case the CABundle of the user would contain any private keys,   that can represent a security risk and the recommendation should be to remove the keys from the bundle. Using a `ConfigMap` rather than a `Secret` would make this more evident to users.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -168,6 +169,13 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return reconcile.Result{}, errors.Wrapf(err, "fails to create InferenceServicesConfig")
 	}
+
+	// Reconcile cabundlesecret
+	caBundleSecretReconciler := cabundlesecret.NewCaBundleSecretReconciler(r.Client, r.Scheme)
```
israel-hdez: If we want to keep using `Secrets` over `ConfigMaps`, I'd suggest to have a way to turn off this reconciler.

Some people is very picky with regards to security. In other projects I've seen instances of people disliking broad privileges (with a strong emphasis around secrets), and requesting trimming down RBAC privileges to the essential ones.

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,100 @@
...
+    {
+        ...
+        "caBundleSecretName": "cabundle",
+        ...
+    }
+  ~~~
+  
+If you update this configuration after, please restart KServe controller pod.  
+
+### Using storage-config Secret
```
israel-hdez, line 42: Would this support the [common storage secret](https://github.com/kserve/kserve/tree/master/docs/samples/storage/storageSpec#create-common-storage-secret) format?

If so, for the common storage we need either (1) a way to attach multiple bundles to the single secret, or (2) be able to specify a bundle for each entry in the single secret. But that is not mentioned here.

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,100 @@
...
+    annotations:
+      serving.kserve.io/s3-cabundle-secret: local-cabundle
+      ...
+    name: storage-config
+    namespace: kserve-demo
+  type: Opaque
+  ~~~
+
+
+## Skip SSL Verification
```
israel-hdez, line 99: Also here, would this support the [common storage secret](https://github.com/kserve/kserve/tree/master/docs/samples/storage/storageSpec#create-common-storage-secret) format?

The ability to skip per entry, or globally in the single secret would also be needed.

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,100 @@
+# KServe with Self Signed Certificate Model Registry
```
israel-hdez, line 1: It is probably better to move these docs to the website for more visibility: https://github.com/kserve/website

israel-hdez (commented):
I'm leaving a few comments of quick scan to the code.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundlesecret/cabundlesecret_reconciler.go
Jooho: the secret is created by kserve controller.

python/kserve/kserve/storage/storage.py
Jooho: No. I don't think there is a need to add more than 1 bundle.

config/configmap/inferenceservice.yaml
Jooho: yes, there will be a new place to specify these fields but we don't know when the new CRD will come, so for now, this configmap is the right place to use.

config/configmap/inferenceservice.yaml
Jooho: good catch! @yuzisun what do you think? if you agree, I will update this source but this is a little much changes though

config/configmap/inferenceservice.yaml
israel-hdez: The CRD is already there: https://github.com/kserve/kserve/blob/master/config/crd/serving.kserve.io_clusterstoragecontainers.yaml

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundlesecret/cabundlesecret_reconciler.go
israel-hdez: If it does not exists, yes.
But are we OK on overwriting if there is an already existent secret that the user may have created if, by chance, has the same name?

Jooho: If users use kserve, then I think they should not use the secret name. like storage-config secret.
```
        "storageSpecSecretName": "storage-config"
```

config/configmap/inferenceservice.yaml
Jooho: is it already implemented?

pkg/controller/v1beta1/inferenceservice/controller.go
Jooho: if we move to configmap, then this issue will be solved.

docs/samples/cabundle/README.md
Jooho: No, it is a special option so annotation would be enough.

config/configmap/inferenceservice.yaml
israel-hdez: I think yes. I've never tried it. But here is the PR that introduced it: https://github.com/kserve/kserve/pull/3060

docs/samples/cabundle/README.md
Jooho: yes good idea

config/configmap/inferenceservice.yaml
Jooho: I convert the secret to configmap.

pkg/controller/v1beta1/inferenceservice/controller.go
Jooho: Converted Secret to configmap

Jooho:
@ckadner could you please check failed CI? looks like it is not related to the source.

Jooho:
@ckadner nvm, I retrigger ci and now it passed all ci :)

docs/samples/cabundle/README.md
Jooho: I don't know if it is possible globally. @yuzisun do you have any thoughts on this?

config/configmap/inferenceservice.yaml
Jooho: @greenmoon55 do you think I should create these fields in the new CRD(clusterstoragecontainers)?

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+If you are using a model registry with a self-signed certificate, you must either skip ssl verify or apply the appropriate cabundle to the storage-initializer to create a connection with the registry.
+This document explains three methods that can be used in KServe.
```
spolti:
```suggestion
This document explains three methods that can be used in KServe, described below:
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Configure CaBundle for storage-initializer
```
spolti:
```suggestion
- Configure CA bundle for storage-initializer
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+  - Global configuration
+  - Using `storage-config` Secret
+
+- Skip SSL Verification
+  
+## Configure CaBundle for storage-initializer  
+### Global Configuration
+
+KServe use `inferenceservice-config` ConfigMap for default configuration. If you want to add `cabundle` cert for every inferenceservice, you can set `caBundleConfigMapName` in the ConfigMap. Before updating the ConfigMap, you have to create a configmap for cabundle certificate in the namespace that KServe controller is running and the data key in the ConfigMap must be cabundle.crt.
```
spolti:
```suggestion
KServe use `inferenceservice-config` ConfigMap for default configuration. If you want to add `cabundle` cert for every inference service, you can set `caBundleConfigMapName` in the ConfigMap. Before updating the ConfigMap, you have to create a ConfigMap for CA bundle certificate in the namespace that KServe controller is running and the data key in the ConfigMap must be cabundle.crt. 
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+  - Using `storage-config` Secret
+
+- Skip SSL Verification
+  
+## Configure CaBundle for storage-initializer  
+### Global Configuration
+
+KServe use `inferenceservice-config` ConfigMap for default configuration. If you want to add `cabundle` cert for every inferenceservice, you can set `caBundleConfigMapName` in the ConfigMap. Before updating the ConfigMap, you have to create a configmap for cabundle certificate in the namespace that KServe controller is running and the data key in the ConfigMap must be cabundle.crt. 
+
+- Create a ConfigMap with the cabundle cert
```
spolti:
```suggestion
- Create a ConfigMap with the CA bundle cert
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+    cabundle.crt: XXXXX
+  kind: ConfigMap
+  metadata:
+    name: local-cabundle
+    namespace: kserve
+  ~~~
+
+- Add an annotation to `storage-config` Secret
+  ~~~
+  apiVersion: v1
```
spolti, line 64: if this secret is already created, what about to add a oc/kubectl patch apply command?

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -78,6 +79,16 @@ var _ = Describe("v1beta1 inference service controller", func() {
                "localGateway": "knative-serving/knative-local-gateway",
                "localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local"
             }`,
+			"storageInitializer": `{
+								"image" : "kserve/storage-initializer:latest",
```
spolti: is the tabulation correct?

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -79,6 +79,16 @@ var _ = Describe("v1beta1 inference service controller", func() {
                "localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local",
                "ingressDomain": "example.com"
             }`,
+			"storageInitializer": `{
```
spolti, line 81: same here

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+
+func NewCaBundleConfigMapReconciler(client client.Client, scheme *runtime.Scheme) *CaBundleConfigMapReconciler {
+	return &CaBundleConfigMapReconciler{
+		client: client,
+		scheme: scheme,
+	}
+}
+
+func (c *CaBundleConfigMapReconciler) Reconcile(isvc *kservev1beta1.InferenceService) error {
+	log.Info("Reconciling CaBundleConfigMap")
```
spolti: it could be debug.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+
+	isvcConfigMap := &corev1.ConfigMap{}
+	err := c.client.Get(context.TODO(), types.NamespacedName{Name: constants.InferenceServiceConfigMapName, Namespace: constants.KServeNamespace}, isvcConfigMap)
+	if err != nil {
+		log.Error(err, "Failed to find config map", "name", constants.InferenceServiceConfigMapName)
```
spolti:
```suggestion
		log.Error(err, "failed to find config map", "name", constants.InferenceServiceConfigMapName)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	storageInitializerConfig := &pod.StorageInitializerConfig{}
+	if storageInitializerConfigValue, ok := isvcConfigMap.Data["storageInitializer"]; ok {
+		err := json.Unmarshal([]byte(storageInitializerConfigValue), &storageInitializerConfig)
+		if err != nil {
+			panic(fmt.Errorf("Unable to unmarshall storage initializer json string due to %w ", err))
```
spolti:
```suggestion
			panic(fmt.Errorf("unable to unmarshal storage initializer json string due to %w ", err))
```

For errros, there is a golang pattern to use lower case.

spolti: is this expected to panic here?
there is no recover state? Wouldn't this break the operator?

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	if equality.Semantic.DeepEqual(desiredConfigMap, existingConfigMap) {
+		return nil
+	}
+
+	// Reconcile differences and update
+	diff, err := kmp.SafeDiff(desiredConfigMap.Data, existingConfigMap.Data)
+	if err != nil {
+		return fmt.Errorf("failed to diff cabundle configmap: %w", err)
+	}
+	log.Info("Reconciling cabundle configmap diff (-desired, +observed):", "diff", diff)
```
spolti: this can be debug as well, otherwise can pollute the logs.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -354,6 +359,63 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		}
 	}
 
+	// Inject CA bundle configMap if caBundleConfigMapName or constants.DefaultGlobalCaBundleConfigMapName annotation set
```
spolti:
```suggestion
	// Inject CA bundle configMap if caBundleConfigMapName or constants.DefaultGlobalCaBundleConfigMapName annotation is set
```

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -354,6 +359,63 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+			Name:  constants.GlobalCaBundleConfigMapNametEnvVarKey,
+			Value: caBundleConfigMapName,
+		})
+
+		caBundleVolumeMountPath := mi.config.CaBundleVolumeMountPath
+		if caBundleVolumeMountPath == "" {
+			caBundleVolumeMountPath = constants.DefaultCaBundleVolumeMountPath
+		}
+
+		for _, envVar := range initContainer.Env {
```
spolti, line 374: there is 2 for loops, maybe check both envs on one for loop, no?

spolti (approved):
just a few comments, not mandatory though.

docs/samples/cabundle/README.md
Jooho: no it will be created by users.

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: what do you mean?

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
Jooho: what do you mean?

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: I think it can be Info to know this reconciliation was triggered easily.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: The configuration must exist for kserve and I don't think kserve can run without this configmap. So panic should be right.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: If there is a difference between the original and copied cabundle, it should be shown up in the log before overwriting.

docs/samples/cabundle/README.md
Jooho: updated

docs/samples/cabundle/README.md
Jooho: updated

docs/samples/cabundle/README.md
Jooho: updated

docs/samples/cabundle/README.md
Jooho: updated

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: updated

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: updated

pkg/webhook/admission/pod/storage_initializer_injector.go
Jooho: updated

pkg/webhook/admission/pod/storage_initializer_injector.go
Jooho: updated

config/configmap/inferenceservice.yaml
Jooho: Discussed this in the community meeting and we agreed that we don't need to add this to `clusterstoragecontainers`

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
israel-hdez: I support @spolti comment. The second log message _"Updating cabundle configmap"_ is OK to be as Info level.

Printing the diff should be reserved for a debug-level log. It is going to be a quite big message, given that the CABundle is a somewhat big base64 file.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
israel-hdez: Minimally, add which namespace is being reconciled. Otherwise, such log line won't be very useful.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := c.ReconcileCaBundleConfigMap(newCaBundleConfigMap); err != nil {
+		return fmt.Errorf("fails to reconcile ca bundle secret: %w", err)
+	}
+
+	return nil
+}
+
+func (c *CaBundleConfigMapReconciler) getCabundleConfigMapForUserNS(caBundleNameInConfig string, kserveNamespace string, isvcNamespace string) (*corev1.ConfigMap, error) {
+	var newCaBundleConfigMap *corev1.ConfigMap
+	log.Info("In caBundleNameInConfig")
```
israel-hdez: This `log.Info` seems was for debugging purposes and should be removed.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.Error(err, "failed to find config map", "name", constants.InferenceServiceConfigMapName)
+		return err
+	}
+
+	storageInitializerConfig := &pod.StorageInitializerConfig{}
+	if storageInitializerConfigValue, ok := isvcConfigMap.Data["storageInitializer"]; ok {
+		err := json.Unmarshal([]byte(storageInitializerConfigValue), &storageInitializerConfig)
+		if err != nil {
+			panic(fmt.Errorf("unable to unmarshal storage initializer json string due to %w ", err))
```
israel-hdez: A `panic` IMO is too severe, since that will fully stop the controller and prevent other ISVCs from being reconciled.

In general, in Go a panic is usually reserved for an unrecoverable error. I don't think this error is non-recoverable.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return nil
+}
+
+func (c *CaBundleConfigMapReconciler) getCabundleConfigMapForUserNS(caBundleNameInConfig string, kserveNamespace string, isvcNamespace string) (*corev1.ConfigMap, error) {
+	var newCaBundleConfigMap *corev1.ConfigMap
+	log.Info("In caBundleNameInConfig")
+
+	// Check if cabundle Secret exist & the cabundle.crt exist in the data in controller namespace
```
israel-hdez:
```suggestion
	// Check if cabundle ConfigMap exist & the cabundle.crt exist in the data in controller namespace
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	var newCaBundleConfigMap *corev1.ConfigMap
+	if storageInitializerConfig.CaBundleConfigMapName == "" {
+		return nil
+	} else {
+		newCaBundleConfigMap, err = c.getCabundleConfigMapForUserNS(storageInitializerConfig.CaBundleConfigMapName, constants.KServeNamespace, isvc.Namespace)
+		if err != nil {
+			return fmt.Errorf("fails to get cabundle secret for creating to user namespace: %w", err)
```
israel-hdez:
```suggestion
			return fmt.Errorf("fails to get cabundle config map for creating to user namespace: %w", err)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	if err := c.ReconcileCaBundleConfigMap(newCaBundleConfigMap); err != nil {
+		return fmt.Errorf("fails to reconcile ca bundle secret: %w", err)
```
israel-hdez:
```suggestion
		return fmt.Errorf("fails to reconcile ca bundle config map: %w", err)
```

pkg/constants/constants.go
```diff
@@ -303,6 +303,21 @@ const (
 // DefaultModelLocalMountPath is where models will be mounted by the storage-initializer
 const DefaultModelLocalMountPath = "/mnt/models"
 
+// Default path to mount CA bundle secret volume
```
israel-hdez:
```suggestion
// Default path to mount CA bundle config map volume
```

pkg/constants/constants.go
```diff
@@ -303,6 +303,21 @@ const (
...
+const DefaultCaBundleVolumeMountPath = "/etc/ssl/custom-certs"
+
+// Default name for CA bundle file
+const DefaultCaBundleFileName = "cabundle.crt"
+
+// Default secret name that will be created in the user namespace.
```
israel-hdez:
```suggestion
// Default config map name that will be created in the user namespace.
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,98 @@
...
+  
+If you update this configuration after, please restart KServe controller pod.  
+
+### Using storage-config Secret
+
+If you want to apply the cabundle only to a specific inferenceservice, you can use a specific annotation on the `storage-config` Secret used by the inferenceservice.
+In this case, you have to create the cabundle ConfigMap in the user namespace before you create the inferenceservice.
+
+
+- Create a Secret with the cabundle cert
```
israel-hdez:
```suggestion
- Create a ConfigMap with the cabundle cert
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return desiredSecret
+}
+
+// ReconcileCaBundleConfigMap will manage the creation, update and deletion of the ca bundle Secret
+func (c *CaBundleConfigMapReconciler) ReconcileCaBundleConfigMap(desiredConfigMap *corev1.ConfigMap) error {
+	log.Info("Reconciling ReconcileCaBundleConfigMap")
+
+	// Create secret if does not exist
```
israel-hdez:
```suggestion
	// Create ConfigMap if does not exist
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+		Data: cabundleData,
+	}
+
+	return desiredSecret
+}
+
+// ReconcileCaBundleConfigMap will manage the creation, update and deletion of the ca bundle Secret
+func (c *CaBundleConfigMapReconciler) ReconcileCaBundleConfigMap(desiredConfigMap *corev1.ConfigMap) error {
+	log.Info("Reconciling ReconcileCaBundleConfigMap")
```
israel-hdez: This log message is going to be somewhat redundant with the one at the `Reconcile` func. Most likely, it should be removed.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil, fmt.Errorf("specified cabundle file %s not found in cabundle secret %s",
+				constants.DefaultCaBundleFileName, caBundleNameInConfig)
+		} else {
+			configData := map[string]string{
+				constants.DefaultCaBundleFileName: caBundleConfigMapData,
+			}
+			newCaBundleConfigMap = getDesiredCaBundleConfigMapForUserNS(constants.DefaultGlobalCaBundleConfigMapName, isvcNamespace, configData)
+		}
+	} else {
+		return nil, fmt.Errorf("can't read cabundle secret %s: %w", constants.DefaultCaBundleFileName, err)
```
israel-hdez:
```suggestion
		return nil, fmt.Errorf("can't read cabundle config map %s: %w", constants.DefaultCaBundleFileName, err)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return newCaBundleConfigMap, nil
+}
+
+func getDesiredCaBundleConfigMapForUserNS(secretName string, namespace string, cabundleData map[string]string) *corev1.ConfigMap {
```
israel-hdez:
```suggestion
func getDesiredCaBundleConfigMapForUserNS(configMapName string, namespace string, cabundleData map[string]string) *corev1.ConfigMap {
```

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -168,6 +169,13 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err != nil {
 		return reconcile.Result{}, errors.Wrapf(err, "fails to create InferenceServicesConfig")
 	}
+
+	// Reconcile cabundleConfigMap
+	caBundleConfigMapReconciler := cabundleconfigmap.NewCaBundleConfigMapReconciler(r.Client, r.Scheme)
```
israel-hdez, line 174: I know I mentioned that by using a ConfigMap, this would be OK.

However, thinking more deeply, for multi-tenant environments I think it is still worth to have a way to fully turn off this.

python/kserve/kserve/storage/storage.py
```diff
@@ -170,6 +170,23 @@ def _download_s3(uri, temp_dir: str):
...
+
+                if os.path.exists(ca_bundle_full_path):
+                    print(f"ca bundle file({ca_bundle_full_path}) exists.")
+                    kwargs.update({"verify": global_ca_bundle_volume_mount_path + "/cabundle.crt"})
```
israel-hdez: If the `AWS_CA_BUNDLE` environment variable is present, this would be set to an erroneous path (because `global_ca_bundle_volume_mount_path` would be `null`).

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+		types.NamespacedName{Name: caBundleNameInConfig, Namespace: kserveNamespace}, caBundleConfigMap); err == nil {
+
+		if caBundleConfigMapData := caBundleConfigMap.Data[constants.DefaultCaBundleFileName]; caBundleConfigMapData == "" {
+			return nil, fmt.Errorf("specified cabundle file %s not found in cabundle secret %s",
+				constants.DefaultCaBundleFileName, caBundleNameInConfig)
+		} else {
+			configData := map[string]string{
+				constants.DefaultCaBundleFileName: caBundleConfigMapData,
+			}
+			newCaBundleConfigMap = getDesiredCaBundleConfigMapForUserNS(constants.DefaultGlobalCaBundleConfigMapName, isvcNamespace, configData)
```
israel-hdez, line 105: The name of the ConfigMap in the user namespace is not configurable, right?

pkg/constants/constants.go
```diff
@@ -303,6 +303,21 @@ const (
...
+const DefaultGlobalCaBundleConfigMapName = "global-ca-bundle"
+
+// Custom CA Secret Environment Variables
+const (
+	GlobalCaBundleConfigMapNametEnvVarKey   = "CA_BUNDLE_CONFIGMAP_NAME"
+	GlobalCaBundleVolumeMountPathtEnvVarKey = "CA_BUNDLE_VOLUME_MOUNT_POINT"
```
israel-hdez: Fix typo

```suggestion
	GlobalCaBundleConfigMapNameEnvVarKey   = "CA_BUNDLE_CONFIGMAP_NAME"
	GlobalCaBundleVolumeMountPathEnvVarKey = "CA_BUNDLE_VOLUME_MOUNT_POINT"
```

Nit: the `Global` prefix is somewhat confusing, gives that this applies to both the global bundle, and also the local one.

israel-hdez (commented):
I did a 2nd review round and left some additional comments.

In general, because of the move to ConfigMap,there are lots of mentions of `secret` while it should be `config map`.  I pointed a few, but there are several left.

In general, with regards functionality, code looks quite fine. However, I feel somewhat reluctant to approve until docs are moved to the website.

pkg/controller/v1beta1/inferenceservice/controller_test.go
spolti: indentation :)
sometimes my english fails LOL

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
spolti: okay, thanks.

Jooho:
this pr needs to include this issue(https://github.com/kserve/kserve/issues/3263) to support json style of storage-config properly.

python/kserve/kserve/storage/storage.py
Jooho: oh it should be 
~~~
    kwargs.update({"verify": ca_bundle_full_path})
~~~
will update.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: No

pkg/constants/constants.go
Jooho: good catch! 
updated
~~~
	CaBundleConfigMapNameEnvVarKey   = "CA_BUNDLE_CONFIGMAP_NAME"
	CaBundleVolumeMountPathEnvVarKey = "CA_BUNDLE_VOLUME_MOUNT_POINT"
~~~

docs/samples/cabundle/README.md
Jooho: this will be solved by this pr(https://github.com/kserve/kserve/pull/3267)

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: updated

pkg/controller/v1beta1/inferenceservice/controller.go
Jooho: @danielezonca could you please take a look at this comment?

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
Jooho: updated

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: updated

Jooho:
[This issue](https://github.com/kserve/kserve/issues/3263) is merged and this pr is updated.

The description of this PR is updated to test 5 use-cases
- namespace scope test
  - using json style storage-config
  -  using annotation style storage-config
- global scope test
  - using `inferenceservice-config` configmap
  - override the global cabundle by `cabundle_configmap` in StorageSpec
  - override the global cabundle by `serving.kserve.io/s3-cabundle-configmap` in storage-config

I tested these cases locally and works fine. In addition to this, I tested `skip validation ssl` using json style and annotation. Then it worked as well.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -1159,6 +1324,742 @@ func TestParsePvcURI(t *testing.T) {
...
+			},
+		},
+		"MountsCaBundleConfigMapVolumeByAnnotationInstreadOfConfigMap": {
+			storageConfig: &StorageInitializerConfig{
+				Image:                 "kserve/storage-initializer:latest",
+				CpuRequest:            "100m",
+				CpuLimit:              "1",
+				MemoryRequest:         "200Mi",
+				MemoryLimit:           "1Gi",
+				CaBundleConfigMapName: "custom-certs", // enable CA bundle secret volume mount
```
israel-hdez:
```suggestion
				CaBundleConfigMapName: "custom-certs", // enable CA bundle config map volume mount
```

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -1159,6 +1324,742 @@ func TestParsePvcURI(t *testing.T) {
...
+		},
+		"SetMountsCaBundleConfigMapVolumePathByAnnotationInstreadOfConfigMap": {
+			storageConfig: &StorageInitializerConfig{
+				Image:                   "kserve/storage-initializer:latest",
+				CpuRequest:              "100m",
+				CpuLimit:                "1",
+				MemoryRequest:           "200Mi",
+				MemoryLimit:             "1Gi",
+				CaBundleConfigMapName:   "custom-certs", // enable CA bundle secret volume mount
+				CaBundleVolumeMountPath: "/path/to",     // set CA bundle secret volume mount path
```
israel-hdez:
```suggestion
				CaBundleConfigMapName:   "custom-certs", // enable CA bundle config map volume mount
				CaBundleVolumeMountPath: "/path/to",     // set CA bundle config map volume mount path
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			copiedConfigs := make(map[string]string)
+			for key, value := range configs {
+				if key == "storageInitializer" {
+					copiedConfigs[key] = `{
+							"image" : "kserve/storage-initializer:latest",
+							"memoryRequest": "100Mi",
+							"memoryLimit": "1Gi",
+							"cpuRequest": "100m",
+							"cpuLimit": "1",
+							"CaBundleConfigMapName": "not-exist-secret",
```
israel-hdez:
```suggestion
							"CaBundleConfigMapName": "not-exist-configmap",
```

config/configmap/inferenceservice.yaml
```diff
@@ -82,6 +84,12 @@ data:
            # cpuLimit is the limits.cpu to set for the storage initializer init container.
            "cpuLimit": "1",
        
+           # caBundleConfigMapName is the ConfigMap will be copied to a user namespace for the storage initializer init container.
+           "caBundleConfigMapName": "",
+
+           # caBundleVolumeMountPath is the mount point for the secret set by caBundleConfigMapName for the storage initializer init container.
```
israel-hdez:
```suggestion
           # caBundleVolumeMountPath is the mount point for the config map set by caBundleConfigMapName for the storage initializer init container.
```

pkg/controller/v1beta1/inferenceservice/controller.go
danielezonca: I think this approach is coherent with the design of KServe so I would leave it a is: the possibility to create a global configuration requires access to the namespace where the controller is so it is already a "protected" action.

As part of this other ticket to review the cluster scoped serving runtime we might consider to review also this option with some opt-in (or opt-out) flag so that the is no _cluster global_ configurations
- https://github.com/kserve/kserve/issues/3154

But for this PR I would keep current design

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -1159,6 +1324,742 @@ func TestParsePvcURI(t *testing.T) {
...
+	// 		config: scenario.storageConfig,
+	// 		client: c,
+	// 	}
+	// 	if err := injector.InjectStorageInitializer(scenario.original); err != nil {
+	// 		t.Errorf("Test %q unexpected result: %s", name, err)
+	// 	}
+	// 	if diff, _ := kmp.SafeDiff(scenario.expected.Spec, scenario.original.Spec); diff != "" {
+	// 		t.Errorf("Test %q unexpected result (-want +got): %v", name, diff)
+	// 	}
+	// }
```
israel-hdez: Remove?

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, isvc)).Should(Succeed())
+
+			caBundleConfigMap := &v1.ConfigMap{}
+			Eventually(func() bool {
```
israel-hdez:
```suggestion
			Consistently(func() bool {
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+							"CaBundleConfigMapName": "not-exist-secret",
+							"caBundleVolumeMountPath": "/etc/ssl/custom-certs",
+							"enableDirectPvcVolumeMount": false						
+					}`
+				} else {
+					copiedConfigs[key] = value
+				}
+			}
+			//To-Do Jhouse
+			fmt.Print(copiedConfigs)
```
israel-hdez: Remove

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+							},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, isvc)).Should(Succeed())
+			defer k8sClient.Delete(ctx, isvc)
+
+			caBundleConfigMap := &v1.ConfigMap{}
+			Eventually(func() bool {
```
israel-hdez:
```suggestion
			Consistently(func() bool {
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
```
israel-hdez:
```suggestion
			Consistently(func() bool {
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -2087,4 +2098,318 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+					Name:      constants.InferenceServiceConfigMapName,
+					Namespace: constants.KServeNamespace,
+				},
+				Data: copiedConfigs,
+			}
+
+			Expect(k8sClient.Create(context.TODO(), configMap)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(context.TODO(), configMap)
+
+			//Create original cabundle configmap with wrong file name
```
israel-hdez:
```suggestion
			//Create original cabundle configmap with right file name
```

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler_test.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestGetDesiredCaBundleConfigMapForUserNS(t *testing.T) {
+	cabundleConfigMapData := make(map[string]string)
+
+	// cabundle data
+	cabundleConfigMapData["cabundle.crt"] = "SAMPLE_CA_BUNDLE"
+	targetNamespace := "test"
+	testCases := []struct {
```
israel-hdez, line 34: Not a blocking comment:
   Although I see this is following the pattern of other files, since there is a single testCase, for legibility, I'd suggest to re-write as a plain test (i.e. remove the `for` loop)

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -65,6 +70,156 @@ var (
 	}
 )
 
+func TestJOOHO(t *testing.T) {
```
israel-hdez: Remove? or Rename?

israel-hdez (commented):
@Jooho I left a few additional comments. I think nothing is a blocker. I'm happy to approve if you think it is all OK

pkg/controller/v1beta1/inferenceservice/controller.go
Jooho: thx for the feedback. I will resolve this.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
Jooho: updated.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
Jooho: updated.

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: updated.

config/configmap/inferenceservice.yaml
Jooho: updated.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
Jooho: good catch! removed

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: good catch! removed

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: updated.

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler_test.go
Jooho: I will keep this way for future possible test cases.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
Jooho: oops removed.

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: I think `Eventually` is fit in this place based on this definition.
```
 Eventually expects the condition to be met at some point in the future, while Consistently expects the condition to be consistently met over time. 
```
It does not matter when it can get the configmap. what we are looking for is at some point, it should be created and be readable.  WDYT?

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: I think `Eventually` is fit in this place based on this definition.
```
 Eventually expects the condition to be met at some point in the future, while Consistently expects the condition to be consistently met over time. 
```
It does not matter when it can get the configmap. what we are looking for is at some point, it should be created and be readable.  WDYT?

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: I think `Eventually` is fit in this place based on this definition.
```
 Eventually expects the condition to be met at some point in the future, while Consistently expects the condition to be consistently met over time. 
```
It does not matter when it can get the configmap. what we are looking for is at some point, it should be created and be readable.  WDYT?

israel-hdez: Unless I'm misunderstanding, the code is checking for absence of the ConfigMap.

To me, it can lead to false positives, because the tested condition is true since the beginning.

Changing to `Consistently` will ensure that it won't be created by mistake.

Jooho: oh right.  line #2403 should be `Eventually` but this one can be `Consistently`

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: oh right.  line #2403 should be `Eventually` but this one can be `Consistently`

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho: oh right.  line #2403 should be `Eventually` but this one can be `Consistently`

israel-hdez (approved):
This PR is now good for me.

Jooho:
@yuzisun any update on this ticket?

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,139 @@
...
+  kubectl create configmap local-cabundle --from-file=/path/to/cabundle.crt
+
+  kubectl get configmap cabundle -o yaml
+  apiVersion: v1
+  data:
+    cabundle.crt: XXXXX
+  kind: ConfigMap
+  metadata:
+    name: local-cabundle
+    namespace: kserve
```
yuzisun: this should be kserve-demo namespace?

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,139 @@
+# KServe with Self Signed Certificate Model Registry
+
+If you are using a model registry with a self-signed certificate, you must either skip ssl verify or apply the appropriate cabundle to the storage-initializer to create a connection with the registry.
```
yuzisun:
```suggestion
If you are using a model registry with a self-signed certificate, you must either skip ssl verify or apply the appropriate CA bundle to the storage-initializer to create a connection with the registry.
```

docs/samples/cabundle/README.md
```diff
@@ -0,0 +1,139 @@
...
+KServe use `inferenceservice-config` ConfigMap for default configuration. If you want to add `cabundle` cert for every inference service, you can set `caBundleConfigMapName` in the ConfigMap. Before updating the ConfigMap, you have to create a ConfigMap for CA bundle certificate in the namespace that KServe controller is running and the data key in the ConfigMap must be `cabundle.crt`. 
+
+- Create a ConfigMap with the CA bundle cert
+  ~~~
+  kubectl create configmap cabundle --from-file=/path/to/cabundle.crt
+
+  kubectl get configmap cabundle -o yaml
+  apiVersion: v1
+  data:
+    cabundle.crt: XXXXX
```
yuzisun, line 24: nit: Kubernetes uses key ca.crt

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2023 The KServe Authors.
```

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -354,6 +359,58 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		}
 	}
 
+	// Inject CA bundle configMap if caBundleConfigMapName or constants.DefaultGlobalCaBundleConfigMapName annotation is set
+	caBundleConfigMapName := mi.config.CaBundleConfigMapName
+	if ok := needCaBundelMount(caBundleConfigMapName, initContainer); ok {
```
yuzisun:
```suggestion
	if ok := needCaBundleMount(caBundleConfigMapName, initContainer); ok {
```

python/kserve/kserve/storage/storage.py
```diff
@@ -171,6 +171,25 @@ def _download_s3(uri, temp_dir: str):
...
+            global_ca_bundle_configmap = os.getenv("CA_BUNDLE_CONFIGMAP_NAME")
+            if global_ca_bundle_configmap:
+                isvc_aws_ca_bundle_path = os.getenv("AWS_CA_BUNDLE")
+                if isvc_aws_ca_bundle_path and isvc_aws_ca_bundle_path != "":
+                    ca_bundle_full_path = isvc_aws_ca_bundle_path
+                else:
+                    global_ca_bundle_volume_mount_path = os.getenv("CA_BUNDLE_VOLUME_MOUNT_POINT")
+                    ca_bundle_full_path = global_ca_bundle_volume_mount_path + "/cabundle.crt"
+                if os.path.exists(ca_bundle_full_path):
+                    print(f"ca bundle file({ca_bundle_full_path}) exists.")
```
yuzisun: use logging instead

python/kserve/kserve/storage/storage.py
```diff
@@ -171,6 +171,25 @@ def _download_s3(uri, temp_dir: str):
         if verify_ssl:
             verify_ssl = not verify_ssl.lower() in ["0", "false"]
             kwargs.update({"verify": verify_ssl})
+        else:
+            verify_ssl = True
+
+        # If verify_ssl is true, then check there is custom ca bundle cert
+        if verify_ssl:
+            global_ca_bundle_configmap = os.getenv("CA_BUNDLE_CONFIGMAP_NAME")
```
yuzisun, line 179: why do we need this env? seems like the presence of `AWS_CA_BUNDLE` is good enough.

docs/samples/cabundle/README.md
Jooho: updated.

docs/samples/cabundle/README.md
Jooho: updated

pkg/controller/v1beta1/inferenceservice/reconcilers/cabundleconfigmap/cabundle_configmap_reconciler.go
Jooho: updated

pkg/webhook/admission/pod/storage_initializer_injector.go
Jooho: updated

python/kserve/kserve/storage/storage.py
Jooho: updated.

python/kserve/kserve/storage/storage.py
Jooho: `AWS_CABUNDLE` represents the path to the CA bundle certificate, while `CA_BUNDLE_CONFIGMAP_NAME` represents the configmap name to attach to the storage initialization container.
For backward compatibility reasons, I did not change the semantics of `AWS_CABUNDLE`.

docs/samples/cabundle/README.md
Jooho: that's right bit I tried to use a slightly different name so as not to overlap with anything else.

yuzisun:
Thanks for the awesome work!! 

please help create an issue on website repo to update the doc.
/lgtm
/approve

Jooho:
doc issue : https://github.com/kserve/website/issues/324

PR #3259, opened 2023-11-22
s3 storage initializer: only set environment variables if variables are set in storage secret json

dtrifiro:
By setting environment variables to the result of `.get(..., "")`, the corresponding env variables are set to an empty value.

This is an issue for some values, such as `AWS_CA_BUNDLE`, which should be set to a path pointing to a valid CA bundle. When set to an empty string, it is propagated all the way down to `botocore.httpsession.URLLib3Session._setup_ssl_cert`,
 which interprets it as `False` and disables verification.

See https://github.com/boto/botocore/blob/6e0ec833714ed88d46e294048cdb0d3869eb2ab5/botocore/httpsession.py#L376-L382

See the original issue on https://github.com/opendatahub-io/kserve/issues/137: `InsecureRequestWarning: Unverified HTTPS request` warnings were being printed when retrieving data from S3 (using https)

Setting  `AWS_CA_BUNDLE` to a non-null value gets rid of the warning.

---
Release notes:

```release-note
fix processing of storage secrets
```

### DIFF
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -112,13 +112,17 @@ class Storage(object):  # pylint: disable=too-few-public-methods
                 storage_secret_json[key] = value
 
         if storage_secret_json.get("type", "") == "s3":
-            os.environ["AWS_ENDPOINT_URL"] = storage_secret_json.get("endpoint_url", "")
-            os.environ["AWS_ACCESS_KEY_ID"] = storage_secret_json.get("access_key_id", "")
-            os.environ["AWS_SECRET_ACCESS_KEY"] = storage_secret_json.get("secret_access_key", "")
-            os.environ["AWS_DEFAULT_REGION"] = storage_secret_json.get("region", "")
-            os.environ["AWS_CA_BUNDLE"] = storage_secret_json.get("certificate", "")
-            os.environ["S3_VERIFY_SSL"] = storage_secret_json.get("verify_ssl", "1")
-            os.environ["awsAnonymousCredential"] = storage_secret_json.get("anonymous", "")
+            for env_var, key in (
+                ("AWS_ENDPOINT_URL", "endpoint_url"),
+                ("AWS_ACCESS_KEY_ID", "access_key_id"),
+                ("AWS_SECRET_ACCESS_KEY", "secret_access_key"),
+                ("AWS_DEFAULT_REGION", "region"),
+                ("AWS_CA_BUNDLE", "certificate"),
+                ("S3_VERIFY_SSL", "verify_ssl"),
+                ("awsAnonymousCredential", "anonymous"),
+            ):
+                if key in storage_secret_json:
+                    os.environ[env_var] = storage_secret_json.get(key)
 
         if storage_secret_json.get("type", "") == "hdfs" or storage_secret_json.get("type", "") == "webhdfs":
             temp_dir = tempfile.mkdtemp()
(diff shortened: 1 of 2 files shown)


bretello:
Added a test with my latest push

dtrifiro:
Added a test with my latest push.

More details about this issue follow:

# How to reproduce and test


```bash
cd python/kserve

poetry install --extras storage --with=test
poetry run pip install 'moto[server]'
```

Now start a moto server with ssl:


```bash
moto_server --ssl

```

This will now listen on `https://127.0.0.1:5000`. Set moto env vars:


```bash
export AWS_ACCESS_KEY_ID='testing'
export AWS_SECRET_ACCESS_KEY='testing'
export AWS_SECURITY_TOKEN='testing'
export AWS_SESSION_TOKEN='testing'
export AWS_DEFAULT_REGION='us-east-1'
export AWS_ENDPOINT_URL="https://127.0.0.1:5000"
```

Save the following to a `issue.py` and run it with `poetry run python issue.py` (make sure the above env vars are set)

```python
import traceback

import os
import json

from kserve.storage import Storage

storage_config = {
    "type": "s3",
    # the following values can be uncommented to get through till AWS_CA_BUNDLE is set to ""
    # "access_key_id": "testing",
    # "secret_access_key": "testing",
    # "endpoint_url": "https://127.0.0.1:5000/",
    # "region": "us-east-1",
}

os.environ['STORAGE_CONFIG'] = json.dumps(storage_config)
for key in "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SECURITY_TOKEN", "AWS_SESSION_TOKEN", "AWS_DEFAULT_REGION", "AWS_ENDPOINT_URL":
    assert os.getenv(key) is not None

# configuration env vars are set, so this should work fine

try:
    Storage.download("s3://testbucket/", "downloaded artifact")
except Exception as exc:
    # this always fails because of missing endpoint url, credentials, etc., such as:
    # ValueError: Invalid endpoint: https://s3..amazonaws.com
    print(traceback.format_exc())


for key in "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SECURITY_TOKEN", "AWS_SESSION_TOKEN", "AWS_DEFAULT_REGION", "AWS_ENDPOINT_URL":
    assert os.getenv(key)
```

If the lines commented keys in `storage_config` are uncommented, one gets the warning from the original issue:

```console
InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. ...`

Jooho:
@dtrifiro I think this issue also can be solved by [this PR](https://github.com/kserve/kserve/pull/3267). WDYT?

dtrifiro:
@Jooho the current behaviour is completely broken and sets (unset) env variables to empty values.

Setting `S3_VERIFY_SSL` does not change `verify`'s behaviour

dtrifiro:
Rebased on #3267 with the last force push. It seems CI failures in e2e tests (`FAILED predictor/test_tensorflow.py::test_tensorflow_kserve - requests.exceptions.HTTPError: 502 Server Error: Bad Gateway for url: http://10.105.59.21/v1/models/isvc-tensorflow:predict`) seem unrelated, also see logs for [e2e-test workflow](https://github.com/kserve/kserve/actions/workflows/e2e-test.yml).

israel-hdez (approved):
This looks fine to me.

israel-hdez:
Somebody with enough privileges will need to re-run the failing test.
Probably @ckadner can help with that.

PR #3273, opened 2023-11-29
fix: compatibility with pydantic v2

dagondagon666:
- Follows Optional fields guidance for pydantic v2 [ref](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields)
- Overwrites protected_namespace in pydantic v2 to get rid of warning logs about namespace conflicts in pydantic [ref](https://docs.pydantic.dev/latest/api/config/#protected-namespaces). An example is shown as follows
```
E   UserWarning: Field "model_name" has conflict with protected namespace "model_".
E
E   You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
```

**Type of changes**

**No image version increase**

**Checklist**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/python/kserve/kserve/protocol/rest/v2_datamodels.py
+++ b/python/kserve/kserve/protocol/rest/v2_datamodels.py
@@ -14,7 +14,7 @@
 from typing import Optional, List, Union, Dict
 
 import orjson
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 
 # TODO: in the future, this file can be auto generated
 # https://pydantic-docs.helpmanual.io/datamodel_code_generator/
@@ -102,7 +102,7 @@ class ModelMetadataResponse(BaseModel):
         }
     """
     name: str
-    versions: Optional[List[str]]
+    versions: Optional[List[str]] = None
     platform: str
     inputs: List[MetadataTensor]
     outputs: List[MetadataTensor]
@@ -136,7 +136,7 @@ class RequestInput(BaseModel):
     name: str
     shape: List[int]
     datatype: str
-    parameters: Optional[Parameters]
+    parameters: Optional[Parameters] = None
     data: List
 
 
@@ -150,7 +150,7 @@ class RequestOutput(BaseModel):
         }
     """
     name: str
-    parameters: Optional[Parameters]
+    parameters: Optional[Parameters] = None
 
 
 class ResponseOutput(BaseModel):
@@ -168,7 +168,7 @@ class ResponseOutput(BaseModel):
     name: str
     shape: List[int]
     datatype: str
-    parameters: Optional[Parameters]
+    parameters: Optional[Parameters] = None
     data: List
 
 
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Could you add a test for this?

python/kserve/kserve/protocol/rest/v2_datamodels.py
```diff
@@ -14,7 +14,7 @@
 from typing import Optional, List, Union, Dict
 
 import orjson
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
```
terrytangyuan, line 17: Should we surround the imports with try-catch in for backwards compatibility purposes?

python/kserve/kserve/protocol/rest/v2_datamodels.py
```diff
@@ -229,10 +229,11 @@ class InferenceResponse(BaseModel):
         }
     """
     model_name: str
-    model_version: Optional[str]
+    model_version: Optional[str] = None
     id: str
-    parameters: Optional[Parameters]
+    parameters: Optional[Parameters] = None
     outputs: List[ResponseOutput]
+    model_config = ConfigDict(protected_namespaces='')
```
yuzisun, line 236: Why do we need the `model_config` ?

timothyjlaurent: This is the new syntax for pydantic 2 that's analogous to the nested Config class in pydantic 1.

My guess is that Pydantic2 uses "model_" a protected namespace which we need for `model_version`

python/kserve/kserve/protocol/rest/v2_datamodels.py
```diff
@@ -229,10 +229,11 @@ class InferenceResponse(BaseModel):
...
 
     class Config:
```
timothyjlaurent, line 238: This is pydantic 1 syntax.

We should move the `schema_extra` into the `json_schema_extra` field in the `model_config` above.

https://docs.pydantic.dev/latest/migration/#changes-to-config

python/kserve/kserve/protocol/rest/v2_datamodels.py
```diff
@@ -183,7 +183,7 @@ class InferenceRequest(BaseModel):
           "outputs" : [ $request_output, ... ] #optional
         }
     """
-    id: Optional[str]
+    id: Optional[str] = None
     parameters: Optional[Parameters] = None
     inputs: List[RequestInput]
     outputs: Optional[List[RequestOutput]] = None
```
timothyjlaurent, line 189: We need to replace Config with model_config if we want it to work in pydantic 2

python/kserve/kserve/protocol/rest/v2_datamodels.py
timothyjlaurent: If we do -- we should likely also likely add if/else for the the class definitions to switch from the nested Config class to model_config

python/kserve/kserve/protocol/rest/v2_datamodels.py
timothyjlaurent: This is to suppress this warning:

```
E   UserWarning: Field "model_name" has conflict with protected namespace "model_".
E
E   You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
```

So it really ought to be an empty tuple.

sivanantha321:
closing in favor of #3374

PR #3294, opened 2023-12-10
fix: Check if HPA has the same scaleTargetRef and behavior

terrytangyuan:
There maybe existing HPAs but incorrectly configured to target non-deployment resources. This PR adds additional check on scaleTargetRef.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
@@ -144,7 +144,8 @@ func (r *HPAReconciler) checkHPAExist(client client.Client) (constants.CheckResu
 }
 
 func semanticHPAEquals(desired, existing *autoscalingv2.HorizontalPodAutoscaler) bool {
-	return equality.Semantic.DeepEqual(desired.Spec.Metrics, existing.Spec.Metrics) &&
+	return equality.Semantic.DeepEqual(desired.Spec.ScaleTargetRef, existing.Spec.ScaleTargetRef) &&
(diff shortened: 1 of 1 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -144,7 +144,8 @@ func (r *HPAReconciler) checkHPAExist(client client.Client) (constants.CheckResu
 }
 
 func semanticHPAEquals(desired, existing *autoscalingv2.HorizontalPodAutoscaler) bool {
-	return equality.Semantic.DeepEqual(desired.Spec.Metrics, existing.Spec.Metrics) &&
+	return equality.Semantic.DeepEqual(desired.Spec.ScaleTargetRef, existing.Spec.ScaleTargetRef) &&
```
Jooho: Good catch!
Then, why don't we add  `Behavior` too? 
```suggestion
	return equality.Semantic.DeepEqual(desired.Spec.ScaleTargetRef, existing.Spec.ScaleTargetRef) &&
	equality.Semantic.DeepEqual(desired.Spec.Behavior, existing.Spec.Behavior) &&
```

terrytangyuan:
@Jooho Great suggestion. Added check for behavior as well!

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
yuzisun: Then this is simply the entire autoscaler spec.

terrytangyuan:
> Then this is simply the entire autoscaler spec.

@yuzisun Thanks for the review. Just updated the PR to check spec directly.

PR #3301, opened 2023-12-13
feat: Expose defaults for several batcher handler parameters

terrytangyuan:
This parameters are currently hard-coded but they would be useful if they are configurable. This PR make them configurable via inference service configmap.

### DIFF
--- a/charts/kserve-resources/templates/configmap.yaml
+++ b/charts/kserve-resources/templates/configmap.yaml
@@ -305,8 +305,7 @@ data:
            "cpuRequest": "1",
            "cpuLimit": "1",
            "maxBatchSize": "32",
-           "maxLatency": "5000",
-           "timeout": "100"
+           "maxLatency": "5000"
        }
      batcher: |-
        {
@@ -328,11 +327,8 @@ data:
            # maxBatchSize is the default maximum batch size for batcher.
            "maxBatchSize": "32",
 
-           # maxLatency is the default maximum latency for batcher.
-           "maxLatency": "5000",
-
-           # timeout is the default timeout for batcher.
-           "timeout": "100"
+           # maxLatency is the default maximum latency in milliseconds for batcher to wait and collect the batch.
+           "maxLatency": "5000"
        }
 
      # ====================================== AGENT CONFIGURATION ======================================
@@ -460,8 +456,7 @@ data:
         "cpuRequest": "1",
         "cpuLimit": "1",
         "maxBatchSize": "32",
-        "maxLatency": "5000",
-        "timeout": "100"
+        "maxLatency": "5000"
     }
   router: |-
     {
--- a/charts/kserve-resources/templates/configmap.yaml
+++ b/charts/kserve-resources/templates/configmap.yaml
@@ -303,7 +303,10 @@ data:
            "memoryRequest": "1Gi",
            "memoryLimit": "1Gi",
            "cpuRequest": "1",
-           "cpuLimit": "1"
+           "cpuLimit": "1",
+           "maxBatchSize": "32",
+           "maxLatency": "5000",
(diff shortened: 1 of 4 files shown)


pkg/batcher/handler.go
```diff
@@ -29,10 +30,10 @@ import (
...
 )
 
-const (
-	SleepTime    = time.Microsecond * 100
-	MaxBatchSize = 32
-	MaxLatency   = 5000
+var (
+	SleepTime, _    = utils.LookupEnvDurationOr("BATCH_HANDLER_SLEEP_TIME", time.Microsecond*100)
+	MaxBatchSize, _ = utils.LookupEnvIntOr("BATCH_HANDLER_MAX_BATCH_SIZE", 32)
+	MaxLatency, _   = utils.LookupEnvIntOr("BATCH_HANDLER_MAX_LATENCY", 5000)
```
yuzisun: These are default values and can be overwritten by the `MaxBatchSize` and `MaxLatency` on the inference service yaml. They shouldn't be environment variables, we can make these defaults configurable in the inference service ConfigMap [here](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L288).

terrytangyuan: Thanks. I missed that and was following a practice from another project. I'll revise the PR to make these configurable via configmap.

terrytangyuan: Updated the PR. Please take another look. Thanks.

pkg/webhook/admission/pod/batcher_injector.go
```diff
@@ -41,6 +41,9 @@ type BatcherConfig struct {
 	CpuLimit      string `json:"cpuLimit"`
 	MemoryRequest string `json:"memoryRequest"`
 	MemoryLimit   string `json:"memoryLimit"`
+	MaxBatchSize  string `json:"maxBatchSize"`
+	MaxLatency    string `json:"maxLatency"`
+	Timeout       string `json:"timeout"`
```
yuzisun: We also need to update the ConfigMap and helm config

https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L499
https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/configmap.yaml#L429

terrytangyuan: Thanks. Updated

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -321,6 +324,15 @@ data:
 
            # cpuLimit is the limits.cpu to set for the batcher container.
            "cpuLimit": "1"
+
+           # maxBatchSize is the default maximum batch size for batcher.
+           "maxBatchSize": "32",
+
+           # maxLatency is the default maximum latency for batcher.
```
yuzisun:
```suggestion
           # maxLatency is the default maximum latency in milliseconds for batcher to wait and collect the batch.
```

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -321,6 +324,15 @@ data:
...
+           "maxLatency": "5000",
+
+           # timeout is the default timeout for batcher.
+           "timeout": "100"
```
yuzisun: This is not actually a timeout, it is used internally for the select loop as a time tick. I think this should not be exposed to user.

terrytangyuan: Removed `timeout`

charts/kserve-resources/templates/configmap.yaml
terrytangyuan: Updated

PR #3305, opened 2023-12-14
feat: Support S3 transfer acceleration

terrytangyuan:
This PRs adds support for [S3 Transfer acceleration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration.html).

### DIFF
--- a/pkg/agent/storage/utils.go
+++ b/pkg/agent/storage/utils.go
@@ -127,10 +127,16 @@ func GetProvider(providers map[Protocol]Provider, protocol Protocol) (Provider, 
 		if ok && strings.ToLower(useVirtualBucketString) == "false" {
 			useVirtualBucket = false
 		}
+		useAccelerateString, ok := os.LookupEnv(s3credential.S3UseAccelerate)
+		useAccelerate := false
+		if ok && strings.ToLower(useAccelerateString) == "true" {
+			useAccelerate = true
+		}
 
 		awsConfig := aws.Config{
 			Region:           aws.String(region),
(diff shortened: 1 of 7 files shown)


sivanantha321:
Can we expose this configuration in [configmap](https://github.com/kserve/kserve/blob/c52753519e1802e73581856b1ae3a0f2f178d457/config/configmap/inferenceservice.yaml#L167) ?

pkg/agent/storage/utils.go
```diff
@@ -127,10 +127,16 @@ func GetProvider(providers map[Protocol]Provider, protocol Protocol) (Provider,
...
+		useAccelerateString, ok := os.LookupEnv(s3credential.S3UseAccelerate)
+		useAccelerate := false
+		if ok && strings.ToLower(useAccelerateString) == "true" {
+			useAccelerate = true
+		}
 
 		awsConfig := aws.Config{
 			Region:           aws.String(region),
 			S3ForcePathStyle: aws.Bool(!useVirtualBucket),
+			S3UseAccelerate:  aws.Bool(useAccelerate),
```
yuzisun, line 139: how about storage initializer? I do not see code handing here
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/storage/storage.py#L143

terrytangyuan: Added

terrytangyuan:
> Can we expose this configuration in [configmap](https://github.com/kserve/kserve/blob/c52753519e1802e73581856b1ae3a0f2f178d457/config/configmap/inferenceservice.yaml#L167) ?

Yep just added.

PR #3306, opened 2023-12-14
fix: Fix typo in S3_USE_VIRTUAL_BUCKET

terrytangyuan:
Small typo fix: `S3_USER_VIRTUAL_BUCKET` should be `S3_USE_VIRTUAL_BUCKET`.

### DIFF
--- a/pkg/credentials/s3/s3_secret.go
+++ b/pkg/credentials/s3/s3_secret.go
@@ -36,7 +36,7 @@ const (
 	S3Endpoint             = "S3_ENDPOINT"
 	S3UseHttps             = "S3_USE_HTTPS"
 	S3VerifySSL            = "S3_VERIFY_SSL"
-	S3UseVirtualBucket     = "S3_USER_VIRTUAL_BUCKET"
+	S3UseVirtualBucket     = "S3_USE_VIRTUAL_BUCKET"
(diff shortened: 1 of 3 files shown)


pkg/credentials/s3/s3_secret.go
```diff
@@ -36,7 +36,7 @@ const (
 	S3Endpoint             = "S3_ENDPOINT"
 	S3UseHttps             = "S3_USE_HTTPS"
 	S3VerifySSL            = "S3_VERIFY_SSL"
-	S3UseVirtualBucket     = "S3_USER_VIRTUAL_BUCKET"
+	S3UseVirtualBucket     = "S3_USE_VIRTUAL_BUCKET"
```
terrytangyuan, line 39: This has the risk of breaking backwards compatibility. @yuzisun Should I make this work for both and add a comment to remove the old one in major releases?

israel-hdez: Hey @terrytangyuan !

This key seems to be configurable via a ConfigMap.

IMO, it is going to be OK to fix the typo. We should, simply, state in release notes that users using the key with the typo can restore the old behavior by setting the desired key in the `inferenceservice-config` ConfigMap.

terrytangyuan:
Closing due to low priority

PR #3312, opened 2023-12-18
Add lint check for go

andyi2it:

Fixes #

The current implementation will lint check and show errors as part of PR.
The lint checks will be done to files that are changed as part of the PR.

**Type of changes**

**Release note**:

```release-note
NONE
```

This is how it would show up in files changed

image

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,35 @@
+# REF: https://github.com/golangci/golangci-lint-action
+
+name: GO Lint Check
+
+on:
+  push:
+    branches: [master, release*]
+  pull_request:
+    branches: []
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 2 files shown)


andyi2it:
@yuzisun 
Currently this lint check would be done only for files that are changed as part of their respective PR?
Should I change it to check all go files even if they are not changed?

yuzisun:
> @yuzisun Currently this lint check would be done only for files that are changed as part of their respective PR? Should I change it to check all go files even if they are not changed?

Can we run the check locally and see how many files are affected ?

terrytangyuan (commented):
It might be better to fix the errors before enabling certain linter. Otherwise, contributors need to fix lint errors that are unrelated to their changes.

andyi2it:
Closing this as we have a different mechanism for flagging lint errors

PR #3315, opened 2023-12-18
build: Switch to use kustomize in kubectl to simplify build process. Fixes #3314

### LINKED ISSUE #3314: Remove unnecessary separate kustomize binary from build process
/kind feature

`kustomize` is part of `kubectl` so we could use `kubectl` directly to simplify our build process.

/assign @terrytangyuan

terrytangyuan:
Fixes #3314.

`kustomize` is part of `kubectl` so we could use `kubectl` directly to simplify our build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,25 +73,27 @@ run: generate fmt vet lint
 deploy: manifests
 	# Remove the certmanager certificate if KSERVE_ENABLE_SELF_SIGNED_CA is not false
 	cd config/default && if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then \
-	kubectl delete --ignore-not-found=true -k ../certmanager; \
-	else kubectl apply -k ../certmanager; fi;
+	echo > ../certmanager/certificate.yaml; \
+	else git checkout HEAD -- ../certmanager/certificate.yaml; fi;
 	kubectl apply -k config/default
 	if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then ./hack/self-signed-ca.sh; fi;
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
 	kubectl apply -k config/clusterresources
+	git checkout HEAD -- ../certmanager/certificate.yaml
 
 
 deploy-dev: manifests
 	./hack/image_patch_dev.sh development
 	# Remove the certmanager certificate if KSERVE_ENABLE_SELF_SIGNED_CA is not false
 	cd config/default && if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then \
-	kubectl delete --ignore-not-found=true -k ../certmanager; \
-	else kubectl apply -k ../certmanager; fi;
+	echo > ../certmanager/certificate.yaml; \
+	else git checkout HEAD -- ../certmanager/certificate.yaml; fi;
 	kubectl apply -k config/overlays/development
 	if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then ./hack/self-signed-ca.sh; fi;
 	# TODO: Add runtimes as part of default deployment
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
 	kubectl apply -k config/clusterresources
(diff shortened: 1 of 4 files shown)


sivanantha321:
Some scripts are using kustomize under hack folder. It would be nice to replace them with kubectl as well.

Makefile
```diff
@@ -75,35 +73,35 @@ run: generate fmt vet lint
 deploy: manifests
 	# Remove the certmanager certificate if KSERVE_ENABLE_SELF_SIGNED_CA is not false
 	cd config/default && if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then \
-	${KUSTOMIZE} edit remove resource ../certmanager; \
-	else ${KUSTOMIZE} edit add resource ../certmanager; fi;
-	${KUSTOMIZE} build config/default | kubectl apply -f -
+	kubectl delete --ignore-not-found=true -k ../certmanager; \
```
sivanantha321: Will it delete the resource definition from kustomize file or from the cluster ? If self signed ca is enabled then, we should not apply the cert manager crd. It should be removed from the resource definition.

terrytangyuan: From the cluster

terrytangyuan:
> Some scripts are using kustomize under hack folder. It would be nice to replace them with kubectl as well.

Done. Thanks!

Makefile
terrytangyuan: @sivanantha321 Would you like to help test the changes in this PR to see if it works for your setup?

sivanantha321: This is not working. We have to remove the [certmanager resources from the kustomize file](https://github.com/kserve/kserve/blob/32bc9d69b60434ed33dd4639d7255a418f445327/config/default/kustomization.yaml#L16) if KSERVE_ENABLE_SELF_SIGNED_CA is true.

sivanantha321: I have tested all other targets. It is working fine.

terrytangyuan: Could you try again?

terrytangyuan: Looks like `kubectl kustomize` does not have sub-commands like `edit`. I thought I could do something like `kubectl edit remove resource -k ../certmanager;` but unfortunately that's not available.

terrytangyuan: I updated the script to remove the content in `certificate.yaml` as a workaround. Let me know if you have better alternatives.

Makefile
```diff
@@ -75,33 +73,35 @@ run: generate fmt vet lint
...
-	${KUSTOMIZE} build config/overlays/development | kubectl apply -f -
+	echo > ../certmanager/certificate.yaml; \
+	else git checkout HEAD -- ../certmanager/certificate.yaml; fi;
+	kubectl apply -k config/overlays/development
 	if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then ./hack/self-signed-ca.sh; fi;
 	# TODO: Add runtimes as part of default deployment
 	kubectl wait --for=condition=ready pod -l control-plane=kserve-controller-manager -n kserve --timeout=300s
-	${KUSTOMIZE} build config/clusterresources | kubectl apply -f -
+	kubectl apply -k config/clusterresources
+	git checkout HEAD -- ../certmanager/certificate.yaml
```
sivanantha321:
```suggestion
	git checkout HEAD -- config/certmanager/certificate.yaml
```

Makefile
sivanantha321: I don't have any other ideas either.

sivanantha321:
LGTM

Makefile
```diff
@@ -75,33 +73,35 @@ run: generate fmt vet lint
 deploy: manifests
 	# Remove the certmanager certificate if KSERVE_ENABLE_SELF_SIGNED_CA is not false
 	cd config/default && if [ ${KSERVE_ENABLE_SELF_SIGNED_CA} != false ]; then \
-	${KUSTOMIZE} edit remove resource ../certmanager; \
-	else ${KUSTOMIZE} edit add resource ../certmanager; fi;
-	${KUSTOMIZE} build config/default | kubectl apply -f -
+	echo > ../certmanager/certificate.yaml; \
```
yuzisun, line 76: This empties the certificate.yaml, `kubectl -k` does not complain?

terrytangyuan: Correct. I tried it locally without any issues.

PR #3316, opened 2023-12-18
Add compatibility for Istio CNI plugin

### LINKED ISSUE #2096: Please consider adding the ability to set the initContainer.securityContext.runAsUser for storage-initializer initContainer
/kind feature

Please consider adding the ability to set the initContainer.securityContext.runAsUser for storage-initializer initContainer. I think this could be solved either by allowing to be set on the `inferenceservice-config` for the `storage-initilializer` context or in the actual `InferenceService` CR in the `predictor`. I just didn't see a way to easily do that [here](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L43-L50). We are currently running 0.7 but I also tried with 0.8 and it didn't work there either.

The `InferenceService` will let us set the `securityContext.runAsUser` to whatever uid we want, and it will set this uid on every container in the `isvc` pod, however this is undesirable, as if we set this value to `1337` then the resulting `isvc` is never on the istio service mesh. 

I think it might be also helpful to consider other initContainers that  might show with other predictors too. For my immediate use case when `istio-cni` is enabled, the tensorflow on the `predictor` the `InferenceService` is unable to connect actually start up due to the `istio-cni` iptables getting in the way. See [here](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers) for the istio docs on how to bypass the iptables by setting this `uid` on the `initContainer`, hence we might want to consider this on other struts too to be used on the `inferenceservice-config` _or_ conversely on the `InferenceService` CR directly when creating a `isvc`.

Below are the logs from the `storage-initializer` `initContainer`

```
k -n kserve logs -f pod/flowers-sample-2-predictor-default-00002-deployment-7797dfk7hqj storage-initializer
/usr/local/lib/python3.7/site-packages/ray/autoscaler/_private/cli_logger.py:61: FutureWarning: Not all Ray CLI dependencies were found. In Ray 1.4+, the Ray CLI, autoscaler, and dashboard will only be usable via `pip install 'ray[default]'`. Please update your install command.
  "update your install command.", FutureWarning)
[I 220317 02:11:23 initializer-entrypoint:13] Initializing, args: src_uri [gs://kfserving-samples/models/tensorflow/flowers] dest_path[ [/mnt/models]
[I 220317 02:11:23 storage:52] Copying contents of gs://kfserving-samples/models/tensorflow/flowers to local
[W 220317 02:11:23 _metadata:104] Compute Engine Metadata server unavailable onattempt 1 of 3. Reason: [Errno 111] Connection refused
[W 220317 02:11:23 _metadata:104] Compute Engine Metadata server unavailable onattempt 2 of 3. Reason: [Errno 111] Connection refused
[W 220317 02:11:23 _metadata:104] Compute Engine Metadata server unavailable onattempt 3 of 3. Reason: [Errno 111] Connection refused
[W 220317 02:11:23 _default:287] Authentication failed using Compute Engine authentication due to unavailable metadata server.
```

If I can provide any more information please let me know.

sample `isvc` I have tried various implementations but this is the generic one in the docs

```
---
kind: InferenceService
metadata:
  name: flowers-sample-2
  namespace: kserve
spec:
  predictor:
    tensorflow:
      name: kserve-container
      resources:
        limits:
          cpu: "1"
          memory: 1Gi
        requests:
          cpu: "1"
          memory: 1Gi
      runtimeVersion: 1.14.0
      storageUri: gs://kfserving-samples/models/tensorflow/flowers
```

I have tried with I didn't want to do this because it would affect traffic once the workflow is online.
```
  annotations:
    traffic.sidecar.istio.io/excludeOutboundIPRanges: "142.250.0.0/24,172.217.0.0/24"
    traffic.sidecar.istio.io/excludeOutboundPorts: "443"
```
I also tried with the below on the object under `spec.predictor.tensorflow` this works but when the workload comes online, its bypassed the istio-cni entirely and is not routed through the mesh at all. So this bit is undesirable as well.

```
      securityContext:
        runAsUser: 1337
        runAsGroup: 1337
```

This can be reproduced by installing istio-cni in a test cluster with kserve 0.7 installed on it and deploying the flower-sample tensorflow that leverages the `storage-initializer` process to pull in a model of some sort in to the deployment… it should also time out when attempting to test

israel-hdez:

When Istio is installed with its CNI plugin, KServe inference services are not capable to start, because networking is not available to the storage-initializer, which is an init-container.

These changes use the approach documented in [Istio CNI docs](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers) for running the storage initializer using the same UserID as the Istio sidecar. The UID is copied from the sidecar container to cover Istio derivatives, like Maistra.

The downside is that the traffic of the storage initializer won't be captured by Istio and won't benefit from Istio features, although this should be OK for the storage-initializer case.

The modification to the UserID of the storage initializer container is done only when Istio CNI is detected.

The reinvocation policy of the mutating webhook for pods had to be changed to `IfNeeded` since the order of execution of mutating webhooks is unpredictable. The change of the policy is to ensure that the webhook would be called after Istio does the injection of the sidecar, allowing to properly set the UID of the storage initializer when needed.

Fixes #2096 

**Type of changes**

Installing Istio with the CNI plugin. See https://istio.io/latest/docs/setup/additional-setup/cni/. Then, deploying KServe with the changes in this PR and validating that the security context of storage initializer is properly configured to use the same UserID as the Istio sidecar.

Also, unit tests were added.

The backwards compatibility validation of the non-CNI case is delegated to existing automated E2E tests.

**Checklist**:

**Release note**:

```release-note
Add compatibility for Istio CNI plugin
```

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -34,6 +34,7 @@ webhooks:
     failurePolicy: Fail
     name: inferenceservice.kserve-webhook-server.pod-mutator
     sideEffects: None
+    reinvocationPolicy: IfNeeded
     admissionReviewVersions: ["v1beta1"]
     namespaceSelector:
       matchExpressions:
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -481,16 +481,127 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		}
 	}
 
+	// Add init container to the spec
+	pod.Spec.InitContainers = append(pod.Spec.InitContainers, *initContainer)
+
+	return nil
+}
+
+// SetIstioCniSecurityContext determines if Istio is installed in using the CNI plugin. If so,
+// the UserID of the storage initializer is changed to match the UserID of the Istio sidecar.
+// This is to ensure that the storage initializer can access the network.
+func (mi *StorageInitializerInjector) SetIstioCniSecurityContext(pod *v1.Pod) error {
+	// Find storage initializer container
+	var storageInitializerContainer *v1.Container
+	for idx, c := range pod.Spec.InitContainers {
+		if c.Name == StorageInitializerContainerName {
+			storageInitializerContainer = &pod.Spec.InitContainers[idx]
+		}
+	}
+
+	// If the storage initializer is not injected, there is no action to do
+	if storageInitializerContainer == nil {
+		return nil
+	}
+
 	// Allow to override the uid for the case where ISTIO CNI with DNS proxy is enabled
 	// See for more: https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers.
 	if value, ok := pod.GetAnnotations()[constants.IstioSidecarUIDAnnotationKey]; ok {
 		if uid, err := strconv.ParseInt(value, 10, 64); err == nil {
-			initContainer.SecurityContext.RunAsUser = ptr.Int64(uid)
+			if storageInitializerContainer.SecurityContext == nil {
+				storageInitializerContainer.SecurityContext = &v1.SecurityContext{}
+			}
+			storageInitializerContainer.SecurityContext.RunAsUser = ptr.Int64(uid)
+		}
+	} else {
+		// When Istio CNI is disabled, the istio-init container would be present.
+		// If it is there, there is no need to touch the security context of the pod.
+		// Reference: https://github.com/istio/istio/blob/d533e52acc54b4721d23b1332aea1f234ecbe3e6/pkg/config/analysis/analyzers/maturity/maturity.go#L134
+		for _, container := range pod.Spec.InitContainers {
+			if container.Name == constants.IstioInitContainerName {
+				return nil
+			}
+		}
+
+		// When Istio CNI is enabled, a sidecar.istio.io/interceptionMode annotation is injected to the pod.
+		// There are three interception modes: REDIRECT, TPROXY and NONE.
+		// It only makes sense to adjust the security context of the storage initializer if REDIRECT mode is
+		// observed, because the Istio sidecar would be injected and traffic would be sent to it, but the
+		// sidecar won't be running at PodInitialization phase.
+		// The TPROXY mode can indicate that Istio Ambient is enabled. The Waypoint proxy would already be running and
+		// captured traffic can go through.
+		// The TPROXY mode can also be set by the user. If this is the case, it is not possible to infer the setup.
+		// Lastly, if interception mode is NONE the traffic is not being captured. This is an advanced mode, and
+		// it is not possible to infer the setup.
+		istioInterceptionMode := pod.Annotations[constants.IstioInterceptionModeAnnotation]
+		if istioInterceptionMode != constants.IstioInterceptModeRedirect {
+			return nil
+		}
+
+		// The storage initializer can only run smoothly when running with the same UID as the Istio sidecar.
+		// First, find the name of the Istio sidecar container. This is found in a status annotation injected
+		// by Istio. If there is no Istio sidecar status annotation, assume that the pod does
+		// not have a sidecar and leave untouched the security context.
+		istioStatus, istioStatusOk := pod.Annotations[constants.IstioSidecarStatusAnnotation]
+		if !istioStatusOk {
+			return nil
+		}
+
+		// Decode the Istio status JSON document
+		var istioStatusDecoded interface{}
+		if err := json.Unmarshal([]byte(istioStatus), &istioStatusDecoded); err != nil {
+			return err
+		}
+
+		// Get the Istio sidecar container name.
+		istioSidecarContainerName := ""
+		istioStatusMap := istioStatusDecoded.(map[string]interface{})
+		if istioContainers, istioContainersOk := istioStatusMap["containers"].([]interface{}); istioContainersOk {
+			if len(istioContainers) > 0 {
+				istioSidecarContainerName = istioContainers[0].(string)
+			}
+		}
+
+		// If there is no Istio sidecar, it is not possible to set any UID.
+		if len(istioSidecarContainerName) == 0 {
+			return nil
+		}
+
+		// Find the Istio sidecar container in the pod.
+		var istioSidecarContainer *v1.Container
+		for idx, container := range pod.Spec.Containers {
+			if container.Name == istioSidecarContainerName {
+				istioSidecarContainer = &pod.Spec.Containers[idx]
+				break
+			}
+		}
+
+		// Set the UserID of the storage initializer to the same as the Istio sidecar
+		if istioSidecarContainer != nil {
+			if storageInitializerContainer.SecurityContext == nil {
+				storageInitializerContainer.SecurityContext = &v1.SecurityContext{}
+			}
+			if istioSidecarContainer.SecurityContext == nil || istioSidecarContainer.SecurityContext.RunAsUser == nil {
+				// If the Istio sidecar does not explicitly have a UID set, use 1337 which is the
+				// UID hardcoded in Istio. This would require privileges to run with AnyUID, which should
+				// be OK because, otherwise, the Istio sidecar also would not work correctly.
+				storageInitializerContainer.SecurityContext.RunAsUser = ptr.Int64(constants.DefaultIstioSidecarUID)
+			} else {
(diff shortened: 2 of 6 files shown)


pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -425,16 +425,124 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
-			initContainer.SecurityContext.RunAsUser = ptr.Int64(uid)
+			if storageInitializerContainer.SecurityContext == nil {
+				storageInitializerContainer.SecurityContext = &v1.SecurityContext{}
+			}
+			storageInitializerContainer.SecurityContext.RunAsUser = ptr.Int64(uid)
+		}
+	} else {
+		// When Istio CNI is enabled, a sidecar.istio.io/interceptionMode annotation
+		// is injected to the pod.
+		istioInterceptionMode, istioCniEnabled := pod.Annotations[constants.IstioInterceptionModeAnnotation]
```
skonto: Is there some other way to detect Istio integration eg. config?

skonto: Also would `k8s.v1.cni.cncf.io/network-status` be a better option here?

israel-hdez: I don't see this annotation or label in `minikube`

israel-hdez: > Is there some other way to detect Istio integration eg. config?

I compared the install YAML with istio-cni enabled and disabled. The most relevant thing I'm able to see is the presence of an `istio-cni` ConfigMap in istio namespace. Not sure if that is really better.

FWIW, I chose this annotation based on this Istio source: https://github.com/istio/istio/blob/f5b75167e06b601ffe50a6e9b5ee2ada10b0c2cf/manifests/charts/istiod-remote/files/injection-template.yaml#L48-L52. The annotations is always there if CNI is enabled.

In a line before it, there is the `k8s.v1.cni.cncf.io/networks` which is related, but that one has a conditional...

skonto: Regarding `k8s.v1.cni.cncf.io/network-status` I was looking at OCP but it seems OCP does not support chained mode so thus the annotation.

>  The annotations is always there if CNI is enabled.

If the annotation is there, is CNI always on?

I also found this, I think that should work given that we check the pod at the webhook side:
https://github.com/istio/istio/blob/26cf55f5d6f576bf8fad436b61c7ec1240fb671d/pkg/config/analysis/analyzers/maturity/maturity.go#L134. 

So with cni instead of the istio-init container, you should get an [istio-validation](https://istio.io/latest/docs/setup/additional-setup/cni/#race-condition--mitigation) one. Verified it on minikube and [OCP](https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-vs-community.html#ossm-cni_ossm-vs-istio) that with cni you get no istio-init container. On OCP btw there is no init container injected at all.

israel-hdez: > If the annotation is there, is CNI always on?
> 

From the template, yes, that's what I understand.

> I also found this, I think that should work given that we check the pod at the webhook side:
> 
> https://github.com/istio/istio/blob/26cf55f5d6f576bf8fad436b61c7ec1240fb671d/pkg/config/analysis/analyzers/maturity/maturity.go#L134. 
> 
> So with cni instead of the istio-init container, you should get an [istio-validation](https://istio.io/latest/docs/setup/additional-setup/cni/#race-condition--mitigation) one. Verified it on minikube and [OCP](https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-vs-community.html#ossm-cni_ossm-vs-istio) that with cni you get no istio-init container. On OCP btw there is no init container injected at all.
> 

I'll check. If it works in both cases, I can change it to align to this code. Yet, as noted in the comments, the check to the interception mode is still needed (to rule out Ambient mode).

yuzisun:
@israel-hdez Have you looked at the `ClusterStorageContainer` where you can now customize the storage initializer container? you can update the security context there. This is quite Istio specific and KServe now does not require Istio to run inference services.

israel-hdez:
> @israel-hdez Have you looked at the `ClusterStorageContainer` where you can now customize the storage initializer container?

@yuzisun I'm aware of it, although I'm not sure if it will fully cover the need. Perhaps for standard Istio `ClusterStorageContainer` can be enough, because the UserID is well known.

The problem is that in OpenShift the sidecar runs without a predictable UID and can change per namespace. We would need controller support for that case, since the UID can only be determined at runtime.

pkg/webhook/admission/pod/storage_initializer_injector.go
israel-hdez: Unfortunately, the `istio-validation` container is missing in Maistra.
So, current approach of checking the interception mode annotations seems to be the most reliable way.

israel-hdez: Oh... Although after re-reading the suggested code, it checks for presence of `istio-init`... let me re-test.

israel-hdez: @skonto OK, I changed the code to use your suggested approach.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -481,16 +481,127 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
...
+		istioInterceptionMode := pod.Annotations[constants.IstioInterceptionModeAnnotation]
+		if istioInterceptionMode != constants.IstioInterceptModeRedirect {
+			return nil
+		}
+
+		// The storage initializer can only run smoothly when running with the same UID as the Istio sidecar.
+		// First, find the name of the Istio sidecar container. This is found in a status annotation injected
+		// by Istio. If there is no Istio sidecar status annotation, assume that the pod does
+		// not have a sidecar and leave untouched the security context.
+		istioStatus, istioStatusOk := pod.Annotations[constants.IstioSidecarStatusAnnotation]
```
yuzisun, line 545: Is this true for all the supported Istio versions ?

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -2852,3 +2852,617 @@ func TestGetContainerWithName(t *testing.T) {
...
+					},
+				},
+			},
+		},
+		"StorageInitializerUidNotSetIfIstioStatusBlank": {
+			original: &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						constants.StorageInitializerSourceUriInternalAnnotationKey: "gs://foo",
+						constants.IstioSidecarStatusAnnotation:                     "{}",
```
yuzisun, line 3232: It seems like all test cases have status annotations, can we add a test case when this annotation does not present ?

config/webhook/manifests.yaml
```diff
@@ -34,6 +34,7 @@ webhooks:
     failurePolicy: Fail
     name: inferenceservice.kserve-webhook-server.pod-mutator
     sideEffects: None
+    reinvocationPolicy: IfNeeded
```
yuzisun, line 37: Can we also update the helm chart configuration here
https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/webhookconfiguration.yaml#L36

israel-hdez: Yes. I fixed.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
israel-hdez: I added two: one when it is fully absent, and another one when it is empty.

pkg/webhook/admission/pod/storage_initializer_injector.go
israel-hdez: Yes, this annotation has been available for quite a while.

israel-hdez:
@luksa, I wonder if you could take a look at this PR. We will appreciate a review from an Istio/CNI expert to know if all this makes sense.

@yuzisun Kindly reminder that I fixed your feedback.

luksa:
@israel-hdez lgtm!

israel-hdez:
Thanks @yuzisun and @luksa !

PR #3319, opened 2023-12-19
ci: publish helm chart to ghcr

davidspek:
**What this PR does / why we need it**:
This PR extends the CI to push the helm charts to GHCR so that users can easily install KServe using them.
Examples of the charts published from this PR can be seen here:
https://github.com/davidspek/kserve/pkgs/container/charts%2Fkserve
https://github.com/davidspek/kserve/pkgs/container/charts%2Fkserve-crd

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
ci: publish helm charts to GHCR
```

PR #3271, opened 2023-11-27
Upgraded poetry version to 1.7.1

andyi2it:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

**Feature/Issue validation/testing**:

**Release note**:

```release-note
NONE
```

PR #3325, opened 2023-12-21
chore: Prepare v0.12.0-rc0 release

terrytangyuan:
This PR follows the documented process in https://github.com/kserve/kserve/blob/master/release/RELEASE_PROCESS_v2.md to prepare for release v0.12.0-rc0.


charts/kserve-resources/values.yaml
```diff
@@ -1,5 +1,5 @@
 kserve:
-  version: &defaultVersion v0.11.1
+  version: &defaultVersion v0.12.0-rc0
   modelmeshVersion: &defaultModelMeshVersion v0.11.0
```
terrytangyuan: Do we want to bump this `modelmeshVersion` as well? Seems like a newer version is available: https://github.com/kserve/modelmesh/releases/tag/v0.11.1

python/VERSION
```diff
@@ -1 +1 @@
-0.11.1
+0.12.0rc0
```
terrytangyuan, line 1: I removed the dash since this needs to comply with the Python package versioning requirement: https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers

sivanantha321:
@terrytangyuan You have to run `./hack/python-release.sh` to update kserve version for model servers. Seems like it is missing from the release process

terrytangyuan:
> @terrytangyuan You have to run `./hack/python-release.sh` to update kserve version for model servers. Seems like it is missing from the release process

Yep, I mentioned that in https://github.com/kserve/kserve/pull/3326#discussion_r1433405873 and Slack. I'll run it and update the PR.

charts/kserve-resources/values.yaml
sivanantha321: yes

python/kserve/Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: test
 
 dev_install:
-	poetry install --with test,dev --extras "storage"
```
terrytangyuan: This is needed to fix build

test/scripts/gh-actions/setup-poetry.sh
```diff
@@ -20,7 +20,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-export POETRY_VERSION=1.4.0
+export POETRY_VERSION=1.7.1
```
sivanantha321: @terrytangyuan We already have a PR for [upgrading poetry](https://github.com/kserve/kserve/pull/3271). I think it is better not to include it here. WDYT ?

terrytangyuan: I had some issues locally with `POETRY_VERSION=1.4.0` for some reason on my Mac. This is the only workaround I found to run `./hack/python-release.sh` successfully.

python/kserve/Makefile
sivanantha321: This is also addressed in [this PR](https://github.com/kserve/kserve/pull/3271). How about merging that PR first ?

terrytangyuan: I had to fix it in my PR to unblock the release. Otherwise, I cannot pass CI build.

terrytangyuan: Happy to revert my changes and rebase if it's merged first.

charts/kserve-resources/values.yaml
terrytangyuan: Bumped to v0.11.1

davidspek:
Do you think we can get https://github.com/kserve/kserve/pull/3319 merged before cutting the next release branch? This way the helm chart will get published to GHCR and we don't need to cherrypick that onto the release branch as well.

terrytangyuan:
> Do you think we can get #3319 merged before cutting the next release branch? This way the helm chart will get published to GHCR and we don't need to cherrypick that onto the release branch as well.

I'll defer that to @yuzisun and others to decide. There are other issues tracked in https://github.com/orgs/kserve/projects/12 that might be worth taking a look as well.

PR #3341, opened 2024-01-03
chore: Remove unused timeout annotation and flag in batcher

terrytangyuan:
This flag is currently missing but batch injector is injecting annotation `/batcher-timeout` as well as appending `--timeout` flag to batcher container args.

### DIFF
--- a/pkg/batcher/handler.go
+++ b/pkg/batcher/handler.go
@@ -30,9 +30,9 @@ import (
 )
 
 const (
+	SleepTime    = time.Microsecond * 100
 	MaxBatchSize = 32
 	MaxLatency   = 5000
-	Timeout      = 1
 )
 
 type Request struct {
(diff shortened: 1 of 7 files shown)


pkg/batcher/handler.go
```diff
@@ -173,7 +174,7 @@ func (handler *BatchHandler) batch() {
 				index,
 			}
 			handler.batcherInfo.CurrentInputLen = len(handler.batcherInfo.Instances)
-		case <-time.After(SleepTime):
+		case <-time.After(time.Duration(handler.Timeout)):
```
yuzisun: This is not actually a timeout for batching, it is a tick time for the select loop. `maxLatency` is the batcher timeout.

terrytangyuan: I had this impression but wasn't sure what the original intention was. Should I remove `BatcherTimeoutInternalAnnotationKey` annotation and the use of `BatcherArgumentTimeout` flag since they are not used and redundant to the existing `maxLatency` flag?

terrytangyuan: I have updated the PR to remove these instead. Let me know if I have missed anything here.

Issue #3342, opened 2024-01-03
Automation of release process


terrytangyuan:
The current release process is pretty manual (e.g. bumping versions in many different files and running various scripts) and we could benefit from adding some automation around it. https://github.com/kserve/kserve/blob/master/release/RELEASE_PROCESS_v2.md

/assign

PR #3353, opened 2024-01-08
Change version for helm charts in README

gawsoftpl:
Update versions for charts in README files.

### DIFF
--- a/charts/kserve-resources/README.md
+++ b/charts/kserve-resources/README.md
@@ -5,11 +5,11 @@
 First install the `kserve-crd` chart
 
 ```console
-helm install kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version 0.11.2
+helm install kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version v0.12.0-rc0
 ```
 
 Then install this chart
 
 ```console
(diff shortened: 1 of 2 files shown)


charts/kserve-resources/README.md
```diff
@@ -5,11 +5,11 @@
 First install the `kserve-crd` chart
 
 ```console
-helm install kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version 0.11.2
+helm install kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version v0.12.0-rc0
```
yuzisun, line 8: This would change every time we release, does it default to the latest version ?

terrytangyuan (commented):
We should automatically bump this version as well as part of https://github.com/kserve/kserve/issues/3342

Also there is a tool to automatically generate docs for helm charts.

yuzisun:
> We should automatically bump this version as well as part of #3342
> 
> Also there is a tool to automatically generate docs for helm charts.

We would need to generate the doc if we want to make helm as official installation method

terrytangyuan:
This is fine to fix the current documentation. I created https://github.com/kserve/kserve/issues/3356 to track the docs generation separately.

/lgtm

PR #3364, opened 2024-01-15
build: Skip already downloaded Istio in quick_install.sh

terrytangyuan:
For cases where it's necessary to run `hack/quick_install.sh` several times due to transient errors, re-running the script would download Istio again. This would avoid having to download Istio again even though it's already downloaded.

### DIFF
--- a/hack/quick_install.sh
+++ b/hack/quick_install.sh
@@ -31,6 +31,7 @@ while getopts ":hsr" option; do
 done
 
 export ISTIO_VERSION=1.19.4
+export ISTIO_DIR=istio-${ISTIO_VERSION}
 export KNATIVE_SERVING_VERSION=knative-v1.10.1
 export KNATIVE_ISTIO_VERSION=knative-v1.10.0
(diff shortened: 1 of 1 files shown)


israel-hdez (approved):
LGTM

hack/quick_install.sh
```diff
@@ -31,14 +31,15 @@ while getopts ":hsr" option; do
...
 export KSERVE_VERSION=v0.12.0-rc1
 export CERT_MANAGER_VERSION=v1.3.0
 export SCRIPT_DIR="$( dirname -- "${BASH_SOURCE[0]}" )"
 
 cleanup(){
-  rm -rf istio-${ISTIO_VERSION}
+  rm -rf ${ISTIO_DIR}
   rm -rf deploy-config-patch.yaml
 }
 trap cleanup EXIT
```
sivanantha321, line 44: won't the cleanup function remove the directory during exit even if the script encounters errors ?

terrytangyuan: Good catch. I moved it to after Istio has been installed successfully

PR #3366, opened 2024-01-16
Adds optional labels and annotations to the controller

guitouni:

Fixes #

Makes it possible to define labels for the deployment and the pods of the kserve-controller-manager.

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Optional labels and annotations can be defined for the deployment and the pods of the kserve-controller-manager.
```

### DIFF
--- a/charts/kserve-resources/templates/deployment.yaml
+++ b/charts/kserve-resources/templates/deployment.yaml
@@ -12,6 +12,9 @@ metadata:
     {{- end }}
   annotations:
     prometheus.io/scrape: 'true'
+    {{- if .Values.kserve.controller.annotations }}
+    {{- toYaml .Values.kserve.controller.annotations | nindent 4 }}
+    {{- end }}
 spec:
   selector:
     matchLabels:
@@ -27,6 +30,9 @@ spec:
         {{- end }}
       annotations:
         kubectl.kubernetes.io/default-container: manager
+        {{- if .Values.kserve.controller.podAnnotations }}
+        {{- toYaml .Values.kserve.controller.podAnnotations | nindent 8 }}
+        {{- end }}
     spec:
       serviceAccountName: kserve-controller-manager
       {{- with .Values.kserve.controller.nodeSelector }}
--- a/charts/kserve-resources/templates/deployment.yaml
+++ b/charts/kserve-resources/templates/deployment.yaml
@@ -7,6 +7,9 @@ metadata:
   labels:
     control-plane: kserve-controller-manager
     controller-tools.k8s.io: "1.0"
+    {{- if .Values.kserve.controller.labels }}
+    {{- toYaml .Values.kserve.controller.labels | nindent 4 }}
+    {{- end }}
   annotations:
     prometheus.io/scrape: 'true'
 spec:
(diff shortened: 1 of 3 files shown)


charts/kserve-resources/templates/deployment.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
   labels:
     control-plane: kserve-controller-manager
     controller-tools.k8s.io: "1.0"
+    {{- if .Values.kserve.controller.labels }}
```
spolti, line 10: from where this value is coming from?
Maybe we can document it to the users?

guitouni: It will be coming from `values.yaml` (see below), and can be overridden by the users.
I actually followed the same pattern as all the other existing parameter. It seems pretty straightforward to me.
Let me know where I should document it if you think otherwise. Happy to do it.

terrytangyuan (commented):
My concern is there are endless options for users to customize, including labels, annotations, etc.

A better approach would be to use kustomize + helm: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md

guitouni:
> My concern is there are endless options for users to customize, including labels, annotations, etc.
> 
> A better approach would be to use kustomize + helm: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md

It's a pattern many projects as it's a very common need. A few examples:

- [Superset](https://github.com/apache/superset/blob/master/helm/superset/templates/deployment.yaml#L69)
- [Istio](https://github.com/hanxiaop/istio/blob/743a49b5636ac4518b5ce387facf925c80994d48/manifests/charts/istio-operator/values.yaml#L51)
- [Grafana](https://github.com/grafana/helm-charts/blob/e733d049a1257d421dd14e8b4f455acdb98c121c/charts/agent-operator/values.yaml#L14)

I don't think this issue alone will justify introducing Kustomize in the stack of many users.
For example in our case, we basically cloned the project to bypass this limitation.

yuzisun:
> > My concern is there are endless options for users to customize, including labels, annotations, etc.
> > A better approach would be to use kustomize + helm: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md
> 
> It's a pattern many projects as it's a very common need. A few examples:
> 
> * [Superset](https://github.com/apache/superset/blob/master/helm/superset/templates/deployment.yaml#L69)
> * [Istio](https://github.com/hanxiaop/istio/blob/743a49b5636ac4518b5ce387facf925c80994d48/manifests/charts/istio-operator/values.yaml#L51)
> * [Grafana](https://github.com/grafana/helm-charts/blob/e733d049a1257d421dd14e8b4f455acdb98c121c/charts/agent-operator/values.yaml#L14)
> 
> I don't think this issue alone will justify introducing Kustomize in the stack of many users. For example in our case, we basically cloned the project to bypass this limitation.

I agree that's common case that people need custom labels and annotations.

yuzisun (commented):
/lgtm
/approve

yuzisun (approved):

PR #3367, opened 2024-01-16
Gracefully shutdown the router server

### LINKED ISSUE #3360: Do graceful shutdown of http server in Inference Graph router
/kind bug

While active traffic is being received, if router pods are getting re-stacked (due to any changes say minReplica change), we observed 5xx for in-flight requests.

First suspicion is that we are not handling SIGTERM signal and not shutting down http server gracefully in [router](https://github.com/kserve/kserve/blob/master/cmd/router/main.go).

Graceful shutdown of http server and in-flight requests should not have been impacted.

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

sivanantha321:

Fixes #3360

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```


cmd/router/main.go
```diff
@@ -330,9 +332,11 @@ func compilePatterns(patterns []string) ([]*regexp.Regexp, error) {
 var (
 	jsonGraph              = flag.String("graph-json", "", "serialized json graph def")
 	compiledHeaderPatterns []*regexp.Regexp
+	timeout                int64
 )
 
 func main() {
+	flag.Int64Var(&timeout, "termination-grace-period", 300, "wait timeout period for in-flight requests during server shutdown")
```
sivanantha321: @yuzisun / @rachitchauhan43 I think we need to introduce a new field for specifying the timeout period. WDYT ?

rachitchauhan43: In serverless mode, i am seeing, due to knative, `terminationGracePeriodSeconds: 300` is getting added to router pods.
But yes, I agree, just like we let users control this in ISVCs, we should let users control this in IG too.

rachitchauhan43: @yuzisun : We are on KServe 0.10.1 and observing this issue. If this is the fix, I would need your help in back-porting this to 0.10.1 or better if we could release 0.10.1.1 for this hotfix.
I f we do not support semver upto hotfixes, then it can be 0.10.3

cmd/router/main.go
```diff
@@ -351,11 +355,47 @@ func main() {
...
 	http.HandleFunc("/", graphHandler)
 
-	err = http.ListenAndServe(":8080", nil)
-	if err != nil {
-		log.Error(err, "failed to listen on 8080")
-		os.Exit(1)
+	go func() {
+		err = server.ListenAndServe()
+		if err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Error(err, fmt.Sprintf("Failed to listen on address %v", server.Addr))
```
rachitchauhan43:
```suggestion
			log.Error(err, fmt.Sprintf("Failed to start server on address %v", server.Addr))
```

cmd/router/main.go
```diff
@@ -351,11 +365,43 @@ func main() {
...
+	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
+
+	sig := <-signalChan
+	log.Info("Received shutdown signal", "signal", sig)
+	// Fail the readiness probe
+	isShuttingDown = true
+	log.Info("Sleeping %v to allow K8s propagation of non-ready state", drainSleepDuration)
+	// Sleep to give networking a little bit more time to remove the pod
+	// from its configuration and propagate that to all loadbalancers and nodes.
+	time.Sleep(drainSleepDuration)
```
sivanantha321, line 409: @rachitchauhan43 / @yuzisun I have added readiness probe for the router. The readiness probe will start failing after receiving the shutdown signal. So the requests should be no longer routed to the pod. The router will wait for 30 seconds before shutting down the server to accommodate the scenario mentioned in the issue #3375. It is worth noting that knative serving already adds a preStop hook to the container which essentially delays the termination to process the accepted requests. https://github.com/knative/serving/blob/51b0337fa29d413302b75acfa13bbd8c765e8fea/pkg/queue/constants.go#L24-L29
Screenshot from 2024-01-27 23-12-25
@rachitchauhan43 Would you mind testing it with these changes ?

cmd/router/main.go
```diff
@@ -351,11 +365,43 @@ func main() {
...
+	// Shut down the server gracefully
+	if err := server.Shutdown(context.Background()); err != nil {
+		log.Error(err, "Failed to shutdown the server gracefully")
```
bmopuri, line 412: If graceful shutdown fails, can we do force close something like this !


```
// Shut down the server forcefully if context deadline exceeded.
		if errors.Is(err, context.DeadlineExceeded) {
			log.Info("Shutting down the server forcefully ...")
			if err := server.Close(); err != nil {
				log.Error(err, "Failed to shutdown the server forcefully")
				os.Exit(1)
			}
			log.Info("Server forcefully shutdown")
		}
```

pkg/controller/v1alpha1/inferencegraph/controller_test.go
```diff
@@ -163,6 +164,23 @@ var _ = Describe("Inference Graph controller test", func() {
 													v1.ResourceMemory: resource.MustParse("100Mi"),
 												},
 											},
+											ReadinessProbe: &v1.Probe{
```
spolti: This very same code is used in different places, maybe we can move to a utils package or something a reuse it.

cmd/router/main.go
sivanantha321: We are using background context which will not timeout, so we don't need this logic.

pkg/controller/v1alpha1/inferencegraph/controller_test.go
sivanantha321: This is used in tests for assert against the actual and expected knative service. I believe it is better to leave it as it is. WDYT ?

spolti: I meant the probes are the same for all the places to instantiate it.
Maybe move this part to a common env or func.

sivanantha321: done

spolti (approved):

PR #3374, opened 2024-01-19
feat: upgrade to new fastapi, update models to handle both pydantic v…

### LINKED ISSUE #3373: Support pydantic 2
/kind feature

The kserve package should support Pydantic 2 and newer FastAPI

timothyjlaurent:
…1 an v2, add matrix conditions for pydantic_v1 and pydantic_v2

This work builds on this PR https://github.com/kserve/kserve/pull/3273 that adds pydantic 2 support to the kserve package.

The following is copied from that PR:
* Follows Optional fields guidance for pydantic v2 [ref](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields)
*Overwrites protected_namespace in pydantic v2 to get rid of warning logs about namespace conflicts in pydantic [ref](https://docs.pydantic.dev/latest/api/config/#protected-namespaces). An example is shown as follows
```
E   UserWarning: Field "model_name" has conflict with protected namespace "model_".
E
E   You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
```

In addition to handling those incompatibilities it also allows for both pydantic 1.x and 2.x versions.

To accomplish this:
- FastAPI is upgraded to the latest version. This version supports both pydantic 1.x and 2.x. As such, projects that add kserve as a dependency will also be able to use both pydantic versions.
- inspired by how https://github.com/docarray/docarray added support for v1 and v2 pydantic an `is_pydantic_2` flag is used to change the underlying pydantic implementations
- each pydantic model that has incompatible syntax, ie nested `Config` class in retains the class if it is pydantic 1 and uses the `model_config` syntax if it is pydantic 2. Keys are changed where needed, eg `schema_extra` -> `json_schema_extra`.
- `protected_namespace=()` is set on the `InferenceResponse` model
- to test, again inspired by docarray's CI, a new matrix component `pydantic-version` is added to CI. As the default version will be pydantic 2, in the case that the version is `pydantic_1` then the make target `install_pydantic_1` is executed which downgrades pydantic to v1.

Fixes #3373 

**Type of changes**

- [ ] Added matrix to test both pydantic 1 and pydantic 2

- Logs

**Checklist**:

**Release note**:

```release-note
Upgrades FastAPI version and now supports Pydantic v1.x and v2.x. This allows dependent projects to upgrade their Pydantic version as needed.
```


.github/workflows/python-test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
+        pydantic-version: ["pydantic_1", "pydantic_2"]
```
terrytangyuan: It's great to see this coverage but I feel like this could significantly increase the build time since we have to test the combo of Python and Pydantic versions.

timothyjlaurent: Ah good point - especially since we don't need to test pydantic-v1 on all of the other python packages.

I can work on making another set of steps for testing the pydantic v1 functionality.

timothyjlaurent: I've added steps for testing this in the main build job and removed the extra matrix arguments

python/kserve/pyproject.toml
```diff
@@ -35,7 +35,7 @@ exclude = ["kserve/storage/test"]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
 uvicorn = { version = "^0.19.0", extras = ["standard"] }
-fastapi = "^0.95.0"
+fastapi = "^0.109.0"
 cloudevents = "^1.6.2"
 six = "^1.16.0"
```
sivanantha321, line 40: I think it is better to explicitly add pydantic dependency here.

sivanantha321:
@timothyjlaurent Can you help rebase the PR and sign off the commits ?

python/kserve/pyproject.toml
sivanantha321: Model servers depends on kserve sdk. So to propagate the version bump, you have to update the lockfiles of serving runtimes. Running `hack/python-releash.sh` would help.

sivanantha321: One more thing, I recently found out that fastapi 0.109.0 breaks ray serve. https://github.com/ray-project/ray/pull/42740. So let's keep it as it is for now.

timothyjlaurent: Well to upgrade pydantic to v2 we need to upgrade to at least 0.100.0.

https://github.com/tiangolo/fastapi/releases/tag/0.100.0 -- it's probably best to increment it maximally so I'll set it to 0.108.0 as was done in the linked PR.

timothyjlaurent: @sivanantha321 I updated the lockfiles with the serving runtimes, added the pydantic 2 syntax to the new models, and added an explicit pydantic dependency.

timothyjlaurent:
@sivanantha321 I've rebased and signed all the commits, can you take another look?

sivanantha321:
@yuzisun Can you help trigger the tests ?

timothyjlaurent:
We had a small bug in the last test and then I somehow borked my rebase -- I've cleaned up the commits on a fresh branch and think it should pass now.

sivanantha321:
@timothyjlaurent I don't see a updated version of pydantic in poetry.lock file ? I think we can upgrade the pydantic version to >2.0.0 for model servers.

timothyjlaurent:
@sivanantha321 Ah this is because we hack/python-release runs with --no-update and the sdk is compatible with pydantic v1 or v2.

I can either run it to update all packages, or I could add pydantic as an explicit dep to each model server. What do you think?

sivanantha321:
> @sivanantha321 Ah this is because we hack/python-release runs with --no-update and the sdk is compatible with pydantic v1 or v2.
> 
> 
> 
> I can either run it to update all packages, or I could add pydantic as an explicit dep to each model server. What do you think?

You can update the packages as long as it does not breaks anything

timothyjlaurent:
@sivanantha321 I've updated the deps.

yuzisun:
@timothyjlaurent Thanks for upgrading pydantic! Can you help take a look a look at the testing failures?
https://github.com/kserve/kserve/actions/runs/7993801929/job/21947239919?pr=3374

python/kserve/kserve/protocol/rest/v2_datamodels.py
```diff
@@ -111,303 +132,376 @@ class ModelMetadataResponse(BaseModel):
...
+      "text_output" : $string,
+      "model_name" : $string,
+      "model_version" : $string #optional,
+      "details": $Details #optional
+    }
     """
+
     text_output: str
     model_name: str
     model_version: Optional[str]
```
gujingit:
```
model_version: Optional[str] = None
details: Optional[Details] = None
```

python/kserve/pyproject.toml
```diff
@@ -35,7 +35,7 @@ exclude = ["kserve/storage/test"]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
 uvicorn = { version = "^0.19.0", extras = ["standard"] }
```
gujingit: uvicorn = { version = "^0.21.1", extras = ["standard"] }

gujingit: @timothyjlaurent Can you help update uvicorn version to "^0.21.1" ？So that PR #3480 can be closed.

python/kserve/kserve/protocol/rest/v2_datamodels.py
timothyjlaurent: Thanks, I've updated all of the optionals

python/kserve/pyproject.toml
timothyjlaurent: Sure thing, added it and updated all dependent packages

timothyjlaurent:
> @timothyjlaurent Thanks for upgrading pydantic! Can you help take a look a look at the testing failures? https://github.com/kserve/kserve/actions/runs/7993801929/job/21947239919?pr=3374

I took the time to figure out how to properly run the test suite locally -- needs to be done from the ./python directory.

I've updated the tests one notable workaround is that the RayServe test asserted `b'{"name":"TestModel","ready":"True"}'` <- notice the stringified bool. In pydantic V2 we correctly get an actual json bool for the ready param. I added a special case to the test and left comments explaining this behavior.

python/kserve/test/test_server.py
```diff
@@ -381,7 +382,14 @@ def test_list_handler(self, http_server_client):
     def test_health_handler(self, http_server_client):
         resp = http_server_client.get('/v1/models/TestModel')
         assert resp.status_code == 200
-        assert resp.content == b'{"name":"TestModel","ready":"True"}'
+        # for some reason the RayServer responds with the stringified python bool
+        # when run on pydantic < 2 and the bool when run on pydantic >= 2
+        # eg {"name":"TestModel","ready":"True"} vs {"name":"TestModel","ready":true}
```
sivanantha321, line 436: I think it is a different bug altogether. Before moving to fastapi, it seems the [health handler returned bool values](https://github.com/kserve/kserve/blob/21ce62bf4522e63ff6a60435250c5cbd55b12b43/python/kserve/kserve/handlers/health.py#L25-L47), 
but after the migration it started [returning string values because how the type is defined](https://github.com/kserve/kserve/blob/e6eede2ea0e94cbe75e90785877cba33f9014216/python/kserve/kserve/protocol/rest/v1_endpoints.py#L39) ?
@yuzisun

yuzisun: The code seems to be returning bool, as @timothyjlaurent mentioned the issue is for the rayserve case?

timothyjlaurent: This is for the RayServe case.

In the previous assertion with pydantic v1 - we were getting `{"name":"TestModel","ready":"True"}` the string value, and testing for that. Seems like it was a bug that got "blessed" by enshrinining it in a testcase.
With the pyanticv2 we're (correctly) getting a bool value for the `ready` key (not a stringified python string) - This, I believe is the correct behavior we should expect.

yuzisun:
@timothyjlaurent seems like it keeps failing after updating the poetry lock file
https://github.com/kserve/kserve/actions/runs/8087460553/job/22214167194?pr=3374

yuzisun:
Tests passed, Thanks @timothyjlaurent for the great work!

/lgtm
/approve

PR #3387, opened 2024-01-24
align cloudevents/sdk-go dependency

spolti:

chore:  The cloudevents/sdk-go was migrated to v2, however, the v1 is still present in the
        go.mod. It was being used bu the pkg/logger/worker.go file to set a field that was
        not being in use anymore, thus it was removed


spolti:
@yuzisun / @terrytangyuan f.y.i

terrytangyuan (approved):
/lgtm

PR #3389, opened 2024-01-25
Update google-cloud-storage dependecy to >=2.3.0,<3.0.0 and ray dependency to >=2.8.1, <3.0.0

sivanantha321:
**What this PR does / why we need it**:

- Fixes google-cloud-storage deprecated warnings:
```
../.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:121
  /home/ubuntu/go/src/github.com/kserve/kserve/.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:121: DeprecationWarning: pkg_resources is deprecated as an API
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)

../.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2870
../.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2870
  /home/ubuntu/go/src/github.com/kserve/kserve/.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2870: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

../.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2870
  /home/ubuntu/go/src/github.com/kserve/kserve/.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2870: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google.cloud')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

../.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2349
  /home/ubuntu/go/src/github.com/kserve/kserve/.venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2349: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(parent)
```
[...]

PR #3394, opened 2024-01-30
chore: Remove unused version update script for Python SDK

terrytangyuan:
These scripts are not being used.


hack/python-sdk/update_release_version.sh
```diff
@@ -33,6 +33,7 @@ dirs=(
   "python/pmmlserver"
   "python/sklearnserver"
   "python/xgbserver"
+  "python/huggingfaceserver"
 )
```
sivanantha321: This script is no longer used. We can remove this script.

hack/python-sdk/update_release_version_helper.py
```diff
@@ -33,6 +33,7 @@
     "python/pmmlserver/pyproject.toml",
     "python/sklearnserver/pyproject.toml",
     "python/xgbserver/pyproject.toml",
```
sivanantha321: This file too can be removed. Instead we are using [this script](https://github.com/kserve/kserve/blob/master/hack/python-release.sh).

sivanantha321:
@terrytangyuan This is already handled in PR #3389

terrytangyuan:
@sivanantha321 ok I’ll use this PR to remove these unused scripts then

yuzisun:
@andyi2it to confirm if these are good to remove

terrytangyuan:
@yuzisun These are not used anywhere in the codebase.

andyi2it:
> @andyi2it to confirm if these are good to remove

@yuzisun These are good to be removed.

PR #3395, opened 2024-01-30
feat: Add HuggingFace runtime out-of-the-box support

terrytangyuan:
Part of https://github.com/kserve/kserve/issues/3393.

- [x] Tests
- [x] Update Python spec
- [x] Regenerate manifests

PR #3415, opened 2024-02-07
Implement support for vllm as alternative backend

gavrissh:
Support vllm as alternative backend for text generation use-cases

Fixes #

**Type of changes**

   Have not added tests as vllm expects gpu driver as prerequiste

- Logs

**Checklist**:

**Release note**:

```release-note

```


python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,44 +89,58 @@ def infer_task_from_model_architecture(model_config_path: str):
         if task is None:
             raise ValueError(f"Task couldn't be inferred from {architecture}. Please manually set `task` option.")
 
+    @staticmethod
+    def infer_vllm_supported_from_model_architecture(model_config_path: str):
```
terrytangyuan, line 97: This is not used?

terrytangyuan (commented):
Should this be in draft?

python/huggingfaceserver/huggingfaceserver/model.py
gavrissh: Used in the init method - https://github.com/kserve/kserve/pull/3415/files#diff-d02c376ad82bad8a11ceaecb4ea815822123e34b961f29fd5050225af82e62b4R76

terrytangyuan: Sorry I missed it

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -73,7 +70,11 @@ def __init__(self, model_name: str, kwargs,
         self.tokenizer = None
         self.model = None
         self.mapping = None
-        self.engine = AsyncLLMEngine.from_engine_args(engine_args) if _vllm else None
+        self.vllm_engine = None
+        self.use_vllm = kwargs.get('use_vllm', False)
```
johnugeorge: should we make it as a default? 
cc @yuzisun

gavrissh: updated to set True by default

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,44 +89,58 @@ def infer_task_from_model_architecture(model_config_path: str):
...
-            elif self.task == MLTask.text_generation.value:
-                self.model = AutoModelForCausalLM.from_pretrained(model_id_or_path)
-            elif self.task == MLTask.text2text_generation.value:
-                self.model = AutoModelForSeq2SeqLM.from_pretrained(model_id_or_path)
-            else:
-                raise ValueError(f"Unsupported task {self.task}. Please check the supported `task` option.")
-            self.model.eval()
-            self.model.to(self.device)
-            logger.info(f"successfully loaded huggingface model from path {model_id_or_path}")
+        if self.vllm_engine is None:
```
yuzisun: nit: for readability we can do 
```python
if self.vllm_engine:
   return True
```

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -73,7 +70,11 @@ def __init__(self, model_name: str, kwargs,
         self.tokenizer = None
         self.model = None
         self.mapping = None
-        self.engine = AsyncLLMEngine.from_engine_args(engine_args) if _vllm else None
+        self.vllm_engine = None
+        self.use_vllm = kwargs.get('use_vllm', True)
+        if self.use_vllm and self.device == torch.device("cuda"):   # vllm needs gpu
+            if self.infer_vllm_supported_from_model_architecture(self.model_id) is not None:
+                self.vllm_engine = AsyncLLMEngine.from_engine_args(engine_args)
```
yuzisun: This can be moved to the load handler?

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,44 +89,58 @@ def infer_task_from_model_architecture(model_config_path: str):
...
 
+    @staticmethod
+    def infer_vllm_supported_from_model_architecture(model_config_path: str):
+        model_config = AutoConfig.from_pretrained(model_config_path)
+        architecture = model_config.architectures[0]
+        model_cls = ModelRegistry.load_model_cls(architecture)
+        if model_cls is not None:
+            return model_cls
+
+        logger.info("vllm unsupported model")
```
yuzisun:
```suggestion
        if model_cls is None:
           logger.info("vllm unsupported model")
        return model_cls
```

        
```

File: python/huggingfaceserver/huggingfaceserver/model.py (hunk 4 above)
Comment (gavrissh) [reply python/huggingfaceserver/huggingfaceserver/model.py]: Yes, I have made this change in latest commit

File: python/huggingfaceserver/huggingfaceserver/model.py (hunk 5 above)
Comment (gavrissh) [reply python/huggingfaceserver/huggingfaceserver/model.py]: Updated with the suggestion

File: python/huggingfaceserver/pyproject.toml [hunk 6]
```diff
@@ -12,9 +12,9 @@ packages = [
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
 kserve = { path = "../kserve", extras = ["storage"], develop = true }
-transformers = "~4.36.0"
+transformers = "~4.37.0"
 torch = "~2.1.2"
-vllm = { version = "^0.2.6", optional = true }
+vllm = "~0.2.7"
```
yuzisun: We can still keep it optional and install it as default in the dockerfile as the downside of doing this is that I can't develop on my Mac

gavrissh: In this case, will the baked image have vllm installed or some manual intervention is needed?

I had also removed code related to ImportError as this was mandatory. Would need to rollback those changes right?

yuzisun: in the dockerfile we can install vllm,  but would like leave as an option to disable vllm for local development. Also you won’t pass all the tests as CI does not have gpu.

gavrissh: I have implemented the above suggestions.

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -160,10 +180,13 @@ async def generate(self, generate_request: GenerateRequest, headers: Dict[str, s
         parameters = generate_request.parameters
         prompt = generate_request.text_input
         request_id = str(uuid.uuid4())
-        if _vllm:
+        if self.vllm_engine is not None:
+            if parameters is None:
+                parameters = {}
```
terrytangyuan: Merge this with line 180?

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -196,6 +219,9 @@ async def generate(self, generate_request: GenerateRequest, headers: Dict[str, s
 
     async def predict(self, input_batch: Union[BatchEncoding, InferRequest], context: Dict[str, Any] = None) \
             -> Union[Tensor, InferResponse]:
+        if self.vllm_engine is not None:
+            raise InferenceError("Use /generate endpoint for vllm runtime")
```
terrytangyuan: Move the error string to a constant since it appears multiple times

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -35,13 +35,14 @@ def list_of_strings(arg):
...
-parser.add_argument('--add_special_tokens', type=bool, default=True,
-                    help='the sequences will be encoded with the special tokens relative to their model')
+parser.add_argument('--disable_lower_case', action='store_true',
+                    help='do not use lower case for the tokenizer')
+parser.add_argument('--disable_special_tokens', action='store_true',
+                    help='the sequences will not be encoded with the special tokens relative to their model')
 parser.add_argument('--tensor_input_names', type=list_of_strings, default=None,
                     help='the tensor input names passed to the model')
 parser.add_argument('--task', required=False, help="The ML task name")
+parser.add_argument('--disable_vllm', action='store_true', help="Do not use vllm as the default runtime")
```
terrytangyuan, line 45: Missing default here

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -50,18 +51,23 @@ def list_of_strings(arg):
...
     _vllm = False
+
 args, _ = parser.parse_known_args()
 
 if __name__ == "__main__":
-    engine_args = AsyncEngineArgs.from_cli_args(args) if _vllm else None
+    engine_args = None
+    if _vllm and not args.disable_vllm:
+        if args.model != args.model_id:   # vllm sets default model
+            args.model = args.model_id
```
terrytangyuan: We can just assign directly to model_id without checking

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -65,15 +68,16 @@ def __init__(self, model_name: str, kwargs,
             assert world_size == tp_degree, f"TP degree ({tp_degree}) doesn't match available GPUs ({world_size})"
         self.model_id = kwargs.get('model_id', None)
         self.model_dir = kwargs.get('model_dir', None)
-        self.do_lower_case = kwargs.get('do_lower_case', True)
-        self.add_special_tokens = kwargs.get('add_special_tokens', True)
+        self.do_lower_case = not kwargs.get('disable_lower_case', False)
+        self.add_special_tokens = not kwargs.get('disable_special_tokens', False)
```
terrytangyuan, line 72: Add default values in parser args in `main.py` instead?

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,7 +92,22 @@ def infer_task_from_model_architecture(model_config_path: str):
...
             raise ValueError(f"Task couldn't be inferred from {architecture}. Please manually set `task` option.")
 
-    def load(self) -> bool:
+    @staticmethod
+    def infer_vllm_supported_from_model_architecture(model_config_path: str):
+        model_config = AutoConfig.from_pretrained(model_config_path)
+        architecture = model_config.architectures[0]
+        model_cls = ModelRegistry.load_model_cls(architecture)
+        if model_cls is None:
+            logger.info("vllm unsupported model")
```
terrytangyuan:
```suggestion
            logger.info("not a supported model by vLLM")
```

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,7 +92,22 @@ def infer_task_from_model_architecture(model_config_path: str):
...
+        return model_cls
+
+    def load(self, engine_args=None) -> bool:
+        if self.use_vllm and self.device == torch.device("cuda"):   # vllm needs gpu
+            if self.infer_vllm_supported_from_model_architecture(self.model_id) is not None:
+                self.vllm_engine = AsyncLLMEngine.from_engine_args(engine_args)
+                self.ready = True
+                return self.ready
```
terrytangyuan, line 109: We don't need to return here. We can use `self.ready` directly

python/huggingfaceserver/huggingfaceserver/__main__.py
gavrissh: This is on/off flag argument not an option that takes value.

python/huggingfaceserver/huggingfaceserver/model.py
gavrissh: The default are set here to facilitate tests.

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -35,13 +35,14 @@ def list_of_strings(arg):
                     help='tensor parallel degree')
 parser.add_argument('--max_length', type=int, default=None,
                     help='max sequence length for the tokenizer')
-parser.add_argument('--do_lower_case', type=bool, default=True,
-                    help='do lower case for the tokenizer')
-parser.add_argument('--add_special_tokens', type=bool, default=True,
-                    help='the sequences will be encoded with the special tokens relative to their model')
+parser.add_argument('--disable_lower_case', action='store_true',
+                    help='do not use lower case for the tokenizer')
+parser.add_argument('--disable_special_tokens', action='store_true',
```
yuzisun, line 40: Is there a reference where these should be default to be disabled?

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -50,18 +51,22 @@ def list_of_strings(arg):
...
     predictor_config = PredictorConfig(args.predictor_host, args.predictor_protocol,
                                        args.predictor_use_ssl,
                                        args.predictor_request_timeout_seconds)
     model = HuggingfaceModel(args.model_name,
                              predictor_config=predictor_config,
-                             kwargs=vars(args), engine_args=engine_args)
+                             kwargs=vars(args))
     try:
-        model.load()
+        model.load(engine_args=engine_args)
```
yuzisun: this changes the load signature, is there a reason why can't be passed to the model instead ?

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -88,7 +92,22 @@ def infer_task_from_model_architecture(model_config_path: str):
...
+    def infer_vllm_supported_from_model_architecture(model_config_path: str):
+        model_config = AutoConfig.from_pretrained(model_config_path)
+        architecture = model_config.architectures[0]
+        model_cls = ModelRegistry.load_model_cls(architecture)
+        if model_cls is None:
+            logger.info("not a supported model by vLLM")
+        return model_cls
+
+    def load(self, engine_args=None) -> bool:
+        if self.use_vllm and self.device == torch.device("cuda"):   # vllm needs gpu
```
yuzisun, line 106: Do we need the `use_vllm` if we can determine the list of models architecture vllm can support?

python/huggingfaceserver/huggingfaceserver/__main__.py
yuzisun: Why the default is true? I think we should default to false unless user wants to explicitly turn it off and let it fallback to huggingface API.

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -157,13 +179,14 @@ def preprocess(self, payload: Union[Dict, InferRequest], context: Dict[str, Any]
 
     async def generate(self, generate_request: GenerateRequest, headers: Dict[str, str] = None) \
             -> Union[GenerateResponse, AsyncIterator[Any]]:
-        parameters = generate_request.parameters
+        parameters = generate_request.parameters or {}
         prompt = generate_request.text_input
         request_id = str(uuid.uuid4())
-        if _vllm:
+        if self.vllm_engine is not None:
```
yuzisun:
```suggestion
        if self.vllm_engine:
```

python/huggingfaceserver/huggingfaceserver/__main__.py
gavrissh: I noticed in latest python versions `parser.add_argument(type=bool, ..)` for type bool() does not work as expected. It sets values as string.

https://docs.python.org/3/library/argparse.html#type
https://stackoverflow.com/questions/60999816/argparse-not-parsing-boolean-arguments
https://docs.python.org/3/library/argparse.html#action

To support it I added the same functionality by inverting as expectation is to have `add_special_tokens` as true by default. Hence unless `--disable_special_tokens` is set add_special_tokens will be True. Similarly for vllm and lower_case

python/huggingfaceserver/huggingfaceserver/model.py
gavrissh: Thought going this route to set whether to use vLLM instead of using the global variable `_vllm`
https://github.com/kserve/kserve/pull/3415/files#diff-d02c376ad82bad8a11ceaecb4ea815822123e34b961f29fd5050225af82e62b4R80
`self.use_vllm = not kwargs.get('disable_vllm', False) if _vllm else False`

This is a pre-check if someone explicitly disables vLLM or does not have vLLM package in local.

python/huggingfaceserver/huggingfaceserver/__main__.py
gavrissh: `action='store_true'` provide capability to set bool value if the flag is set. In this case if `--disable_vllm` is set, the value of `arg.disable_vllm` will be true and in all other cases it is false

https://docs.python.org/3/library/argparse.html#action

The experience is if `--disable_vllm` is set only then vLLM will not be used explicitly. if this flag is not set then vLLM to be used, with  huggingface API as fallback

terrytangyuan:
HuggingFace out-of-box support has been added in case you'd like to test it together. @gavrishp https://github.com/kserve/kserve/pull/3395

gavrissh:
> HuggingFace out-of-box support has been added in case you'd like to test it together. @gavrishp #3395

That's great. I'll give it a try

python/huggingfaceserver/README.md
```diff
@@ -92,6 +93,7 @@ spec:
       - --model_id=bert-base-uncased
       - --predictor_protocol=v2
       - --tensor_input_names=input_ids
+      - --disable_vllm
```
yuzisun: I think we probably we do not want to confuse user that they have to set this config to get Bert to work. I think a better example would be the case that vllm is by default and show an example that it can be disabled.

gavrissh: I have added a disable vllm example

python/huggingfaceserver/README.md
```diff
@@ -102,3 +102,66 @@ spec:
...
+        requests:
+          cpu: "6"
+          memory: 24Gi
+          nvidia.com/gpu: "1"
+```
+
+Perform the inference for vllm specific runtime
+
+vllm runtime deployments only support `/generate` endpoint for inference. Kserve python runtime supports all protocols
+Text generate Schema details - https://github.com/kserve/open-inference-protocol/blob/main/specification/protocol/generate_rest.yaml
```
yuzisun:
```suggestion
vllm runtime deployments only support `/generate` endpoint for inference. Please refer to [text generation API schema](https://github.com/kserve/open-inference-protocol/blob/main/specification/protocol/generate_rest.yaml) for more details.
```

yuzisun:
Great work on this @gavrishp !

/lgtm
/approve

python/huggingfaceserver/README.md
lizzzcai: I just trying out the new huggingface runtime and notice this. I feel like a bit werid to have vllm as default runtime (as the PR mentioned vllm as alternative backend).

how about `--backend=vllm`, having `huggingface` as default and vllm as alternative and future can support other backend like `tensorRT`

what do you think? @yuzisun

PR #3417, opened 2024-02-08
ci: Add dependabot configuration for security updates. Fixes #3416

### LINKED ISSUE #3416: Re-enable dependabot
/kind feature

This was discussed in the last community meeting. It might be a good idea to enable dependabot to fix issues faster automatically. 

I saw that dependabot was enabled before but then disabled. Was there any reason for that? Can we re-enable it?

cc @yuzisun @rachitchauhan43

terrytangyuan:
Fixes #3416.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "saturday"
+    # ignore all non-security updates: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit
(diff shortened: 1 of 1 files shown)


spolti:
+1
it will address most of the alerts we have on the python subprojects as well.

Thanks @terrytangyuan

spolti (approved):

ckadner:
KServe is already using Dependabot -- with the GitHub recommended standard configuration

I added a few more details on the issue you created for this:
https://github.com/kserve/kserve/issues/3416#issuecomment-1942912955

There are currently 126 reported potential vulnerabilities, 886 previously addressed/closed ones

Do you intend to add custom configurations that are not possible to configure using the GitHub repo settings? If not, I would prefer to keep using the standard setting, without having to maintain another .yaml file.

terrytangyuan:
Let's discuss in https://github.com/kserve/kserve/issues/3416

PR #3378, opened 2024-01-22
Fixed go lint error using golangci-lint tool.

andyi2it:
**What this PR does / why we need it**:
Fixes go lint errors

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3426

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

**Release note**:

```release-note
NONE
```

PR #3429, opened 2024-02-13
Typos and minor fixes

alpe:
Minor chore. I was reading the file and added some fixes for typos or godoc standard
```

### CODE REVIEW

File: pkg/controller/v1beta1/inferenceservice/controller.go [hunk 1]
```diff
@@ -188,8 +186,8 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		result, err := reconciler.Reconcile(isvc)
 		if err != nil {
 			r.Log.Error(err, "Failed to reconcile", "reconciler", reflect.ValueOf(reconciler), "Name", isvc.Name)
-			r.Recorder.Eventf(isvc, v1.EventTypeWarning, "InternalError", err.Error())
-			r.updateStatus(isvc, deploymentMode)
+			r.Recorder.Event(isvc, v1.EventTypeWarning, "InternalError", err.Error())
+			_ = r.updateStatus(isvc, deploymentMode)
```
alpe: I stumbled upon this line and was wondering why this update call exists. The error was not handled, so it would be good to mark it explicit to be ignored. I am not sure if this is the right behaviour though.

sivanantha321: This is handled in PR #3378

terrytangyuan (approved):
/lgtm

PR #3434, opened 2024-02-14
Fix Raw Logger E2E

israel-hdez:

This is a small fix to the test_raw_logger.py/test_kserve_logger E2E test.

This E2E test was creating a `message-dumper-raw` that was dependent on a Serverless setup, leading to the E2E to fail on an environment where Istio nor KNative are available.

The fix is to add the right annotation to `message-dumper-raw` to deploy it in Raw mode, and adjust the logger URL of the `isvc-logger-raw` to point to the Kubernetes Service, rather than the KNative service.

**Release note**:

```release-note
NONE
```


terrytangyuan (commented):
/lgtm

israel-hdez:
There seems to be an issue with paddle which, IMO, is unrelated to these changes.
@terrytangyuan I wonder if you have privileges to re-run the check?

israel-hdez:
All checks are passing
cc @yuzisun

PR #3435, opened 2024-02-14
Reduce E2Es dependency on CI environment

israel-hdez:

Some code of the E2Es assume the environment is GitHub, because it is referring to GitHub-specific variables.

This PR focuses on references to the `kserve/image-transformer` image. This image is built in the CI flow and made available to the runner, so that a pull from an external registry is not needed.

The references to this image are changed to an environment variable that is more agnostic to the runner, in an effort to make E2Es more compatible in other environments. The e2e-test.yml is modified to declare the new variable.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -13,6 +13,7 @@ on:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  IMAGE_TRANSFORMER_IMG: "kserve/image-transformer:${{ github.sha }}"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


israel-hdez:
@sivanantha321 since this touches the collocation E2Es, I'd like to have your review in this PR.

.github/workflows/e2e-test.yml
```diff
@@ -13,6 +13,7 @@ on:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  IMAGE_TRANSFORMER_IMG: "kserve/image-transformer:${{ github.sha }}"
```
sivanantha321, line 16: Can we set the env variable in python tests itself using fixtures or something ? This will make the local testing bit easier. As for the rest LGTM

sivanantha321 (approved):

terrytangyuan (approved):
/lgtm

PR #3444, opened 2024-02-17
fix: Correct args override behavior in runtime and predictor containers

### LINKED ISSUE #3453: `predictor.model.args` defined in `InferenceService` does not override `containers.args` in `ServingRuntime`
/kind bug

 When specifying both `args` field in `InferenceService` and `ServingRuntime`, the both `args` field are combined in a list which is then consumed by the predictor deployment  in the example shown below. In this example, the inference-predictor cannot start because the ` text-generation-launcher` command can only takes ` port value. 

```
  containers:
  - args:
    - --port=3000 
    - --port=3000
    - --grpc-port=8033
    - --num-shard=1
     command:
      - text-generation-launcher
```

 I'd expect the `predictor.model.args` defined in `InferenceService` to override `containers.args` in `ServingRuntime`

See above

**Environment:**

- Istio Version: N/A
- Knative Version: N/A
- KServe Version: 0.11
- Kubeflow version: N/A
- Cloud Environment: ibm
- Minikube/Kind version: N/A
- Kubernetes version: OCP 4.13 / K8s 1.26
- OS (e.g. from `/etc/os-release`): Linux

terrytangyuan:
Fixes https://github.com/kserve/kserve/issues/3453. The current args from two container specs are being merged in a naive way and cannot handle complex use cases.
1. Key and value are separated to different args, e.g. `--foo bar`  vs `--foo=newBar`
1. Boolean flags, especially when it comes to merging/overriding, e.g. `--bool-flag`
1. Handle duplicates and different ordering of args

The code can be complex if we want to cover these cases.

if there's an overlap, we don't merge them since the merge result will be incorrect and contain duplicates. In addition, instead of overriding individual args, we should simply replace the entire list of args by the list of args defined in `InferenceService`. This PR corrects the behavior.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -211,7 +211,30 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
 	if mergedContainer.Name == "" {
 		mergedContainer.Name = runtimeContainerName
 	}
+
+	// Only concatenate args when there's no overlap. See discussions in https://github.com/kserve/kserve/pull/3444
+	if len(findIntersection(runtimeContainer.Args, predictorContainer.Args)) == 0 {
+		mergedContainer.Args = append(append([]string{}, runtimeContainer.Args...), predictorContainer.Args...)
+	}
+
 	return &mergedContainer, nil
+}
+
+func findIntersection(arr1, arr2 []string) []string {
+	intersection := make([]string, 0)
+
+	// Traverse each element in arr1
+	for _, num1 := range arr1 {
+		// Check if num1 exists in arr2
+		for _, num2 := range arr2 {
+			if num1 == num2 {
+				intersection = append(intersection, num1)
+				break
+			}
+		}
+	}
+
+	return intersection
 }
 
 // MergePodSpec Merge the predictor PodSpec struct with the runtime PodSpec struct, allowing users
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -46,7 +46,7 @@ import (
 // Constants
 var (
 	SupportedStorageURIPrefixList = []string{"gs://", "s3://", "pvc://", "file://", "https://", "http://", "hdfs://", "webhdfs://", "oci://"}
-	log                           = logf.Log.WithName("CredentialBuilder")
+	log                           = logf.Log.WithName("InferenceServiceUtils")
 )
 
 const (
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -222,18 +222,16 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
 
 func findIntersection(arr1, arr2 []string) []string {
 	intersection := make([]string, 0)
-
 	// Traverse each element in arr1
-	for _, num1 := range arr1 {
-		// Check if num1 exists in arr2
-		for _, num2 := range arr2 {
-			if num1 == num2 {
-				intersection = append(intersection, num1)
+	for _, elem1 := range arr1 {
+		// Check if elem1 exists in arr2
+		for _, elem2 := range arr2 {
+			if elem1 == elem2 {
+				intersection = append(intersection, elem1)
 				break
 			}
 		}
 	}
-
 	return intersection
 }
 
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils.go
@@ -40,11 +40,13 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


spolti (approved):

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -208,6 +208,16 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
 
 	// Strategic merge patch will replace args but more useful behaviour here is to concatenate
 	mergedContainer.Args = append(append([]string{}, runtimeContainer.Args...), predictorContainer.Args...)
+	// Remove duplicates
+	seenArgs := make(map[string]bool)
+	var result []string
+	for _, arg := range mergedContainer.Args {
+		if _, ok := seenArgs[arg]; !ok {
+			seenArgs[arg] = true
+			result = append(result, arg)
```
yuzisun: What happens for the case that key and value are separated to different args ?

For example
—foo
dummy
—foo
test

terrytangyuan: I realized that the current args from two container specs are being merged in a naive way and cannot handle complex use cases.
1. Key and value are separated to different args, e.g. `--foo bar`  vs `--foo=newBar`
1. Boolean flags, especially when it comes to merging/overriding, e.g. `--bool-flag`
1. Handle duplicates and different ordering of args

The code can be complex if we want to cover these cases.

Instead of overriding individual args, we should simply replace the entire list of args by the list of args defined in `InferenceService`.

I have just updated the PR based on the new approach.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -211,10 +211,6 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
 	if mergedContainer.Name == "" {
 		mergedContainer.Name = runtimeContainerName
 	}
-
-	// Strategic merge patch will replace args but more useful behaviour here is to concatenate
-	mergedContainer.Args = append(append([]string{}, runtimeContainer.Args...), predictorContainer.Args...)
```
terrytangyuan, line 216: Looks like tests are failing because test cases expect that runtime container args need to exist in the merged container args. If our users are already relying on this, this will be a breaking change.

Should we consider adding a flag to turn this on/off?

spolti: Is there a default command that needs to be always added by default?

If not, maybe instead to creating a flag we could make a conditional configuration, if there is no args don't set the args, maybe set the field as nil, otherwise set with the provided ones.

terrytangyuan: Ok CI passing now. So basically if there's an overlap, we don't merge them since the merge result will be incorrect and contain duplicates.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -212,12 +212,29 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
 		mergedContainer.Name = runtimeContainerName
 	}
 
-	// Strategic merge patch will replace args but more useful behaviour here is to concatenate
-	mergedContainer.Args = append(append([]string{}, runtimeContainer.Args...), predictorContainer.Args...)
+	// Only concatenate args when there's no overlap. See discussions in https://github.com/kserve/kserve/pull/3444
```
danielezonca, line 218: It should be probably better to document this behaviour and at least properly log both branches (if and else): it might be misleading with the update of a resource with scenarios like
1. `ServingRuntime` and `InferenceService` has no overlap -> merge
1. `ServingRuntime` is updated to add a new default configuration -> now overlap so no more merge and replace the command entirely
1. `InferenceService` has overlap but then remove one of the options -> no overlap anymore so now merge

Wdyt?

terrytangyuan: Great idea. Added docs and log.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -212,12 +215,32 @@ func MergeRuntimeContainers(runtimeContainer *v1.Container, predictorContainer *
...
+	if len(findIntersection(runtimeContainer.Args, predictorContainer.Args)) == 0 {
+		mergedContainer.Args = append(append([]string{}, runtimeContainer.Args...), predictorContainer.Args...)
+		log.Info("Concatenated args from runtime and predictor containers since there's no overlap")
+	} else {
+		log.Info("Predictor container args will override the runtime container args since an overlap is found")
```
yuzisun, line 223: This might not be backwards compatible as there could be cases when you need take args from the serving runtime while only overriding the `--model_name` arg.

pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
```diff
@@ -765,6 +765,72 @@ func TestMergeRuntimeContainers(t *testing.T) {
 				},
 			},
 		},
+		"OverlappingContainerArgs": {
+			containerBase: &v1.Container{
+				Name:  "kserve-container",
+				Image: "default-image",
+				Args: []string{
+					"--foo=bar",
+					"--test=dummy",
```
yuzisun, line 774: dropping args from runtime is not the expected behavior

terrytangyuan:
We no longer need this fix/enhancement. Closing for now.

PR #3450, opened 2024-02-19
Fix: Add padding and truncation in huggingface tokenizer

### LINKED ISSUE #3438: ValueError when running HuggingfaceModel with multiple inputs of different length
/kind bug

- Modify `test_bert_sequence_classification()` in `huggingfaceserver/test_model.py` such that the two input instances have different length:
```
% git diff huggingfaceserver/test_model.py
--- a/python/huggingfaceserver/huggingfaceserver/test_model.py
+++ b/python/huggingfaceserver/huggingfaceserver/test_model.py
@@ -62,7 +62,8 @@ def test_bert_sequence_classification():
     model.load()
 
     request = "Hello, my dog is cute."
-    response = asyncio.run(model({"instances": [request, request]}, headers={}))
+    request_2 = "Hello, my dog is cute. I love him."
+    response = asyncio.run(model({"instances": [request, request_2]}, headers={}))
     assert response == {"predictions": [1, 1]}
```
- Run the modified test: `pytest -v -W ignore -k "test_bert_sequence_classification"`
- This will fail with:
```
ValueError: Unable to create tensor, you should probably activate truncation and/or padding with 'padding=True' 'truncation=True'
to have batched tensors with the same length. Perhaps your features (`input_ids` in this case) 
have excessive nesting (inputs type `list` where type `int` is expected).
```
Full logs: [test.txt](https://github.com/kserve/kserve/files/14293621/test.txt)

- In short, the problem is that because the inputs have different lengths, they can’t be converted to fixed-size tensors

I expect that we are able to supply inputs of different length in a single batch.

**Potential fix:**
Based on the error, it seems like we need to add paddings, which is documented [here](https://huggingface.co/docs/transformers/en/pad_truncation) by hugging face.

> Batched inputs are often different lengths, so they can’t be converted to fixed-size tensors. Padding and truncation are strategies for dealing with this problem, to create rectangular tensors from batches of varying lengths. Padding adds a special padding token to ensure shorter sequences will have the same length as either the longest sequence in a batch or the maximum length accepted by the model. Truncation works in the other direction by truncating long sequences.
>
> In most cases, padding your batch to the length of the longest sequence and truncating to the maximum length a model can accept works pretty well. However, the API supports more strategies if you need them. The three arguments you need to are: padding, truncation and max_length.

So we should add `padding=True` (or `"longest"`, same behaviour, but more explicit) when calling `self.tokenizer()`, which will pad the inputs to the longest sequence in the batch, so that all inputs will have the same length.

I'm willing to contribute a fix for this. I've tested it locally and it works, but I'll add more rigorous tests as well, perhaps also see if we need truncation as well or just padding.

**Environment:**

- KServe Version: commit [e21b913](https://github.com/kserve/kserve/commit/e21b913fe10b377d72e67fbf92955dbcedd6f465)

kevinmingtarja:
Add padding and truncation in huggingface tokenizer. Without this padding, we will run into a `ValueError` when we pass multiple inputs of different length to a `HuggingfaceModel` because they can’t be converted to fixed-size tensors. And without truncation, we will also run into a `RuntimeError` when we pass an input with a token length that is larger than the model's max token length. More details can be found in the issue linked below.

Fixes #3438 

**Type of changes**

- [x] Added two tests in `python/huggingfaceserver/huggingfaceserver/test_model.py`

**Checklist**:


terrytangyuan (commented):
LGTM

/cc @yuzisun to double check

python/huggingfaceserver/huggingfaceserver/model.py
```diff
@@ -133,6 +133,8 @@ def preprocess(self, payload: Union[Dict, InferRequest], context: Dict[str, Any]
                 max_length=self.max_length,
                 add_special_tokens=self.add_special_tokens,
                 return_tensors=TensorType.NUMPY,
+                padding=True,
```
yuzisun, line 156: @kevinmingtarja is this always true or we should make them as options

python/huggingfaceserver/huggingfaceserver/test_model.py
```diff
@@ -89,5 +89,33 @@ def test_bloom():
...
+    response = asyncio.run(model({"instances": [request_one, request_two]}, headers={}))
+    assert response == {"predictions": [1, 1]}
+
+
+def test_input_truncation():
+    model = HuggingfaceModel("bert-base-uncased-yelp-polarity",
+                             {"model_id": "textattack/bert-base-uncased-yelp-polarity",
+                              "task": MLTask.sequence_classification.value})
+    model.load()
+
```
yuzisun: Linting error
blank line contains whitespace

python/huggingfaceserver/huggingfaceserver/model.py
kevinmingtarja: I think it's safe to always set to true, i.e. even if we only provide a single input, no padding will be applied, so we don't need to handle that case separately.

Reference: https://huggingface.co/docs/transformers/en/pad_truncation

python/huggingfaceserver/huggingfaceserver/test_model.py
kevinmingtarja: Fixed!

yuzisun:
thanks @kevinmingtarja !!

kevinmingtarja:
thanks for the review as well!

PR #3472, opened 2024-02-22
Enhance controller setup based on available CRDs

### LINKED ISSUE #3470: Enhance manager setup based on available CRDs
/kind feature

This issue specifically refers to the code of this func: https://github.com/kserve/kserve/blob/99ac7b2050fafb14b7114b94ad6e3fd7ecfe3d15/pkg/controller/v1beta1/inferenceservice/controller.go#L298

Despite the KServe website has instructions to setup KServe under a Serverless environment and also under Raw, the manager setup uses configs to decide what set of CRDs are going to be watched.

Specifically, the decisions around the default deployment mode are not optimal. Users may still prefer using RawDeployment by default while using Serverless mode for the workloads where KNative-related features are needed. However, configuring RawDeployment as the default mode will lead to not watching KNative and Istio related resources, which is probably not OK because users may modify them and break the setup and the controller won't fix the broken resources.

Therefore, rather than using the default deployment mode as a helper to set what resources to watch, it may be better to resolve if the CRDs are available to watch them. In case the user configures Serverless as default deployment mode, the controller would still run fine (right now it continuously restarts because of the missing CRDs) if the user has the policy to annotate all the ISVCs with RawDeployment mode and, if the annotation is forgotten, the manager would start logging errors and that would be fine.

israel-hdez:

This enhances the setup of the InferenceService controller and the InferenceGraph controller. Instead of relying on the `defaultDeploymentMode` configuration to determine what CRDs to watch, the setup now checks whether KNative Services and Istio VirtualServices are available in the cluster and setup the watches (invoke `Owns`) accordingly.

This enhancement has the following advantages:
* A crashloop is prevented if the CRDs are missing in the cluster. The user would still be able to create InferenceServices by taking care of annotating the ISVC for RawDeployment mode.
* If RawDeployment mode is configured as the default mode, the controllers would still watch for KNative and Istio resources if these components are available. This will let the controller watch for changes for the dependent resources if the user uses Serverless mode for some of the InferenceServices.
* In the InferenceService controller, the watch for the VirtualServices is still conditioned to the value of the `disableVirtualHost` configuration.

Fixes #3470 

**Type of changes**

- Install KServe in a cluster where Istio and KNative are missing, and still configure Serverless as default deployment mode. See the controller outputting in the logs messages like `The InferenceGraph controller won't watch serving.knative.dev/v1/Service resources because the CRD is not available.` and verified there no errors are logged.
  - Verify that InferenceServices with the RawDeployment mode annotation deploy successfully.
- Do a regular Serverless install but change the default deployment mode to Raw
  - Verify that an InferenceService can both be created in Raw mode and Serverless mode.
  - For the InferenceService deployed in Serverless mode, verify that modifying the KServe created VirtualService and/or the KNative Service would trigger a reconcile and KServe would fix the resources.

**Checklist**:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -23,6 +23,7 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	"github.com/kserve/kserve/pkg/utils"
 	istio_networking "istio.io/api/networking/v1beta1"
 	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
 	v1 "k8s.io/api/core/v1"
@@ -159,13 +160,26 @@ func main() {
 		setupLog.Error(err, "unable to get ingress config.")
 		os.Exit(1)
 	}
-	if deployConfig.DefaultDeploymentMode == string(constants.Serverless) {
+
+	ksvcFound, ksvcCheckErr := utils.IsCrdAvailable(cfg, knservingv1.SchemeGroupVersion.String(), constants.KnativeServiceKind)
+	if ksvcCheckErr != nil {
+		setupLog.Error(ksvcCheckErr, "error when checking if Knative Service kind is available")
+		os.Exit(1)
+	}
+	if ksvcFound {
 		setupLog.Info("Setting up Knative scheme")
 		if err := knservingv1.AddToScheme(mgr.GetScheme()); err != nil {
 			setupLog.Error(err, "unable to add Knative APIs to scheme")
 			os.Exit(1)
 		}
-		if !ingressConfig.DisableIstioVirtualHost {
+	}
+	if !ingressConfig.DisableIstioVirtualHost {
+		vsFound, vsCheckErr := utils.IsCrdAvailable(cfg, istioclientv1beta1.SchemeGroupVersion.String(), constants.IstioVirtualServiceKind)
+		if vsCheckErr != nil {
+			setupLog.Error(vsCheckErr, "error when checking if Istio VirtualServices are available")
+			os.Exit(1)
+		}
+		if vsFound {
 			setupLog.Info("Setting up Istio schemes")
 			if err := istioclientv1beta1.AddToScheme(mgr.GetScheme()); err != nil {
 				setupLog.Error(err, "unable to add Istio v1beta1 APIs to scheme")
--- a/pkg/controller/v1alpha1/inferencegraph/controller.go
+++ b/pkg/controller/v1alpha1/inferencegraph/controller.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/kserve/kserve/pkg/utils"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -37,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
@@ -55,10 +57,11 @@ import (
 // InferenceGraphReconciler reconciles a InferenceGraph object
 type InferenceGraphReconciler struct {
 	client.Client
-	Clientset kubernetes.Interface
-	Log       logr.Logger
-	Scheme    *runtime.Scheme
-	Recorder  record.EventRecorder
+	ClientConfig *rest.Config
+	Clientset    kubernetes.Interface
+	Log          logr.Logger
+	Scheme       *runtime.Scheme
+	Recorder     record.EventRecorder
 }
 
 // InferenceGraphState describes the Readiness of the InferenceGraph
@@ -193,6 +196,18 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		logger.Info("Inference graph raw before propagate status")
 		PropagateRawStatus(&graph.Status, deployment, url)
 	} else {
+		// Abort if Knative Services are not available
+		ksvcAvailable, checkKsvcErr := utils.IsCrdAvailable(r.ClientConfig, knservingv1.SchemeGroupVersion.String(), constants.KnativeServiceKind)
+		if err != nil {
+			return reconcile.Result{}, checkKsvcErr
+		}
+
+		if !ksvcAvailable {
+			r.Recorder.Event(graph, v1.EventTypeWarning, "ServerlessModeRejected",
+				"It is not possible to use Serverless deployment mode when Knative Services are not available")
+			return reconcile.Result{Requeue: false}, reconcile.TerminalError(fmt.Errorf("the resolved deployment mode of InferenceGraph '%s' is Serverless, but Knative Serving is not available", graph.Name))
+		}
+
 		// @TODO check raw deployment mode
 		desired := createKnativeService(graph.ObjectMeta, graph, routerConfig)
 		err = controllerutil.SetControllerReference(graph, desired, r.Scheme)
@@ -268,15 +283,22 @@ func inferenceGraphReadiness(status v1alpha1api.InferenceGraphStatus) bool {
 }
 
 func (r *InferenceGraphReconciler) SetupWithManager(mgr ctrl.Manager, deployConfig *v1beta1api.DeployConfig) error {
-	if deployConfig.DefaultDeploymentMode == string(constants.RawDeployment) {
-		return ctrl.NewControllerManagedBy(mgr).
-			For(&v1alpha1api.InferenceGraph{}).
-			Owns(&appsv1.Deployment{}).
-			Complete(r)
-	} else {
-		return ctrl.NewControllerManagedBy(mgr).
-			For(&v1alpha1api.InferenceGraph{}).
-			Owns(&knservingv1.Service{}).
-			Complete(r)
+	r.ClientConfig = mgr.GetConfig()
+
+	ksvcFound, err := utils.IsCrdAvailable(r.ClientConfig, knservingv1.SchemeGroupVersion.String(), constants.KnativeServiceKind)
+	if err != nil {
+		return err
 	}
+
+	ctrlBuilder := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1api.InferenceGraph{}).
+		Owns(&appsv1.Deployment{})
+
+	if ksvcFound {
+		ctrlBuilder = ctrlBuilder.Owns(&knservingv1.Service{})
+	} else {
+		r.Log.Info("The InferenceGraph controller won't watch serving.knative.dev/v1/Service resources because the CRD is not available.")
+	}
+
+	return ctrlBuilder.Complete(r)
 }
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
@@ -87,10 +88,11 @@ const (
 // InferenceServiceReconciler reconciles a InferenceService object
 type InferenceServiceReconciler struct {
 	client.Client
-	Clientset kubernetes.Interface
-	Log       logr.Logger
-	Scheme    *runtime.Scheme
-	Recorder  record.EventRecorder
+	ClientConfig *rest.Config
+	Clientset    kubernetes.Interface
+	Log          logr.Logger
+	Scheme       *runtime.Scheme
+	Recorder     record.EventRecorder
 }
 
 func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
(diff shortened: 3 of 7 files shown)


pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -296,26 +296,33 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1api.InferenceServiceStatus, deplo
...
-			Owns(&knservingv1.Service{}).
-			Owns(&istioclientv1beta1.VirtualService{}).
-			Owns(&appsv1.Deployment{}).
-			Complete(r)
+	ksvcFound, err := utils.IsCrdAvailable(mgr.GetConfig(), "serving.knative.dev/v1", "Service")
+	if err != nil {
+		return err
+	}
+
+	vsFound, err := utils.IsCrdAvailable(mgr.GetConfig(), "networking.istio.io/v1beta1", "VirtualService")
```
Jooho, line 304: If istio changes api version, it would be failed. why don't we check all api versions that have the kind of CRD?

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -266,15 +267,20 @@ func inferenceGraphReadiness(status v1alpha1api.InferenceGraphStatus) bool {
 }
 
 func (r *InferenceGraphReconciler) SetupWithManager(mgr ctrl.Manager, deployConfig *v1beta1api.DeployConfig) error {
-	if deployConfig.DefaultDeploymentMode == string(constants.RawDeployment) {
-		return ctrl.NewControllerManagedBy(mgr).
-			For(&v1alpha1api.InferenceGraph{}).
-			Owns(&appsv1.Deployment{}).
-			Complete(r)
+	ksvcFound, err := utils.IsCrdAvailable(mgr.GetConfig(), "serving.knative.dev/v1", "Service")
```
Jooho: If istio changes api version, it would behavior something wrong. Why don't we check all api versions that have the kind of CRD?

Jooho (commented):
by the way, don't we need tests for this change? crashlooping is a big issue so adding tests would be a good idea.

danielezonca:
@israel-hdez 
Nice job with the PR, I have a couple of questions

- What about a scenario where Knative/Istio are not installed when KServe is installed but then they are? I think it is fine to mention this scenario is not supported for now if it makes too complex this PR, I just want to understand the overall behaviour.
- Have you tested the error if a user tries to deploy a Serverless deployment `InferenceService` without having Knative after this PR? Before the PR the controller was failing checking the CRD but what about now? I would expect a proper error in the `InferenceService` status/event to mention that it is not possible to deploy but at least that the controller doesn't crash and the IS status reports the failed status

pkg/controller/v1beta1/inferenceservice/controller.go
terrytangyuan: Move these strings to constants and consider using a loop to check all of these CRDs since we may need to check them for different versions, etc.

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -266,15 +267,20 @@ func inferenceGraphReadiness(status v1alpha1api.InferenceGraphStatus) bool {
...
+		Owns(&appsv1.Deployment{})
+
+	if ksvcFound {
+		ctrlBuilder = ctrlBuilder.Owns(&knservingv1.Service{})
 	} else {
-		return ctrl.NewControllerManagedBy(mgr).
-			For(&v1alpha1api.InferenceGraph{}).
-			Owns(&knservingv1.Service{}).
-			Complete(r)
+		r.Log.Info("The InferenceGraph controller won't watch serving.knative.dev/v1/Service resources because the CRD is not available.")
```
terrytangyuan, line 300: Should this be warning?

israel-hdez:
@danielezonca 

> @israel-hdez Nice job with the PR, I have a couple of questions
> 
>     * What about a scenario where Knative/Istio are not installed when KServe is installed but then they are? I think it is fine to mention this scenario is not supported for now if it makes too complex this PR, I just want to understand the overall behaviour.
> 

I think detecting the change in the setup would add some complexity. I'm not sure if it is worth to do that work, given changes to the stack may not be that common.

For this PR, if a Serverless setup is done after KServe is installed, the only additional action needed would be to restart the KServe controller pod and it should start using KNative as normally.

>     * Have you tested the error if a user tries to deploy a Serverless deployment `InferenceService` without having Knative after this PR? Before the PR the controller was failing checking the CRD but what about now? I would expect a proper error in the `InferenceService` status/event to mention that it is not possible to deploy but at least that the controller doesn't crash and the IS status reports the failed status

I just tried it. There is no crash, although the controller will try to reconcile continuously. I do see that it tries to write something in the status field, but it fails for some reason (looks like a bug) - so, looks like the error message is hidden by another one.

I agree that it should fail more gracefully. Given the detected CRDs it shouldn't reconcile continuously a Serverless InferenceService when the stack is absent. I'll do the needed code changes for this.

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: The `r.Log` does not have a warning function. Only `Info` and `Error`.

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: To manipulate the KNative-related and Istio-related resources, their API libraries are imported and used in `go.mod`. Right now, [Istio library version 1.19.4 is being used](https://github.com/kserve/kserve/blob/f86bc3d307ce3d9d8e5f6a17a8092d13273ca97d/go.mod#L29) and [KNative 0.39.3 is being used](https://github.com/kserve/kserve/blob/f86bc3d307ce3d9d8e5f6a17a8092d13273ca97d/go.mod#L39). Such libraries would not contain any future versions of the Istio or KNative APIs. Furthermore, we can watch (invoke `Own`) only on a specific API version of the CRDs.

So, if either Istio or KNative change their APIs, KServe would require more code changes than just changing this check (see an example in #3150). So, to me, it does not make sense to check for all versions, while the rest of the code would only work against a specific version. 

BTW, there is a [recommended version matrix](https://kserve.github.io/website/master/admin/serverless/serverless/#recommended-version-matrix) in the website which seems to be outdated. I'd think it should work as a compatibility matrix, as I don't think it is feasible to support any combinations.

pkg/controller/v1alpha1/inferencegraph/controller.go
terrytangyuan: Oh interesting

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: I do agree on moving to constants, I'll do it.

About a loop for checking different versions, I'm not sure that's useful. Read more in [my reply to Jooho](https://github.com/kserve/kserve/pull/3472/files#r1503095746).

terrytangyuan: SGTM

pkg/controller/v1alpha1/inferencegraph/controller.go
spolti: I guess warnings would looks like:

```go
V(0).Info(...
```

pkg/controller/v1beta1/inferenceservice/controller.go
spolti: Guess the best way is to have a const (array in case of more versions) and check those.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -296,26 +296,33 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1api.InferenceServiceStatus, deplo
...
 	} else {
-		return ctrl.NewControllerManagedBy(mgr).
-			For(&v1beta1api.InferenceService{}).
-			Owns(&knservingv1.Service{}).
-			Owns(&appsv1.Deployment{}).
-			Complete(r)
+		r.Log.Info("The InferenceService controller won't watch networking.istio.io/v1beta1/VirtualService resources because the CRD is not available.")
 	}
 
+	return ctrlBuilder.Complete(r)
```
spolti, line 350: Are you planning to add tests to check if the configuration is correctly applied after the reconcile runs?

israel-hdez: What configuration are you referring to?

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: I modified to using constants and 3rd-party library-provided variables

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: Would `V(0)` be meaningful?
I'm not sure it will have any outcome in the logs.

israel-hdez:
> * I would expect a proper error in the `InferenceService` status/event to mention that it is not possible to deploy

I just added some code to abort reconciliation early and log an `Event` that should be visible with `kubectl describe`̣ .

israel-hdez:
@Jooho @terrytangyuan @spolti The code is ready. Please, do another review round.

I added tests around the impacted `Reconcile` functions.

I failed to find a way to add tests for the `SetupWithManager` functions. If you know how to do it, I'll appreciate your advice.

pkg/utils/utils.go
```diff
@@ -169,3 +175,62 @@ func AppendPortIfNotExists(slice []v1.ContainerPort, elems ...v1.ContainerPort)
...
+
+// IsCrdAvailable checks if a given CRD is present in the cluster by verifying the
+// existence of its API.
+func IsCrdAvailable(config *rest.Config, groupVersion, kind string) (bool, error) {
+	gvResources, err := GetAvailableResourcesForApi(config, groupVersion)
+	if err != nil {
+		return false, err
+	}
+
+	found := false
```
spolti, line 187: this could be reduced to something like:

```
return gvResources != nil && any(crd.Kind == kind for crd := range gvResources.APIResources), nil
```

pkg/utils/utils.go
```diff
@@ -169,3 +175,62 @@ func AppendPortIfNotExists(slice []v1.ContainerPort, elems ...v1.ContainerPort)
...
+			return nil, newClientErr
+		}
+
+		var getGvResourcesErr error
+		gvResources, getGvResourcesErr = discoveryClient.ServerResourcesForGroupVersion(groupVersion)
+		if getGvResourcesErr != nil && !apierr.IsNotFound(getGvResourcesErr) {
+			return nil, getGvResourcesErr
+		}
+
+		SetAvailableResourcesForApi(groupVersion, gvResources)
```
spolti, line 221: Is SetAvailableResourcesForApi used outside this package?
If not you could, instead create a func to store it, just put them in the cache here.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -299,26 +316,35 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1api.InferenceServiceStatus, deplo
...
+	}
+
+	ctrlBuilder := ctrl.NewControllerManagedBy(mgr).
+		For(&v1beta1api.InferenceService{}).
+		Owns(&appsv1.Deployment{})
+
+	if ksvcFound {
+		ctrlBuilder = ctrlBuilder.Owns(&knservingv1.Service{})
+	} else {
+		r.Log.Info("The InferenceService controller won't watch serving.knative.dev/v1/Service resources because the CRD is not available.")
```
spolti, line 341: wouldn't be good to publish it to events as well?
Most of the time we don't usually see logs, having it in the events would be helpful.

pkg/controller/v1beta1/inferenceservice/controller.go
spolti: I was referring to the isvc, e.g.
if you change the isvc and force knative, and if it is disabled, the isvc should not be reconciled, right? so the confs should be the same than before running the reconcile loop again.

does it make sense?

pkg/controller/v1alpha1/inferencegraph/controller.go
spolti: It is not clear in the documentation, but digging a little bit about it, V(0) seems to be a way to add verbosity in the logs.

terrytangyuan: Yes it's the verbosity level.

terrytangyuan: So if you run the controller you can do something like `--gloglevel 5` to print out only logs that falls into that verbosity boundary

pkg/utils/utils.go
spolti: ok, it is. :)

spolti (approved):

cmd/manager/main.go
```diff
@@ -157,13 +158,26 @@ func main() {
 		setupLog.Error(err, "unable to get ingress config.")
 		os.Exit(1)
 	}
-	if deployConfig.DefaultDeploymentMode == string(constants.Serverless) {
+
+	ksvcFound, ksvcCheckErr := utils.IsCrdAvailable(cfg, knservingv1.SchemeGroupVersion.String(), constants.KnativeServiceKind)
+	if ksvcCheckErr != nil {
+		setupLog.Error(ksvcCheckErr, "error when checking if Knative Services are available")
```
terrytangyuan: Minor: seems like not all changes are synced from https://github.com/opendatahub-io/kserve/pull/265?

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: I reviewed this and I think we should not raise the verbosity for this one as it helps to see why the controller would not watch the involved CRD -- and it is printed only once for the pod lifetime.

So, I think it is worth to have it as a regular Info, given `r.Log` does not have a warning function.

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: I rarely deal with the Events API. So, I don't know. 
Personally, looking at the logs would be the first thing I'd do to diagnose things. Then, I typically look at Events of some resource. If we want to record an event, I'm not sure which resource the event should belong to?

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: At some point in time, I added to this PR some units in `pkg/controller/v1beta1/inferenceservice/controller_test.go` and `pkg/controller/v1alpha1/inferencegraph/controller_test.go`. Maybe that's enough?

pkg/utils/utils.go
israel-hdez: I copied and pasted and it does not build....
I guess this is pseudocode. If it is, then yes, what you provided is the short form of all that code.

cmd/manager/main.go
israel-hdez: Fixed

israel-hdez:
@terrytangyuan @spolti I finally had time to finish this PR.
Please, take another look at this.

israel-hdez:
I pushed an empty commit to re-run CI, which is now passing.
@spolti @terrytangyuan Please, re-approve if you are still OK with this PR.

terrytangyuan (approved):
/lgtm

israel-hdez:
Per prow instructions:

/assign @yuzisun

israel-hdez:
@yuzisun This is rebased to resolve conflicts, and CI is passing.

yuzisun:
Thanks @israel-hdez for the nice PR descriptions for easier review!

/approve

PR #3483, opened 2024-02-27
fix: Service existence check should check all spec fields

terrytangyuan:
There is an edge case where a service with the same name exists but they have different configurations. The PR adds a more strict check.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
@@ -157,8 +157,7 @@ func (r *ServiceReconciler) checkServiceExist(client client.Client) (constants.C
 }
 
 func semanticServiceEquals(desired, existing *corev1.Service) bool {
-	return equality.Semantic.DeepEqual(desired.Spec.Ports, existing.Spec.Ports) &&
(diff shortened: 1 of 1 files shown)


sivanantha321 (approved):

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -157,8 +157,7 @@ func (r *ServiceReconciler) checkServiceExist(client client.Client) (constants.C
 }
 
 func semanticServiceEquals(desired, existing *corev1.Service) bool {
-	return equality.Semantic.DeepEqual(desired.Spec.Ports, existing.Spec.Ports) &&
-		equality.Semantic.DeepEqual(desired.Spec.Selector, existing.Spec.Selector)
+	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec)
```
yuzisun, line 160: KServe does not use all the service spec fields, so there could be diffs generated by the Kubernetes defaults which causes infinite reconciliation loop.

terrytangyuan: Good point. Closing this

PR #3486, opened 2024-02-28
fix: Skip domain validation if ingress creation is disabled. Fixes #3485

### LINKED ISSUE #3485: Domain validation should be skipped if ingress creation is disabled
/kind bug

When `disableIngressCreation` is set to `true`, if ISVC name is long, the domain name generated exceeds the limit.

```
"fails to reconcile component: fails to create NewRawKubeReconciler for predictor: failed creating host name: invalid domain name \"granite-13b-chat-v2-inference-service-predictor-watsonx-huggingface.example.com\": url: Invalid value: \"granite-13b-chat-v2-inference-service-predictor-watsonx-huggingface\": must be no more than 63 characters"
```

We should disable validation when ingress creation is disabled.

/assign

terrytangyuan:
Fixes #3485

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
@@ -55,10 +55,11 @@ func GenerateDomainName(name string, obj metav1.ObjectMeta, ingressConfig *v1bet
 		return "", fmt.Errorf("error rendering the domain template: %w", err)
 	}
 
-	urlErrs := validation.IsFullyQualifiedDomainName(field.NewPath("url"), buf.String())
-	if urlErrs != nil {
-		return "", fmt.Errorf("invalid domain name %q: %w", buf.String(), urlErrs.ToAggregate())
+	if !ingressConfig.DisableIngressCreation {
+		urlErrs := validation.IsFullyQualifiedDomainName(field.NewPath("url"), buf.String())
+		if urlErrs != nil {
+			return "", fmt.Errorf("invalid domain name %q: %w", buf.String(), urlErrs.ToAggregate())
+		}
 	}
-
 	return buf.String(), nil
 }
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain_test.go
@@ -123,11 +123,12 @@ func TestGenerateDomainName(t *testing.T) {
 		{
 			name: "long domain name with validation",
 			args: args{
-				name: "super-super-super-super-super-super-super-super-super-super-long-model-name",
+				name: "super-super-super-super-super-super-super-super-super-super-long-model-name-with-validation",
 				obj:  obj,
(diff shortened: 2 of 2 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
```diff
@@ -55,10 +55,11 @@ func GenerateDomainName(name string, obj metav1.ObjectMeta, ingressConfig *v1bet
 		return "", fmt.Errorf("error rendering the domain template: %w", err)
 	}
 
-	urlErrs := validation.IsFullyQualifiedDomainName(field.NewPath("url"), buf.String())
-	if urlErrs != nil {
-		return "", fmt.Errorf("invalid domain name %q: %w", buf.String(), urlErrs.ToAggregate())
+	if !ingressConfig.DisableIngressCreation {
+		urlErrs := validation.IsFullyQualifiedDomainName(field.NewPath("url"), buf.String())
+		if urlErrs != nil {
+			return "", fmt.Errorf("invalid domain name %q: %w", buf.String(), urlErrs.ToAggregate())
```
spolti, line 61: isn't better to return nil instead ""?

terrytangyuan: The return type is string. This value is also what's used in `url.Host` which also requires a string.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
```diff
@@ -55,10 +55,11 @@ func GenerateDomainName(name string, obj metav1.ObjectMeta, ingressConfig *v1bet
 		return "", fmt.Errorf("error rendering the domain template: %w", err)
 	}
 
-	urlErrs := validation.IsFullyQualifiedDomainName(field.NewPath("url"), buf.String())
-	if urlErrs != nil {
-		return "", fmt.Errorf("invalid domain name %q: %w", buf.String(), urlErrs.ToAggregate())
+	if !ingressConfig.DisableIngressCreation {
```
yuzisun, line 58: I do not think the validation should be disabled, when ingress is turned off you still have an internal in cluster service url, you do not want that to exceed the dns limit.

danielezonca, line 58: As far as I see this is going to skip the validation but not the generation so the method is going to return an invalid value in this case.

What about move this check where `GenerateDomainName` function is invoked to skip the invocation (while if the function is invoked the error is returned)?
My main concern is that this function `GenerateDomainName` in the case of `DisableIngressCreation=true` can return an invalid value with the assumption that it will not be used because of this configuration

terrytangyuan:
Discussed offline that we should not skip this validation. Closing this.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
spolti: Isn't it just for generate a domain?
Internal services dont use domains, right? 

With that said, it would make sense to enable only when the ingress is being exposed, no?

PR #3443, opened 2024-02-16
chore: Remove Seldon Alibi dependencies. Fixes #3380

terrytangyuan:
https://github.com/kserve/kserve/issues/3380

PR #3492, opened 2024-03-01
feat: Provide minimal distribution of CRDs

### LINKED ISSUE #3487: InferenceService CRD is too large
/kind bug

KServe's [ISVC CRD](https://github.com/kserve/kserve/blob/master/config/crd/serving.kserve.io_inferenceservices.yaml) is too large and it's impossible to add new fields to the spec given the limitation of K8s CRD size in general.

This currently blocks a couple of community PRs, including https://github.com/kserve/kserve/pull/3479 and https://github.com/kserve/kserve/pull/3418.

Even though the error message is pointing to the annotations, the actual problem might be the size of the entire CRD. 
`CustomResourceDefinition "[inferenceservices.serving.kserve.io](http://inferenceservices.serving.kserve.io/)" is invalid: metadata.annotations: Too long: must have at most 262144 bytes`

terrytangyuan:
Fixes https://github.com/kserve/kserve/issues/3487.

Trying out a similar approach to what we are doing for Argo Workflows https://github.com/argoproj/argo-workflows


hack/main.go
```diff
@@ -0,0 +1,14 @@
+package main
```
yuzisun: Can we move these files under a different directory?  we mostly store scripts under `hack`

config/crd/full/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
yuzisun, line 1: We also need to update the crd helm chart, probably need to a make a new chart for minimal crd

hack/main.go
terrytangyuan: Moved to `cmd/crd-gen`

config/crd/full/kustomization.yaml
terrytangyuan: Added kserve-crd-minimal chart

terrytangyuan:
@yuzisun Addressed comments. PTAL.

terrytangyuan:
@yuzisun @sivanantha321 Just rebased and resolved conflicts after the merge of https://github.com/kserve/kserve/pull/3443.

hack/crdgen.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sivanantha321: This script only generates minimal crd right ? If yes we can name the script more specifically

sivanantha321:
I am just curious, How is the validation handled for minimal crd version. Is it through validation webhook ?

hack/crdgen.sh
terrytangyuan: Renamed!

terrytangyuan:
@sivanantha321 Yes and basically leave that responsibility to something else, e.g. front-end, validation webhook, etc.

PR #3493, opened 2024-03-01
remove unnecessary Istio settings from quick_install.sh

peterj:
Removing the `useMCP` and `PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING` as both of these already default to the values they are being set to.

**Type of changes**

Checked the minimal Istio profile where the `useMCP` defaults to false (e.g. `istioctl manifest generate --set profile=minimal | grep useMCP`). The `PILOT_ENABLE_CONFIG_DISTRIBUTION_TRACKING` flag defaults to false in [Istio's source](https://github.com/istio/istio/blob/a8eae20529e2f24a9c07a9f9caf03411d194a035/pilot/pkg/features/experimental.go#L6).

### DIFF
--- a/hack/quick_install.sh
+++ b/hack/quick_install.sh
@@ -78,7 +78,6 @@ spec:
     global:
       proxy:
         autoInject: disabled
-      useMCP: false
 
   meshConfig:
     accessLogFile: /dev/stdout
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
Can you point me to any references to defaults?

peterj:
The distribution tracking variable defaults to `false` [here](https://github.com/istio/istio/blob/a8eae20529e2f24a9c07a9f9caf03411d194a035/pilot/pkg/features/experimental.go#L64). 

The `useMCP` value isn't actually used anywhere in the code anymore. Here's [one reference to the PR](https://github.com/istio/istio/pull/48332) that removed the same setting from the installation profiles.

terrytangyuan (approved):
Thank you! 

/lgtm

PR #3495, opened 2024-03-02
Add Google Cloud Storage support to Storage Spec

tjandy98:
Add gcs service_account support to the existing storage spec

```
apiVersion: v1
stringData:
  gcs: |
    {
    "type": "service_account",
    "project_id": "",
    "private_key_id": "",
    "private_key": "",
    "client_email": "",
    "client_id": "",
    "auth_uri": "https://accounts.google.com/o/oauth2/auth",
    "token_uri": "https://oauth2.googleapis.com/token",
    "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
    "client_x509_cert_url": "",
    "universe_domain": "googleapis.com"
    }
kind: Secret
metadata:
  name: storage-config
type: Opaque
```

**Type of changes**

Documentation: https://github.com/kserve/website/pull/351


pkg/credentials/service_account_credentials_test.go
```diff
@@ -1429,7 +1429,7 @@ func TestCredentialBuilder_CreateStorageSpecSecretEnvs(t *testing.T) {
 					Name:      "storage-secret",
 					Namespace: namespace,
 				},
-				StringData: map[string]string{"minio": "{\n     \"type\": \"gs\",\n      \"access_key_id\": \"minio\",\n      \"secret_access_key\": \"minio123\",\n      \"endpoint_url\": \"http://minio-service.kubeflow:9000\",\n      \"bucket\": \"test-bucket\",\n      \"region\": \"us-south\"\n    }"},
+				StringData: map[string]string{"minio": "{\n     \"type\": \"gss\",\n      \"access_key_id\": \"minio\",\n      \"secret_access_key\": \"minio123\",\n      \"endpoint_url\": \"http://minio-service.kubeflow:9000\",\n      \"bucket\": \"test-bucket\",\n      \"region\": \"us-south\"\n    }"},
```
terrytangyuan, line 1432: Is this intentional? Why wasn't this failing before?

tjandy98: > Is this intentional? Why wasn't this failing before?

Hi @terrytangyuan , thanks for your feedback.

Yes, this change is intentional. Before this PR, the 'gs' type wasn't supported, the test would fail.

terrytangyuan (dismissed):
/lgtm

lizzzcai (dismissed):
Hi @tjandy98, thanks for the PR. I think the doc for gcs storage is missing, do you mind to update [it](https://github.com/kserve/website/tree/main/docs/modelserving/storage) as well.

python/kserve/kserve/storage/storage.py
```diff
@@ -139,6 +139,21 @@ def _update_with_storage_spec():
                     f.write(value)
                     f.flush()
 
+        if storage_secret_json.get("type", "") == "gs":
```
yuzisun: Can we add a test?

python/kserve/kserve/storage/storage.py
```diff
@@ -139,6 +139,21 @@ def _update_with_storage_spec():
...
+            temp_dir = tempfile.mkdtemp()
```
yuzisun: Can you explain how this works, why do we write to the temp directory?

python/kserve/kserve/storage/storage.py
spolti: Maybe move gs and other supported types to constants?

python/kserve/kserve/storage/storage.py
tjandy98: [Google ADC](https://cloud.google.com/docs/authentication/application-default-credentials) uses `GOOGLE_APPLICATION_CREDENTIALS` env var to obtain location of the credential JSON file

pkg/credentials/service_account_credentials_test.go
israel-hdez: Quite related to [Dan comment](https://github.com/kserve/kserve/pull/3495/files#r1547636720), but I don't think it makes sense to force this to fail, by doing this change. The test should be reworked to properly validate the GS support (or fully remove it, and create a new one from stracth).

python/kserve/kserve/storage/storage.py
israel-hdez: I, personally, don't feel that comfortable writing the credentials to a file. IMO, it is better to either:
1. Mount the credentials as a file from kserve-controller... although this may not be trivial, since this is about StorageSpec.
2. Load in memory with `service_account.Credentials.from_service_account_info` (see [docs](https://google-auth.readthedocs.io/en/latest/user-guide.html#service-account-private-key-files)) and use the returned value when creating the client in https://github.com/kserve/kserve/blob/f917fa18fc9746d48e6c5ff3e15e156b1f2cd5bc/python/kserve/kserve/storage/storage.py#L302)

If possible, please use (2).

spolti: As @israel-hdez mentioned as well, from the same doc, SA is supported, https://cloud.google.com/docs/authentication/application-default-credentials#attached-sa

imho, that is the best option.

spolti: Guess we need to add docs for this.

tjandy98: Thanks @spolti and @israel-hdez for the feedback.

@spolti [docs](https://github.com/kserve/website/pull/351)

spolti: Thanks @tjandy98 :)

spolti:
@tjandy98 can you please send a empty commit to retrigger CI?
The ci error might be caused by a instability.

python/kserve/kserve/storage/storage.py
```diff
@@ -152,6 +154,24 @@ def _update_with_storage_spec():
                     f.write(value)
                     f.flush()
 
+        if storage_secret_json.get("type", "") == "gs":
+            if storage_secret_json.get("base64_service_account", "") != "":
```
yuzisun: @tjandy98  `base64_service_account` name is a bit confusing, from what I read this is a service account key file
https://cloud.google.com/iam/docs/keys-create-delete#iam-service-account-keys-create-console

tjandy98: Thanks for your feedback. The idea is to support base64 encoded version of the credential. Do you have a suggestion?

jefshe:
It's a bit strange to me that the docs website has instructions on this but the code itself has not been merged: https://kserve.github.io/website/latest/modelserving/storage/gcs/gcs/

I've wasted a few hours reading the source code here only to find out this change isn't actually supported. Will this PR ever be merged?

python/kserve/kserve/storage/storage.py
jefshe: @yuzisun the idea is that the file is stored as a string secret and then mounted as a volume in GKE. Google's own docs do a similar thing: https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine#service-account-key-file

yuzisun: Should we rename to base64_service_account_key_file ?

python/kserve/kserve/storage/storage.py
```diff
@@ -158,6 +160,24 @@ def _update_with_storage_spec():
...
+            if storage_secret_json.get("base64_service_account_key_file", "") != "":
+                try:
+                    base64_service_account_key_file = storage_secret_json.get(
+                        "base64_service_account_key_file", ""
+                    )
+                    service_account_str = base64.b64decode(
+                        base64_service_account_key_file
+                    ).decode("utf-8")
+                    service_account_dict = json.loads(service_account_str)
+                    os.environ["GOOGLE_SERVICE_ACCOUNT"] = json.dumps(
```
yuzisun: Seems like we should use the environment variable `GOOGLE_APPLICATION_CREDENTIALS` to point to the service account key file.

https://cloud.google.com/docs/authentication/application-default-credentials

yuzisun: We can mount the secret in the container for storage spec, we already support it here custom secrets. In this way we do not need the special code path for storage spec.
https://github.com/kserve/kserve/blob/e879a397b38fdcbebd8ec004c4da87916fb30a03/pkg/credentials/service_account_credentials.go#L263

tjandy98: Hi @yuzisun, can I clarify if I am understanding correctly:

In `CreateStorageSpecSecretEnvs`, if the storage spec secret type is gcs, call `mountSecretCredential` which calls `gcs.BuildSecretVolume`. The function will need to be modified to extract the specific gcs object from the storage spec secret , and build the volume based off its `base64_service_account_key_file` value.

yuzisun: @tjandy98 sorry for the late reply, you are right. We build the volume to point to the mounted base64 sa key file and env `GOOGLE_APPLICATION_CREDENTIALS` will be taking effect when calling the GCS API.

spolti:
Is this PR ready to be reviewed? (Still on draft status).

tjandy98:
I would like to see guidance on how to proceed with integrating GCS support in the storage spec. In the latest code of the pull request, the stringData field of the GCS service account within the Secret is mounted to a specific path defined by the key using the keyToPath configuration. The type field is expected to remain consistent with existing supported storage specification types, such as s3. However, in this case, the service_account type is required as part of the service account file.

One potential solution is to store the base64-encoded credential in the data field.  However, there is currently no method, at least that I know of, to extract and mount only the value of the data field without resorting to an inelegant approach.

```
apiVersion: v1
kind: Secret
metadata:
  name: storage-config
type: Opaque
stringData:
  gcs: |
    {
      "type": "service_account",
      "data": "<base64 encoded credential>"
    }
```

davekpatrick:
Any update on this issue & pull request ? And further to the comment, https://github.com/kserve/kserve/pull/3495#issuecomment-2305076941. If this storage config spec

https://kserve.github.io/website/latest/modelserving/storage/gcs/gcs/?query=GCS#create-google-secret is not supported , can it be removed from the documentation please

https://github.com/kserve/website/blob/317538df678203beaaf189f285dbe3f51f0ad0d3/docs/modelserving/storage/gcs/gcs.md?plain=1#L39

tjandy98:
> Any update on this issue & pull request ? And further to the comment, [#3495 (comment)](https://github.com/kserve/kserve/pull/3495#issuecomment-2305076941). If this storage config spec
> 
> https://kserve.github.io/website/latest/modelserving/storage/gcs/gcs/?query=GCS#create-google-secret is not supported , can it be removed from the documentation please
> 
> https://github.com/kserve/website/blob/317538df678203beaaf189f285dbe3f51f0ad0d3/docs/modelserving/storage/gcs/gcs.md?plain=1#L39

There is an open [PR ](https://github.com/kserve/website/pull/404)to update the documentation but it has not been merged

yuzisun:
> > Any update on this issue & pull request ? And further to the comment, [#3495 (comment)](https://github.com/kserve/kserve/pull/3495#issuecomment-2305076941). If this storage config spec
> > https://kserve.github.io/website/latest/modelserving/storage/gcs/gcs/?query=GCS#create-google-secret is not supported , can it be removed from the documentation please
> > https://github.com/kserve/website/blob/317538df678203beaaf189f285dbe3f51f0ad0d3/docs/modelserving/storage/gcs/gcs.md?plain=1#L39
> 
> There is an open [PR ](https://github.com/kserve/website/pull/404)to update the documentation but it has not been merged

@tjandy98 so the plan is to use the generic secret right, Do we still need this PR?

PR #3539, opened 2024-03-21
Initial commit for OCI dist spec v1.1.0 agent support

rchincha:

    Initial commit for OCI dist spec v1.1.0 agent support

    Partially addresses https://github.com/kubeflow/community/pull/682

    OCI image and distribution specs v1.1.0 has added support for pushing
    and pulling arbitrary artifacts to a conformant registry, and not just
    container images.

    Since a registry is already needed to deploy inference workloads as
    containers, and that it would be desirable to avoid another piece of
    infrastructure just to store inference data, a OCI conformant registry
    could become that ideal store to combine and colocate both use cases.

    This plugin adds that support.

    https://kserve.github.io/website/latest/modelserving/storage/storagecontainers/

    References:

    https://opencontainers.org/posts/blog/2024-03-13-image-and-distribution-1-1/

Fixes #

https://github.com/kubeflow/community/pull/682

**Type of changes**

- [*] New feature (non-breaking change which adds functionality)
- [*] This change requires a documentation update

- [ ] Deploy a conformant registry such as (`zot`), push model data and then deploy a workload that consumes the data.

- Logs

**Checklist**:

**Release note**:

```release-note

```


rchincha:
Cross-posting here ...

https://kccnceu2024.sched.com/event/1YeLi
^ This idea is spreading around I suppose ... https://github.com/kubecon EU 2024

lizzzcai:
Hi @rchincha , glad to see this will be supported in KServe. Just want to check is there a plan to support it in storage initializer in this PR? (like what I have previously commented [here](https://github.com/kserve/kserve/issues/3043#issuecomment-1662550251)). what is the difference between `storage initializer` and `agent`?

terrytangyuan (commented):
Could you update the PR description?

rchincha:
> Hi @rchincha , glad to see this will be supported in KServe. Just want to check is there a plan to support it in storage initializer in this PR? (like what I have previously commented [here](https://github.com/kserve/kserve/issues/3043#issuecomment-1662550251)). what is the difference between `storage initializer` and `agent`?

@lizzzcai Thanks for pointing out https://github.com/kserve/kserve/issues/3043 - lots of relevant discussion there.  I am still learning about kserve so hazarding a possiblly incorrect guess. This PR should give the ability to serve models/artifacts from a OCI registry - OCI dist-spec v1.1.0 supports arbitrary artifacts. An example is perhaps this: https://kserve.github.io/website/0.7/modelserving/storage/s3/s3/

There is still a kubeflow workflow to push artifacts into a OCI registry to consider.

rchincha:
> Could you update the PR description?

Done, pls take a look.

pkg/agent/storage/provider.go
```diff
@@ -29,9 +29,11 @@ const (
 	// File  Protocol = "file://"
 	HTTPS Protocol = "https://"
 	HTTP  Protocol = "http://"
+	// OCI dist spec v1.1.0
+	OCI Protocol = "oci://"
```
rhuss: Modelcars also use `oci://` as a schema prefix, so that might be conflicting here. Modelcars in fact, are conventional OCI images as they have also an active component for keeping a container alive (notably `sleep` and also a `ln -s` for creating the symbolic link in the emptyDir volume to point to the model data).

I'm not sure how the PR proposed here infers with the logic of the modelcar as defined in the storage initializer.

rchincha: The images here are going to be stored from a OCI artifact registry, so indeed they are also oci images. The question then is the transport schema. Maybe "oci-registry://" similar to "model-registry://"?

rhuss: @rchincha sorry, I'm not deep in this PR, but how do you plan to make the modeldata in the OCI image accessible to the runtime ? For me it looks like that the existing [modelcar feature](https://kserve.github.io/website/latest/modelserving/storage/oci/) is already quite similar what you are trying to achieve with this PR. It would be cool, if we could merge your work with that.

If not, what would be the difference to https://kserve.github.io/website/latest/modelserving/storage/oci/ ? (just for my understanding)

rchincha: @rhuss indeed, the "effect" is the same, except the modelcar approach needs a running container whereas this PR is downloading model data and purely a storage approach. Also my slow progress on this PR as I am trying to understand more. No question that the modelcar approach although one can argue is a "hack", solves lots of pertinent issues and very effective.

Just FYI that there are some discussions happening in the opencontainers (OCI) side about this use case - pulling large things and making them available to container workloads.

rhuss: I agree, I see modelcars also only as a temporary solution until a proper OCI image volume mount type is available for Kubernetes (see also KEP https://github.com/kubernetes/enhancements/pull/4642) 

I haven't yet looked into your PR. Still, I'm really curious how direct OCI image access could work today (as many have tried but not really succeeded, except via that `shareProcessNamespace` trick, if you follow the long list of comments on https://github.com/kubernetes/kubernetes/issues/831), so I will definitely have a look. 

Let's check how we can combine our efforts to cover the current state while being prepared for future Kubernetes improvements (especially when it comes to which schema to select).

rchincha: Also to be aware of, fyi ...

https://github.com/opencontainers/distribution-spec/issues/536
https://github.com/opencontainers/image-spec/issues/1190

https://github.com/kubernetes/enhancements/pull/4642
wrt ^ we are trying to see if we can solve it in a way that benefits non-k8s world (yes, there is one!) also

rchincha: Some thoughts we are exploring in the OCI community to help with this.
Overall latency = pull time + unpack time + startup/run time

For "pull time", clients already pull multiple layers in parallel. But in this inference model case, there are low single-digit, very large files, likely in a single layer. Good news is we already have "Range:" based pulls for a single layer, so it is possible to pull parts on a large layer in parallel. 

For "unpack time", we can actually make that zero if the layer is already a full compressed filesystem such as squashfs or erofs, so they can just be mounted.

Once the layers are pulled, they are cached anyways, so pod restarts happen as usual.

~200 GB inference model sizes? geez!

rchincha: FYI, https://github.com/opencontainers/distribution-spec/pull/537 has been merged with some data points.

rchincha: https://github.com/opencontainers/distribution-spec/issues/494
^ introducing blake3 helps with large artifact generation and verification, because blake3 thanks to merkle structure is very fast.

PR #3578, opened 2024-04-05
build: Add flake8 and black to pre-commit hooks

terrytangyuan:
For local development, these checks would run automatically when committing changes and we can easily add additional checks/hooks.


terrytangyuan:
Need to update based on https://github.com/kserve/kserve/commit/6e88700ad4efd9890d85b1a7d29224e8c6c4d277

spolti (approved):
Just got a issue with black lint on odh:
https://github.com/opendatahub-io/kserve/actions/runs/8573660036/job/23498794321?pr=289

.github/workflows/python-test.yml
```diff
@@ -28,9 +28,8 @@ jobs:
       - name: Lint with flake8
         uses: py-actions/flake8@v2
         with:
-          args: "--config python/.flake8"
+          args: "--config .flake8"
```
terrytangyuan, line 31: Moved .flake8 file to top level to work with pre-commit

sivanantha321:
I believe it is a good thing to add the `pre-commit` as a dev dependency. WDYT ?

terrytangyuan:
> I believe it is a good thing to add the `pre-commit` as a dev dependency. WDYT ?

Good idea but I think it should only be recommended but not required. I can submit a PR to the website later.

.flake8
```diff
@@ -0,0 +1,17 @@
+[flake8]
+max-line-length = 120
```
lizzzcai, line 2: I saw [this](https://github.com/kserve/kserve/blob/96ca17a152d4603b1c2b52aa5bd02ff6d2f45140/python/kserve/setup.cfg#L2), is it something still required? maybe can delete it for consistency.

terrytangyuan: Yep. Deleted.

sivanantha321 (approved):

lizzzcai (approved):

PR #3579, opened 2024-04-05
docs: Move Alibi explainer to docs

terrytangyuan:
Alibi explainer was removed in https://github.com/kserve/kserve/pull/3443 but the source code should be moved to docs.


terrytangyuan:
cc @sivanantha321

Issue #3489, opened 2024-02-28
Make storage initializer install only what is needed for it to run


spolti:
/kind feature

**Describe the solution you'd like**
Today, when the `storage-initializer` container image is built, besides installing the required dependencies, it also brings all other dependencies needed by `KServe`, as it uses the `KServe` _poetry.lock_ to install the needed dependencies:

https://github.com/kserve/kserve/blob/master/python/storage-initializer.Dockerfile#L23

This will bring not needed dependencies, like `ray`.
Which can introduce false/positive security alerts when scanning this container for CVEs.

**Anything else you would like to add:**
One option to address it, would be moving the `storage.py` outside `KServe` and maintain it as another library so it can be built independently and only the needed dependencies would be installed.

For this, we would need to update the other components that uses it as well.

spolti:
This PR is related: https://github.com/kserve/kserve/pull/3834

However beside ray, a lot more of unneeded dependencies will not be installed as part of the storage image build.

spolti:
Hey all, as discussed in the last community meeting, we decided to rename the storage package.

As `storage` is already taken, we might want to prefix it with some string that refers to the KServe project.

I noticed that for names with multiple strings, there are the usage of `-` and `_ `, however, the majority uses `-`.
With that said, here is a suggested names list.

- kserve-storage  👍
- ks-storage 😄
- k-storage 🎉
- kserve-storage-utils 🚀
- kserve-storage-lib 👀


Please cast your vote by adding the corresponding reaction of the name of your preference in this comment.

spolti:
@yuzisun @sivanantha321 @andyi2it fyi.

PR #3584, opened 2024-04-09
Implement Huggingface model download in storage initializer

### LINKED ISSUE #3545: Support download huggingface models in storage initializer
/kind feature

I would like to use storage initializer to pull the huggingface [models](https://huggingface.co/models) which fits the KServe experience. StorageUri will be in this format: `hf://<repo>/<model>:<hash(optional)>`

```yaml
kind: InferenceService
metadata:
  name: bloom
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      #runtime: kserve-huggingface
      storageUri: hf://bigscience/bloom-560m:<hash>
      resources:
        limits:
          cpu: "2"
          memory: 4Gi
        requests:
          cpu: "1"
          memory: 2Gi
```

Another option is to support any `Git LFS`.

andyi2it:

Fixes #3545 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```


python/kserve/kserve/storage/storage.py
```diff
@@ -272,8 +271,28 @@ def _download_s3(uri, temp_dir: str):
             if mimetype in ["application/x-tar", "application/zip"]:
                 Storage._unpack_archive_file(target, mimetype, temp_dir)
 
+    @staticmethod
+    def _download_hf(uri, temp_dir: str):
+        from transformers import AutoTokenizer, AutoConfig, AutoModel
```
lizzzcai: instead of loading model and then saving the model to local, which means you need resources (lot of memory) to load the model, maybe [snapshot_download](https://github.com/vllm-project/vllm/blob/65bf2ac165734fb6339210c4b2b8ce68d2391b77/vllm/model_executor/model_loader/weight_utils.py#L203-L210) SDK will be better.

.github/workflows/e2e-test.yml
```diff
@@ -24,6 +24,7 @@ env:
   # Controller images
   CONTROLLER_IMG: "kserve-controller"
   STORAGE_INIT_IMG: "storage-initializer"
+  HF_STORAGE_INIT_IMG: "hf-storage-initializer"
```
lizzzcai: why have a new image here? I assume the feature will be in the storage initializer and we should use the single image for all types of storage.

andyi2it: @lizzzcai 
The problem is storage initializer image size grows significantly to support huggingface server, from 500+ Mb to 1.2+ Gb approximately. That would be fairly large increase in image size for a feature only a few subset of users may use.
So, we decided to use a separate image for huggingface server.

config/storagecontainers/default.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         memory: 100Mi
         cpu: 100m
       limits:
-        memory: 1Gi
+        memory: 3Gi
```
yuzisun: any reason to bump up to 3Gi?

docs/samples/storage/hf/hf_isvc.yaml
```diff
@@ -0,0 +1,24 @@
...
+  predictor:
+    serviceAccountName: hfserviceacc
+    model:
+      modelFormat:
+        name: huggingface
+      args:
+      - --model_name=llama2
+      - --model_dir=hf://meta-llama/Llama-2-7b-chat-hf
+      - --tensor_input_names=input_ids
+      - --disable_vllm
```
yuzisun: this flag is already removed and replaced by --backend

docs/samples/storage/hf/hf_isvc.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: huggingface-llama2
+spec:
+  predictor:
+    serviceAccountName: hfserviceacc
+    model:
+      modelFormat:
+        name: huggingface
+      args:
+      - --model_name=llama2
+      - --model_dir=hf://meta-llama/Llama-2-7b-chat-hf
```
yuzisun: This should be using `storageUri` field

python/kserve/kserve/storage/storage.py
```diff
@@ -154,6 +148,9 @@ def _update_with_storage_spec():
 
     @staticmethod
     def get_S3_config():
+        from botocore import UNSIGNED
+        from botocore.client import Config
+
```
yuzisun, line 159: I think maybe better to separate them out into different files now?

python/hf-storage-initializer.Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Activate virtual env
+ARG VENV_PATH
+ENV VIRTUAL_ENV=${VENV_PATH}
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+COPY kserve/pyproject.toml kserve/poetry.lock kserve/
+RUN cd kserve && poetry install --no-root --no-interaction --no-cache --extras "hf-storage"
+COPY kserve kserve
```
yuzisun: does storage initializer actually need the kserve dependency?

python/kserve/kserve/storage/test/test_storage.py
```diff
@@ -235,17 +235,16 @@ def test(_):
     mock.patch("requests.get", return_value=response)(test)()
 
 
-@mock.patch(STORAGE_MODULE + ".storage")
-def test_mock_gcs(mock_storage):
+@mock.patch("google.cloud.storage.Client")
```
yuzisun: this test should be in `test_gcs_storage.py`

python/kserve/kserve/storage/test/test_storage.py
```diff
@@ -235,17 +235,16 @@ def test(_):
...
+def test_mock_gcs(mock_client):
```
yuzisun: same here

python/kserve/kserve/storage/test/test_storage.py
```diff
@@ -255,21 +254,21 @@ def test_storage_blob_exception():
         Storage.download(blob_path)
 
 
-@mock.patch(STORAGE_MODULE + ".boto3")
-def test_storage_s3_exception(mock_boto3):
+@mock.patch("boto3.resource")
+def test_storage_s3_exception(mock_resource):
```
yuzisun: this test should be in `test_s3_storage.py`

python/kserve/kserve/storage/test/test_storage.py
```diff
@@ -303,3 +302,29 @@ def test_unpack_zip_file():
     Storage._unpack_archive_file(tar_file, mimetype, out_dir)
     assert os.path.exists(os.path.join(out_dir, "model.pth"))
     os.remove(os.path.join(out_dir, "model.pth"))
+
+
+def test_download_hf():
```
yuzisun: move this test to `test_hf_storage.py`

python/kserve/pyproject.toml
```diff
@@ -61,6 +66,8 @@ azure-storage-blob = { version = "^12.10.0", optional = true }
 azure-storage-file-share = { version = "^12.7.0", optional = true }
 azure-identity = { version = "^1.8.0", optional = true }
 boto3 = { version = "^1.21.0", optional = true }
+transformers = { version = "^4.37.0", optional = true }
+torch = { version = "^2.3.0+cpu", optional = true, source = "pytorch"}
```
yuzisun: I think the hf storage can be standalone and does not need to depend on kserve package

.github/workflows/e2e-test.yml
lizzzcai: Hi @andyi2it , to just download huggingface repo, I think [huggingface_hub](https://github.com/huggingface/huggingface_hub) should be enough, and the size should be acceptable. In this case, I think we don't need to split the image.

python/kserve/kserve/storage/storage.py
lizzzcai: Hi @andyi2it , as mentioned above, [snapshot_download](https://huggingface.co/docs/huggingface_hub/v0.24.2/en/package_reference/file_download#huggingface_hub.snapshot_download) sdk will be the better option to download the model artifacts.

python/kserve/kserve/storage/storage.py
```diff
@@ -278,8 +277,29 @@ def _download_s3(uri, temp_dir: str) -> str:
...
+        components = uri.split("://")[1].split("/")
+
+        repo = components[0]
+        model, _, hash_value = components[1].partition(":")
+        revision = hash_value if hash_value else None
+
+        tokenizer = AutoTokenizer.from_pretrained(f"{repo}/{model}", revision=revision)
+        model_config = AutoConfig.from_pretrained(f"{repo}/{model}", revision=revision)
+        model = AutoModel.from_config(model_config)
```
juhyeon-cha: custom code models must set the trust_remote_code argument to true.
need to use either the inference service's model arguments or the HF hub model API.

docs/samples/storage/hf/hf_secret.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
spolti, line 1: isn't better to call this file `hf_storage_config.yaml`?

pkg/credentials/hf/hf_secret.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+const (
+	HFTokenKey = "HF_TOKEN"
+)
+
+func BuildSecretEnvs(secret *v1.Secret) []v1.EnvVar {
+	envs := make([]v1.EnvVar, 0)
+
+	if token, ok := secret.Data[HFTokenKey]; ok {
```
spolti, line 30: Isn't this env mandatory?
Maybe we could print a warning or even panic.

python/hf-storage-initializer.Dockerfile
```diff
@@ -0,0 +1,52 @@
+ARG PYTHON_VERSION=3.9
+ARG BASE_IMAGE=python:${PYTHON_VERSION}-slim-bullseye
+ARG VENV_PATH=/prod_venv
+
+FROM ${BASE_IMAGE} as builder
+
+# Install Poetry
+ARG POETRY_HOME=/opt/poetry
+ARG POETRY_VERSION=1.7.1
```
spolti: I think we should use 1.8.2

python/storage-initializer.Dockerfile
```diff
@@ -48,6 +48,8 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN useradd kserve -m -u 1000 -d /home/kserve
 
+RUN apt-get -y update && apt-get install ncdu
```
spolti: why is this needed?

python/hf-storage-initializer.Dockerfile
spolti: a question, can't the existing storage initializer image be reused?
Thinking that users would need to switch manually the initializer images, if so, how should it be done?

andyi2it: By default the non hf storage initializer's image will be pulled. If the storage uri contains hf prefix the hf-storage-initializer image will be pulled. The end user needn't switch manually.

The reason why we have 2 images is because hf storage initializer image is twice the size.

andyi2it: @spolti ☝️

docs/samples/storage/hf/hf_secret.yaml
andyi2it: We were following the same sample pattern as s3.

pkg/credentials/hf/hf_secret.go
andyi2it: I suppose this secret is mandatory only in case of pulling private HuggingFace repository.

python/hf-storage-initializer.Dockerfile
lizzzcai: is it possible to reduce the image size by using some other light-weighted sdk? adding a new image means the operation team needs to maintain a new image(security scan etc) just for models from HF.

spolti: okay, thanks.

spolti: btw, I am working on this https://github.com/kserve/kserve/issues/3489
it is related to it.

the plan is to decouple the storage package from KServe and make a hard copy of the logging.py file, so the storage-initializer can be completely independent from KServe avoiding importing/installing a few unneeded dependencies, like ray and others.
Like

python/kserve/kserve/storage/storage.py
```diff
@@ -284,8 +283,26 @@ def _download_s3(uri, temp_dir: str) -> str:
...
+        from huggingface_hub import snapshot_download
+
+        components = uri.split("://")[1].split("/")
+
+        repo = components[0]
+        model, _, hash_value = components[1].partition(":")
+        revision = hash_value if hash_value else None
+
+        snapshot_download(
+            repo_id=f"{repo}/{model}", revision=revision, local_dir=temp_dir
```
lizzzcai, line 314: HF repo usually contains multiple formats, for example, `safetensors`, `bin` etc. Most of the case, only one should be downloaded to reduce the download time (should save half of the time). 

Usually I will add the allow/ignore patterns, for example:

```python
# here safetensors is the preferred format.
allow_patterns = ["*.json", "*.safetensors", "*.model"]
ignore_patterns = ["*.msgpack", "*.h5", "*.bin"]

output_dir = snapshot_download(
      repo_id=model_name,
      revision=revision,
      allow_patterns=allow_patterns,
      ignore_patterns=ignore_patterns,
      local_dir=models_path,
)
```
I think it is quite needed, but I am not sure the best way to pass it, for example via the `ALLOW_PATTERNS` and `IGNORE_PATTERNS` in the secret?

lizzzcai:
/lgtm
/assign @yuzisun 
the initial support for hf model download looks good to me. for the allow/ignore patterns, can be done in another PR.

python/kserve/kserve/storage/storage.py
```diff
@@ -284,8 +283,26 @@ def _download_s3(uri, temp_dir: str) -> str:
                 temp_dir = Storage._unpack_archive_file(target, mimetype, temp_dir)
         return temp_dir
 
+    @staticmethod
+    def _download_hf(uri, temp_dir: str) -> str:
+        from huggingface_hub import snapshot_download
+
+        components = uri.split("://")[1].split("/")
```
yuzisun: @andyi2it Can you add a check here to validate the hf url  ?

python/kserve/kserve/storage/test/test_hf_storage.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest.mock as mock
+
+from kserve.storage import Storage
+
+
+@mock.patch("huggingface_hub.snapshot_download")
+def test_download_model(mock_snapshot_download):
```
yuzisun, line 22: Add a test for invalidate hf url

yuzisun (approved):

PR #3587, opened 2024-04-09
Fix model unload in server stop method

sivanantha321:
Fixes model unload in model server stop method

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

### DIFF
--- a/python/kserve/kserve/model_server.py
+++ b/python/kserve/kserve/model_server.py
@@ -326,8 +326,8 @@ class ModelServer:
         if self._grpc_server:
             logger.info("Stopping the grpc server")
             await self._grpc_server.stop(sig)
-        for _, model in self.registered_models.get_models().items():
-            model.unload()
+        for model_name in list(self.registered_models.get_models().keys()):
+            self.registered_models.unload(model_name)
 
     def register_exception_handler(
         self,


spolti:
Hi @sivanantha321, is there a issue or something that describes the problem or adds a background to it?
In this case, any loaded model remains there indefinitely?

sivanantha321:
> Hi @sivanantha321, is there a issue or something that describes the problem or adds a background to it? In this case, any loaded model remains there indefinitely?

There isn't a issue for this. In the code `model.unload()` is called but model does not have unload method. I just fixed that and regarding your question about model remaining in memory, I haven't tested that. But I believe that is not the case.

spolti:
> > Hi @sivanantha321, is there a issue or something that describes the problem or adds a background to it? In this case, any loaded model remains there indefinitely?
> 
> There isn't a issue for this. In the code `model.unload()` is called but model does not have unload method. I just fixed that and regarding your question about model remaining in memory, I haven't tested that. But I believe that is not the case.

I see, thanks.

/lgtm

spolti (approved):

python/kserve/kserve/model_server.py
```diff
@@ -326,8 +326,8 @@ async def stop(self, sig: Optional[int] = None):
         if self._grpc_server:
             logger.info("Stopping the grpc server")
             await self._grpc_server.stop(sig)
-        for _, model in self.registered_models.get_models().items():
-            model.unload()
+        for model_name in list(self.registered_models.get_models().keys()):
```
yuzisun, line 329: if `get_models()` is returning a map then you only need to do following, as the iterator loop through keys by default
```python
for model_name in self.registered_models.get_models():
    self.registered_models.unload(model_name)
```

sivanantha321: Python does not allows modifying the dict while iterating, otherwise we will get a runtime error. For example,
```python
  File "main.py", line 11, in <module>
    for model in map:
RuntimeError: dictionary changed size during iteration
```

terrytangyuan (approved):
/lgtm

PR #3477, opened 2024-02-26
OpenAI schema

tessapham:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3419.

**Type of changes**
Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

PR #3590, opened 2024-04-09
Support OpenAIModel in ModelRepository

grandbora:

Adds `OpenAIModel` support to `ModelRepository`.

In https://github.com/kserve/kserve/pull/3477 we introduced openai endpoints to kserve. These endpoints are registered [when there are any models](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/protocol/rest/openai/config.py#L48-L51) of type `OpenAIModel` in the `ModelRepository`. In order to take advantage of the openai endpoints, ModelRepository should support `OpenAIModel` type.

Fixes #

**Type of changes**

`cd python/kserve && make test`

### DIFF
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -44,7 +44,7 @@ PREDICTOR_V2_URL_FORMAT = "{0}://{1}/v2/models/{2}/infer"
 EXPLAINER_V2_URL_FORMAT = "{0}://{1}/v2/models/{2}/explain"
 
 
-class BaseKserveModel(ABC):
+class BaseKServeModel(ABC):
     """
     A base class to inherit all of the kserve models from.
 
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -104,7 +104,7 @@ class PredictorConfig:
         self.predictor_request_timeout_seconds = predictor_request_timeout_seconds
 
 
-class Model(BaseKserveModel):
+class Model(BaseKServeModel):
     def __init__(self, name: str, predictor_config: Optional[PredictorConfig] = None):
         """KServe Model Public Interface
 
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from abc import ABC
 import inspect
 import time
 from enum import Enum
 from typing import Dict, List, Union, Optional, AsyncIterator, Any
-
 import grpc
 import httpx
 import orjson
@@ -42,6 +42,24 @@ PREDICTOR_URL_FORMAT = "{0}://{1}/v1/models/{2}:predict"
 EXPLAINER_URL_FORMAT = "{0}://{1}/v1/models/{2}:explain"
 PREDICTOR_V2_URL_FORMAT = "{0}://{1}/v2/models/{2}/infer"
 EXPLAINER_V2_URL_FORMAT = "{0}://{1}/v2/models/{2}/explain"
+
+
+class BaseKserveModel(ABC):
+    """
+    A base class to inherit all of the kserve models from.
+
+    This class implements the expectations of model repository and model server.
+    """
+
+    def __init__(self, name: str):
+        """
+        Adds the required attributes
+
+        Args:
+            name: The name of the model.
+        """
+        self.name = name
+        self.ready = False
 
 
 class InferenceVerb(Enum):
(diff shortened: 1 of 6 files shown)


grandbora:
Tagging the people with most knowledge on openai support: @tessapham @cmaddalozzo @yuzisun

python/kserve/kserve/model_repository.py
```diff
@@ -30,7 +31,7 @@ class ModelRepository:
     """
 
     def __init__(self, models_dir: str = MODEL_MOUNT_DIRS):
-        self.models: Dict[str, Union[Model, DeploymentHandle]] = {}
+        self.models: Dict[str, Union[Model, OpenAIModel, DeploymentHandle]] = {}
```
cmaddalozzo: I think ideally we would have a base abstract class that all models inherit from. This may be a much larger change so adding `OpenAIModel` to the `Unions` works for now.

cmaddalozzo (approved):

python/kserve/kserve/model_repository.py
grandbora: Agree. Happy to add the base class in a follow up PR.

python/kserve/kserve/model_server.py
```diff
@@ -226,7 +227,7 @@ def start(self, models: Union[List[Model], Dict[str, Deployment]]) -> None:
         """
         if isinstance(models, list):
             for model in models:
-                if isinstance(model, Model):
+                if isinstance(model, Union[Model, OpenAIModel]):
```
grandbora: I wanted to make this change separately and add unit tests to it. However the model registry logic is too tightly couple with starting the server processes. Hence I couldn't unit test them. Since this change turned out to be small and related, I am appending it to this PR.

grandbora: cc @cmaddalozzo let me know ifyou have feedback on https://github.com/kserve/kserve/pull/3590/commits/b16fc8ddef46ca496343d5a45a562d07834b1ecd

python/kserve/test/test_model_server.py
```diff
@@ -0,0 +1,25 @@
...
+from kserve import ModelServer
+
+
+def test_model_server_start_no_models():
+    server = ModelServer()
+
+    with pytest.raises(RuntimeError) as exc:
+        server.start(models=None)
+
+    assert exc.value.args[0] == "Unknown model collection types"
```
yuzisun: Can we make this error message as a constant ?

python/kserve/kserve/model_server.py
yuzisun: Let's make `isinstance(model, Union[Model, OpenAIModel])` a function so we do not need to change so many places next time.

python/kserve/test/test_model_server.py
grandbora: Addressed

python/kserve/kserve/model_server.py
grandbora: I made it a type: `BaseModelType = Union[Model, OpenAIModel]` Next time we can only update this type.

grandbora: Let me know if this is not addressing the concern.

grandbora:
Looks like some of the test failures are due to different python version I was using when developing. I'll make sure tests pass in python 3.9.

python/kserve/kserve/model.py
```diff
@@ -44,6 +44,24 @@
 EXPLAINER_V2_URL_FORMAT = "{0}://{1}/v2/models/{2}/explain"
 
 
+class BaseKserveModel(ABC):
```
yuzisun: Kserve word is redudant here as it is already in the kserve package

yuzisun:
```suggestion
class BaseKServeModel(ABC):
```

grandbora: `BaseModel` was my first choice, but it conflicts with pydantic's `BaseModel` in too many places. We wouldn't want to deal with aliases. Do you have another suggestion?

yuzisun: I see, this is probably ok I think the base model shouldn’t be used in a lot of the places. Can you help fix the case of KServe ?

grandbora:
I tested this branch with a transformer I ran locally. I was able to get a response from the openai legacy completion endpoint. @yuzisun

python/kserve/kserve/model_repository.py
grandbora: I eventually had to convert the implementation to use a base class. Otherwise openai model was exposed to other components. This then required them to pull the openai dependency. I believe we don't want to impose openai support anywhere else.

PR #3479, opened 2024-02-26
feat: Support customizable deployment spec for RawDeployment mode

terrytangyuan:
Fixes https://github.com/kserve/kserve/issues/3452.

PR #3603, opened 2024-04-15
feat: Support customizable deployment strategy for RawDeployment mode. Fixes #3452

### LINKED ISSUE #3452: Ability to change parameters of of the default `Deployment` created from `ServingRuntime` and `InferenceService`
/kind feature

I'd like to have the ability to modify some of the `Deployment.spec` for the predictor deployment created from  `ServingRuntime` and `InferenceService`. These parameters include `strategy` for adjusting rolling update strategy, `template.spec.securityContext` for adjusting the security context to name a few. The ideal solution would be to override any `Deployment.spec` inside the `ServingRuntime` and `InferenceService`.

Currently, the deployment created by `ServingRuntime` and `InferenceService` use the default Kubernetes deployment values. 

terrytangyuan:
Fixes #3452. Supersedes https://github.com/kserve/kserve/pull/3479


spolti:
Maybe we could migrate the formatting changes to another PR?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -106,6 +107,10 @@ type ComponentExtensionSpec struct {
 	// More info: http://kubernetes.io/docs/user-guide/annotations
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// The deployment strategy to use to replace existing pods with new ones.
+	// +optional
+	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
```
spolti, line 113: defaults to?

terrytangyuan:
> Maybe we could migrate the formatting changes to another PR?

The changes that get reformatted are introduced in this PR

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -503,6 +503,420 @@ var _ = Describe("v1beta1 inference service controller", func() {
 			}
 			Expect(actualHPA.Spec).To(gomega.Equal(expectedHPA.Spec))
 		})
+		It("Should have ingress/service/deployment/hpa created with DeploymentStrategy", func() {
+			By("By creating a new InferenceService with DeploymentStrategy in PredictorSpec")
```
spolti, line 507: can't be this test broken into smaller `By` steps?

spolti:
> > Maybe we could migrate the formatting changes to another PR?
> 
> The changes that get reformatted are introduced in this PR

even the jupyter notebook ones?

terrytangyuan:
> > > Maybe we could migrate the formatting changes to another PR?
> 
> > 
> 
> > The changes that get reformatted are introduced in this PR
> 
> 
> 
> even the jupyter notebook ones?

Oh didn't notice those. I'll remove them.

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
terrytangyuan: Nope. It would require refactoring existing tests

pkg/apis/serving/v1beta1/component.go
terrytangyuan: It's optional and defaults to whatever appsv1.DeploymentStrategy has.

terrytangyuan:
@spolti Actually these notebook changes will have to be part of this PR to pass the CI build.

spolti (approved):

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -77,6 +77,9 @@ func createRawDeployment(componentMeta metav1.ObjectMeta,
 			},
 		},
 	}
+	if componentExt.DeploymentStrategy != nil {
+		deployment.Spec.Strategy = *componentExt.DeploymentStrategy
```
Jooho, line 81: From my understanding, rawDeployment is using HPA as the default autoscaler. This means that HPA will overwrite deployment replicas. So I guess hpa and this controller will continuously try to update this deployment, what do you think?

terrytangyuan: It should be fine since HPA will only update the replicas field. Here we are only modifying the deployment strategy so it won't be overwritten.

Jooho: Ok it only changed strategy so this source is ok.

pkg/apis/serving/v1beta1/component.go
yuzisun, line 113: What happens when this field is set for the Knative mode?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -106,6 +107,10 @@ type ComponentExtensionSpec struct {
 	// More info: http://kubernetes.io/docs/user-guide/annotations
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// The deployment strategy to use to replace existing pods with new ones.
```
yuzisun: Can we add a note that this field is only applicable for raw deployment mode

pkg/apis/serving/v1beta1/component.go
terrytangyuan: This is only used in `NewDeploymentReconciler.createRawDeployment` so it won't affect knative mode.

pkg/apis/serving/v1beta1/component.go
terrytangyuan: Done

docs/samples/client/kfserving_sdk_v1beta1_sample.ipynb
```diff
@@ -22,7 +22,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from kubernetes import client \n",
+    "from kubernetes import client\n",
```
yuzisun, line 25: Next time let's separate these changes into a separate PR which can make PR review easier

pkg/apis/serving/v1beta1/component.go
yuzisun: the field is still exposed on the isvc which user can set for the knative mode, but I guess knative probably validate this.

docs/samples/client/kfserving_sdk_v1beta1_sample.ipynb
terrytangyuan: Somehow I cannot avoid it (haven't figured out why yet). These changes happen automatically by black+pre-commit and the build would fail without including the changes.

pkg/apis/serving/v1beta1/component.go
yuzisun: I think we need to add a validation that this field is not supported for knative otherwise user will expect this to work but knative performs blue/green deployment.

bmopuri:
@terrytangyuan   Thank you for making these changes.  I see these deployment strategy targets inference services in raw deployment mode.  Starting with kserve v12 we support raw deployment for inference graphs as well.

Could you please extend this change to inference graph raw deployment as well?

pkg/apis/serving/v1beta1/component.go
terrytangyuan: Added

terrytangyuan:
>Could you please extend this change to inference graph raw deployment as well?

I am trying to narrow down the scope of this PR. Extending this to IG would be a separate PR.

PR #3604, opened 2024-04-15
build: Allow pre-commit to keep changes in reformatted code

terrytangyuan:
Removing the two args `--check, --diff` that prevent `pre-commit` from writing the changes back:

```
  --check                         Don't write the files back, just return the
                                  status. Return code 0 means nothing would
                                  change. Return code 1 means some files would
                                  be reformatted. Return code 123 means there
                                  was an internal error.
  --diff                          Don't write the files back, just output a
                                  diff to indicate what changes Black would've
                                  made. They are printed to stdout so

```

Writing the changes back would be more convenient since developers don't have to run `black` commands locally. They only need to do `git commit ...` once `pre-commit` is installed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
     rev: 24.3.0
     hooks:
       - id: black
-        args: [--check, --diff, --config, python/pyproject.toml]
+        args: [--config, python/pyproject.toml]
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.2.3
     hooks:


spolti:
WHen CI fails because lint, will it show what is different?
Or it this case, it is not importante as now it will write things?

terrytangyuan:
An example:
```
pre-commit run
Helm Docs............................................(no files to check)Skipped
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted python/huggingfaceserver/huggingfaceserver/async_generate_stream.py

All done! ✨ 🍰 ✨
1 file reformatted.

Flake8...................................................................Passed
```

No diff will be shown but I don't think it's needed either since they will be fixed automatically.

You can see the diff via `git diff`.

spolti (approved):

PR #3605, opened 2024-04-15
Added secure gRPC server option to kserve

cbalbera-imaige:
This PR gives KServe the ability to use gRPC's built-in support for ssl/tls to both authenticate requests & encrypt data.

N/A

**Type of changes**:
New feature (non-breaking change which adds functionality)

- Existing test suite runs unchanged, as desired (as the option for secure gRPC should not impact existing - insecure - gRPC channels).  gRPC vs. HTTP server options are not currently scoped in test.

**Checklist**:

- [X] Have you added unit/e2e tests that prove your fix is effective or that this feature works? - covered by existing unit tests (all passing)
- [X] Has code been commented, particularly in hard-to-understand areas? - yes
- [ ] Have you made corresponding changes to the documentation? - I have not updated any README docs, as the gRPC functionality is not noted in any docs as it currently exists

**Release note**:

```release-note
Added option for gRPC requests to KServe to use gRPC's built-in dual authentication/encryption functionality that leverages SSL/TLS.  This release adds the option (which is excluded by default) to use this by passing a flag that the secure option is desired as well as the required certs to the ModelServer class via either the command line or constructor when it is started.  Note that the existing corresponding InferenceServerClient class already supports passing SSL/TLS certs to make authenticated & encrypted requests, so this update to the server class is sufficient to enable fully authenticated & encrypted channels.
```


python/kserve/kserve/model_server.py
```diff
@@ -20,7 +21,7 @@
 import socket
 import sys
 from multiprocessing import Process
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, Callable, Any, IO
```
spolti, line 24: maybe keed in order?

python/kserve/kserve/model_server.py
```diff
@@ -154,21 +173,26 @@
...
-        enable_grpc: bool = args.enable_grpc,
-        enable_docs_url: bool = args.enable_docs_url,
-        enable_latency_logging: bool = args.enable_latency_logging,
-        configure_logging: bool = args.configure_logging,
-        log_config: Optional[Union[Dict, str]] = args.log_config_file,
-        access_log_format: str = args.access_log_format,
-    ):
+
+    def __init__(self, http_port: int = args.http_port,
+                 grpc_port: int = args.grpc_port,
```
spolti, line 178: formatting issue?

python/kserve/kserve/model_server.py
```diff
@@ -200,10 +227,46 @@ def __init__(
...
         self._rest_server = None
+        self.secure_grpc_server = secure_grpc_server
+        self.grpc_ssl_key = server_key
+        self.grpc_ssl_cert = server_cert
+        self.grpc_ssl_ca_cert = ca_cert
         if self.enable_grpc:
-            self._grpc_server = GRPCServer(
-                grpc_port, self.dataplane, self.model_repository_extension
-            )
+            if self.secure_grpc_server:
```
spolti, line 235: Maybe we can move these validations to a util file?
certs_utils?

python/kserve/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "kserve"
-version = "0.12.0"
+version = "0.13.0"
```
spolti, line 3: was this intentional?

python/kserve/kserve/protocol/grpc/server.py
```diff
@@ -58,8 +63,18 @@ async def start(self, max_workers):
             inference_servicer, self._server
         )
 
-        listen_addr = f"[::]:{self._port}"
-        self._server.add_insecure_port(listen_addr)
+        listen_addr = f'[::]:{self._port}'
+
+        if self._secure_server:
+            server_credentials = grpc_ssl_server_credentials(
```
spolti, line 69: can we have a test for this?

python/kserve/kserve/model_server.py
cbalbera-imaige: Can keep in order - thanks!

python/kserve/kserve/model_server.py
cbalbera-imaige: ah, good catch, I'll fix that

python/kserve/pyproject.toml
cbalbera-imaige: ah, good catch - have been updating this locally so the package that depends on my updates to kserve pulls my version rather than the latest (0.12.0) from pypi.  will revert.

python/kserve/kserve/protocol/grpc/server.py
cbalbera-imaige: I hadn't seen any existing tests for the server start functionality (neither gRPC or REST), so was unsure what the scope of testing was supposed to be here.  can write a few if that would be best, let me know

python/kserve/kserve/model_server.py
cbalbera-imaige: this would be fine.  could put them in the regular utils file (kserve/utils/utils.py) if you would like - let me know.

python/kserve/kserve/protocol/grpc/server.py
spolti: One simple test could be done by creating the certificates and spawn a ssl protect server and try to make any request.

python/kserve/kserve/model_server.py
spolti: That's sounds good, but let wait to hear from others.

spolti:
You might need to sign your commit.

python/kserve/kserve/model_server.py
cbalbera-imaige: did not realize there's already a creds_utils file, going to put them in there

python/kserve/kserve/protocol/grpc/server.py
cbalbera-imaige: Going to test this - as well as the helpers - by creating a folder with a set of dummy SSL certs for testing purposes.  Just noting this for posterity (& will document as well) because they will eventually expire and require refreshing

PR #3607, opened 2024-04-16
Add option for returning probabilities in huggingface server

### LINKED ISSUE #3570: Add option to skip postprocess in huggingface server
/kind feature

Add option to skip the default `postprocess` method of huggingface server, as users may want to postprocess the raw prediction results their own way, which may not be possible right now.

For example, let's take the `sequence_classification` task:
https://github.com/kserve/kserve/blob/f6782431cd0751b3dc504fd51d56da75e6e12027/python/huggingfaceserver/huggingfaceserver/model.py#L352-L358

Currently, the `postprocess` makes it output only the predicted idx (i.e. `0`, or `1`). So let's say if a user wants to get the probabilities of each idx (i.e. `{0: 0.61, 1: 0.39}`), it is not possible even if they implement their own `postprocess`, because the default `postprocess` will run first.

andyi2it:
Add flag to return raw prediction results

Fixes #3570

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```


python/huggingfaceserver/huggingfaceserver/test_model.py
```diff
@@ -203,6 +217,21 @@ async def test_bert_sequence_classification(bert_base_yelp_polarity):
     assert response == {"predictions": [1, 1]}
 
 
+@pytest.mark.asyncio
+async def test_bert_sequence_classification_return_probabilities(bert_base_return_prob):
```
yuzisun, line 269: Can you also help add the test case for token classification and mask filling?

PR #3615, opened 2024-04-19
Enabled the multiple domains support on an inference service

### LINKED ISSUE #2747: Allow setting multiple domains on an inference service
/kind feature

Currently KServe supports setting only a single external host for inference services. It would be useful if we could support multiple hosts.

We could enable this feature on a global level or on a per inference service basis.

#### Globally:
Add a section to the ingress config in the `inferenceservice-config` configmap that allows users to configure additional ingress domains. e.g.

```yaml
  ingress: |-
    {
        "ingressGateway" : "istio-system/ingress-gateway",
        "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
        "localGateway" : "istio-system/local-gateway",
        "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
        "ingressDomain"  : "example.com",
        "additionalIngressDomains": [
             "other-example.com" 
        ]
    }
```

#### Per inference service
A user can specify an "alias" hostname for the inference service. This will be used alongside the automatically generated domain.

```yaml
kind: InferenceService
metadata:
  name: test-model
spec:
  predictor:
    # predict definition...
  hostAliases:
    - test-model.other-example.com
```

For serverless mode these hosts will be added to the `VirtualService` that KServe creates. This means that only the canonical name of the inference service will have the alias domain attached. i.e. the predictor, explainer, and transformer specific hosts will not be created for the alias domains.

houshengbo:
Currently KServe supports setting only a single external host for inference services. After enabling the multi-host support, we can facilitate the use cases like migrating the inference services from one to another. This PR enables this feature on a global level with the field `additionalIngressDomains` in the configMap `inferenceservice-config`.

Fixes #2747

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Enable the multiple domains support on an inference service.
```


pkg/apis/serving/v1beta1/configmap_test.go
```diff
@@ -87,10 +87,29 @@ func TestNewIngressConfig(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 	clientset := fakeclientset.NewSimpleClientset(&v1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{Name: constants.InferenceServiceConfigMapName, Namespace: constants.KServeNamespace},
+		Data: map[string]string{
+			IngressConfigKeyName: `{
+				"ingressGateway": "knative-serving/knative-ingress-gateway",
+				"ingressService": "test-destination",
+				"localGateway": "knative-serving/knative-local-gateway",
+				"localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local",
+                "urlScheme": "https",
```
spolti: missing indentation

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -432,6 +432,11 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
 		})
 		// Include ingressDomain to the domains (both internal and external) derived by KNative
 		hosts = append(hosts, url.Host)
+
+		// Include additional ingressDomain to the domains (both internal and external)
+		if config.AdditionalIngressDomains != nil && len(*config.AdditionalIngressDomains) > 0 {
```
spolti: Guess we need to do some validation here before append.
We could relay in the k8s validator api, or, does istio validates it?

houshengbo: This one should work the same as `ingressConfig.IngressDomain`. We just read the values from the CM and set them in this `config`. I did not find any validation for other fields. Do you see validation for other fields?

houshengbo: We will create or update the ingress returned by `createIngress`. If there is validation issue, the creation or update will fail. That's my assumption, but need to verify.

sukumargaonkar: would help to verify if the values passed in additionalIngressDomain are valid URLs. This includes their length and presence of special characters

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -431,7 +431,25 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
...
+	s := strings.Split(serviceHost, isvc.Namespace)
+	if len(s) > 1 && config.AdditionalIngressDomains != nil && len(*config.AdditionalIngressDomains) > 0 {
+		// len(s) > 1 means serviceHost contains the namespace.
+		// If the list of the additionalIngressDomains is not empty, we will append the valid host created by the
+		// additional ingress domain.
+		prefix := fmt.Sprintf("%s%s", s[0], isvc.Namespace)
+		for _, domain := range *config.AdditionalIngressDomains {
+			host := fmt.Sprintf("%s.%s", prefix, domain)
+			if !IsValidHostURL(fmt.Sprintf("%s://%s", config.UrlScheme, host)) {
+				log.Error(fmt.Errorf("The domain name %s in the additionalIngressDomains is not a valid domain name.", domain),
```
spolti:
```suggestion
				log.Error(fmt.Errorf("The domain name %s in the additionalIngressDomains is not valid", domain),
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
```diff
@@ -62,3 +63,8 @@ func GenerateDomainName(name string, obj metav1.ObjectMeta, ingressConfig *v1bet
 
 	return buf.String(), nil
 }
+
+func IsValidHostURL(hostURL string) bool {
+	_, err := url.ParseRequestURI(hostURL)
```
spolti: there is a package called `validation` in the apimachinery tool that could help with it.

IIRC: 
```go
validation.IsDNS1123Subdomain()
```

That might be useful here, as we are validating domains for ingress controller.

wdyt @israel-hdez @yuzisun ?

houshengbo: I change the way for us to validate the domain or subdomain.
The tricky part is how to get the subdomain. The safest way is to get the domain name for the existing hosts that knative serving generated, remove it and append the new additional ingress domain to get the new hosts.

pkg/apis/serving/v1beta1/configmap_test.go
```diff
@@ -87,10 +87,29 @@ func TestNewIngressConfig(t *testing.T) {
...
+				"ingressDomain": "example.com",
+				"additionalIngressDomains": ["other-example.com", "other-example-second.com"]
+			}`,
+		},
 	})
 	ingressCfg, err := NewIngressConfig(clientset)
 	g.Expect(err).Should(gomega.BeNil())
 	g.Expect(ingressCfg).ShouldNot(gomega.BeNil())
+
+	g.Expect(ingressCfg.IngressGateway).To(gomega.Equal("knative-serving/knative-ingress-gateway"))
```
alexagriffith: some of these values are hard coded multiple times like "knative-serving/knative-ingress-gateway", is there a constant you can refer to or do you think it would be helpful to make them?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -53,13 +57,15 @@ var (
...
+	clientset     kubernetes.Interface
 	scheme        *runtime.Scheme
 	ingressConfig *v1beta1.IngressConfig
 }
 
-func NewIngressReconciler(client client.Client, scheme *runtime.Scheme, ingressConfig *v1beta1.IngressConfig) *IngressReconciler {
+func NewIngressReconciler(client client.Client, clientset kubernetes.Interface, scheme *runtime.Scheme, ingressConfig *v1beta1.IngressConfig) *IngressReconciler {
 	return &IngressReconciler{
 		client:        client,
+		clientset:     clientset,
```
alexagriffith, line 70: may be nice to add a note as to why we want to use both client and clientset, just reading the code it can be confusing to see, but (correct me if im wrong) its because client/clientset have access to different functions/resources

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -434,6 +440,32 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
...
+			if found {
+				subdomain = res
+				break
+			}
+		}
+	}
+	if len(subdomain) != 0 && config.AdditionalIngressDomains != nil && len(*config.AdditionalIngressDomains) > 0 {
+		// len(subdomain) != 0 means we have found the subdomain.
+		// If the list of the additionalIngressDomains is not empty, we will append the valid host created by the
+		// additional ingress domain.
```
alexagriffith: +1 to comment here

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -65,17 +65,18 @@ type InferenceServicesConfig struct {
...
-	DisableIstioVirtualHost bool    `json:"disableIstioVirtualHost,omitempty"`
-	PathTemplate            string  `json:"pathTemplate,omitempty"`
-	DisableIngressCreation  bool    `json:"disableIngressCreation,omitempty"`
+	IngressGateway           string    `json:"ingressGateway,omitempty"`
+	IngressServiceName       string    `json:"ingressService,omitempty"`
+	LocalGateway             string    `json:"localGateway,omitempty"`
+	LocalGatewayServiceName  string    `json:"localGatewayService,omitempty"`
+	IngressDomain            string    `json:"ingressDomain,omitempty"`
+	IngressClassName         *string   `json:"ingressClassName,omitempty"`
+	AdditionalIngressDomains *[]string `json:"additionalIngressDomains,omitempty"`
```
alexagriffith, line 74: im not sure if this applies here but do you need to also update the charts/ file for helm releases like here https://github.com/kserve/kserve/blob/cefdc2ecff9764b318d975f1bfa31e71c6874362/charts/kserve-resources/templates/configmap.yaml#L491 or anything? I ran into this issue once, but im not sure if it applies here..

I just did a search for the variable above it `IngressClassName` to get an idea of all the files it is in https://github.com/search?q=repo%3Akserve%2Fkserve%20IngressClassName&type=code

alexagriffith (approved):
lgtm but im not 100% familiar with this part of the code, may be good to get another review as well!

pkg/apis/serving/v1beta1/configmap.go
houshengbo: You were correct. I did not cover the chart.

pkg/apis/serving/v1beta1/configmap_test.go
houshengbo: I will make it neater.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
houshengbo: Yes, I will add, since two different clients are used to access different types of resources.

yuzisun: I think we only need to use controller runtime client as we do not use lower level clientset across kserve project.

charts/kserve-resources/values.yaml
```diff
@@ -35,6 +35,8 @@ kserve:
     rbacProxyImage: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
     gateway:
       domain: example.com
+      additionalIngressDomains:
+        - example.com
```
yuzisun: The default should be empty, otherwise we get duplicated entries

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler_test.go
```diff
@@ -641,6 +643,128 @@ func TestCreateVirtualService(t *testing.T) {
 					},
 				},
 			},
+		}, {
+			name: "found predictor status with path template and the additional ingress domains",
```
yuzisun: path template is used for url path based routing, it does not need the additional domains which is only for host based routing

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler_test.go
```diff
@@ -641,6 +643,128 @@ func TestCreateVirtualService(t *testing.T) {
...
+							Match: []*istiov1beta1.HTTPMatchRequest{
+								{
+									Uri: &istiov1beta1.StringMatch{
+										MatchType: &istiov1beta1.StringMatch_Prefix{
+											Prefix: fmt.Sprintf("/serving/%s/%s/", namespace, serviceName),
+										},
+									},
+									Authority: &istiov1beta1.StringMatch{
+										MatchType: &istiov1beta1.StringMatch_Regex{
+											Regex: constants.HostRegExp("my-domain.com"),
```
yuzisun, line 750: The additional domains also should appear in the matches otherwise it won't work

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -185,6 +185,7 @@ data:
            "localGateway" : "knative-serving/knative-local-gateway",
            "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
            "ingressDomain"  : "example.com",
+           "additionalIngressDomains": ["additional-example.com", "additional-example-1.com"],
```
yuzisun, line 188: also need to update the configmap in kustomize directory
https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -453,6 +487,26 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
...
+// getDomainList gets all the available domain names available with Knative Serving.
+func getDomainList(clientset kubernetes.Interface) *[]string {
+	res := new([]string)
+	ns := constants.DefaultNSKnativeServing
+	if namespace := os.Getenv(system.NamespaceEnvKey); namespace != "" {
+		ns = namespace
+	}
+
+	// Leverage the clientset to access the configMap to get all the available domain names
+	configMap, err := clientset.CoreV1().ConfigMaps(ns).Get(context.TODO(),
```
yuzisun, line 529: You can use controllerRuntime client to get the configMap as well, just to be consistent with the rest of code in the kserve repo

spolti (commented):
leaving /lgtm

It sounds good once @yuzisun's are addressed.

israel-hdez:
I quickly scanned the code and looks like it would work.

However, I'm wondering if we could try to rely less on Istio implementation and prefer using Knative resources? Knative already has some features to manage and customize domains.

I'd like to write a proposal for this. If it would be accepted, this PR code may become obsolete.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
houshengbo: c.client does not work in terms of accessing the configmaps.

houshengbo:
Kserve leverages istio as the default ingress, and knative serving is just one of options provisioning the services. It may not be enough if you only think of knative.

yuzisun:
> I quickly scanned the code and looks like it would work.
> 
> However, I'm wondering if we could try to rely less on Istio implementation and prefer using Knative resources? Knative already has some features to manage and customize domains.
> 
> I'd like to write a proposal for this. If it would be accepted, this PR code may become obsolete.

I think we still want an optional routing layer for KServe specific needs, we have implemented features like path based routing and logic for routing to different inference components. The virtual service can be disabled if not needed and in future we can migrate to gateway API so it does not rely on istio.

yuzisun:
Nice work on the first major kserve PR! @houshengbo 
/lgtm
/approve

israel-hdez:
> we have implemented features like path based routing

Hmm... OK, I hadn't this feature in mind.
Sorry for the noise.

PR #3620, opened 2024-04-22
build: Fix CRD copying in generate-install.sh

terrytangyuan:
This was missed in https://github.com/kserve/kserve/pull/3492 and breaks `generate-install.sh` for preparing releases.


spolti (approved):
/lgtm

PR #3635, opened 2024-04-25
adding metadata on requests

### LINKED ISSUE #3634: Add metadata to logger system
/kind feature

Sometimes when logging request/responses using the logger system it might be needed to log extra data. This might be metadata related to the request (location, language, etc)

Would be nice to be able to specify some prefix in the config so that when users send a request with headers containing that prefix the values get sent to the logger in the metadata field

gcemaj:
Attempts to close #3634.
The PR adds a new configuration to the logger system to set a prefix string such that any header in the request that contains the header gets parsed and sent to the cloud event in a new metadata field.

Changes consists of
- Adds a new configuration to the logger field
- Adds a new field in the CE to hold the metadata
- Adds  system to parse request headers and extract name/value pairs containing prefix and passes the map to the CE

Fixes #3634

**Type of changes**

- [x] Added unit test to verify that cloud event gets metadata field based on request headers

**Checklist**:

**Release note**:

```release-note
Add new field (Metadata) in cloud event emmited by the request when using request-logging.
Metadata is populated by passing in headers in the request. Headers containing a prefix defined in the configuration will get parsed and sent along as Metadata.
```


spolti:
I am not sure about it.
Isn't better to dump the request, similar how it is done with Apache?

gcemaj:
> I am not sure about it. Isn't better to dump the request, similar how it is done with Apache?

dump the headers you mean? that would also work, have a usecase where we need some extra metadata to be associated with the request.

gcemaj:
> I am not sure about it. Isn't better to dump the request, similar how it is done with Apache?

could also be an allowlist of headers?

spolti:
I might no be fully understand the reason of this change, is there a specific use case that just dumping the request headers in the logs is not enough?

gcemaj:
> I might no be fully understand the reason of this change, is there a specific use case that just dumping the request headers in the logs is not enough?

security concerns where there might be a access token in the headers

spolti:
> > I might no be fully understand the reason of this change, is there a specific use case that just dumping the request headers in the logs is not enough?
> 
> security concerns where there might be a access token in the headers

I see, maybe just dump the request direct in the logs would be enough, wdyt?

@yuzisun @terrytangyuan

gcemaj:
> > > I might no be fully understand the reason of this change, is there a specific use case that just dumping the request headers in the logs is not enough?
> > 
> > 
> > security concerns where there might be a access token in the headers
> 
> I see, maybe just dump the request direct in the logs would be enough, wdyt?
> 
> @yuzisun @terrytangyuan

i think we can just have a allowlist for the headers? so you can explicitly configure what you want sent over? Does that seem reasonable? If so i can make the change in this PR

spolti:
Let's wait others to review.

lizzzcai:
option to dump/forward headers + allowlist on headers will be something more flexible. prefix may not be so convenient as not all headers have the same prefix.

gcemaj:
> option to dump/forward headers + allowlist on headers will be something more flexible. prefix may not be so convenient as not all headers have the same prefix.

ill make this change

gcemaj:
@lizzzcai mind giving this a review?

pkg/logger/handler_test.go
```diff
@@ -78,7 +78,69 @@ func TestLogger(t *testing.T) {
...
+		g.Expect(b).To(gomega.Or(gomega.Equal(predictorRequest), gomega.Equal(predictorResponse)))
+		_, err = rw.Write(predictorResponse)
+		g.Expect(err).To(gomega.BeNil())
+	}))
+	// Close the server when test finishes
+	defer predictor.Close()
+
+	reader := bytes.NewReader(predictorRequest)
+	r := httptest.NewRequest("POST", "http://a", reader)
+	r.Header.Add("my-meta/foo", "bar")
```
lizzzcai: how does this test case validate that this header is passed and other headers not in the allowlist will be filtered?

cmd/agent/main.go
```diff
@@ -107,6 +107,7 @@ type loggerArgs struct {
 	namespace        string
 	endpoint         string
 	component        string
+	headerAllowList  []string
```
yuzisun: This can be a global configuration as well ?

yuzisun: I think it is a platform config more than a user parameter, for global configuration it can be added here
https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L284

pkg/logger/handler_test.go
gcemaj: fixed this test. Was also getting some errors when passing the metadata so i opted to json encode it

pkg/logger/worker.go
```diff
@@ -96,6 +98,13 @@ func (w *Worker) sendCloudEvent(logReq LogRequest) error {
 	event.SetExtension(ComponentAttr, logReq.Component)
 	event.SetExtension(EndpointAttr, logReq.Endpoint)
 
+	encodedMetadata, err := json.Marshal(logReq.Metadata)
+	fmt.Println(logReq.Metadata, encodedMetadata)
```
yuzisun, line 102: Remove the print line

yuzisun:
```suggestion
```

pkg/apis/serving/v1beta1/inference_service.go
```diff
@@ -60,7 +60,8 @@ type LoggerSpec struct {
 	// - "request": log only request; <br />
 	// - "response": log only response <br />
 	// +optional
-	Mode LoggerType `json:"mode,omitempty"`
+	Mode           LoggerType `json:"mode,omitempty"`
+	HeaderMetadata []string   `json:"headerMetadata,omitempty"`
```
yuzisun: suggest renaming to `headersForPropagation"

yuzisun:
```suggestion
        // Matched headers for propagating to logger cloud events
        // +optional
	HeaderMetadata []string   `json:"headerMetadata,omitempty"`
```

yuzisun (approved):
/lgtm
/approve

PR #3650, opened 2024-04-29
fix: HPA equality check should include annotations

terrytangyuan:
This PR fixes an issue where:

Correct behavior: If it's a new ISVC with the annotation `serving.kserve.io/autoscalerClass: "external"`, no HPA will be created. 

Incorrect behavior: However, if it's an **_existing_** ISVC where HPA is already created with annotation `serving.kserve.io/autoscalerClass: "hpa"`, our users want to change the annotation to "external" and expect that the existing HPA to be deleted.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/autoscaler/autoscaler_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/autoscaler/autoscaler_reconciler.go
@@ -83,10 +83,8 @@ func createAutoscaler(client client.Client,
 	componentExt *v1beta1.ComponentExtensionSpec) (Autoscaler, error) {
 	ac := getAutoscalerClass(componentMeta)
 	switch ac {
-	case constants.AutoscalerClassHPA:
+	case constants.AutoscalerClassHPA, constants.AutoscalerClassExternal:
 		return hpa.NewHPAReconciler(client, scheme, componentMeta, componentExt), nil
-	case constants.AutoscalerClassExternal:
-		return &NoOpAutoscaler{}, nil
 	default:
 		return nil, fmt.Errorf("unknown autoscaler class type: %v", ac)
 	}
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
@@ -149,10 +149,8 @@ func semanticHPAEquals(desired, existing *autoscalingv2.HorizontalPodAutoscaler)
 	var autoscalerClassChanged bool
 	if hasDesiredAutoscalerClass && hasExistingAutoscalerClass {
 		autoscalerClassChanged = desiredAutoscalerClass != existingAutoscalerClass
-	} else {
-		if hasDesiredAutoscalerClass || hasExistingAutoscalerClass {
-			autoscalerClassChanged = true
-		}
+	} else if hasDesiredAutoscalerClass || hasExistingAutoscalerClass {
+		autoscalerClassChanged = true
 	}
 
 	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec) && !autoscalerClassChanged
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
@@ -144,8 +144,18 @@ func (r *HPAReconciler) checkHPAExist(client client.Client) (constants.CheckResu
 }
 
 func semanticHPAEquals(desired, existing *autoscalingv2.HorizontalPodAutoscaler) bool {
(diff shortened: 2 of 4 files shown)


spolti (approved):
/lgtm

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -144,7 +144,8 @@ func (r *HPAReconciler) checkHPAExist(client client.Client) (constants.CheckResu
 }
 
 func semanticHPAEquals(desired, existing *autoscalingv2.HorizontalPodAutoscaler) bool {
-	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec)
+	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec) &&
+		equality.Semantic.DeepEqual(desired.ObjectMeta.Annotations, existing.ObjectMeta.Annotations)
```
yuzisun: Annotations can include `last-applied` annotation which can be always different

terrytangyuan: Updated to only check specific annotation instead.

yuzisun: Can you help explain what the issue is ? I am not sure how the annotation on HPA matters, as if it is set to `external` then it won't create the HPA.

terrytangyuan: If it's a new ISVC with the annotation `serving.kserve.io/autoscalerClass: "external"`, no HPA will be created. 

However, if it's an **_existing_** ISVC where HPA is already created with annotation `serving.kserve.io/autoscalerClass: "hpa"`, our users want to change the annotation to "external" and expect that the existing HPA to be deleted.

yuzisun: But I do not think we have logic to delete the existing HPA

terrytangyuan: Oh good catch. Just updated.

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -140,11 +140,28 @@ func (r *HPAReconciler) checkHPAExist(client client.Client) (constants.CheckResu
...
+	if hasDesiredAutoscalerClass && hasExistingAutoscalerClass {
+		autoscalerClassChanged = desiredAutoscalerClass != existingAutoscalerClass
+	} else if hasDesiredAutoscalerClass || hasExistingAutoscalerClass {
+		autoscalerClassChanged = true
+	}
+
+	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec) && !autoscalerClassChanged
+}
+
+func shouldDeleteHPA(desired *autoscalingv2.HorizontalPodAutoscaler) bool {
```
yuzisun, line 165: I think we should only delete if it is changed from enabled to external otherwise delete throws error

yuzisun: Actually nvm, `shouldDeleteHPA` is called only when HPA exists. But when autoscaler is set to `external` it creates `NoOpAutoscaler` and does not reconcile

yuzisun: have you tested this? I think it will not work

terrytangyuan: Will fix and test it. Marking as draft for now

terrytangyuan: @yuzisun Please take another look. I tested a couple of scenarios and they worked as expected.

spolti (approved):
/lgtm

pkg/controller/v1beta1/inferenceservice/reconcilers/autoscaler/autoscaler_reconciler.go
```diff
@@ -83,10 +83,8 @@ func createAutoscaler(client client.Client,
 	componentExt *v1beta1.ComponentExtensionSpec) (Autoscaler, error) {
 	ac := getAutoscalerClass(componentMeta)
 	switch ac {
-	case constants.AutoscalerClassHPA:
+	case constants.AutoscalerClassHPA, constants.AutoscalerClassExternal:
 		return hpa.NewHPAReconciler(client, scheme, componentMeta, componentExt), nil
-	case constants.AutoscalerClassExternal:
-		return &NoOpAutoscaler{}, nil
```
yuzisun, line 89: So should we delete the NoOpAutoscaler class now if it is no longer used?

PR #3658, opened 2024-04-30
Add headers to predictor exception logging

grandbora:
Adds the headers to exception logging, which includes request and trace ids. This maybe useful in debugging issues.

**Type of changes**

**Checklist**:

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -345,9 +345,10 @@ class Model(BaseKServeModel):
                 predict_url, timeout=self.timeout, headers=predict_headers, content=data
             )
         except Exception as exc:
+            request_id = predict_headers.get("x-request-id", "unknown-request-id")
             logger.error(
                 f"Could not send a request to predictor at url {predict_url} "
-                f"with headers {predict_headers} "
+                f"for {request_id=} "
                 f"due to exception {exc}"
             )
             raise exc
--- a/python/kserve/kserve/model.py
+++ b/python/kserve/kserve/model.py
@@ -347,6 +347,7 @@ class Model(BaseKServeModel):
         except Exception as exc:
             logger.error(
                 f"Could not send a request to predictor at url {predict_url} "
+                f"with headers {predict_headers} "
                 f"due to exception {exc}"
             )
             raise exc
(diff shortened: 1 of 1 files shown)


python/kserve/kserve/model.py
```diff
@@ -347,6 +347,7 @@ async def _http_predict(
         except Exception as exc:
             logger.error(
                 f"Could not send a request to predictor at url {predict_url} "
+                f"with headers {predict_headers} "
```
grandbora: No information other than `Content-Type`, `x-request-id` and `x-b3-traceid` exist in these headers. This should be safe to log.

spolti: I am not sure if it really is all the headers that are added, however, for the Id ones, it might be good to log a small portion of it, if it is big enough to split in smaller pieces.

grandbora: It is actually all the headers added. The UI may not show it, it is defined a few lines above. Nevertheless, this code can change. To be more future proof I pulled the request id from the headers and logged only that.

spolti:
@sivanantha321 can you please take a look?

sivanantha321:
@grandbora Can you help sign-off the commit ?

grandbora:
Signed the commit. Thank you @sivanantha321

python/kserve/kserve/model.py
```diff
@@ -345,8 +345,10 @@ async def _http_predict(
                 predict_url, timeout=self.timeout, headers=predict_headers, content=data
             )
         except Exception as exc:
+            request_id = predict_headers.get("x-request-id", "unknown-request-id")
             logger.error(
                 f"Could not send a request to predictor at url {predict_url} "
+                f"for {request_id=} "
```
yuzisun, line 351:
```suggestion
                f"for {request_id} "
```

yuzisun: Let's be consistent with predictive models here?
https://github.com/kserve/kserve/blob/9c6a6b8f0d8d431ebae828b132097cb8d4ea61ca/python/kserve/kserve/model.py#L161

grandbora: Updated the default value to `N.A.`

grandbora: @yuzisun the equals above was intentional (https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging). Is there a reason not to use it? We are at python 3.9 as far as i see.

yuzisun: Ah ok ye we deprecated 3.8

PR #3671, opened 2024-05-07
upgrade vllm/transformers version

johnugeorge:
Upgrade vllm and transformers version to support latest models


terrytangyuan (approved):
/lgtm

PR #3672, opened 2024-05-07
introduce service configuration at configmap level

spolti:
chore: This PR introduces a new configuration in the inferenceservice-config
	configMap to enable user to set the service `ClusterIP` to `None` to rely on
	headless services.
	With this configuration, users are enabled to rely on third party balancers
	having total control over it.
	The reasoning behind opening a new configuration support is to allow it easier
	to customize any other important flag at the service creation in the future

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -84,6 +85,12 @@ type DeployConfig struct {
 	DefaultDeploymentMode string `json:"defaultDeploymentMode,omitempty"`
 }
 
+type ServiceConfig struct {
+	// ServiceClusterIPNone is a flag to indicate if the service should have a clusterIP set to None.
+	// If the DeploymentMode is Raw, the default value for ServiceClusterIPNone is true when the value is absent.
+	ServiceClusterIPNone *bool `json:"serviceClusterIPNone,omitempty"`
+}
+
```
terrytangyuan: @yuzisun @sivanantha321 any thoughts on this?

terrytangyuan:
test-graph is failing. Maybe push empty commits to see if it's flaky.

config/configmap/inferenceservice.yaml
```diff
@@ -425,7 +425,21 @@ data:
          # ModelMesh https://kserve.github.io/website/master/admin/modelmesh/
          "defaultDeploymentMode": "Serverless"
        }
-     
+    
+     # ====================================== SERVICE CONFIGURATION ======================================
+     # Example
+     service: |-
```
israel-hdez, line 436: I wonder if it isn't better to reuse the `deploy` key, rather than creating a new one just for the Service?

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -131,6 +133,16 @@ func createService(componentMeta metav1.ObjectMeta, componentExt *v1beta1.Compon
...
 		},
 	}
+
+	if serviceConfig != nil {
+		if (serviceConfig.ServiceClusterIPNone == nil && raw) ||
+			(serviceConfig.ServiceClusterIPNone != nil && *serviceConfig.ServiceClusterIPNone) {
+			service.Spec.ClusterIP = corev1.ClusterIPNone
+		}
+	} else if raw {
+		service.Spec.ClusterIP = corev1.ClusterIPNone
```
israel-hdez, line 170: I'm not sure this is a safe default. Clients would need to know how to handle the multiple DNS entries. Also, KServe defaults would diverge from Kubernetes defaults...

Most importantly, I see this as a backwards incompatible change.

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -415,6 +415,20 @@ data:
...
+     # Example
+     service: |-
+       {
+         "serviceClusterIPNone": true
+       }
+     service: |-
+       {
+          # ServiceClusterIPNone is a flag to indicate if the service should have a clusterIP set to None.
+          # This is particularly useful for headless services to allow users to use out of the box balancers.
+          # If the DeploymentMode is Raw, the default value for ServiceClusterIPNone is true
```
israel-hdez:
```suggestion
          # ServiceClusterIPNone indicates if KServe should create  headless services (clusterIP=None).
          # If the DeploymentMode is Raw, the default value for ServiceClusterIPNone is true
```

External LBs is probably just one use case for a headless service. I'd omit such details.

config/configmap/inferenceservice.yaml
spolti: We had defined a new option in the config to not mix the configuration purpose, this affects only the service created, for now only Raw.
But in the future, more fields can be easily added here if needed..

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
spolti: It is the default only when raw deployments and the flag is not explicitly set to false. However, it is just a proposal atm, we need to define first the way we will be handling this configuration

@yuzisun @terrytangyuan

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -84,6 +85,12 @@ type DeployConfig struct {
 	DefaultDeploymentMode string `json:"defaultDeploymentMode,omitempty"`
 }
 
+type ServiceConfig struct {
+	// ServiceClusterIPNone is a boolean flag to indicate if the service should have a clusterIP set to None.
+	// If the DeploymentMode is Raw, the default value for ServiceClusterIPNone is true when the value is absent.
+	ServiceClusterIPNone *bool `json:"serviceClusterIPNone,omitempty"`
```
yuzisun: We probably want another config for the ServiceType, this option only makes sense when type is `ClusterIP`

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -37,27 +37,29 @@ var log = logf.Log.WithName("ServiceReconciler")
...
 		scheme:       scheme,
-		Service:      createService(componentMeta, componentExt, podSpec),
+		Service:      createService(componentMeta, componentExt, podSpec, serviceConfig, raw),
 		componentExt: componentExt,
 	}
 }
 
 func createService(componentMeta metav1.ObjectMeta, componentExt *v1beta1.ComponentExtensionSpec,
-	podSpec *corev1.PodSpec) *corev1.Service {
+	podSpec *corev1.PodSpec, serviceConfig *v1beta1.ServiceConfig, raw bool) *corev1.Service {
```
yuzisun:
```suggestion
	podSpec *corev1.PodSpec, serviceConfig *v1beta1.ServiceConfig, rawDeploymentMode bool) *corev1.Service {
```

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
yuzisun: I think if the service config does not exist we should retain the existing behavior

charts/kserve-resources/templates/configmap.yaml
spolti: I see, okay

config/configmap/inferenceservice.yaml
spolti: Is this ok @israel-hdez

pkg/apis/serving/v1beta1/configmap.go
spolti: for now, it might be okay to keep this way as we define the service as ClusterIP, with this flag the only thing we do is set it to None.

Or, what do you have in mind? Open the configuration for more properties?

config/configmap/inferenceservice.yaml
israel-hdez: To me, the `deploy` is the key that should hold the config, as my thinking is that this is a config related to deploying models. I see first-level keys in this ConfigMap to be high-level fields, rather than low-level resource-focused ones.

But I think I prefer to defer to @yuzisun that decision.

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
spolti: okay, did as you suggested, PTAL.

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
spolti: As it is not needed anymore (as per as the last update)\, I forgot to revert it in my last update, will do it tomorrow morning.

spolti: removed the `raw` parameter.

spolti:
The transformer issue will be addressed as part of https://github.com/kserve/kserve/pull/3758

spolti:
@yuzisun applied your suggestions, can you please take a look again?

spolti:
Hey @yuzisun , in case you have missed the previous message :)

config/configmap/inferenceservice.yaml
yuzisun: sorry for the late reply, I think we this is actually related to the ingress config section. The whole purpose to set ClusterIP to None is that you want to enable your LB right?

spolti: @yuzisun, it might not be related to ingress, in this case, it is just to configure headless services.

> enable your LB right

Yes

spolti:
@terrytangyuan do you think we still need it?

.github/workflows/e2e-test.yml
```diff
@@ -788,11 +788,25 @@ jobs:
...
+          kubectl describe configmaps -n kserve inferenceservice-config
+
       - name: Run E2E tests
         timeout-minutes: 30
         run: |
           ./test/scripts/gh-actions/run-e2e-tests.sh "raw" "6"
 
+      - name: Patch inferenceservice config for cluster ip none
+        run: |
+          kubectl patch configmaps -n kserve inferenceservice-config --patch-file config/overlays/test/configmap/inferenceservice-enable-cluster-ip.yaml
```
israel-hdez, line 802: Is a controller restart needed after patching?
...it is typically needed for other changes to the ConfigMap.

config/configmap/inferenceservice.yaml
israel-hdez: I still would lean more towards `deploy`. Our use case is related to ingress. That said, it could very well be an in-cluster service doing client-side load balancing (e.g. python's [ginepro library](https://docs.rs/ginepro/latest/ginepro/struct.LoadBalancedChannel.html)).

That said, I have no strong opinion.

pkg/apis/serving/v1beta1/configmap.go
israel-hdez: Doesn't it make more sense to have a hard `bool` instead of a pointer?
AFAIK, it will default to `false`. So, if the config is absent, you already will get `false` by default and you no longer need to test for `nil` in the code.

pkg/apis/serving/v1beta1/configmap_test.go
```diff
@@ -110,3 +113,39 @@ func TestNewDeployConfig(t *testing.T) {
...
+
+func TestNewServiceConfig(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
+	// nothing declared
+	empty := fakeclientset.NewSimpleClientset(&v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: constants.InferenceServiceConfigMapName, Namespace: constants.KServeNamespace},
+	})
+	emp, err := NewServiceConfig(empty)
+	g.Expect(err).Should(gomega.BeNil())
+	g.Expect(emp).ShouldNot(gomega.BeNil())
```
israel-hdez, line 125: If you opt for the variable no longer being a pointer, maybe add an assertion for the expected default:
```go
g.Expect(*emp.ServiceClusterIPNone).Should(gomega.BeFalse())
```

test/e2e/logger/test_raw_logger.py
```diff
@@ -30,14 +32,67 @@
...
+                requests={"cpu": "10m", "memory": "128Mi"},
+                limits={"cpu": "100m", "memory": "256Mi"},
+            ),
+        ),
+    )
+    base_test(msg_dumper, service_name, predictor, rest_v1_client)
+
+
+@pytest.mark.rawcipn
+async def test_kserve_logger_cipn(rest_v1_client):
```
israel-hdez, line 68: IMHO, [test_raw_deployment.py](https://github.com/kserve/kserve/blob/master/test/e2e/predictor/test_raw_deployment.py) would had been a better location for the new test.

israel-hdez (approved):
I'm happy with the code. I left some comments, but I don't consider them to be blockers.

We've been testing this more deeply. There is a side effect: since kube-proxy isn't involved when using headless services, clients require to use the open ports on the destination pods (the typical port 80 no longer works). I think a follow-up PR will be required to correctly reflect this in the `status` of the InferenceService... @VedantMahabaleshwarkar may already be working on this.

test/e2e/logger/test_raw_logger.py
spolti: I don't remember exactly the reason that led me to put this test here, I guess it was something related to the logs messages to make sure those are also added for clusterIP.

.github/workflows/e2e-test.yml
spolti: good catch, however, I used the same patch strategy for other tests, e.g. istio: https://github.com/kserve/kserve/pull/3672/files#diff-5cc2c864012004f60ee3f9078fa1c4a8721c6b841e386c57f12bea58b692e6e2R881

config/configmap/inferenceservice.yaml
spolti: I don't have a strong opinion as well, both work for me. Initially, I did this way because, if for another reason more customizations are needed in the service object, this could be used.

pkg/apis/serving/v1beta1/configmap.go
spolti: Good idea.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #3676, opened 2024-05-10
Update Dockerfile and Readme

gavrissh:
Update Dockerfile for HuggingFace and Readme update

### DIFF
--- a/python/huggingfaceserver/README.md
+++ b/python/huggingfaceserver/README.md
@@ -111,7 +111,7 @@ spec:
           cpu: 100m
           memory: 2Gi
 ```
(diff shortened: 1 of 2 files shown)


python/huggingfaceserver/README.md
```diff
@@ -111,7 +111,7 @@ spec:
           cpu: 100m
           memory: 2Gi
 ```
-3. Serve the huggingface model using vllm runtime. vllm is the default runtime. Note - Model need to be supported by vllm otherwise KServe python runtime will be used as a failsafe.
+3. Serve the huggingface model using vllm runtime. Note - Model need to be supported by vllm otherwise KServe python runtime will be used as a failsafe.
```
spolti, line 114: Please align with this PR

https://github.com/kserve/kserve/pull/3677

@alexagriffith fyi

PR #3680, opened 2024-05-12
Fix: model id and model dir check order

yuzisun:
Fix the condition check order, we should always check if `model_id` is None first as `model_dir` can be always specified.
Default `--model_dir` to `/mnt/models`, so user does not need to add this additional argument on inference service yaml. 

Fixes #

**Type of changes**

1. This depends on https://github.com/kserve/kserve/pull/3679 which should be merged first

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -55,7 +55,9 @@ parser.add_argument(
     default="/mnt/models",
     help="A URI pointer to the model binary",
 )
-parser.add_argument("--model_id", required=False, default=None, help="Huggingface model id")
+parser.add_argument(
+    "--model_id", required=False, default=None, help="Huggingface model id"
+)
 parser.add_argument(
     "--model_revision", required=False, default=None, help="Huggingface model revision"
 )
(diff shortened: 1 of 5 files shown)


terrytangyuan (approved):
/lgtm
/hold

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -52,10 +52,12 @@ def list_of_strings(arg):
 parser.add_argument(
     "--model_dir",
     required=False,
-    default=None,
+    default="/mnt/models",
```
ghost, line 55: Wouldn't this be problematic as [L136](https://github.com/kserve/kserve/pull/3680/files#diff-2ed97434a3b43aac01ea6d9f2364d6c176a3716fcff260115226a61260d460f9L133)/[L156](https://github.com/kserve/kserve/blob/c660972ea153a197a483efb23b5c5e036eb4504c/python/huggingfaceserver/huggingfaceserver/__main__.py#L156) would always set `args.model` to `/mnt/models`?
Should the line be then updated to `args.model = args.model_id or args.model_dir`

PR #3688, opened 2024-05-14
Fix:vLLM Model Supported check throwing circular dependency

gavrissh:
After vLLM update to version 4.2
The load of Llama Arch was causing a circular dependency error.

This PR fixes that issue

**Type of changes**

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)

Please describe the tests you ran to verify your changes and the relevant result summary. Provide instructions so it can be reproduced.

- [x] Ran sanity checks on Llama, Mistral, and Gemma arch Models

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/vllm/utils.py
+++ b/python/huggingfaceserver/huggingfaceserver/vllm/utils.py
@@ -44,6 +44,7 @@ def infer_vllm_supported_from_model_architecture(
 
     if architecture not in ModelRegistry.get_supported_archs():
         logger.info("not a supported model by vLLM")
+        return False
     return True
 
 
--- a/python/huggingfaceserver/huggingfaceserver/vllm/utils.py
+++ b/python/huggingfaceserver/huggingfaceserver/vllm/utils.py
@@ -13,11 +13,10 @@
 # limitations under the License.
 
 from argparse import ArgumentParser
-from typing import Any, Optional, Type, Union
+from typing import Any, Union
 from pathlib import Path
 
 from kserve.logging import logger
-from torch import nn
 
 try:
     from vllm.engine.arg_utils import AsyncEngineArgs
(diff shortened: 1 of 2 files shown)


python/huggingfaceserver/huggingfaceserver/vllm/utils.py
```diff
@@ -37,15 +36,15 @@ def vllm_available() -> bool:
 
 def infer_vllm_supported_from_model_architecture(
     model_config_path: Union[Path, str],
-) -> Optional[Type[nn.Module]]:
+) -> bool:
     if not _vllm:
-        return None
+        return False
     model_config = AutoConfig.from_pretrained(model_config_path)
     architecture = model_config.architectures[0]
```
spolti: don't we need to check all archs here?

python/huggingfaceserver/huggingfaceserver/vllm/utils.py
```diff
@@ -37,15 +36,15 @@ def vllm_available() -> bool:
...
-    model_cls = ModelRegistry.load_model_cls(architecture)
-    if model_cls is None:
+
+    if architecture not in ModelRegistry.get_supported_archs():
         logger.info("not a supported model by vLLM")
```
spolti: shouldn't it return false if there is no supported archs?

gavrissh: missed it, Thanks

python/huggingfaceserver/huggingfaceserver/vllm/utils.py
gavrissh: We generally noticed a case of a single arch in response. 
Updated, with changes to check all.

PR #3694, opened 2024-05-16
Add precaution again running v1 endpoints on openai models

grandbora:
# What this PR does
Adds error handling for executing V1 endpoints on OpenAI models.

Kserve allows users to send V1 endpoint requests to OpenAI models. This leads to a crash on the server side. We should gracefully fail on these requests.

Before the PR kserve was returning a `500` error to the user. After the PR kserve is returning a `400` error to the user. 

## Logs
500 err before the PR:
```
curl -v \
    -H "Content-Type: application/json" \
    $ENDPOINT \
    -d \
    '{...}'
*   Trying 0.0.0.0:8080...
* Connected to 0.0.0.0 (127.0.0.1) port 8080
> POST /v1/models/local-test-completion:predict HTTP/1.1
> Host: 0.0.0.0:8080
> User-Agent: curl/8.4.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 329
> 
< HTTP/1.1 500 Internal Server Error
< date: Thu, 16 May 2024 19:37:54 GMT
< server: uvicorn
< content-length: 71
< content-type: application/json
< 
* Connection #0 to host 0.0.0.0 left intact
{"error":"TypeError : 'CompletionsTransformer' object is not callable"}%   
```

400 err after the PR:
```
curl -v \
    -H "Content-Type: application/json" \
    $ENDPOINT \
    -d \
    '{...}'
*   Trying 0.0.0.0:8080...
* Connected to 0.0.0.0 (127.0.0.1) port 8080
> POST /v1/models/local-test-completion:predict HTTP/1.1
> Host: 0.0.0.0:8080
> User-Agent: curl/8.4.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 329
> 
< HTTP/1.1 400 Bad Request
< date: Thu, 16 May 2024 19:35:56 GMT
< server: uvicorn
< content-length: 97
< content-type: application/json
< 
* Connection #0 to host 0.0.0.0 left intact
{"error":"Model local-test-completion is of type OpenAIModel. It does not support infer method."}%                                                                                                                                                                                         
```

### Future Work
PS1. Ideally server should return a 404.
PS2. The vice versa is still not handled gracefully. Sending an openai request to a non openai model creates a 500.
Happy to address both of the above in follow up PRs.

**Type of changes**

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -364,6 +364,11 @@ class DataPlane:
         """
         # call model locally or remote model workers
         model = self.get_model(model_name)
+        if isinstance(model, OpenAIModel):
+            logger.warning(
+                f"Model {model_name} is of type OpenAIModel. It does not support the explain method."
+                " A request exercised this path and will cause a server crash."
+            )
         if isinstance(model, DeploymentHandle):
             response = await model.remote(request, verb=InferenceVerb.EXPLAIN)
         else:
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -364,9 +364,6 @@ class DataPlane:
         """
         # call model locally or remote model workers
         model = self.get_model(model_name)
-        if isinstance(model, OpenAIModel):
-            error_msg = f"Model {model_name} is of type OpenAIModel. It does not support the explain method."
-            raise InvalidInput(reason=error_msg)
         if isinstance(model, DeploymentHandle):
             response = await model.remote(request, verb=InferenceVerb.EXPLAIN)
         else:
--- a/python/kserve/kserve/protocol/dataplane.py
+++ b/python/kserve/kserve/protocol/dataplane.py
@@ -30,6 +30,7 @@ from ..model import InferenceVerb, Model
 from ..model_repository import ModelRepository
 from ..utils.utils import create_response_cloudevent, is_structured_cloudevent
 from .infer_type import InferRequest, InferResponse
+from .rest.openai import OpenAIModel
 
 JSON_HEADERS = [
     "application/json",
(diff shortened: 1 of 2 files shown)


grandbora:
cc @cmaddalozzo @yuzisun

yuzisun:
We can choose to not register v1 endpoints for the model server, but there could be edge cases that people create both `KServeModel` and `OpenAIModel` which need both predictive and generative endpoints, so changing to 400 is ok but let's fix the other case calling `KServeModel` via openai endpoints to be consistent.

grandbora:
I will carve some time to handle the opposite case. In the spirit of small incremental improvements I'd suggest we merge this, unless there is any other feedback.

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -360,6 +364,9 @@ async def explain(
         """
         # call model locally or remote model workers
         model = self.get_model(model_name)
+        if isinstance(model, OpenAIModel):
+            error_msg = f"Model {model_name} is of type OpenAIModel. It does not support the explain method."
+            raise InvalidInput(reason=error_msg)
```
yuzisun: I am not sure about this, but openai does not support explanation so maybe we should leave this out, do want to block people who wants to implement this for `OpenAIModel`

grandbora:
Hi @yuzisun , I removed the check from explain. Though I have these reservations:

* If we want to allow a user to use the `/v1/models/{model_name}:explain` path on an openai model then they will need to implement the explain logic under the [`__call__` method](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model.py#L154-L196). Openai model doesn't come with any of these.

* By leaving this code unchecked, any user can make the server crash with an http request. This is not good practice, it can lead to other issues.

If we want to support this I think the model should come with a stub that returns a `not implemented` error by default.

yuzisun:
> Hi @yuzisun , I removed the check from explain. Though I have these reservations:
> 
> * If we want to allow a user to use the `/v1/models/{model_name}:explain` path on an openai model then they will need to implement the explain logic under the [`__call__` method](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model.py#L154-L196). Openai model doesn't come with any of these.
> * By leaving this code unchecked, any user can make the server crash with an http request. This is not good practice, it can lead to other issues.
> 
> If we want to support this I think the model should come with a stub that returns a `not implemented` error by default.

ok let's log an issue and address in a separate PR

grandbora:
Added a warning log to `explain` method. @yuzisun this is back to you.

grandbora:
Bumping this. @yuzisun whenever you get a chance, please take a look.

PR #3695, opened 2024-05-17
Add FP16 datatype support for OIP grpc

### LINKED ISSUE #3643: Inference gRPC/Rest client to support FP16
/kind feature

Triton client supports FP16, kserve rest/grpc should also support transporting FP16.
https://github.com/kserve/kserve/blob/master/python/kserve/kserve/protocol/infer_type.py#L278-L296

sivanantha321:

Fixes #3643

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Add FP16 datatype support for OIP grpc
```

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/protocol/grpc/servicer.py
+++ b/python/kserve/kserve/protocol/grpc/servicer.py
@@ -22,6 +22,8 @@ from kserve.utils.utils import to_headers
 
 from grpc import ServicerContext
 
+from ...errors import InvalidInput
+
 
 class InferenceServicer(grpc_predict_v2_pb2_grpc.GRPCInferenceServiceServicer):
 
@@ -33,6 +35,21 @@ class InferenceServicer(grpc_predict_v2_pb2_grpc.GRPCInferenceServiceServicer):
         super().__init__()
         self._data_plane = data_plane
         self._mode_repository_extension = model_repository_extension
+
+    @classmethod
+    def validate_grpc_request(cls, request: pb.ModelInferRequest):
+        raw_inputs_length = len(request.raw_input_contents)
+        if raw_inputs_length != 0 and len(request.inputs) != raw_inputs_length:
+            raise InvalidInput(
+                f"the number of inputs ({len(request.inputs)}) does not match the expected number of "
+                f"raw input contents ({raw_inputs_length}) for model '{request.model_name}'."
+            )
+        if raw_inputs_length != 0:
+            for input_ in request.inputs:
+                if input_.HasField("contents"):
+                    raise InvalidInput(
+                        f"contents field must not be specified when using raw_input_contents for input '{input_.name}' for model '{request.model_name}'"
+                    )
 
     async def ServerMetadata(self, request: pb.ServerMetadataRequest, context):
         metadata = self._data_plane.metadata()
@@ -96,6 +113,7 @@ class InferenceServicer(grpc_predict_v2_pb2_grpc.GRPCInferenceServiceServicer):
         self, request: pb.ModelInferRequest, context: ServicerContext
     ) -> pb.ModelInferResponse:
         headers = to_headers(context)
+        self.validate_grpc_request(request)
         infer_request = InferRequest.from_grpc(request)
         response_body, _ = await self._data_plane.infer(
             request=infer_request, headers=headers, model_name=request.model_name
--- a/python/kserve/kserve/protocol/infer_type.py
+++ b/python/kserve/kserve/protocol/infer_type.py
@@ -885,7 +885,11 @@ class InferResponse:
         if _contains_fp16_datatype(self):
             use_raw_outputs = True
         for infer_output in self.outputs:
-            if use_raw_outputs and infer_output.data:
+            if (
+                use_raw_outputs
+                and infer_output.data
+                and isinstance(infer_output.data, list)
+            ):
                 infer_output.data = infer_output.as_numpy()
             if isinstance(infer_output.data, np.ndarray):
                 infer_output.set_data_from_numpy(infer_output.data, binary_data=True)
--- a/python/kserve/kserve/protocol/infer_type.py
+++ b/python/kserve/kserve/protocol/infer_type.py
@@ -15,7 +15,6 @@
 import struct
 from typing import Optional, List, Dict, Union
 
-import numpy
 import numpy as np
 import pandas as pd
 import uuid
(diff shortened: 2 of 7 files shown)


python/kserve/kserve/protocol/infer_type.py
```diff
@@ -562,7 +647,14 @@ def as_numpy(self) -> numpy.ndarray:
         if dtype is None:
             raise InvalidInput("invalid datatype in the input")
         if self._raw_data is not None:
-            np_array = np.frombuffer(self._raw_data, dtype=dtype)
+            if self.datatype == "BYTES":
+                # String results contain a 4-byte string length
+                # followed by the actual string characters. Hence,
+                # need to decode the raw bytes to convert into
+                # array elements.
+                np_array = deserialize_bytes_tensor(self._raw_data)
```
yuzisun, line 652: This is only for the REST case right? for gRPC we can just call np.frombuffer

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -32,43 +31,85 @@
 )
 from ..utils.numpy_codec import to_np_dtype, from_np_dtype
 
+# Dictionary that maps model names to a boolean indicating whether their output uses the FP16 datatype
+_model_name_to_fp16_map: Dict[str, bool] = {}
```
yuzisun: Why this is a global map?

yuzisun:
Let's add an e2e test for this

python/kserve/kserve/protocol/infer_type.py
sivanantha321: removed

python/kserve/kserve/protocol/infer_type.py
sivanantha321: Triton client uses serialization for both http and grpc

python/kserve/kserve/protocol/grpc/servicer.py
```diff
@@ -34,6 +36,21 @@ def __init__(
...
+            raise InvalidInput(
+                f"the number of inputs ({len(request.inputs)}) does not match the expected number of "
+                f"raw input contents ({raw_inputs_length}) for model '{request.model_name}'."
+            )
+        if raw_inputs_length != 0:
+            for input_ in request.inputs:
+                if input_.HasField("contents"):
+                    raise InvalidInput(
+                        f"contents field must not be specified when using raw_input_contents for input '{input_.name}' for model '{request.model_name}'"
+                    )
```
yuzisun, line 52: I think only FP16 and BF16 requires setting raw inputs, for other types user do not have to set raw input contents

sivanantha321: If raw_inputs_contents specified, then the contents field should be empty. https://github.com/kserve/open-inference-protocol/blob/52528cf4c48368114874e99217d25b9e006e494a/specification/protocol/open_inference_grpc.proto#L203-L204

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -173,7 +211,14 @@ def as_numpy(self) -> np.ndarray:
         if dtype is None:
             raise InvalidInput(f"invalid datatype {dtype} in the input")
         if self._raw_data is not None:
-            np_array = np.frombuffer(self._raw_data, dtype=dtype)
+            if self.datatype == "BYTES":
+                # String results contain a 4-byte string length
+                # followed by the actual string characters. Hence,
+                # need to decode the raw bytes to convert into
+                # array elements.
+                np_array = deserialize_bytes_tensor(self._raw_data)
```
yuzisun, line 219: How `deserialize_bytes_tensor` is different from `np.frombuffer` ?

sivanantha321: Usually strings are serialized as a 1-dimensional array containing the 4-byte byte size followed by the actual element bytes. All elements are concatenated together. To deserialize this format we are using deserialize_byte_tensor

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -791,8 +880,14 @@ def to_grpc(self) -> ModelInferResponse:
         """
         infer_outputs = []
         raw_output_contents = []
+        use_raw_outputs = False
+        # If FP16 datatype is present in the outputs use raw outputs.
+        if _contains_fp16_datatype(self):
+            use_raw_outputs = True
```
yuzisun, line 886: What happens if only one of the outputs use FP16?

python/kserve/test/test_grpc_server.py
```diff
@@ -0,0 +1,662 @@
...
+        self.name = name
+        self.ready = False
+
+    def load(self):
+        self.ready = True
+
+    async def predict(self, request, headers=None):
+        outputs = pd.DataFrame(
+            {
+                "fp16_output": request.get_input_by_name("fp32_input")
```
yuzisun, line 81: fp16_input

python/kserve/test/test_grpc_server.py
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+
+@pytest.mark.asyncio
+@patch(
+    "kserve.protocol.grpc.servicer.to_headers", return_value=[]
+)  # To avoid NotImplementedError from trailing_metadata function
+async def test_grpc_raw_inputs(mock_to_headers, server):
+    """
+    If we receive raw inputs then, the response also should be in raw output format.
```
yuzisun, line 302: Is this the right assumption ? wondering how triton behaves.

python/kserve/kserve/protocol/infer_type.py
sivanantha321: The response will be returned as raw outputs

python/kserve/kserve/protocol/infer_type.py
yuzisun: what happens if user set the string dtype np.array bytes to the raw content field?

python/kserve/test/test_grpc_server.py
sivanantha321: Yes, I just followed the triton behaviour

python/kserve/kserve/protocol/infer_type.py
yuzisun: same for output according to 
https://github.com/kserve/open-inference-protocol/blob/52528cf4c48368114874e99217d25b9e006e494a/specification/protocol/open_inference_grpc.proto#L257

sivanantha321: I mean the all the outputs will be sent as bytes which will be stored in raw_ouput_contents field. The contents field will be empty.

python/kserve/test/test_grpc_server.py
```diff
@@ -0,0 +1,662 @@
...
+    """
+    raw_input_contents = [
+        np.array([6.8, 2.8, 4.8, 1.4, 6.0, 3.4, 4.5, 1.6], dtype=np.float32).tobytes(),
+        np.array([6, 2, 4, 1, 6, 3, 4, 1], dtype=np.int32).tobytes(),
+        serialize_byte_tensor(
```
yuzisun: Wondering if this is the same as `ndarray.tobytes`
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html

python/kserve/kserve/protocol/infer_type.py
sivanantha321: serialize_bytes_tensor only used if the strings is sent in raw_input_contents. If we use contents we can just directly provide strings as list of bytes (e.g. [b"cat", b"dog"] )

python/kserve/test/test_grpc_server.py
```diff
@@ -0,0 +1,662 @@
...
+                "parameters": {"binary_data_size": {"int64_param": "64"}},
+            },
+            {
+                "name": "fp16_output",
+                "datatype": "FP16",
+                "shape": ["8"],
+                "parameters": {"binary_data_size": {"int64_param": "16"}},
+            },
+        ],
+        "raw_output_contents": [
```
yuzisun, line 549: Can we add the asserts to decode raw output content to InferResponse/InferOutput so deserializeBytesTensor is tested?

python/kserve/kserve/protocol/infer_type.py
```diff
@@ -791,8 +880,14 @@ def to_grpc(self) -> ModelInferResponse:
...
         for infer_output in self.outputs:
-            if isinstance(infer_output.data, numpy.ndarray):
+            if use_raw_outputs and infer_output.data:
+                infer_output.data = infer_output.as_numpy()
+            if isinstance(infer_output.data, np.ndarray):
                 infer_output.set_data_from_numpy(infer_output.data, binary_data=True)
```
yuzisun: seems like we are decoding back to numpy on line 889 and then serializing to bytes again on line 891. If it is already in bytes then we can simply pass them? For simple data it is fine but might be an overhead for large output tensor

python/kserve/test/test_grpc_server.py
sivanantha321: Added

python/kserve/kserve/protocol/infer_type.py
sivanantha321: This case is for list data. Added type check for list

PR #3678, opened 2024-05-10
Update huggingface readme

alexagriffith:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

PR #3698, opened 2024-05-18
 Merge changes from master to release-0.13 branch 

johnugeorge:
All fixes from master are pulled to release-0.13 branch


johnugeorge:
@yuzisun There is a commit in https://github.com/kserve/kserve/pull/3678 which is unsigned. I am wondering how did this get into master without DCO check?

yuzisun:
> @yuzisun There is a commit in #3678 which is unsigned. I am wondering how did this get into master without DCO check?

I probably did not notice that and manually merged it. The auto merge stopped working for some reason, will need to look into that.

PR #3711, opened 2024-05-22
Prevent the PassthroughCluster for clients/workloads in the service mesh

israel-hdez:
# What this PR does / why we need it

## Background
The KServe Ingress VirtualServices are created with configurations targeting only the Gateways. Although this works, the omission of the Istio sidecars has the following downsides for client/workloads that belong to the Istio mesh (i.e. have an Istio sidecar):

* Requests to InferenceServices will be treated as going to external services (i.e. not part of the mesh), because the sidecars are unaware of the routing rules.
* In consequence, the requests will be hanlded as with any external (non-mesh) workload: the ingress gateway will first receive the request and will forward it to itself doing the URL rewrite to the relevant `-predictor`, `-explainer` or `-transformer` hostname. Such forwarding can be avoided (for mesh-workloads) and the rewrite can be performed by the sidecars with the right VirtualService configuration.

This can be verified in the metrics that [Istio emits](https://istio.io/latest/docs/reference/config/metrics/). For example, the `istio_requests_total` metric would be emitted like this (some labels omitted for brevity):

```
istio_requests_total{
 destination_canonical_revision="latest",
 destination_canonical_service="unknown",
 destination_cluster="unknown",
 destination_principal="unknown",
 destination_service="sklearn-v2-iris.kserve-test.svc.cluster.local",
 destination_service_name="PassthroughCluster",
 destination_service_namespace="unknown",
 destination_workload="unknown",
 destination_workload_namespace="unknown",
 [...]}
```

Specifically, the `destination_service_name="PassthroughCluster"` reveals that the requests are being hanlded as leaving the service mesh, despite the client-workload has a sidecar. Such requests would be blocked if Istio would be configured with `REGISTRY_ONLY`.
* Learn more about the PassthroughCluster in [this Istio blog post](https://istio.io/v1.13/blog/2019/monitoring-external-service-traffic/#what-are-blackhole-and-passthrough-clusters), and about REGISTRY_ONLY [in Istio docs](https://istio.io/latest/docs/tasks/traffic-management/egress/egress-control/#change-to-the-blocking-by-default-policy).

When using the [Kiali project](https://kiali.io/), its graph will reveal that observability is potentially lost. For example, the following image shows that the `curl-inside-mesh` workload has a sidecar and all its requests are going to the PassthroghCluster, despite everything is in the mesh.

image

## Fix

This is adding the missing configurations in the KServe-created VirtualService, so that Istio sidecars are aware of the KServe services/hostnames and workloads with an Istio sidecar will correctly handle the requests as mesh-internal traffic. Also, there is the added benefit that the URL rewrite will be done in the sidecar, rather than delaying/deferring the rewrite to the Gateway and this saves one request forwarding (potentially, slightly better performance).

With the fixed configs, the following is an example of a metric that will be emitted by Istio:

```
istio_requests_total{
 destination_app="unknown",
 destination_canonical_revision="data-science-smcp",
 destination_canonical_service="istio-ingressgateway",
 destination_cluster="Kubernetes",
 destination_principal="unknown",
 destination_service="knative-local-gateway.istio-system.svc.cluster.local",
 destination_service_name="knative-local-gateway",
 destination_service_namespace="istio-system",
 destination_version="unknown",
 destination_workload="istio-ingressgateway",
 destination_workload_namespace="istio-system",
 [...]}
```

Such metric no longer involves the PassthroughCluster, which should be understood as Istio handling the requests as mesh-internal. Furthermore, if mTLS is enabled in Istio, the metrics will also have _principal_ labels populated (requests are authenticated); e.g:

```
istio_requests_total{
 [...]
 destination_principal="spiffe://cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account",
 source_principal="spiffe://cluster.local/ns/kserve-workloads/sa/default"}
```

...and this gives the advantage of being able to use Istio security features, if needed (like AuthorizationPolicies). Also, better observability may also be possible. For example, the Kiali graph would now be like this:

image

**Type of changes**

# Feature/Issue validation/testing

* Relevant unit tests were adapted.
* Manual validation with Istio metrics was performed, as mentioned previously.

**Checklist**:

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -231,6 +231,7 @@ const (
 
 var (
 	LocalGatewayHost = "knative-local-gateway.istio-system.svc." + network.GetClusterDomainName()
+	IstioMeshGateway = "mesh"
 )
 
 // InferenceService Component enums
--- a/pkg/controller/v1beta1/inferenceservice/controller_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller_test.go
@@ -316,6 +316,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 				Spec: istiov1beta1.VirtualService{
 					Gateways: []string{
 						constants.KnativeLocalGateway,
+						constants.IstioMeshGateway,
 						constants.KnativeIngressGateway,
 					},
 					Hosts: []string{
@@ -326,7 +327,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 						{
 							Match: []*istiov1beta1.HTTPMatchRequest{
 								{
-									Gateways: []string{constants.KnativeLocalGateway},
+									Gateways: []string{constants.KnativeLocalGateway, constants.IstioMeshGateway},
 									Authority: &istiov1beta1.StringMatch{
 										MatchType: &istiov1beta1.StringMatch_Regex{
 											Regex: constants.HostRegExp(network.GetServiceHostname(serviceKey.Name, serviceKey.Namespace)),
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -289,7 +289,7 @@ func createHTTPMatchRequest(prefix, targetHost, internalHost string, additionalH
 					Regex: constants.HostRegExp(internalHost),
 				},
 			},
-			Gateways: []string{config.LocalGateway},
+			Gateways: []string{config.LocalGateway, constants.IstioMeshGateway},
 		},
 	}
 	if !isInternal {
@@ -431,6 +431,7 @@ func createIngress(isvc *v1beta1.InferenceService, useDefault bool, config *v1be
 
 	gateways := []string{
 		config.LocalGateway,
+		constants.IstioMeshGateway,
 	}
 	if !isInternal {
 		hosts = append(hosts, serviceHost)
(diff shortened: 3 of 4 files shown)


pkg/constants/constants.go
```diff
@@ -231,6 +231,7 @@ const (
 
 var (
 	LocalGatewayHost = "knative-local-gateway.istio-system.svc." + network.GetClusterDomainName()
+	IstioMeshGateway = "mesh"
```
spolti, line 234: is it safe to hardcode to mesh?
I mean, it is isolated peer namespace, but the MeshConfig can have m ore than one namespace, I don't think it is the case for KServe, just double checking :)

israel-hdez: It should be safe.

The _mesh_ gateway is a keyword in the VirtualService. It means that the configuration should be applied to the sidecars.

bartoszmajsak: One little nit - it is applied to ALL sidecars in the mesh.

bartoszmajsak:
One thing I wonder is if "mesh" was deliberately skipped in the original architecture or if it was overlooked. I can't answer this question but I wonder if this has any impact on the traffic flow and alternates some rules or is it working as expected.

yuzisun:
> One thing I wonder is if "mesh" was deliberately skipped in the original architecture or if it was overlooked. I can't answer this question but I wonder if this has any impact on the traffic flow and alternates some rules or is it working as expected.

It was probably an oversight, I think this should work fine as it is adding the mesh gateway to resolve the routing in the sidecar instead of going to the passthrough.

Jooho:
@yuzisun looks like this pr is ready to merge. could you please approve it?

spolti (approved):
/lgtm

israel-hdez:
@yuzisun OK, I have one approval.
Let me know if you need something else before merging this one.

PR #3715, opened 2024-05-30
Fixup max_length for HF and model info for vLLM

Datta0:

1. As of now, if you pass any model_id param to create vLLM model, it fails because the args.model is being set to `/mnt/model`, the default value of `args.model_dir`. So for creating `args.model`, let `args.model_id` take precedence over `args.model_dir`.
2. By default while initialising and loading models, we're setting `self.max_length` (that denotes max num of tokens that can be processed) to [`self.model_config.max_length`](https://huggingface.co/docs/transformers/en/main_classes/configuration#transformers.PretrainedConfig.max_length) which is a parameter read from generation_config.json, that defaults to 20. But a lot of models like [mistral](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2/blob/main/generation_config.json), [gemma](https://huggingface.co/google/gemma-2b-it/blob/main/generation_config.json) don't have that parameter (llama has). So let's rely on `self.max_position_embeddings` for the same?

**Type of changes**

**Checklist**:

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/test_model.py
+++ b/python/huggingfaceserver/huggingfaceserver/test_model.py
@@ -257,12 +257,13 @@ async def test_bloom_completion(bloom_model: HuggingfaceGenerativeModel):
         prompt="Hello, my dog is cute",
         stream=False,
         echo=True,
+        max_tokens=100,  # bloom doesn't have any field specifying context length. Our implementation would default to 2048. Testing with something longer than HF's default max_length of 20
     )
     request = CompletionRequest(params=params)
     response = await bloom_model.create_completion(request)
     assert (
         response.choices[0].text
-        == "Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute"
+        == "Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey,' == 'Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute"
     )
 
 
--- a/python/huggingfaceserver/huggingfaceserver/test_model.py
+++ b/python/huggingfaceserver/huggingfaceserver/test_model.py
@@ -257,13 +257,29 @@ async def test_bloom_completion(bloom_model: HuggingfaceGenerativeModel):
         prompt="Hello, my dog is cute",
         stream=False,
         echo=True,
+    )
+    request = CompletionRequest(params=params)
+    response = await bloom_model.create_completion(request)
+    assert (
+        response.choices[0].text
+        == "Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute"
+    )
+
+
+@pytest.mark.asyncio
+async def test_bloom_completion_max_tokens(bloom_model: HuggingfaceGenerativeModel):
+    params = CreateCompletionRequest(
+        model="bloom-560m",
+        prompt="Hello, my dog is cute",
+        stream=False,
+        echo=True,
         max_tokens=100,  # bloom doesn't have any field specifying context length. Our implementation would default to 2048. Testing with something longer than HF's default max_length of 20
     )
     request = CompletionRequest(params=params)
     response = await bloom_model.create_completion(request)
     assert (
         response.choices[0].text
-        == "Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute.\n- Hey,' == 'Hello, my dog is cute.\n- Hey, my dog is cute.\n- Hey, my dog is cute"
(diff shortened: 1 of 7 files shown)


sivanantha321:
> By default while initialising and loading models, we're setting self.max_length (that denotes max num of tokens that can be processed) to [self.model_config.max_length](https://huggingface.co/docs/transformers/en/main_classes/configuration#transformers.PretrainedConfig.max_length) which is a parameter read from generation_config.json, that defaults to 20. But a lot of models like [mistral](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2/blob/main/generation_config.json), [gemma](https://huggingface.co/google/gemma-2b-it/blob/main/generation_config.json) don't have that parameter (llama has). So let's rely on self.max_position_embeddings for the same?

`max_position_embeddings` is not a standard either. That being said, the Huggingface's pipelines relies on [`tokenizer.model_max_length`](https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer.model_max_length) for restricting the input token length, but if this value is not provided then it is set to a large int value (1e30). This may cause some issues with models that does not provide this configuration. I think it is best to always provide the `--max_length` cmd arg to the model server. WDYT ?

Datta0:
@sivanantha321 
> `max_position_embeddings` is not a standard either. Do we have any example for that? 

Setting it to any large value would not serve the purpose. If some request is larger than it is supposed to be, either the backend might break or the generation might be non sensical

sivanantha321:
> @sivanantha321
> 
> > `max_position_embeddings` is not a standard either. Do we have any example for that?
> 
> Setting it to any large value would not serve the purpose. If some request is larger than it is supposed to be, either the backend might break or the generation might be non sensical

Yeah, that's what I am saying. I think it is best to always provide the --max_length cmd arg to the model server instead of relying on defaults

Datta0:
Users should not be mandated to enter such values which are deeply technical. One who wants to run llama should not need to know its context window size or anything right?
Ease of use will get effected if we force that. 

And afaik, max_position_embeddings is a required parameter. Please let me know if there are any cases where that isn't present.

sivanantha321:
> Users should not be mandated to enter such values which are deeply technical. One who wants to run llama should not need to know its context window size or anything right? Ease of use will get effected if we force that.
> 
> And afaik, max_position_embeddings is a required parameter. Please let me know if there are any cases where that isn't present.

If max_position_embeddings is a required parameter, then I guess it's fine.

sivanantha321:
It looks like bloom-560 does not have the max_position_embedding. E       AttributeError: 'BloomConfig' object has no attribute 'max_position_embeddings'. Should this be only applicable to generative models ?

Datta0:
Then ig we'll try to fetch that value and if not present, rely on a default value which is good enough but not too large like 512 or 2048?

Datta0:
[vLLM](https://github.com/vllm-project/vllm/blob/a22dea54d3e80bf069cfeed8002a193ef8b18e1b/vllm/config.py#L1160) for example has a dynamic way of checking for the same and a default value in case there's none
We can have something similar

johnugeorge:
Currently, gemma is broken since it doesn't have this parameter.  vllm method seems to be a good balance

python/huggingfaceserver/huggingfaceserver/test_model.py
```diff
@@ -257,12 +257,13 @@ async def test_bloom_completion(bloom_model: HuggingfaceGenerativeModel):
         prompt="Hello, my dog is cute",
         stream=False,
         echo=True,
+        max_tokens=100,  # bloom doesn't have any field specifying context length. Our implementation would default to 2048. Testing with something longer than HF's default max_length of 20
```
yuzisun, line 276: Can we add a separate test for max_tokens 100?

yuzisun: the chat completion test seems to be failing

python/huggingfaceserver/huggingfaceserver/utils.py
```diff
@@ -0,0 +1,123 @@
...
+        derived_max_model_len = min(derived_max_model_len, sliding_window_len)
+
+    # If none of the keys were found in the config, use a default and
+    # log a warning.
+    if derived_max_model_len == float("inf"):
+        if max_model_len is not None:
+            # If max_model_len is specified, we use it.
+            return max_model_len
+
+        default_max_len = 2048
```
yuzisun, line 71: Is this expected to default to a very large Max length?

python/huggingfaceserver/huggingfaceserver/test_model.py
yuzisun: just did a test, seems like for `CreateChatCompletionRequest` it defaults `max_tokens` to None and then kserve sets to `params.max_tokens = self.max_length - num_input_tokens`

```python
>>> b = CreateChatCompletionRequest(model="test", messages=[{"role": "system", "content": "testtestst"}])
>>> b
CreateChatCompletionRequest(messages=[ChatCompletionRequestSystemMessage(content='testtestst', role='system', name=None)], model='test', frequency_penalty=0, logit_bias=None, logprobs=False, top_logprobs=None, max_tokens=None, n=1, presence_penalty=0, response_format=None, seed=None, stop=None, stream=False, temperature=1, top_p=1, tools=None, tool_choice=None, user=None, function_call=None, functions=None)
```

yuzisun: Interestingly chat completion defaults max_token to None while completion request defaults to 16

https://github.com/kserve/kserve/blob/16c90170721e6c2ed121204fa848d41e9f4f4e96/python/kserve/kserve/protocol/rest/openai/types/openapi.py#L2643

python/huggingfaceserver/huggingfaceserver/utils.py
Datta0: My idea was if none of the keys are mentioned in the configs, we should default to something that is manageable by most of the contemporary models. 2048 is something even LlaMA-1 would support. Anything higher and we'd run the risk of letting users generate more than the model's context window. Anything lower and we're under-powering the user. 

We can go for a different default value but 2048 looked good enough to me and as this implementation is based off of vLLM, I'm also using the same.

python/huggingfaceserver/huggingfaceserver/test_model.py
Datta0: I think openai's expectation is that chat complete models are supposed to end when they generate <eos_token> (or equivalent) because they are trained to finish their responses.
So things should be better if we use a chat/instruct model for this test?
Something like [DialoGPT](https://huggingface.co/microsoft/DialoGPT-small) which is small yet generates dialogues that are limited in length.
The responses can be non-sensical but there'll be a response

Datta0: [tinyllama(&chat)](huggingface.co/tinyllama/tinyllama-1.1b-chat-v1.0) is a good model but its very big in comparison to the above one.
And 2x as big as bloom.

yuzisun:
Thanks @Datta0 !!

/lgtm
/approve

PR #3716, opened 2024-05-30
Extract openai predict logic into smaller methods

grandbora:
## What does this PR do?
This PR is purely a refactor. It does not change the logic. It puts the the prediction code into smaller methods.
## Why do we want it?
1. Having fine grained methods yields a better user experience since the child classes have greater flexibility on what logic they want to override.
1. The internal implementation is now closer to the `pre-process / predict / post-process` pattern, the users are already accustomed to from v1 endpoints.
## Example Usage
In the V1 endpoints, we override the [predict logic](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/model.py#L398) to wrap it with custom latency measurements and exception handling. It looks like below:

```python
async def predict(self, payload: dict, headers: dict[str, str] = None) -> dict:
  with self.latency_tracker.report_predict(tag1=my_tag_1, tag2=my_tag2):
    try:
      return await super().predict(payload, headers)
    except Exception as exc:
      # log, inc counters etc.
      raise exc
```

With the current implementation of OpenAIProxyModel, we have to override the whole `create_completion` method to apply this pattern. It is better to be able to override only the prediction logic and measure that.

**Feature/Issue validation/testing**: `make test`

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
+++ b/python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
@@ -254,15 +254,19 @@ class OpenAIProxyModel(OpenAIModel):
             )
             return it
         else:
-            response = await self._http_client.send(req)
-            response.raise_for_status()
-            if self.skip_upstream_validation:
-                obj = response.json()
-                completion = Completion.model_construct(**obj)
-            else:
-                completion = Completion.model_validate_json(response.content)
+            completion = await self.predict_completion(req)
             self.postprocess_completion(completion, request)
             return completion
+
+    async def predict_completion(self, req: httpx.Request) -> Completion:
+        response = await self._http_client.send(req)
+        response.raise_for_status()
+        if self.skip_upstream_validation:
+            obj = response.json()
+            completion = Completion.model_construct(**obj)
+        else:
+            completion = Completion.model_validate_json(response.content)
+        return completion
 
     @error_handler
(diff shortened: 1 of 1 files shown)


grandbora:
cc @cmaddalozzo @yuzisun I'd appreciate a review.

python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
```diff
@@ -254,16 +254,20 @@ async def create_completion(
...
-            if self.skip_upstream_validation:
-                obj = response.json()
-                completion = Completion.model_construct(**obj)
-            else:
-                completion = Completion.model_validate_json(response.content)
+            completion = await self.predict_completion(req)
             self.postprocess_completion(completion, request)
             return completion
 
+    async def predict_completion(self, req: httpx.Request) -> Completion:
```
yuzisun: not sure if predict is the right verb, should we name it `generate_completion` ?

yuzisun: Also we should't expose http client to the function signature

grandbora:
@yuzisun Addressed the comments.

grandbora:
Hey @yuzisun this has been waiting for a while. I'd appreciate a review.

cmaddalozzo (approved):

PR #3721, opened 2024-06-04
Configurable label annotation

### LINKED ISSUE #3710: Make label and annotation propagation configurable
/kind feature

When KServe creates the underlying Knative revisions for an Inference Service's predictor and transformer we propagate the annotations and labels that are attached to the Inference Service. We have a [hardcoded disallow list](https://github.com/kserve/kserve/blob/690e269ac1aa88d6e15872164c966305cd518169/pkg/constants/constants.go#L332) of annotations we do not want to propagate. We should make this list configurable on a cluster level as well as allow configuring a disallow list for labels. We can accomplish this by adding a section to the `inferenceservice-config` ConfigMap.

```yaml
     deploy: |-
       {
         "defaultDeploymentMode": "Serverless",
         "serviceAnnotationDisallowedList": [
             "autoscaling.knative.dev/min-scale",
             "autoscaling.knative.dev/max-scale",
             "internal.serving.kserve.io/storage-initializer-sourceuri",
             "kubectl.kubernetes.io/last-applied-configuration"
         ],
         "serviceLabelDisallowedList": []
       }
```

Before propagating an annotation or label we must ensure it's not present on the respective disallow list.

andyi2it:

Fixes #3710 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


config/configmap/inferenceservice.yaml
```diff
@@ -546,7 +546,16 @@ data:
 
   deploy: |-
     {
-      "defaultDeploymentMode": "Serverless"
+      "defaultDeploymentMode": "Serverless",
+      "serviceAnnotationDisallowedList": [
```
yuzisun: Let's separate out to a different config item, maybe we can call `labelAnnotationPropagation`

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -148,9 +148,6 @@ func createKnativeService(componentMeta metav1.ObjectMeta,
 		}
 		trafficTargets = append(trafficTargets, latestTarget)
 	}
-	labels := utils.Filter(componentMeta.Labels, func(key string) bool {
```
yuzisun, line 151: this is no longer needed?

config/configmap/inferenceservice.yaml
```diff
@@ -548,6 +561,19 @@ data:
...
+      "autoscaling.knative.dev/min-scale",
+      "autoscaling.knative.dev/max-scale",
+      "internal.serving.kserve.io/storage-initializer-sourceuri",
+      "kubectl.kubernetes.io/last-applied-configuration"
+    ]
+
+  labelsPropagationDisallowList: |-
+    [
+      "networking.knative.dev/visibility"
+    ]
```
terrytangyuan: Would it be better to support regex instead?

config/configmap/inferenceservice.yaml
```diff
@@ -425,7 +425,20 @@ data:
          # ModelMesh https://kserve.github.io/website/master/admin/modelmesh/
          "defaultDeploymentMode": "Serverless"
        }
-     
+
+      annotationsPropagationDisallowList: |-
+        [
+          "autoscaling.knative.dev/min-scale",
+          "autoscaling.knative.dev/max-scale",
+          "internal.serving.kserve.io/storage-initializer-sourceuri",
```
israel-hdez: I've seen several KServe internal annotations and labels. I so so remember some of those internal annotations/labels would control injections on the pods.

So, I wonder if it is worth to disallow the KServe internal ones even if they are not on the list...

config/configmap/inferenceservice.yaml
andyi2it: Sure, agree... Would add support

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
andyi2it: We have moved it to a different place.

config/configmap/inferenceservice.yaml
```diff
@@ -548,6 +569,19 @@ data:
     {
       "defaultDeploymentMode": "Serverless"
     }
+  
+  annotationsPropagationDisallowList: |-
```
andyi2it: Can be moved under inferenceService ?

spolti:
Hi @andyi2it, looks like we did the same thing, I sent this PR last week https://github.com/kserve/kserve/pull/4030
we could merge both, on my version I decided to put it inside the inferenceServiceConfig because it is already injected in every place we need to use it.

Please let me know how you want to handle it.

andyi2it:
> Hi @andyi2it, looks like we did the same thing, I sent this PR last week #4030 we could merge both, on my version I decided to put it inside the inferenceServiceConfig because it is already injected in every place we need to use it.
> 
> Please let me know how you want to handle it.

Hi @spolti 
We can use yours as long as it meets the same purpose. I will close this one. Can you verify if your PR addresses all tests?
I will leave my comment there.

spolti:
@andyi2it sure, I'll handle it this week.

andyi2it:
Is resolved by https://github.com/kserve/kserve/pull/4030

PR #3737, opened 2024-06-12
Allow KServe to have its own local gateways for Serverless mode

israel-hdez:

These changes introduce the possibility to configure KServe with its own Istio local gateway, to partially decouple KServe from the Knative local gateway.

Typically, it is OK to re-use the already configured Knative local gateway for KServe uses (as long as configs do not conflict). However, there are cases where having a dedicated local gateway for KServe is beneficial. Just to give some examples:
* To have the ability to use strict mTLS in Istio
* To reduce some pressure on the Knative local gateway by having a dedicated gateway deployment (it still would hit Knative gateway, but only once, rather than twice)
* To be able to configure TLS on cluster-local hostnames (Knative support is still experimental)

To have a dedicated Gateway in KServe, similar configurations to Knative are need to be done. At the very least, and if not having a dedicated gateway deployment, a v1/Service and an Istio Gateway resource need to be created for KServe. Such resources would need to be configured in _localGateway_ and _localGatewayService_. KServe still needs to rely on Knative routing for the KSVCs it creates. Thus, after handling an incoming request and resolving its target, it needs to be forwarded to be handled by Knative. This is the reason for introducing a new `knativeLocalGatewayService` in the ConfigMap.

The removed `ingressService` seems to be unused. Apparently, it became unused when the v1alpha2 API of the InferenceServices was deprecated and removed.

**Type of changes**

**TODO**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Allow KServe to have its own local gateways for Serverless mode
```

### DIFF
--- a/charts/kserve-resources/templates/configmap.yaml
+++ b/charts/kserve-resources/templates/configmap.yaml
@@ -181,7 +181,6 @@ data:
      ingress: |-
        {
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
            "localGateway" : "knative-serving/knative-local-gateway",
            "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
            "ingressDomain"  : "example.com",
@@ -199,9 +198,15 @@ data:
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
 
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway service.
+           # The default KServe configurations are re-using the Istio local gateways for Knative. In this case, this
+           # knativeLocalGatewayService field can be left unset. When unset, the value of "localGatewayService" will be used.
+           # However, sometimes it may be better to have local gateways specifically for KServe (e.g. when enabling strict mTLS in Istio).
+           # Under such setups where KServe is needed to have its own local gateways, the values of the "localGateway" and
+           # "localGatewayService" should point to the KServe local gateways. Then, this knativeLocalGatewayService field
+           # should point to the Knative's local gateway service.
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
+           "knativeLocalGatewayService": "",
 
            # localGateway specifies the gateway which handles the network traffic within the cluster.
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
@@ -495,7 +500,7 @@ data:
   ingress: |-
     {
         "ingressGateway" : "{{ .Values.kserve.controller.gateway.ingressGateway.gateway }}",
-        "ingressService" : "{{ .Values.kserve.controller.gateway.ingressGateway.gatewayService }}",
+        "knativeLocalGatewayService" : "{{ .Values.kserve.controller.gateway.localGateway.knativeGatewayService }}",
         "localGateway" : "{{ .Values.kserve.controller.gateway.localGateway.gateway }}",
         "localGatewayService" : "{{ .Values.kserve.controller.gateway.localGateway.gatewayService }}",
         "ingressClassName" : "{{ .Values.kserve.controller.gateway.ingressGateway.className }}",
--- a/config/configmap/inferenceservice.yaml
+++ b/config/configmap/inferenceservice.yaml
@@ -192,7 +192,6 @@ data:
      ingress: |-
        {
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
            "localGateway" : "knative-serving/knative-local-gateway",
            "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
            "ingressDomain"  : "example.com",
@@ -210,9 +209,15 @@ data:
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
      
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway service.
+           # The default KServe configurations are re-using the Istio local gateways for Knative. In this case, this
+           # knativeLocalGatewayService field can be left unset. When unset, the value of "localGatewayService" will be used.
+           # However, sometimes it may be better to have local gateways specifically for KServe (e.g. when enabling strict mTLS in Istio).
+           # Under such setups where KServe is needed to have its own local gateways, the values of the "localGateway" and
+           # "localGatewayService" should point to the KServe local gateways. Then, this knativeLocalGatewayService field
+           # should point to the Knative's local gateway service.
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
+           "knativeLocalGatewayService": "",
      
            # localGateway specifies the gateway which handles the network traffic within the cluster.
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
@@ -494,7 +499,6 @@ data:
   ingress: |-
     {
         "ingressGateway" : "knative-serving/knative-ingress-gateway",
-        "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
         "localGateway" : "knative-serving/knative-local-gateway",
         "localGatewayService" : "knative-local-gateway.istio-system.svc.cluster.local",
         "ingressDomain"  : "example.com",
--- a/charts/kserve-resources/README.md
+++ b/charts/kserve-resources/README.md
@@ -27,9 +27,9 @@ $ helm install kserve oci://ghcr.io/kserve/charts/kserve --version v0.13.0
 | kserve.controller.gateway.domainTemplate | string | `"{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}"` |  |
 | kserve.controller.gateway.ingressGateway.className | string | `"istio"` |  |
 | kserve.controller.gateway.ingressGateway.gateway | string | `"knative-serving/knative-ingress-gateway"` |  |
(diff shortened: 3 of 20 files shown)


charts/kserve-resources/templates/configmap.yaml
```diff
@@ -199,9 +198,15 @@ data:
...
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway serivice.
+           # The default KServe configurations are re-using the Istio local gateways for Knative. In this case, this
+           # knativeLocalGatewayService field can be left unset. When unset, the value of "localGatewayService" will be used.
+           # However, sometimes it may be better to have local gateways specifically for KServe (e.g. when enabling strict mTLS in Istio).
+           # Under such setups where KServe is needed to have its own local gateways, the values of the "localGateway" and
+           # "localGatewayService" should point to the KServe local gateways. Then, this knativeLocalGatewayService field
+           # should point to the Knative's local gateway service.
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
+           "knativeLocalGatewayService": "",
```
lizzzcai, line 209: why not point to the `knative-local-gateway.istio-system.svc.cluster.local` as well, like localGatewayService?

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -199,9 +198,15 @@ data:
...
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
 
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway serivice.
+           # The default KServe configurations are re-using the Istio local gateways for Knative. In this case, this
+           # knativeLocalGatewayService field can be left unset. When unset, the value of "localGatewayService" will be used.
+           # However, sometimes it may be better to have local gateways specifically for KServe (e.g. when enabling strict mTLS in Istio).
+           # Under such setups where KServe is needed to have its own local gateways, the values of the "localGateway" and
+           # "localGatewayService" should point to the KServe local gateways. Then, this knativeLocalGatewayService field
+           # should point to the Knative's local gateway service.
```
lizzzcai, line 207: why need to point to knative's local gateway here even if there is a dedicated gateway for KServe?
I saw KNative support [non-default local gateway](https://knative.dev/docs/install/installing-istio/#updating-the-config-istio-configmap-to-use-a-non-default-local-gateway) and seems like it can support multiple entries? if we updated the dedicated KServe localgateway here as well, do we still need to point to knative local gateway?

lizzzcai:
just something come up in my mind maybe can be a future enhancement if it is useful. 
Do you think having a dedicated gateway CRD to manage the gateway creation and configuration here will be useful? so that users don't need to look into how to configure Istio and KNative, and maybe will be helpful for the raw deployment mode (to support different gateway implementations?)

charts/kserve-resources/templates/configmap.yaml
israel-hdez: It is related to the top-level virtual service that KServe creates. There is some explanation here: https://kserve.github.io/website/latest/admin/serverless/servicemesh/#disable-top-level-virtual-service

charts/kserve-resources/templates/configmap.yaml
israel-hdez: I can do it in this part, as this is just an example, to be explicit about what would be the default value.

Although, I'm not sure if it is OK to be explicit in the actual configs. I'm not sure if that would lead to configuration mistakes for users that would be migrating and using a different name for the gateways. I'm trying to be backwards-compat friendly.

config/configmap/inferenceservice.yaml
```diff
@@ -210,9 +209,15 @@ data:
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
      
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway serivice.
```
spolti:
```suggestion
           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway service.
```

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -199,9 +198,15 @@ data:
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
 
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway serivice.
```
spolti:
```suggestion
           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway service.
```

spolti (approved):

israel-hdez:
I have a pull request in the website that may give better idea about why these changes are relevant: https://github.com/kserve/website/pull/374.

israel-hdez:
> just something come up in my mind maybe can be a future enhancement if it is useful. Do you think having a dedicated gateway CRD to manage the gateway creation and configuration here will be useful? so that users don't need to look into how to configure Istio and KNative, and maybe will be helpful for the raw deployment mode (to support different gateway implementations?)

Wel... I like abstractions to make things easier for users. But, unfortunately, there are many ways to expose services and it is probably better and less problematic to let the users do it themselves using their favorite method.

That said, I expect [Gateway API](https://gateway-api.sigs.k8s.io/) will be that needed abstraction.

spolti (approved):
/lgtm

charts/kserve-resources/templates/configmap.yaml
israel-hdez: oops! Fixed.

config/configmap/inferenceservice.yaml
israel-hdez: Fixed :-)

israel-hdez:
@yuzliu I have an approval. Would you like to review?

israel-hdez:
Sorry... in my previous comment I wanted to tag @yuzisun

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -199,9 +198,15 @@ data:
            # NOTE: This configuration only applicable for serverless deployment with Istio configured as network layer.
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
 
-           # ingressService specifies the hostname of the ingress service.
+           # knativeLocalGatewayService specifies the hostname of the Knative's local gateway service.
+           # The default KServe configurations are re-using the Istio local gateways for Knative. In this case, this
+           # knativeLocalGatewayService field can be left unset. When unset, the value of "localGatewayService" will be used.
```
yuzisun, line 203: This is not backwards compatible if user has customized the existing `localGatewayService`

yuzisun: oh I see you have
```go
if len(ingressConfig.KnativeLocalGatewayService) == 0 {
     ingressConfig.KnativeLocalGatewayService = ingressConfig.LocalGatewayServiceName
}
```

charts/kserve-resources/templates/configmap.yaml
yuzisun: @israel-hdez Can you explain a bit more how this works when top level virtual service is enabled by using an additional gateway? I am a bit concerned that this adds additional configuration complexity to kserve admin.

yuzisun: If I understand correctly, the additional local gateway is needed for handling the non service mesh traffic when mutual tls is on. But I am not sure if this is a real use case as if the cluster is turned on for service mesh then mutual tls should work in cluster. You can use ingress gateway to handle the non service mesh traffic so we eliminate the need of the additional local gateway.

yuzisun:
@israel-hdez @terrytangyuan I am not convinced we need this additional configuration but I could miss something, maybe we can schedule a call to discuss.

israel-hdez:
@yuzisun We'll reach you on Slack either on Monday 8th or Tuesday 9th to schedule the call.

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -181,7 +181,6 @@ data:
      ingress: |-
        {
            "ingressGateway" : "knative-serving/knative-ingress-gateway",
-           "ingressService" : "istio-ingressgateway.istio-system.svc.cluster.local",
```
yuzisun, line 184: just a note that this config is no longer used by kserve hence we can remove it

charts/kserve-resources/templates/configmap.yaml
yuzisun: I am just thinking about a better name for this as by default other gateway configurations are reusing knative gateways, so this could lead to some confusions.

charts/kserve-resources/templates/configmap.yaml
israel-hdez: Yes, the code is what makes it backwards compatible.

charts/kserve-resources/templates/configmap.yaml
israel-hdez: The way I think it is that, by default, _kserve gateway_ is also the Knative's one :slightly_smiling_face: .

I was unable to think a better name, given in the de-coupled configuration, this name seems to represent its goal. But if there is a better naming, I'm happy to change it.

charts/kserve-resources/templates/configmap.yaml
israel-hdez: > But I am not sure if this is a real use case as if the cluster is turned on for service mesh then mutual tls should work in cluster.

@yuzisun  Hopefully, it is more clear after our offline call.

For the record, and in short, when locking the mesh with mTLS using both _PeerAuthentication_ and also _DestinationRule_ traffic becomes blocked on the ingress-gateway when using the top-level virtual service. This is because the top-level virtual service is configuring a loop on the ingress gateway. The listening port of the ingress gateway is using plain-text. In the default configurations, this loopback connection of the top-level virtualservice is possible, because the connection happens as plain-text. However, when locking the mesh with a DestinationRule with mTLS, this loopback connection no longer succeeds because it is tried with mTLS while the gateway is still listening with plain-text.

By allowing KServe to have its own Gateway, it becomes possible to open a new port in the gateway that would listen with mTLS rather than plain-text. Thus, it becomes feasible to configure the top-level virtualservice to connect to the mTLS port.

israel-hdez:
@yuzisun After some long vacations, I'm retaking this. I wonder if there is something else that needs to be clarified for this PR?

yuzisun:
> @yuzisun After some long vacations, I'm retaking this. I wonder if there is something else that needs to be clarified for this PR?

other than the configuration naming it looks good

yuzisun (approved):

PR #3749, opened 2024-06-17
[WIP] Model cache Example

greenmoon55:
1. Introduce model cache cr with storage uri and name
2. Add pvc name in storage initializer for all models. (which might be temporary if we switch to one pvc per model)
3. When the inference service has a annotation, AND the storage uri matches the one in model cache cr, the storage initializer injector would modify the storage uri to be a pvc url and the storage initializer injection is skipped. The pvc path is "pvc://`pvc name from configmap`/models/`model cache name from cr`".

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the storage initializer init container, and the protocols it supports.
+// +k8s:openapi-gen=true
+type ClusterCachedModelSpec struct {
+	// Container spec for the storage initializer init container
+
+	StorageUri string `json:"storageUri" validate:"required"`
```
yuzisun, line 28: Maybe we can consider naming this model uri now

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,57 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope="Cluster"
+type ClusterCachedModel struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ClusterCachedModelSpec `json:"spec,omitempty"`
+
+	// +optional
+	Disabled *bool `json:"disabled,omitempty"`
```
yuzisun, line 43: Also need the model status field for storing the cached pvc url and potential other fields.

pkg/constants/constants.go
```diff
@@ -94,6 +94,7 @@ var (
 	DefaultPrometheusPath                       = "/metrics"
 	QueueProxyAggregatePrometheusMetricsPort    = 9088
 	DefaultPodPrometheusPort                    = "9091"
+	EnableModelCache                            = KServeAPIGroupName + "/enable-model-cache"
```
yuzisun, line 97: Can we name it model-cache instead so we have the flexibility of turning this into a non-Boolean field

yuzisun: For example we might want to explicitly bind to the model cache resource

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -66,6 +66,7 @@ type StorageInitializerConfig struct {
 	EnableDirectPvcVolumeMount bool   `json:"enableDirectPvcVolumeMount"`
 	EnableOciImageSource       bool   `json:"enableModelcar"`
 	UidModelcar                *int64 `json:"uidModelcar"`
+	ModelCachePvcName          string `json:"modelCachePvcName"`
```
yuzisun, line 69: If we go with the pvc per model cache CR approach then we do not need this configuration.

PR #3753, opened 2024-06-19
inferenceservice controller: fix error check in Serverless mode

dtrifiro:
It seems that `err` was being used instead of `checkKsvcErr`


yuzisun:
Good catch!

/lgtm
/approve

terrytangyuan (commented):
Thanks for catching this!

Issue #3644, opened 2024-04-28
Add vLLM backend e2e test


yuzisun:
/kind feature

**Describe the solution you'd like**
vLLM 0.4.1 now support cpu backend, so we can add the e2e tests without depending on GPUs.
https://github.com/vllm-project/vllm/pull/3993

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

gavrissh:
@yuzisun @sivanantha321 This can be closed now?

sivanantha321:
> [@yuzisun](https://github.com/yuzisun) [@sivanantha321](https://github.com/sivanantha321) This can be closed now?

Yes

PR #3771, opened 2024-07-01
Add tests for vLLM

sivanantha321:

- Related #3644 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py
```diff
@@ -343,12 +357,15 @@ def apply_chat_template(
         self,
         messages: Iterable[ChatCompletionRequestMessage,],
     ):
-        return self.tokenizer.apply_chat_template(conversation=messages, tokenize=False)
+        return self.tokenizer.apply_chat_template(
+            conversation=messages, tokenize=False, add_generation_prompt=True
```
gavrissh: Ideally this should be a request specific parameter with True as default. For that we need to modify the CreateChatCompletionRequest. [vLLM does that here](https://github.com/vllm-project/vllm/blob/7cd2ebb0251fd1fd0eec5c93dac674603a22eddd/vllm/entrypoints/openai/protocol.py#L177)

python/huggingfaceserver/tests/test_vlllm_model.py
```diff
@@ -0,0 +1,2898 @@
...
+        request = CompletionRequest(request_id=request_id, params=params, context={})
+        response_iterator = await opt_model.create_completion(request)
+        completion = ""
+        log_probs = Logprobs(
+            text_offset=[], token_logprobs=[], tokens=[], top_logprobs=[]
+        )
+        async for resp in response_iterator:
+            assert resp.id == request_id
+            completion += resp.choices[0].text
+            assert len(resp.choices) == 1
```
Datta0, line 1232: Should we assert on length before doing `completion += resp.choices[0].text`?

python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py
sivanantha321: I agree. We have a separate PR. Let's take this discussion to #3775

python/huggingfaceserver/tests/test_vlllm_model.py
```diff
@@ -0,0 +1,2898 @@
...
+                ",": -1.6155686378479004,
+            },
+            {
+                " fluffy": -5.6248779296875,
+                " a": -1.4977400302886963,
+                " tiny": -3.006150484085083,
+            },
+            {" white": -2.2152767181396484, " fur": -1.9012728929519653},
+        ]
+        assert response.id == expected.id
```
Datta0: Should we have these comparisons as a function and do 
`assert compare_response_to_expected(response,expected)`
```
def compare_response_to_expected(actual,expected, fields_to_compare=[default_fields]) -> bool:
   for field in fields_to_compare:
      if actual.field!=expected.field: 
       return False
  return True
```
I see that we're using this at multiple places

python/huggingfaceserver/tests/test_vlllm_model.py
sivanantha321: done

python/huggingfaceserver/tests/test_vlllm_model.py
sivanantha321: done it for non stream cases. for stream each case has different scenarios, making it as a common function will be little bit complex.

yuzisun:
Thanks for adding the tests!!

/lgtm
/approve

yuzisun (approved):

PR #3774, opened 2024-07-03
Torchserve security update 

udaij12:
Security issue in Torchserve require new arguments to be added when torchserve starts in order to maintain the same functionality as before. Users should look to https://github.com/pytorch/serve for full explanation of the security update.

### DIFF
--- a/docs/samples/v1beta1/torchserve/v1/gpu.yaml
+++ b/docs/samples/v1beta1/torchserve/v1/gpu.yaml
@@ -10,4 +10,7 @@ spec:
         limits:
           memory: 4Gi
           nvidia.com/gpu: "1"
+      args:
+        - --disable-token-auth
+        - --enable-model-api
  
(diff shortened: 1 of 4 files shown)


terrytangyuan (approved):
Thanks!

/lgtm
/approve

yuzisun:
I think we need to bump up the tochserve version here

yuzisun:
Cc @andyi2it to review

andyi2it:
These args are compatible with torchserve 0.9, however they could be redundant in version 0.9 as that is the default in 0.9. 
However, 0.11.x changes the behavior and hence may be required.
https://github.com/pytorch/serve/releases/tag/v0.11.1

/lgtm

yuzisun:
> These args are compatible with torchserve 0.9, however they could be redundant in version 0.9 as that is the default in 0.9. However, 0.11.x changes the behavior and hence may be required. https://github.com/pytorch/serve/releases/tag/v0.11.1
> 
> /lgtm

@andyi2it seems like this only updates the doc, but we should really update the torch serve runtime yaml with these arguments right?

yuzisun:
> > These args are compatible with torchserve 0.9, however they could be redundant in version 0.9 as that is the default in 0.9. However, 0.11.x changes the behavior and hence may be required. https://github.com/pytorch/serve/releases/tag/v0.11.1
> > /lgtm
> 
> @andyi2it seems like this only updates the doc, but we should really update the torch serve runtime yaml with these arguments right?

Discussed with @andyi2it we can add in torch serve 0.11

PR #3779, opened 2024-07-08
Make Virtual Service case-insensitive

### LINKED ISSUE #3726: VirtualService regex match should be case insensitive
/kind bug

I have an InferenceService `my-classifier`.

When i try to `curl` from another pod inside the same namespace, it works fine when the domain name is all lowercase (expectedly), but it doesn't when there are some uppercase letters (404 not found). 

But if i try to `curl` the private service created by knative (`my-classifier-predictor-00001-private`), it works for both cases (i'm guessing bcos it doesn't go through Istio).
```
% curl http://my-classifier.test.svc.cluster.local | jq
{
  "status": "alive"
}

% curl http://MY-Classifier.test.svc.cluster.local -v  
> ...
> Host: MY-classifier.test.svc.cluster.local
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< server: envoy
< content-length: 0

% curl http://MY-Classifier-predictor-00001-private.test.svc.cluster.local | jq
{
  "status": "alive"
}
```

I expect `curl http://MY-Classifier.test.svc.cluster.local` to work, since DNS names aren't case sensitive. 

My use case is that the application code i'm writing which will be talking to kserve models may have uppercase letters in the domain name. I can always make it all lowercase in the application code too, but I feel like we should solve this on the kserve/knative/istio side.

```yaml
kind: InferenceService
metadata:
  name: my-classifier
  namespace: test
spec:
  ...
status:
  ...
  url: http://my-classifier.test.svc.cluster.local
```

This is the `VirtualService` created by KServe for this `InferenceService`:
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: my-classifier
  namespace: test
  ownerReferences:
  - apiVersion: serving.kserve.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: InferenceService
    name: my-classifier
  ...
spec:
  gateways:
  - knative-serving/knative-local-gateway
  hosts:
  - my-classifier.test.svc.cluster.local
  http:
  - headers:
      request:
        set:
          Host: my-classifier-predictor.test.svc.cluster.local
    match:
    - authority:
        regex: ^my-classifier\.test(\.svc(\.cluster\.local)?)?(?::\d{1,5})?$
      gateways:
      - knative-serving/knative-local-gateway
    route:
    - destination:
        host: knative-local-gateway.istio-system.svc.cluster.local
        port:
          number: 80
      weight: 100
```

If i add `(?i)` to the `regex` match (so `^(?i)my-classifier\.test(\.svc(\.cluster\.local)?)?(?::\d{1,5})?$`), the failing `curl http://MY-Classifier.test.svc.cluster.local` will work because the "i" modifier specifies a case-insenitive match, so maybe we should do that?

**Environment:**

- Istio Version: v1.20.3
- Knative Version: v1.13.1
- KServe Version: v0.12.0
- Cloud Environment: k8s_istio
- Kubernetes version: (use `kubectl version`): v1.27.13-eks-3af4770

andyi2it:

Fixes #3726

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -579,13 +579,13 @@ const portMatch = `(?::\d{1,5})?`
 // HostRegExp returns an ECMAScript regular expression to match either host or host:<any port>
 // for clusterLocalHost, we will also match the prefixes.
 func HostRegExp(host string) string {
-	localDomainSuffix := ".svc." + network.GetClusterDomainName()
+	localDomainSuffix := "(?i).svc." + network.GetClusterDomainName()
 	if !strings.HasSuffix(host, localDomainSuffix) {
 		return exact(regexp.QuoteMeta(host) + portMatch)
 	}
 	prefix := regexp.QuoteMeta(strings.TrimSuffix(host, localDomainSuffix))
-	clusterSuffix := regexp.QuoteMeta("." + network.GetClusterDomainName())
-	svcSuffix := regexp.QuoteMeta(".svc")
+	clusterSuffix := regexp.QuoteMeta("(?i)." + network.GetClusterDomainName())
+	svcSuffix := regexp.QuoteMeta("(?i).svc")
 	return exact(prefix + optional(svcSuffix+optional(clusterSuffix)) + portMatch)
 }
 


lizzzcai:
I saw Istio support [ignoreUriCase](https://istio.io/latest/docs/reference/config/networking/virtual-service/#:~:text=%22/ratings/v2/%22-,ignoreUriCase,-%3A%20true%0A%20%20%20%20route), seems like KServe not have it as well if I am checking correctly. not sure if this should be included as well?

PR #3808, opened 2024-07-19
Increase timeout to make unit test stable

Jooho:

kserve agent watcher unit test is not stable and it fails a lot locally. I am not sure it is related to my system but it always fails with the following errors:
~~~
2024/07/18 07:03:18 Deleted temp dir /tmp/example3994517843

------------------------------
• [FAILED] [1.001 seconds]
Watcher Use HTTP(S) Downloader Getting new model events [It] should download and load the new models
/home/kserve/pkg/agent/watcher_test.go:679

  [FAILED] Timed out after 1.000s.
  Expected
      <int>: 0
  to equal
      <int>: 1
  In [It] at: /home/kserve/pkg/agent/watcher_test.go:739 @ 07/18/24 07:03:18.357
~~~

After increasing the timeout to 2 minutes, it always succeeds. This is just a timeout increase, so if a unit test passes faster than the timeout, it will take the same amount of time as before.

slack discussion: https://cloud-native.slack.com/archives/C06AH2C3K8B/p1721286710777179

**Type of changes**

no code changes. so it does not need to be tested.
```
make test
```

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/agent/watcher_test.go
+++ b/pkg/agent/watcher_test.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	gstorage "cloud.google.com/go/storage"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


sivanantha321 (approved):

Jooho:
@yuzisun if you have time, please review this. it is not code change. just increase the timeout for the test.

PR #3814, opened 2024-07-23
Add support overriding model mount path in model server container 

### LINKED ISSUE #3606: Support overriding model mount path in model server container
/kind feature

Currently it is not possible to specify at what path the downloaded model should be available in the model server container. The downloaded model is always mounted in the model server container at `/mnt/models`. 

We should allow the user to specify the path inside the model server container the volume should be mounted at. e.g.

```yaml
kind: InferenceService
metadata:
  annotations:
    ds.bloomberg.com/gpuSubtype: bloomberg.com/gpu-l4
    identities.bloomberg.com/bcs: platform-bcosv2
  name: huggingface-t5-hfs
  namespace: s-dsplatform
spec:
  predictor:
    containers:
    - args:
      - --model
      - llama-2-70b
      - --num_gpus=4
      command:
      - nemollm_inference_ms
      env:
      - name: STORAGE_URI
        value: s3://models/llama-7b
      - name: STORAGE_MOUNT_PATH
        value: /model-store
      image: nvcr.io/ohlfw0olaadg/ea-participants/nim_llm:24.02
```

Since serving runtimes may require models to be mounted at a specific location we should also make it configurable on a runtime level. e.g.

```yaml
kind: ClusterServingRuntime
metadata:
  name: nim
spec:
  containers:
  - command:
    - nemollm_inference_ms
    image: nvcr.io/ohlfw0olaadg/ea-participants/nim_llm:24.02
    name: kserve-container
    ports:
    - containerPort: 9999
      protocol: TCP
  modelMountPath: /model-store
  supportedModelFormats:
  - autoSelect: true
    name: nim
```

davidliu40313:

Fixes #3606 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -519,6 +519,117 @@ func TestCustomSpecStorageUriInjection(t *testing.T) {
...
+					},
+				},
+			},
+			expectedStorageMountPathEnvVar: &v1.EnvVar{
+				Name:  constants.CustomSpecStorageMountPathKey,
+				Value: "/custom/models",
+			},
+			expectedMountPath: "/custom/models",
+		},
+		// "CustomSpecStorageMountPathEmpty": {
```
yuzisun: do we still need this test?

davidliu40313: > do we still need this test?

My idea is this test is necessary as it verifies whether the user-defined storage mount path is correctly injected.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -169,8 +175,13 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *v1.Pod) error {
 
 	// Create the modelcar that is used as a sidecar in Pod and add it to the end
 	// of the containers (but only if not already have been added)
+	modelContainer := mi.createModelContainer(image, constants.DefaultModelLocalMountPath)
 	if getContainerWithName(pod, ModelcarContainerName) == nil {
-		modelContainer := mi.createModelContainer(image, constants.DefaultModelLocalMountPath)
```
spolti, line 181: why move outside the if condition?

davidliu40313: > why move outside the if condition?

I moved the declaration of modelContainer outside the if statement to improve code readability and help developers better understand its state changes. However, I realize this may cause unnecessary resource usage when the if condition isn't met. If you think it's better to keep the declaration inside the if, I'm happy to adjust. Thank you for your review

spolti: I think it adds a new unneeded step since it will check every container.

spolti (commented):
changes look good to me.
Let's wait others to take a look.

However, if this gets accepted, do not forget to add a note about it in the docs.

davidliu40313:
@yuzisun Sorry to bother you.Do you have any suggestions for the PR

PR #3819, opened 2024-07-24
Add support for Azure DNS zone endpoints

tjandy98:
Azure [added support for DNS zone endpoints
](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-query-endpoint-srp?tabs=dotnet)


lizzzcai:
is it possible to add some test cases for this as well?

lizzzcai (approved):

yuzisun (approved):

PR #3827, opened 2024-07-30
Rename max_length parameter to max_model_len to be in sync with vLLM

Datta0:
vLLM uses parameter called `max-model-len` to denote the max length allowed for tokenisation and processing. We have similar parameter for HF case but under a diff name called `max_length`. Rename that to use the same parameter for both backends for consistency.
Now we're adding `--max_model_len` which would essentially write to `args.max_model_len`. We'll slowly move away from `--max_length` .

**Checklist**:

```release-note
Add `--max_model_len` to huggingface runtime and deprecate `--max_length`
```

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -133,13 +133,6 @@ if not vllm_available():
         choices=dtype_choices,
         help=f"data type to load the weights in. One of {dtype_choices}. Defaults to float16 for GPU and float32 for CPU systems",
     )
-    # vLLM uses max-model-len as paramter to denote max tokens. Register the same for HuggingFace (if vllm not available)
-    parser.add_argument(
-        "--max-model-len",
-        required=False,
-        type=int,
-        help="max number of tokens the model can process/tokenize. If not mentioned, uses model's max position encodings",
-    )
 
 
 args, _ = parser.parse_known_args()
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -70,6 +70,13 @@ parser.add_argument(
     help="Huggingface tokenizer revision",
 )
 parser.add_argument(
+    "--max_length",
+    dest="max_model_len",
+    type=int,
+    required=False,
+    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-length",
+)
+parser.add_argument(
     "--disable_lower_case",
     action="store_true",
     help="do not use lower case for the tokenizer",
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -74,7 +74,13 @@ parser.add_argument(
     dest="max_model_len",
     type=int,
     required=False,
-    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-len",
+    help="max sequence length for the tokenizer. will be deprecated in favour of --max_model_len",
+)
+parser.add_argument(
+    "--max_model_len",
+    type=int,
+    required=False,
+    help="max number of tokens the model can process/tokenize. If not mentioned, uses model's max position encodings",
 )
 parser.add_argument(
     "--disable_lower_case",
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -74,7 +74,7 @@ parser.add_argument(
     dest="max_model_len",
     type=int,
     required=False,
-    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-length",
+    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-len",
 )
 parser.add_argument(
     "--disable_lower_case",
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -70,12 +70,6 @@ parser.add_argument(
     help="Huggingface tokenizer revision",
 )
 parser.add_argument(
-    "--max_length",
-    type=int,
-    required=False,
-    help="max sequence length for the tokenizer",
-)
-parser.add_argument(
     "--disable_lower_case",
     action="store_true",
     help="do not use lower case for the tokenizer",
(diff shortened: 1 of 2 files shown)


python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -126,6 +120,14 @@ def list_of_strings(arg):
         choices=dtype_choices,
         help=f"data type to load the weights in. One of {dtype_choices}. Defaults to float16 for GPU and float32 for CPU systems",
     )
+    # vLLM uses max-model-len as paramter to denote max tokens. Register the same for HuggingFace (if vllm not available)
+    parser.add_argument(
+        "--max-model-len",
```
sivanantha321: All the other args are using `_`(underscore). We should be consistent with the naming.

Datta0: vLLM uses `-` hyphenated separation. Using the same to ensure we can make use of only single parameter name.

sivanantha321: Wouldn't be a breaking change for current users ?

Datta0: Yeah this would be a breaking change for existing users. Should we allow both for time being and warn that `max_length` would be deprecated in favour of `max-model-len` while giving the latter precedence?
I'm curious if they're internally handling the setting of parameters with diff name for diff backends.

sivanantha321: @yuzisun What's your thought on this ?

yuzisun: if there any strong reason to take the naming convention from vllm ?

Datta0: Someone using kserve need not switch between 2 parameter names if they are switching between HF and vLLM backends. Ensures consistency.

yuzisun: We can potentially support both conventions if there is a way we can easily convert between the two.

Datta0: If we want to support both, one of them should take precedence right?
As for converting, it is a 1-1 swap. For users, it'd be easier to manage one parameter ig.

Datta0: @yuzisun thoughts on this?

spolti: this should be possible:

```python
arg_parser.add_argument('--example-one', '--example_one')
```

Datta0: So I just added `max_length` back which internally sets `args.max_model_len`. 
Let me know if this works

Datta0: @yuzisun @sivanantha321 @spolti please take a look

yuzisun: Can we support both `args.max_model_len` and `args.max-model-len` and deprecate `max_length` ?

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -126,6 +127,14 @@ def list_of_strings(arg):
...
```
yuzisun:
```suggestion
        "--max-model-len",
        "--max_model_len",
```

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -71,9 +71,10 @@ def list_of_strings(arg):
 )
 parser.add_argument(
     "--max_length",
+    dest="max_model_len",
```
yuzisun, line 74:
```suggestion
```

python/huggingfaceserver/huggingfaceserver/__main__.py
Datta0: We're just trying to have the same argument as [vLLM](https://arc.net/l/quote/ufttkhkf). Hence the hyphen `-` instead of underscore `_`

python/huggingfaceserver/huggingfaceserver/__main__.py
Datta0: This is to make sure we're writing to same argument and hence avoid doing something like
 `final_max_len = args.max_length or args.max_model_len`

python/huggingfaceserver/huggingfaceserver/__main__.py
Datta0: Also the problem with adding `--max_model_len` here is, this would not work if vLLM exists. 
If we want to support `--max_model_len`, we'd need to add it separately. This ends up creating 3 CLI flags for one variable.

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -71,9 +71,10 @@ def list_of_strings(arg):
...
     type=int,
     required=False,
-    help="max sequence length for the tokenizer",
+    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-length",
```
yuzisun:
```suggestion
    help="max sequence length for the tokenizer. will be deprecated in favour of --max-model-len",
```

python/huggingfaceserver/huggingfaceserver/__main__.py
yuzisun: I thought `--max-model-len` is parsed to `max_model_len` as well

Datta0: Yeah --max-model-len is parsed to args.max_model_len
Just that if we want to support it, it should be added outside of `if _vllm:`

yuzisun:
```python
parser.add_argument('--foo-bar')
parser.add_argument('--foo_bar')
args = parser.parse_args(['--foo-bar', '24', '--foo_bar', '30'])
Namespace(foo_bar='30')
```

Datta0: if vLLM exists, we import that parser and hence it won't let us add `--max-model-len` *again*
so the final code would end up looking like
```python
parser.add_argument('--max_length', dest='max_model_len')
parser.add_argument('--max_model_len')
if not vllm_available():
    parser.add_argument('--max-model-len')
```

All three for essentially the same variable. Do we want this?

yuzisun: vllm actually implements a flexible arg parser so it can automatically convert underscore to dashes
https://github.com/vllm-project/vllm/blob/80162c44b1d1e59a2c10f65b6adb9b0407439b1f/vllm/utils.py#L1088

yuzisun (approved):
/lgtm
/approve

PR #3834, opened 2024-08-01
Make ray an optional dependency

### LINKED ISSUE #2403: Improve Python SDK dependency management strategy
/kind feature

As discussed in WG meeting ([notes](https://docs.google.com/document/d/1KZUURwr9MnHXqHA08TFbfVbM8EAJSJjmaMhnvstvi-k/edit#heading=h.x518nkl3k9ww)), it'd be better if we could improve our kserve python SDK dependency management strategy for easier customer adoption. Specifically, the current pain points are:
1. strict equal (==, [boto3](https://github.com/kserve/kserve/blob/release-0.9/python/kserve/requirements.txt#L18)) and tilde equal (~=, [numpy](https://github.com/kserve/kserve/blob/release-0.9/python/kserve/requirements.txt#L12)) do not perform well in terms of backward & forward compatibility, and introduce unnecessary restrictions as a "middle-layer" SDK (downstream customer may have specific reasons to use other versions of boto3/numpy, thus making KServe SDK hard to adopt)
2. Similarly, dependency requirements (even under the >= cases) should ideally be as relaxed as possible, so that customers using some old version of dependencies can still enjoy later version of KServe SDK.
3. Not all dependencies declared as a requirement are actually used by customers, e.g., some customers may not need boto3/azure/google libraries depending on their cloud provider choice; some customers may not need ray[serve] if they are not using parallel inference feature. Such too broad hard dependency pattern can introduce additional risk on adoption, e.g., if one unnecessary dependency can't be installed ([example](https://github.com/ray-project/tune-sklearn/issues/169#issuecomment-1163872731))or introduces conflict in customer env.

Suggestions for dependency management:
1. Use >= instead of ~=/== if possible
2. The dependency version requirement should be as relaxed as possible (e.g. the lower the version number the better) so the python sdk could be more adoptable (give the freedom to pick dependency versions to downstream packages)
3. Introduce optional dependencies ([example](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)) to separate out core functionalities from optional functionalities. Some candidates for optional dependency can be ray and cloud provide specific packages.

### LINKED ISSUE #3661: AttributeError: 'Deployment' object has no attribute 'deploy'
I'm trying to run model server using kserve but I'm getting the following error:
```
  File "/project/e2e_servers/pre_asr_kserve.py", line 215, in <module>
    ModelServer(http_port=8000).start({'PRE-ASR': PreASRServer})
  File "/home/xxx/.conda/envs/my_env/lib/python3.9/site-packages/kserve/model_server.py", line 164, in start
    models[key].deploy()
AttributeError: 'Deployment' object has no attribute 'deploy'
```

I installed kserve using pip:
`pip install kserve==0.12.1`

So in my conda env installed:
kserve 0.12.1
ray 2.12.0
starlette 0.36.3

My snippet code:
```
from kserve import Model, ModelServer
from ray import serve

@serve.deployment(name='PRE-ASR', num_replicas=2)
class PreASRServer(Model):
    def __init__(self):
        self.name = 'PRE-ASR'
        super().__init__(self.name)
        self.model = None
        self.ready = False
        self.load()

    def load(self):
        self.model = load_model()
        self.model.eval()
        self.ready = True

    async def predict(self, payload: Dict, headers: Dict[str, str] = None):
        'some code'

if __name__ == "__main__":
    ModelServer().start({'PRE-ASR': PreASRServer})
```

Am I doing something wrong?

Please advise

### LINKED ISSUE #3794: Remove ray dependency for the storage-initializer
/kind bug

Rightnow storage-initializer installs ray as a prerequisite because the kserve module imports but just for the storage-initializer module we don't need ray package. Following patch proves that:

```
[root@mkumatag-x86-fedora python]# git diff kserve/kserve/__init__.py
diff --git a/python/kserve/kserve/__init__.py b/python/kserve/kserve/__init__.py
index 638bacbc..74658fa1 100644
--- a/python/kserve/kserve/__init__.py
+++ b/python/kserve/kserve/__init__.py
@@ -15,10 +15,10 @@
 from __future__ import absolute_import from .model import Model
-from .model_server import ModelServer
+#from .model_server import ModelServer
 from .inference_client import InferenceGRPCClient, InferenceRESTClient, RESTConfig
 from .protocol.infer_type import InferRequest, InferInput, InferResponse, InferOutput
-from .model_repository import ModelRepository
+#from .model_repository import ModelRepository
 from .constants import constants
 from .utils import utils[root@mkumatag-x86-fedora python]# git diff kserve/pyproject.toml
diff --git a/python/kserve/pyproject.toml b/python/kserve/pyproject.toml
index 406b42b5..1d592c09 100644
--- a/python/kserve/pyproject.toml
+++ b/python/kserve/pyproject.toml
@@ -42,7 +42,7 @@ kubernetes = ">=23.3.0"
 python-dateutil = "^2.8.0"
 numpy = "^1.23.5"
 psutil = "^5.9.0"
-ray = { version = "~2.10.0", extras = ["serve"] }
+#ray = { version = "~2.10.0", extras = ["serve"] }
 grpcio = "^1.49.1"
 protobuf = "^3.19.0"
 prometheus-client = "^0.20.0" 
```

We need to find a way to avoid this hard dependency, may be by some conditional import or some other way..

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

sivanantha321:
Make ray an optional dependency #3668
This PR rebases and adds e2e test, some minor fixes to the PR #3668. Thanks for the awesome work @cmaddalozzo.

This PR implements the changes required to make Ray Serve an optional dependency. The code is fully backward compatible **for users that don't use Ray**. Users who are currently using KServe with Ray will need to make some minor changes to their code when upgrading.

Where before you could register a Ray handle directly and KServe would start the ray server:
```python
ModelServer().start({"custom-model": AlexNetModel})
```

Now the user must wrap their deployment handle in a RayModel and start the Ray server:
```python
app = AlexNetModel.bind()
handle = serve.run(app)
model = RayModel(name="custom-model", handle=handle)
model.load()
ModelServer().start([model])
```
While we could continue to have KServe manage the lifecycle of the Ray server I believe it's cleaner and more flexible to require the user to perform this operation.

- Fixes #3794 
- Fixes #3661
- Fixes #2403

**Type of changes**

**Checklist**:

**Release note**:

```release-note
Make Ray Serve an optional dependency
```

**Re-running failed tests**


terrytangyuan (commented):
> The code is fully backward compatible for users that don't use Ray. Users who are currently using KServe with Ray will need to make some minor changes to their code when upgrading.

Can you clarify what "using Ray" means here? I am a bit concerned that this might be a breaking change for any runtimes that rely on Ray.

sivanantha321:
> Can you clarify what "using Ray" means here? I am a bit concerned that this might be a breaking change for any runtimes that rely on Ray.

Ray serve can be used with KServe for parallel inference with multiple models. The default serving runtimes that comes with KServe doesn't support parallel inferencing with ray, So It will not affect them. This will only affect the users who are using ray for parallel inference in their custom model runtimes. You can find the related docs [here](https://kserve.github.io/website/latest/modelserving/v1beta1/custom/custom_model/#parallel-model-inference)

yuzisun (approved):

PR #3839, opened 2024-08-06
Cluster Local Model CR

greenmoon55:
This PR introduces a cluster model cache CR and a node group CR. With these two PRs, you can define a model that would be downloaded to multiple nodes which are defined by a node group CR (node selectors).

Example Model Cache CR
```yaml
kind: ClusterLocalModel
metadata:
  name: iris
spec:
  modelSize: 1Gi
  nodeGroups:
    - gpu1
  sourceModelUri: gs://kfserving-examples/models/sklearn/1.0/model
```

NodeGroup CR
```yaml
kind: LocalModelNodeGroup 
metadata:
  name: gpu1
spec:
  persistentVolumeSpec:
    accessModes:
    - ReadWriteOnce
    volumeMode: Filesystem
    capacity:
      storage: 2Gi
    hostPath:
      path: /models
      type: ""
    persistentVolumeReclaimPolicy: Delete
    storageClassName: standard
    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: node.kubernetes.io/instance-type
            operator: In
            values:
              - gpu1
  persistentVolumeClaimSpec:
    accessModes:
    - ReadWriteOnce
    resources:
      requests:
        storage: 2Gi
    storageClassName: standard
    volumeMode: Filesystem
    volumeName: blah

```

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1alpha1/cached_model_status.go
```diff
@@ -0,0 +1,63 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type CachedModelStatus struct {
+	OverallStatus OverallStatus         `json:"overallStatus,omitempty"`
+	NodeStatus    map[string]NodeStatus `json:"nodeStatus,omitempty"`
```
greenmoon55: example: nodeX: Downloading

pkg/apis/serving/v1alpha1/cached_model_status.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// How many nodes have the model available
+	// +optional
+	ModelCopies       *ModelCopies     `json:"copies,omitempty"`
```
greenmoon55: for example: Suceessful: 10 total: 20

pkg/apis/serving/v1alpha1/cached_model_status.go
```diff
@@ -0,0 +1,63 @@
...
+	InferenceServices []NamespacedName `json:"infereneceServices,omitempty"`
```
greenmoon55: ns1: isvc1, ns2: isvc2 uses this cached model

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the storage initializer init container, and the protocols it supports.
+// +k8s:openapi-gen=true
+type ClusterCachedModelSpec struct {
+	// Container spec for the storage initializer init container
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="StorageUri is immutable"
+	StorageUri string            `json:"storageUri" validate:"required"`
```
greenmoon55: Full match vs match prefix like "storageUriPrefix"

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,80 @@
...
+	ModelSize  resource.Quantity `json:"modelSize" validate:"required"`
+	NodeGroups []string          `json:"nodeGroups" validate:"required"`
```
greenmoon55: one model can be cached on multiple nodegroups (gpu-1, gpu-2, gpu-3)

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,80 @@
...
+	// only local is supported for now
+	StorageType   StorageType   `json:"storageType" validate:"required"`
+	CleanupPolicy CleanupPolicy `json:"cleanupPolicy" validate:"required"`
```
greenmoon55: Maybe we also need a "download policy" that expects the model is on the node without creating jobs. Just for special maintenances.

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,80 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageContainerSpec defines the container spec for the storage initializer init container, and the protocols it supports.
+// +k8s:openapi-gen=true
+type ClusterCachedModelSpec struct {
```
greenmoon55: better name?

pkg/apis/serving/v1alpha1/node_group_types.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ModelCacheNodeGroupSpec defines a group of nodes for to download the model to.
+// +k8s:openapi-gen=true
+type ModelCacheNodeGroupSpec struct {
+	StorageLimit resource.Quantity `json:"storageLimit"`
```
greenmoon55: We can use the storageLimit on the PersistentVolume instead of this field. However that field is immutable which is less flexible.

pkg/apis/serving/v1alpha1/node_group_types.go
```diff
@@ -0,0 +1,61 @@
...
+	NodeSelector map[string]string `json:"nodeSelector"`
```
greenmoon55: Discussed with @sukumargaonkar. Two approaches here.
Approach 1 - It's up to the user to make sure they use the correct label selector so two node groups never have overlaps.
Approach 2 - Require a specific label set on nodes. e.g. kserve-nodegroup-name: xyz
Otherwise if we allow overlapping in nodegroups. storage size calculation would be tricky.

pkg/apis/serving/v1alpha1/node_group_types.go
```diff
@@ -0,0 +1,61 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="persistentVolume is immutable"
+	// PV spec template
+	PersistentVolume corev1.PersistentVolume `json:"persistentVolume"`
```
greenmoon55: We will always create PVs for now. In the future, users might be able to leverage PersistentVolumeClaim with a custom CSI driver which makes sure different PVs map to the same disk location.

pkg/apis/serving/v1alpha1/cached_model_types.go
greenmoon55: Currently this is cluster scope only.

pkg/apis/serving/v1alpha1/node_group_types.go
greenmoon55: In the future we may have a namespace-scoped CR for NFS backed PV which does not share across namespaces. For non-local disks, sharing across namespaces is tricky as you would need different PVs and the volume plugin need to make sure they map to the same storage.

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,80 @@
...
+	// Container spec for the storage initializer init container
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="StorageUri is immutable"
+	StorageUri string            `json:"storageUri" validate:"required"`
+	ModelSize  resource.Quantity `json:"modelSize" validate:"required"`
+	NodeGroups []string          `json:"nodeGroups" validate:"required"`
+	// only local is supported for now
+	StorageType   StorageType   `json:"storageType" validate:"required"`
+	CleanupPolicy CleanupPolicy `json:"cleanupPolicy" validate:"required"`
+}
```
greenmoon55: Envvars for auth can be put in a configmap and added to the pod spec.
```
envFrom:
  - configMapRef:
      name: <config-file>
```

```
    - name: AWS_ACCESS_KEY_ID
      valueFrom:
        secretKeyRef:
          key: access-key
          name: name
```

pkg/apis/serving/v1alpha1/cached_model_types.go
greenmoon55: Should we remove this field?

pkg/apis/serving/v1alpha1/node_group_types.go
yuzisun: How can we make it namespace scoped? Now the PV/PVC is defined on the node group CR but this probably never should be made namespace scoped.

pkg/apis/serving/v1alpha1/cached_model_types.go
yuzisun: I still prefer the name `ClusterLocalModel` as the implementation of this CR is based on caching models locally.

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,85 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:openapi-gen=true
+type ClusterCachedModelSpec struct {
+	// Container spec for the storage initializer init container
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="StorageUri is immutable"
+	// Original StorageUri
+	StorageUri string `json:"storageUri" validate:"required"`
```
yuzisun: suggest renaming to `SourceModelUri`

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,85 @@
...
+	// Model size to make sure it does not exceed the disk space reserved for local models. The limit is defined on the NodeGroup.
+	ModelSize resource.Quantity `json:"modelSize" validate:"required"`
+	// group of nodes to cache the model on.
+	NodeGroups []string `json:"nodeGroups" validate:"required"`
+	// only local is supported for now
+	StorageType StorageType `json:"storageType" validate:"required"`
```
yuzisun: I think this CR is very specifically designed for local model cache, for other type of storage like AWS EBS fields like NodeGroup, cleanup policy would not apply I think.

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,85 @@
...
+// +kubebuilder:validation:Enum="";LocalPV
+type StorageType string
+
+// StorageType Enum values
+const (
+	LocalPV StorageType = "LocalPV"
+)
+
+// CleanupPolicy enum
+// +kubebuilder:validation:Enum="";DeleteModel;Ignore
```
yuzisun: what are the other possible values? this sounds like a flag

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -32,6 +32,7 @@ type StorageContainerSpec struct {
 
 	// List of URI formats that this container supports
 	SupportedUriFormats []SupportedUriFormat `json:"supportedUriFormats" validate:"required"`
+	UseCase             UseCase              `json:"useCase" validate:"required"`
```
yuzisun: suggest `workloadType`: InitContainer or Job

pkg/apis/serving/v1alpha1/cached_model_status.go
```diff
@@ -0,0 +1,64 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type CachedModelStatus struct {
+	OverallStatus OverallStatus         `json:"overallStatus,omitempty"`
```
yuzisun: can we just call status?

pkg/apis/serving/v1alpha1/cached_model_status.go
```diff
@@ -0,0 +1,64 @@
...
+	Name      string `json:"name,omitempty"`
+}
+
+// NodeStatus enum
+// +kubebuilder:validation:Enum="";NodeReady;NodeDownloading;NodeDownloadError;Deleting;NodeDeletionError;NodeDeleted
+type NodeStatus string
+
+// NodeStatus Enum values
+const (
+	NodeReady         NodeStatus = "NodeReady"
```
yuzisun: What does NodeReady mean? no models cache?

pkg/apis/serving/v1alpha1/cached_model_status.go
yuzisun: suggest `NodeModelStatus`

pkg/apis/serving/v1alpha1/cached_model_status.go
greenmoon55: Actually it means that it is cached. Maybe `Downloaded` is a better name.

pkg/apis/serving/v1alpha1/cached_model_types.go
greenmoon55: I was thinking about temporarily disabling deletion jobs for some reason, say we want to delete and then recreate the CR but keeping the files on the disk.

pkg/apis/serving/v1alpha1/cached_model_types.go
greenmoon55: yeah, I can't think of other values for now.

pkg/apis/serving/v1alpha1/node_group_types.go
greenmoon55: another namespace-scoped model cache CR that doesn't use nodegroup cr.

pkg/apis/serving/v1alpha1/node_group_types.go
```diff
@@ -0,0 +1,59 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ModelCacheNodeGroupSpec defines a group of nodes for to download the model to.
+// +k8s:openapi-gen=true
+type ModelCacheNodeGroupSpec struct {
+	NodeSelector map[string]string `json:"nodeSelector"`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="persistentVolume is immutable"
+	// PV spec template
+	PersistentVolume corev1.PersistentVolume `json:"persistentVolume"`
```
greenmoon55: Note that we rely on the storage limit on the PV spec which is immutable.

pkg/apis/serving/v1alpha1/cached_model_types.go
```diff
@@ -0,0 +1,85 @@
...
+type ClusterCachedModelSpec struct {
+	// Container spec for the storage initializer init container
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="StorageUri is immutable"
+	// Original StorageUri
+	StorageUri string `json:"storageUri" validate:"required"`
+	// Model size to make sure it does not exceed the disk space reserved for local models. The limit is defined on the NodeGroup.
+	ModelSize resource.Quantity `json:"modelSize" validate:"required"`
+	// group of nodes to cache the model on.
+	NodeGroups []string `json:"nodeGroups" validate:"required"`
```
greenmoon55: Ideally it should be mutable, with support for adding and removing a node group.

pkg/apis/serving/v1alpha1/cluster_local_model_node_group_types.go
```diff
@@ -21,39 +21,36 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// ModelCacheNodeGroupSpec defines a group of nodes for to download the model to.
+// ClusterLocalModelNodeGroupSpec defines a group of nodes for to download the model to.
```
greenmoon55: LocalModelNodeGroup?

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -46,14 +46,15 @@ const (
...
+// +kubebuilder:validation:Enum="";ModelReady;ModelNotReady;ModelPartiallyReady;ModelDeleting
+type ClusterLocalModelOverallStatus string
 
-// OverallStatus Enum values
+// ClusterLocalModelOverallStatus Enum values
 const (
-	ModelReady       OverallStatus = "ModelReady"
-	ModelDownloading OverallStatus = "ModelDownloading"
-	ModelDeleting    OverallStatus = "ModelDeleting"
+	ModelReady          ClusterLocalModelOverallStatus = "ModelReady"
```
greenmoon55: ClusterLocalModelStatus

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -46,14 +46,15 @@ const (
...
+	ModelNotReady       ClusterLocalModelOverallStatus = "ModelNotReady"
+	ModelPartiallyReady ClusterLocalModelOverallStatus = "ModelPartiallyReady"
+	ModelDeleting       ClusterLocalModelOverallStatus = "ModelDeleting"
 )
 
 type ModelCopies struct {
```
greenmoon55, line 50: total copies counts all nodes matching the node selector labels on the node group.
Total, available - download success, failed - download failures.

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
greenmoon55: Remove this overall status

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -32,7 +32,7 @@ type StorageContainerSpec struct {
 
 	// List of URI formats that this container supports
 	SupportedUriFormats []SupportedUriFormat `json:"supportedUriFormats" validate:"required"`
-	UseCase             UseCase              `json:"useCase" validate:"required"`
+	WorkloadType        WorkloadType         `json:"workloadType" validate:"required"`
```
greenmoon55: optional? default to initcontainer

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -32,6 +32,7 @@ type StorageContainerSpec struct {
 
 	// List of URI formats that this container supports
 	SupportedUriFormats []SupportedUriFormat `json:"supportedUriFormats" validate:"required"`
+	WorkloadType        WorkloadType         `json:"workloadType,omitempty"`
```
greenmoon55: not sure how to set a default here, maybe treat empty/null as initContainer in the code.

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
```diff
@@ -0,0 +1,56 @@
...
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LocalModelNodeGroupSpec defines a group of nodes for to download the model to.
+// +k8s:openapi-gen=true
+type LocalModelNodeGroupSpec struct {
```
greenmoon55, line 27: note that we cannot change the size limit on node groups because it is immutable on the pv spec.

greenmoon55:
Updated this PR and added node group status.

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
```diff
@@ -0,0 +1,57 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="persistentVolume is immutable"
+	PersistentVolumeSpec corev1.PersistentVolumeSpec `json:"persistentVolumeSpec"`
```
yuzisun, line 31: What happens if user makes a mistake and the node group is already in use. Is it safe to delete and recreate?

pkg/apis/serving/v1alpha1/storage_container_types.go
yuzisun: we can't use `// +kubebuilder:default` here?

greenmoon55: oh, I think we can, let me add it!

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
greenmoon55: good question! I haven't thought about it before. Maybe we can ignore "node group not found" errors in the model cache reconciler then. It would not affect existing models or services since the pvs and pvcs are per-model. Existing pvs and pvcs are not updated. However it would mess up with node storage space calculation, which we do not have yet.

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2024 The KServe Authors.
```

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,54 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type ClusterLocalModelStatus struct {
+	NodeStatus map[string]NodeStatus `json:"nodeStatus,omitempty"`
```
yuzisun, line 21: Add documentation for NodeStatus

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// How many nodes have the model available
+	// +optional
+	ModelCopies       *ModelCopies     `json:"copies,omitempty"`
+	InferenceServices []NamespacedName `json:"infereneceServices,omitempty"`
```
yuzisun: Add documentation for InferenceServices

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
```diff
@@ -0,0 +1,60 @@
...
+	// Original StorageUri
+	SourceModelUri string `json:"sourceModelUri" validate:"required"`
+	// Model size to make sure it does not exceed the disk space reserved for local models. The limit is defined on the NodeGroup.
+	ModelSize resource.Quantity `json:"modelSize" validate:"required"`
+	// group of nodes to cache the model on.
+	NodeGroup string `json:"nodeGroup" validate:"required"`
+}
+
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
yuzisun, line 36: deepcopy-gen is deprecated
https://kubebuilder.io/migration/legacy/migration_guide_v1tov2#migrate-the-apis

yuzisun: replaced with `+kubebuilder:object:root=true`

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
```diff
@@ -0,0 +1,60 @@
...
+type ClusterLocalModel struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ClusterLocalModelSpec   `json:"spec,omitempty"`
+	Status ClusterLocalModelStatus `json:"status,omitempty"`
+}
+
+// +k8s:openapi-gen=true
+// +kubebuilder:object:root=true
```
yuzisun, line 50: We only need this line

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:openapi-gen=true
+type ClusterLocalModelSpec struct {
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="StorageUri is immutable"
+	// Original StorageUri
+	SourceModelUri string `json:"sourceModelUri" validate:"required"`
```
yuzisun, line 28: maybe we need to consider supporting multiple source modelUris

pkg/apis/serving/v1alpha1/local_model_node_group_status.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright 2021 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2024 The KServe Authors.
```

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2023 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2024 The KServe Authors.
```

pkg/apis/serving/v1alpha1/local_model_node_group_status.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/api/resource"
+
+type LocalModelNodeGroupStatus struct {
+	Used      resource.Quantity `json:"used,omitempty"`
+	Available resource.Quantity `json:"available,omitempty"`
```
yuzisun, line 25: Add documentation

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
```diff
@@ -0,0 +1,55 @@
...
+// LocalModelNodeGroupSpec defines a group of nodes for to download the model to.
+// +k8s:openapi-gen=true
+type LocalModelNodeGroupSpec struct {
+	PersistentVolumeSpec      corev1.PersistentVolumeSpec      `json:"persistentVolumeSpec"`
+	PersistentVolumeClaimSpec corev1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec"`
+}
+
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
```
yuzisun, line 38:
```suggestion
```

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
greenmoon55: Somehow we still call `deepcopy-gen` and removing it would cause an error while generating swagger files.

pkg/apis/serving/v1alpha1/local_model_node_group_status.go
greenmoon55: Added, btw I'm planning to implement the storage size tracking later.

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
greenmoon55: We can make it a list of models here or support multiple local model crs for one inference service.

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,56 @@
...
+	NodeStatus map[string]NodeStatus `json:"nodeStatus,omitempty"`
+
+	// How many nodes have the model available locally
+	// +optional
+	ModelCopies *ModelCopies `json:"copies,omitempty"`
+	// Inference services using this local model
+	InferenceServices []NamespacedName `json:"inferenceServices,omitempty"`
+}
+
+type NamespacedName struct {
```
greenmoon55, line 30: Defining a new struct here because I got `encountered struct field "Namespace" without JSON tag in type "NamespacedName"` if I use `types.NamespacedName`.

pkg/apis/serving/v1alpha1/cluster_local_model_types.go
greenmoon55: same as above

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,56 @@
...
+	Namespace string `json:"namespace,omitempty"`
+	Name      string `json:"name,omitempty"`
+}
+
+// NodeStatus enum
+// +kubebuilder:validation:Enum="";NodeNotReady;NodeDownloading;NodeDownloaded;NodeDownloadError;NodeDeleting;NodeDeletionError;NodeDeleted
+type NodeStatus string
+
+// NodeStatus Enum values
+const (
```
greenmoon55, line 40: ~With the daemonset in the future, Node status should be `ModelExists` or `ModelDoesNotExist` reported by the daemonset. For other status, we can look at individual jobs?~
It still makes sense if we do not go down the daemonset approach.

pkg/apis/serving/v1alpha1/cluster_local_model_status.go
```diff
@@ -0,0 +1,56 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type ClusterLocalModelStatus struct {
+	// Status of the model on a node, like NodeDownloaded or NodeNotReady
+	NodeStatus map[string]NodeStatus `json:"nodeStatus,omitempty"`
```
yuzisun, line 21: I think we can create a LocalModelNodeGroup controller and create the LocalModelNode resource per node, in this way we can spawn the model reconciler daemonset and update the status on each LocalModelNode.

yuzisun:
Discussed with @greenmoon55 to create a follow up PR to add a LocalModelNode custom resource to track the model status for each node in the group.

yuzisun (approved):

PR #3863, opened 2024-08-15
Add TLS support for Inference Loggers

### LINKED ISSUE #3837: Support for TLS enabled logger endpoints
/kind feature

KServe's Inference Logger producer does not support TLS-enabled endpoints, as the [logger handler](https://github.com/kserve/kserve/blob/master/pkg/logger/handler.go) does not implement TLS-related parameters and implementing the necessary logic to establish secure connections.

Althought this feature proposal does not include the mechanism to pass certificates to the agent, we can assume that if those certificates are available, then the logger could properly configure the transport according to the protocol (ie. either `http` or `https`).

cc @terrytangyuan @danielezonca

ruivieira:

This PR adds support for TLS in Inference Loggers.
It adds two new optional fields to the global ConfigMap: `tlsBundle` and `tlsCertName`.

If a ConfigMap matching `tlsBundle` is present in the Inference Service's namespace, it will be mounted as a volume on the agent. 
* If the logger's endpoint protocol is of type `https`, the logger will use the CA cert specified with `tlsCertName` to configure the transport.
  * If TLS is configured in ConfigMap, but the bundle is not available or the CA cert is not found, the logger will continue to behave as before this PR (no TLS transport configured)
* If the logger's endpoint protocol is of type `http`, the inference logger will behave as previously

_why we need it_

The current implementation does not configure HTTP transport for TLS.

Fixes https://github.com/kserve/kserve/issues/3837

**Type of changes**

- [x] Exploratory testing on with development deployments
- [x] Agent injection test

- Logs

**Checklist**:

**Release note**:

```release-note
TLS support for Inference Logger by specifying certificate bundle in the global configuration
```

**Re-running failed tests**


cmd/agent/main.go
```diff
@@ -74,6 +74,7 @@ var (
 	readinessProbeTimeout = flag.Duration("probe-period", -1, "run readiness probe with given timeout") //nolint: unused
 	// This creates an abstract socket instead of an actual file.
 	unixSocketPath = "@/kserve/agent.sock"
+	tlsCertName    = flag.String("log-tls-cert", "service-ca.crt", "The TLS certificate name")
```
yuzisun: How are we mounting the ca certificate to the container?

yuzisun: it is probably better to be mounted as a volume

cmd/agent/main.go
```diff
@@ -286,6 +288,7 @@ func startLogger(workers int, logger *zap.SugaredLogger) *loggerArgs {
 		endpoint:         *endpoint,
 		namespace:        *namespace,
 		component:        *component,
+		tlsCertName:      *tlsCertName,
```
yuzisun: to be precise this is CA certificate

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -56,6 +60,8 @@ type LoggerConfig struct {
 	MemoryRequest string `json:"memoryRequest"`
 	MemoryLimit   string `json:"memoryLimit"`
 	DefaultUrl    string `json:"defaultUrl"`
+	TlsBundle     string `json:"tlsBundle"`
```
yuzisun: Suggest renaming to CABundle

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -277,6 +288,30 @@ func (ag *AgentInjector) InjectAgent(pod *v1.Pod) error {
 		},
 	}
 
+	// If the Logger TLS bundle ConfigMap is specified, mount it
+	if injectLogger && ag.loggerConfig.TlsBundle != "" {
+		// Optional. If the ConfigMap is not found, this will not make the Pod fail
+		configMapVolume := v1.VolumeSource{
+			ConfigMap: &v1.ConfigMapVolumeSource{
+				LocalObjectReference: v1.LocalObjectReference{
+					Name: ag.loggerConfig.TlsBundle,
```
yuzisun: Is the CA bundle configmap available in each namespace?

cmd/agent/main.go
yuzisun: Also this is a file not a name

cmd/agent/main.go
ruivieira: Renamed to `caCertName`

pkg/webhook/admission/pod/agent_injector.go
ruivieira: Renamed to `CaBundle`

pkg/webhook/admission/pod/agent_injector.go
ruivieira: Hi @yuzisun, addressing this question and the one [above](https://github.com/kserve/kserve/pull/3863#discussion_r1720545006).

The CA bundle's name is specified in the global ConfigMap.
When the agent is created, a ConfigMap from the same namespace as the InferenceService and with that matching name is mounted as a volume and used to configure the transport.
Since the previous behaviour (before this PR) was that `https` could be specified as a logger endpoint, although with no TLS configuration, the following is expected in order keep this backwards compatible with the previous behaviour:
- If the ConfigMap is _not_ present in the InferenceService's namespace and the logger endpoint is `https`, continue using that endpoint but with no TLS configured, i.e. the ConfigMap is optional and it won't cause the agent to fail if missing

pkg/logger/worker.go
```diff
@@ -85,6 +90,26 @@ func (w *Worker) sendCloudEvent(logReq LogRequest) error {
 		return fmt.Errorf("while creating http transport: %w", err)
 	}
 
+	if logReq.Url.Scheme == "https" {
+		caCertFile, err := os.ReadFile(filepath.Join(CaCertPath, logReq.TlsCertName))
+		// Do not fail if certificates not found, for backwards compatibility
```
spolti, line 96: maybe log a warning might be good, missing the cert might be not intentional.

ruivieira: Added log warning for the case when logger endpoint is https but CA cert file not found

ruivieira:
@yuzisun @spolti Is there anything still needing to be addressed from the review's comments? Thank you!

terrytangyuan:
Can you fix the DCO check?

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -39,6 +40,9 @@ const (
 	LoggerArgumentNamespace        = "--namespace"
 	LoggerArgumentEndpoint         = "--endpoint"
 	LoggerArgumentComponent        = "--component"
+	LoggerCaBundleVolume           = "agent-ca-bundle"
+	LoggerCaCertMountPath          = "/etc/tls/logger"
```
yuzisun: move the path to a constant, it is used in multiple places.

pkg/logger/worker.go
```diff
@@ -85,6 +90,29 @@ func (w *Worker) sendCloudEvent(logReq LogRequest) error {
...
+			clientCertPool := x509.NewCertPool()
+			if !clientCertPool.AppendCertsFromPEM(caCertFile) {
+				return fmt.Errorf("while parsing CA certificate")
+			}
+
+			tlsTransport := &http.Transport{
+				TLSClientConfig: &tls.Config{
+					RootCAs:            clientCertPool,
+					MinVersion:         tls.VersionTLS12,
+					InsecureSkipVerify: false,
```
yuzisun: Should we also consider proving an option to skip TLS verification?

pkg/webhook/admission/pod/agent_injector_test.go
```diff
@@ -55,6 +56,12 @@ var (
 		Image:      "gcr.io/kfserving/agent:latest",
 		DefaultUrl: "http://httpbin.org/",
 	}
+	loggerConfigTls = &LoggerConfig{
```
yuzisun:
```suggestion
	loggerTLSConfig = &LoggerConfig{
```

pkg/logger/types.go
```diff
@@ -32,4 +32,5 @@ type LogRequest struct {
 	Component        string
 	Endpoint         string
 	Metadata         map[string][]string
+	TlsCertName      string
```
Jooho: Is it a server certificate? so I think certName is ok.

```suggestion
	certName      string
```

pkg/logger/handler.go
```diff
@@ -43,10 +43,11 @@ type LoggerHandler struct {
 	endpoint         string
 	next             http.Handler
 	metadataHeaders  []string
+	tlsCertName      string
```
Jooho: this is the same 
```suggestion
	certName      string
```

pkg/logger/handler.go
```diff
@@ -43,10 +43,11 @@ type LoggerHandler struct {
...
 }
 
 func New(logUrl *url.URL, sourceUri *url.URL, logMode v1beta1.LoggerType,
-	inferenceService string, namespace string, endpoint string, component string, next http.Handler, metadataHeaders []string) http.Handler {
+	inferenceService string, namespace string, endpoint string, component string, next http.Handler, metadataHeaders []string, tlsCertName string) http.Handler {
```
Jooho:
```suggestion
	inferenceService string, namespace string, endpoint string, component string, next http.Handler, metadataHeaders []string, certName string) http.Handler {
```

pkg/logger/handler.go
```diff
@@ -59,6 +60,7 @@ func New(logUrl *url.URL, sourceUri *url.URL, logMode v1beta1.LoggerType,
 		endpoint:         endpoint,
 		next:             next,
 		metadataHeaders:  metadataHeaders,
+		tlsCertName:      tlsCertName,
```
Jooho: same 
```suggestion
		certName:      certName,
```

pkg/logger/handler.go
```diff
@@ -113,6 +115,7 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			Endpoint:         eh.endpoint,
 			Component:        eh.component,
 			Metadata:         metadata,
+			TlsCertName:      eh.tlsCertName,
```
Jooho:
```suggestion
			certName:      eh.certName,
```

pkg/logger/handler.go
```diff
@@ -141,6 +144,7 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				Namespace:        eh.namespace,
 				Endpoint:         eh.endpoint,
 				Component:        eh.component,
+				TlsCertName:      eh.tlsCertName,
```
Jooho:
```suggestion
				certName:      eh.certName,
```

terrytangyuan (commented):
PTAL @yuzisun

ruivieira:
@terrytangyuan This is don't have the support for skipping TLS verification, I hope to push that today and comment here when done.
@yuzisun @terrytangyuan Another question, at the moment this PR does include a default value for the cert fields in the ConfigMap. Is this something that should be added as part of this PR, or left without a default CM?

ruivieira:
@terrytangyuan @yuzisun The changes are ready for review now.

terrytangyuan:
We can probably add default later. WDYT? @yuzisun

yuzisun (dismissed):
/lgtm
/approve

yuzisun:
@ruivieira Can you help update the logger doc for this on the kserve website repo?

ruivieira:
@yuzisun @terrytangyuan Had to add a new commit to fix the linting errors:

- Restore newline at the end of charts/kserve-resources/README.md
- Remove import of github.com/kserve/kserve/pkg/constants from `worker.go` and replace with local constant for the CA mount path
- `InsecureSkipVerify: logReq.TlsSkipVerify` was triggering gosec's G402 with "potential 'true' for `logReq.TlsSkipVerify`". Since this value is allowed to be true, this specific line was excluded from the checks and an explanatory comment added
- Remove import of `k8s.io/utils/ptr` and replace with a pointer `&optionalVolume`

I'm unsure about the gosec error in particular, since if InsecureSkipVerify is to be configurable it can by definition be `true` and will always trigger this warning.

ruivieira:
> @ruivieira Can you help update the logger doc for this on the kserve website repo?

Of course :+1:

yuzisun (approved):

PR #3885, opened 2024-08-22
Set the volume mount's readonly annotation based on the ISVC annotation

### LINKED ISSUE #3687: Allow PVC Model Mount in ReadWrite Mode
/kind feature

I would like to add an additional flag to specify a volume to be ReadWrite or ReadOnly.

Currently the implementation for PVC model stores [here](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L255) expects that a PV containing model files is always mounted to a `ServingRuntime` as `ReadOnly`. I have a use case where my server has some internal logic that I would like to dynamically download model components and cache them to the local PV.

As an alternative, this could be a use case for the new model-car, but my download logic is somewhat complex and I need to run the logic in many locations outside of KServe, so keeping the logic inside the container works best for me. This may also be a use case for a custom StorageContainer, but I think the change is small enough and common enough that other users might find this feature useful.

**Design proposal:**
Add a readonly flag under storageUri:
```
storageUri: "pvc://pvc-name/model-store/" 
  readonly: true # default true
```

See Slack for more details: https://cloud-native.slack.com/archives/C06AH2C3K8B/p1714883766854429

hdefazio:
Fixes #3687

**Type of changes**
 Bug fix (non-breaking change which fixes an issue)

Test Setup:
- Created a cluster
- Installed Service Mesh, Serverless
- Replaced the image in deployments -> kserve-controller  with my local image in the  opendatahub namespace
- Created a serving runtime via yaml config
- Created an ISVC via a yaml config

Test Cases:
storage.kserve.io/readonly: "false"
readonly_false

storage.kserve.io/readonly: "true"
readonly_true

storage.kserve.io/readonly was not provided
readonly_unset

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -307,7 +307,7 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 		pvcSourceVolumeMount := v1.VolumeMount{
 			Name:      PvcSourceMountName,
 			MountPath: PvcSourceMountPath,
-			ReadOnly:  true,
+			ReadOnly:  isvcReadonlyStringFlag,
 		}
 		storageInitializerMounts = append(storageInitializerMounts, pvcSourceVolumeMount)
 
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -184,7 +184,7 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *v1.Pod) error {
 
 // InjectStorageInitializer injects an init container to provision model data
 // for the serving container in a unified way across storage tech by injecting
-// a provisioning INIT container. This is a work around because KNative does not
+// a provisioning INIT container. This is a workaround because KNative does not
 // support INIT containers: https://github.com/knative/serving/issues/4307
 func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) error {
 	// Only inject if the required annotations are set
@@ -207,6 +207,15 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 	for _, container := range pod.Spec.InitContainers {
 		if strings.Compare(container.Name, StorageInitializerContainerName) == 0 {
 			return nil
+		}
+	}
+
+	// Update volume mount's readonly annotation based on the ISVC annotation
+	isvcReadonlyStringFlag := true
+	isvcReadonlyString, ok := pod.ObjectMeta.Annotations[constants.StorageReadonlyAnnotationKey]
+	if ok {
+		if isvcReadonlyString == "false" {
+			isvcReadonlyStringFlag = false
 		}
 	}
 
@@ -252,7 +261,7 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 				MountPath: constants.DefaultModelLocalMountPath,
 				// only path to volume's root ("") or folder is supported
 				SubPath:  pvcPath,
-				ReadOnly: true,
+				ReadOnly: isvcReadonlyStringFlag,
 			}
 
 			// Check if PVC source URIs is already mounted
@@ -361,7 +370,7 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 	sharedVolumeReadMount := v1.VolumeMount{
 		Name:      StorageInitializerVolumeName,
 		MountPath: constants.DefaultModelLocalMountPath,
-		ReadOnly:  true,
+		ReadOnly:  isvcReadonlyStringFlag,
 	}
 	userContainer.VolumeMounts = append(userContainer.VolumeMounts, sharedVolumeReadMount)
 	if transformerContainer != nil {
--- a/pkg/webhook/admission/pod/storage_initializer_injector_test.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector_test.go
@@ -202,6 +202,131 @@ func TestStorageInitializerInjector(t *testing.T) {
 				},
 			},
 		},
+
+		"StorageInitializerInjectedReadOnlyFalse": {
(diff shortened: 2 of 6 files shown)


terrytangyuan (commented):
cc @yuzisun @lizzzcai @supertetelman

yuzisun:
@hdefazio Can you help sign off the commits ?

hdefazio:
@yuzisun I have signed the commits and DCO is now passing

terrytangyuan:
I just manually triggered the CI builds.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -252,7 +261,7 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *v1.Pod) erro
 				MountPath: constants.DefaultModelLocalMountPath,
 				// only path to volume's root ("") or folder is supported
 				SubPath:  pvcPath,
-				ReadOnly: true,
+				ReadOnly: isvcReadonlyStringFlag,
```
lizzzcai, line 264: how about this [ReadOnly](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector.go#L310) parameter, need to it as well?

spolti: Hi @lizzzcai this one refers to the storage Initializer, I guess it only reads the model's directory.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -202,6 +202,192 @@ func TestStorageInitializerInjector(t *testing.T) {
 				},
 			},
 		},
+
+		"StorageInitializerInjectedReadOnlyUnset": {
```
lizzzcai, line 206: can add tests under the [directVolumeMountForPvc](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector_test.go#L2085) scenario as well?

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -202,6 +202,192 @@ func TestStorageInitializerInjector(t *testing.T) {
...
+						constants.StorageInitializerSourceUriInternalAnnotationKey: "gs://foo",
+					},
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name: constants.InferenceServiceContainerName,
+							VolumeMounts: []v1.VolumeMount{
+								{
+									Name:      "kserve-provision-location",
```
lizzzcai: nit: should we use `constants.StorageInitializerVolumeName`?

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: noob question: Could you clarify your use case? From https://github.com/kserve/kserve/issues/3687,  Do you download in the predictor container? 
In our use case, we would like to use the storage initializer to download to the pvc. So we'd like to make this [ReadOnly](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector.go#L310) parameter configurable, and ideally keep these two ReadOnly (L264 and L373) false.

greenmoon55: noob question: Could you clarify your use case? From #3687, Do you download in the predictor container?
In our use case, we would like to download to pvc in the storage initializer, which means we probably only want this [ReadOnly](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector.go#L310) parameter to be `false`. Is it a good idea to differentiate these two use cases? Like using a different annotation?

lizzzcai: Hi @spolti I am not sure if there is a scenario that the user want to modify the existing file in the PVC, and whether it will work with the `ReadOnly: true` [here](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector.go#L310) and the symlink file under the storage initializer approach of mounting pvc.
With the changes [here](https://github.com/kserve/kserve/blob/b3019774251f8c64d3c1863c6564c2ef7b6592b5/pkg/webhook/admission/pod/storage_initializer_injector.go#L264), the above scenario should work with the `directPvcVolumeMount` which is the default way of mounting PVC in KServe now.

Hi @greenmoon55, if my understanding correctly, your scenario is downloading a model from cloud storage via storage initializer to a PVC rather than emptyDir, like what I described [here](https://github.com/kserve/kserve/issues/3687#issuecomment-2109717856). If yes, I think it is a different use case from this PR.

spolti: To satisfy all use cases we might need to set it for storage-initializer as well.

@hdefazio can you please handle it?

Jooho:
```
Created a cluster
Installed Service Mesh, Serverless, Opendatahub operators (in that order)
Created a DSCI and waited for it to be Ready
Created a DSC and waited for it to be Ready
Scaled deployments -> opendatahub-operator-controller-manager down to zero in the openshift-operators namespace
Replaced the image in deployments -> kserve-controller with my local image in the opendatahub namespace
Created a serving runtime via yaml config
Created an ISVC via a yaml config
```
@hdefazio This is not Opendatahub kserve repository. Could you please update Test Setup steps?

pkg/webhook/admission/pod/storage_initializer_injector.go
hdefazio: @spolti @lizzzcai I have updated this to include the storage-initializer if you'd like to double check

lizzzcai:
@hdefazio please signoff your commits.

/lgtm

spolti (approved):

yuzisun (approved):

PR #3886, opened 2024-08-23
Fix the 'tokens exceeding model limit' error response in vllm server

saileshd1402:
The error returned when total token length exceeds context length is currently returning an InternalServerError (500) when it should return a BadRequest error (400) in VLLM backend with huggingfaceserver predictor. (This is consistent with the error status that vLLM returns)

Example:
Current - InternalServerError (500):
```
{
  "error": {
    "code": "500",
    "message": "This model's maximum context length is 8192 tokens. However, you requested 50014 tokens (14 in the messages, 50000 in the completion). Please reduce the length of the messages or completion.",
    "param": "",
    "type": "OpenAIError"
  }
}
```

Fix - BadRequest error (400):
```
{
  "error": {
    "code": "400",
    "message": "This model's maximum context length is 8192 tokens. However, you requested 50014 tokens (14 in the messages, 50000 in the completion). Please reduce the length of the messages or completion.",
    "param": "",
    "type": "BadRequest"
  }
}
```

Fixes #

**Type of changes**

Testing the error for Completions and ChatCompletions with and without streaming

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py
+++ b/python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py
@@ -25,6 +25,7 @@ from typing import (
     Union,
     Iterator,
 )
+from http import HTTPStatus
 
 import torch
 from vllm import PoolingParams
@@ -50,8 +51,7 @@ from kserve.protocol.rest.openai.types.openapi import (
     CreateCompletionResponse as Completion,
     Logprobs,
 )
-from kserve.errors import InvalidInput
-from kserve.protocol.rest.openai.errors import OpenAIError
+from kserve.protocol.rest.openai.errors import OpenAIError, create_error_response
 from kserve.protocol.rest.openai import ChatCompletionRequestMessage, CompletionRequest
 
 
@@ -159,7 +159,7 @@ class OpenAIServingCompletion:
                     )
                 )
         except Exception as e:
-            raise OpenAIError(str(e))
+            raise e if isinstance(e, OpenAIError) else OpenAIError(str(e))
 
         result_generator: AsyncIterator[Tuple[int, RequestOutput]] = (
             merge_async_iterators(*generators)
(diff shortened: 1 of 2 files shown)


python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py
```diff
@@ -388,13 +388,17 @@ def _validate_input(
             request.max_tokens = self.max_model_len - token_num
 
         if token_num + request.max_tokens > self.max_model_len:
-            raise InvalidInput(
```
yuzisun, line 391: https://github.com/kserve/kserve/blob/e82beb6c4cdac44d15ebc900d49d1be947861d01/python/kserve/kserve/protocol/rest/server.py#L88

hmm InvalidInput is supposed to return 400
https://github.com/kserve/kserve/blob/e82beb6c4cdac44d15ebc900d49d1be947861d01/python/kserve/kserve/errors.py#L110

saileshd1402: That exception is being caught here and getting converted to OpenAIError: https://github.com/kserve/kserve/blob/7dd4ae2e5c6c821a4a3b875ffcc35a70b18a3234/python/huggingfaceserver/huggingfaceserver/vllm/vllm_completions.py#L162

Even if we return it as InvalidInput error, the format will be in kserve error format which is inconsistent with the rest of the errors returned here. I mentioned the difference in an issue I created: https://github.com/kserve/kserve/issues/3887

yuzisun: Gotcha, for openai completion and chat completion endpoints we should return the OpenAIError. We will check the case for huggingface backend.

yuzisun (approved):
/lgtm
/approve

PR #3902, opened 2024-08-28
storage initializer segregation

### LINKED ISSUE #3489: Make storage initializer install only what is needed for it to run

spolti:
Fixes https://github.com/kserve/kserve/issues/3489

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


terrytangyuan (commented):
Can you add the motivation in the PR description? 

Also, removing it would break compatibility since the package currently includes it in the distribution so we might need to include this change in a major version

spolti:
> Can you add the motivation in the PR description?

It is in the link in the pr description, should I replicate it here?


> Also, removing it would break compatibility since the package currently includes it in the distribution so we might need to include this change in a major version

indeed, I would need help on how to publish it as part of the Python SDK (https://pypi.org/project/kserve/)

python/kserve/pyproject.toml
```diff
@@ -53,35 +53,20 @@ pandas = ">=1.3.5"
...
-
 # Logging dependencies. They can be opted into by apps.
-asgi-logger = { version = "^0.1.0", optional = true, python = ">3.8.0,<3.11" }
+asgi-logger = { version = "^0.1.0", optional = true, python = ">3.8.0,<3.12" }
 
 # Ray dependency
 ray = { version = "~2.10.0", extras = ["serve"] }
 
 [tool.poetry.extras]
-storage = [
```
sivanantha321, line 76: We can still include the storage package as an optional dependency , So that it won't be a breaking change. This will give users some time to migrate to the new storage package. Maybe then we can remove it from Kserve sdk.

python/storage/logging.py
```diff
@@ -0,0 +1,140 @@
...
+            "()": "logging.Formatter",
+            "fmt": KSERVE_LOGGER_FORMAT,
+            "datefmt": KSERVE_LOGGER_DATE_FORMAT,
+        },
+        "kserve_trace": {
+            "()": "logging.Formatter",
+            "fmt": KSERVE_TRACE_LOGGER_FORMAT,
+            "datefmt": KSERVE_LOGGER_DATE_FORMAT,
+        },
+        "uvicorn": {
```
sivanantha321: I think we don't need to configure uvicorn in storage initializer

python/storage/logging.py
```diff
@@ -0,0 +1,140 @@
...
+KSERVE_LOG_CONFIG = {
+    "version": 1,
+    "disable_existing_loggers": False,
+    "formatters": {
+        "kserve": {
+            "()": "logging.Formatter",
+            "fmt": KSERVE_LOGGER_FORMAT,
+            "datefmt": KSERVE_LOGGER_DATE_FORMAT,
+        },
+        "kserve_trace": {
```
sivanantha321: Do we need kserve_trace ?

python/storage-initializer/scripts/initializer-entrypoint
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 import sys
+from logging import configure_logging, logger
+from storage import Storage
 
-from kserve.storage import Storage
-from kserve.logging import configure_logging, logger
 
 configure_logging()
```
sivanantha321, line 6: How is the logger configured here ? I mean we don't have any specific args

sivanantha321:
Fixes #3905

python/storage-initializer/scripts/initializer-entrypoint
spolti: It is a try to fix the storage_initializer container image:
```
ImportError: cannot import name 'configure_logging' from 'logging' (/usr/local/lib/python3.11/logging/__init__.py)
```

python/storage/logging.py
spolti: for now, I just replicated the logging.py file.
We can probably clean it.

python/kserve/pyproject.toml
spolti: do you mean keep storage module around for a while?

python/storage/logging.py
spolti: I've cleaned it a few, please take a look

python/storage/logging.py
spolti: removed.

python/kserve/pyproject.toml
spolti: added back the storage package inside `python/kserve`.

python/storage-initializer/scripts/initializer-entrypoint
spolti: Worked on a local build, hopefully CI will be green as well :)

python/kserve/pyproject.toml
spolti: @sivanantha321  If I add this to the `kserve's` _pyproject,toml_:

```yaml
# Storage dependencies. They can be opted into by apps.
storage = { path = "../storage", extras = ["storage"], optional = true, develop = true }
```

The dependencies will be updated to:

```
--- a/python/pmmlserver/poetry.lock
+++ b/python/pmmlserver/poetry.lock
@@ -1148,7 +1148,7 @@ uvicorn = {version = "^0.21.1", extras = ["standard"]}
 [package.extras]
 logging = ["asgi-logger (>=0.1.0,<0.2.0)"]
 ray = ["ray[serve] (>=2.10.0,<2.11.0)"]
-storage = ["azure-identity (>=1.8.0,<2.0.0)", "azure-storage-blob (>=12.10.0,<13.0.0)", "azure-storage-file-share (>=12.7.0,<13.0.0)", "boto3 (>=1.21.0,<2.0.0)", "google-cloud-storage (>=2.3.0,<3.0.0)", "huggingface-hub (>=0.24.5,<0.25.0)", "requests (>=2.20.0,<3.0.0)"]
+storage = ["storage[storage] @ file:///Users/fspolti/data/dev/sources/kserve/python/storage"]
```

python/storage/pyproject.toml
```diff
@@ -0,0 +1,54 @@
+[tool.poetry]
+name = "storage"
```
sivanantha321, line 2: Maybe rename it to kserve-storage ?

python/kserve/pyproject.toml
sivanantha321: it should be something like this
```
storage = { path = "../storage", optional = true, develop = true }

[tool.poetry.extras]
storage = [
"storage"
]

```

python/lgbserver/lgbserver/model.py
```diff
@@ -20,7 +20,7 @@
 
 from kserve import Model
 from kserve.errors import InferenceError, ModelMissingError
-from kserve.storage import Storage
+from storage import Storage
```
sivanantha321, line 23: In the kserve/storage directory, we can do something like this for backward compatibility
```
__init__.py
    try:
       from storage import *
    except ImportError as e:
       Print("Please install storage package")
       raise from e
       
```

python/storage/pyproject.toml
spolti: I am ok with both names, let's see what others think :)

cc @yuzisun @terrytangyuan

python/kserve/pyproject.toml
spolti: That was what I did.
https://github.com/spolti/kserve/commit/c54fcd37464a1a32f08f596b9081c93c175f523b

python/lgbserver/lgbserver/model.py
spolti: That's a good idea, but I was not able to make it work, can you please take a look here and see what I might be missing: 
https://github.com/spolti/kserve/commit/c54fcd37464a1a32f08f596b9081c93c175f523b

python/kserve/pyproject.toml
spolti: @sivanantha321 this line specifically:
https://github.com/spolti/kserve/commit/c54fcd37464a1a32f08f596b9081c93c175f523b#diff-ad2d8c87e4697eeeb1827029aaf92553c8ce992e2a65850374e3d1448e27f144R658

python/storage/pyproject.toml
spolti: @yuzisun @terrytangyuan

python/lgbserver/lgbserver/model.py
spolti: @sivanantha321 I guess it worked now, can you please check the last commit?

spolti: Just adding the storage module as optional should be enough not to break backward compatibility.

spolti:
@terrytangyuan @yuzisun can you please review it again?

python/storage/storage/storage_logging.py
```diff
@@ -0,0 +1,109 @@
...
+            loaded_config = json.load(file)
+            logging.config.dictConfig(loaded_config)
+    elif log_config.endswith((".yaml", ".yml")):
+        with open(log_config) as file:
+            loaded_config = yaml.safe_load(file)
+            logging.config.dictConfig(loaded_config)
+    else:
+        # See the note about fileConfig() here:
+        # https://docs.python.org/3/library/logging.config.html#configuration-file-format
+        logging.config.fileConfig(log_config, disable_existing_loggers=False)
```
rnetser, line 109: how about putting this block under `isinstance(log_config, str)` else raise ?

spolti: Hi @rnetser, I think I didn't understand what you meant.

python/kserve/pyproject.toml
spolti: @sivanantha321 I've readded the _init_.py for backward compatibility; however, for our modules, I've updated everything to the new path.

python/kserve/kserve/storage/__init__.py
```diff
@@ -11,7 +11,15 @@
...
-from kserve.storage.storage import Storage
+# Keep backwards capability, try import storage package.
+# This way, existing projects using the Python SDK will not break, allowing
+# users to upgrade the SDK without changing their code.
+# We might need to inform users about the segregation of the storage package
+# so they have time to update the code and change the imports to the new package.
+try:
+    from storage import *
+except ImportError as e:
+    print("Please install the storage package")
```
terrytangyuan: No need to print the duplicate message

python/kserve/kserve/storage/__init__.py
```diff
@@ -11,7 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 # flake8: noqa
 
-from kserve.storage.storage import Storage
+# Keep backwards capability, try import storage package.
+# This way, existing projects using the Python SDK will not break, allowing
+# users to upgrade the SDK without changing their code.
```
terrytangyuan: Can you test the backwards compatibility?

python/kserve/kserve/storage/__init__.py
```diff
@@ -11,7 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 # flake8: noqa
 
-from kserve.storage.storage import Storage
+# Keep backwards capability, try import storage package.
```
terrytangyuan: compatibility?

test/scripts/gh-actions/setup-kserve.sh
```diff
@@ -49,3 +49,7 @@ echo "Installing KServe Python SDK ..."
 pushd python/kserve >/dev/null
     poetry install --with=test --no-interaction
 popd
+echo "Installing KServe Storage ..."
```
terrytangyuan: storage module

python/kserve/kserve/storage/__init__.py
spolti: I've tested manually, let me try to add a test case.

spolti: added a test.

python/kserve/kserve/storage/__init__.py
spolti: removed.

python/kserve/kserve/storage/__init__.py
spolti: yes, typo :)
thanks.

test/scripts/gh-actions/setup-kserve.sh
spolti: updated.

python/kserve/pyproject.toml
spolti: @sivanantha321 please take a look, I've also added a new git action to test without storage and with the storage module installed.

python/kserve/kserve/storage/__init__.py
spolti: https://github.com/kserve/kserve/actions/runs/12678720187/job/35336887961?pr=3902

spolti: @sivanantha321 ^

python/kserve/pyproject.toml
```diff
@@ -54,38 +53,27 @@ pandas =  "^2.2.0"
...
-requests = { version = "^2.32.2", optional = true }
-google-cloud-storage = { version = "^2.14.0", optional = true }
-azure-storage-blob = { version = "^12.20.0", optional = true }
-azure-storage-file-share = { version = "^12.16.0", optional = true }
-azure-identity = { version = "^1.15.0", optional = true }
-boto3 = { version = "^1.29.0", optional = true }
-huggingface-hub = { version = "^0.24.5", extras = ["hf-transfer"], optional = true }
-
+# KServe Storage module. They can be opted into by apps.
+storage = { path = "../storage", optional = true, develop = true }
```
sivanantha321: I am little confused here. If we publish the kserve package to pypi, how this will work

python/storage-initializer/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+    "Operating System :: OS Independent",
+]
+
+packages = [
+    { include = "scripts/initializer-entrypoint" }
+]
+
+[tool.poetry.dependencies]
+python = ">=3.9,<3.13"
+storage = { path = "../storage", extras = ["logging"], develop = true }
```
sivanantha321, line 20: is there any reason to separate storage and storage-initializer

spolti: No, I just kept it as it is, and moved the storage module outside KServe.

python/kserve/kserve/storage/tests/test_storage_initializer_backwards_compatibility.py
```diff
@@ -0,0 +1,30 @@
...
+        import kserve.storage as Storage
+
+        Storage.download("s3://test", "test")
+
+
+def test_kserve_module_installed():
+    import kserve.storage as Storage
+
+    # it should return TypeError: Storage.download() missing 1 required positional argument: 'uri'
+    with pytest.raises(TypeError):
```
rnetser: `pytest.raises` takes `match` so you can make sure that it fails on something specific and not any `TypeError`

python/storage/storage/storage_logging.py
```diff
@@ -0,0 +1,109 @@
...
+    },
+}
+
+logger = logging.getLogger(STORAGE_INITIALIZER_LOGGER_NAME)
+trace_logger = logging.getLogger(STORAGE_INITIALIZER_TRACE_LOGGER_NAME)
+
+
+def configure_logging(log_config: Optional[Union[Dict, str]] = None):
+    """
+    Configures Kserve and Uvicorn loggers.
```
rnetser:
```suggestion
    Configures Kserve storage loggers.
```
?

python/storage/storage/storage_logging.py
rnetser: @spolti resolving this; `fileConfig` will do the validations

python/storage-initializer/pyproject.toml
spolti: Do you think we can keep it this way, as the storage_initializer is just a wrapper for calling the storage methods?

python/kserve/pyproject.toml
spolti: Good point, it shouldn't be here. updated.

python/kserve/kserve/storage/tests/test_storage_initializer_backwards_compatibility.py
spolti: Very nice @rnetser, updated, please take a look.

python/kserve/pyproject.toml
```diff
@@ -54,38 +53,20 @@ pandas =  "^2.2.0"
 pydantic = [{ version = "^2.5.0", python = "3.12" }, { version = ">1.0,<3", python = ">=3.9,<3.12" }]
 pyyaml = "^6.0.0"
 
-# Storage dependencies. They can be opted into by apps.
```
israel-hdez, line 57: Although this is successfully fixing all code within this repository, IMO this is still a breaking change impacting users.

All the code under `python/kserve` is called the _KServe Python SDK_ (see the [README.md](https://github.com/kserve/kserve/blob/master/python/kserve/README.md) inside that directory). It is being published to PyPi here: https://pypi.org/project/kserve/. Notice that in the README.md, we claim that [the SDK supports some storage providers](https://github.com/kserve/kserve/blob/96f43de5ca4d91ceb4bd9a5b9295ad82a40e206f/python/kserve/README.md?plain=1#L46-L66).

By extracting the code under `python/kserve/kserve/storage` directory and moving it to its own package under `python/storage`, the moved code is no longer available in the KServe Python SDK. That would lead to a broken [KServe Python Server](https://github.com/kserve/kserve/blob/96f43de5ca4d91ceb4bd9a5b9295ad82a40e206f/python/kserve/README.md#kserve-python-server) and there won't be an easy workaround to fix it.

Moving around the code is not that much of a problem on itself. The issue is that users that would like to use the KServe SDK won't be able to have a working setup, because there will be no way to install the `storage` package without cloning the source code and manually installing the missing python package (because this is not publishing it to PyPi). Thus, the SDK in PyPi becomes less useful to users that need the KServe Python Server (instead of installing one part from PyPi and the missing part from sources, if becomes easier to get everything from sources).

Thus, IMO, the work here for extracting the storage part is fine. What I see missing is that
* The new storage package needs to be published to PyPi (maybe it needs to be renamed to `kserve-storage` because there is [already a package named `storage`](https://pypi.org/project/storage/) in PyPi)
* Once that package is published to PyPi, it can be a mandatory dependency here in this file.
* The storage-initializer container can be built by using only the `kserve-storage` package (as already done in this PR).

With those missing steps, my expectation is that running `pip install kserve` (or `poetry install`) should lead to a working setup without any additional commands to run. So, several changes of this PR to `Dockerfiles` and `.toml` files could be reverted.

sivanantha321: Agreed. I thought the plan was to publish the storage as a package to PyPi.

spolti: Many thanks for your review @israel-hdez.

> By extracting the code under the python/kserve/kserve/storage directory and moving it to its package under python/storage, the moved code is no longer available in the KServe Python SDK.

The storage module dependencies were always optional, not installed by default, you need to provide `--extras storage` to get it installed otherwise the user won`t be able to use it anyways.  The user would need to install the dependencies manually as well. With that said, once we have this work finished and the storage module published, it will not be a problem as we have the old storage module importing the new storage module to keep backward compatibility, so users using the old SDK will not need to change their imports if they want to update to the new version.

> That would lead to a broken [KServe Python Server](https://github.com/kserve/kserve/blob/96f43de5ca4d91ceb4bd9a5b9295ad82a40e206f/python/kserve/README.md#kserve-python-server) and there won't be an easy workaround to fix it.

See this line where the storage dependencies are installed separately:
https://github.com/kserve/kserve/blob/96f43de5ca4d91ceb4bd9a5b9295ad82a40e206f/python/kserve/README.md#poetry

> KServe SDK won't be able to have a working setup, because there will be no way to install the storage package without cloning the source code and manually installing the missing Python package

The plan is to publish it, so it should not be a breaking change, there were no decisions on how we would move it forward and publish the storage package. However, one thing that we could probably do differently, for now, is to rename the package, and once merged update the docs and the KServe dependency.


> The new storage package needs to be published to PyPi (maybe it needs to be renamed to kserve-storage because there is [already a package named storage](https://pypi.org/project/storage/) in PyPi)

Definitely, we should do it.

>  So, several changes of this PR to Dockerfiles and .toml files could be reverted.
While I don't have a strong opinion about it, our Dockerfiles are built using the local modules, with that said, installing them passing `--extra kserve-storage` or installing them as done in this PR, would be basically the same thing.


So, for now, can we do the following:

1. Rename the storage module to `kserve-storage`
2. Update the code and imports to reflect this change
3. Once we get it merged:
   - Update the Docs (https://github.com/kserve/kserve/blob/master/python/kserve/README.md)
   - Publish the kserve-storage
   - Update KServe dependencies to still include the storage as optional.


@terrytangyuan @sivanantha321 @israel-hdez what do you think?

spolti: > Agreed. I thought the plan was to publish the storage as a package to PyPi.

It is the plan, although I was waiting for this PR to get reviewed so we could discuss the next steps, this was one of my first questions on this PR :)

python/storage/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+version = "0.14.0"
+description = "A storage initializer for KServe"
```
Jooho, line 4: I am confused of this description because it is the same as this(https://github.com/kserve/kserve/pull/3902/files#diff-932ce623b0d496fa19859e372bc304d7f01400e0fc1e575c36c2959d700e03acR4)
it would be better to give more detail description to distinguish with the others

python/storage/storage/storage_logging.py
```diff
@@ -0,0 +1,109 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import logging.config
+from typing import Optional, Union, Dict
+
+import yaml
+
+STORAGE_INITIALIZER_LOGLEVEL = "DEBUG"
```
Jooho, line 21: by default, we need to use DEBUG log level?

python/storage/test/test_storage.py
```diff
@@ -22,9 +22,9 @@
 
 import pytest
 
-from kserve.storage import Storage
+from storage import Storage
 
-STORAGE_MODULE = "kserve.storage.storage"
+STORAGE_MODULE = "storage.storage"
```
Jooho, line 27: I think some comments already proposed this but I just added here again. 
`storage.storage` does not clear to me so why don't we use something like `kserve-storage.storage`

Jooho (commented):
There's nothing particularly serious here, I just added a few comments on minor details.

python/kserve/pyproject.toml
israel-hdez: Perhaps, then, may it be better to make sure that the new package is pushed to pypi, first, before merging the removal?

We just got a v0.15.0-rc0. I think soon there will be a final v0.15.0 release. I think it is better to keep the SDK stable, than to potentially delay v0.15 (or to release it broken).

> The storage module dependencies were always optional, not installed by default, you need to provide --extras storage to get it installed otherwise the user won`t be able to use it anyways. 

I understand this part. My point is that users simply need to run a new pip/poetry command to install the optional deps to get a working setup. You would not require to download source code.

With this PR, you can still install the optional dependencies by enumerating them with pip/poetry. But you won't have a working setup unless you download the sources to get the `storage` package.

yuzisun:
@spolti Can you help resolve the conflicts? thanks

sivanantha321:
- completed  in #4391

PR #3903, opened 2024-08-29
For vllm users, our parser should be able to support both - and _

hustxiayang:
The parser of kserve can only accept "-", not "_". Thus, when we deploy an inference service using vllm as backend,  it is not consistent. For example, many optimizations (for example, speculative decoding) require "use_v2_block_manager", if you set all other parameters properly but using "--use_v2_block_manager", all other setups are not used.

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


python/huggingface_server.Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG POETRY_HOME=/opt/poetry
 ARG POETRY_VERSION=1.8.3
 
 # Install vllm
-ARG VLLM_VERSION=0.5.5
+ARG VLLM_VERSION=0.6.0
```
terrytangyuan: Has this been tested? Do we want to bump version in a separately PR?

terrytangyuan (commented):
Just a comment on the version bump. Otherwise LGTM.

python/huggingface_server.Dockerfile
hustxiayang: No, yes, this was an error, and I should do it separately

PR #3907, opened 2024-08-29
update fluid kserve sample to use huggingface servingruntime

lizzzcai:
**What this PR does / why we need it**:

Update sample to use huggingface servingruntime and GPU model.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Type of changes**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

Issue #3791, opened 2024-07-14
When reasoning with huggingface server, NCCL error occurs when GPU>1


MichaelXcc:
/kind bug

**What steps did you take and what happened:**
I try to run the https://huggingface.co/Qwen/Qwen2-7B-Instruct model inference and using multiple gpu，I downloaded the model locally and mounted it into pod, but miraculously my service worked fine when I had nvidia.com/gpu = 1, and failed and reported errors when I had nvidia.com/gpu > 1.

Here is the full log of my container starting to stopping

```
WARNING 07-14 08:10:30 config.py:1155] Casting torch.bfloat16 to torch.float16.
2024-07-14 08:10:33,156	WARNING services.py:1996 -- WARNING: The object store is using /tmp instead of /dev/shm because /dev/shm has only 67108864 bytes available. This will harm performance! You may be able to free up space by deleting files in /dev/shm. If you are inside a Docker container, you can increase /dev/shm size by passing '--shm-size=10.24gb' to 'docker run' (or add it to the run_options list in a Ray cluster config). Make sure to set this to more than 30% of available RAM.
2024-07-14 08:10:34,257	INFO worker.py:1743 -- Started a local Ray instance. View the dashboard at 127.0.0.1:8265
INFO 07-14 08:10:35 llm_engine.py:161] Initializing an LLM engine (v0.4.3) with config: model='/mnt/models/Qwen2-7B-Instruct', speculative_config=None, tokenizer='/mnt/models/Qwen2-7B-Instruct', skip_tokenizer_init=False, tokenizer_mode=auto, revision=None, rope_scaling=None, tokenizer_revision=None, trust_remote_code=False, dtype=torch.float16, max_seq_len=32768, download_dir=None, load_format=LoadFormat.AUTO, tensor_parallel_size=2, disable_custom_all_reduce=False, quantization=None, enforce_eager=False, kv_cache_dtype=auto, quantization_param_path=None, device_config=cuda, decoding_config=DecodingConfig(guided_decoding_backend='outlines'), seed=0, served_model_name=/mnt/models/Qwen2-7B-Instruct)
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
hugging-model-1-predictor-00001-deployment-5c86c57756-p7ghj:1:1 [0] NCCL INFO Bootstrap : Using eth0:10.233.70.13<0>
[...]

PR #3910, opened 2024-09-01
mount /dev/shm volume to huggingfaceserver

lizzzcai:
Address the comment [here](https://github.com/kserve/kserve/pull/3907#discussion_r1739963803). add [shm](https://github.com/huggingface/text-generation-inference?tab=readme-ov-file#a-note-on-shared-memory-shm) to huggingface server which is used for multi-gpu communication.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/charts/kserve-resources/templates/clusterservingruntimes.yaml
+++ b/charts/kserve-resources/templates/clusterservingruntimes.yaml
@@ -461,14 +461,20 @@ spec:
         limits:
           cpu: "1"
           memory: 2Gi
+      {{- if .Values.kserve.servingruntime.huggingfaceserver.devShm.enabled }}
       volumeMounts:
-      - name: devshm
-        mountPath: /dev/shm
+        - name: devshm
+          mountPath: /dev/shm
+      {{- end }}
+  {{- if .Values.kserve.servingruntime.huggingfaceserver.devShm.enabled }}
   volumes:
-  - name: devshm
-    emptyDir:
-      medium: Memory
-      sizeLimit: {{ .Values.kserve.servingruntime.huggingfaceserver.devShmSizeLimit }}
+    - name: devshm
+      emptyDir:
+        medium: Memory
+        {{- with .Values.kserve.servingruntime.huggingfaceserver.devShm.sizeLimit }}
+        sizeLimit: {{ . }}
+        {{- end }}
+  {{- end }}
   {{- with .Values.kserve.servingruntime.huggingfaceserver.imagePullSecrets }}
   imagePullSecrets:
     {{- toYaml . | nindent 4 }}
--- a/charts/kserve-resources/templates/clusterservingruntimes.yaml
+++ b/charts/kserve-resources/templates/clusterservingruntimes.yaml
@@ -461,6 +461,14 @@ spec:
         limits:
           cpu: "1"
           memory: 2Gi
+      volumeMounts:
+      - name: devshm
+        mountPath: /dev/shm
+  volumes:
+  - name: devshm
+    emptyDir:
+      medium: Memory
+      sizeLimit: {{ .Values.kserve.servingruntime.huggingfaceserver.devShmSizeLimit }}
   {{- with .Values.kserve.servingruntime.huggingfaceserver.imagePullSecrets }}
   imagePullSecrets:
     {{- toYaml . | nindent 4 }}
--- a/charts/kserve-resources/README.md
+++ b/charts/kserve-resources/README.md
@@ -80,6 +80,7 @@ $ helm install kserve oci://ghcr.io/kserve/charts/kserve --version v0.14.0-rc0
 | kserve.servingruntime.art.defaultVersion | string | `"v0.14.0-rc0"` |  |
 | kserve.servingruntime.art.image | string | `"kserve/art-explainer"` |  |
 | kserve.servingruntime.art.imagePullSecrets | list | `[]` |  |
+| kserve.servingruntime.huggingfaceserver.devShmSizeLimit | string | `"1Gi"` |  |
 | kserve.servingruntime.huggingfaceserver.image | string | `"kserve/huggingfaceserver"` |  |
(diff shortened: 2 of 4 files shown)


charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -421,6 +421,14 @@ spec:
         limits:
           cpu: "1"
           memory: 2Gi
+      volumeMounts:
```
spolti, line 465: shouldn't be a condition here to verify if the target serving runtime is hugging face?

lizzzcai: Hi @spolti , this change only applies to the [kserve-huggingfaceserver](https://github.com/kserve/kserve/blob/8272bde3c1a22e8601e77e9555d4e27a848a773d/charts/kserve-resources/templates/clusterservingruntimes.yaml#L399C9-L399C33) clusterservingruntime, do think it is enough?

spolti: oh, I didn't expand the file, yes, it is enough :D

spolti (approved):

terrytangyuan (commented):
Requested by @yuzisun so please take a look

/assign @yuzisun

Datta0:
@lizzzcai what we observed is that the shm requirement changes with the number of GPUs you use in that deployment. 
From whatever we tried, 1GiB is not always enough and is sometimes unused.

Also, do we need this for CPU only deployments? I don't think so. In such case, aren't we unnecessarily taking up >1GiB of  memory?

Any info about how this would effect multi `vGPU` setups?

Also curious if you plan to do the same change for [`HuggingfaceRuntimeSpec`](https://github.com/kserve/kserve/blob/67373adee41fa97c07f2b42246dcce35210a35a0/pkg/apis/serving/v1beta1/predictor_huggingfaceruntime.go#L27)

lizzzcai:
Hi @Datta0,

> From whatever we tried, 1GiB is not always enough and is sometimes unused.

The value is configurable, but it is a bit tricky as the runtime is shared across difference inference services.

> Also, do we need this for CPU only deployments? I don't think so. In such case, aren't we unnecessarily taking up >1GiB of memory?

My understanding is that the emptydir here share the memory with the container, `SizeLimit` is to limit the size of this [emptydir](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#memory-backed-emptydir), if it is not being used, it will not occupy the memory.  So it is probably ok to create an emptydir even it is not being used. Correct me if I am wrong here. 

> Any info about how this would effect multi vGPU setups?

this one I am not sure, so far I haven't play around with vGPU.

> Also curious if you plan to do the same change for [HuggingfaceRuntimeSpec](https://github.com/kserve/kserve/blob/67373adee41fa97c07f2b42246dcce35210a35a0/pkg/apis/serving/v1beta1/predictor_huggingfaceruntime.go#L27)

I think about it initially but I just pick the most easy way. If we want to make it only work with GPU instances, then we need the controller to handle it (similar logic was done to add the GPU taint already). Tricky part is probably is the value for the SizeLimit.

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -421,6 +421,14 @@ spec:
...
+      - name: devshm
+        mountPath: /dev/shm
+  volumes:
```
yuzisun, line 470: We can add a flag to enable injecting the devshm volume and volume mount

yuzisun:
> Hi @Datta0,
> 
> > From whatever we tried, 1GiB is not always enough and is sometimes unused.
> 
> The value is configurable, but it is a bit tricky as the runtime is shared across difference inference services.
> 
> > Also, do we need this for CPU only deployments? I don't think so. In such case, aren't we unnecessarily taking up >1GiB of memory?
> 
> My understanding is that the emptydir here share the memory with the container, `SizeLimit` is to limit the size of this [emptydir](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#memory-backed-emptydir), if it is not being used, it will not occupy the memory. So it is probably ok to create an emptydir even it is not being used. Correct me if I am wrong here.
> 
> > Any info about how this would effect multi vGPU setups?
> 
> this one I am not sure, so far I haven't play around with vGPU.
> 
> > Also curious if you plan to do the same change for [HuggingfaceRuntimeSpec](https://github.com/kserve/kserve/blob/67373adee41fa97c07f2b42246dcce35210a35a0/pkg/apis/serving/v1beta1/predictor_huggingfaceruntime.go#L27)
> 
> I think about it initially but I just pick the most easy way. If we want to make it only work with GPU instances, then we need the controller to handle it (similar logic was done to add the GPU taint already). Tricky part is probably is the value for the SizeLimit.

User can still override the SizeLimit from isvc as kserve controller merges the container specs from isvc and serving runtime. The value specified on isvc spec takes precedence.

yuzisun:
We also need the knative fix for enabling hostIPC
https://github.com/kserve/kserve/issues/3791

charts/kserve-resources/templates/clusterservingruntimes.yaml
lizzzcai: added a flag to enable the devsh mount. however I set the default value false and set the default sizeLimit to empty as well. As I found the value can be very high from the Knative IPC issue. If user want to enable with default value, then they can enable in the helmchart. Otherwise they can mount directly in the isvc as well.

yuzisun: Tested with 10GB which works for llama3 70B model using tensor parallelism 8

lizzzcai: Hi @yuzisun, do you expect a default sizeLimit being set here? and what value you proposed.

yuzisun (approved):

PR #3932, opened 2024-09-09
Modelcar race condition mitigation with an init container

### LINKED ISSUE #3927: Modelcar: Add no-op init-container to pre-pull model image
/kind feature

When a model runtime starts when using modelcars for initializing the model, there can be a race condition when the OCI image with the model needs to be pulled, so that the runtime cannot find it when it starts. 

https://github.com/kserve/kserve/issues/3646 solves this, but this works only if K8s sidecars are possible (k8s > 1.29 & feature enabled).

An alternative solution is to add just an init-container whose `command:` does nothing, but forces the model car image to be pulled already in the init phase. 
In this situation, the likelihood that the runtime container starts before modelcar is minimized, since the image is already there.
And even when the runtime might start faster, this will result possibly only in one restart of the runtime during the time the modelcar image is pulled (which can be many minutes).
 
Ideally this feature is included in the storage-initializer's support for modelcars.
Just adapt the modified `spec:` to add an additional init-container that just references the `storageUri` OCI image and does nothing in its `command:` (like something like `sh -c true` or similar).

israel-hdez:

This adds the model container as an init-container to mitigate a race condition that would happen if the model container is not present on the cluster-node. The race condition happens if the cluster is able to fetch and start the runtime container before the modelcar is fetched. This would lead to the runtime to terminate with error.

By configuring the model container as an init-container the runtime won't start until the modelcar is fetched. Although there is still the risk of a race condition when the cluster schedules the runtime container first, the pod should stabilize after a few restarts of the runtime container and should either prevent a CrashLoopBackOff event on the pod, or the crash event would finish quickly.

This improves compatibility with the runtimes which can now stay agnostic to the modelcar implementation, until better techniques (like native sidecars, and oci volume mounts) become mature.

Fixes #3927 

**Type of changes**

- Smoke test that the init container is properly added when using OCI to pre-fetch the model, and that the modelcar works properly.
- Smoke test that cloud storage is not affected.

**Checklist**:

**Release note**:
```release-note
Mitigate race condition when using OCI for model storage (modelcar)
```

### DIFF
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -48,6 +48,7 @@ const (
 	PvcSourceMountPath                      = "/mnt/pvc"
 	CaBundleVolumeName                      = "cabundle-cert"
 	ModelcarContainerName                   = "modelcar"
+	ModelcarInitContainerName               = "modelcar-init"
 	ModelInitModeEnv                        = "MODEL_INIT_MODE"
 	CpuModelcarDefault                      = "10m"
 	MemoryModelcarDefault                   = "15Mi"
@@ -172,6 +173,11 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *v1.Pod) error {
 	if getContainerWithName(pod, ModelcarContainerName) == nil {
 		modelContainer := mi.createModelContainer(image, constants.DefaultModelLocalMountPath)
 		pod.Spec.Containers = append(pod.Spec.Containers, *modelContainer)
+
+		// Add the model container as an init-container to pre-fetch the model before
+		// the runtimes starts.
+		modelInitContainer := mi.createModelInitContainer(image)
+		pod.Spec.InitContainers = append(pod.Spec.InitContainers, *modelInitContainer)
 	}
 
 	// Enable process namespace sharing so that the modelcar's root filesystem
@@ -678,6 +684,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
 		modelContainer.SecurityContext = &v1.SecurityContext{
 			RunAsUser: mi.config.UidModelcar,
 		}
+	}
+
+	return modelContainer
+}
+
+func (mi *StorageInitializerInjector) createModelInitContainer(image string) *v1.Container {
+	cpu := mi.config.CpuModelcar
+	if cpu == "" {
+		cpu = CpuModelcarDefault
+	}
+	memory := mi.config.MemoryModelcar
+	if memory == "" {
+		memory = MemoryModelcarDefault
+	}
+
+	modelContainer := &v1.Container{
+		Name:  ModelcarInitContainerName,
+		Image: image,
+		Args: []string{
+			"sh",
+			"-c",
+			// Check that the expected models directory exists
+			"echo 'Pre-fetching modelcar " + image + ": ' && [ -d /models ] && [ \"$$(ls -A /models)\" ] && echo 'OK ... Prefetched and valid (/models exists)' || (echo 'NOK ... Prefetched but modelcar is invalid (/models does not exist or is empty)' && exit 1)",
+		},
+		Resources: v1.ResourceRequirements{
+			Limits: map[v1.ResourceName]resource.Quantity{
+				// Could possibly be reduced to even less
+				v1.ResourceCPU:    resource.MustParse(cpu),
+				v1.ResourceMemory: resource.MustParse(memory),
+			},
+			Requests: map[v1.ResourceName]resource.Quantity{
+				v1.ResourceCPU:    resource.MustParse(cpu),
+				v1.ResourceMemory: resource.MustParse(memory),
+			},
+		},
+		TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
 	}
 
 	return modelContainer
--- a/pkg/webhook/admission/pod/storage_initializer_injector_test.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector_test.go
@@ -2723,6 +2723,33 @@ func TestInjectModelcar(t *testing.T) {
 			t.Errorf("Expected two containers but got %d", len(pod.Spec.Containers))
 		}
 
+		// Check that an init container has been injected, and it is the model container
+		if len(pod.Spec.InitContainers) != 1 {
+			t.Errorf("Expected one init container but got %d", len(pod.Spec.InitContainers))
+		} else if pod.Spec.InitContainers[0].Name != ModelcarInitContainerName {
+			t.Errorf("Expected the init container to be the model but got %s", pod.Spec.InitContainers[0].Name)
+		} else {
+			// Check that resources are correctly set.
+			if _, ok := pod.Spec.InitContainers[0].Resources.Limits[v1.ResourceCPU]; !ok {
+				t.Error("The model container does not have CPU limit set")
+			}
+			if _, ok := pod.Spec.InitContainers[0].Resources.Limits[v1.ResourceMemory]; !ok {
+				t.Error("The model container does not have Memory limit set")
+			}
+			if _, ok := pod.Spec.InitContainers[0].Resources.Requests[v1.ResourceCPU]; !ok {
+				t.Error("The model container does not have CPU request set")
+			}
+			if _, ok := pod.Spec.InitContainers[0].Resources.Requests[v1.ResourceMemory]; !ok {
+				t.Error("The model container does not have Memory request set")
+			}
+
+			// Check args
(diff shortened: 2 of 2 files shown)


israel-hdez:
@rhuss Here is the init container to pre-fetch the OCI mode image. I will appreciate your review.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -683,6 +689,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
...
 
+func (mi *StorageInitializerInjector) createModelInitContainer(image string) *v1.Container {
+	cpu := mi.config.CpuModelcar
+	if cpu == "" {
+		cpu = CpuModelcarDefault
+	}
+	memory := mi.config.MemoryModelcar
+	if memory == "" {
+		memory = MemoryModelcarDefault
+	}
```
rhuss, line 700: I'm unsure if you need to set those limits to the same value as the real model car. But it's ok as both the command for checking the conformance of the modelcar image and the actual command executed when running as a model car should be similar and minimal. It should also not affect the scheduling of the Pod, as for the scheduling decision, the max(resources of init-containers) is used, while for sidecars they are added up (as they run in parallel).

However, I would probably set both `requests` and `limits` to the same values to avoid the Pod not being put into the QoS class of "Guaranteed" (which is the highest possible class). Let's do this, too, for the modelcar when running as a sidecar (if not already)

See https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/ for more details.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -683,6 +689,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
...
+			"sh",
+			"-c",
+			// Check that the expected models directory exists
+			"[ -d /models ] && [ \"$$(ls -A /models)\" ] && echo 'Found /models' || (echo 'Path /models does not exist or is empty' && exit 1)",
+		},
+		Resources: v1.ResourceRequirements{
+			Limits: map[v1.ResourceName]resource.Quantity{
+				// Could possibly be reduced to even less
+				v1.ResourceCPU:    resource.MustParse(cpu),
+				v1.ResourceMemory: resource.MustParse(memory),
```
rhuss, line 715: I would add `limits` here as well

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -683,6 +689,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
...
+			},
+			Requests: map[v1.ResourceName]resource.Quantity{
+				v1.ResourceCPU:    resource.MustParse(cpu),
+				v1.ResourceMemory: resource.MustParse(memory),
```
rhuss, line 719: I would limits here as well

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -683,6 +689,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
...
+			},
+		},
+		TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
```
rhuss, line 722: 👍

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -683,6 +689,42 @@ func (mi *StorageInitializerInjector) createModelContainer(image string, modelPa
...
+	}
+
+	modelContainer := &v1.Container{
+		Name:  ModelcarInitContainerName,
+		Image: image,
+		Args: []string{
+			"sh",
+			"-c",
+			// Check that the expected models directory exists
+			"[ -d /models ] && [ \"$$(ls -A /models)\" ] && echo 'Found /models' || (echo 'Path /models does not exist or is empty' && exit 1)",
```
rhuss: Nice! This also helps validate a model car image early on. I'm not fully sure what happens if the init-container exits with 1, and if the the full Pod is restarted or just marked with an error. I would avoid that as it does not make sense to restart with the same image as the error won't go away.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -2723,6 +2723,13 @@ func TestInjectModelcar(t *testing.T) {
 			t.Errorf("Expected two containers but got %d", len(pod.Spec.Containers))
 		}
 
+		// Check that an init container has been injected, and it is the model container
+		if len(pod.Spec.InitContainers) != 1 {
+			t.Errorf("Expected one init container but got %d", len(pod.Spec.InitContainers))
+		} else if pod.Spec.InitContainers[0].Name != ModelcarInitContainerName {
+			t.Errorf("Expected the init container to be the model but got %s", pod.Spec.InitContainers[0].Name)
+		}
+
```
rhuss, line 2752: You might want to add some more checks on the structure of the init-container, e.g. that both requests & limits are set, and that the commands contains some key aspects (I rarely check on literal strings for equality, but e.g. that the messaging contains important keywords that might help the user to troubleshoot, too.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: Can you also add some more context information to the error message? For example, the image name should be added here, too. That makes it easier to troubleshoot.

rhuss (commented):
Looks good to me, thanks! 

Some minor comments for improvements added, but the overall structure looks great.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: You could also add the purpose of the container to the success message, so that people understand why this dummy container is started overall.

Something like:
```suggestion
			"echo 'Pre-fetching modelcar " + image + ": ' && [ -d /models ] && [ \"$$(ls -A /models)\" ] && echo 'OK ... Prefetched and valid (/models exists)' || (echo 'NOK ... Prefetched but modelcar is invalid (/models does not exist or is empty)' && exit 1)",
```

pkg/webhook/admission/pod/storage_initializer_injector.go
israel-hdez: I think somewhat confused. [Some lines later](https://github.com/kserve/kserve/pull/3932/files#diff-19ab20f87e5f9bff37c2eada3e341c832cf91029b78a75b9f70acf6a22dacb8cR711), resources and limits are set. I see you have some comments around those lines :-) 

From [this docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container), I'm not seeing any additional attributes to set, unless I'm not seeing well. So, I think I need you to elaborate a little bit more.

pkg/webhook/admission/pod/storage_initializer_injector.go
israel-hdez: I simply used your suggestion :-)

pkg/webhook/admission/pod/storage_initializer_injector_test.go
israel-hdez: OK. I added the additional checks.

pkg/webhook/admission/pod/storage_initializer_injector.go
israel-hdez: > I'm not fully sure what happens if the init-container exits with 1, and if the the full Pod is restarted or just marked with an error.

In my tests, the cluster restarts the init container, and the pod persists. It will keep restarting the init contaienr, and will mark the pod with `Init:Error` status. After some additional restarts, the status is downgraded to a CrashLoopBackOff.

israel-hdez:
@rhuss OK, I fixed what I could. Please, take another look.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: Sorry, my bad :) Forget my comments, I just was confused.

pkg/webhook/admission/pod/storage_initializer_injector.go
rhuss: Do you think it would be possible that if the init-container fails because e.g. the format is wrong, that we could bring the whole Pod in an error status ? That would be a nice to have, though. I think we can live with the current solution.

rhuss:
Thanks! lgtm from my side

/approve
/lgtm

israel-hdez:
@yuzisun This is ready for a final review/approval.

israel-hdez:
@sivanantha321 I wonder if you would help me giving this PR a final review?
It's been sitting here for some time and I'd like it to catch the 0.14 release.

terrytangyuan (approved):

PR #3933, opened 2024-09-10
For vllm users, our parser should be able to support both - and _

hustxiayang:
In huggingfaceserver, we do not support "_" but vLLM does support it. This is inconsistent and may cause performance effect: for example, if we have a typo "--use_v2_block_manager" then, this is unset and many other optimization parameters would be unset, which would cause unexpected formance.

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/__main__.py
+++ b/python/huggingfaceserver/huggingfaceserver/__main__.py
@@ -50,7 +50,12 @@ def list_of_strings(arg):
     return arg.split(",")
 
 
-parser = argparse.ArgumentParser(parents=[kserve.model_server.parser])
+try:
+    from vllm.utils import FlexibleArgumentParser
+
+    parser = FlexibleArgumentParser(parents=[kserve.model_server.parser])
+except ImportError:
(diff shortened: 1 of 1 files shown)


hustxiayang:
@terrytangyuan  Hi, could you please review this small change? Thanks a lot!

yuzisun:
> @terrytangyuan Hi, could you please review this small change? Thanks a lot!

Could you help add a test in the e2e directory?

terrytangyuan (commented):
Could you add tests as suggested above?

terrytangyuan:
> vLLM does support it

Can you provide a reference to this? What was the motivation for vLLM to support both?

yuzisun:
> > vLLM does support it
> 
> Can you provide a reference to this? What was the motivation for vLLM to support both?

 kserve uses snake case while vllm backend uses dashes, so it is not consistent if user specifies both kserve and vllm arguments. vllm accepts both conventions so we should do the same in kserve. 
 
see https://github.com/vllm-project/vllm/pull/5718

terrytangyuan:
Thanks for the reference!

yuzisun (approved):

PR #3953, opened 2024-09-23
Add combined logging feature

### LINKED ISSUE #3938: Combined Request/Response Logging
/kind feature
# **Describe the solution you'd like**
Add a new logging configuration which emits a single CloudEvent with a combined request / response. The current implementation emits separate events for the request and response, this adds complexity for downstream consumers which need to use the combined request / response as a single entity. 

The [LogRequest](https://github.com/kserve/kserve/blob/9ec6842546682b338f0141bba2bda769a953a850/pkg/logger/types.go#L23-L35) struct fields are mostly the same between the request and response types. The only fields that differ are:
* ContentType
* Bytes
* Metadata (not present on the response)
* ReqType

## Combining requests and response
The primary issue in combining the request / response is how to handle combining their data. CloudEvent uses the `ContentType` to encode different objects into the event body, since a combined Request/Response would have a single body the logger would have to make an opinionated guess as to how to combined the request and response. Due to the number of content-types that kserve supports as well as user edgecases we need to determine the best approach. I've found two potential solutions.

### Option 1: Only emit combined logs if both are JSON
The primary issues with combining logs comes from cases where the request and response have mismatched content-types or for complicated binary encodings like grpc. When both messages are JSON content type we can easily merge them into a single message:
```json
{
    "request_body": {...},
    "response_body": {...}
}
```
This option has the easiest implementation, though we will need handle cases where the schema is specified in the content-type ex. `"application/schema+json"`. 

### Option 2: Leverage cloud event batching
With cloud event batching we could keep the current `LogRequests` and send both as a list. From the [cloud event spec](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#event-format)
>A "batch-mode message" is one where multiple (zero or more) events are encoded in a single message body, according to a specific event format. Not all event formats or protocol bindings support batch-mode messages. The CloudEvents within a batch are largely independent from one another: there is no restriction that they have the same source, producer, content type etc. The only restriction is that all CloudEvents within the same batch MUST have the same value for the specversion attribute.

Batch events are supported via the cloud event go-sdk, though only. through low level primitives for sending and receiving. There isn't a way to send multiple events from the `Client` object currently used. You can see the PR where they were added [here](https://github.com/cloudevents/sdk-go/pull/829/files), there's a lot of good context there. 

It would be possible to use the `NewHttpRequestFromEvents` to support sending multiple and `NewHttpRequestFromEvent` to satisfy the existing usecases. In addition to using these new function the `WorkQueue` would need to be changed to `chan []LogRequest` from `chan LogRequest`. 

## Usage and configuration

We could add a new ReqType option for this to differentiate it from the other CloudEvent types:
* `org.kubeflow.serving.inference.requestresponse` or `org.kubeflow.serving.inference.combined`

Users could opt in to this by specifying a new LoggingType `combined` in their kserve config:
```yaml
kind: InferenceService
metadata:
  name: sklearn-iris
spec:
  predictor:
    logger:
      mode: combined
      url: http://message-dumper.default/
    model:
      modelFormat:
        name: sklearn
```
# **Anything else you would like to add:**

Looking at the handler, I don't believe this should cause additional memory overhead, it can also be implemented easily within the current handler layout without requiring large modifications. 

I'm going to begin a rough draft implementation of what I outline above, I'll make sure to keep in line with where discussion goes on this post. 

# **Links to the design documents:**

benkawecki:
This PR attempts to address #3938. This PR adds a combined logging mode which emits a batch CloudEvent containing the log for the InferenceRequest and InferenceResponse. 

Summary of changes:
* Add a new configuration to the logger field 
* Refactors the worker to support sending a single cloud event or batch cloud events
  * Changes the worker queue to use a slice of LogRequest instead of a single LogRequest
  * Use existing functionality when a single LogRequest is sent of the channel
  * Call a `sendBatchCloudEvent` method when two LogRequests are sent over the channel
  * Refactor various parts of the `sendCloudEvent` method into other methods to share functionality between `sendCloudEvent` and `sendBatchCloudEvent`

Fixes #3938 

**Type of changes**

- [ ] Modified existing unittest to ensure the timefield on cloud events.
- [ ] Added unittest to ensure the batch CloudEvent is sent and is parsable with `NewEventsFromHTTPRequest`.
- [ ] Added an e2e test to ensure batch events are emitted properly.

**Checklist**:

**Release note**:

```release-note
Add a new "combined" log mode that emits a batch cloud event consisting of the inference request and response.
```

**Re-running failed tests**

### DIFF
--- a/pkg/logger/worker.go
+++ b/pkg/logger/worker.go
@@ -122,7 +122,9 @@ func (w *Worker) prepareEvent(logReq LogRequest) (cloudevents.Event, error) {
 
 // buildClient constructs an http client and transport to send cloudevent request.
 func (w *Worker) buildClient(url *url.URL, certName string, tlsSkipVerify bool) (*http.Client, error) {
-	c := &http.Client{}
+	c := &http.Client{
+		Timeout: time.Second * 10,
+	}
 
 	if url.Scheme == "https" {
 		caCertFilePath := filepath.Join(LoggerCaCertMountPath, certName)
@@ -220,9 +222,12 @@ func (w *Worker) sendCloudEvent(logReq LogRequest) error {
 		if err != nil {
 			w.Log.Warnf("unable to read body of response: %w", err)
 		}
-		err = fmt.Errorf(string(body))
+
+		w.Log.Errorf("Sent with status code %d, error: %v", res.StatusCode, fmt.Errorf(string(body)))
+	} else {
+		w.Log.Infof("Sent with status code %d", res.StatusCode)
 	}
-	w.Log.Infof("Sent with status code %d, error: %v", res.StatusCode, err)
+
 	return nil
 }
 
@@ -253,7 +258,7 @@ func (w *Worker) Start() {
 					}
 
 				} else {
-					w.Log.Errorf("Support for more than 2 work items not permitted")
+					w.Log.Errorf("Invalid amount of work items, number of work items must be 1 or 2, not %d", len(work))
 				}
 
 			case <-w.QuitChan:
--- a/pkg/logger/worker.go
+++ b/pkg/logger/worker.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -90,7 +91,7 @@ type Worker struct {
 	CeCtx       context.Context
 }
 
-// prepareEvent creates a cloudevent from the LogRequest setting the extentions
+// prepareEvent creates a cloudevent from the LogRequest setting the extensions
 // body and metadata. Event defaulter functions should be applied here since
 // a cloudevent client is not used.
 func (w *Worker) prepareEvent(logReq LogRequest) (cloudevents.Event, error) {
@@ -184,11 +185,12 @@ func (w *Worker) sendBatchCloudEvent(logReqs []LogRequest) error {
 		if err != nil {
 			w.Log.Warnf("unable to read body of response: %w", err)
 		}
-		err = fmt.Errorf(string(body))
-	}
-	w.Log.Infof("Sent with status code %d, error: %v", res.StatusCode, err)
-	return nil
 
+		w.Log.Errorf("Sent with status code %d, error: %v", res.StatusCode, errors.New(string(body)))
+	} else {
+		w.Log.Infof("Sent with status code %d", res.StatusCode)
+	}
+	return nil
 }
 
 func (w *Worker) sendCloudEvent(logReq LogRequest) error {
@@ -222,8 +224,7 @@ func (w *Worker) sendCloudEvent(logReq LogRequest) error {
 		if err != nil {
 			w.Log.Warnf("unable to read body of response: %w", err)
 		}
-
-		w.Log.Errorf("Sent with status code %d, error: %v", res.StatusCode, fmt.Errorf(string(body)))
+		w.Log.Errorf("Sent with status code %d, error: %v", res.StatusCode, errors.New(string(body)))
 	} else {
 		w.Log.Infof("Sent with status code %d", res.StatusCode)
 	}
@@ -241,7 +242,8 @@ func (w *Worker) Start() {
 
 			select {
 			case work := <-w.Work:
-				if len(work) == 1 {
+				switch len(work) {
+				case 1:
 					workItem := work[0]
 					// Receive a work request.
 					w.Log.Infof("Received work request %d, url: %s, requestId: %s", w.ID, workItem.Url.String(), workItem.Id)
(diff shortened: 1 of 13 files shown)


pkg/logger/worker.go
```diff
@@ -167,11 +236,24 @@ func (w *Worker) Start() {
...
+					// Receive a work request.
+					w.Log.Infof("Received work request %d, url: %s, requestIds: %s %s", w.ID, work[0].Url.String(), work[0].Id, work[1].Id)
+
+					if err := w.sendBatchCloudEvent(work); err != nil {
+						w.Log.Error(err, "Failed to send cloud event, url: %s", work[0].Url.String())
+					}
 
-				if err := w.sendCloudEvent(work); err != nil {
-					w.Log.Error(err, "Failed to send cloud event, url: %s", work.Url.String())
+				} else {
```
yuzisun: Is there a possible case with 0 work items ?

test/e2e/logger/test_logger.py
```diff
@@ -115,3 +115,89 @@ async def test_kserve_logger(rest_v1_client):
...
+        api_version=constants.KSERVE_V1BETA1,
+        kind=constants.KSERVE_KIND,
+        metadata=client.V1ObjectMeta(name=msg_dumper, namespace=KSERVE_TEST_NAMESPACE),
+        spec=V1beta1InferenceServiceSpec(predictor=predictor),
+    )
+
+    kserve_client.create(isvc)
+    kserve_client.wait_isvc_ready(msg_dumper, namespace=KSERVE_TEST_NAMESPACE)
+
+    service_name = "isvc-logger"
```
yuzisun: Suggest changing to a different name as tests can run concurrently and conflict with other tests

pkg/logger/worker.go
```diff
@@ -75,85 +84,145 @@ func NewWorker(id int, workerQueue chan chan LogRequest, logger *zap.SugaredLogg
...
-		} else {
-			w.Log.Infof("Send did not return an HTTP response: %s", res)
+	// Generate Http request from event
+	httpReq, err := cloudevents.NewHTTPRequestFromEvent(w.CeCtx, logReq.Url.String(), event)
+	if err != nil {
+		return fmt.Errorf("while creating http request: %w", err)
+	}
+
+	// Send event
+	res, err := c.Do(httpReq)
```
yuzisun, line 216: Do we have timeout for the http request

pkg/logger/worker.go
```diff
@@ -75,85 +84,145 @@ func NewWorker(id int, workerQueue chan chan LogRequest, logger *zap.SugaredLogg
...
 
-	encodedMetadata, err := json.Marshal(logReq.Metadata)
+	if res.StatusCode != http.StatusOK {
+		body, err := io.ReadAll(res.Body)
+		if err != nil {
+			w.Log.Warnf("unable to read body of response: %w", err)
+		}
+		err = fmt.Errorf(string(body))
+	}
+	w.Log.Infof("Sent with status code %d, error: %v", res.StatusCode, err)
```
yuzisun: Do we want to print error level for the requests that are failed  ?

pkg/logger/worker.go
```diff
@@ -75,85 +84,145 @@ func NewWorker(id int, workerQueue chan chan LogRequest, logger *zap.SugaredLogg
...
+
+	for _, lr := range logReqs {
+		event, err := w.prepareEvent(lr)
+		if err != nil {
+			return err
+		}
+		events = append(events, event)
+	}
+
+	c, err := w.buildClient(logReqs[0].Url, logReqs[0].CertName, logReqs[0].TlsSkipVerify)
```
yuzisun, line 166: These fields looks like should be common to all the log requests

pkg/logger/worker.go
benkawecki: Right now we control access to the queue through the functions `sendCloudEvent` and `sendBatchCloudEvent`, but I agree we should be defensive on this. I'll make this error log more inclusive to account for the 0 case.

pkg/logger/worker.go
benkawecki: I thought the same thing, but the previous behavior was to log all response at the info level, optionally there was an error to include it. I can split this up to log at error level when status code isn't 200 and info otherwise. I'll update this.

pkg/logger/worker.go
benkawecki: The default behavior for the client is no timeout. I just checked if we have a timeout in the current way we send cloud events and I don't believe that has a timeout either. The http implementation of the send protocol which the CloudEvent client uses doesn't seem to set a default timeout. 

I put a timeout of 10 seconds, since that seems like a sensible default. I could see a case for increasing that to 30-60 seconds to be closer to the current behavior of no timeout. @yuzisun what are you thoughts?

pkg/logger/worker.go
benkawecki: I agree these fields are shared and infact they are static through the lifetime of the logger. The current implementation creates a client and loads the certs each time it sends a log request. Not only are we sending more data over the channel, but we are not re-using the client connections. 

I debated refactoring this in this PR, but it would likely require some larger changes to the dispatcher, worker and handler. I think including it in this PR would make the change too large and complicate the logic. I think its worth while doing that refactor in a separate PR.

PR #3964, opened 2024-09-27
bump to vllm0.6.2 and add explicit chat template

hustxiayang:
Bump to vLLM0.6.2 to support multimodal models. 
vLLM0.6.2 requires newer version of transformers '>=4.45.0`, some related changes are:

1. It does not support default chat templates for models: https://github.com/huggingface/transformers/pull/31733. 
Thus,  add an optional `chat template` member in `CreateChatCompletionRequest` class, which is similar to https://github.com/vllm-project/vllm/blob/9aaf14c62e16a7c74b5192a44d01a78125dab2fc/vllm/entrypoints/openai/protocol.py#L239

2. In tokenizer, `clean_up_tokenization_spaces` would be set as false by default, check details: https://github.com/huggingface/transformers/issues/31884. Thus, change the assertions in test `test_input_padding_with_pad_token_not_specified`

Also added a simple unit test of `chat_template` for tokenizer.
 

**Type of changes**

```release-note
add explicit chat template
clean_up_tokenization_spaces were set as False by transformers
```

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/generative_model.py
+++ b/python/huggingfaceserver/huggingfaceserver/generative_model.py
@@ -384,7 +384,7 @@ class HuggingfaceGenerativeModel(
         return GenerationConfig(**kwargs)
 
     def apply_chat_template(
-        self, messages: Iterable[ChatCompletionRequestMessage]
+        self, messages: Iterable[ChatCompletionRequestMessage], chat_template: Optional[str] = None,
     ) -> ChatPrompt:
         """
         Given a list of chat completion messages, convert them to a prompt.
@@ -394,6 +394,7 @@ class HuggingfaceGenerativeModel(
                 str,
                 self._tokenizer.apply_chat_template(
                     [m.model_dump() for m in messages],
+                    chat_template=chat_template,
                     tokenize=False,
                     add_generation_prompt=True,
                 ),
--- a/python/huggingfaceserver/huggingfaceserver/generative_model.py
+++ b/python/huggingfaceserver/huggingfaceserver/generative_model.py
@@ -384,7 +384,9 @@ class HuggingfaceGenerativeModel(
         return GenerationConfig(**kwargs)
 
     def apply_chat_template(
-        self, messages: Iterable[ChatCompletionRequestMessage], chat_template: Optional[str] = None,
+        self,
+        messages: Iterable[ChatCompletionRequestMessage],
+        chat_template: Optional[str] = None,
     ) -> ChatPrompt:
         """
         Given a list of chat completion messages, convert them to a prompt.
--- a/test/e2e/data/opt_125m_input_generate.json
+++ b/test/e2e/data/opt_125m_input_generate.json
@@ -3,11 +3,11 @@
     "messages": [
       {
         "role": "system",
-        "content": "You are a friendly chatbot who always responds in the style of a pirate",
+        "content": "You are a friendly chatbot who always responds in the style of a pirate"
       },
       {
         "role": "user",
-        "content": "How many helicopters can a human eat in one sitting?",
+        "content": "How many helicopters can a human eat in one sitting?"
       }
     ],
     "temperature": 0,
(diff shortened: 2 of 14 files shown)


hustxiayang:
With newer versions of transformers, it does not support default chat templates for models: https://github.com/huggingface/transformers/pull/31733, thus I explicitly give a chat template in tests

terrytangyuan (commented):
Can you update the PR title and description to also include any features that you'd like to bring in this release?

hustxiayang:
@terrytangyuan thanks a lot for your reply. I've updated the title and description!

spolti (commented):
The changes look good to me, however, my only concern is backward compatibility.
For users migrating to the new KServe version which brings it, will user be required to do anything else or?

hustxiayang:
@spolti I thought the added parameter has a default value `None`, so it would not affect the users? If it's not set, the model's tokenizer should have some chat templates, which behave the same as `tokenizer`'s `apply_chat_template` method? but I will double check it.

johnugeorge:
We should have  0.14.0 stable release cut while we thoroughly test these new change separately.

hustxiayang:
> We should have 0.14.0 stable release cut while we thoroughly test these new change separately.

@johnugeorge Hi, thanks for your comment! Do you mean that I should expect some more PRs on `0.14.0 stable release` beyond https://github.com/kserve/kserve/pull/3970?

yuzisun:
@sivanantha321 can you help review the test changes

test/e2e/data/opt_125m_input_generate.json
```diff
@@ -11,5 +11,6 @@
       }
     ],
     "temperature": 0,
-    "max_tokens": 20
+    "max_tokens": 20,
+    "chat_template": "{% for message in messages %}{{ message.content }}{{ eos_token }}{% endfor %}"
```
yuzisun, line 15: The test should still work without adding chat_template field right?

python/huggingfaceserver/huggingfaceserver/generative_model.py
```diff
@@ -218,6 +218,7 @@ def load(self) -> bool:
             tokenizer_kwargs["trust_remote_code"] = True
 
         model_kwargs["torch_dtype"] = self.dtype
+        tokenizer_kwargs["clean_up_tokenization_spaces"] = True
```
yuzisun: Does this apply to all model types?

test/e2e/data/opt_125m_input_generate.json
hustxiayang: @yuzisun for this model, its tokenizer does not have a default chat_template. The default chat_template was removed in src/transformers/models/gpt2/tokenization_gpt2.py in https://github.com/huggingface/transformers/pull/31733. Thus, you either assign a chat_template to the tokenizer as in tests/models/gpt2/test_tokenization_gpt2.py in https://github.com/huggingface/transformers/pull/31733, or give an explicit chat_template parameter as tests/async_engine/test_chat_template.py in https://github.com/vllm-project/vllm/pull/7238 (also this pr). Otherwise it raise error : https://github.com/huggingface/transformers/blob/37ea04013b34b39c01b51aeaacd8d56f2c62a7eb/src/transformers/tokenization_utils_base.py#L1964C5-L1964C34 and this test would not pass.

python/huggingfaceserver/huggingfaceserver/generative_model.py
hustxiayang: @yuzisun Thanks a lot for the comment! I added this because test `test_input_padding_with_pad_token_not_specified` failed as `clean_up_tokenization_spaces` is false by default in transformers v4.45. I thought in most cases, we do not expect a whitespace before punctuations, so I enforce it to be true, but you are right, I think I should leave it as `False` by default to be consistent with transformers, and just update the assertions in the test?

yuzisun (approved):
/lgtm
/approve

yuzisun:
Thanks @hustxiayang ! great work on this

PR #3981, opened 2024-10-09
Fix kserve local testing env

yuzisun:
`make test` was failing locally as the crd paths are different for tests from different packages, we should pass the crd path to the `SetupEnvTest` function instead of enumerating all the possible paths there.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/openapi/swagger.json
```diff
@@ -1169,7 +1169,7 @@
       "type": "object",
       "properties": {
         "annotations": {
-          "description": "Annotations that will be add to the component pod. More info: http://kubernetes.io/docs/user-guide/annotations",
+          "description": "Annotations that will be added to the component pod. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/",
```
sivanantha321, line 1172: Wondering how was the CI didn't complain about this ? Maybe not rebased ? Should we have a check for rebase ?

yuzisun: I think the merged PR probably was not rebased and it was created before the verification was added

sivanantha321: Oh Okay. Also I am thinking we can add a auto retry mechanism for Infra setup actions.https://github.com/marketplace/actions/retry-action. WDYT ?

yuzisun: Sounds good

PR #4000, opened 2024-10-16
Feature: Add hf transfer

tjandy98:
Add support for fast download from Huggingface by enabling [hf_transfer](https://huggingface.co/docs/huggingface_hub/en/guides/download#faster-downloads)


spolti (approved):
/lgtm

PR #4003, opened 2024-10-17
Fix  InferenceService state when Predictor pod in CrashLoopBackOff 

### LINKED ISSUE #1859: kfserving inference service state shows Ready=True when pods are in CrashLoopBackoff state
/kind bug

Inference service resource was created as the deployment was created successfully with 1 pod as shown below:
k get pods -A | grep d5ce6b94087b41
`xxx-namespace-xxx     d5ce6b94087b41-predictor-default-vcwhr-deployment-xxxx   3/3     Running              0          11h`

k get deploy -A | grep d5ce6b94087b41
`xxx-namespace-xxx    d5ce6b94087b41-predictor-default-vcwhr-deployment     1/1     1            1           20d`

k get isvc -A | grep d5ce6b94087b41
`xxx-namespace-xxx          d5ce6b94087b41     http://d5ce6b94087b41-xxx-namespace-xxx-xxx.com       True           100                              d5ce6b94087b41-predictor-default-vcwhr    20d`

Due to some scenario, the deployment got restarted. But this time, as there was some problem with the model download storage-initialisation phase, pod got stuck with CrashLoopBackoff state.

Below is the current state of the resources:
k get pods -A | grep d5ce6b94087b41
`xxx-namespace-xxx        d5ce6b94087b41-predictor-default-vcwhr-deployment-oooo    0/3    Init:CrashLoopBackOff   356        40m46s`

k get deploy -A | grep d5ce6b94087b41
`xxx-namespace-xxx          d5ce6b94087b41-predictor-default-vcwhr-deployment     0/1     1            0           20d`

k get isvc -A | grep d5ce6b94087b41
`xxx-namespace-xxx          d5ce6b94087b41     http://d5ce6b94087b41-xxx-namespace-xxx-xxx.com       True           100                              d5ce6b94087b41-predictor-default-vcwhr    20d`

storage-initialiser error:
`Traceback (most recent call last):
  File "./scripts/initializer-entrypoint", line 14, in <module>
    kfserving.Storage.download(src_uri, dest_path)
  File "/usr/lib/python3.6/site-packages/kfserving/storage.py", line 62, in download
    Storage._download_s3(uri, out_dir)
  File "/usr/lib/python3.6/site-packages/kfserving/storage.py", line 101, in _download_s3
    The path or model %s does not exist." % (uri))
RuntimeError: Failed to fetch model. The path or model s3://bucket-name/test_data/gender does not exist.`

The inference service should reflect the similar status i.e, it should be Ready=False as there are no pods in Running state to serve the requests.

Could reproduce the scenario.
Step1: Create a deployment with a model available in the object store s3. Deployment/Pod/InferenceService are in Running and Ready=True state.
Step2: Delete the model from the object store that is referenced by the deployment.
Step3: Kill the pod and the new pod gets stuck with CrashLoopBackoff state. InferenceService is still in Ready=True state.

No logs were observed in kfserving-controller during the deployment restart and the new pod was trying to come up. Is this an expected behaviour as the Ready=True state is misleading though the deployment/pods are not in Running state?

**Environment:**

- Istio Version: v1.9.7
- Knative Version: v0.18.0
- KFServing Version: v0.5.1
- KFserving-storage-initializer: v0.5.1

`Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.9", GitCommit:"7a576bc3935a6b555e33346fd73ad77c925e9e4a", GitTreeState:"clean", BuildDate:"2021-07-15T20:56:38Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}`

hdefazio:
When deploying a model with an error (e.g. a nonexistent model path) the ISVC modelStatus was stuck in Pending / InProgress. This happened because the Predictor's reconcile function was catching the failing pod's storage initializer container in the running state and never rechecked in order to catch the container in the terminated state. This PR will requeue the Predictor's reconcile function if the storage initializer container is caught in the running state and there was a previous error ( a hint that something may be wrong even if our current status is running). 
The ideal scenario is that we now catch the storage initializer in the terminated state and update the modelStatus to FailedToLoad  / BlockedByFailedLoad. However, the PR also adds a secondary check for serverless deployment so that if we miss the terminated state and the deployment has scaled down to 0, we can retrieve the error status from knative. 

**Type of changes**

Fixes #1859

- [ ] Added a unit test to cover the case where we need to use the knative error information to update the ISVC modelStatus
- [ ] Live test:
1. Deployed a model via ODH UI and waited for it to scale to 1
2. Edited the model config to point to a nonexistent model path 
model_config
3. After the deployment has scaled to 0, double checked that the final ISVC modelStatus was FailedToLoad  / BlockedByFailedLoad.
ISVC_final_modelStatus

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -565,24 +565,27 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
 	return true
 }
 
-func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeployment bool) {
+func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeployment bool, serviceStatus *knservingv1.ServiceStatus) bool {
 	// Check at least one pod is running for the latest revision of inferenceservice
 	totalCopies := len(podList.Items)
 	if totalCopies == 0 {
-		ss.UpdateModelRevisionStates(Pending, totalCopies, nil)
-		return
-	}
-	// Update model state to 'Loaded' if inferenceservice status is ready.
-	// For serverless deployment, the latest created revision and the latest ready revision should be equal
-	if ss.IsReady() {
-		if rawDeployment {
-			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
-			return
-		} else if statusSpec.LatestCreatedRevision == statusSpec.LatestReadyRevision {
-			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
-			return
+		if !rawDeployment {
+			// Make sure we haven't scaled down to 0 because of an error
+			for _, knativeCond := range serviceStatus.Conditions {
+				if knativeCond.Status == "False" {
+					// If any of the knative statuses are False, the model failed
+					// Hopefully the lastFailureInfo already has the info we need, so we don't update it here
+					ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, nil)
+					return true
+				}
+			}
 		}
+
+		// If we made it here then hopefully there are 0 pods because we're just getting started and therefore Pending seems appropriate
+		ss.UpdateModelRevisionStates(Pending, totalCopies, nil)
+		return true
 	}
+
 	// Update model state to 'Loading' if storage initializer is running.
 	// If the storage initializer is terminated due to error or crashloopbackoff, update model
 	// state to 'ModelLoadFailed' with failure info.
@@ -590,25 +593,47 @@ func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatu
 		if cs.Name == constants.StorageInitializerContainerName {
 			switch {
 			case cs.State.Running != nil:
-				ss.UpdateModelRevisionStates(Loading, totalCopies, nil)
-				return
+				// Double check that we aren't missing an error because the cs is looping between an error state and running
+				if cs.LastTerminationState.Terminated != nil {
+					// Requeue the Predictor reconcile loop if there was a previous error reported
+					// to make sure we aren't just temporarily in the cs.State.Running case before moving to the cs.State.Terminated case
+					return false
+				} else {
+					// If there is no previous error, we should be okay to move into the Loading state
+					ss.UpdateModelRevisionStates(Loading, totalCopies, nil)
+					return true
+				}
+
 			case cs.State.Terminated != nil && cs.State.Terminated.Reason == constants.StateReasonError:
 				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:   ModelLoadFailed,
 					Message:  cs.State.Terminated.Message,
 					ExitCode: cs.State.Terminated.ExitCode,
 				})
-				return
+				return true
 			case cs.State.Waiting != nil && cs.State.Waiting.Reason == constants.StateReasonCrashLoopBackOff:
 				ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, &FailureInfo{
 					Reason:   ModelLoadFailed,
 					Message:  cs.LastTerminationState.Terminated.Message,
 					ExitCode: cs.LastTerminationState.Terminated.ExitCode,
 				})
-				return
+				return true
 			}
 		}
 	}
+
+	// Update model state to 'Loaded' if inferenceservice status is ready.
+	// For serverless deployment, the latest created revision and the latest ready revision should be equal
+	if ss.IsReady() {
+		if rawDeployment {
+			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
+			return true
+		} else if statusSpec.LatestCreatedRevision == statusSpec.LatestReadyRevision {
+			ss.UpdateModelRevisionStates(Loaded, totalCopies, nil)
+			return true
+		}
+	}
+
 	// If the kserve container is terminated due to error or crashloopbackoff, update model
 	// state to 'ModelLoadFailed' with failure info.
 	for _, cs := range podList.Items[0].Status.ContainerStatuses {
@@ -631,4 +656,5 @@ func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatu
 			}
 		}
 	}
+	return true
 }
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -332,6 +332,7 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	var podLabelValue string
 
 	// Here we allow switch between knative and vanilla deployment
+	kstatus := &knservingv1.ServiceStatus{}
 	if p.deploymentMode == constants.RawDeployment {
 		rawDeployment = true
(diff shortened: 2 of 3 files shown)


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -332,6 +333,8 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	var podLabelValue string
 
 	// Here we allow switch between knative and vanilla deployment
+	kstatus := knservingv1.ServiceStatus{}
```
spolti: why duplicate the envs here?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -371,7 +374,10 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 		if err := controllerutil.SetControllerReference(isvc, r.Service, p.scheme); err != nil {
 			return ctrl.Result{}, errors.Wrapf(err, "fails to set owner reference for predictor")
 		}
-		status, err := r.Reconcile()
+
+		var err error
+		status, err = r.Reconcile()
+		fmt.Println("updating status to ", *status)
```
spolti: left over?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -565,50 +566,73 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
 	return true
 }
 
-func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeployment bool) {
+func (ss *InferenceServiceStatus) PropagateModelStatus(statusSpec ComponentStatusSpec, podList *v1.PodList, rawDeployment bool, serviceStatus *knservingv1.ServiceStatus) bool {
```
spolti, line 568: looking here: https://github.com/kserve/kserve/blob/master/Makefile#L336C1-L337C125

there is no difference between returning true or false, the reconciliation will happen anyway, can you please elaborate on why returning a bool is required?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -565,50 +566,73 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
...
 	// If the storage initializer is terminated due to error or crashloopbackoff, update model
 	// state to 'ModelLoadFailed' with failure info.
 	for _, cs := range podList.Items[0].Status.InitContainerStatuses {
 		if cs.Name == constants.StorageInitializerContainerName {
 			switch {
 			case cs.State.Running != nil:
-				ss.UpdateModelRevisionStates(Loading, totalCopies, nil)
-				return
+				// Double check that we aren't missing an error
+				if cs.LastTerminationState.Terminated != nil {
```
spolti, line 597: I am not sure if this extra check here has any effect.

hdefazio: The intention was to only re-trigger the reconcile loop if there might be an issue (catch the case where the storage-initializer is looping between an error and running. I was seeing crashloop -> running -> terminated). But if there's no error then it would be a waste to re-check the statuses

pkg/apis/serving/v1beta1/inference_service_status.go
hdefazio: I'm open to a better solution, I returned bool just for this code block https://github.com/kserve/kserve/blob/cd84424ffd3d14b207c92575f9f026c5693867f3/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L395 - so I would only re-trigger the reconcile loop again if there might be an issue with the storage-initilaizer running status

pkg/apis/serving/v1beta1/inference_service_status.go
spolti: Ok, please add this explanation as comment.

spolti: instead of triggering a new loop, what happens if you set the revision state?

hdefazio: Without this change, the last time the reconcile loop happens it catches the storage-initializer in the running state and sets the revision state to Loading. 

I didn't directly just set the revision state to a Failure instead because I really wanted to catch the terminated state (that felt like the most correct sequence of events). The knative failure check is just a backup to make sure that if we still end up missing the terminated state that we don't see the same status issue that this PR is addressing (if we missed the terminated state and didn't check for knative failures we would end up as Pending because the deployment scaled down to 0).

spolti: ok, make sense.

spolti:
cc @sivanantha321

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -565,50 +565,75 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
...
+		if !rawDeployment {
+			// Make sure we haven't scaled down to 0 because of an error
+			for _, knativeCond := range serviceStatus.Conditions {
+				if knativeCond.Status == "False" {
+					// If any of the knative statuses are False, the model failed
+					// Hopefully the lastFailureInfo already has the info we need, so we don't update it here
+					ss.UpdateModelRevisionStates(FailedToLoad, totalCopies, nil)
+					return true
+				}
+			}
```
israel-hdez, line 581: I'm not sure if I'm missing it, but I don't manage to see coverage for these lines in the units. Is it possible to add a unit for this case?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -565,50 +565,75 @@ func (ss *InferenceServiceStatus) SetModelFailureInfo(info *FailureInfo) bool {
...
 		if cs.Name == constants.StorageInitializerContainerName {
 			switch {
 			case cs.State.Running != nil:
-				ss.UpdateModelRevisionStates(Loading, totalCopies, nil)
-				return
+				// Double check that we aren't missing an error because the cs is looping between an error state and running
+				if cs.LastTerminationState.Terminated != nil {
+					// Requeue the Predictor reconcile loop if there was a previous error reported
+					// to make sure we aren't just temporarily in the cs.State.Running case before moving to the cs.State.Terminated case
+					return false
```
israel-hdez, line 600: Also, I'm not seeing coverage for this branch in the units.

From code flow, I'd infer the reported status would be `Pending`. I'm a little bit confused about if that would be a good state to report if the storage initializer is running.

israel-hdez (commented):
I think the approach is good.

I left two comments

israel-hdez:
@hdefazio May you rewrite your testing instructions keeping aside the ODH UI?
The ODH UI is not applicable to this project.

sivanantha321:
@hdefazio Can you help sign the commits?  The DCO check is failing.

hdefazio:
@sivanantha321  Just fixed it!

pkg/apis/serving/v1beta1/inference_service_status.go
sivanantha321: Resolved

sivanantha321:
Nice work!
/lgtm
/approve

PR #4005, opened 2024-10-17
Update quick install script

johnugeorge:
Update quick install script with latest version of Istio and Knative


hack/quick_install.sh
```diff
@@ -17,11 +17,11 @@ Help() {
    echo
 }
 
-export ISTIO_VERSION=1.20.4
-export KNATIVE_OPERATOR_VERSION=v1.14.5
-export KNATIVE_SERVING_VERSION=1.13.1
+export ISTIO_VERSION=1.23.2
```
spolti, line 20: We are actually on Istion v1.23.0, as it is a minor version update, it should be fine.

spolti (approved):
just a minor comment.

/lgtm

hack/quick_install.sh
```diff
@@ -17,11 +17,11 @@ Help() {
...
+export KNATIVE_OPERATOR_VERSION=v1.15.7
```
Jooho, line 21: KNATIVE 1.16 is planned to release next week. why don't we wait 1 week and use that?

PR #4030, opened 2024-11-01
Make label and annotation propagation configurable

### LINKED ISSUE #3710: Make label and annotation propagation configurable

spolti:
chore:	Make the DisallaowedAnnotations and Labels configurable through
	ConfigMap so users can configured it quickly.

fixes https://github.com/kserve/kserve/issues/3710

The rationale behind using the `InferenceServiceConfig` key instead of the `deploy` as suggested initially in the `ConfigMap` is because the isvc is basically, already injected on every place that requires a change to read it from the configuration instead of the constants.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


spolti:
It is still missing the `python-sdk` and tests updates, I will handle it on Monday. Only tested manually so far.

spolti:
Hi @yuzisun @terrytangyuan @andyi2it, can you please take a look at this PR once you have a chance?
@israel-hdez fyi

config/configmap/inferenceservice.yaml
```diff
@@ -41,7 +41,38 @@ data:
...
+          "my.custom.annotation/1"  
+        ],
+        "serviceLabelDisallowedList": [
+          "my.custom.label.1"  
+        ]
+       }
+    # Example of isvc configuration
+    inferenceservice: |-
+      {
+        # ServiceAnnotationDisallowedList is a list of annotations that are not allowed to be propagated to Knative
```
terrytangyuan, line 58: This is only applicable for Knative revisions?

spolti: Yes

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -31,11 +31,12 @@ import (
 
 // ConfigMap Keys
 const (
-	ExplainerConfigKeyName = "explainers"
-	IngressConfigKeyName   = "ingress"
-	DeployConfigName       = "deploy"
-	LocalModelConfigName   = "localModel"
-	SecurityConfigName     = "security"
+	InferenceServiceConfigKeyName = "inferenceService"
```
israel-hdez: I'm seeing here `S` in uppercase, but the manifest example in `inferenceservice.yaml` is all lowercase. Should the case match?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -76,9 +76,12 @@ func (p *Predictor) Reconcile(isvc *v1beta1.InferenceService) (ctrl.Result, erro
 	var sRuntimeAnnotations map[string]string
 
 	annotations := utils.Filter(isvc.Annotations, func(key string) bool {
-		return !utils.Includes(constants.ServiceAnnotationDisallowedList, key)
+		return !utils.Includes(p.inferenceServiceConfig.ServiceAnnotationDisallowedList, key)
 	})
 
+	fmt.Printf("\n Predictor custom annotations %v\n", p.inferenceServiceConfig.ServiceAnnotationDisallowedList)
+	fmt.Printf("\n Predictor custom labels %v\n", p.inferenceServiceConfig.ServiceLabelDisallowedList)
```
israel-hdez: Use `p.Log.Info` instead of `fmt.Printf`?

israel-hdez (commented):
I have one small comment.

pkg/apis/serving/v1beta1/configmap.go
spolti: Nice catch, the test didn't fail because I used the value in the camel case as defined here.
updated the example in the yaml.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
spolti: Oh, it is a left over from my manual tests, I think we can keep it as debug.

spolti:
@israel-hdez @terrytangyuan @andyi2it @yuzisun Hi, can you please take a look when you have a chance?

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #4048, opened 2024-11-12
LocalModelNode Agent

HotsauceLee:
Implement local model node agent to reconcile the expected model cache list against the actual models in the local node volumes

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


charts/kserve-resources/templates/localmodelnode/deployment.yaml
```diff
@@ -45,12 +45,25 @@ spec:
...
           limits:
             cpu: 100m
             memory: 300Mi
           requests:
             cpu: 100m
             memory: 200Mi
+      volumes:
+        - name: local-models
+          hostPath:
+            path: /mnt/models
```
greenmoon55, line 66: you can make it configurable in helm but doesn't need to be in this pr.

charts/kserve-resources/templates/localmodelnode/deployment.yaml
```diff
@@ -45,12 +45,25 @@ spec:
...
+            type: Directory
```
greenmoon55, line 67: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-volume-types
type: DirectoryOrCreate because we don't have this folder

greenmoon55: We also need to test if `fsgroupChangepolicy` works for hostpath. i.e. it changes the ownership of the path to a non-root user.

go.sum
```diff
@@ -1,3 +1,4 @@
+cel.dev/expr v0.15.0/go.mod h1:TRSuuV7DlVCE/uwv5QbAiW/v8l5O8C4eEPHeu7gf7Sg=
```
greenmoon55, line 1: run `go mod tidy` to get rid of these changes.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +55,237 @@ type LocalModelNodeReconciler struct {
...
+	return defaultContainer, nil
+}
+
+func (c *LocalModelNodeReconciler) DownloadModel(ctx context.Context, localModelNode *v1alpha1api.LocalModelNode, jobNamespace string) error {
+	c.Log.Info("Downloading models to", "node", localModelNode.ObjectMeta.Name)
+
+	for _, modelInfo := range localModelNode.Spec.LocalModels {
+		jobName := modelInfo.ModelName + "-" + localModelNode.ObjectMeta.Name
+		c.Log.Info("Launch download job", "name", jobName)
+		job, err := c.launchDownloadJob(jobName, jobNamespace, localModelNode, modelInfo, modelInfo.ModelName, localModelNode.ObjectMeta.Name)
```
greenmoon55, line 200: @yuzisun 
In the old code, we launch a download job if no job exists AND model status != download successful (no job was successful before).
Launch a download job when
Option 1: if (no job exists AND model status != download successful) or model folder does not exist 
Option 2: if no job exists or model folder does not exist
Isn't option 2 better because it doesn't depend on it's own status field if we assume download jobs are never deleted.

we don't need to change anything in this pr.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +55,237 @@ type LocalModelNodeReconciler struct {
...
+	localModels := map[string]struct{}{}
+	for _, entry := range entries {
+		// Models could only exist in sub dir
+		if entry.IsDir() {
+			localModels[entry.Name()] = struct{}{}
+		}
+	}
+	c.Log.Info("Found existing local models", "models", maps.Keys(localModels))
+
+	// 2. Compare with list of models from LocalModelNode CR
```
greenmoon55, line 267: Could you add tests for these deletion cases?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +55,237 @@ type LocalModelNodeReconciler struct {
...
+	// 2. Kick off download jobs for all models in spec
+	localModelConfig, err := v1beta1.NewLocalModelConfig(c.Clientset)
+	if err != nil {
+		c.Log.Error(err, "Failed to get local model config")
+		return reconcile.Result{}, err
+	}
+	defaultJobImage = localModelConfig.DefaultJobImage
+	FSGroup = localModelConfig.FSGroup
+	if err := c.DownloadModel(ctx, localModelNode, localModelConfig.JobNamespace); err != nil {
+		c.Log.Error(err, "Model download err")
```
greenmoon55, line 250: retry by `return reconcile.Result{}, err`?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +55,237 @@ type LocalModelNodeReconciler struct {
...
+	}
+	// Delete models from disk if they are not on the CR.
+	// 1. Scan model dir and get a list of downloaded models
+	entries, err := os.ReadDir(localModelContainerVol + "/models")
+	if err != nil {
+		c.Log.Error(err, "Failed to list model directory", "dir", localModelContainerVol+"/models")
```
greenmoon55, line 256: ditto , return here

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: Another thing - when we deploy this change, our existing jobs are owned by the controller, not the agent. It should reconcile from existing successful jobs and update the status. Worst case it would download all models again which is fine.

yuzisun: I agree, with the node agent we should no longer depend on the download job status. I am not sure if it is safe to assume that we never delete the download jobs, we can check the local model node resource if the model has been successfully downloaded.

pkg/controller/v1alpha1/localmodelnode/controller.go
yuzisun: Let's also wrap deletion code block into a function

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +55,237 @@ type LocalModelNodeReconciler struct {
...
+			}
+		}
+	}
+
 	return reconcile.Result{}, nil
 }
 
 func (c *LocalModelNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1api.LocalModelNode{}).
```
greenmoon55, line 288: (different pr) might be better filter by node name in the predicate  https://github.com/kubernetes-sigs/controller-runtime/blob/c1331a5c7c5880c35223d5dec0d87ac166554d46/pkg/builder/controller.go#L83

PR #4055, opened 2024-11-15
Update max_model_len calculation and fixup encoder pooling

Datta0:

[vLLM updated their max_model_len](https://github.com/vllm-project/vllm/blob/2ac6d0e75bc846998da56b50bf4f8853cb36d484/vllm/config.py#L1870) calculations and hence we'd want to use the same. 
Also, `_mean_pooling` wasn't always happening on the right device. so fix it.

### DIFF
--- a/python/huggingfaceserver/huggingfaceserver/utils.py
+++ b/python/huggingfaceserver/huggingfaceserver/utils.py
@@ -13,19 +13,25 @@
 # limitations under the License.
 
 import torch
+import os
 from transformers import PretrainedConfig
-from typing import Optional
+from typing import Optional, Union, List, Any
 
 from kserve.logging import logger
 
 
+ALLOW_LONG_MAX_MODEL_LEN = "ALLOW_LONG_MAX_MODEL_LEN"
+
+
 # "This implementation is based on vLLM's _get_and_verify_max_len "
-# "https://github.com/vllm-project/vllm/blob/a377f0bd5e1fa0ca069e3dbf28f4de5af64d0bb1/vllm/config.py#L1160"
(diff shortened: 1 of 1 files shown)


python/huggingfaceserver/huggingfaceserver/utils.py
```diff
@@ -13,19 +13,25 @@
...
 
 from kserve.logging import logger
 
 
+ALLOW_LONG_MAX_MODEL_LEN = "ALLOW_LONG_MAX_MODEL_LEN"
+
+
 # "This implementation is based on vLLM's _get_and_verify_max_len "
-# "https://github.com/vllm-project/vllm/blob/a377f0bd5e1fa0ca069e3dbf28f4de5af64d0bb1/vllm/config.py#L1160"
+# "https://github.com/vllm-project/vllm/blob/2ac6d0e75bc846998da56b50bf4f8853cb36d484/vllm/config.py#L1870"
```
gavrissh, line 27: ~~I believe you can consume these methods through the vLLM pkg now~~

HFModel does not consume vLLM pkg.

Datta0:
@yuzisun @sivanantha321  can you please take a look at this?

python/huggingfaceserver/huggingfaceserver/utils.py
yuzisun: I have the same question if we can reuse this function from vllm ?  It is hard to get keep up to date as vLLM frequently changes this

Datta0: As of now, we don't mandate installation/existence of vLLM package. Its an extra and if one chooses to run on CPU or just via HuggingfaceGenerativeModel, it'd be an unnecessary.
Do we want to make vLLM mandataory?

PR #4072, opened 2024-11-19
chore: use patch instead of update for finalizer changes

whynowy:

In general, when adding/removing finalizers, a `patch` operation should be used instead of an `update` to avoid other fields getting updated unexpectedly.

Fixes #

**Type of changes**

- [ x] Bug fix (non-breaking change which fixes an issue)

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -139,15 +142,17 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		// The object is not being deleted, so if it does not have our finalizer,
 		// then lets add the finalizer and update the object. This is equivalent
 		// registering our finalizer.
-		if !utils.Includes(isvc.ObjectMeta.Finalizers, finalizerName) {
-			isvc.ObjectMeta.Finalizers = append(isvc.ObjectMeta.Finalizers, finalizerName)
-			if err := r.Update(context.Background(), isvc); err != nil {
+		if !controllerutil.ContainsFinalizer(isvc, finalizerName) {
+			controllerutil.AddFinalizer(isvc, finalizerName)
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.ObjectMeta.Finalizers, ",") + "]"
```
whynowy, line 147: Works for no finalizers as well.

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	k8s.io/code-generator v0.30.4
 	k8s.io/component-helpers v0.30.4
 	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.130.1
```
whynowy: The changes in `go.mod` and `go.sum` were caused by running `go mod tidy`. 

Obviously both `v1` and `v2` are used in this project, which can be updated separately.

whynowy:
@terrytangyuan @lizzzcai - Can any of you give an approval to run the CI and review?

pkg/controller/v1beta1/inferenceservice/controller.go
greenmoon55: noob question, is it the best way to create a patch?

whynowy: There are multiple ways to patch, depends on the `types.xxxxPatchType`. This approach with `MergePatchType` will never go wrong - It clearly specifies patching `finalizers` only, replacing everything under it with the values specified.

whynowy:
@lizzzcai - Could you approve the CI again?

go.mod
terrytangyuan: I wonder why two versions are used

terrytangyuan:
Good to see you here, Derek!

go.mod
whynowy: https://github.com/kserve/kserve/blob/master/cmd/spec-gen/main.go#L26
https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/agent_injector.go#L26
https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/suite_test.go#L27

I don't want to touch them in this PR.

pkg/controller/v1beta1/inferenceservice/controller.go
greenmoon55: Sorry, I should be more specific. 
I tried https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#MergeFrom so I don't need to write the patch yaml on my own.
```
original := client.MergeFrom(localModel.DeepCopy())
localModel.ObjectMeta.Finalizers = utils.RemoveString(localModel.ObjectMeta.Finalizers, finalizerName)
if err := c.Patch(context.TODO(), localModel, original); err != nil {
	return ctrl.Result{}, err
}
```
I got this approach from
https://stackoverflow.com/questions/57712941/what-is-the-proper-way-to-patch-an-object-with-controller-runtime
Your code looks good to me too! Thanks!

whynowy: Thanks, understood.

whynowy:
Can I get a review?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -156,8 +161,10 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			}
 
 			// remove our finalizer from the list and update it.
-			isvc.ObjectMeta.Finalizers = utils.RemoveString(isvc.ObjectMeta.Finalizers, finalizerName)
-			if err := r.Update(context.Background(), isvc); err != nil {
+			controllerutil.RemoveFinalizer(isvc, finalizerName)
```
terrytangyuan, line 164: Can we link the reference to best practice in the comment?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -156,8 +161,10 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(isvc.ObjectMeta.Finalizers, ",") + "]"
+			patchJson, _ := yaml.YAMLToJSON([]byte(patchYaml))
+			if err := r.Patch(ctx, isvc, client.RawPatch(types.MergePatchType, patchJson)); err != nil {
```
terrytangyuan, line 167: Anything needs to be updated in https://github.com/kserve/kserve/tree/master/config/rbac?

whynowy: I guess no.  It already has `patch` privilege. https://github.com/kserve/kserve/blob/master/config/rbac/role.yaml#L141-L153

pkg/controller/v1beta1/inferenceservice/controller.go
whynowy: I don't have a link unfortunately. I personally saw some small problems with `update`, for example, for fields like `resources.requests.memory`, if you set a value `1024Mi`, after a `update`, it will be updated to `1Gi`. Even though there's no difference, but tools like Argo CD consider there's a diff.

Things like this is a not big problem but kind of annoying. But using patch to only update the specified scope of the spec is much safer.

terrytangyuan:
Thanks! 

/lgtm
/approve

whynowy:
> Thanks!
> 
> /lgtm /approve

Thank you @terrytangyuan ! CI approval?

terrytangyuan:
Triggered

whynowy:
@terrytangyuan - The CI failures don't seem to be created by this PR?

terrytangyuan:
Merged manually

PR #4075, opened 2024-11-20
LocalModelNode Agent that creates download jobs and update statuses from jobs

greenmoon55:
This is an update on #4048 . 
Local model node agent reconciles the expected model cache list against the actual models in the local disk.
Todo items not in this pr
1. Make local path configurable in helm
2. Handle deletion
3. Check local disk and redownload models if they are not cached

Fixes #

**Type of changes**

Added a new test
Manually create a LocalModelNode CR and check it is downloaded my the agent on the correct node.
```
- apiVersion: serving.kserve.io/v1alpha1
  kind: LocalModelNode
  metadata:
    creationTimestamp: "2024-11-06T21:45:53Z"
    generation: 4
    name: cluster2-worker2
    resourceVersion: "2141934"
    uid: 69a67abe-81c5-4088-a842-d261c7dafee0
  spec:
    localModels:
    - modelName: iris2
      sourceModelUri: gs://kfserving-examples/models/sklearn/1.0/model
  status:
    modelStatus:
      iris2: ModelDownloaded
```

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +53,209 @@ type LocalModelNodeReconciler struct {
...
+var (
+	defaultJobImage    = "kserve/storage-initializer:latest" // Could be overwritten by the value in the configmap
+	FSGroup            *int64                                // Could be overwritten by the value in the configmap
+	jobNamespace       string
+	nodeName           = os.Getenv("NODE_NAME") // Name of current node
+	localModelHostPath = "/mnt/models"          // Host node directory to store local models
+)
+
+// Launches a job if not exist, or return the existing job
+func (c *LocalModelNodeReconciler) launchJob(jobName string, localModelNode *v1alpha1api.LocalModelNode, modelInfo v1alpha1api.LocalModelInfo, claimName string) (*batchv1.Job, error) {
```
greenmoon55: Still using pvc for now, this is mostly copied from the controller.

config/localmodelnodes/manager.yaml
```diff
@@ -21,6 +21,8 @@ spec:
       annotations:
         kubectl.kubernetes.io/default-container: manager
     spec:
+      nodeSelector:
```
yuzisun, line 24: Can this be configured ?

greenmoon55: updated it, but I haven't test it yet. 😅

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +53,209 @@ type LocalModelNodeReconciler struct {
...
+		{
+			MountPath: mountPath,
+			Name:      "kserve-pvc-source",
+			ReadOnly:  false,
+			SubPath:   "models/" + modelInfo.ModelName,
+		},
+	}
+	expectedJob := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      jobName + "dryrun",
```
yuzisun, line 101: Why do we need to add "dryrun" into the job name?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +53,209 @@ type LocalModelNodeReconciler struct {
...
+	if err != nil {
+		if apierr.IsNotFound(err) {
+			jobFound = false
+		} else {
+			c.Log.Error(err, "Failed to get job", "name", jobName)
+			return job, err
+		}
+	}
+
+	container.Name = jobName
```
yuzisun: container name can be static

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +53,209 @@ type LocalModelNodeReconciler struct {
...
+			c.Log.Error(err, "Failed to delete job.", "name", job.Name)
+			return nil, err
+		}
+	}
+
+	if err := controllerutil.SetControllerReference(localModelNode, expectedJob, c.Scheme); err != nil {
+		c.Log.Error(err, "Failed to set controller reference", "name", modelInfo.ModelName)
+		return nil, err
+	}
+	expectedJob.Name = jobName
```
yuzisun, line 149: it is not necessary to change the job name back as we can use the original name for dry run

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -45,13 +53,209 @@ type LocalModelNodeReconciler struct {
...
+	}
+	return defaultContainer, nil
+}
+
+func (c *LocalModelNodeReconciler) downloadModels(ctx context.Context, localModelNode *v1alpha1api.LocalModelNode) error {
+	c.Log.Info("Downloading models to", "node", localModelNode.ObjectMeta.Name)
+
+	for _, modelInfo := range localModelNode.Spec.LocalModels {
+		if status, ok := localModelNode.Status.ModelStatus[modelInfo.ModelName]; ok {
+			if status == v1alpha1api.ModelDownloaded {
```
yuzisun, line 195: How do we handle the case if the model was downloaded but then node is rebuilt ?

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: probably to avoid name conflict if the job already exists.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: To remove the "dryrun" suffix. Maybe we can remove this dry run logic since it is not very helpful.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: I'm going to handle it in a different PR. We will probably check if the folder exists here.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: If the job already exists, I remember dryrun would fail because of the name conflict.

yuzisun (approved):
/lgtm
/approve

PR #4095, opened 2024-12-03
Detect missing models and redownload models

greenmoon55:
Now we could have multiple download jobs. We use labels and creation timestamp to create latest  
For model cache, it detects missing models and recreate download jobs.

image

- If the folder exists
  - If download had succeeded before (model status = download successful), return early
  - If no job exists, create a download job. 
  - If job exists, update the model status with job status. This could happen if the job just completed.
- If the folder does not exist
   -  If no job exists or the job was successful (job not failed, failed pods would be retried automatically), create a download job to download again.
   -  Otherwise the job could be running.

Fixes #

**Type of changes**

Test case for detecting missing local model folder with mocked file system.
Manually tested download and deletion for localmodel controller and localmodelnode agent in a kind cluster. 

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -195,7 +195,7 @@ func (c *LocalModelNodeReconciler) getContainerSpecForStorageUri(ctx context.Con
 	return defaultContainer, nil
 }
 
-func (c *LocalModelNodeReconciler) getLatestJob(ctx context.Context, modelName string, nodeName string) (*batchv1.Job, error) {
+func (c *LocalModelNodeReconciler) getLatestJob(ctx context.Context, modelName string, nodeName string, excludeSucceeded bool) (*batchv1.Job, error) {
 	jobList := &batchv1.JobList{}
 	labelSelector := map[string]string{
 		"model": modelName,
@@ -210,11 +210,13 @@ func (c *LocalModelNodeReconciler) getLatestJob(ctx context.Context, modelName s
 	c.Log.Info("Found jobs", "model", modelName, "num of jobs", len(jobList.Items))
 	var latestJob *batchv1.Job
 	for i, job := range jobList.Items {
+		if excludeSucceeded && job.Status.Succeeded > 0 {
+			continue
+		}
 		if latestJob == nil || job.CreationTimestamp.After(latestJob.CreationTimestamp.Time) {
 			latestJob = &jobList.Items[i]
 		}
 	}
-	c.Log.Info("found job", "job", latestJob)
 	return latestJob, nil
 }
 
@@ -250,13 +252,12 @@ func (c *LocalModelNodeReconciler) downloadModels(ctx context.Context, localMode
 				}
 				continue
 			}
-			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName)
+			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName, false)
 			if err != nil {
 				c.Log.Error(err, "Failed to getLatestJob", "model", modelInfo.ModelName, "node", nodeName)
 				return err
 			}
 			// If job is not found, create a new one. Because download could be incomplete.
-			// Otherwise model status is updated by the job status outside this if statement
 			if job == nil {
 				c.Log.Info("Model folder exists, creating download job", "model", modelInfo.ModelName)
 				job, err = c.launchJob(ctx, *localModelNode, modelInfo)
@@ -268,10 +269,13 @@ func (c *LocalModelNodeReconciler) downloadModels(ctx context.Context, localMode
 		} else {
 			// Folder does not exist
 			c.Log.Info("Model folder not found", "model", modelInfo.ModelName)
-			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName)
+			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName, true)
 			if err != nil {
 				c.Log.Error(err, "Failed to getLatestJob", "model", modelInfo.ModelName, "node", nodeName)
 				return err
+			}
+			if job != nil {
+				c.Log.Info("model status from latest job", "model", modelInfo.ModelName, "status", getModelStatusFromJobStatus(job.Status))
 			}
 			// Recreate job if it has been terminated because the model is missing locally
 			if job == nil || job.Status.Succeeded > 0 {
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -63,12 +63,13 @@ const (
 )
 
 var (
-	defaultJobImage  = "kserve/storage-initializer:latest" // Can be overwritten by the value in the configmap
-	FSGroup          *int64
-	jobNamespace     string
-	nodeName         = os.Getenv("NODE_NAME") // Name of current node, passed as an env variable via downward API
-	modelsRootFolder = filepath.Join(MountPath, `models`)
-	fsHelper         fileSystemInterface
+	defaultJobImage            = "kserve/storage-initializer:latest" // Can be overwritten by the value in the configmap
+	FSGroup                    *int64
+	jobNamespace               string
+	jobTTLSecondsAfterFinished int32 = 86400                  // Can be overwritten by the value in the configmap
+	nodeName                         = os.Getenv("NODE_NAME") // Name of current node, passed as an env variable via downward API
+	modelsRootFolder                 = filepath.Join(MountPath, "models")
+	fsHelper                   fileSystemInterface
 )
 
 type fileSystemInterface interface {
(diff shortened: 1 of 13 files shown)


pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -133,13 +165,26 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, jobName string
...
 	if jobFound {
-		bg := metav1.DeletePropagationBackground
+		// If the job is still running, do not delete it
+		if recreateJob && job.Status.Succeeded == 0 {
+			c.Log.Info("New job is not successful yet, skipping deletion", "name", job.Name)
+			return job, nil
+		}
+		c.Log.Info("Deleting job.", "name", job.Name)
+		policy := metav1.DeletePropagationForeground
 		err = jobs.Delete(ctx, job.Name, metav1.DeleteOptions{
```
greenmoon55: One small issue - in my tests the delete request is successful, but it cannot create a new job immediately because the deletion is still in progress. It would work in the next reconciliation loop.

pkg/controller/v1alpha1/localmodelnode/controller_test.go
```diff
@@ -214,20 +278,86 @@ var _ = Describe("CachedModel controller", func() {
...
+
+			// Manually trigger reconcillation
+			patch := client.MergeFrom(localModelNode.DeepCopy())
+			localModelNode.Annotations = map[string]string{"foo": "bar"}
+			Expect(k8sClient.Patch(ctx, localModelNode, patch)).Should(Succeed())
+
+			newJob := &batchv1.Job{}
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: modelName + "-" + nodeName, Namespace: modelCacheNamespace}, newJob)
+				// Todo: Figure out why we have to manually remove finalizers in the test. It works in kind clusters.
```
greenmoon55: In this test the finalizer seems never removed, but it works in my kind cluster.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: Alternatively we could use generate name on the job and use labels for filtering, but then I don't want to have duplicate jobs.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -133,13 +165,26 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, jobName string
 	if err != nil {
 		return nil, err
 	}
-	if job != nil && reflect.DeepEqual(job.Spec.Template.Spec, dryrunJob.Spec.Template.Spec) {
+
+	// If we do not recreate job on purpose, return the existing job if it is the same as the expected job
+	if !recreateJob && job != nil && reflect.DeepEqual(job.Spec.Template.Spec, dryrunJob.Spec.Template.Spec) {
```
HotsauceLee: would checking `jobFound` better than`job != nil` if it's checking existence of the job?

pkg/controller/v1alpha1/localmodelnode/controller.go
HotsauceLee: IMO creating the job in the next reconciliation loop is fine since this shouldn't happen a lot.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -199,14 +244,24 @@ func (c *LocalModelNodeReconciler) downloadModels(ctx context.Context, localMode
...
+		recreateJob := false
 		if status, ok := localModelNode.Status.ModelStatus[modelInfo.ModelName]; ok {
 			if status == v1alpha1api.ModelDownloaded {
-				continue
+				newStatus[modelInfo.ModelName] = v1alpha1api.ModelDownloaded
+				hasModelFolder, err := fsHelper.hasModelFolder(modelInfo.ModelName)
+				if err != nil {
+					c.Log.Error(err, "Failed to check model folder", "model", modelInfo.ModelName, "folder", getModelFolder(modelInfo.ModelName))
+				}
+				if hasModelFolder {
```
HotsauceLee: `hasModelFolder` will be false if `err != nil`, so it will recreate the job when failed to scan the dir for reasons other than non-exist. I'm assuming that's not what we want?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -199,14 +244,24 @@ func (c *LocalModelNodeReconciler) downloadModels(ctx context.Context, localMode
 
 	newStatus := map[string]v1alpha1api.ModelStatus{}
 	for _, modelInfo := range localModelNode.Spec.LocalModels {
+		recreateJob := false
 		if status, ok := localModelNode.Status.ModelStatus[modelInfo.ModelName]; ok {
 			if status == v1alpha1api.ModelDownloaded {
-				continue
+				newStatus[modelInfo.ModelName] = v1alpha1api.ModelDownloaded
+				hasModelFolder, err := fsHelper.hasModelFolder(modelInfo.ModelName)
+				if err != nil {
```
HotsauceLee, line 235:
```suggestion
				if err != nil || hasModelFolder {

```

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: added `return err` above

pkg/controller/v1alpha1/localmodelnode/controller_test.go
greenmoon55: basically the job cannot be cleaned up in the test for some reason

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	k8s.io/code-generator v0.30.4
 	k8s.io/component-helpers v0.30.4
 	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.130.1
```
greenmoon55: not sure why this got added.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -67,12 +68,43 @@ var (
...
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return false, err
+}
+
+func (c *LocalModelNodeReconciler) getOrLaunchJob(ctx context.Context, jobName string, localModelNode *v1alpha1api.LocalModelNode, modelInfo v1alpha1api.LocalModelInfo, recreateJob bool) (*batchv1.Job, error) {
```
yuzisun: I found it is  a bit odd to pass the `recreateJob` flag to a function and make the code harder to read.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -62,34 +62,53 @@ const (
...
+	folder := getModelFolder(modelName)
+	_, err := f.readDir(folder)
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return false, err
+}
```
yuzisun: This code should be moved out of the controller.go

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -192,60 +162,127 @@ func (c *LocalModelNodeReconciler) getContainerSpecForStorageUri(ctx context.Con
...
+					newStatus[modelInfo.ModelName] = v1alpha1api.ModelDownloaded
+				}
+				continue
+			}
+			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName, false)
+			if err != nil {
+				c.Log.Error(err, "Failed to getLatestJob", "model", modelInfo.ModelName, "node", nodeName)
+				return err
+			}
+			// If job is not found, create a new one. Because download could be incomplete.
```
yuzisun, line 231: job can also be cleaned up by TTL, do we want to create download job in this case?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -192,60 +162,127 @@ func (c *LocalModelNodeReconciler) getContainerSpecForStorageUri(ctx context.Con
...
+			job, err = c.getLatestJob(ctx, modelInfo.ModelName, nodeName, true)
+			if err != nil {
+				c.Log.Error(err, "Failed to getLatestJob", "model", modelInfo.ModelName, "node", nodeName)
+				return err
+			}
+			if job != nil {
+				c.Log.Info("model status from latest job", "model", modelInfo.ModelName, "status", getModelStatusFromJobStatus(job.Status))
+			}
+			// Recreate job if it has been terminated because the model is missing locally
+			if job == nil || job.Status.Succeeded > 0 {
```
yuzisun, line 254: There could also be the case that the job is failed without creating the folder.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: Yes, job can also be cleaned up by TTL if the job was successful, we would have already updated the status of the model to be downloaded, which would bypass this piece of code. See the `continue` above.

pkg/controller/v1alpha1/localmodelnode/controller.go
greenmoon55: Yeah, then we do not retry these jobs until they are manually removed. What do you think?

HotsauceLee (approved):
lgtm

PR #4099, opened 2024-12-05
Support datetime object serialization in v1/v2 response

### LINKED ISSUE #4096: KServe 0.14 fails to serialize datetime
/kind bug

```python
import kserve
import datetime

class MyKserveModel(kserve.Model):
    def predict(self, data, headers = {}):
        return {"my_date": datetime.date(2024, 12, 3)}
```

and getting error response for 
```
TypeError : Object of type date is not JSON serializable
```

The error should not happen

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

sivanantha321:
Serializes datetime to Iso datetime format with tz info.

Fixes #4096

**Type of changes**

**Checklist**:

**Release note**:

```release-note
Support datetime object serialization in v1/v2 response
```

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/protocol/rest/v1_endpoints.py
+++ b/python/kserve/kserve/protocol/rest/v1_endpoints.py
@@ -16,7 +16,7 @@ from typing import Optional, Union, Dict, List, AsyncIterator
 
 from fastapi import Request, Response, FastAPI, APIRouter
 from starlette.responses import StreamingResponse
-from fastapi.responses import JSONResponse
+from fastapi.responses import ORJSONResponse
 
 from kserve.errors import ModelNotReady
 from ..dataplane import DataPlane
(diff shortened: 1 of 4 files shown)


python/kserve/kserve/protocol/rest/v1_endpoints.py
```diff
@@ -96,7 +96,7 @@ async def predict(self, model_name: str, request: Request) -> Union[Response, Di
             return Response(content=response, headers=response_headers)
         if isinstance(response, AsyncIterator):
             return StreamingResponse(content=response)
-        return JSONResponse(content=response, headers=response_headers)
+        return ORJSONResponse(content=response, headers=response_headers)
```
yuzisun, line 99: `ORJSONResponse` is also much more performant than JSONResponse, is there any other place still using JSONResponse

yuzisun: looks like these are the only places and by default it should be returning `ORJSONResponse`, see 
https://github.com/kserve/kserve/blob/e75491c986473ef86d66e5c0377a81de29e3462c/python/kserve/kserve/model_server.py#L188

PR #4121, opened 2024-12-12
Enable transformer deeper readiness check tests

sivanantha321:
Enable transformer deeper readiness check tests

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/python/kserve/kserve/utils/inference_client_factory.py
+++ b/python/kserve/kserve/utils/inference_client_factory.py
@@ -14,12 +14,14 @@
 
 from typing import Optional
 from ..inference_client import InferenceGRPCClient, InferenceRESTClient, RESTConfig
+from .utils import is_v2
 
 
 class InferenceClientFactory:
     _instance = None
     _grpc_client: Optional[InferenceGRPCClient] = None
-    _rest_client: Optional[InferenceRESTClient] = None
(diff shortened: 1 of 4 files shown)


yuzisun:
@sivanantha321 How is change related to enable deeper readiness check?

sivanantha321:
> @sivanantha321 How is change related to enable deeper readiness check?

It enables tests for deeper readiness check which were previously skipped.

python/kserve/kserve/utils/inference_client_factory.py
```diff
@@ -14,12 +14,14 @@
...
 from typing import Optional
 from ..inference_client import InferenceGRPCClient, InferenceRESTClient, RESTConfig
+from .utils import is_v2
 
 
 class InferenceClientFactory:
     _instance = None
     _grpc_client: Optional[InferenceGRPCClient] = None
-    _rest_client: Optional[InferenceRESTClient] = None
+    _rest_v1_client: Optional[InferenceRESTClient] = None
```
sivanantha321, line 23: This change to avoid the situation where v1 client is used by v2 tests. This is because of the tests are running in single session and the client being singleton. The client is created with the v1 protocol as it is the first test to run and when the v2 test runs it simply returns the v1 client .

PR #4129, opened 2024-12-16
Add Support for OpenAI-compatible Embeddings API

### LINKED ISSUE #4025: Cannot run embedding model
/kind bug

I created an inference service with an embedding model, but I cannot access any endpoint.

I expected the service to respond to calls to `/openai/v1/models` and `/openai/v1/embeddings`.

```
kind: InferenceService
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"serving.kserve.io/v1beta1","kind":"InferenceService","metadata":{"annotations":{"sidecar.istio.io/inject":"false"},"name":"embedding","namespace":"kf-llama-backend"},"spec":{"predictor":{"maxReplicas":1,"minReplicas":1,"model":{"args":["--model_name=embedding","--model_id=thenlper/gte-base","--model_dir=/mnt/models/","--task=text_embedding"],"env":[{"name":"KSERVE_LOGLEVEL","value":"WARNING"}],"modelFormat":{"name":"huggingface"},"resources":{"limits":{"cpu":2,"memory":"8Gi"},"requests":{"cpu":2,"memory":"8Gi"}}},"serviceAccountName":"default-editor"}}}
    sidecar.istio.io/inject: "false"
  creationTimestamp: "2024-10-29T13:46:23Z"
  finalizers:
  - inferenceservice.finalizers
  generation: 1
  name: embedding
  namespace: kf-llama-backend
  resourceVersion: "222824233"
  uid: 9fe8685c-75ee-4359-94b0-28576a6763d2
spec:
  predictor:
    automountServiceAccountToken: false
    maxReplicas: 1
    minReplicas: 1
    model:
      args:
      - --model_name=embedding
      - --model_id=thenlper/gte-base
      - --model_dir=/mnt/models/
      - --task=text_embedding
      env:
      - name: KSERVE_LOGLEVEL
        value: WARNING
      modelFormat:
        name: huggingface
      name: ""
      resources:
        limits:
          cpu: "2"
          memory: 8Gi
        requests:
          cpu: "2"
          memory: 8Gi
    serviceAccountName: default-editor
status:
  address:
    url: http://embedding.kf-llama-backend.svc.cluster.local
  components:
    predictor:
      address:
        url: http://embedding-predictor.kf-llama-backend.svc.cluster.local
      latestCreatedRevision: embedding-predictor-00001
      latestReadyRevision: embedding-predictor-00001
      latestRolledoutRevision: embedding-predictor-00001
      traffic:
      - latestRevision: true
        percent: 100
        revisionName: embedding-predictor-00001
      url: http://embedding-predictor.kf-llama-backend.svc.cluster.local
  conditions:
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    status: "True"
    type: IngressReady
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    severity: Info
    status: "True"
    type: LatestDeploymentReady
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    severity: Info
    status: "True"
    type: PredictorConfigurationReady
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    status: "True"
    type: PredictorReady
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    severity: Info
    status: "True"
    type: PredictorRouteReady
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-10-29T13:48:35Z"
    severity: Info
    status: "True"
    type: RoutesReady
  modelStatus:
    copies:
      failedCopies: 0
      totalCopies: 1
    states:
      activeModelState: Loaded
      targetModelState: Loaded
    transitionStatus: UpToDate
  observedGeneration: 1
  url: http://embedding.kf-llama-backend.svc.cluster.local
```

I can acces the base URL of the service and get a `status: "alive"`. Trying any other OpenAI-compatible endpoint results in `{"detail":"Not Found"}`. 

**Environment:**

- Istio Version: 1.22.1
- Knative Version: knative-serving v1.12.4, knative-eventing v1.12.6
- KServe Version: 0.14.0
- Kubeflow version: v1.9.0
- Cloud Environment: GKE on version 1.30.4
- Kubernetes version: (use `kubectl version`): Client Version: v1.30.5-dispatcher / Server Version: v1.30.4-gke.1348001

FabianScheidt:

This PR extends the KServe protocol to support an [OpenAI-compatible endpoint for text embeddings](https://platform.openai.com/docs/api-reference/embeddings). It adds a simple implementation to the HuggingFace serving runtime, which essentially wraps the existing implementation for the open inference protocol. 

Fixes #4025 

**Type of changes**

Embeddings models can be served using the HuggingFace Serving Runtime. E.g., like this

```bash
python -m huggingfaceserver \
    --model_name=embedding \
    --model_id=mixedbread-ai/mxbai-embed-large-v1 \
    --task=text_embedding
```

Previously, it was only possible to call the embedding model using the open inference protocol. We're retaining this functionality:

```bash
curl -H "Content-Type: application/json" \
    "http://localhost:8080/v1/models/embedding:predict" \
    -d '{"instances": ["Hello World"]}' | jq
```

Now, we're introducing an OpenAI-compatible endpoint:

```bash
curl http://localhost:8080/openai/v1/embeddings \
  -H "Content-Type: application/json" \
  -d '{
    "input": ["Hello World"],
    "model": "embedding",
    "encoding_format": "float"
  }' | jq
```

As a result, we're able to call the endpoint with [OpenAI's python library](https://pypi.org/project/openai/):

```python
from openai import OpenAI

client = OpenAI(base_url='http://localhost:8080/openai/v1', api_key='no-api-key-needed')
res = client.embeddings.create(model="embedding", input=["Hello World"])
print(res)
```

This yields the following output:

```
CreateEmbeddingResponse(data=[Embedding(embedding=[0.015021147206425667, 0.020543351769447327, ...], index=0, object='embedding')], model='embedding', object='list', usage=Usage(prompt_tokens=4, total_tokens=4))
```

This PR includes lots of boilerplate changes to the OpenAI data plane. In essence, we're separating the `OpenAIModel` class into `OpenAICompletionModel` and `OpenAIEmbeddingModel`. We're using `OpenAICompletionModel` in all the places where we previously had `OpenAIModel`. 

The actual implementation of the OpenAI-support in the HuggingFace serving runtime is simple: We're inheriting from `OpenAIEmbeddingModel` and implement a method to wrap the existing implementation.

**Checklist**:

I've added tests for the model registration. Not sure what other tests would be beneficial. Happy to make changes if you have suggestions. In case this is merged, we should add a remark to the [API endpoints section of the documentation](https://kserve.github.io/website/latest/modelserving/v1beta1/llm/huggingface/#api-endpoints).

**Release note**:

```release-note
Add Support for OpenAI-compatible Embeddings API in HuggingFace Serving Runtime
```


mginfn:
Hi @FabianScheidt, thanks for this work!
In my case, I'm implementing a fully-hosted QA solution and I need OpenAI-compatible Embeddings API.
Are there any plans to review/approve this PR?
Thanks

yuzisun:
> Hi @FabianScheidt, thanks for this work! In my case, I'm implementing a fully-hosted QA solution and I need OpenAI-compatible Embeddings API. Are there any plans to review/approve this PR? Thanks

@Datta0 @gavrissh Can you help review this?

python/huggingfaceserver/huggingfaceserver/encoder_model.py
```diff
@@ -340,3 +349,60 @@ def _log_request(self, request_id: str, prompt: list[str]) -> None:
                 request_id,
                 prompt=prompt,
             )
+
+    async def create_embedding(self, request: EmbeddingRequest) -> Embedding:
+        params = request.params
+        if params.input is None:
+            raise ValueError("input is required")
```
gavrissh: Let's use `OpenAIError()`, to return OpenAI Error Response json

python/huggingfaceserver/huggingfaceserver/encoder_model.py
```diff
@@ -340,3 +349,60 @@ def _log_request(self, request_id: str, prompt: list[str]) -> None:
...
+
+        # The OpenAI documentation allows the input of token lists instead of strings. As the tokenization is specific
+        # to the model, it is most likely different from the ones used by OpenAI (e.g., tiktoken). Libraries like
+        # Llangchain attempt to determine the proper tokenization based on the model name and will fall back to the
```
gavrissh: nit: Langchain?

python/huggingfaceserver/huggingfaceserver/encoder_model.py
```diff
@@ -340,3 +349,60 @@ def _log_request(self, request_id: str, prompt: list[str]) -> None:
...
+        # accepting the mismatch, we rather raise an exception.
+        if isinstance(params.input, list) and len(params.input) > 0:
+            input_is_int = isinstance(params.input[0], int)
+            input_is_list_of_int = (
+                isinstance(params.input[0], list)
+                and len(params.input[0]) > 0
+                and isinstance(params.input[0][0], int)
+            )
+            if input_is_int or input_is_list_of_int:
+                raise NotImplementedError("Input of token lists is not supported.")
```
gavrissh: Use `OpenAIError()`

python/kserve/kserve/protocol/rest/openai/endpoints.py
```diff
@@ -137,6 +139,41 @@ async def stream_results() -> AsyncGenerator[str, None]:
...
+        self,
+        raw_request: Request,
+        request_body: CreateEmbeddingRequest,
+        response: Response,
+    ) -> Response:
+        """Create embedding handler.
+
+        Args:
+            raw_request (Request): fastapi request object,
+            model_name (str): Model name.
```
gavrissh: model_name is not an arg here?

python/kserve/kserve/protocol/dataplane.py
```diff
@@ -432,8 +432,8 @@ async def infer(
         # call model locally or remote model workers
         response_headers = {}
         model = await self.get_model(model_name)
-        if isinstance(model, OpenAIModel):
-            error_msg = f"Model {model_name} is of type OpenAIModel. It does not support the infer method."
+        if isinstance(model, OpenAICompletionModel):
```
gavrissh, line 435: Is this change required?

python/huggingfaceserver/huggingfaceserver/encoder_model.py
FabianScheidt: If you're referring to the typo in the library name, that's fixed now :) If you're referring to the comment itself: It was the second library I experimented with after OpenAI's official library and I guess it's fairly common to use. I want to make sure that other people don't stumble into weird errors...

python/kserve/kserve/protocol/dataplane.py
FabianScheidt: Yes, otherwise we'd break the current endpoints for v1 and v2 for no good reason.

python/kserve/kserve/protocol/rest/openai/endpoints.py
FabianScheidt: This was actually outdated across all doc strings in the file. I now have updated all of them.

python/huggingfaceserver/huggingfaceserver/encoder_model.py
```diff
@@ -340,3 +351,72 @@ def _log_request(self, request_id: str, prompt: list[str]) -> None:
...
+            )
+
+        # The OpenAI documentation allows the input of token lists instead of strings. As the tokenization is specific
+        # to the model, it is most likely different from the ones used by OpenAI (e.g., tiktoken). Libraries like
+        # LangChain attempt to determine the proper tokenization based on the model name and will fall back to the
+        # default "cl100k_base" tokenization, which will certainly not match the deployed model. Instead of silently
+        # accepting the mismatch, we rather raise an exception.
+        if isinstance(params.input, list) and len(params.input) > 0:
+            input_is_int = isinstance(params.input[0], int)
+            input_is_list_of_int = (
```
Datta0: From what I understand, [`List[List[Any]]`](https://github.com/kserve/kserve/blob/654d314f15af5a4ed5fd6114dbd2be1cfdc6d5a0/python/kserve/kserve/protocol/rest/openai/types/openapi.py#L773) is valid for this. So if we want to enforce that only string or list of strings are input, shouldn't we directly check the same instead of checking if its not `int/List[int]` ?
Something like {'input':[[{"1":"2"}],[{"3":"4"}]] } would be valid according to this check right?
And it'd throw error down the line at tokenization but we want to avoid that ig?

python/huggingfaceserver/huggingfaceserver/encoder_model.py
```diff
@@ -340,3 +351,72 @@ def _log_request(self, request_id: str, prompt: list[str]) -> None:
...
+
+        # Optionally encode result to base64
+        if params.encoding_format == "base64":
+            for i, o in enumerate(embedding_out):
+                embedding_bytes = [struct.pack("<f", el) for el in o]
+                embedding_base64 = base64.b64encode(b"".join(embedding_bytes))
+                embedding_out[i] = embedding_base64.decode("ascii")
+
+        # The autogenerated OpenAPI specification doesn't seem to include the base64 encoding option. Add it.
+        class EmbeddingObjectWithBase64Support(EmbeddingObject):
```
Datta0: NIT: Can we move it to [__init__.py](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/protocol/rest/openai/types/__init__.py) or [openai_model.py](https://github.com/kserve/kserve/blob/master/python/kserve/kserve/protocol/rest/openai/openai_model.py)

python/huggingfaceserver/huggingfaceserver/encoder_model.py
FabianScheidt: Good catch! I've adjusted the type validation to make use of pydantic, so it's more expressive. I have also adjusted the initial error message to match the one from the OpenAI API.

python/huggingfaceserver/huggingfaceserver/encoder_model.py
FabianScheidt: I have moved it to \_\_init.py__. As a result, we also no longer need to lengthy name, but can simply override the existing class name. Hope that's alright with you :)

Datta0 (approved):
LGTM!
Thanks for these changes :)

gavrissh (approved):
/lgtm
Thank you!

yuzisun:
@FabianScheidt Can you help fix the CI ? Let's also add the embedding unit test to `python/kserve/test/test_openai.py` and the e2e test https://github.com/kserve/kserve/blob/master/test/e2e/predictor/test_huggingface.py#L294

FabianScheidt:
> @FabianScheidt Can you help fix the CI ? Let's also add the embedding unit test to `python/kserve/test/test_openai.py` and the e2e test https://github.com/kserve/kserve/blob/master/test/e2e/predictor/test_huggingface.py#L294

Sure, done!

- The failing CI was caused by a newer language feature for type hints, which I've adjusted.
- I have created a unit test in a separate file and renamed the existing one to differentiate between *completion* and *embedding*. That way, I don't need to rename all occurrences of the mocked model class.
- I have added an e2e test which validates both output formats *float* and *base64*.

FabianScheidt:
Sorry, I didn't realize that there are numerical inaccuracies even if we're using the base64 encoding option. I've updated the e2e test to decode the response and match against the approximation. Now the tests should pass.

yuzisun:
@FabianScheidt Thanks for the awesome contributions! 
/lgtm
/approve

PR #4140, opened 2024-12-19
Clean up jobs in model cache agent

greenmoon55:
With the introduction of model cache agent, the download job is owned by the agent instead of the model cache CR, which means it would never get deleted once the model cache CR is deleted. 

In this PR, the agent filter jobs by the node label and then delete jobs if the model name on the job does not exist in the LocalModelNode spec.

Fixes #

**Type of changes**

Added a new test case.
Manually create a model cache CR and delete it. Then make sure the jobs is removed.

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -330,7 +330,7 @@ func (c *LocalModelNodeReconciler) cleanupJobs(ctx context.Context, localModelNo
 	}
 
 	// 3. Delete jobs that are not in the spec
-	for i, _ := range jobs.Items {
+	for i := range jobs.Items {
 		job := jobs.Items[i]
 		modelName, ok := job.Labels["model"]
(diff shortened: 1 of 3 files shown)


pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -314,6 +314,41 @@ func (c *LocalModelNodeReconciler) deleteModels(localModelNode v1alpha1api.Local
 	return nil
 }
 
+func (c *LocalModelNodeReconciler) cleanupJobs(ctx context.Context, localModelNode v1alpha1api.LocalModelNode) error {
+	// 1. Get all jobs for the LocalModelNode
+	jobs := &batchv1.JobList{}
+	labelSelector := map[string]string{"node": localModelNode.Name}
```
yuzisun, line 320: Do we need to select with the model name label as well?

PR #4184, opened 2025-01-13
Add hf to storageuri prefix list

tjandy98:
`hf://` storage uri type is already supported, but it is not whitelisted in the prefix list


spolti (approved):
/lgtm

gavrissh (approved):
/lgtm

Thanks!

PR #4186, opened 2025-01-13
Add option to disable volume management in localModel config

ritzdevp:

Adding a label check `disableVolumeManagement` in LocalModelCache that lets users disable volume management. This is to support the use case for when there is a requirement for creating LocalModelCache on a cluster that does not create inference services. This will also allow users to write their own logic for managing PV/PVCs for the cache irrespective of the existence of inference services.

Fixes #

**Type of changes**

Updated unit tests to support this feature. Refactored unit tests for local model controller.

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1alpha1/localmodel/controller.go
```diff
@@ -296,6 +300,10 @@ func (c *LocalModelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		c.Log.Error(err, "Create PVC err", "name", pv.Name)
 	}
 
+	disableReconcileForIsvcs, ok := localModel.Labels["disableReconcileForIsvcs"]
```
lizzzcai: not going into the feature itself, but the label name format can follow some of the conventions [here](https://github.com/kserve/kserve/blob/aad4e454e4a47fbda5a8db0e986d3f92f317f042/pkg/constants/constants.go).

pkg/controller/v1alpha1/localmodel/controller.go
```diff
@@ -296,6 +300,10 @@ func (c *LocalModelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		c.Log.Error(err, "Create PVC err", "name", pv.Name)
 	}
 
+	disableReconcileForIsvcs, ok := localModel.Labels[constants.DisableIsvcReconciliation]
```
yuzisun: I think this should be a configuration rather than label

ritzdevp: Updated. Added `disableVolumeManagement` as a flag in localModel configMap.

pkg/controller/v1alpha1/localmodel/controller.go
```diff
@@ -296,6 +300,10 @@ func (c *LocalModelReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		c.Log.Error(err, "Create PVC err", "name", pv.Name)
 	}
 
+	if localModelConfig.DisableVolumeManagement {
```
greenmoon55, line 303: maybe we should not watch inference service on L470 when this flag is enabled. I'm fine either way.

ritzdevp: Thanks for reviewing! Added a change to NOT watch for isvc if disableVolumeManagement is true in localmodel config. This is because the only reason we watch for isvc is to create pv/pvcs, if that step is not happening then watching isvcs becomes redundant. Thanks for the suggestion!

pkg/controller/v1alpha1/localmodel/controller_test.go
```diff
@@ -219,32 +196,9 @@ var _ = Describe("CachedModel controller", func() {
...
 		It("Should create pvs and pvcs for inference services", func() {
-			var configMap = &v1.ConfigMap{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      constants.InferenceServiceConfigMapName,
-					Namespace: constants.KServeNamespace,
-				},
-				Data: configs,
-			}
-			Expect(k8sClient.Create(context.TODO(), configMap)).NotTo(HaveOccurred())
+			configMap, clusterStorageContainer, nodeGroup := genericSetup(configs, clusterStorageContainerSpec, localModelNodeGroupSpec)
```
greenmoon55: Thank you for cleaning it up!

pkg/controller/v1alpha1/localmodel/controller_test.go
```diff
@@ -349,36 +303,131 @@ var _ = Describe("CachedModel controller", func() {
...
+				return err == nil && persistentVolume != nil
+			}, timeout, interval).Should(BeTrue())
+
+			persistentVolumeClaim := &v1.PersistentVolumeClaim{}
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, pvcLookupKey, persistentVolumeClaim)
+				return err == nil && persistentVolumeClaim != nil
+			}, timeout, interval).Should(BeTrue())
+		})
+		It("Should delete pvs and pvcs if localmodel config value DisableVolumeManagement does not exist", func() {
```
greenmoon55: nit: I think we already test this in L325. https://github.com/kserve/kserve/pull/4186/files#diff-a1dae2643d8f0e64fd1dddfb23a8076f6f92ef5711a55825b37522e373d55b69L325
Would it be better to remove one of them?

ritzdevp: Sure I agree, removed it
Thanks!

pkg/controller/v1alpha1/localmodel/controller.go
```diff
@@ -453,14 +467,17 @@ func (c *LocalModelReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		For(&v1alpha1api.LocalModelCache{}).
 		// Ownes PersistentVolumes and PersistentVolumeClaims that is created by this local model controller
 		Owns(&v1.PersistentVolume{}).
-		Owns(&v1.PersistentVolumeClaim{}).
-		// Creates or deletes pv/pvcs when isvcs got created or deleted
-		Watches(&v1beta1.InferenceService{}, handler.EnqueueRequestsFromMapFunc(c.isvcFunc), builder.WithPredicates(isvcPredicates)).
-		// Downloads models to new nodes
+		Owns(&v1.PersistentVolumeClaim{})
+
+	if !localModelConfig.DisableVolumeManagement {
```
ritzdevp, line 476: Only watch for isvc if disableVolumeManagement is NOT true

pkg/controller/v1alpha1/localmodel/suite_test.go
```diff
@@ -102,27 +95,6 @@ var _ = BeforeSuite(func() {
 	}
 	Expect(k8sClient.Create(context.Background(), kserveNamespaceObj)).Should(Succeed())
 	Expect(k8sClient.Create(context.Background(), jobsNamespaceObj)).Should(Succeed())
-
-	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
```
ritzdevp, line 106: Moved to controller_test.go as we need to initialize localmodelconfig with and without disableVolumeManagement

greenmoon55 (approved):

yuzisun (approved):
/lgtm
/approve

PR #4194, opened 2025-01-17
add DeploymentMode to InferenceService and InferenceGraph status and prevent deploymentMode change

### LINKED ISSUE #4191: Changing the defaultDeploymentMode retroactively changes the deploymentMode of existing InferenceServices
/kind bug

- Deploy Kserve with `defaultDeploymentMode = Serverless`
- Deploy a few InferenceServices without explicit annotation for deploymentMode
- Change `defaultDeploymentMode = RawDeployment` 
- Restart Kserve Controller
- All active InferenceServices have their deploymentMode changed from Serverless to RawDeployment

- Changing the `defaultDeploymentMode` does not affect existing InferenceServices. 

```
kind: "InferenceService"
metadata:
  name: "sklearn-v2-iris"
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      protocolVersion: v2
      runtime: kserve-sklearnserver
```

VedantMahabaleshwarkar:

Fixes #4191 

**Type of changes**

***Original Behavior***
- set `defaultDeploymentMode : Serverless`
- Deploy
```
kind: "InferenceService"
metadata:
  name: "sklearn-v2-iris-basic"
  annotations:
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      protocolVersion: v2
      runtime: kserve-sklearnserver
```
- verify successful deployment of isvc with Serverless mode (3/3 containers)
- change `defaultDeploymentMode : RawDeployment`
- restart `kserve-controller-manager`
- check that the InferenceService now has 2 deployments (one for serverless mode, one for rawdeployment mode) 

***New Behavior***
- set `defaultDeploymentMode : Serverless`
- Deploy
```
kind: "InferenceService"
metadata:
  name: "sklearn-v2-iris-basic"
  annotations:
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      protocolVersion: v2
      runtime: kserve-sklearnserver
```
- verify successful deployment of isvc with Serverless mode (3/3 containers)
- verify `isvc.Status.deploymentMode = Serverless`
- change `defaultDeploymentMode : RawDeployment`
- restart `kserve-controller-manager`
- verify that `isvc.Status.deploymentMode = Serverless` continues to be true
- check that the InferenceService still only has 1 deployment, corresponding to Serverless mode. 

**Checklist**:

**Release note**:

```release-note
added deploymentMode to inferenceservice and inferencegraph status
```

**Re-running failed tests**

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -82,8 +82,15 @@ func (v *InferenceServiceValidator) ValidateUpdate(ctx context.Context, oldObj, 
 		validatorLogger.Error(err, "Unable to convert object to InferenceService")
 		return nil, err
 	}
+	oldIsvc, err := convertToInferenceService(oldObj)
+	if err != nil {
+		validatorLogger.Error(err, "Unable to convert object to InferenceService")
+	}
 	validatorLogger.Info("validate update", "name", isvc.Name)
-
+	err = validateDeploymentMode(isvc, oldIsvc)
+	if err != nil {
+		return nil, err
+	}
 	return validateInferenceService(isvc)
 }
 
@@ -431,6 +438,19 @@ func validateCollocationStorageURI(predictorSpec PredictorSpec) error {
 				}
 			}
 			break
+		}
+	}
+	return nil
+}
+
+// validates if the deploymentMode specified in the annotation is not different from the one recorded in the status
+func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService) error {
+	statusDeploymentMode := oldIsvc.Status.DeploymentMode
+	if len(statusDeploymentMode) != 0 {
+		annotations := newIsvc.Annotations
+		annotationDeploymentMode, ok := annotations[constants.DeploymentMode]
+		if ok && annotationDeploymentMode != statusDeploymentMode {
+			return fmt.Errorf("update rejected: deploymentMode cannot be changed from '%s' to '%s'", statusDeploymentMode, annotationDeploymentMode)
 		}
 	}
 	return nil
--- a/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
@@ -21,6 +21,9 @@ import (
 	"fmt"
 	"time"
 
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	"github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/components"
+	otelv1beta1 "github.com/open-telemetry/opentelemetry-operator/apis/v1beta1"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/kserve/kserve/pkg/utils"
@@ -567,6 +570,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -1117,6 +1121,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -1645,6 +1650,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -2021,6 +2027,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -2590,6 +2597,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -3425,6 +3433,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
@@ -4369,6 +4378,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 					TransitionStatus:    "InProgress",
 					ModelRevisionStates: &v1beta1.ModelRevisionStates{TargetModelState: "Pending"},
 				},
+				DeploymentMode: string(constants.RawDeployment),
 			}
 			Eventually(func() string {
 				isvc := &v1beta1.InferenceService{}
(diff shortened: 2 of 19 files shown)


terrytangyuan (commented):
@yuzisun Is the current behavior documented in https://github.com/kserve/kserve/issues/4191 intentional? 

@VedantMahabaleshwarkar If we change the behavior, we'll lose the capability to change all ISVC's deployment mode at once, right?

VedantMahabaleshwarkar:
> If we change the behavior, we'll lose the capability to change all ISVC's deployment mode at once, right?
 
 @terrytangyuan Yes, however in my opinion this behavior is unintentional. When the deploymentMode of an isvc is changed, the old deployment is not removed. In such a case all those isvcs will have duplicate deployments (one corresponding to each deploymentMode)

terrytangyuan:
Thanks. I agree that the existing behavior is not ideal. This seems like a good approach then. Any thoughts? @yuzisun @sivanantha321

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -173,12 +173,18 @@ case 2: serving.kserve.org/deploymentMode is set
...
 */
-func GetDeploymentMode(annotations map[string]string, deployConfig *v1beta1.DeployConfig) constants.DeploymentModeType {
+func GetDeploymentMode(statusDeploymentMode string, annotations map[string]string, deployConfig *v1beta1.DeployConfig) constants.DeploymentModeType {
+	// First priority for deploymentMode is explicit request via annotation
 	deploymentMode, ok := annotations[constants.DeploymentMode]
 	if ok && (deploymentMode == string(constants.RawDeployment) || deploymentMode ==
 		string(constants.Serverless) || deploymentMode == string(constants.ModelMeshDeployment)) {
 		return constants.DeploymentModeType(deploymentMode)
 	}
+	// Second priority (to preserve original deploymentMode used when InferenceService was deployed)
```
israel-hdez: I wonder if the value in the `status` should be the first priority??

On PR description, you mention that by changing the default mode, KServe won't do cleanup. So, I may think that manually switching with the annotation from Serverless to Raw mode may also not clean the no longer needed resources for Serverless mode. If this is the case, it may be better to not allow switching the deployment mode (by honoring what was recorded in the `status`). KServe may later improve to do proper clean-up and the explicit annotation can have priority.

This would be my only comment. But, otherwise the approach here is looking good.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -173,12 +173,18 @@ case 2: serving.kserve.org/deploymentMode is set
 	        if the mode is "RawDeployment", "Serverless" or "ModelMesh", return it.
 			else return config.deploy.defaultDeploymentMode
 */
-func GetDeploymentMode(annotations map[string]string, deployConfig *v1beta1.DeployConfig) constants.DeploymentModeType {
+func GetDeploymentMode(statusDeploymentMode string, annotations map[string]string, deployConfig *v1beta1.DeployConfig) constants.DeploymentModeType {
+	// First priority for deploymentMode is explicit request via annotation
```
sivanantha321: If the user changes the deployment mode using the annotation, will it still result in duplicate deployments ?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
VedantMahabaleshwarkar: agreed, changed the priority accordingly

pkg/controller/v1beta1/inferenceservice/utils/utils.go
VedantMahabaleshwarkar: https://github.com/kserve/kserve/pull/4194#discussion_r1925855888 
according to this comment, this PR will not allow changing the deploymentMode after initial creation (since that results in uncleaned duplicate resources)

VedantMahabaleshwarkar:
I also added logic to the validating webhook to prevent an isvc to change it's deployment mode after creation.

yuzisun:
> Thanks. I agree that the existing behavior is not ideal. This seems like a good approach then. Any thoughts? @yuzisun @sivanantha321

The approach looks good, thanks @VedantMahabaleshwarkar for addressing this.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -353,6 +360,19 @@ func validateCollocationStorageURI(predictorSpec PredictorSpec) error {
 	return nil
 }
 
+// validates if the deploymentMode specified in the annotation is not different from the one recorded in the status
+func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService) error {
+	statusDeploymentMode := oldIsvc.Status.DeploymentMode
+	if len(statusDeploymentMode) != 0 {
+		annotations := newIsvc.Annotations
+		annotationDeploymentMode, ok := annotations[constants.DeploymentMode]
```
yuzisun, line 451: We also need to check the configmap, deployment mode annotation is not required. If the annotation is not specified it gets the default deployment mode from configmap.

VedantMahabaleshwarkar: If the isvc does not have a deployment mode annotation, the default would be inferred from the configmap [here](https://github.com/kserve/kserve/pull/4194/files#diff-57baeaa559989f6eb9da5ebbc7d34374151c8f8c9cdc26be56595dae29e0c760R190) and it would be [recorded](https://github.com/kserve/kserve/pull/4194/files#diff-f14605793ed521dabc5c4794584c2961365d2d8324ad41f009aa44b1513d50feR292) in the status. 

The webhook is to ensure that this isvc cannot be modified later to add deployment mode annotations that would contradict the mode recorded in the status when the isvc was created. 

Please let me know if I'm missing something

Jooho: From user perspective, what is the right way to change deploymentMode?

Jooho: For example,  some users want to change deloymentMode of existing isvc—something like from serverless to raw or raw to serverless. With current change, `status.deploymentMode` is higher priority if it is different from current annotation. To me , the only way to change it is recreating isvc.

How about adding annotations unconditionally when creating isvc? So that when changing the deployment mode of isvc, the user can easily change it by deleting or changing the annotation. Is it really necessary to add the deploymentMode of status?

VedantMahabaleshwarkar: @Jooho Kserve does not correctly change the deploymentMode even if the user tries to do it via annotation(resources for the new deploymentMode are created but the resources for the old mode are not deleted). Thus, according to @israel-hdez's [suggestion](https://github.com/kserve/kserve/pull/4194#discussion_r1925855888) I added logic so that the deploymentMode of an isvc cannot be changed after creation.

VedantMahabaleshwarkar: > From user perspective, what is the right way to change deploymentMode?

IMO, first the bugs related to changing the deploymentMode should be fixed and then changing deploymentMode via annotation can be allowed again.

spolti: >  added logic so that the deploymentMode of an isvc cannot be changed after creation.


+1 on this.
it should be the correct way, serverless and raw depend and install different resources, by creating a new one we can ensure that happens.

Jooho: OK then recreating isvc would be right way to change mode when people change default `deploymentMode`. If we agree with it, I am ok with this way.

israel-hdez: > We also need to check the configmap, deployment mode annotation is not required.

@yuzisun I think @VedantMahabaleshwarkar is right. Given this new func is invoked only in `ValidateUpdate`, it should be expected that the deploymentMode is already recorded in the status of the ISVC.

yuzisun:
@VedantMahabaleshwarkar can you help fix the generated diff ?

pkg/apis/serving/v1beta1/inference_service_validation.go
yuzisun: What happens if the status is not reconciled due to other validations? this blocks making other changes from the isvc spec.

sivanantha321:
@VedantMahabaleshwarkar Can you help resolve the conflicts ?

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -576,7 +577,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 				if err := k8sClient.Get(context.TODO(), serviceKey, isvc); err != nil {
 					return err.Error()
 				}
-				return cmp.Diff(&expectedIsvcStatus, &isvc.Status, cmpopts.IgnoreTypes(apis.VolatileTime{}))
+				return cmp.Diff(&expectedIsvcStatus, &isvc.Status, cmpopts.IgnoreTypes(apis.VolatileTime{}), cmpopts.IgnoreFields(v1beta1.InferenceServiceStatus{}, "DeploymentMode"))
```
yuzisun: Why do we need to ignore the deploymentMode field? it should be in the expected isvc status

pkg/apis/serving/v1beta1/inference_service_validation.go
israel-hdez: IIUC, no validation will be performed here, because of the `if len(statusDeploymentMode) != 0`

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
israel-hdez: I've fixed in the replacement PR #4423

spolti: If it this PR was replaced, can we close this one?

VedantMahabaleshwarkar:
replaced by #4423

PR #4196, opened 2025-01-20
keep replicas when autoscaler set external

Jooho:
When autoscaler set external, we should not set nil for replica in deployment. 
If we set nil, it will be set 1 always when isvc is updated.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


spolti (commented):
can we have a simple unit test for this?

Jooho:
@spolti, regarding unit tests for the hack file, I’m not entirely sure if they’re necessary in this case. The change primarily checks whether GOBIN is defined, and if it is, it gives it priority over GOPATH/bin and uses its value.

Let me know your thoughts!

https://go.dev/doc/code
```
The install directory is controlled by the GOPATH and GOBIN [environment variables](https://go.dev/cmd/go/#hdr-Environment_variables). If GOBIN is set, binaries are installed to that directory. If GOPATH is set, binaries are installed to the bin subdirectory of the first directory in the GOPATH list. Otherwise, binaries are installed to the bin subdirectory of the default GOPATH ($HOME/go or %USERPROFILE%\go).
```

```
GOBIN
	The directory where 'go install' will install a command.
```

Jooho:
@spolti added unit tests

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/go-logr/logr v1.4.2
 	github.com/go-logr/zapr v1.3.0
 	github.com/gofrs/uuid/v5 v5.3.0
+	github.com/golang/mock v1.4.4
```
spolti: looks like `go mod tidy` is needed.

spolti (commented):
/lgtm

code looks good, however, the `verify-go-mod` GHA is complaining.

go.mod
Jooho: updated.

PR #4197, opened 2025-01-20
Enhance multinode health_check python and manifests

### LINKED ISSUE #4067: Enhance vLLM multinode ServingRuntime
/kind feature

Multi-node servingruntime is working fine with deploying multi-node model.
However, when we think about more use cases, it causes some issues.
This should supports the following use-cases:
- rollingUpgrade 
- auto healing when head is killed
- auto healing when worker is killed

Jooho:
This PR enhance Multi node manifest to work properly for rollingUpdate.
Moreover, health check python script is enhanced to check service properly.

Fixes https://github.com/kserve/kserve/issues/4067

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


Jooho:
this need final test with gpu cluster but I could not have the cluster yet. I will test it ASAP.

Jooho:
I did final test and it works as expected.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

israel-hdez (approved):

python/huggingfaceserver/health_check.py
```diff
@@ -13,41 +13,88 @@
...
-        ray.init(address="auto")
-        return "Ray initialized"
-    else:
-        return "Ray already initialized"
+def initialize_ray_cluster(ray_address="auto"):
+    try:
+        if ray.is_initialized():  # Check if Ray is already initialized
+            return
+        else:
+            ray.init(address=ray_address)
```
spolti, line 30: Is it a good idea ti initialize the ray cluster in a health_check script?

spolti (approved):
minor comment, otherwise looks good.

python/huggingfaceserver/health_check.py
Jooho: initialize is for connection to the head node. it is not starting a new ray cluster

spolti: yeah, just realized that ray.init with no parameters would try to start a new one :)
Thanks.

config/runtimes/kserve-huggingfaceserver-multinode.yaml
```diff
@@ -16,35 +16,33 @@ spec:
...
           if [[ ! -z ${MODEL_DIR} ]]
           then
-            MODEL=${MODEL_DIR}
+            export MODEL=${MODEL_DIR}
           fi
 
-          python3 -m huggingfaceserver --model_name=${MODEL_NAME}  --model_dir=${MODEL} --tensor-parallel-size=${TENSOR_PARALLEL_SIZE} --pipeline-parallel-size=${PIPELINE_PARALLEL_SIZE}
+          export RAY_ADDRESS=${POD_IP}:${RAY_PORT}
+          ray start --head --disable-usage-stats --include-dashboard false 
+          python ./huggingfaceserver/health_check.py registered_nodes --retries 200  --probe_name runtime_start
```
yuzisun, line 31: Why it needs 200 retries? some comments might be helpful

python/huggingfaceserver/health_check.py
```diff
@@ -13,41 +13,88 @@
...
+
+    if (
+        check_registered_nodes_status == "Healthy"
+        and check_runtime_health_status == "Healthy"
+    ):
         return "Healthy"
-    except Exception as e:
-        logger.error(f"Ray is NOT accessible: {e}")
+    else:
+        return "Unhealthy"
```
yuzisun, line 77: it is better to return boolean instead unless there is a third state

python/huggingfaceserver/health_check.py
```diff
@@ -13,41 +13,88 @@
...
+def check_registered_node_and_runtime_models(
+    pipeline_parallel_size, runtime_url, ray_address, isvc_name
+):
+    result1 = check_registered_nodes(pipeline_parallel_size, ray_address)
+    result2 = check_runtime_models(runtime_url, isvc_name)
+    logger.debug(f"check_registered_nodes: {result1}, check_runtime_models: {result2}")
+    # Check both results
+    if (result1 == "Healthy" or result1 is True) and (
+        result2 == "Healthy" or result2 is True
+    ):
```
yuzisun, line 89: I am confused about this condition, why it can return either Healthy or boolean ?

config/runtimes/kserve-huggingfaceserver-multinode.yaml
Jooho: Previously, there is no timeout so I added 200s by adding 200 retries. 
Basically, I don't think it would take more than 10~30 seconds but I would give enough time to wait.

python/huggingfaceserver/health_check.py
Jooho: Right. I was also worried about this, but it seemed like the method name and return value didn't match, so I did it like this, but I'll think about it some more and change it later.

python/huggingfaceserver/health_check.py
Jooho: This is for a similar reason to the above.

PR #4198, opened 2025-01-20
Fix golangci-lint binary path selection based on GOBIN

Jooho:
### Description:
This PR updates the script to dynamically select the golangci-lint binary path. It first checks if the `GOBIN` environment variable is set and uses that as the binary path. If `GOBIN` is not set, it falls back to using `GOPATH/bin`. This change ensures that the correct binary path is used depending on the user's environment configuration.

### Changes:
- Modified the script to check for `GOBIN` and use it if set.
- If `GOBIN` is not set, the script falls back to `GOPATH/bin`.

This fix helps improve compatibility with different Go environment setups and ensures that the correct golangci-lint binary is used during the linting process.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -18,7 +18,11 @@ set -o errexit
 set -o pipefail
 
 # golangci-lint binary path
+# Check if GOBIN is set, if not use GOPATH/bin
 golangci_lint_binary="$(go env GOPATH)/bin/golangci-lint"
+if [ -n "$(go env GOBIN)" ]; then
+    golangci_lint_binary="$(go env GOBIN)/golangci-lint"
+fi
 
 # Check if golangci-lint is already installed
 if ! command -v golangci-lint &> /dev/null; then


hack/verify-golint.sh
```diff
@@ -18,7 +18,11 @@ set -o errexit
 set -o pipefail
 
 # golangci-lint binary path
+# Check if GOBIN is set, if not use GOPATH/bin
 golangci_lint_binary="$(go env GOPATH)/bin/golangci-lint"
+if [ -n "$(go env GOBIN)" ]; then
+    golangci_lint_binary="$(go env GOBIN)/golangci-lint"
```
spolti, line 24: not sure if it is a good approach, by default the go_path structure should have bin, src and pkg directories.
GOBIN points to the golang binaries itself.

Jooho: By default, the GOPATH structure does include bin, src, and pkg directories, which is correct. However, GOBIN does not point to the Go binary itself (e.g., the go command). Instead, it specifies the directory where binaries installed via go install are placed.

If GOBIN is set, it takes precedence over the default GOPATH/bin for storing installed binaries. This behavior is explicitly documented in the Go documentation, and it allows users to customize the installation path for their binaries while keeping the Go toolchain itself separate. Using GOBIN in this way is a standard and recommended approach when a custom binary path is needed.

spolti: by `go binary iteself`, I meant the place where binaries are placed when you run `go install`, which is by default gobin :)
sry for the confusion.

spolti (approved):

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #4208, opened 2025-01-24
Fix service name more than 63

luohua13:

Fix service name more than 63, for example: when `len({isvc-name}+"-predictor")` is more than 63, will report an error.

**Type of changes**

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strings"
 
+	"knative.dev/pkg/kmeta"
 	"knative.dev/serving/pkg/apis/autoscaling"
 
 	"knative.dev/pkg/network"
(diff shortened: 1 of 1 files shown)


pkg/constants/constants.go
```diff
@@ -543,51 +544,51 @@ func InferenceServiceHostName(name string, namespace string, domain string) stri
 }
 
 func DefaultPredictorServiceName(name string) string {
-	return name + "-" + string(Predictor) + "-" + InferenceServiceDefault
+	return kmeta.ChildName(name, "-"+string(Predictor)+"-"+InferenceServiceDefault)
 }
 
 func PredictorServiceName(name string) string {
-	return name + "-" + string(Predictor)
+	return kmeta.ChildName(name, "-"+string(Predictor))
```
sivanantha321, line 551: Linking the doc for reference.
```go
// ChildName generates a name for the resource based upon the parent resource and suffix.
// If the concatenated name is longer than K8s permits the name is hashed and truncated to permit
// construction of the resource, but still keeps it unique.
// If the suffix itself is longer than 31 characters, then the whole string will be hashed
// and `parent|hash|suffix` will be returned, where parent and suffix will be trimmed to
// fit (prefix of parent at most of length 31, and prefix of suffix at most length 30).
```

sivanantha321:
@luohua13 Can you add a test for this ?

Jooho:
I wonder if this impact existing services? what if user already have a isvc that have more than 63 length?

pkg/constants/constants.go
mturley: > when len({isvc-name}+"-predictor") is more than 63, will report an error

If I understand the ChildName doc correctly, is this really the behavior? It sounds like there will be no error and the service name will be truncated if necessary.

Jooho: good catch! Just took a look at the api and you are right, it returns string. no errors.
https://github.com/knative/pkg/blob/dcf159339de2/kmeta/names.go#L42

spolti:
> I wonder if this impact existing services? what if user already have a isvc that have more than 63 length?

k8s should prevent that, not sure if that would be possible.

pkg/constants/constants.go
spolti: we also have this class that could be helpful on validating the names
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go

Jooho:
> k8s should prevent that, not sure if that would be possible

The main problem is
- isvc can be created with more than 63 length
- hostname will be failed to create because of long length.
- if it is for internal access service, there is no hostname so i am worried this use-case

pkg/constants/constants.go
Jooho: A little bit research this function ChildName. it gives us unique name that is less than 63 but the string does not look right to me.
It should look like
```
superlongparentname6b3a55-e0d14b5a39f33a9d91b7da69
parent6b3a55e0d14b5a39f33a9d91b7da69-predictornameaverylongsuffixbeyondthirtycharacters
parentsuffix
```
do we really like this ?

I didn't execute sample code. just read the code (https://github.com/knative/pkg/blob/c8bea7c326ff/kmeta/names.go#L42)

spolti:
@luohua13 can you please rebase?

PR #4233, opened 2025-02-06
[CHART] Use correct secret in webhooks cert-manager.io/inject-ca-from

Nabsku:

When deploying KServe via the supplied Helm charts, the Webhooks do not work due to reading the wrong secret. This PR fixes this by using the correct secret, which is created by cert-manager. 

**Type of changes**

- [X] Deploy Helm Chart with more or less default values, Webhooks pull the correct secret.

- Logs

**Checklist**:

**Release note**:

```release-note
Fixed cert-manager.io/inject-ca-from annotation to properly reflect the name of the Secret that cert-manager generates.
```

**Re-running failed tests**

### DIFF
--- a/charts/kserve-resources/templates/webhookconfiguration.yaml
+++ b/charts/kserve-resources/templates/webhookconfiguration.yaml
@@ -4,7 +4,7 @@ kind: MutatingWebhookConfiguration
 metadata:
   name: inferenceservice.serving.kserve.io
   annotations:
-    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/serving-cert
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/kserve-webhook-server-cert
 webhooks:
   - clientConfig:
       service:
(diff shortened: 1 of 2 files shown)


charts/kserve-resources/templates/webhookconfiguration.yaml
```diff
@@ -4,7 +4,7 @@ kind: MutatingWebhookConfiguration
 metadata:
   name: inferenceservice.serving.kserve.io
   annotations:
-    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/serving-cert
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/kserve-webhook-server-cert
```
yuzisun, line 7: the name is correct, see https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/certificate.yaml#L4

yuzisun: It is the certificate resource name not the underlying secret name

yuzisun: for secret name cert manager support `cert-manager.io/inject-ca-from-secret`.
An injectable resource MUST have one of these annotations: cert-manager.io/inject-ca-from, cert-manager.io/inject-ca-from-secret, or cert-manager.io/inject-apiserver-ca, depending on the injection source. 
https://cert-manager.io/docs/concepts/ca-injector.

PR #4234, opened 2025-02-07
Better compatibility with in-place upgrades

israel-hdez:

For in-place upgrades the manifests of the new version are applied in a cluster where KServe may already be installed. The old version of the kserve-controller is terminated and the new version is deployed.

The default strategy is Rolling. This has the consequence that during an in-place upgrade, there will be two different versions of kserve-controller running in parallel. Most importantly, the two instances will be running with different configs, which include references to different versions of kserve-agent, kserve-router, storage-initializer, etc. During an in-place upgrade, the new version of the controller will upgrade the deployed models to use the new versions of the images. However, since the old version of kserve-controller is running in parallel, this old version will do rollback. The two instances will conflict with each other, leading to Deployments/KSVCs being changed very quickly which will lead to the cluster spawning pods without control.

Eventually, the cluster will terminate the old version of kserve-controller. However, because of the very, very quick updates to deployments the resources of the cluster can be exhausted leading to instability.

The leader election solves mentioned issues because the new version of the manager won't fully boot until it is capable of getting the lease for being the leader. The manifests enable leader election by default, because `manager_auth_proxy_patch.yaml` is used by default. However, the kustomization.yaml file gives the alternative to use `manager_prometheus_metrics_patch.yaml` and leader election won't be enabled in such case. This is enabling leader election in the manager, by default for all cases. 

N/A

**Type of changes**

I tested manually by modifying the deployment strategy in an existent setup. I observed the expected behavior about the old pod being fully terminated before the new one is created.

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -30,5 +30,7 @@ spec:
           name: https
       - name: manager
         args:
+        # When changing args, also update manager.yaml because the following
+        # list of args will fully override the arguments in manager.yaml.
         - "--metrics-addr=127.0.0.1:8080"
         - "--leader-elect"
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -27,6 +27,10 @@ spec:
       containers:
       - command:
         - /manager
+        args:
+        # When changing arguments, make sure to review the args in manager_auth_proxy_patch.yaml
+        # and update as needed.
+        - "--leader-elect"
         image: ko://github.com/kserve/kserve/cmd/manager
         imagePullPolicy: Always
         name: manager


yuzisun:
@israel-hdez I think we should actually turn on the HA mode to run multiple replicas, the leader election is already supported.

israel-hdez:
> @israel-hdez I think we should actually turn on the HA mode to run multiple replicas, the leader election is already supported.

Hmmmm... That would reduce the time when there is no controller running, right? If so @yuzisun, yes I agree that will be a more appropriate approach.

Let me do testing to ensure the commented conflicts will no longer happen and I will update the PR.

israel-hdez:
@yuzisun I updated the PR to use leader election. It works to solve the issue about the conflicting reconciles.

sivanantha321:
> @yuzisun I updated the PR to use leader election. It works to solve the issue about the conflicting reconciles.

It is already using leader election.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -31,4 +31,3 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
-        - "--leader-elect"
```
sivanantha321, line 34: This is a patch file. I think we can keep the manger args here or we can move both args to manager.yaml.

spolti: it is for the auto-proxy, right?

israel-hdez: Hey @sivanantha321. Thanks for your comment. You made me do a deeper review.

It turns out that we are being affected, because we don't use `kube-rbac-proxy`. So, we are following the comment here: https://github.com/kserve/kserve/blob/895f9fd944939f3646b940a16cefdeb2d5824e5a/config/default/kustomization.yaml#L194-L202.

So, by default, users that are using the manifests as is won't see any issue, because `manager_auth_proxy_patch.yaml` is included. Thus, leader elect will be enabled. In our case, since we have opted out from kube-rbac-proxy, we are not including `manager_auth_proxy_patch.yaml`. This also has the side effect of leader-elect being turned off.

So, I'm keeping changes in `manager.yaml` and added a comment there. Also, I'm reverting changes here in `manager_auth_proxy_patch.yaml` but also adding a comment.

With this, leader-elect should always be enabled despite opting out `manager_auth_proxy_patch.yaml`. The helm charts are, apparently, unaffected as I don't see any conditional around the leader-elect argument: https://github.com/kserve/kserve/blob/895f9fd944939f3646b940a16cefdeb2d5824e5a/charts/kserve-resources/templates/deployment.yaml#L88-L90. So, this should be better aligning kustomize manifests and helm charts.

sivanantha321: Nice catch!

Issue #4231, opened 2025-02-06
[Inquery] Default value of pathTemplate


DeepFlame-JR:
In https://github.com/kserve/kserve/blob/release-0.14/charts/kserve-resources/templates/configmap.yaml#L264,
we can expect that the `/serving/{{ .Namespace }}/{{ .Name }}` value is default value of pathTemplate.

But if the pathTemplate value is empty in `inferenceservice-config`, the pathTemplate is not applied.
I think it's related to the code below.
https://github.com/kserve/kserve/blob/release-0.14/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/path.go#L36

Inquery
1. Is the `/serving/{{ .Namespace }}/{{ .Name }}` default value of pathTemplate?
1. In helm chart, pathTemplate is in values.yaml. Can you add it? (https://github.com/kserve/kserve/blob/release-0.14/charts/kserve-resources/templates/configmap.yaml#L518)

lizzzcai:
Hi @DeepFlame-JR , from [here](https://github.com/kserve/kserve/blob/release-0.14/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go#L152), it looks like host-based routing is used when pathTemplate is empty, and use path-based routing when it is not empty. So I will say the default value of pathTemplate is `""`, as host-based routing is the default behaviour for KServe and KNative.

for 2. feel free to raise a PR to add it to the helm chart.

DeepFlame-JR:
Hello. @lizzzcai , I raised the PR https://github.com/kserve/kserve/pull/4237.

PR #4237, opened 2025-02-09
Add pathTemplate field to values.yaml for increased flexibility of helm chart

DeepFlame-JR:
- I needed to set the pathTemplate value while installing via helm chart, but it didn't exist.
- I just add the pathTemplate in helm chart.

- https://github.com/kserve/kserve/issues/4231

**Type of changes**

**Release note**:
```release-note
NONE
```


charts/kserve-resources/templates/configmap.yaml
```diff
@@ -273,7 +274,7 @@ data:
            # Namespace of the inference service ( {{- "{{ .Namespace }}" -}} )
            # For more info https://github.com/kserve/kserve/issues/2257.
            # NOTE: This configuration only applicable to serverless deployment.
-           "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}"
+           "pathTemplate": ""
```
israel-hdez: IMO this is just an example that shows how to use the placeholders, and can stay unchanged. The previous block where you added this field should be enough to show what is the default value.

charts/kserve-resources/values.yaml
```diff
@@ -154,6 +154,9 @@ kserve:
       # -- Ingress domain template for RawDeployment mode, for Serverless mode it is configured in Knative.
       domainTemplate: "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}"
 
+      # -- Path template for RawDeployment mode, for Serverless mode it is configured in Knative.
```
israel-hdez, line 157: I'm confused, because the comments of the `configmap.yaml` that you also changed tell this field is for Serverless. But, here, you say it is for Raw. Either, here is wrong, or `configmap.yaml` is wrong.

sivanantha321: @DeepFlame-JR path based routing is supported in both serverless and raw deployment (with gateway api).

charts/kserve-resources/templates/configmap.yaml
lizzzcai: I think can keep it as example for reference.

lizzzcai:
Hi @DeepFlame-JR can you sign off your commit

DeepFlame-JR:
@lizzzcai @israel-hdez @sivanantha321 
Can you please check the changes?

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -272,7 +273,7 @@ data:
            # Name of the inference service ( {{- "{{ .Name }}" -}} )
            # Namespace of the inference service ( {{- "{{ .Namespace }}" -}} )
            # For more info https://github.com/kserve/kserve/issues/2257.
-           # NOTE: This configuration only applicable to serverless deployment.
+           # If path template is empty the default path template is ""
```
sivanantha321, line 276: I think this line is confusing. It should be something like `If the path template is an empty string "" then path based routing is disabled.`

charts/kserve-resources/values.yaml
sivanantha321, line 157:
```suggestion
      # -- pathTemplate specifies the template for generating path based url for each inference service. If the value is an empty string, then path based routing will be disabled.
      
```

PR #4255, opened 2025-02-19
Collocation transformer and predictor spec

### LINKED ISSUE #3269: Collocate transformer and predictor for simplifing the deployment
/kind feature

While deploying transformer/predictor as separate services gives you the flexibility of running different replicas for transformer and predictor. KServe 0.11 already starts to support [collocating transformer and predictor](https://kserve.github.io/website/master/modelserving/v1beta1/transformer/collocation/) in the same pod for custom containers, we'd like to provide an option for collocating transformer with supported serving runtimes as well(e.g triton , TorchServe).

## Proposed Inference Service Spec Option 1
```yaml
kind: InferenceService
metadata:
  name: torch-transformer
  annotations:
      serving.kserve.io/collocate: true
spec:
  transformer:
    containers:
    - image: kserve/nlp-tokenizer:latest
      name: transformer-container
      command:
        - "python"
        - "-m"
        - "tokenizer"
      args:
        - --tokenizer
        - nlptokenizer
  predictor:
    model:
      modelFormat:
        name: pytorch
      storageUri: gs://kfserving-examples/models/nlp
```

## Proposed Inference Service Spec Option 2
```yaml
kind: InferenceService
metadata:
  name: torch-transformer
spec:
  predictor:
    model:
      modelFormat:
        name: pytorch
      storageUri: gs://kfserving-examples/models/nlp
    containers:
    - name: transformer-container
       args:
        - mapper
        - --task=embedding
        - --tokenizer=/mnt/models/triton/tokenizer_files/
        - predictor-config
        - --predictor-host=localhost:9000
        - model
        - --predictor-model-name=roberta-large-cased-bn-bfw-fp16
        - --name=benchmark-test
        - --return-bytes
```
## Benefits:
- Reduce networking overhead between node to node communications
- Reduce complexity of deploying transformer and predictor as separate services and tune the ratio for best performance
- Reduce the risk of non-compatible changes between transformer and predictor, now they can be deployed as a single unit.

## Other considerations:
* Compare the performance between collocating transformer and predictor vs launching as separate services
* We would also like to rename transformer as there is always confusion with the "Transformer Model" from google's paper.

sivanantha321:

Fixes #3269

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -67,6 +67,9 @@ type PredictorSpec struct {
 	PodSpec `json:",inline"`
 	// Component extension defines the deployment configurations for a predictor
 	ComponentExtensionSpec `json:",inline"`
+
+	// Enables collocation of transformer with predictor
+	Processors []corev1.Container `json:"processors,omitempty"`
```
yuzisun: How do we plan to merge the processor with the multiple containers defined on serving runtime ?

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -67,6 +67,9 @@ type PredictorSpec struct {
 	PodSpec `json:",inline"`
 	// Component extension defines the deployment configurations for a predictor
 	ComponentExtensionSpec `json:",inline"`
+
+	// Enables collocation of transformer with predictor
```
yuzisun:
```suggestion
	// Enables collocation of pre/post processors with predictor
```

pkg/apis/serving/v1beta1/predictor.go
yuzisun: As discussed with @sivanantha321 `containers` are already defined on PodSpec, we can use that as the additional processor containers to the model container (kserve-container)

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -59,11 +57,12 @@ type PredictorSpec struct {
...
-	// 1) Provide a full PodSpec for custom predictor.
-	// The field PodSpec.Containers is mutually exclusive with other predictors (i.e. TFServing). <br />
-	// 2) Provide a predictor (i.e. TFServing) and specify PodSpec
-	// overrides, you must not provide PodSpec.Containers in this case. <br />
+	// This spec serves three purposes. <br />
+	// 1) To provide a full PodSpec for a custom predictor.
+	//    The field PodSpec.Containers is mutually exclusive with other predictors (e.g., TFServing). <br />
+	// 2) To provide a predictor (e.g., TFServing) and specify PodSpec overrides. <br />
+	// 3) To provide a pre/post-processing container for a predictor. <br />
+	// You must not specify a predictor unless you are using a custom predictor.
```
yuzisun: What does this mean ?

pkg/apis/serving/v1beta1/predictor_custom.go
```diff
@@ -93,6 +93,10 @@ func (c *CustomPredictor) GetStorageSpec() *StorageSpec {
 func (c *CustomPredictor) GetContainer(metadata metav1.ObjectMeta, extensions *ComponentExtensionSpec, config *InferenceServicesConfig,
 	predictorHost ...string,
 ) *corev1.Container {
+	// If only one container is found, assume it is the predictor container
+	if len(c.Containers) == 1 {
```
yuzisun: I think this is equivalent to the following logic checking the kserve-container name.

pkg/apis/serving/v1beta1/predictor.go
sivanantha321: It means the user should not provide 'kserve-container' unless it is a custom predictor.

pkg/apis/serving/v1beta1/predictor_custom.go
sivanantha321: Previously we supported providing predictor container without name for custom predictor. So even if it does not have the name 'kserve-container' we assume it is a predictor container

yuzisun: I think we always sets the default container name 
https://github.com/kserve/kserve/blob/0a24d8669975c30a963f9de359cc0db462e64b38/pkg/apis/serving/v1beta1/predictor_custom.go#L69

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -198,51 +203,103 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+			if err = isvcutils.ReplacePlaceholders(&sRuntime.Containers[containerIndexInSR], isvc.ObjectMeta); err != nil {
+				isvc.Status.UpdateModelTransitionStatus(v1beta1.InvalidSpec, &v1beta1.FailureInfo{
+					Reason:  v1beta1.InvalidPredictorSpec,
+					Message: fmt.Sprintf(ErrInvalidPlaceholder, *isvc.Spec.Predictor.Model.Runtime, sRuntime.Containers[containerIndexInSR].Name),
+				})
+				return ctrl.Result{}, errors.Wrapf(err, ErrInvalidPlaceholder, *isvc.Spec.Predictor.Model.Runtime, sRuntime.Containers[containerIndexInSR].Name)
+			}
+		}
+		switch {
+		// Transformer container is present in both ServingRuntime and InferenceService
```
yuzisun: Can we add the test cases for these ?

yuzisun: this method is getting big, might want to break down to small testable functions

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -198,51 +203,103 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+		// Transformer container is present in InferenceService
+		case containerIndexInIS != -1:
+			// Append transformer container from InferenceService
+			podSpec.Containers = append(podSpec.Containers, isvc.Spec.Predictor.Containers[containerIndexInIS])
+			// Append all containers except the transformer container from InferenceService
+			podSpec.Containers = append(podSpec.Containers, isvc.Spec.Predictor.Containers[:containerIndexInIS]...)
+			podSpec.Containers = append(podSpec.Containers, isvc.Spec.Predictor.Containers[containerIndexInIS+1:]...)
+			// Append all containers except predictor container from ServingRuntime
+			podSpec.Containers = append(podSpec.Containers, sRuntime.Containers[:kserveContainerIdx]...)
+			podSpec.Containers = append(podSpec.Containers, sRuntime.Containers[kserveContainerIdx+1:]...)
```
yuzisun: We can simplify this logic across all the cases by excluding the kserve and transformer container from both serving runtime and isvc and merge if exists in both, then add the rest of the containers.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
sivanantha321: > Can we add the test cases for these ?

test cases are already present in the controller_test.go

sivanantha321: > this method is getting big, might want to break down to small testable functions

sure

pkg/apis/serving/v1beta1/predictor.go
yuzisun: Is there a validation if kserve-container is specified when using the model spec?

pkg/apis/serving/v1beta1/predictor_custom.go
sivanantha321: It wasn't working when I tested it. Let me recheck

pkg/apis/serving/v1beta1/predictor.go
yuzisun:
```suggestion
	// You must not specify kserve-container on pod spec unless you are using a custom predictor.
```

sivanantha321: Yes. [This](https://github.com/kserve/kserve/pull/4255/files#diff-4f94af97ce7181059eb881e5d3ad65043b99878eceb917d7877e80ca0570ad3eR140) will produce more than one implementation and it will make the validation fail in [here](https://github.com/kserve/kserve/blob/88c98d92132ea09d1c86c02b243503d32cd00db6/pkg/apis/serving/v1beta1/inference_service_validation.go#L128). I think it is missing a test case, will add it shortly.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -172,41 +179,3 @@ func (p *PredictorExtensionSpec) GetStorageUri() *string {
 func (p *PredictorExtensionSpec) GetStorageSpec() *StorageSpec {
 	return p.Storage
 }
-
-// GetPredictorImplementations GetPredictor returns the implementation for the predictor
-func (s *PredictorSpec) GetPredictorImplementations() []ComponentImplementation {
```
yuzisun, line 177: @sivanantha321 this is no longer needed right?

yuzisun (approved):

pkg/apis/serving/v1beta1/predictor.go
sivanantha321: Yes. It is a dead code.

PR #4270, opened 2025-02-26
Merging kfp into kserve docs. 

mholder6:
Merging the [pipelines/components/kserve](https://github.com/kubeflow/pipelines/tree/master/components/kserve) directory here as it was unmaintained in pipelines


docs/kfp/requirements.txt
```diff
@@ -0,0 +1,2 @@
+kserve==0.12.0
```
spolti: we can probably update to 0.14.1, no?

mholder6: Yes we can, will update accordingly

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #4275, opened 2025-02-27
(Bug #4273) quick_install.sh failed to uninstall incomplete installation and has small syntax bug

### LINKED ISSUE #4273: quick_install.sh uninstall components in wrong sequence and has a small syntax bug
/kind bug

I noticed two flaws of quick_install.sh script.

1. In a clear install with quick_install.sh, I got the following small bash error which complains a small syntax bug.

```bash
$curl -s "https://raw.githubusercontent.com/kserve/kserve/release-0.14/hack/quick_install.sh" | bash

...
./quick_install.sh: line 133: [: =: unary operator expected
...

```
The error is caused by that an unset parameter `installKserve` is used in test command. 

2. If the quick_install.sh failed in mid of installation, I would like to redo the install on the same kube env, so I need to clean up the env first. When I use `quick_install.sh -u` to clean up the env, it always failed at the first few steps, couldn't completely clean up the resources like Istio or certmanager if they were installed in the pervious failed installation.

```bash
./quick_install.sh -u
release "kserve" uninstalled
release "kserve-crd" uninstalled
😀 Successfully uninstalled KServe
error: the server doesn't have a resource type "KnativeServing"
```
This is because the script install and uninstall the resources in a different order, like Knative is installed at the second latest step, but it is always uninstalled at the second place. 

I expect the uninstall works, and no syntax error in script.

**Environment:**

- Istio Version: any
- Knative Version: any
- KServe Version: any
- Kubeflow version: any
- Minikube/Kind version: kind v0.24.0 go1.22.6 darwin/amd64
- OS (e.g. from `/etc/os-release`): darwin 14.6.1

zozowell:

1. quick_install.sh has a small syntax bug.
...
./quick_install.sh: line 133: [: =: unary operator expected
...
The error is caused by that an unset parameter installKserve is used in test command.

2. The `./quick_install.sh -u` failed to uninstall an incomplete installation. This is because the script install and uninstall the resources in a different order, like Knative is installed at the second latest step, but it is always uninstalled at the second place.

Fixes #4273 

**Type of changes**

- [x] Ran a clean installation, it ran clean without the syntax bug anymore.
- [x] Tested `quick_install.sh -u` with an incomplete installation, it cleanup all remanent components on the test env.

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


spolti (approved):

hack/quick_install.sh
```diff
@@ -115,7 +115,7 @@ helm install \
 echo "😀 Successfully installed Cert Manager"
 
 # Install Knative
-if [ $deploymentMode = "Serverless" ]; then
+if [ "${deploymentMode}" = "Serverless" ]; then
```
Jooho, line 118: To me, the original one is ok except the `=`.
Could you please check this? 
```suggestion
if [ "${deploymentMode}" == "Serverless" ]; then
```

hack/quick_install.sh
```diff
@@ -134,7 +134,7 @@ EOF
    echo "😀 Successfully installed Knative"
 fi
 
-if [ $installKserve = false ]; then
+if [ "${installKserve}" = false ]; then
```
Jooho, line 137: In bash, it should always check the string.
```suggestion
if [ "${installKserve}" == "false" ]; then
```

Jooho (changes requested):
I think the main root cause is the condition `=` .

hack/quick_install.sh
zozowell: @Jooho, in bash test, `=` is used as string comparison operator, and it's safe for both `[ ]` and `[[ ]]` test command. In contrast `==` is not safe for POSIX standard bash. The root cause here is that the lhs operand is not set, so the command is expanded to : `if [ = "Severless"]`.  See the test:

```bash
bash-3.2$ if [ $deploymentMode = "Serverless" ]; then echo "yes"; else echo "no"; fi
bash: [: =: unary operator expected
no
bash-3.2$ if [ $deploymentMode == "Serverless" ]; then echo "yes"; else echo "no"; fi
bash: [: ==: unary operator expected
no
bash-3.2$ export deploymentMode=local
bash-3.2$ if [ $deploymentMode = "Serverless" ]; then echo "yes"; else echo "no"; fi
no
bash-3.2$ unset deploymentMode
bash-3.2$ echo $deploymentMode

bash-3.2$ if [ "${deploymentMode}" = "Serverless" ]; then echo "yes"; else echo "no"; fi
no
```

hack/quick_install.sh
zozowell: same as above.

sivanantha321:
@zozowell can you help sign off the commits ? The DCO check is failing

zozowell:
@sivanantha321 thanks for reviewing this PR! I have added the sign-off and rebased the branch.

sivanantha321 (approved):

sivanantha321:
Thanks!
/lgtm
/approve

Issue #4229, opened 2025-02-05
Allow MultiNode ServingRuntime to add options through args


Jooho:
/kind feature

**Describe the solution you'd like**
Currently, MultiNode ServingRuntime executes the script using command + args, but we need to transition to using command instead.

**Reasons for Change:**
- User Flexibility: Users should be able to pass additional arguments via args, but the current setup prevents this.
- Standardization: Other ServingRuntime implementations already use command for execution, so aligning with this approach will improve consistency across runtimes.

hugginfaceserver
~~~
     args:
        - --model_name={{.Name}}
~~~
hugginfaceserver-multinode
~~~
      command: ["bash", "-c"]
      args:
        - |
          export MODEL=${MODEL_ID}
          if [[ ! -z ${MODEL_DIR} ]]
          then
            export MODEL=${MODEL_DIR}
          fi

          export RAY_ADDRESS=${POD_IP}:${RAY_PORT}
          ray start --head --disable-usage-stats --include-dashboard false 
          python ./huggingfaceserver/health_check.py registered_nodes --retries 200  --probe_name runtime_start

          python -m huggingfaceserver --model_name=${MODEL_NAME} --model_dir=${MODEL} --tensor-parallel-size=${TENSOR_PARALLEL_SIZE} --pipeline-parallel-size=${PIPELINE_PARALLEL_SIZE} 
~~~    

**Proposed Changes:**
- Modify MultiNode ServingRuntime to use command only for script execution instead of args.
- Ensure users can still pass custom arguments via args.

PR #4289, opened 2025-03-05
Move arguments from 'args' to 'command' for huggingface server multnode SR

Jooho:
This changes multinode servingruntime to accept a new args. It used args to run the servingruntime but with this, it was not allowed to add new args by isvc.  So now it is using `command`  and allows to add new args by isvc.

Fixes #https://github.com/kserve/kserve/issues/4229

**Type of changes**

- [X] Test A
Setup:
- Deploy multi node servingruntime
- add a new arg by isvc
- check if head runtime pod has the isvc

servingruntime
```
  containers:
  - args:
    - --model_name={{ .Name }}
```

isvc
```
    predictor:
      model:
        args:
        - --max-batch-size=123
```

pod 
```
 containers:
  - args:
    - --model_name=vllm-llama3-8b
    - --max-batch-size=123
 ```

- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


spolti:
Which arg takes precedence? Let's say there is a default arg in the servingRuntime, but the user wants to override it with the isvc.

Jooho:
The default args(the below) should be updated by kserve so it is not allowed to override.
```
--model_dir=${MODEL} 
--tensor-parallel-size=${TENSOR_PARALLEL_SIZE} 
--pipeline-parallel-size=${PIPELINE_PARALLEL_SIZE}
```

Jooho:
if you are talking about the `model_name`, isvc args should be a high priority so it can overwrite but it is the same situation. it is not a good idea to overwrite the default args.

spolti:
I am not talking about default (internal) ones, but any arg that a user add to the servingRuntime and instead updating the serving runtime, he wants to override it in the isvc...

Jooho:
It should be managed by kserve but from my test, it is not working as expected. like others,  ISVC should be higher priority and it will override the same argument in ServingRunmtime. However, the function is not related to this PR though. it should be solved with another ticket.

spolti:
ok, maybe we can open the issue :)

/lgtm

spolti (approved):

yuzisun:
@Jooho Could you help add this runtime to the helm chart as well?

Jooho:
@yuzisun can I send the helm chart with another PR?

yuzisun (approved):

PR #4295, opened 2025-03-07
Keda integration external scaler

### LINKED ISSUE #4301: Add support to scale inference service using Keda External Scaler
/kind feature

Given that Keda is integrated with KServe, ideally we would like to have support for Keda External Scaler.
https://keda.sh/docs/2.16/concepts/external-scalers/

**Why?**
Keda by default uses a polling mechanism to monitor trigger sources, its typical response time is significantly higher. One option is setting the polling interval to a lower value, however the downside to this will be increasing load to kubenetes api or prometheus.
https://keda.sh/docs/2.16/reference/scaledobject-spec/#pollinginterval

**How?**

Keda defines an interface that has to be implemented for external scaling. One of the ways this will be faster and close to real time would be because there are mechanisms that enables streaming of the state if scaling is required using `StreamIsActive`.
Keda can then get metrics immediately and decide the scale count.

While planning the implementation of Keda External Scaler, we wanted to bring in the community's input on how we envision the external scaler to be implemented. Should it be running as a separate global service along with controller? Should it be running as a service per inference service similar to ksvc.

There are pros and cons to each approach.

**Global Pros**
1. Centralized Management
2. Effective resource usage
3. Easier scaling as grpc service endpoint need not be managed as it is global.

**Global Cons**
1. Complex code logic as all the inference service's metrics has to be handled.
2. Performance bottlenecks can occur beyond a certain scale.

**Service per Inference Service Pros**
1. Dedicated scaler, fast response time.
2. Higher Fault toleraance.
3. Higher scope for Fine tuning to specific needs.

 **Service per Inference Service Cons**
1. Maintenance can be a pain.
2. ...

Also, we are also contemplating about providing scaling out of the box without keda or knative in the near future. So, this implementation can be used as a base for that. 

Please add your thoughts on it.

I will add the initial design doc and attach here for reference.

andyi2it:

Support KEDA external scaler by using OTEL addon to push the metrics to KEDA. This allows faster autoscaler and removes the dependency on external metrics backend like Prometheus or Graphite. 
https://github.com/kedify/otel-add-on

Fixes #4301 

**Type of changes**

- [X] Example isvc spec using KEDA and OTEL
```yaml
kind: "InferenceService"
metadata:
 name: "sklearn-v2-iris"
 annotations:
   serving.kserve.io/deploymentMode: "RawDeployment"
   serving.kserve.io/autoscalerClass: "keda"
spec:
 predictor:
   model:
     modelFormat:
       name: sklearn
     protocolVersion: v2
     runtime: kserve-sklearnserver
   minReplicas: 1
   maxReplicas: 5
   autoScaling:
      metrics:
        - type: PodMetric
          podmetric:
            metric:
              backend: "opentelemetry"
              metricNames: 
                - http_request_duration_seconds_count
              query: "avg(http_request_duration_seconds_count)"
            target:
              type: Value
              value: "4"
```
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


hack/quick_install.sh
```diff
@@ -127,8 +127,12 @@ if [ $installKeda = true ]; then
    helm repo add kedacore https://kedacore.github.io/charts
    helm install keda kedacore/keda --version ${KEDA_VERSION} --namespace keda --create-namespace --wait
    echo "😀 Successfully installed KEDA"
+
+   helm upgrade -i kedify-otel oci://ghcr.io/kedify/charts/otel-add-on --version=v0.0.5
```
houshengbo: I am glad you are testing and using this otel-add-on here :-)

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -105,6 +106,17 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
 						trigger.Metadata["namespace"] = metric.External.Metric.Namespace
 					}
 				}
+				if triggerType == "opentelemetry" {
+					trigger.Type = "external"
+					trigger.Metadata["clampMin"] = strconv.Itoa(int(minReplicas))
+					trigger.Metadata["clampMax"] = strconv.Itoa(int(maxReplicas))
+					trigger.Metadata["metricQuery"] = query
+					trigger.Metadata["targetValue"] = strconv.Itoa(targetValue)
```
houshengbo: I guess you have got almost all the keys for the metadata, except there is another one called `operationOverTime`. The value for this key will define, over certain time period, what operation we can leverage to aggregate the metrics.

For example, `operationOverTime: "avg"` means during the past 60 seconds for instance, we pick up the average value for all the same metrics.

`operationOverTime: "last_one"` means during the past 60 seconds for instance, we pick up the last metric record no matter how many we received.

There are other operation options for this key as well.
Suggest you can add it.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -294,6 +294,7 @@ func validateKEDAMetrics(metric ScaleMetric) error {
 
 // Validate of autoscaler KEDA metrics
 func validateKEDAMetricBackends(backend MetricsBackend) error {
+	println("---------------------------------------", backend)
```
yuzisun: remove the print line

python/otel-metrics-collector.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+# Install dependencies
+RUN pip install requests opentelemetry-api opentelemetry-sdk opentelemetry-exporter-otlp prometheus_client
+
+# Copy the script into the container
+WORKDIR /app
+COPY otel_metrics_collector.py .
+
+# Run the script
+CMD ["python", "otel_metrics_collector.py"]
```
yuzisun: not sure if we need this, Otel collector has a way to inject the sidecar to collect metrics
https://github.com/open-telemetry/opentelemetry-operator?tab=readme-ov-file#sidecar-injection

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -105,6 +106,17 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
 						trigger.Metadata["namespace"] = metric.External.Metric.Namespace
 					}
 				}
+				if triggerType == "opentelemetry" {
+					trigger.Type = "external"
```
yuzisun, line 136: here should create a separate Keda scaleTrigger

pkg/apis/serving/v1beta1/component.go
```diff
@@ -236,7 +236,7 @@ const (
 
 type MetricSource struct {
 	// MetricsBackend defines the scaling metric type watched by autoscaler
-	// possible values are prometheus, graphite.
+	// possible values are prometheus, graphite, opentelemetry.
```
yuzisun: We should also consider adding the PodMetricsSource type?

python/kserve/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
yuzisun: accidental change?

python/huggingfaceserver/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
yuzisun: this change is not relevant

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (o *OtelReconciler) Reconcile(ctx context.Context) error {
+	desired := o.OTelCollector
+	existing := &otelv1beta1.OpenTelemetryCollector{}
+	err := o.client.Get(ctx, types.NamespacedName{Name: desired.Name, Namespace: desired.Namespace}, existing)
+	if err != nil {
+		if apierr.IsNotFound(err) {
+			log.Info("Creating OTel Collector resource", "name", desired.Name)
+			if err := o.client.Create(ctx, desired); err != nil {
+				log.Error(err, "Failed to create predictor HttpRoute", "name", desired.Name)
```
yuzisun:
```suggestion
				log.Error(err, "Failed to create otel collector resource", "name", desired.Name)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -0,0 +1,167 @@
...
+	}, nil
+}
+
+func createOtelCollector(componentMeta metav1.ObjectMeta,
+	metric v1beta1.MetricsSpec,
+	otelConfig v1beta1.OtelCollectorConfig,
+) *otelv1beta1.OpenTelemetryCollector {
+	metricQuery := metric.External.Metric.Query
+
+	otelCollector := &otelv1beta1.OpenTelemetryCollector{
```
yuzisun, line 73: The main difference between inference service Otel collector config is the metrics query right?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -197,6 +203,18 @@ type ExternalMetricSource struct {
 	Target MetricTarget `json:"target,omitempty"`
 }
 
+// PodsMetricSource indicates how to scale on a metric describing each pod in
+// the current scale target (for example, transactions-processed-per-second).
+// The values will be averaged together before being compared to the target
+// value.
+type PodsMetricSource struct {
+	// metric identifies the target metric by name and selector
+	Metric MetricSource `json:"metric,omitempty"`
```
yuzisun: I think we might want a separate MetricSource type for PodMetrics, the backend for PodMetrics is opentelemetry while for ExtenralMetrics it is prometheus/graphite.

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -100,9 +103,36 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
...
+					trigger.Metadata["namespace"] = metric.External.Metric.Namespace
+				}
+
+				triggers = append(triggers, trigger)
+
+			case v1beta1.MetricSourceType(constants.AutoScalerPodMetric):
+				otelConfig, _ := v1beta1.NewOtelCollectorConfig(configMap)
+				OTelScalerEndpoint := otelConfig.OTelScalerEndpoint
+
+				triggerType := string(*metric.External.Metric.Backend)
```
yuzisun: For the PodMetric case it should use the `metrics.PodMetric`

pkg/apis/serving/v1beta1/component.go
```diff
@@ -160,6 +160,12 @@ type MetricsSpec struct {
...
 	External *ExternalMetricSource `json:"external,omitempty"`
+
+	// pods refers to a metric describing each pod in the current scale target
+	// (for example, transactions-processed-per-second).  The values will be
+	// averaged together before being compared to the target value.
+	// +optional
+	Pods *PodsMetricSource `json:"podmetric,omitempty"`
 }
 
 // MetricSourceType indicates the type of metric.
```
yuzisun, line 171: Here it is missing PodMetric

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -100,9 +103,36 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
...
+				query := metric.External.Metric.Query
+				targetValue = int(metric.External.Target.Value.AsApproximateFloat64())
+
+				trigger := kedav1alpha1.ScaleTriggers{
+					Metadata: map[string]string{},
+				}
+
+				if triggerType == "opentelemetry" {
```
yuzisun: use the constant for opentelemetry

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -100,9 +103,36 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
...
+					trigger.Type = "external"
+					trigger.Metadata = map[string]string{
```
yuzisun, line 137: Add a comment where these metadata are coming from 
https://github.com/kedify/otel-add-on/blob/b01a40125ab930c22576ff04f57b6572c3a6a81d/util/handler_specific.go#L14

pkg/apis/serving/v1beta1/component.go
```diff
@@ -248,6 +266,9 @@ type MetricSource struct {
 	// For namespaced query
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
+	// OperationOverTime specifies the operation to aggregate the metrics over time
+	// +optional
+	OperationOverTime string `json:"operationOverTime,omitempty"`
```
yuzisun, line 292: It is not clear what are the possible values here, what's the default operation (last or aggregate  over time)

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -0,0 +1,167 @@
...
+				Receivers: otelv1beta1.AnyConfig{Object: map[string]interface{}{
+					"prometheus": map[string]interface{}{
+						"config": map[string]interface{}{
+							"scrape_configs": []interface{}{
+								map[string]interface{}{
+									"job_name":        "otel-collector",
+									"scrape_interval": otelConfig.ScrapeInterval,
+									"static_configs": []interface{}{
+										map[string]interface{}{
+											"targets": []interface{}{"localhost:8080"},
```
yuzisun: this is the metrics endpoint right? it may not be always 8080, we need to get it from serving runtime custom resource.
https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/clusterservingruntimes.yaml#L337

python/kserve/poetry.lock
andyi2it: The pre-commit hook, forced this change. Not sure why.

pkg/apis/serving/v1beta1/component.go
```diff
@@ -248,6 +270,30 @@ type MetricSource struct {
...
 	Namespace string `json:"namespace,omitempty"`
+	// OperationOverTime specifies the operation to aggregate the metrics over time
+	// possible values are last_one, avg, max, min, rate, count. Default is 'last_one'.
+	// +optional
+	OperationOverTime string `json:"operationOverTime,omitempty"`
+}
+
+type PodsMetricSource struct {
+	// MetricsBackend defines the scaling metric type watched by autoscaler
+	// possible values are prometheus, graphite, opentelemetry.
```
yuzisun: For PodsMetricsSource, we do not need prometheus and graphite right? the only possible value is  opentelemetry

pkg/apis/serving/v1beta1/component.go
```diff
@@ -248,6 +270,30 @@ type MetricSource struct {
 	// For namespaced query
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
+	// OperationOverTime specifies the operation to aggregate the metrics over time
+	// possible values are last_one, avg, max, min, rate, count. Default is 'last_one'.
+	// +optional
+	OperationOverTime string `json:"operationOverTime,omitempty"`
```
yuzisun: This is only relevant to PodMetricsSource

pkg/apis/serving/v1beta1/component.go
yuzisun: Now you can remove opentelemetry, ExternalMetricSource means autoscaler get the metrics from external source like prometheus, graphite. For PodMetricsSource autoscaler gets directly from Pod.

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -100,9 +103,36 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
...
+
+				if triggerType == string(constants.AutoScalerMetricsPrometheus) && metric.External.Metric.Namespace != "" {
+					trigger.Metadata["namespace"] = metric.External.Metric.Namespace
+				}
+
+				triggers = append(triggers, trigger)
+
+			case v1beta1.MetricSourceType(constants.AutoScalerPodMetric):
+				otelConfig, _ := v1beta1.NewOtelCollectorConfig(configMap)
+				OTelScalerEndpoint := otelConfig.OTelScalerEndpoint
```
yuzisun: This can also come from the serverAddress defined from the `PodMetricsSource` if the global config is not configured.

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -0,0 +1,172 @@
...
+		metric:        metric,
+	}, nil
+}
+
+func createOtelCollector(componentMeta metav1.ObjectMeta,
+	metric v1beta1.MetricsSpec,
+	otelConfig v1beta1.OtelCollectorConfig,
+) *otelv1beta1.OpenTelemetryCollector {
+	metricName := metric.PodMetric.Metric.MetricName
+	port, ok := componentMeta.Annotations["prometheus.kserve.io/port"]
```
yuzisun, line 67: Most of time this is not defined on the isvc spec, it is defined on serving runtime
https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/clusterservingruntimes.yaml#L7

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -0,0 +1,172 @@
...
+							},
+						},
+					},
+				}},
+				Processors: &otelv1beta1.AnyConfig{Object: map[string]interface{}{
+					"filter/ottl": map[string]interface{}{
+						"error_mode": "ignore",
+						"metrics": map[string]interface{}{
+							"metric": []interface{}{
+								fmt.Sprintf(`name != "%s"`, metricName),
```
yuzisun: what happens if metricName is not specified on PodMetricsSource ?

python/huggingfaceserver/poetry.lock
yuzisun: I think local you have installed poetry 1.8.3 ?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -248,6 +270,30 @@ type MetricSource struct {
...
+	// MetricsBackend defines the scaling metric type watched by autoscaler
+	// possible values are prometheus, graphite, opentelemetry.
+	// +optional
+	Backend *PodsMetricsBackend `json:"backend,omitempty"`
+	// Address of MetricsBackend server.
+	// +optional
+	ServerAddress string `json:"serverAddress,omitempty"`
+	// MetricName is the name of the metric in the backend
+	// +optional
+	MetricName string `json:"metricName,omitempty"`
```
yuzisun: the query may have multiple metrics, also if this is left not specified then it filters all the metrics.

andyi2it: Have modified this to be a list

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
andyi2it: All the metrics from the pod will be pushed to otel collector.

pkg/apis/serving/v1beta1/component.go
```diff
@@ -197,6 +207,18 @@ type ExternalMetricSource struct {
...
+// PodMetricSource indicates how to scale on a metric describing each pod in
+// the current scale target (for example, transactions-processed-per-second).
+// The values will be averaged together before being compared to the target
+// value.
+type PodMetricSource struct {
+	// metric identifies the target metric by name and selector
+	Metric PodsMetricSource `json:"metric,omitempty"`
+
+	// target specifies the target value for the given metric
+	Target MetricTarget `json:"target" protobuf:"bytes,2,name=target"`
```
yuzisun: We do not support protobuf, I guess this is copied from kube?

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -100,9 +103,39 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
...
-						trigger.Metadata["namespace"] = metric.External.Metric.Namespace
+
+				if triggerType == string(constants.AutoScalerMetricsPrometheus) && metric.External.Metric.Namespace != "" {
+					trigger.Metadata["namespace"] = metric.External.Metric.Namespace
+				}
+
+				triggers = append(triggers, trigger)
+
+			case v1beta1.MetricSourceType(constants.AutoScalerPodMetric):
+				otelConfig, _ := v1beta1.NewOtelCollectorConfig(configMap)
```
yuzisun: Need to capture the error, if it fails to ummarshall then otelConfig is empty

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
yuzisun: discussed, this presents as it is after the serving runtime and isvc merge.

yuzisun (approved):

Issue #3768, opened 2024-07-01
RollingUpdate strategy is not effective for RawDeployment mode


Syulin7:
/kind bug

**What steps did you take and what happened:**
When updating the inferenceService, deployment's replicas are not set, the deployment replicas will initially be set to 1, and then the HPA (Horizontal Pod Autoscaler) will adjust the replicas to the minReplicas value.

For example, in the case below, where minReplicas is set to 4, even under the default update strategy(maxUnavailable: 25%), it would simultaneously delete 3 pods, rather than just one.
```
sklearn-iris-predictor   4/4     4            4           3d19h
sklearn-iris-predictor   4/1     4            4           3d19h
sklearn-iris-predictor   4/1     0            4           3d19h
sklearn-iris-predictor   4/1     0            4           3d19h
sklearn-iris-predictor   1/1     0            1           3d19h
sklearn-iris-predictor   1/1     1            1           3d19h
sklearn-iris-predictor   1/4     1            1           3d19h
sklearn-iris-predictor   1/4     1            1           3d19h
sklearn-iris-predictor   1/4     1            1           3d19h
sklearn-iris-predictor   1/4     1            1           3d19h
sklearn-iris-predictor   1/4     3            1           3d19h
sklearn-iris-predictor   2/4     3            2           3d19h
sklearn-iris-predictor   3/4     3            3           3d19h
sklearn-iris-predictor   3/4     3            3           3d19h
sklearn-iris-predictor   3/4     3            3           3d19h
sklearn-iris-predictor   3/4     3            3           3d19h
sklearn-iris-predictor   3/4     4            3           3d19h
sklearn-iris-predictor   4/4     4            4           3d19h
sklearn-iris-predictor   4/4     4            4           3d19h
sklearn-iris-predictor   3/4     4            3           3d19h
sklearn-iris-predictor   4/4     4            4           3d19h
```

**What did you expect to happen:**
When creating a deployment, set replicas=minReplicas.

**What's the InferenceService yaml:**
[To help us debug please run `kubectl get isvc $name -n $namespace -oyaml` and paste the output]
```
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: sklearn-iris
  namespace: default
spec:
  predictor:
    minReplicas: 4
    model:
      modelFormat:
        name: sklearn
      storageUri: gs://kfserving-examples/models/sklearn/1.0/model
```

**Anything else you would like to add:**
[...]

PR #4297, opened 2025-03-07
Envs removed from ISVC are not removed from deployment

spolti:
chore:  Fixes the reconcile loop to effectivelly update the EnvVars and replicas count.
        Also address the issue when the replicas are increased or decreased in the ISVC.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -230,13 +232,15 @@ func (r *DeploymentReconciler) checkDeploymentExist(ctx context.Context, client
 		log.Error(err, "Failed to perform dry-run update of deployment", "Deployment", deployment.Name)
 		return constants.CheckResultUnknown, nil, err
 	}
+
 	if diff, err := kmp.SafeDiff(deployment.Spec, existingDeployment.Spec, ignoreFields); err != nil {
+		log.Error(err, "Failed to diff deployments", "Deployment", deployment.Name)
```
terrytangyuan: Is this needed? err will be returned by the function so maybe it will be printed there

spolti: yeah,  not needed.

hack/verify-golint.sh
```diff
@@ -27,11 +27,11 @@ fi
 # Check if golangci-lint is already installed
 if ! command -v golangci-lint &> /dev/null; then
     echo "installing golangci-lint"
-    go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.63
+    go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.6
```
Jooho: This is not related to the PR, right? and it is already fixed with https://github.com/kserve/kserve/pull/4302

so this pr seems to be updated with the latest commits

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/google/go-cmp/cmp"
+
```
Jooho, line 26: Blank lines are added here. Is this intended?

Jooho (commented):
just found some blank lines and need to be updated to the latest commits.
The others looks good to me

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
spolti: added by go fmt imports formater.

hack/verify-golint.sh
spolti: yeah, this PR is a few days old and at the time that pr wasn't merged yet, will fix the conflict later.

spolti: @Jooho updated.

Jooho (approved):

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -381,31 +384,59 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deploym
...
 				return nil, err
 			}
+
 			// Generate the strategic merge patch between the current and modified JSON
 			patchByte, err := strategicpatch.StrategicMergePatch(curJson, modJson, appsv1.Deployment{})
 			if err != nil {
 				return nil, err
 			}
 
+			// override the envs that needs to be removed with  "$patch": "delete"
```
yuzisun, line 458: This is getting a little complex, so we will have the same problem for command line argument list ?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -381,31 +384,59 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deploym
...
+			patchByte = []byte(strings.ReplaceAll(string(patchByte), "\"value\":\""+utils.PLACEHOLDER_FOR_DELETION+"\"", "\"$patch\":\"delete\""))
+
 			// Patch the deployment object with the strategic merge patch
-			opErr = r.client.Patch(ctx, deployment, client.RawPatch(types.StrategicMergePatchType, patchByte))
+			opErr = r.client.Patch(ctx, deployment, kclient.RawPatch(types.StrategicMergePatchType, patchByte))
```
yuzisun, line 462: Not sure what was the original change to use patch, maybe we should reconsider update operation as patch is primarily used for targeted changes which is not the case for updating deployment spec.

yuzisun (changes requested):
I think we need to reevaluate the decision to use patch operation as it is making the code much more complex and it is primarily for targeted changes, however for updating deployment spec it can getting quite complex as argument list or other fields may suffer the same problem as environment variables.

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
spolti: I didn't test it, but recently we got another point of things not getting updated; the secret name was one of it.
@andresllh is there anything else that you have spotted?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
spolti: Yes, I had the same question. While looking at it, debugging it a little bit, the patch bytes contains the entire deployment model, breaking the advantage of using merge patch, which allows us to patch just a tiny portion of the object instead passing the complete one.

So, in this case, using update would prevent these issues and keep the code easier.

+1 from me to move to update.

spolti:
> I think we need to reevaluate the decision to use patch operation as it is making the code much more complex and it is primarily for targeted changes, however for updating deployment spec it can getting quite complex as argument list or other fields may suffer the same problem as environment variables.

Agreed, +1 from me.

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
andresllh: We will probably have a problem with anything that is an array. Latest one was the secrets as @spolti mentioned. I agree with moving forward with an update operation as @yuzisun mentioned down below.

yuzisun:
@spolti Now I think this PR is problematic https://github.com/kserve/kserve/issues/3768, we also found an issue that readiness probe can not be updated if initially not specified. cc @andyi2it

yuzisun:
> @spolti Now I think this PR is problematic #3768, we also found an issue that readiness probe can not be updated if initially not specified. cc @andyi2it

Whenever some fields need to be removed then the update will fail if using strategic merge

spolti:
If my understanding is correct, the mentioned PR moves to the strategic merge patch to make it work for rolling deployments, and if we move back, we will re-introduce the bug.

spolti:
@yuzisun @andyi2it I've created this test [PR](https://github.com/kserve/kserve/pull/4447) with `update` based on the changes in this PR.
As the test case was added before to test the rolling strategy, we might have a notion if the `update` will break things.

spolti:
Replaced by https://github.com/kserve/kserve/pull/4445

Issue #4228, opened 2025-02-05
Allow MultiNode Setup with Only tensorParallelSize Set


Jooho:
/kind feature

**Describe the solution you'd like**
Currently, pipelineParallelSize is a required parameter in the MultiNode setup. However, we should allow users to specify only `tensorParallelSize`, and automatically calculate `pipelineParallelSize` if it is not set.

If `pipelineParallelSize` is not provided, it should be derived as:
```
pipelineParallelSize = tensorParallelSize/number of GPU nodes
```
The number of GPU nodes will be determined by finding nodes with the GPU resource key (e.g., nvidia.com/gpu).

The uses case is from vllm official doc(https://docs.vllm.ai/en/latest/serving/distributed_serving.html)
```
You can also use tensor parallel without pipeline parallel, just set the tensor parallel size to the number of GPUs in the cluster. For example, if you have 16 GPUs in 2 nodes (8 GPUs per node), you can set the tensor parallel size to 16:

vllm serve /path/to/the/model/in/the/container \
     --tensor-parallel-size 16
```
**Proposed Changes:**
- Make pipelineParallelSize optional.
- If pipelineParallelSize is not set, compute it dynamically based on the number of GPU nodes.
- Identify GPU nodes by checking for the specified GPU resource key in the node labels.

sukumargaonkar:
i think pipelineParallel=2 + tensorParallel=8 and pipelineParallel=1 + tensorParallel=16 are two different (and valid) usecases
docs from vllm
Usecase 1 ([vllm-docs](https://docs.vllm.ai/en/latest/serving/distributed_serving.html))
```
For example, if you have 16 GPUs in 2 nodes (8 GPUs per node), you can set the tensor parallel size to 8 and the pipeline parallel size to 2:

 vllm serve /path/to/the/model/in/the/container \
     --tensor-parallel-size 8 \
     --pipeline-parallel-size 2
```

Usecase 2 ([vllm-docs](https://docs.vllm.ai/en/latest/serving/distributed_serving.html))
```
For example, if you have 16 GPUs in 2 nodes (8 GPUs per node), you can set the tensor parallel size to 16:

vllm serve /path/to/the/model/in/the/container \
     --tensor-parallel-size 16
```

also
```
pipelineParallelSize = tensorParallelSize/number of GPU nodes
```
do you mean ?
```
pipelineParallelSize = tensorParallelSize/number of GPUs *per* node
```

Jooho:
[...]

PR #4303, opened 2025-03-11
Refactor: Consolidate duplicate custom gpu type checking logic

Jooho:

- Add 'multiNode' Field to inferenceService-config:
  - A new field, multiNode, has been added to the inferenceService-config ConfigMap. This field is intended to support multi-node setups, providing more flexibility in GPU resource management within multi-node runtime pods.
- Refactor: Consolidate Duplicate Custom GPU Type Checking Logic:
  - The previous implementation involved repetitive code for checking custom GPU resource types. In this update, this duplicate logic has been refactored into a reusable function.
  - The refactor improves code readability and reduces redundancy.
  - Additionally, unit tests have been added to verify the functionality of the newly introduced functions an updated existing unit tests.

Fixes # part of https://github.com/kserve/kserve/issues/4228

**Type of changes**

Setting
- Deploy Kserve
- Create a test namespace "vllm-multinode"

- [X] Test A -1(unit test included)

custom GPU type without annotation
```
cat <<EOF| oc apply -f -
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      name: ""
      resources: 
        requests:
	  custom.com/gpu: 1         
    workerSpec: 
      pipelineParallelSize: 2
      tensorParallelSize: 1
EOF
```
**Expected result: denied to create isvc by webhook**

- [X] Test A-2 (unit test included)
custom GPU type with annotation
```
cat <<EOF| kubectl apply -f -
kind: InferenceService
metadata:
  annotations:
    serving.kserve.io//gpu-resource-types: '["custom.com/gpu"]
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      name: ""
      resources: 
        requests:
	  custom.com/gpu: 1         
    workerSpec: 
      pipelineParallelSize: 2
      tensorParallelSize: 1
EOF
```
**Expected result: allowed to create isvc by webhook**

clean up
```
kubectl delete isvc --all -n vllm-multinode
```

- [X] Test B - with inferenceservice-config configmap (unit test included)

update configmap
```
kubectl edit  cm inferenceservice-config -n kserve
..
   multiNode: |-
      {
        "customGPUResourceTypeList": [
          "custom.com/gpu"
        ]
      }
..

cat <<EOF| kubectl apply -f -
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      name: ""
      resources: 
        requests:
	  custom.com/gpu: 1         
    workerSpec: 
      pipelineParallelSize: 2
      tensorParallelSize: 1
EOF
```
**Expected result: allowed to create isvc by webhook**

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


cmd/manager/main.go
```diff
@@ -167,6 +167,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Update Global GPU Resource Type List when custom GPU resource types are provided
```
spolti, line 170: Shouldn't this piece of code be in the configmap.go where all the other configs are read?

config/configmap/inferenceservice.yaml
```diff
@@ -73,6 +73,24 @@ data:
           "my.custom.label.1"  
         ]
       }  
+    # ====================================== MultiNode CONFIGURATION ======================================
```
spolti, line 76: I was thinking about, whether this extra config is needed, I mean, if the user uses a custom GPU and sets it in the isvc or the SR, that should be enough, IMHO.
Because, if I understand it correctly, the user will need to add the custom GPU in the config manually, later in the isvc, and this config will just check if the custom GPU is valid.

Despite the default ones, I am not sure if validating custom ones is necessary. It would make sense if the intention is to prevent users from setting any other GPU type than the ones defined here.
I will probably be missing something.

pkg/utils/utils.go
```diff
@@ -276,46 +277,49 @@ func IsUnknownGpuResourceType(resources corev1.ResourceRequirements, customGpuRe
...
 	// Return true if there are unknown GPU resources
-	return len(possibleGPUResourceType) > 0
+	return len(possibleGPUResourceType) > 0, nil
 }
 
 // IsValidCustomGPUArray checks if the input string is a valid JSON array of strings.
 // It returns false if the array is empty, contains empty strings, or any non-string elements.
-func IsValidCustomGPUArray(s string) bool {
+// Otherwise, it returns true and the list of custom GPU types.
+func IsValidCustomGPUArray(s string) ([]string, bool) {
```
spolti, line 299: isn't better to treat it as array of string instead json?

pkg/utils/utils.go
```diff
@@ -326,3 +330,51 @@ func StringToInt32(number string) (int32, error) {
...
+				existingTypes[t] = struct{}{}
+			}
+		}
+	}
+	return updatedGPUResourceTypes, nil
+}
+
+// UpdateGlobalGPUResourceTypeList adds new GPU resource types from inferenceservice-config to constants.GPUResourceTypeList.
+func UpdateGlobalGPUResourceTypeList(newGPUResourceTypes []string) error {
+	// Use a set to avoid duplicates
```
spolti:
```suggestion
	// Use a map to avoid duplicates
```

pkg/utils/utils_test.go
```diff
@@ -656,51 +677,74 @@ func TestIsUnknownGpuResourceType(t *testing.T) {
...
+			resources: corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{
+					corev1.ResourceCPU:                    resource.MustParse("1"),
+					corev1.ResourceMemory:                 resource.MustParse("1Gi"),
+					corev1.ResourceName("nvidia.com/gpu"): resource.MustParse("1"),
+				},
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:                       resource.MustParse("1"),
+					corev1.ResourceMemory:                    resource.MustParse("1Gi"),
+					corev1.ResourceName("unknown.com-2/gpu"): resource.MustParse("1"),
```
spolti, line 693: does your code detect if in the requests GPU is the same as limits?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -156,14 +156,9 @@ func validateMultiNodeVariables(isvc *InferenceService) error {
...
 
-			customGPUResourceTypes := isvc.GetAnnotations()[constants.CustomGPUResourceTypesAnnotationKey]
-			if customGPUResourceTypes != "" {
-				if !utils.IsValidCustomGPUArray(customGPUResourceTypes) {
-					return fmt.Errorf(InvalidCustomGPUTypesAnnotationFormatError, isvc.Name, constants.CustomGPUResourceTypesAnnotationKey)
-				}
-			}
-
-			if utils.IsUnknownGpuResourceType(isvc.Spec.Predictor.Model.Resources, customGPUResourceTypes) {
+			if isUnknownGPUType, err := utils.IsUnknownGpuResourceType(isvc.Spec.Predictor.Model.Resources, isvc.Annotations); err != nil {
```
VedantMahabaleshwarkar, line 159: [here](https://github.com/kserve/kserve/pull/4303/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R281) the resourceTypeList is updated with the annotation from the isvc (if the annotation is not in the default list). So when will this function ever return false?

VedantMahabaleshwarkar: Question : [here](https://github.com/kserve/kserve/pull/4303/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R364) the constants list is updated with any new entries from the configmap, so if the isvc annotations has a new custom gpu annotation that is not in the constants list, is that a valid GPU type or unknown?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -330,11 +330,11 @@ func addGPUResourceToDeployment(deployment *appsv1.Deployment, targetContainerNa
 	// If CustomGPUResourceTypeAnnotationKey is set, the specified custom GPU resource will be added to the available GPUResourceTypeList.
 	customGPUResourceTypes := deployment.GetAnnotations()[constants.CustomGPUResourceTypesAnnotationKey]
 	if customGPUResourceTypes != "" {
-		constants.GPUResourceTypeList = append(constants.GPUResourceTypeList, strings.Split(customGPUResourceTypes, ",")...)
+		constants.DefaultGPUResourceTypeList = append(constants.DefaultGPUResourceTypeList, strings.Split(customGPUResourceTypes, ",")...)
```
VedantMahabaleshwarkar, line 333: This is already done in https://github.com/kserve/kserve/pull/4303/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R364 right?

pkg/utils/utils_test.go
Jooho: no it does not have.

pkg/utils/utils.go
Jooho: From my two cents, it is similar. It needs parsing anyway to make a string array. Are there specific benefits if I use a string array?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
Jooho: No it is different. this add the gpuType from annotations to DefaultGPUResourceTypeList. 
The one you pointed out is adding the custom GPU list in the inferenceservice-config to the DefaultGPUResourceTypeList.

pkg/apis/serving/v1beta1/inference_service_validation.go
Jooho: >here the resourceTypeList is updated with the annotation from the isvc (if the annotation is not in the default list). So when will this function ever return false?

By the fault, allowed GPU types are 4. If the user uses a custom GPU type like `custom.com/gpu: 1` and it returns false

> Question : here the constants list is updated with any new entries from the configmap, so if the isvc annotations has a new custom gpu annotation that is not in the constants list, is that a valid GPU type or unknown?

The custom GPU type in annotation will be added as a valid GPU type but it is not added to constants.
It returns [a new string array](https://github.com/kserve/kserve/pull/4303/files/7a6404c238e5747e7f38510b1a8dbe6a46a50747#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R360)

So it will be valid GPU type

cmd/manager/main.go
Jooho: Some configurations also uses [main.go](https://github.com/kserve/kserve/blob/master/cmd/manager/main.go#L154-L168)

The multinode part has only one configuration for now, and it doesn’t need to be called every time—just once to update the constants.  By the way, I already added the multinode part in configmap.go but I missed unit test for it. I will follow up it

config/configmap/inferenceservice.yaml
Jooho: Multinode in KServe needs to determine the number of GPUs by calculating tensorParallelSize and pipelineParallelSize. To specify GPUs, we need to know the exact GPU resource names.

However, GPU resource names can be arbitrarily defined by vendors, so we cannot determine them in advance. That’s why I believe this approach is the best way to identify which resources in the configuration are intended for GPUs.

Even within NVIDIA GPUs, there can be hundreds of different custom resource names, making it difficult to identify them. While we could use nvidia.com to detect them, as I mentioned earlier, this wouldn’t work for custom GPUs from new vendors. That’s why I think allowing users to define them manually is the most reasonable approach.

To make this process more convenient, I added a new field in inferenceservice-config so that all ISVCs can check GPU types automatically.

pkg/utils/utils.go
spolti: What do you mean by, `it needs parsing anyway`?

config/configmap/inferenceservice.yaml
spolti: yeah, that makes sense.

pkg/utils/utils_test.go
spolti: it might be useful if requests is greater than limits, if that happens, respect the limits to set the tensor.

Jooho: Your comment adds value, but I feel this feature might be unnecessary at this point. If a real need arises in the future, we can consider adding it then.

pkg/utils/utils.go
Jooho: Annotations in Kubernetes can only store strings, so even if we input something like ["a","b","c"], it still needs to be parsed into a string array.

In the end, whether we store it as a JSON string or a plain string array, parsing is required either way. Are there specific advantages to treating it as an array in this case?

cmd/manager/main.go
Jooho: updated

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #4313, opened 2025-03-17
[Bug] Default "dtype" was overriding vLLM "auto" dtype mode

gavrissh:
When passing dtype as "auto" in the vLLM backend it was getting overridden within the function to either fp16 or fp32

Fixes #

**Type of changes**

```
Tested creation and inference with `dtype=auto` for both (vllm and huggingface) backends
```

**Release note**:

```release-note

```

**Re-running failed tests**


python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -225,6 +221,9 @@ def load_model():
             "float": torch.float32,
         }
 
+        if "dtype" in args and args.dtype == "auto":
+            args.dtype = default_dtype
+
         model_config = AutoConfig.from_pretrained(
             str(model_id_or_path),
```
Datta0, line 227: We're loading the model config here. So let's set the dtype to float16 if we don't see anything in the model config?
Otherwise let the one in config take precedence?
Examples:
1. https://huggingface.co/meta-llama/Llama-2-13b-chat-hf/blob/main/config.json#L21
2. https://huggingface.co/google/gemma-3-1b-it/blob/main/config.json#L33
3. https://huggingface.co/meta-llama/Llama-3.2-11B-Vision-Instruct/blob/main/config.json#L216

Datta0: Something like:
```
if hasattr(model_config, torch_dtype):
    if model_config.torch_dtype=='float32':
		dtype = 'float16' # override to ensure consistency with vLLM
	else:
		dtype = model_config.torch_dtype
else:
	dtype='float16'
```

Datta0 (approved):
LGTM !

python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -217,19 +222,26 @@ def load_model():
...
         )
+
+        # Convert dtype from string to torch dtype. Default to float16
+        dtype = kwargs.get("dtype", default_dtype)
+        if dtype == "auto":
+            if (
+                hasattr(model_config, "torch_dtype")
+                and model_config.torch_dtype is not None
+            ):
+                dtype = model_config.torch_dtype
```
yuzisun, line 239: @gavrissh is this the main change to honor the dtype from model config?

gavrissh: For vLLM backend during `auto` mode, we'll let vLLM engine decide. Previously it was getting overridden due to [this](https://github.com/kserve/kserve/pull/4313/files#diff-2ed97434a3b43aac01ea6d9f2364d6c176a3716fcff260115226a61260d460f9L191-L192) 

For HF backend during `auto` mode, we will honor the type from the model config or set to our `default_dtype`
We have found in some cases the dtype in the model config can be None

yuzisun (approved):

PR #4326, opened 2025-03-21
Fix: Model cache uri requires subpath match instead of string match

greenmoon55:
Previously we did prefix matching for model cache uri. e.g. `s3://models/a/b` in the inference service can match a cached model `s3://models/a/`. 

However, we could ran into a scenario where we cache:
* `s3://models/a`

But the inference service with `s3://models/ab` matches the model cache `s3://models/a`, causing the service to load the incorrect model.

This PR updates the match logic to match by subpath instead.

For `s3://models/a`
* `s3://models/ab` is NOT a match
* `s3://models/a/b` is a match

For non-standard uris like `foo:bar:baz`, an exact match is required.

Fixes #

**Type of changes**

Updated unit tests

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/apis/serving/v1alpha1/local_model_cache_types.go
+++ b/pkg/apis/serving/v1alpha1/local_model_cache_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strings"
+
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -60,4 +62,18 @@ type LocalModelCacheList struct {
 
 func init() {
 	SchemeBuilder.Register(&LocalModelCache{}, &LocalModelCacheList{})
+}
+
+func (spec *LocalModelCacheSpec) MatchStorageURI(storageUri string) bool {
+	cachedUri := strings.TrimSuffix(spec.SourceModelUri, "/")
+	isvcStorageUri := strings.TrimSuffix(storageUri, "/")
+	if strings.HasPrefix(isvcStorageUri, cachedUri) {
+		if len(isvcStorageUri) == len(cachedUri) {
+			return true
+		}
+		if string(isvcStorageUri[len(cachedUri)]) == "/" {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/apis/serving/v1alpha1/local_model_cache_types.go
+++ b/pkg/apis/serving/v1alpha1/local_model_cache_types.go
@@ -64,6 +64,7 @@ func init() {
 	SchemeBuilder.Register(&LocalModelCache{}, &LocalModelCacheList{})
 }
 
+// If the storageUri from inference service matches the sourceModelUri of the LocalModelCache or is a subdirectory of the sourceModelUri, return true
 func (spec *LocalModelCacheSpec) MatchStorageURI(storageUri string) bool {
 	cachedUri := strings.TrimSuffix(spec.SourceModelUri, "/")
 	isvcStorageUri := strings.TrimSuffix(storageUri, "/")
@@ -71,7 +72,9 @@ func (spec *LocalModelCacheSpec) MatchStorageURI(storageUri string) bool {
 		if len(isvcStorageUri) == len(cachedUri) {
 			return true
 		}
-		if string(isvcStorageUri[len(cachedUri)]) == "/" {
+
+		// If the storageUri is a subdirectory of the cachedUri, the next character after the cachedUri should be a "/"
+		if len(cachedUri) < len(isvcStorageUri) && string(isvcStorageUri[len(cachedUri)]) == "/" {
 			return true
 		}
 	}
--- a/pkg/apis/serving/v1beta1/inference_service_defaults.go
+++ b/pkg/apis/serving/v1beta1/inference_service_defaults.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
-	"strings"
 
 	"google.golang.org/protobuf/proto"
 	corev1 "k8s.io/api/core/v1"
@@ -457,8 +456,6 @@ func (isvc *InferenceService) setLocalModelLabel(models *v1alpha1.LocalModelCach
 	if models == nil {
 		return
 	}
-	// Todo: support multiple storage uris?
-	var storageUri string
 	var predictor ComponentImplementation
 	if predictor = isvc.Spec.Predictor.GetImplementation(); predictor == nil {
 		return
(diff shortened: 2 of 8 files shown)


pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -484,7 +484,6 @@ func (isvc *InferenceService) setLocalModelLabel(models *v1alpha1.LocalModelCach
 		isvc.Annotations = make(map[string]string)
 	}
 	isvc.Labels[constants.LocalModelLabel] = localModel.Name
-	isvc.Annotations[constants.LocalModelSourceUriAnnotationKey] = localModel.Spec.SourceModelUri
```
spolti: Any reason of removing it?

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -3997,39 +3997,22 @@ func TestLocalModelPVC(t *testing.T) {
...
-			localModelSourceUriLabel: "s3://foo",
-			pvcName:                  "model-h100",
-			expectedSubPath:          "models/bar/",
+			storageUri:      "s3://foo",
+			localModelLabel: "bar",
+			pvcName:         "model-h100",
+			expectedSubPath: "models/bar/",
 		},
 		"extra / at the end": {
-			storageUri:               "s3://foo/",
```
spolti: maybe rework these instead remove?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -247,12 +247,8 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *corev1.Pod)
 
 	// Mount pvc directly if local model label exists
 	if modelName, ok := pod.ObjectMeta.Labels[constants.LocalModelLabel]; ok {
-		subPath, _ := strings.CutPrefix(srcURI, pod.ObjectMeta.Annotations[constants.LocalModelSourceUriAnnotationKey])
-		if !strings.HasPrefix(subPath, "/") {
-			subPath = "/" + subPath
-		}
 		if pvcName, ok := pod.ObjectMeta.Annotations[constants.LocalModelPVCNameAnnotationKey]; ok {
-			srcURI = "pvc://" + pvcName + "/models/" + modelName + subPath
+			srcURI = "pvc://" + pvcName + "/models/" + modelName + "/"
```
spolti: what happens with models that have big names?
e.g. OpenVino might fail with this change.

pkg/apis/serving/v1beta1/inference_service_defaults_test.go
```diff
@@ -897,7 +897,8 @@ func TestLocalModelAnnotation(t *testing.T) {
 					Predictor: PredictorSpec{
 						PyTorch: &TorchServeSpec{
 							PredictorExtensionSpec: PredictorExtensionSpec{
-								StorageURI: proto.String("gs://testbucket2/testmodel"),
+								// StorageURI has to match the one in LocalModelCache, extra suffix is not considered as a match
+								StorageURI: proto.String("gs://testbucket/testmodel/foo"),
```
spolti: it would be nice to have a test that loads two similar models and test if the correct ones were loaded where it is needed.

greenmoon55: good point. Thank you!

pkg/apis/serving/v1beta1/inference_service_defaults.go
greenmoon55: It was used to construct the pvc subPath on the pod, without prefix matching this annotation is not useful any more.

pkg/webhook/admission/pod/storage_initializer_injector.go
greenmoon55: Could you elaborate? I'm not familiar with OpenVino. We are making the srcURI shorter in the PR btw.

spolti: okay, thanks.

pkg/apis/serving/v1beta1/inference_service_defaults.go
spolti: okay, thanks.

pkg/webhook/admission/pod/storage_initializer_injector.go
spolti: OpenVino expects a very specific path to be able to load the models, e.g. /\<path\>/\<version\>/files:
```
# filesystem will look like:
/mnt/models/mnist/1/<model_files>
```

And, the definition in the storage initializer for it will be, e.g.: `pvc://mnt/models/mnist`

So, wondering, how the modelCache would behave in this case.

greenmoon55: Thank you. I'll update the PR to support subpath.

pkg/apis/serving/v1beta1/inference_service_defaults_test.go
```diff
@@ -897,27 +942,18 @@ func TestLocalModelAnnotation(t *testing.T) {
...
 						},
 					},
 				},
 			},
-			labelMatcher:      gomega.Not(gomega.HaveKeyWithValue(constants.LocalModelLabel, localModelName)),
+			labelMatcher:      gomega.Not(gomega.HaveKey(constants.LocalModelLabel)),
 			annotationMatcher: gomega.Not(gomega.HaveKey(constants.LocalModelPVCNameAnnotationKey)),
 		},
 	}
-	localModel := &v1alpha1.LocalModelCache{
```
greenmoon55, line 910: moved to the top

yuzisun:
@greenmoon55 Can we update doc for the matching logic ?

pkg/apis/serving/v1alpha1/local_model_cache_types.go
```diff
@@ -61,3 +63,17 @@ type LocalModelCacheList struct {
...
 }
+
+func (spec *LocalModelCacheSpec) MatchStorageURI(storageUri string) bool {
+	cachedUri := strings.TrimSuffix(spec.SourceModelUri, "/")
+	isvcStorageUri := strings.TrimSuffix(storageUri, "/")
+	if strings.HasPrefix(isvcStorageUri, cachedUri) {
+		if len(isvcStorageUri) == len(cachedUri) {
+			return true
+		}
+		if string(isvcStorageUri[len(cachedUri)]) == "/" {
```
yuzisun: Can you add a comment for the matching case this check is handling

pkg/webhook/admission/pod/storage_initializer_injector.go
gavrissh: I am curious to know, what if there are two paths like these?
- `pvc://mnt/models/meta-llama/Llama-3.2-1B-Instruct`
-  `pvc://mnt/models/meta-llama/Llama-3.2-3B-Instruct`

Both of these will match with `pvc://mnt/models/meta-llama/`?

greenmoon55: @gavrissh Yes, I updated the PR. Ideally we should create separate cache CRs for these two models. It's mostly for the case @spolti mentioned where the inference service uses some files under the cached folder.

yuzisun (approved):
/lgtm
/approve

PR #4337, opened 2025-03-26
Stop and resume a model by adding a new annotation [Serverless] 

### LINKED ISSUE #4207: Support stop InferenceService
/kind feature

In a private cloud environment, GPU cards are very scarce. Sometimes I create an InferenceService. When not in use, I hope to stop it and release the resources for others to use in a timely manner. This can efficiently utilize the limited GPU resources. Although Serverless is used, but I want to completely stop it to prevent accidental access. When I need to use it later, I can just start it again without creating a new InferenceService instance again.

In the design, we can add an annotation to the InferenceService, for example `kserve.io/stop: true`,  and then the controller component stops the service, and the InferenceService status field will also show stopped.

hdefazio:
 This adds the ability to stop and resume services without permanently deleting configurations or incurring delays caused by gradual scaling processes

Fixes #4207 

**Type of changes**

cluster:
Client Version: v1.30.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.15+ff493be

Test setup:
```
$ kubectl create ns stop-resume-demo
$ kubectl config set-context --current --namespace stop-resume-demo 
$ kustomize build ./config/runtimes/| sed 's/ClusterServingRuntime/ServingRuntime/g' |kubectl create -n stop-resume-demo -f -

$ podman build -t kserve:stop-resume . 
$ podman push kserve:stop-resume  quay.io/hdefazio/kserve:stop-resume

Edited kserve-controller-manager to use quay.io/hdefazio/kserve:stop-resume
```

Test: 

1. Normal deployment --> add the annotation

create a isvc without the annotation
```
kind: InferenceService
metadata:
  annotations:
  name: "sklearndemo1"
  namespace: stop-resume-demo
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
EOF
```

observe that the inference service is Ready
```
$ kubectl get isvc -n stop-resume-demo
NAME           URL                                                                                         READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION            AGE
sklearndemo1   https://sklearndemo1-stop-resume-demo.apps.rosa.han-stop-resume.wtsd.p3.openshiftapps.com   True           100                              sklearndemo1-predictor-00002   2m8s
```

```
$ kubectl get pods -n stop-resume-demo
NAME                                                       READY   STATUS    RESTARTS   AGE
sklearndemo1-predictor-00002-deployment-679bd4754f-mwvsv   3/3     Running   0          4m50s
```

Stop the model
```
$ kubectl edit isvc/sklearndemo1 -n stop-resume-demo 
Add annotation `serving.kserve.io/stop: 'true'`
```

It should remove all objects except the isvc

Observe that the ISVC now has a Stopped condition
```
kubectl describe isvc/sklearndemo1 -n stop-resume-demo 
```

2. Stopped model--> set the annotation to false
Resume the model
```
$ kubectl edit isvc/sklearndemo1 -n stop-resume-demo 
Set annotation `serving.kserve.io/stop: 'false'`
```

Observe that the KSVC, deployment, and pods are created and ready. 

**Checklist**:

**Re-running failed tests**

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -94,7 +94,7 @@ var (
 	TargetUtilizationPercentage                 = KServeAPIGroupName + "/targetUtilizationPercentage"
 	MinScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/min-scale"
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/max-scale"
-	StopResumeAnnotationKey                     = KServeAPIGroupName + "/stop"
+	StopAnnotationKey                           = KServeAPIGroupName + "/stop"
 	RollOutDurationAnnotationKey                = KnativeServingAPIGroupName + "/rollout-duration"
 	KnativeOpenshiftEnablePassthroughKey        = "serving.knative.openshift.io/enablePassthrough"
 	EnableMetricAggregation                     = KServeAPIGroupName + "/enable-metric-aggregation"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -94,6 +94,7 @@ var (
 	TargetUtilizationPercentage                 = KServeAPIGroupName + "/targetUtilizationPercentage"
 	MinScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/min-scale"
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/max-scale"
+	StopResumeAnnotationKey                     = KServeAPIGroupName + "/stop"
 	RollOutDurationAnnotationKey                = KnativeServingAPIGroupName + "/rollout-duration"
 	KnativeOpenshiftEnablePassthroughKey        = "serving.knative.openshift.io/enablePassthrough"
 	EnableMetricAggregation                     = KServeAPIGroupName + "/enable-metric-aggregation"
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -64,10 +64,11 @@ type Predictor struct {
 	inferenceServiceConfig *v1beta1.InferenceServicesConfig
 	deploymentMode         constants.DeploymentModeType
 	Log                    logr.Logger
+	forceStopRuntime       bool
 }
 
 func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *runtime.Scheme,
-	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType,
+	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType, forceStopRuntime bool,
 ) Component {
 	return &Predictor{
 		client:                 client,
@@ -76,6 +77,7 @@ func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *
 		inferenceServiceConfig: inferenceServiceConfig,
 		deploymentMode:         deploymentMode,
 		Log:                    ctrl.Log.WithName("PredictorReconciler"),
+		forceStopRuntime:       forceStopRuntime,
 	}
 }
 
@@ -190,16 +192,11 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 	} else {
 		var err error
 		podLabelKey = constants.RevisionLabel
+
 		if kstatus, err = p.reconcileKnativeDeployment(ctx, isvc, &objectMeta, &podSpec); err != nil {
 			return ctrl.Result{}, err
 		}
-
-		forceStopRuntime := "false"
-		if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
-			forceStopRuntime = val
-		}
-
-		if strings.EqualFold(forceStopRuntime, "true") {
+		if p.forceStopRuntime {
 			// Exit early if we have already set the status to stopped
 			existing_stopped_condition := isvc.Status.GetCondition(v1beta1.Stopped)
 			if existing_stopped_condition != nil && existing_stopped_condition.Status == corev1.ConditionTrue {
@@ -209,6 +206,19 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 			// Clear all statuses
 			p.Log.Info("Clearing ISVC status")
 			isvc.Status = v1beta1.InferenceServiceStatus{}
+
+			// Set the ready conditions to false
+			predictor_ready_condition := &apis.Condition{
+				Type:   v1beta1.PredictorReady,
+				Status: corev1.ConditionFalse,
+			}
+			isvc.Status.SetCondition(v1beta1.PredictorReady, predictor_ready_condition)
+
+			ingress_ready_condition := &apis.Condition{
+				Type:   v1beta1.IngressReady,
+				Status: corev1.ConditionFalse,
+			}
+			isvc.Status.SetCondition(v1beta1.IngressReady, ingress_ready_condition)
 
 			// Add the stopped condition
 			stopped_condition := &apis.Condition{
@@ -727,6 +737,8 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
 	if err != nil {
 		return nil, errors.Wrapf(err, "fails to reconcile predictor")
 	}
-	isvc.Status.PropagateStatus(v1beta1.PredictorComponent, kstatus)
+	if !p.forceStopRuntime {
+		isvc.Status.PropagateStatus(v1beta1.PredictorComponent, kstatus)
+	}
 	return kstatus, nil
 }
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -195,7 +195,7 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		}
 
 		forceStopRuntime := "false"
-		if val, exist := isvc.Annotations[constants.StopResumeAnnotationKey]; exist {
+		if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
 			forceStopRuntime = val
 		}
 
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -200,7 +200,13 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		}
 
 		if strings.EqualFold(forceStopRuntime, "true") {
-			// Clear all statuses if the ISVC is stopped
+			// Exit early if we have already set the status to stopped
+			existing_stopped_condition := isvc.Status.GetCondition(v1beta1.Stopped)
+			if existing_stopped_condition != nil && existing_stopped_condition.Status == corev1.ConditionTrue {
+				return ctrl.Result{}, nil
+			}
+
+			// Clear all statuses
 			p.Log.Info("Clearing ISVC status")
 			isvc.Status = v1beta1.InferenceServiceStatus{}
 
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -204,18 +204,11 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 			// Clear all statuses
 			isvc.Status = v1beta1.InferenceServiceStatus{}
 
-			// Set the ready condition to false via the ready condition's dependencies PredictorReady and IngressReady
 			predictor_ready_condition := &apis.Condition{
 				Type:   v1beta1.PredictorReady,
 				Status: corev1.ConditionFalse,
 			}
 			isvc.Status.SetCondition(v1beta1.PredictorReady, predictor_ready_condition)
-
-			ingress_ready_condition := &apis.Condition{
-				Type:   v1beta1.IngressReady,
-				Status: corev1.ConditionFalse,
-			}
-			isvc.Status.SetCondition(v1beta1.IngressReady, ingress_ready_condition)
 
 			// Add the stopped condition
 			stopped_condition := &apis.Condition{
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -201,9 +201,15 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 				return ctrl.Result{}, nil
 			}
 
+			deployMode := isvc.Status.DeploymentMode
+
 			// Clear all statuses
 			isvc.Status = v1beta1.InferenceServiceStatus{}
 
+			// Preserve the deployment mode value
+			isvc.Status.DeploymentMode = deployMode
+
+			// Set the ready condition
 			predictor_ready_condition := &apis.Condition{
 				Type:   v1beta1.PredictorReady,
 				Status: corev1.ConditionFalse,
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -64,11 +64,10 @@ type Predictor struct {
 	inferenceServiceConfig *v1beta1.InferenceServicesConfig
 	deploymentMode         constants.DeploymentModeType
 	Log                    logr.Logger
-	forceStopRuntime       bool
 }
 
 func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *runtime.Scheme,
-	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType, forceStopRuntime bool,
+	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType,
 ) Component {
 	return &Predictor{
 		client:                 client,
@@ -77,7 +76,6 @@ func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *
 		inferenceServiceConfig: inferenceServiceConfig,
 		deploymentMode:         deploymentMode,
 		Log:                    ctrl.Log.WithName("PredictorReconciler"),
-		forceStopRuntime:       forceStopRuntime,
 	}
 }
 
@@ -196,7 +194,7 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		if kstatus, err = p.reconcileKnativeDeployment(ctx, isvc, &objectMeta, &podSpec); err != nil {
 			return ctrl.Result{}, err
 		}
-		if p.forceStopRuntime {
+		if isvc.GetForceStopRuntime() {
 			// Exit early if we have already set the status to stopped
 			existing_stopped_condition := isvc.Status.GetCondition(v1beta1.Stopped)
 			if existing_stopped_condition != nil && existing_stopped_condition.Status == corev1.ConditionTrue {
@@ -736,7 +734,7 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
 	if err != nil {
 		return nil, errors.Wrapf(err, "fails to reconcile predictor")
 	}
-	if !p.forceStopRuntime {
+	if !isvc.GetForceStopRuntime() {
 		isvc.Status.PropagateStatus(v1beta1.PredictorComponent, kstatus)
 	}
 	return kstatus, nil
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -204,10 +204,9 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 			}
 
 			// Clear all statuses
-			p.Log.Info("Clearing ISVC status")
 			isvc.Status = v1beta1.InferenceServiceStatus{}
 
-			// Set the ready conditions to false
+			// Set the ready condition to false via the ready condition's dependencies PredictorReady and IngressReady
 			predictor_ready_condition := &apis.Condition{
 				Type:   v1beta1.PredictorReady,
 				Status: corev1.ConditionFalse,
--- a/pkg/controller/v1beta1/inferenceservice/components/predictor.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/predictor.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
+	"knative.dev/pkg/apis"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -192,6 +193,32 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		if kstatus, err = p.reconcileKnativeDeployment(ctx, isvc, &objectMeta, &podSpec); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		forceStopRuntime := "false"
+		if val, exist := isvc.Annotations[constants.StopResumeAnnotationKey]; exist {
+			forceStopRuntime = val
+		}
+
+		if strings.EqualFold(forceStopRuntime, "true") {
+			// Clear all statuses if the ISVC is stopped
+			p.Log.Info("Clearing ISVC status")
+			isvc.Status = v1beta1.InferenceServiceStatus{}
+
+			// Add the stopped condition
+			stopped_condition := &apis.Condition{
+				Type:   v1beta1.Stopped,
+				Status: corev1.ConditionTrue,
+			}
+			isvc.Status.SetCondition(v1beta1.Stopped, stopped_condition)
+
+			return ctrl.Result{}, nil
+		} else {
+			resume_condition := &apis.Condition{
+				Type:   v1beta1.Stopped,
+				Status: corev1.ConditionFalse,
+			}
+			isvc.Status.SetCondition(v1beta1.Stopped, resume_condition)
+		}
 	}
 
 	statusSpec := isvc.Status.Components[v1beta1.PredictorComponent]
@@ -204,6 +231,7 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 	if err != nil {
 		return ctrl.Result{}, errors.Wrapf(err, "fails to list inferenceservice pods by label")
 	}
+
 	if isvc.Status.PropagateModelStatus(statusSpec, predictorPods, rawDeployment, kstatus) {
 		return ctrl.Result{}, nil
 	} else {
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -220,14 +220,9 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return reconcile.Result{}, err
 	}
 
-	forceStopRuntime := false
-	if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
-		forceStopRuntime = strings.EqualFold(val, "true")
-	}
-
 	reconcilers := []components.Component{}
 	if deploymentMode != constants.ModelMeshDeployment {
-		reconcilers = append(reconcilers, components.NewPredictor(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode, forceStopRuntime))
+		reconcilers = append(reconcilers, components.NewPredictor(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode))
 	}
 	if isvc.Spec.Transformer != nil {
 		reconcilers = append(reconcilers, components.NewTransformer(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode))
@@ -259,7 +254,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if isvc.Spec.Explainer != nil {
 			componentList = append(componentList, v1beta1.ExplainerComponent)
 		}
-		if !forceStopRuntime {
+		if !isvc.GetForceStopRuntime() {
 			isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.RoutesReady)
 			isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.LatestDeploymentReady)
 		}
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -220,9 +220,14 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return reconcile.Result{}, err
 	}
 
+	forceStopRuntime := false
+	if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
+		forceStopRuntime = strings.EqualFold(val, "true")
+	}
+
 	reconcilers := []components.Component{}
 	if deploymentMode != constants.ModelMeshDeployment {
-		reconcilers = append(reconcilers, components.NewPredictor(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode))
+		reconcilers = append(reconcilers, components.NewPredictor(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode, forceStopRuntime))
 	}
 	if isvc.Spec.Transformer != nil {
 		reconcilers = append(reconcilers, components.NewTransformer(r.Client, r.Clientset, r.Scheme, isvcConfig, deploymentMode))
@@ -254,8 +259,10 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if isvc.Spec.Explainer != nil {
 			componentList = append(componentList, v1beta1.ExplainerComponent)
 		}
-		isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.RoutesReady)
-		isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.LatestDeploymentReady)
+		if !forceStopRuntime {
+			isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.RoutesReady)
+			isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.LatestDeploymentReady)
+		}
 	}
 	// Reconcile ingress
 	ingressConfig, err := v1beta1.NewIngressConfig(isvcConfigMap)
--- a/pkg/controller/v1beta1/inferenceservice/controller_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller_test.go
@@ -596,14 +596,7 @@ var _ = Describe("v1beta1 inference service controller", func() {
 				return err == nil
 			}, 30*time.Second).Should(BeTrue())
 
-			// Check that the ISVC was updated
-			updatedIsvc := &v1beta1.InferenceService{}
-			Eventually(func() bool {
-				err := k8sClient.Get(ctx, serviceKey, updatedIsvc)
-				return err == nil
-			}, timeout, interval).Should(BeTrue())
-
-			// Add a url to the knative service so the virtual services can be made
+			// Add a url to the knative service so the services can be made
 			copiedKsvc := actualService.DeepCopy()
 			predictorUrl, _ := apis.ParseURL("http://" + constants.InferenceServiceHostName(constants.DefaultPredictorServiceName(serviceKey.Name), serviceKey.Namespace, domain))
 			copiedKsvc.Status.URL = predictorUrl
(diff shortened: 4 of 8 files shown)


pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -188,6 +189,41 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		if kstatus, err = p.reconcileKnativeDeployment(ctx, isvc, &objectMeta, &podSpec); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		stop, ok := isvc.Annotations[constants.StopResumeAnnotationKey]
+		if !ok {
+			stop = "false"
+		}
```
Jooho: I recommend using runtimeShutdown as the variable name because it clearly conveys that this annotation controls the shutdown state of the runtime. This improves code readability and makes the purpose of the variable immediately understandable.

```suggestion
                forceStopRuntime :=  "false"
                
		if val, exist := isvc.Annotations[constants.StopResumeAnnotationKey];  exist {
			forceStopRuntime = val
		}
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,170 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			Expect(k8sClient.Create(context.TODO(), isvc)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, isvc)
+
+			// Knative service
+			actualService := &knservingv1.Service{}
+			predictorServiceKey := types.NamespacedName{
+				Name:      constants.PredictorServiceName(serviceKey.Name),
+				Namespace: serviceKey.Namespace,
+			}
+			Eventually(func() error { return k8sClient.Get(context.TODO(), predictorServiceKey, actualService) }, timeout).
```
Jooho: This should use `Consistently` to check if the ksvc keep staying or not

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,170 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+
+			// Check that the KSVC does not exist
+			time.Sleep(60 * time.Second)
+			actualService := &knservingv1.Service{}
+			predictorServiceKey := types.NamespacedName{
+				Name:      constants.PredictorServiceName(serviceKey.Name),
+				Namespace: serviceKey.Namespace,
+			}
+			Eventually(func() error { return k8sClient.Get(context.TODO(), predictorServiceKey, actualService) }, timeout).
+				ShouldNot(Succeed())
```
Jooho: To understand better, what do you think about using IsNotFound?
```suggestion
			Eventually(func() bool {
    return apierrors.IsNotFound(k8sClient.Get(context.TODO(), predictorServiceKey, actualService))
}, timeout).Should(BeTrue(),"The ksvc should eventually be deleted"))
```

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,170 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+
+			// Define InferenceService
+			isvc := defaultIsvc(serviceKey.Namespace, serviceKey.Name, storageUri)
+			isvc.Annotations = map[string]string{}
+			isvc.Annotations[constants.StopResumeAnnotationKey] = "true"
+			Expect(k8sClient.Create(context.TODO(), isvc)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, isvc)
+
+			// Check that the KSVC does not exist
+			time.Sleep(60 * time.Second)
```
Jooho: Why do we need to wait 60s? it sets the annotation "true" from the beginning so it should not create KSVC.

pkg/controller/v1beta1/inferenceservice/controller_test.go
Jooho, line 1: Could you please add a check if virtualService and Svc exists ?  (https://github.com/kserve/kserve/issues/4207#issuecomment-2625150999)

Based on my tests, it still left.
```
kubectl get vs,svc
NAME                                                    GATEWAYS                                           HOSTS                                                  AGE
virtualservice.networking.istio.io/sklearn-iris-burst   ["knative-serving/knative-local-gateway","mesh"]   ["sklearn-iris-burst.kserve-demo.svc.cluster.local"]   11m

NAME                         TYPE           CLUSTER-IP   EXTERNAL-IP                                            PORT(S)   AGE
service/sklearn-iris-burst   ExternalName   <none>       knative-local-gateway.istio-system.svc.cluster.local   <none>    11m
```

Jooho: Plus, I think it would be a good idea to add use cases
1. Normal deployment --> add the annotation
- create a isvc without the annotation
- update the isvc the annotation "true"
- It should remove all objects except the isvc

2. stopped annotation --> restart

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -231,6 +232,11 @@ func reconcileKsvc(desired *knservingv1.Service, existing *knservingv1.Service)
 
 func (r *KsvcReconciler) Reconcile(ctx context.Context) (*knservingv1.ServiceStatus, error) {
 	desired := r.Service
+	stop, ok := desired.Spec.Template.Annotations[constants.StopResumeAnnotationKey]
```
Jooho: it is the same

Jooho (commented):
I test this PR and have some weird stuff.

**Error Events**
~~~
LAST SEEN   TYPE      REASON                     OBJECT                                                               MESSAGE
3s          Normal    Killing                    pod/sklearn-iris-burst-predictor-00001-deployment-6d876dd66b-dpcq4   Stopping container kserve-container
3s          Normal    Killing                    pod/sklearn-iris-burst-predictor-00001-deployment-6d876dd66b-dpcq4   Stopping container queue-proxy
1s          Warning   Unhealthy                  pod/sklearn-iris-burst-predictor-00001-deployment-6d876dd66b-dpcq4   Readiness probe failed: HTTP probe failed with statuscode: 503
3s          Normal    FinalizerUpdate            route/sklearn-iris-burst-predictor                                   Updated "sklearn-iris-burst-predictor" finalizers
3s          Normal    FinalizerUpdate            ingress/sklearn-iris-burst-predictor                                 Updated "sklearn-iris-burst-predictor" finalizers
3s          Warning   InternalError              route/sklearn-iris-burst-predictor                                   failed to remove route annotation to /, Kind= "sklearn-iris-burst-predictor-00001": revisions.serving.knative.dev "sklearn-iris-burst-predictor-00001" not found
3s          Warning   UpdateFailed               route/sklearn-iris-burst-predictor                                   Failed to update status for "sklearn-iris-burst-predictor": routes.serving.knative.dev "sklearn-iris-burst-predictor" not found
3s          Warning   InferenceServiceNotReady   inferenceservice/sklearn-iris-burst                                  InferenceService [sklearn-iris-burst] is no longer Ready because of: Ready
0s          Warning   Unhealthy                  pod/sklearn-iris-burst-predictor-00001-deployment-6d876dd66b-dpcq4   Readiness probe failed: HTTP probe failed with statuscode: 503
0s          Warning   Unhealthy                  pod/sklearn-iris-burst-predictor-00001-deployment-6d876dd66b-dpcq4   Readiness probe failed: HTTP probe failed with statuscode: 503
~~~
can we clean up this error messages when we add the annotation? or is it expected?

** Message is weird**
```			
 InferenceService [sklearn-iris-burst] is no longer Ready because of: Ready
```
I think this message is something wrong. This is from https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/controller.go#L439

pkg/constants/constants.go
```diff
@@ -94,6 +94,7 @@ var (
 	TargetUtilizationPercentage                 = KServeAPIGroupName + "/targetUtilizationPercentage"
 	MinScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/min-scale"
 	MaxScaleAnnotationKey                       = KnativeAutoscalingAPIGroupName + "/max-scale"
+	StopResumeAnnotationKey                     = KServeAPIGroupName + "/stop"
```
spolti: I guess it would be better to name it as `StopAnnotationKey` to reflect better on its purpose.

gavrissh: Agree, with the  naming of `StopAnnotationKey`

pkg/controller/v1beta1/inferenceservice/controller_test.go
hdefazio: without waiting I get ` runtime error: invalid memory address or nil pointer dereference`... I will decrease the time to 10 seconds though

Jooho: I don't understand why the error should happen? From the first round reconcile, it should not create the ksvc, am I right? where does the panic come from?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -301,6 +308,36 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			}
+			return ctrl.Result{}, err
+		} else {
+			r.Log.Info("Delete services")
+			if err := r.Delete(ctx, existingService); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+
+		// Delete the virtualservice
```
Jooho: I’m not sure if it will work, as the VirtualService might be recreated by another reconciler. Have you had a chance to test it? I believe the VirtualService is managed by the ingress reconciler.

hdefazio: yes tested and it was deleted and stayed deleted on my end

Jooho: Still, I am not sure it is a good idea to delete the object from here because the object should be managed by the right reconciller. The VS are managed by ingress reconcile, right? could we think to manage it from there?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -187,6 +188,32 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
 			return ctrl.Result{}, err
 		}
+
+		forceStopRuntime := "false"
+		if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
+			forceStopRuntime = val
+		}
+
+		if strings.EqualFold(forceStopRuntime, "true") {
+			// Clear all statuses if the ISVC is stopped
```
VedantMahabaleshwarkar, line 198: Is there a reason for doing this in the predictor rather than the top level controller? IIRC all the existing logic for setting isvc statuses is in the [top level controller](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/controller.go#L297). I think it would be better to be consistent and move this logic there unless I'm missing something.

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,191 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+					fmt.Println((stopped_cond))
+					if stopped_cond.Status == corev1.ConditionTrue {
+						return true
+					}
+				}
+				return false
+			}, timeout, interval).Should(BeTrue(), "The stopped condition should be set to true")
+		})
+	})
+
```
VedantMahabaleshwarkar, line 1043: Testcase for updating an existing isvc to add the stopped annotation is missing. I think that will be the most common usecase.

VedantMahabaleshwarkar (changes requested):
Might be good to have e2e tests as well

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,49 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
 	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
 	disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost
+
+	// Check if the isvc is stopped
+	forceStopRuntime := "false"
+	if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
+		forceStopRuntime = val
+	}
+	log.Info("Stopped: ")
+	log.Info(forceStopRuntime)
```
brettmthompson:
```suggestion
	log.Info("Stopped: ", forceStopRuntime)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,49 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
+
+	if strings.EqualFold(forceStopRuntime, "true") {
+		// Delete the service
+		existingService := &corev1.Service{}
+		if err := ir.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, existingService); err != nil {
+			if apierr.IsNotFound(err) {
+				return nil
```
brettmthompson: Should we continue here instead of returning nil to ensure the virtualservice deletion logic is also executed in a scenario where the service is not found?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
hdefazio: ooops I need to clean this up, not supposed to be included

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,263 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			Expect(k8sClient.Create(context.TODO(), isvc)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, isvc)
+
+			// Check that the KSVC does not exist
+			actualService := &knservingv1.Service{}
+			predictorServiceKey := types.NamespacedName{
+				Name:      constants.PredictorServiceName(serviceKey.Name),
+				Namespace: serviceKey.Namespace,
+			}
+			Eventually(func() bool {
```
Jooho: I think you should use `Consistently` instead to check if the ksvc was created or not .

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,263 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			Expect(k8sClient.Create(context.TODO(), servingRuntime)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, servingRuntime)
+
+			// Define InferenceService
+			isvc := defaultIsvc(serviceKey.Namespace, serviceKey.Name, storageUri)
+			isvc.Annotations = map[string]string{}
+			isvc.Annotations[constants.StopAnnotationKey] = "false"
+			Expect(k8sClient.Create(context.TODO(), isvc)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, isvc)
+			time.Sleep(20 * time.Second)
```
Jooho: Instead of using Sleep, consider using Eventually to more reliably check if the object has been created.

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,263 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, expectedRequest.NamespacedName, actualIsvc)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			// Stop the inference service
+			updatedIsvc := actualIsvc.DeepCopy()
+			updatedIsvc.Annotations[constants.StopAnnotationKey] = "true"
+			Expect(k8sClient.Update(ctx, updatedIsvc)).NotTo(HaveOccurred())
+			time.Sleep(10 * time.Second)
```
Jooho: this too

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
 	serviceHost := getServiceHost(isvc)
 	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
 	disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost
+
+	// Check if the isvc is stopped
```
Jooho: The current comment might be misleading. To improve clarity, consider changing it to:

```suggestion
	// If the "serving.kserve.io/stop" annotation is present in isvc, follow the stop-handling logic
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
+				return err
+			}
+		} else if !apierr.IsNotFound(err) {
+			return err
+		}
+
+		// Delete the virtualservice
+		existingVService := &istioclientv1beta1.VirtualService{}
+		if err := ir.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, existingVService); err == nil {
+			log.Info("Delete vservices")
```
Jooho: why don't we give more detailed information  something like
```
 The InferenceService <isvc.name> is marked as stopped — delete its associated VirtualService
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
+	forceStopRuntime := "false"
+	if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
+		forceStopRuntime = val
+	}
+
+	if strings.EqualFold(forceStopRuntime, "true") {
+		// Delete the service
+		existingService := &corev1.Service{}
+		if err := ir.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, existingService); err == nil {
+			log.Info("Delete services")
```
Jooho: why don't we give more detailed information  something like
```
 The InferenceService <isvc.name> is marked as stopped — delete its associated Service
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -189,9 +192,49 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
 		}
+		if p.forceStopRuntime {
+			// Exit early if we have already set the status to stopped
+			existing_stopped_condition := isvc.Status.GetCondition(v1beta1.Stopped)
+			if existing_stopped_condition != nil && existing_stopped_condition.Status == corev1.ConditionTrue {
+				return ctrl.Result{}, nil
+			}
+
+			// Clear all statuses
+			p.Log.Info("Clearing ISVC status")
```
Jooho: I think we don't need to log this or we should increase log level to debug.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -189,9 +192,49 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+			isvc.Status = v1beta1.InferenceServiceStatus{}
+
+			// Set the ready conditions to false
```
Jooho: Could you please add more information why we add predictor_ready and ingress_ready conditions here?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
Jooho, line 1: It would be great to add more unit tests to verify that both the Service and VirtualService are properly removed.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -63,10 +64,11 @@ type Predictor struct {
 	inferenceServiceConfig *v1beta1.InferenceServicesConfig
 	deploymentMode         constants.DeploymentModeType
 	Log                    logr.Logger
+	forceStopRuntime       bool
 }
 
 func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *runtime.Scheme,
-	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType,
+	inferenceServiceConfig *v1beta1.InferenceServicesConfig, deploymentMode constants.DeploymentModeType, forceStopRuntime bool,
```
israel-hdez: Since the `Reconcile` func receives an instance of the ISVC, for cleanliness and to avoid a new argument here, I wonder if we could resolve this flag from the provided ISVC on the `Reconcile` func?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -254,8 +259,10 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if isvc.Spec.Explainer != nil {
 			componentList = append(componentList, v1beta1.ExplainerComponent)
 		}
-		isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.RoutesReady)
-		isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.LatestDeploymentReady)
+		if !forceStopRuntime {
```
israel-hdez: Without the conditional, wouldn't the resulting status be accurate?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
 	serviceHost := getServiceHost(isvc)
 	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
 	disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost
+
+	// If the "serving.kserve.io/stop" annotation is present in isvc, follow the force stop logic
+	forceStopRuntime := "false"
+	if val, exist := isvc.Annotations[constants.StopAnnotationKey]; exist {
+		forceStopRuntime = val
+	}
```
israel-hdez: There seems to be more than one instance of this code, and not all instances are equal. The other one is using `forceStopRuntime = strings.EqualFold(val, "true")` (although I see the following conditional is using the EqualFold helper)

I suggest to extract these lines to a func/method in the `InferenceService` struct.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
+		// Delete the service
+		existingService := &corev1.Service{}
+		if err := ir.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, existingService); err == nil {
+			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated Service")
+			if err := ir.client.Delete(ctx, existingService); err != nil {
+				return err
+			}
+		} else if !apierr.IsNotFound(err) {
+			return err
+		}
```
israel-hdez: For better separation of concerns, I suggest moving this code inside the `reconcileExternalService` func.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -78,9 +78,43 @@ func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infere
...
+
+		// Delete the virtualservice
+		existingVService := &istioclientv1beta1.VirtualService{}
+		if err := ir.client.Get(ctx, types.NamespacedName{Name: isvc.Name, Namespace: isvc.Namespace}, existingVService); err == nil {
+			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated VirtualService")
+			if err := ir.client.Delete(ctx, existingVService); err != nil {
```
israel-hdez: To prevent deleting a user-owned resource, I suggest validating the fecthed `VirtualService` is owned by the `InferenceService`.

This comment would also apply to the `Service`.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -189,9 +192,48 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
 		}
+		if p.forceStopRuntime {
+			// Exit early if we have already set the status to stopped
+			existing_stopped_condition := isvc.Status.GetCondition(v1beta1.Stopped)
+			if existing_stopped_condition != nil && existing_stopped_condition.Status == corev1.ConditionTrue {
+				return ctrl.Result{}, nil
+			}
+
+			// Clear all statuses
+			isvc.Status = v1beta1.InferenceServiceStatus{}
```
israel-hdez, line 207: IMO, we should not clear the whole Status field, as it can carry data from other reconcilers.

It happens that, currently, the `Predictor` reconciler is the first one to run. But just as defensive coding, I think it is better to not blindly clear everything.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -189,9 +192,48 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+				Type:   v1beta1.PredictorReady,
+				Status: corev1.ConditionFalse,
+			}
+			isvc.Status.SetCondition(v1beta1.PredictorReady, predictor_ready_condition)
+
+			ingress_ready_condition := &apis.Condition{
+				Type:   v1beta1.IngressReady,
+				Status: corev1.ConditionFalse,
+			}
+			isvc.Status.SetCondition(v1beta1.IngressReady, ingress_ready_condition)
```
israel-hdez: I think this condition should be handled in `ingress_reconciler`.

israel-hdez (commented):
This is a good start, but it is partial.

Once this is merged, you should create follow-up PRs dealing with the Transformer and the Explainer.

pkg/controller/v1beta1/inferenceservice/controller.go
hdefazio: If we don't skip this when the ISVC is stopped, these just show up with "Unknown" statuses. We're skipping adding them to avoid a messy amount of unhelpful information

pkg/controller/v1beta1/inferenceservice/components/predictor.go
hdefazio: Is there a case where the Predictor's status field would have any useful information updated after it's stopped? Even if we're worried about previous failure information being wiped, I still think it's clearer to remove it since otherwise it may be difficult to see the difference between a stopped ISVC and a failed ISVC without putting too much effort into deciphering the status.
 
By clearing it, we are making it much more legible that the ISVC is stopped and that the predictor cannot be used. Without wiping the predictor's status, the conditions are all reported with status "Unknown", which is not helpful, and the model status should be wiped since it can no longer be queried

pkg/controller/v1beta1/inferenceservice/components/predictor.go
hdefazio: not against it per se, but is it cleaner to have it in the same place as the "Predictor ready" status since the end goal of setting both is to get the "Ready" status to report as false as these are its dependent statuses?

pkg/controller/v1beta1/inferenceservice/controller.go
hdefazio: Add comment explaining that the routes and deployments are removed / never made , so the conditions are not needed

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,366 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			serviceKey := expectedRequest.NamespacedName
+			storageUri := "s3://test/mnist/export"
+
+			servingRuntime := createServingRuntime(serviceKey.Namespace, "tf-serving")
+			Expect(k8sClient.Create(context.TODO(), servingRuntime)).NotTo(HaveOccurred())
+			defer k8sClient.Delete(ctx, servingRuntime)
+
+			// Define InferenceService
+			isvc := defaultIsvc(serviceKey.Namespace, serviceKey.Name, storageUri)
+			isvc.Annotations = map[string]string{}
```
brettmthompson: This will remove the `serving.kserve.io/deploymentMode` annotation you are setting in the `defaultIsvc` method. This applies to all added tests

brettmthompson:
In the unit tests, I think we should validate that the `Service` is deleted in addition to the `VirtualService` to be thorough.

pkg/controller/v1beta1/inferenceservice/controller_test.go
hdefazio: good catch - just fixed

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,366 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, serviceKey, updatedIsvc)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			// Add a url to the knative service so the virtual services can be made
+			copiedKsvc := actualService.DeepCopy()
+			predictorUrl, _ := apis.ParseURL("http://" + constants.InferenceServiceHostName(constants.DefaultPredictorServiceName(serviceKey.Name), serviceKey.Namespace, domain))
+			copiedKsvc.Status.URL = predictorUrl
+			copiedKsvc.Status.Conditions = duckv1.Conditions{
```
brettmthompson, line 603: Adding the following lines will avoid the test from erroring due to an unexpected revision name:
```
copiedKsvc.Status.LatestCreatedRevisionName = "revision-v1"
copiedKsvc.Status.LatestReadyRevisionName = "revision-v1"
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
hdefazio: is this needed? We are setting the owner as the inference service through SetControllerReference - I don't see how a virtual service could get made/updated and skip the SetControllerReference call

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -75,57 +75,30 @@ func NewIngressReconciler(client client.Client, clientset kubernetes.Interface,
...
 
 func (ir *IngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.InferenceService) error {
-	serviceHost := getServiceHost(isvc)
-	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
 	disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost
-	if serviceHost == "" || serviceUrl == "" {
-		return nil
+
+	if err := ir.reconcileVirtualService(ctx, isvc); err != nil {
+		return errors.Wrapf(err, "fails to reconcile external name service")
```
Jooho: error message is the same as the next reconciliation. is it expected?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -75,57 +75,30 @@ func NewIngressReconciler(client client.Client, clientset kubernetes.Interface,
...
+	}
+	// Create external service which points to local gateway
+	if err := ir.reconcileExternalService(ctx, isvc, ir.ingressConfig); err != nil {
+		return errors.Wrapf(err, "fails to reconcile external name service")
```
Jooho, line 85: error message is the same as the previous reconciliation. is it expected?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -75,57 +75,30 @@ func NewIngressReconciler(client client.Client, clientset kubernetes.Interface,
...
-				err = ir.client.Update(ctx, deepCopy)
-			}
-		}
-		if err != nil {
-			return errors.Wrapf(err, "fails to create or update ingress")
-		}
+	serviceHost := getServiceHost(isvc)
+	serviceUrl := getServiceUrl(isvc, ir.ingressConfig)
+	if serviceHost == "" || serviceUrl == "" {
+		log.Info("service host and serviceurl are empty..... returning")
```
Jooho: do we need `....`??

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,394 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+				Namespace: serviceKey.Namespace,
+			}
+			Consistently(func() bool {
+				err := k8sClient.Get(context.TODO(), predictorServiceKey, actualService)
+				return apierr.IsNotFound(err)
+			}, timeout).Should(BeTrue(), "The ksvc should not be created")
+
+			// Check that the virtual service is deleted
+			virtualService := &istioclientv1beta1.VirtualService{}
+			Eventually(func() bool {
```
Jooho: Consistently would be the right one to check if the VS will not be created or not.

pkg/controller/v1beta1/inferenceservice/controller_test.go
```diff
@@ -474,6 +474,394 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+				err := k8sClient.Get(context.TODO(), serviceKey, virtualService)
+				return apierr.IsNotFound(err)
+			}, timeout).
+				Should(BeTrue(), "The virtual service should be deleted")
+
+			// Check that the service is deleted
+			service := &corev1.Service{}
+			Eventually(func() bool {
```
Jooho: Consistently would be the right one to check if the VS will not be created or not.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
israel-hdez: I can only think about keeping the logs if the stopped service was failing. Since the child resources are being fully deleted, only here we would have a copy about what was failing inside `status.modelStatus.lastFailureInfo.message`.

The previous is just feedback, not mandatory though if you prefer simplicity. Still, you may have to re-think this when you deal with the Transformer and Explainer.

The only field it is worth to think twice is `status.deploymentMode`. By clearing it, there may be an impact on #4191.

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: > Add comment explaining that the routes and deployments are removed / never made , so the conditions are not needed

I'm not sure what this meant... But, anyway, I think I now understand the rationale for the conditional :-)

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
```diff
@@ -268,7 +241,74 @@ func getHostBasedServiceUrl(isvc *v1beta1.InferenceService, config *v1beta1.Ingr
...
+
+			if err != nil {
+				return errors.Wrapf(err, "fails to create or update ingress")
+			}
+		}
+	} else {
+		// Delete the virtualservice
+		if getExistingErr == nil {
+			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated VirtualService")
+			if err := ir.client.Delete(ctx, existing); err != nil {
```
israel-hdez: Following up on https://github.com/kserve/kserve/pull/4337#discussion_r2062391117:

The very first statement in the func is `disableIstioVirtualHost := ir.ingressConfig.DisableIstioVirtualHost`. If it evaluates to `true` no virtual service is created and users may freely create their own.

Here, there is no check of the `disableIstioVirtualHost` nor a check for ownership. Thus, this code can delete a user-owned VirtualService. 

I prefer checking for ownership as that is more accurate than the flag.

israel-hdez: This same comment applies inside `reconcileExternalService`.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
israel-hdez: I see this code is no longer present in the current revision. So, I assume you moved it around?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -693,6 +727,8 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
 	if err != nil {
 		return nil, errors.Wrapf(err, "fails to reconcile predictor")
 	}
-	isvc.Status.PropagateStatus(v1beta1.PredictorComponent, kstatus)
+	if !isvc.GetForceStopRuntime() {
```
israel-hdez, line 736: If you are planning on keeping the code that resets the `status`. This change is not needed IMO.

hdefazio: The PropagateStatus call actually isn't removed - it got moved to the `reconcileKnativeDeployment` function in a different PR after my PR was opened. So this is just the result of a weird git status thing

pkg/controller/v1beta1/inferenceservice/components/predictor.go
hdefazio: updated to keep the status.DeploymentMode value

pkg/controller/v1beta1/inferenceservice/components/predictor.go
hdefazio: yes it got moved to the ingress reconciler

pkg/controller/v1beta1/inferenceservice/components/predictor.go
hdefazio: I'd agree with not needed - but isn't it better to avoid a bunch of calls to update the status field if we are just wiping it afterwards anyway?

israel-hdez: I think we can have a debate on this change :yum: 

Let's keep your change.

israel-hdez (approved):
/lgtm

Jooho (approved):
/lgtm

terrytangyuan (approved):
/lgtm
/approve

VedantMahabaleshwarkar (approved):
/lgtm

PR #4342, opened 2025-03-27
Auto-update annotation for isvc.

### LINKED ISSUE #3299: Update InferenceService when ServingRuntimes are updated
/kind feature

Currently, updating a ServingRuntime doesn't appear to immediately update any InferenceService that uses the runtime. I was expecting a new revision to be created immediately, but instead it looks like the InferenceService only picks up the new runtime config the next time it's updated.

The actual underlying problem is we use helm to install our applications, and I want to be able to configure affinities/tolerations/resources etc. for the model servers from the helm chart so they can be managed under gitops. The application itself creates new model versions which aren't managed by helm. Currently, we've got the ServingRuntime in a helm chart defining the affinities/tolerations/resources and the application is creating/updating InferenceServices that reference the runtime. I want to be able to update the helm release to trigger new revisions in any existing models. It doesn't currently trigger the update, so we're looking at using a helm hook to manually update the isvc, but it would be nice to have it handled automatically by kserve.

andresllh:

Fixes [#3299](https://github.com/kserve/kserve/issues/3299) 

**Type of changes**

- [x] Test A No Annotation
1. Created a serving runtime
2. Created an isvc that uses that serving runtime and no annotation
3. Verified that the deployment has been created with the image specified in the serving runtime
4. Updated the serving runtime spec.containers[0].image with a new image
5. Verified that the deployment now had the updated image

- [x] Test B Annotation set to auto-update=true
1. Created a serving runtime
2. Created an isvc that uses that serving runtime with the annotation set to true
3. Verified the deployment was created correctly
4. Updated the serving runtime spec.containers[0].image with a new image
5. Verified that the deployment now had the updated image

- [x] Test C Annotation set to auto-update=false
1. Created a serving runtime
2. Created an isvc that uses that serving runtime with the annotation set to true
3. Verified the deployment was created correctly
4. Updated the serving runtime spec.containers[0].image with a new image
5. Verified that the deployment DID NOT have the updated image

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


spolti (approved):
/lgtm

terrytangyuan (dismissed):
Thanks!

/lgtm
/approve

terrytangyuan (dismissed):
/lgtm
/approve

yuzisun:
@andresllh from what I understand even when the annotation auto-update is set to false, the kserve controller can still “auto” update the deployment with period resync which happens every 10 hours. It is a bit counterintuitive for users with this behavior.

andresllh:
> @andresllh from what I understand even when the annotation auto-update is set to false, the kserve controller can still “auto” update the deployment with period resync which happens every 10 hours. It is a bit counterintuitive for users with this behavior.

@yuzisun based my implementation on this open issue. Specifically @Jooho comment. https://github.com/kserve/kserve/issues/3299#issuecomment-2108541652

terrytangyuan:
Let’s make sure the comment from @yuzisun is addressed. @yuzisun Can you take another look?

sivanantha321:
@andresllh I think this needs a rebase from master

spolti:
@andresllh can you please sign your commits?

andresllh:
@yuzisun do you still have any comments or concerns regarding this?

israel-hdez:
> @andresllh from what I understand even when the annotation auto-update is set to false, the kserve controller can still “auto” update the deployment with period resync which happens every 10 hours. It is a bit counterintuitive for users with this behavior.

I agree with this. If we want a consistent experience, we need a way to freeze the deployment.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -363,8 +373,60 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1.InferenceServiceStatus, deployme
...
+		if annotations != nil {
+			if autoUpdate, found := annotations[constants.AutoUpdateAnnotationKey]; found && autoUpdate == "false" {
+				r.Log.Info("Auto-update is disabled for InferenceService", "InferenceService", isvc.Name)
+				continue
+			}
+		}
+		if isvc.Spec.Predictor.Model.ProtocolVersion == nil {
+			continue
+		}
+		if slices.Contains(supportedModelFormatNames, isvc.Spec.Predictor.Model.ModelFormat.Name) && slices.Contains(protocolVersionsAsStrings, string(*isvc.Spec.Predictor.Model.ProtocolVersion)) {
```
yuzisun: runtime matching also includes priority, if the lower priority runtime is updated the inference service probably should not be reconciled. Also there is case that user explicitly the serving runtime name. 

cc @sivanantha321 to take a look at this, feel it is duplicating with the logic we have in other places.

andresllh: Sounds good, let me know how you would like me to proceed with this. I'll wait for a response from @sivanantha321 as well before I make any changes.

sivanantha321: @andresllh How about adding a field for runtime in the ISVC status ? This will be also useful for the users to track what runtime is used by the current revision of ISVC. In this way we can avoid the complexities here. If the runtime name matches with the status field, we trigger the reconcile request for that ISVC.

andresllh: @sivanantha321 that is a great idea, any objections @israel-hdez @yuzisun? 
If not, I'll start on making this update.

spolti: One question, though, shouldn't it be updated in case the SR is updated, regardless of whether the ISVC has the annotation?

andresllh: Correct, that is the current behavior. If the SR is updated and ISVC has the auto-update annotation set to true, or if it's missing, then it will update the ISVC.

andresllh: Only if the auto-update annotation is set to false will it skip over.

spolti: right.

pkg/constants/constants.go
```diff
@@ -107,6 +107,7 @@ var (
 	QueueProxyAggregatePrometheusMetricsPort    = "9088"
 	DefaultPodPrometheusPort                    = "9091"
 	NodeGroupAnnotationKey                      = KServeAPIGroupName + "/nodegroup"
+	AutoUpdateAnnotationKey                     = KServeAPIGroupName + "/auto-update"
```
spolti: If only the annotation is present, the reconciliation will not be triggered. How about renaming it to something like, `disable-auto-update`?

pkg/controller/v1beta1/inferenceservice/controller.go
spolti: Also, the idea proposed by @sivanantha321 would make things easier as well.
+1 to the suggested approach.

pkg/constants/constants.go
andresllh: Done

andresllh:
@spolti @yuzisun @sivanantha321 I've made the changes. Can you please review when you get a chance?

andresllh:
> > @andresllh from what I understand even when the annotation auto-update is set to false, the kserve controller can still “auto” update the deployment with period resync which happens every 10 hours. It is a bit counterintuitive for users with this behavior.
> 
> I agree with this. If we want a consistent experience, we need a way to freeze the deployment.

@israel-hdez my new implementation can "freeze" the deployment. Please take a look.

spolti (approved):
Thanks @andresllh , imho it looks good.
/lgtm

spolti:
@andresllh looks like there are some conflicts, can you please help with fixing them?

yuzisun:
@andresllh the test is failing 
```
Summarizing 1 Failure:
  [FAIL] v1beta1 inference service controller When Updating a Serving Runtime [It] InferenceService should not reconcile the deployment if auto-update is disabled
  /home/runner/work/kserve/kserve/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go:2330
```

andresllh:
> @andresllh the test is failing
> 
> ```
> Summarizing 1 Failure:
>   [FAIL] v1beta1 inference service controller When Updating a Serving Runtime [It] InferenceService should not reconcile the deployment if auto-update is disabled
>   /home/runner/work/kserve/kserve/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go:2330
> ```

@yuzisun  Yeah, I'm looking into this. It passes when I run it locally, but for some reason it consistently fails in the CI.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -374,8 +386,44 @@ func inferenceServiceStatusEqual(s1, s2 v1beta1.InferenceServiceStatus, deployme
...
+		}
+		if isvc.Status.ServingRuntimeName == runtimeObj.Name {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: isvc.Namespace,
+					Name:      isvc.Name,
+				},
+			})
+		}
+	}
```
Jooho: I noticed that with this change, the default behavior is effectively disableAutoUpdate == false, which means updates will be triggered when the ServingRuntime changes. I'm wondering if that's okay, since this could lead to all related InferenceService pods being restarted after this PR is merged — which might be a change from the current behavior.

Would it make sense to keep the default behavior aligned with the current state (i.e., not triggering updates), and instead introduce a global config option in inferenceservice-config to enable auto-updates explicitly? That way, we maintain backward compatibility while allowing flexibility for those who want to opt into the new behavior.

What do you think?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -268,7 +280,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			isvc.Status.PropagateCrossComponentStatus(componentList, v1beta1.LatestDeploymentReady)
 		}
 	}
-	// Reconcile ingress
+
```
israel-hdez: Removed comment?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -330,6 +332,8 @@ func (p *Predictor) reconcileModel(ctx context.Context, isvc *v1beta1.InferenceS
 		// Get first supporting runtime.
 		sRuntime = runtimes[0].Spec
 		isvc.Spec.Predictor.Model.Runtime = &runtimes[0].Name
+		// Set the ServingRuntimeName in the status
+		isvc.Status.ServingRuntimeName = runtimes[0].Name
```
israel-hdez: AFAIK, it is possible to have a cluster-scoped `ClusterServingRuntime` with same name as a namespace-scoped `ServingRuntime`. I think you need a way to distinguish them.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -453,7 +525,7 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployCo
 		ctrlBuilder = ctrlBuilder.Owns(&netv1.Ingress{})
 	}
 
-	return ctrlBuilder.Complete(r)
+	return ctrlBuilder.Watches(&v1alpha1.ServingRuntime{}, handler.EnqueueRequestsFromMapFunc(r.servingRuntimeFunc), builder.WithPredicates(servingRuntimesPredicate)).Complete(r)
```
israel-hdez: You may also want to add a `Watch` for the `ClusterServingRuntime` case.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -197,6 +200,14 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		// Stop reconciliation as the item is being deleted
 		return ctrl.Result{}, nil
 	}
+	// Check if auto-update is disabled, this will skip the reconciliation if the annotation is present.
+	// Used for when k8s autoreconciles the InferenceService.
+	if annotations != nil {
+		if disableAutoUpdate, found := annotations[constants.DisableAutoUpdateAnnotationKey]; found && disableAutoUpdate == "true" && isvc.Status.IsReady() {
```
israel-hdez, line 207: I'm not sure I agree with this. It is very hard freeze to the `InferenceService` rather than just freezing the `ServingRuntime.

If the annotation is set since the creation of the ISVC, I don't think the controller would even create any resource. Added later, it blocks the controller from doing any status updates, self-healing, etc.

I think I was expecting the ISVC to continue to operate normally, as if it would have its own copy of the Runtime.

andresllh: The isReady() check on the isvc makes sure that all resources get created before pausing any reconciliation.

pkg/controller/v1beta1/inferenceservice/controller.go
andresllh: Done

pkg/controller/v1beta1/inferenceservice/components/predictor.go
andresllh: Done

pkg/controller/v1beta1/inferenceservice/controller.go
andresllh: Done

pkg/controller/v1beta1/inferenceservice/controller.go
andresllh: Readded

pkg/controller/v1beta1/inferenceservice/controller.go
andresllh: That is the default behavior already. Kubernetes auto-reconciles resources every 10 hours by default. This allows users to pause that auto-reconciliation. This will just propagate changes a lot quicker which is more in line with making sure that the desired state matches the current cluster state rather than waiting for the auto-reconcile to happen after 10 hours.

Jooho: Ok that was the default behavior, but I’m not sure if it makes sense. Since we’re using rollingUpdate by default, a change in the ServingRuntime would trigger the creation of new pods. This could temporarily double the resource usage, and in environments with a large number of InferenceServices, it might lead to unexpected resource exhaustion. That’s why I believe the default behavior should be reconsidered.

Jooho: Wouldn't this be a problem in production? @yuzisun @sivanantha321 @terrytangyuan wdyt?

Jooho: We discussed it and decided to keep the current behavior

PR #4348, opened 2025-03-28
Replace WorkerSpec.PodSpec to WorkerSpec.Container

### LINKED ISSUE #4346: WorkerSpec resources not applied correctly in isvc, GPU resource request missing in worker pod
/kind bug

When I set resources in the ISVC workerSpec, it should be used but it is removed.
~~~
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      resources:  {}
    workerSpec:       
      resources: 
        requests:
          nvidia.com/gpu: "2"   
~~~
With this inferenceserive, worker pod should use nvidia.com/gpu: 2 under resources. But the resource field is removed from isvc so it is not in the pod as well.

It is because WorkerSpec use PodSpec and it does not have resoure field. API should use Container instead of PodSpec.

worker pod should use nvidia.com/gpu: 2

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

Jooho:
Change WorkerSpec PodSpec to Container

Fixes https://github.com/kserve/kserve/issues/4346

**Type of changes**

- [X] Test A
  - Setup multinode servingruntime with kserve raw
  - create a isvc
```
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      resources:  {}
    workerSpec:       
      resources: 
        requests:
          nvidia.com/gpu: "2"   
```
Check ISVC and the resource part should be there.
 
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -69,7 +69,7 @@ type PredictorSpec struct {
 }
 
 type WorkerSpec struct {
-	PodSpec `json:",inline"`
+	corev1.Container `json:",inline"`
```
yuzisun, line 72: Pod has an array of containers, are you trying to limit to a single container?

spolti: It is not clear to me either.
Seeing the test that was removed, it looks like the WorkerSpec does not accept more than 1 container.

Jooho: > Pod has an array of containers, are you trying to limit to a single container?

The regular InferenceService also uses [Container](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/predictor.go#L101), which allows setting resources directly. However, WorkerSpec currently uses Containers, making it impossible to set resources directly. I want to align this behavior so that WorkerSpec follows the same approach as InferenceService.

I also considered whether WorkerSpec would ever need to support multiple containers, but I don't think so. This is because the controller role belongs to the head node, while worker nodes act as helpers. That said, I may be missing some aspects.

Using Containers provides more flexibility but requires explicitly defining each container. On the other hand, using Container is less flexible but allows direct resource configuration, which is a significant advantage.

I've submitted a PR, but I'm open to feedback and discussion because it is about API change.

israel-hdez: This API is just made public with v0.15 now live. I'd now perceive this as a breaking change.
It is not acceptable for me, but I'll defer that decision to others.

Jooho: agreed. I will close this PR for now. if we need this, then we can work it at that time.

PR #4356, opened 2025-04-03
MultiNode change a logic to calculate ray node count and gpu count

Jooho:
Changes:
- PipelineParallelSize no longer affects worker nodes, and TensorParallelSize no longer impacts the GPU count. Instead, use the new environment variable "RAY_NODE_COUNT".
- Update health_check.py to use RAY_NODE_COUNT
- Create a calculation logic to determine the head GPU count, worker GPU count, and rayNodeCount.
- Allow pipelineParallelSize to be unset in the webhook(inferenceService/Servingruntime).
- Update/add unit tests

Fixes # https://github.com/kserve/kserve/issues/4228

**Type of changes**

- [] New feature (non-breaking change which adds functionality)

- Deploy KServe
- Delete multi node serving runtime and create a new one 
```
kubectl delete clusterservingruntime --all

git clone git@github.com:kserve/kserve.git
cd kserve
export kserve_home=$PWD

kubectl create

oc delete configmap health-check ;oc create configmap health-check --from-file=${kserve_home}/python/huggingfaceserver/health_check.py

sed 's/ClusterServingRuntime/ServingRuntime/g' ${kserve_home}/config/runtimes/kserve-huggingfaceserver-multinode.yaml|sed "s|image: .*$|image: kserve/huggingfaceserver:latest-gpu|g"  | oc apply -f -


kubectl patch servingruntime kserve-huggingfaceserver-multinode --type='json' -p='[
  {
    "op": "add",
    "path": "/spec/containers/0/volumeMounts/-",
    "value": {
      "mountPath": "/huggingfaceserver/debug",
      "name": "health-check"
    }
  },
  {
    "op": "add",
    "path": "/spec/volumes/-",
    "value": {
      "name": "health-check",
      "configMap": {
        "name": "health-check"
      }
    }
  },
  {
    "op": "add",
    "path": "/spec/workerSpec/containers/0/volumeMounts/-",
    "value": {
      "mountPath": "/huggingfaceserver/debug",
      "name": "health-check"
    }
  },
  {
    "op": "add",
    "path": "/spec/workerSpec/volumes/-",
    "value": {
      "name": "health-check",
      "configMap": {
        "name": "health-check"
      }
    }
  }
]'

oc get servingruntimes.serving.kserve.io  -o yaml|sed 's|health_check.py|debug/health_check.py|g' |oc apply -f -
```

- Create a fake pvc
```
kubectl create ns vllm-multinode
kubectl config set-context --current --namespace vllm-multinode

# fake pvc
cat <<EOF| oc create -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: llama-3-8b-pvc
spec:
  accessModes:
    - ReadWriteOnce 
  resources:
    requests:
      storage: 10Gi 
EOF
```

- sample isvc
1. simple
~~~
cat <<EOF| oc create -f -
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      name: ""
    workerSpec: {}
EOF
~~~

2. complex
~~~
kind: InferenceService
metadata:
  annotations:
  name: vllm-llama3-8b
  namespace: vllm-multinode
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      resources: 
        requests:
          nvidia.com/gpu: "8"
    workerSpec:    
      pipelineParallelSize: 1
      tensorParallelSize: 16
      containers:
      - name: worker-container
        resources: 
          requests:
            nvidia.com/gpu: "8"
~~~

- You can test various isvc(please refer to tested scenarioes)

**Tested Scenarios**
- ServingRuntime
  - pipelineParallelSize(2)/tensorParallelSize(1)

Test cases ( unit test are created for these cases):
**default multinode inferenceservice: Success**
   1. expected: Success => 1 head(1 gpu) + 1 worker(1 gpu)

**default isvc with custom pipelineParallelSize/tensorParallelSize**
  1. pipelineParallelSize(3)/tensorParallelSize(1) 
     1. expected: Success => 1 head(1 gpu) + 2 worker(1 gpu)

  2. pipelineParallelSize(1)/tensorParallelSize(4) 
     1. expected: Success => 1 head(1 gpu) + 3 worker(1 GPU)

  3. tensorParallelSize(4) 
     1. expected: Success => 1 head(1 gpu) + 3 worker(1 gpu)    #  pipelineParallelSize will use 2 because it is in ServingRuntime
    
 **default isvc with custom pipelineParallelSize/tensorParallelSize + head/worker resources** 
  1. pipelineParallelSize(1)/tensorParallelSize(4) with head resources(2 gpu)/ worker resources(3 GPU) 
     1. expected: Success => 1 head(2 gpu) + 1 worker(3 GPU) 
     2. It still deploy multiNode/multiGpu deployment even though it assigned 5 GPU but it is larger than what it needs(4 GPU)

  2. pipelineParallelSize(1)/tensorParallelSize(4) with worker resources(3 gpu) 
     1. expected: Success => 1 head(1 gpu) + 1 worker(3 GPU) 
     2. head gpu will be set automatically
    
  3. pipelineParallelSize(1)/tensorParallelSize(4) with with head resources(1 gpu)/ worker resources(2 gpu) 
     1. expected: Fail => 1 head(1 gpu) + 1 worker(2 GPU)
     2. It did not create deployments with the following err msg    
     3. expected error:  total required number of GPUs(4) is less than the number of GPUs assigned to the head node(1) + worker node(2)
    
  4. pipelineParallelSize(1)/tensorParallelSize(2) with with head resources(2 GPU) 
      1. expected: Success => 1 head(2 gpu) + 0 worker(0 GPU)  
      2. It deploys a head node only when total required gpu is the same as specified gpu resource
      3. worker deployment is just template not assigned any gpu/cpu/memory.
    
  5.  pipelineParallelSize(1)/tensorParallelSize(2) with with worker resources(2 GPU) 
      1. expected: Success => 1 head(2 gpu) + 0 worker(0 GPU)  
      2. It deploys a head node only when total required gpu is the same as specified gpu resource
      3. worker deployment is just template not assigned any gpu/cpu/memory.

- Logs

**Checklist**:

 ==> I sent a pr to doc - https://github.com/kserve/website/pull/467

**Release note**:

```release-note

```

**Re-running failed tests**


spolti:
@Jooho can you please add a little more information about scenarios like:

```
pipelineParallelSize(1)/tensorParallelSize(4)
```

What would be the expected behavior since there are 2 GPUs available?

pkg/constants/constants.go
```diff
@@ -506,16 +506,19 @@ const (
...
 )
 
-// Model Parallel Options Default value
+// MultiNode default values
 const (
-	DefaultTensorParallelSize   = "1"
-	DefaultPipelineParallelSize = "2"
+	DefaultTensorParallelSize   = 1
+	DefaultPipelineParallelSize = 1
+	DefaultGPUCount             = 1
```
spolti: What would be the difference between `DefaultTensorParallelSize| DefaultPipelineParallelSize ` and `DefaultGPUCount`?
if there are two GPUs in the cluster, tensor should not be bigger than 2, right?
What is the usage of defaultGPUCount?

pkg/apis/serving/v1beta1/inference_service_validation_test.go
```diff
@@ -1173,12 +1173,89 @@ func TestValidateMultiNodeVariables(t *testing.T) {
 					},
 				},
 			},
-			expected: gomega.Equal(fmt.Errorf(InvalidUnknownGPUTypeError, "foo-6")),
+			expected: gomega.Equal(fmt.Errorf(InvalidUnknownGPUTypeError, "foo-3-2")),
 		},
-		"When unsupported storageURI set, then it should return error": {
+		"When customGPUResourceTypes set to annotations, unknownGPUResource in Predictor.Model do not return an error and function correctly": {
```
spolti, line 1178: why unknown gpu should work correctly?

spolti: ok, I got it.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -455,6 +458,29 @@ func (p *Predictor) reconcileWorker(sRuntime v1alpha1.ServingRuntimeSpec, isvc *
...
 
 func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.InferenceService, podSpec *corev1.PodSpec, annotations map[string]string, isvcGeneration string) (*corev1.PodSpec, error) {
+	var workerContainer *corev1.Container
+	var mergedWorkerPodSpec *corev1.PodSpec
+	var err error
+
+	// Initialize PipelineParallelSize and TensorParallelSize if not set
+	if sRuntime.WorkerSpec.PipelineParallelSize == nil {
+		sRuntime.WorkerSpec.PipelineParallelSize = new(int)
+		*sRuntime.WorkerSpec.PipelineParallelSize = constants.DefaultPipelineParallelSize
```
spolti: Why don't you assign it directly?
e.g. add a helper function in the `pkg/utils/types.go` to return a pointer for you based on a input.

spolti: or just `ptr.To()`

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -470,41 +496,44 @@ func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.Infere
...
-	var workerContainer *corev1.Container
-	var mergedWorkerPodSpec *corev1.PodSpec
-	var err error
-
 	targetisvcContainer := corev1.Container{}
 	if isvc.Spec.Predictor.WorkerSpec.Containers != nil {
 		targetisvcContainer = isvc.Spec.Predictor.WorkerSpec.Containers[0]
 	}
-	_, workerContainer, mergedWorkerPodSpec, err = isvcutils.MergeServingRuntimeAndInferenceServiceSpecs(sRuntime.WorkerSpec.Containers, targetisvcContainer, isvc, constants.WorkerContainerName, sRuntime.WorkerSpec.ServingRuntimePodSpec, isvc.Spec.Predictor.WorkerSpec.PodSpec)
+	_, workerContainer, mergedWorkerPodSpec, err = isvcutils.MergeServingRuntimeAndInferenceServiceSpecs(sRuntime.WorkerSpec.Containers, targetisvcContainer, isvc, constants.WorkerContainerName, sRuntime.WorkerSpec.ServingRuntimePodSpec, isvc.Spec.Predictor.PodSpec)
```
spolti: so, the the workerspec has changed?

spolti (commented):
just minor comments, otherwise lgtm

Jooho:
> @Jooho can you please add a little more information about scenarios like:

>pipelineParallelSize(1)/tensorParallelSize(4)
>What would be the expected behavior since there are 2 GPUs available?

If gpu is insufficient, pod will be pending by kubernetes scheduler

pkg/constants/constants.go
Jooho: `defaultGPU` is gpu resource count per `deployment`. In this case, it is head and worker node. 

`DefaultTensorParallelSize`, `DefaultPipelineParallelSize` : these are fo the parallelism for multi-node/multi-gpu.

> if there are two GPUs in the cluster, tensor should not be bigger than 2, right?

Just because there are two GPUs available doesn't necessarily mean we can use both — they might already be assigned to other pods. Accounting for all these possibilities would place too much responsibility on KServe. We believe it's important to fully trust the Kubernetes platform, specifically the Kubernetes scheduler, and delegate this responsibility to it.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: updated

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: oops my bad. updated

config/configmap/inferenceservice.yaml
```diff
@@ -699,7 +699,7 @@ data:
 
   deploy: |-
     {
-      "defaultDeploymentMode": "Serverless"
+      "defaultDeploymentMode": "RawDeployment"
```
spolti: why changing the default deployment mode?

pkg/constants/constants.go
spolti: Where is it being used?
You added on this PR, but I didn't find it anywhere.

config/configmap/inferenceservice.yaml
Jooho: oops.. it happened.. to test .. updated.

pkg/constants/constants.go
Jooho: Ah.. you are right. I changed it at some point, so I will remove it.

spolti (approved):

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -162,6 +166,7 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		var err error
 		workerObjectMeta, workerPodSpec, err = p.reconcileWorker(sRuntime, isvc, &podSpec, annotations, predictorAnnotations, isvcGeneration)
 		if err != nil {
+			isvc.Status.PropagateRawStatusWithMessages(v1beta1.PredictorComponent, v1beta1.InvalidGPUAllocation, err.Error(), corev1.ConditionFalse)
```
VedantMahabaleshwarkar, line 169: question: this executes if `multiNodeEnabled` is True. Is that enough? Should there also be a rawdeployment mode check here? Not sure if now/in the future multinode can be used with serverless

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -324,17 +329,30 @@ func setDefaultDeploymentSpec(spec *appsv1.DeploymentSpec) {
 	}
 }
 
-func addGPUResourceToDeployment(deployment *appsv1.Deployment, targetContainerName string, tensorParallelSize string) {
+// addGPUResourceToDeployment assigns GPU resources to a specific container in a Deployment.
```
VedantMahabaleshwarkar, line 332: nit: comments above function definition for consistency

VedantMahabaleshwarkar (approved):
minor comment, otherwise lgtm

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: multinode/multigpu supports only for raw and no plan to support serverless.

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
Jooho: I believe the comment is already above the function — is there something specific you were referring to?

VedantMahabaleshwarkar: my bad, I misread the diff.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: We provide the condition in the doc so I think it should be ok.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -521,21 +549,101 @@ func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.Infere
...
+	if workerNodeGPUCount > 0 {
+		rayNodeCount = computeRayNodes(totalRequestGPUCount, workerNodeGPUCount)
+		newHeadNodeGPUCount := totalRequestGPUCount - workerNodeGPUCount*(rayNodeCount-1)
+
+		if headNodeGPUCount > 0 {
+			rayNodeCountByHeadGpuCount := computeRayNodes(totalRequestGPUCount, headNodeGPUCount)
+			if rayNodeCountByHeadGpuCount == 1 {
+				rayNodeCount = 1
+			}
+		} else if headNodeGPUCount == 0 || headNodeGPUCount == newHeadNodeGPUCount {
```
israel-hdez: The second conditional `headNodeGPUCount == newHeadNodeGPUCount` seems to be void because the code flow would enter into the  `if headNodeGPUCount > 0` branch, right?

So, you can remove the `headNodeGPUCount == newHeadNodeGPUCount`conditional to keep the code clean.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -521,21 +549,101 @@ func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.Infere
...
+	workerNodeGPUCount := getGPUResourceQty(mergedWorkerPodSpec)
+
+	// Case 1 & 2: At least worker GPUs are set
+	if workerNodeGPUCount > 0 {
+		rayNodeCount = computeRayNodes(totalRequestGPUCount, workerNodeGPUCount)
+		newHeadNodeGPUCount := totalRequestGPUCount - workerNodeGPUCount*(rayNodeCount-1)
+
+		if headNodeGPUCount > 0 {
+			rayNodeCountByHeadGpuCount := computeRayNodes(totalRequestGPUCount, headNodeGPUCount)
+			if rayNodeCountByHeadGpuCount == 1 {
```
israel-hdez, line 608: Maybe the code can be simpler to understand with the following?
```suggestion
			if totalRequestGPUCount <= headNodeGPUCount {
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -521,21 +549,101 @@ func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.Infere
...
+	}
+
+	// Case 3: Only head GPU is set
+	if headNodeGPUCount > 0 {
+		remainingGPUs := totalRequestGPUCount - headNodeGPUCount
+		rayNodeCount = 1
+		workerNodeGPUCount = 1
+
+		if remainingGPUs > 0 {
+			rayNodeCount = computeRayNodes(remainingGPUs, 1) + 1 // Single GPU worker nodes
```
israel-hdez: Simplification suggestion:

```suggestion
			rayNodeCount += remainingGPUs // Single GPU worker nodes
```

pkg/webhook/admission/servingruntime/servingruntime_webhook.go
```diff
@@ -223,39 +220,31 @@ func validateMultiNodeSpec(newSpec *v1alpha1.ServingRuntimeSpec, existingSpec *v
...
-		}
-
-		if newSpec.WorkerSpec.TensorParallelSize == nil {
-			return errors.New(MissingTensorParallelSizeValueError)
-		}
-
 		// WorkerSpec.PipelineParallelSize should not be less than 2.
 		pipelineParallelSize := *newSpec.WorkerSpec.PipelineParallelSize
-		if pipelineParallelSize < 2 {
+		if pipelineParallelSize < constants.DefaultPipelineParallelSize {
```
israel-hdez, line 245: We don't have a defaulting webhook, right?
If not, I think we, at least, need the `nil` checks. Otherwise, the de-reference of `*newSpec.WorkerSpec.PipelineParallelSize` will generate a panic.

This also applies to the following lines.

python/huggingfaceserver/health_check.py
```diff
@@ -222,32 +218,32 @@ def main():
...
 
-    # Check if registered node is the same as pipelineParalleSize/ runtime health subcommand
+    # Check if registered node is the same as rayNodeCount/ runtime health subcommand
     registered_node_and_runtime_health_parser = subparsers.add_parser(
         "registered_node_and_runtime_health",
         help="Check node counts and runtime health",
     )
     registered_node_and_runtime_health_parser.add_argument(
-        "--pipeline_parallel_size",
+        "--ray_node_count",
```
israel-hdez, line 243: Just as non-actionable feedback: I think these renamings are good as long as this feature is not advertised as stable (thus no changes requested for now).

Once we say multi-node/mult-gpu is stable we should be more careful, as these kind of renamings will lead to re-creation of any existing pods, which sometimes is not very tolerable in production envs.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: good catch! updated

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: That should work as well, but I think the original code is more readable to me.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
Jooho: This is the same reason why I want to keep original code but I want to update like this
```
rayNodeCount = computeRayNodes(remainingGPUs, workerNodeGPUCount) + 1 // Single GPU worker nodes
```
want to keep the same format that is used in the func

pkg/webhook/admission/servingruntime/servingruntime_webhook.go
Jooho: nil check is already there at the first line
```
	if existingSpec.WorkerSpec != nil && newSpec.WorkerSpec == nil {
		return errors.New(DisallowedRemovingWorkerSpecFromServingRuntimeError)
	}

	if newSpec.WorkerSpec != nil {
```

Jooho: right. I only checked workerSpec. I updated

python/huggingfaceserver/health_check.py
Jooho: Yes, I agree with your point.
That said, when we introduce a change to core logic like this, users typically need to restart the pod anyway in order to pick up the new environment variable. So from that perspective, the impact is somewhat inevitable.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -521,21 +549,101 @@ func multiNodeProcess(sRuntime v1alpha1.ServingRuntimeSpec, isvc *v1beta1.Infere
...
+		}
+
+		if err := validateGPUAllocation(rayNodeCount, headNodeGPUCount, workerNodeGPUCount); err != nil {
+			return 0, 0, 0, err
+		}
+		return rayNodeCount, workerNodeGPUCount, headNodeGPUCount, nil
+	}
+
+	// Case 4: No GPUs found → Default values
+	return totalRequestGPUCount, 1, 1, nil
```
israel-hdez, line 644: Hmm... so, I was about to approve but I'm thinking twice about this specific line.
Wouldn't this lead to a change in behavior?

The way I understand it is that if both the `ServingRuntime` and the `InferenceService` does **not** include the gpu resources, and only pipelineParallelSize and tensorParallelSize are specified, the old behavior was:
* Set the GPU resources to tensorParallelSize
* The number of pods spawned (workers + head) is the pipelineParallelSize number

The new behavior seems to be:
* Set the GPU resources to one
* The number of pods spawned (workers + head) is the multiplication of pipelineParallelSize*tensorParallelSize.

If I am understanding correctly, are we good with the change in behavior?

Jooho: Yes that is the right understanding. The goal of this PR is to give users more options, so they can use multinode in the way that best fits their needs.
Previously, as you mentioned, there were hard limits(GPU count use TP value, PP must be set and it must be more than 2 and so on) in place, which sometimes made it difficult to support the user's preferred setup — leading to some inconvenience.

israel-hdez (approved):
OK, I'm approving.

There is a behavior change, but as I am not seeing multi-node/multi-gpu as a stable API we can adapt.

terrytangyuan (commented):
/lgtm
/approve

terrytangyuan (approved):

PR #4373, opened 2025-04-08
Fix for KEDA scaledobject target value is set to pointer instead of specified value

### LINKED ISSUE #4414: KEDA scaledobject target value is set to pointer instead of specified value
/kind bug

I have created an InferenceService with KEDA autoscaler, and defined the CPU averageUtilization threshold in the spec. However, the resulting ScaledObject resource sets the metric value to a pointer-like number instead of the provided value of 50.

The ScaledObject created from the InferenceService spec should reflect the given CPU averageUtilization value of 50.

```yaml
apiVersion: "[serving.kserve.io/v1beta1](http://serving.kserve.io/v1beta1)"
kind: "InferenceService"
metadata:
  name: "sklearn-v2-iris"
  annotations:
    [serving.kserve.io/deploymentMode](http://serving.kserve.io/deploymentMode): "RawDeployment"
    [serving.kserve.io/autoscalerClass](http://serving.kserve.io/autoscalerClass): "keda"
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      protocolVersion: v2
      runtime: kserve-sklearnserver
    minReplicas: 1
    maxReplicas: 5
    autoScaling:
      metrics:
        - type: Resource
          resource:
            name: cpu
            target:
              type: "Utilization"
              averageUtilization: 50
```

Image

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

andyi2it:
The ScaledObject created from the InferenceService spec should reflect the given CPU averageUtilization instead of using the pointer as the target value.

Fixes #4414 

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
@@ -88,16 +88,14 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
 							averageUtil = &constants.DefaultCPUUtilization
 						}
 					}
-					if metric.Resource.Target.AverageUtilization != nil {
-						targetValue = fmt.Sprintf("%d", averageUtil)
-					}
+					targetValue = strconv.Itoa(int(*averageUtil))
 				case v1beta1.AverageValueMetricType:
 					if metric.Resource.Target.AverageValue != nil {
 						targetValue = metric.Resource.Target.AverageValue.String()
 					}
 				case v1beta1.ValueMetricType:
 					if metric.Resource.Target.Value != nil {
-						targetValue = fmt.Sprintf("%f", metric.Resource.Target.Value.AsApproximateFloat64())
+						targetValue = metric.Resource.Target.Value.String()
 					}
 				}
 				triggers = append(triggers, kedav1alpha1.ScaleTriggers{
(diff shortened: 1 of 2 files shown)


yuzisun:
@andyi2it can you update the PR description ?

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -88,16 +88,14 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
 							averageUtil = &constants.DefaultCPUUtilization
 						}
 					}
-					if metric.Resource.Target.AverageUtilization != nil {
-						targetValue = fmt.Sprintf("%d", averageUtil)
-					}
+					targetValue = strconv.Itoa(int(*averageUtil))
```
andyi2it: image
Value is being set to the pointer value rather than the provided value of 50

yuzisun: Can you help add a test to cover this?  also the average util needs to be integer ?

yuzisun: We still need the nil check as there is no default value for memory

yuzisun (approved):

PR #4381, opened 2025-04-09
[WIP] Add support for writing model request/response logs to object stores

### LINKED ISSUE #4380: Add support for writing model request/response logs to object stores
/kind feature

## Why?

## How?

jared-cloudera:
## **What this PR does / why we need it**:

Loading credentials used by the SDKs is accomplished by three new annotations in the `serving.kserve.io` space:
- **serving.kserve.io/logger-secret-name** - The name of the K8s Secret containing the credentials document. Defaults to **logger-credentials**
- **serving.kserve.io/logger-secret-path** - The path at which to mount the credential file. Defaults to **/etc/secrets**
- **serving.kserve.io/logger-secret-file** - The filename of the credential configuration. Defaults to **credentials.yaml**. YAML and JSON are both supported.

So by default, the logger will attempt to look for a Secret named **logger-credentials** which will then be available to the agent container at `/etc/secrets/credentials.yaml`.

The agent injector loads the secret, creates the volume and volume mount, then passes the credential file location to the agent via its arguments.

Fixes #4380

## **Type of changes**

### Unit tests
- [X] Added additional test to `logger/handler_test.go`
- [X] Added new test, `logger/store_test.go`
- [X] Added new agent command argument to `agent_injector_test.go`

### Feature testing

I've been using configs such as
```yaml
kind: InferenceService
metadata:
  name: sklearn-iris
spec:
  predictor:
    serviceAccountName: logger-sa
    logger:
      mode: all
      url: s3://inference-logs
    model:
      modelFormat:
        name: sklearn
```      

The key elements here are:

- The url which has an `s3://bucket` instead of a HTTP url
- A service account, which is used to load a Secret for the store's credential.

Customizing the location of the credentials:

```yaml
kind: InferenceService
metadata:
  name: sklearn-iris
spec:
  predictor:
    annotations:
      serving.kserve.io/logger-secret-name: logger-credentials-json
      serving.kserve.io/logger-secret-file: credentials.json
    serviceAccountName: logger-sa
    logger:
      mode: all
      url: s3://inference-logs
    model:
      modelFormat:
        name: sklearn
```

An example `logger-credentials.yaml` could look like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: logger-credentials
type: Opaque
stringData:
  credentials.yaml: |
    s3:
      s3AccessKeyIDName: foo
      s3SecretAccessKeyName: bar
      s3Endpoint: http://192.168.1.2:9000
    s3ForcePathStyle: true
```

This Secret could then be bound to a Service Account such as referenced above:
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: logger-sa
secrets:
  - name: logger-credentials
``` 

I have primarily used a local MinIO server to test, as well as S3 itself.

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


yuzisun:
@jared-cloudera annotation is hard to validate as part of the schema, the general guideline is to specify first class citizen fields if we think this is a core logging feature in the long term. There is already existing storage spec which can be reused and added to the logger struct for specifying the storage secrets which can be potentially different from the model storage secret.

jared-cloudera:
> @jared-cloudera annotation is hard to validate as part of the schema, the general guideline is to specify first class citizen fields if we think this is a core logging feature in the long term. There is already existing storage spec which can be reused and added to the logger struct for specifying the storage secrets which can be potentially different from the model storage secret.

Great, thanks Dan. I wasn't sure if adding new fields here was acceptable or not. We will revisit this

cjohannsen-cloudera:
@yuzisun Would you mind reviewing now?  I have refactored to use a top level StorageSpec on the logger config as you suggested.

jared-cloudera:
Continuing as https://github.com/kserve/kserve/pull/4422

PR #4391, opened 2025-04-11
Initial segregation of the storage module from KServe SDK

### LINKED ISSUE #3489: Make storage initializer install only what is needed for it to run

spolti:
chore:  This PR aims to create a new storage module, called `kserve-storage`
	     that will be totally independent from the existing one which is part
	     of the current KServe Python SDK.


How to test it?
This python package was published here: https://test.pypi.org/project/kserve-storage/
It can be tested by:
```bash
# create a new venv:
virtualenv -p python3.12 /tmp/venv1 && source /private/tmp/venv1/bin/activate
# Then install it:
 pip install kserve-storage==0.15.0 \                                                                                                                                                                                                      
  --index-url https://test.pypi.org/simple \
  --extra-index-url https://pypi.org/simple
```
Note, the extra index is necessary because some of the dependencies are not available in the test index.

Later, test the `Storage` module with this command:
```bash
python -c "from kserve_storage import Storage; from kserve_storage import configure_logging, logger; Storage.download('hf://google/t5-small-lm-adapt', '/tmp/model/')"
```



Partially Fixes https://github.com/kserve/kserve/issues/3489

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

The documentation changes will be done together with this PR after we agreed to move forward, or in a subsequent PR where all needed changes will be made.

**Release note**:
<!--  Write your release note:
1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
2. If no release note is required, just write "NONE".
-->
```release-note
Storage module segregation - it now has its own module  called `kserve-storage`, available at Pypi.org
```

**Re-running failed tests**


spolti:
@yuzisun @sivanantha321 @Jooho

python/storage/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Hugging Face Hub
+
+```python
+model_dir = Storage.download("hf://org-name/model-name")
+# With specific revision
+model_dir = Storage.download("hf://org-name/model-name:revision")
+```
+
+## Environment Variables
```
Jooho, line 102: How to set HF token to pull a model from Huggingface hub private repo?

Jooho (commented):
Most of the source code seems to be from the existing implementation, and the same goes for the tests, right? Since there are no issues with the unit tests, I believe the overall changes are fine.
I left a comment on one point I was curious about, but aside from that, everything looks good to me.

Moreover, I tested this based on what you described and it worked fine

python/storage/README.md
spolti: It is the same exposed by the library: https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables

spolti:
> chore: This PR aims to create a new storage module, called `kserve-storage` that will be totally independent from the existing one which is part of the current KServe Python SDK.
> 
> How to test it? This python package was published here: https://test.pypi.org/project/kserve-storage/ It can be tested by:

Correct, after we get this step done, then we can delete the storage module from kserve/kserve and use this instead.

python/storage/README.md
Jooho: Yes I know how to set the token but my question was, we don't need to add the part in this doc like others or not

spolti: Maybe just link the huggingface docs, what do you think?

Jooho (dismissed):
/lgtm

python/storage/README.md
spolti: Done, thanks, sry for missing the 👍

terrytangyuan (approved):
LGTM. PTAL @yuzisun @sivanantha321

python/storage/pyproject.toml
```diff
@@ -0,0 +1,40 @@
+[tool.poetry]
+name = "kserve-storage"
+version = "00001"
```
sivanantha321, line 3: how does the version is updated ?

spolti: it will use the KServe actual version, I did it for testing when pushing the module to the test pypi instance.

PR #4394, opened 2025-04-11
Improve Handling of Knative Autoscaler Confguration

### LINKED ISSUE #4389: Handle Knative Autoscaler Configuration When Creating Knative Services
/kind feature

**Gaps Identified With Kserve's Handling of the Knative Autoscaler Configuration** 
Currently, Kserve does not take into account the knative autoscaler configuration when creating knative services for Inference Services or Inference Graphs. This leads to several gaps in the way kserve handles the creation of knative services:

1. kserve does not account for the global knative configuration of `initial-scale`. This can lead to scenarios where the number of replicas created to initialize a knative revision does not fall into the range of the requested min or max replicas for the Inference Service or Inference Graph. Consider the following scenarios:

    1. knative is globally configured with an `initial-scale` value that exceeds the max replicas requested for an Inference Service or Inference Graph. In this scenario, the knative revision created will be initialized with a greater number of replicas created than what the user requested as the maximum value.
    2. knative is globally configured with an `initial-scale` value that exceeds the min replicas requested for an Inference Service, but is less than the requested max replicas. knative will always choose the larger of `min-scale` and `initial-scale` as the initial target scale for a knative revision. So, in this scenario the knative revision created will be initialized with a greater number of replicas created than the user requested minimum value. A prominent case of this scenario would be when the min replicas requested is 0. The default `initial-scale` value used by knative is 1, so if a user installs knative and kserve out of the box, there will always be 1 replica created to initialize the revision when min replicas 0 is requested.

2. kserve only [adds](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go#L92) the knative `autoscaling.knative.dev/max-scale` annotation to a knative service's template spec if the max replicas value requested for an Inference Service is a non zero value (**kserve never adds this annotation for Inference Graphs**). Due to this, if an Inference Service is requested with 0 max replicas, knative will use the globally configured `max-scale` value for the created knative revision, alternatively for Inference Graphs, the globally configured `max-scale` value will **always** be used. This presents a bug for the following scenario:

    1. knative is globally configured with a `max-scale-limit` value of 0, i.e. infinite scaling is allowed, and a `max-scale` value that is greater than 0. If a user desires to allow a requested Inference Service or Inference Graph to scale up infiinitely by requesting for the max replicas to be 0, this would not be possible since no `autoscaling.knative.dev/max-scale`  annotation is added, knative will fall back to using the global configuration which has a `max-scale` value that is greater than 0, effectively placing a limit on the scale.

**Proposed Solutions**
The scenarios mentioned in [1] above, while still leading to a confusing user experience, can be argued as the proper behavior a defined by knative. It is up to debate whether these represent gaps that should be addressed. On the other hand, the scenario mentioned in [2] above presents a possible bug. That being said, I have worked on implementing solutions for all scenarios mentioned above, which are listed below:

1.i. When a knative service is created for an Inference Service or Inference Graph, read the  knative autoscaler configuration defined in the knativeserving custom resource to get the `initial-scale` value. If the requested max replicas is greater than 0 and less than the configured `initial-scale` value, then add the `autoscaling.knative.dev/initial-scale` annotation to the created knative service's template spec with a value equal to the requested max replicas. Alternatively, we can output a log in the kserve-controller-manager pod to inform the user that the globally configured knative initial scale value exceeds the requested max replicas.

1.ii. When a knative service is created for an Inference Service or Inference Graph, by default add the `autoscaling.knative.dev/initial-scale` annotation to the created knative service's template spec with a value equal to the requested min replicas (also solves above.). To prevent limiting the users ability of customizing the initial scale, we can allow users to set the `autoscaling.knative.dev/initial-scale` annotation in the Inference Service or Inference Graph to override the default initial scale functionality. To handle the edge case when 0 initial scale is requested with knative configured to not allow zero initial scale, we can set the initial-scale to 1 since this is the next closest value.

2.i. Always add the `autoscaling.knative.dev/max-scale` annotation to the created knative service's template spec, for **both** Inference Services and Inference Graphs, with a value equal to the requested max replicas.

**References:**
I have a [PR](https://github.com/opendatahub-io/kserve/pull/537) open downstream in opendatahub-io/kserve that implements the changes mentioned in 1.b., specifically for the scenario where 0 min replicas are requested.
[Knative Configuring Scale Bounds Documentation](https://knative.dev/docs/serving/autoscaling/scale-bounds/)

brettmthompson:

fixes #4389 

The changes in this PR encompass the following:

1. Aligning the logic used to set autoscaling annotations in both  ISVCs and IGs
2. Now checking the `allow-zero-initial-scale` value in the knative serving autoscaler configuration when creating a new knative service for Inference Services or Graphs. This configuration is defined in the `config-autoscaler` configmap. This configmap will always have the name 'config-autoscaler', but there is no guarantee as to which namespace it exists within. To accommodate this, adding support for a new  `KNATIVE_CONFIG_AUTOSCALER_NAMESPACE` env variable to be set if this configmap exists in a namespace other than the default of 'knative-serving'.
3. By default, now adding the `autoscaling.knative.dev/initial-scale` annotation to created knative revisions with a value equal to the requested min replicas. End users have the ability to override this default behavior by setting the `autoscaling.knative.dev/inital-scale` annotation within an Inference Service or Graph resource.
5. In a scenario where knative is configured to not allow zero initial scale and an initial scale of 0 is requested , now setting the `autoscaling.knative.dev/inital-scale` annotation to 1 because this is the closest allowed value to the requested initial scale. Once the knative revision is initialized with 1 replica, it will be scaled down to 0.
4. Now always adding a `autoscaling.knative.dev/max-scale annotation` equal to the value of the requested max replicas.

**Type of changes**

- [x] Unit Tests
- [ ] E2E Tests
- [x] Manual Integration Tests

**Test Strategy:**

Inference Service:
1. Install Knative Serving and Kserve in a cluster
2. Deploy sample ISVC with min replicas value of 1 and max replicas value of 1
3. Update the KnativeServing operator and apply the following value under the config key. (Alternatively, if KnativeServing CR is not installed in the cluster, this value can be updated in the config-autoscaler configmap directly):
```    
autoscaler:
      allow-zero-initial-scale: "true"
````
5. Edit the kserve-controller-manager-deployment and update the image to be quay.io/bretthom/kserve-dev:autoscaling
6. Edit the ISVC to change minReplicas to 0
7. Observe that no new pods are created for the new knative revision and the old pod is terminated
8. Add the `autoscaling.knative.dev/inital-scale` annotation to the ISVC with a value of 2
9. Observe that 2 new pods are created to initialize the new knative revision, then get scaled down to 0

Inference Graph:
1. Install Knative Serving and Kserve in a cluster
2. Deploy sample IG with min replicas value of 1 and max replicas value of 1
3. Update the KnativeServing operator and apply the following value under the config key. (Alternatively, if KnativeServing CR is not installed in the cluster, this value can be updated in the config-autoscaler configmap directly):
```    
autoscaler:
      allow-zero-initial-scale: "true"
````
5. Edit the kserve-controller-manager-deployment and update the image to be quay.io/bretthom/kserve-dev:autoscaling
6. Edit the IG to change min replicas to 0
7. Observe that no new pods are created for the new knative revision and the old pod is terminated
8. Add the `autoscaling.knative.dev/inital-scale` annotation to the IG with a value of 2
9. Observe that 2 new pods are created to initialize the new knative revision, then get scaled down to 0

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -218,10 +220,26 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		knServingFound, knServingCheckErr := utils.IsCrdAvailable(r.ClientConfig, operatorv1beta1.SchemeGroupVersion.String(), constants.KnativeServingKind)
+		if knServingCheckErr != nil {
+			return reconcile.Result{}, knServingCheckErr
+		}
+
+		if !knServingFound {
+			r.Recorder.Event(graph, corev1.EventTypeWarning, "ServerlessModeRejected",
+				"It is not possible to use Serverless deployment mode when Knative KnativeServings are not available")
+			return reconcile.Result{Requeue: false}, reconcile.TerminalError(fmt.Errorf("the resolved deployment mode of InferenceGraph '%s' is Serverless, but Knative KnativeServings are not available", graph.Name))
+		}
```
pierDipi: Knative serving can also be installed without the operator and in that case the warning "It is not possible to use Serverless deployment mode when Knative KnativeServings are not available" is not quite accurate

pierDipi (commented):
I'm wondering if we need to also handle the case where Knative Serving is not installed using the operator or does Kserve _requires_ the Knative operator for serveless mode?

spolti:
We check the CRD, I think in this case, the instalation mode does not matter since the CRDs are available, this way, we also support customer that already have Serverless installations and wants to reuse it.

brettmthompson:
> I'm wondering if we need to also handle the case where Knative Serving is not installed using the operator or does Kserve _requires_ the Knative operator for serveless mode?

@pierDipi, this is a great point. The knative [documentation](https://knative.dev/docs/install/yaml-install/serving/install-serving-with-yaml/) describes how knative can be installed without the knative operator. In this scenario the KnativeServing CRD will not be installed. 

To handle all knative installations, would it be better to rely on the config-autoscaler configmap as a source of truth for the knative autoscaler configuration?

Jooho:
>To handle all knative installations, would it be better to rely on the config-autoscaler configmap as a source of truth for the knative autoscaler configuration?

How about defining a priority order for the check?
For example, if the CRD exists, we check the CR. If the CRD is not installed, we fall back to checking the ConfigMap.
This way, it could also be reused downstream.

pierDipi:
I created a PR for the Knative operator to instantly reconcile configmap changes if made directly to underlying resources: https://github.com/knative/operator/pull/2062.

The source of truth for Knative is the ConfigMap, the fact that the operator isn't applying those instantly is a bug of the operator and I wouldn't complicate the logic here with priorities and fallback but rather use the ConfigMap and decouple us from how Knative is installed.

pkg/controller/v1alpha1/utils/utils.go
```diff
@@ -30,3 +41,106 @@ func CheckNodeAffinity(pvSpec *corev1.PersistentVolumeSpec, node corev1.Node) (b
...
+	// Always use the first knativeserving resource returned.
+	// We are operating under the assumption that there should be a single knativeserving custom resource created on the cluster.
+	kserving := kservingList.Items[0]
+	if kserving.Spec.Config != nil {
+		if kservingAutoscalerConfig, ok := kserving.Spec.Config[constants.AutoscalerKey]; ok {
+			if configuredAllowZeroInitialScale, ok := kservingAutoscalerConfig[constants.AutoscalerAllowZeroScaleKey]; ok {
+				allowZeroInitialScale = configuredAllowZeroInitialScale
+			}
+		}
+	}
```
pierDipi: Another downside of using KnativeServing CRD for this check is that users can set configurations for the autoscaler with both (equivalent with a priority for the shorter form) but we're only checking one key:

```yaml
spec:
  config:
    autoscaler: ...
```
and
```yaml
spec:
  config:
    config-autoscaler: ...
```
which couples this logic to whatever the operator offers/does/will do.

Ref: https://github.com/knative/operator/blob/5bef76f94348e361563e37bad282ce662090f212/pkg/reconciler/common/config_maps.go#L26-L41

brettmthompson:
Based on the conversation above, I believe it is best to update the implementation to rely on the autoscaler configmap instead of the KnativeServing CR. This configmap is required for knative autoscaling functionality, whereas the CR is not a requirement. Additionally, after @pierDipi's [PR](https://github.com/knative/operator/pull/2062) is merged, this configmap will always reflect the values present in the KnativeServing CR if it exists. As Pierangelo stated, this configmap should be treated as the source of truth.

This configmap must be named "config-autoscaler" so the only unknown is what namespace it exists within. We can allow admins to configure this namespace as an env variable, within the kserve-controller-manager deployment and fall back to using the default "knative-serving" namespace if the variable is not present.

@Jooho @pierDipi @spolti

pkg/constants/constants.go
```diff
@@ -38,7 +38,10 @@ const (
 	KnativeServingAPIGroupName       = KnativeServingAPIGroupNamePrefix + ".dev"
 )
 
-var KServeNamespace = getEnvOrDefault("POD_NAMESPACE", "kserve")
+var (
+	KServeNamespace              = getEnvOrDefault("POD_NAMESPACE", "kserve")
+	AutoscalerConfigmapNamespace = getEnvOrDefault("CONFIG_AUTOSCALER_NAMESPACE", DefaultKnServingNamespace)
```
pierDipi: nit: since we also leverage KEDA for autoscaling a more precise name would be `CONFIG_KNATIVE_AUTOSCALER_NAMESPACE`

pkg/constants/constants.go
```diff
@@ -388,6 +392,7 @@ const (
 
 var (
 	ServiceAnnotationDisallowedList = []string{
+		autoscaling.InitialScaleAnnotationKey,
```
pierDipi: Instead of disallowing it completely, should we allow overriding it while setting a default value if not?

brettmthompson: If we want to allow overriding of the initial scale value used (the default value being the min replicas requested) via the knative `initial-scale` annotation, I think we should also support allowing overriding of the initial scale using the global knative autoscaler configuration as well.

brettmthompson: I agree,  allowing end users to override the default initial scale functionality is a good thing to have. I will add this

pierDipi (commented):
/lgtm

spolti (approved):

terrytangyuan (dismissed):
/lgtm
/approve

terrytangyuan (approved):

yuzisun:
@brettmthompson Why `CheckZeroInitialScaleAllowed` is done deep in the reconciliation functions? this makes ksvc reconciliation unit tests having to depend on reading the knative configmap, and the k8s clientset has to be passed all the way down.

brettmthompson:
> @brettmthompson Why `CheckZeroInitialScaleAllowed` is done deep in the reconciliation functions? this makes ksvc reconciliation unit tests having to depend on reading the knative configmap, and the k8s clientset has to be passed all the way down.

@yuzisun thank you for the review, I understand your point. Would it be better to decouple the `CheckZeroInitialScaleAllowed` call from the `SetAutoScalingAnnotations` function and instead expose a new `zeroInitialScaleAllowed` boolean input variable? Then we could call `CheckZeroInitialScaleAllowed` earlier in the respective reconciliation loops and pass the function output, instead of the k8s client, all the way down.

yuzisun:
> > @brettmthompson Why `CheckZeroInitialScaleAllowed` is done deep in the reconciliation functions? this makes ksvc reconciliation unit tests having to depend on reading the knative configmap, and the k8s clientset has to be passed all the way down.
> 
> @yuzisun thank you for the review, I understand your point. Would it be better to decouple the `CheckZeroInitialScaleAllowed` call from the `SetAutoScalingAnnotations` function and instead expose a new `zeroInitialScaleAllowed` boolean input variable? Then we could call `CheckZeroInitialScaleAllowed` earlier in the respective reconciliation loops and pass the function output, instead of the k8s client, all the way down.

I think so that will make the code cleaner and easier to test.

brettmthompson:
> > > @brettmthompson Why `CheckZeroInitialScaleAllowed` is done deep in the reconciliation functions? this makes ksvc reconciliation unit tests having to depend on reading the knative configmap, and the k8s clientset has to be passed all the way down.
> > 
> > 
> > @yuzisun thank you for the review, I understand your point. Would it be better to decouple the `CheckZeroInitialScaleAllowed` call from the `SetAutoScalingAnnotations` function and instead expose a new `zeroInitialScaleAllowed` boolean input variable? Then we could call `CheckZeroInitialScaleAllowed` earlier in the respective reconciliation loops and pass the function output, instead of the k8s client, all the way down.
> 
> I think so that will make the code cleaner and easier to test.

Thank you for the advice. I will open a separate PR to implement this change.

PR #4421, opened 2025-04-24
Implement authorization for Raw InferenceGraphs

israel-hdez:

Authorization is implemented by using the TokenReview and the SubjectAccessReview Kubernetes APIs. A Middleware function is set up when some arguments are specified that trigger plugging-in the middleware func.

Some additional reconciliation is added toInferenceGraph controller to:
* Switch to a different ServiceAccount (in case the user hasn't customized it) so that privileges for using the cluster APIs are granted.
  * Creating the needed ServiceAccount for the auth-protected InferenceGraph to run.
* Managing a ClusterRoleBinding to give the required privileges for auth verification.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/apis/serving/v1alpha1/inference_graph.go
+++ b/pkg/apis/serving/v1alpha1/inference_graph.go
@@ -17,10 +17,14 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
+
+	"github.com/kserve/kserve/pkg/constants"
 )
 
 // InferenceGraph is the Schema for the InferenceGraph API for multiple models
@@ -334,4 +338,26 @@ type InferenceGraphList struct {
 
 func init() {
 	SchemeBuilder.Register(&InferenceGraph{}, &InferenceGraphList{})
+}
+
+func (ig *InferenceGraph) HasAuthEnabled() bool {
+	return ig.GetAnnotations()[constants.EnableAuthAnnotationKey] == "true"
+}
+
+// ResolveServiceAccountName returns the ServiceAccount name to use for the InferenceGraph pod.
+// If the user has customized the name of the ServiceAccount, the custom name is used. If there is no
+// custom ServiceAccount the returned name depends on whether the IG is auth-protected:
+//   - If the IG is *not* auth-protected, `default` is returned (backwards compatibility)
+//   - If the IG is *not* auth-protected, the returned value is the name of the IG with `-auth-verifier` suffix
+//     in a try to stay close to principle of least privilege.
+func (ig *InferenceGraph) ResolveServiceAccountName() string {
+	if len(ig.Spec.ServiceAccountName) != 0 {
+		return ig.Spec.ServiceAccountName
+	}
+
+	if ig.HasAuthEnabled() {
+		return fmt.Sprintf("%s-auth-verifier", ig.GetName())
+	}
+
+	return "default"
 }
--- a/pkg/apis/serving/v1alpha1/inference_graph.go
+++ b/pkg/apis/serving/v1alpha1/inference_graph.go
@@ -17,8 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"fmt"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
@@ -356,7 +354,7 @@ func (ig *InferenceGraph) ResolveServiceAccountName() string {
 	}
 
 	if ig.HasAuthEnabled() {
-		return fmt.Sprintf("%s-auth-verifier", ig.GetName())
+		return ig.GetName() + "-auth-verifier"
 	}
 
 	return "default"
--- a/pkg/controller/v1alpha1/inferencegraph/controller.go
+++ b/pkg/controller/v1alpha1/inferencegraph/controller.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 // +kubebuilder:rbac:groups=serving.kserve.io,resources=inferencegraphs;inferencegraphs/finalizers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=serving.kserve.io,resources=inferencegraphs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=create;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=create;get;update;patch,resourceNames=kserve-inferencegraph-auth-verifiers
 // +kubebuilder:rbac:groups=serving.knative.dev,resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=serving.knative.dev,resources=services/finalizers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=serving.knative.dev,resources=services/status,verbs=get;update;patch
@@ -25,6 +27,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -45,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/yaml"
 
 	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
@@ -67,8 +71,9 @@ type InferenceGraphReconciler struct {
 type InferenceGraphState string
 
 const (
-	InferenceGraphNotReadyState InferenceGraphState = "InferenceGraphNotReady"
-	InferenceGraphReadyState    InferenceGraphState = "InferenceGraphReady"
+	InferenceGraphControllerName string              = "inferencegraph-controller"
+	InferenceGraphNotReadyState  InferenceGraphState = "InferenceGraphNotReady"
+	InferenceGraphReadyState     InferenceGraphState = "InferenceGraphReady"
 )
 
 type RouterConfig struct {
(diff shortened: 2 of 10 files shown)


pkg/controller/v1alpha1/inferencegraph/raw_ig.go
```diff
@@ -188,6 +208,183 @@ func handleInferenceGraphRawDeployment(ctx context.Context, cl client.Client, cl
...
+
+	// Check if the SA is owned by the IG.
+	// TODO: For now, do a custom check. After updating to controller-runtime v0.20, change the check to use HasOwnerReference util func.
+	isOwner := false
+	saOwnerReference := existingServiceAccount.GetOwnerReferences()
+	if len(saOwnerReference) > 0 {
+		isOwner = saOwnerReference[0].APIVersion == graph.APIVersion &&
+			saOwnerReference[0].Kind == graph.Kind &&
+			saOwnerReference[0].Name == graph.GetName() &&
+			saOwnerReference[0].UID == graph.UID
```
pierDipi, line 367: Should we check all owner refs as there could be more than one in some cases and ours is not necessarily the "first" one to make this check more robust?

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -294,6 +341,16 @@ func inferenceGraphReadiness(status v1alpha1.InferenceGraphStatus) bool {
...
+	if err := removeAuthPrivilegesFromGraphServiceAccount(ctx, r.Clientset, graph); err != nil {
+		return err
+	}
+	if err := deleteGraphServiceAccount(ctx, r.Clientset, graph); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (r *InferenceGraphReconciler) SetupWithManager(mgr ctrl.Manager, deployConfig *v1beta1.DeployConfig) error {
```
pierDipi, line 354: Shouldn't we also "watch" and requeue associated InferenceGraph(s) when the newly reconciled / owned resources change (eg ServiceAccount, ClusterRoleBinding, etc)?

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -335,3 +337,25 @@ type InferenceGraphList struct {
...
+
+func (ig *InferenceGraph) HasAuthEnabled() bool {
+	return ig.GetAnnotations()[constants.EnableAuthAnnotationKey] == "true"
+}
+
+// ResolveServiceAccountName returns the ServiceAccount name to use for the InferenceGraph pod.
+// If the user has customized the name of the ServiceAccount, the custom name is used. If there is no
+// custom ServiceAccount the returned name depends on whether the IG is auth-protected:
+//   - If the IG is *not* auth-protected, `default` is returned (backwards compatibility)
+//   - If the IG is *not* auth-protected, the returned value is the name of the IG with `-auth-verifier` suffix
```
spolti, line 349: Maybe one is protected and the other is not?

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -335,3 +337,25 @@ type InferenceGraphList struct {
...
+//     in a try to stay close to principle of least privilege.
+func (ig *InferenceGraph) ResolveServiceAccountName() string {
+	if len(ig.Spec.ServiceAccountName) != 0 {
+		return ig.Spec.ServiceAccountName
```
spolti, line 353: At this point, I suppose the validation to make sure that the SA exists was already done when the pod was created, right?

pkg/controller/v1alpha1/inferencegraph/controller.go
spolti: While I think it should happen under normal circumstances, maybe, when security is involved, allowing it would open a security breach to updating the roles and the associated SAs, wdyt?

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -147,6 +152,42 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return reconcile.Result{}, err
 	}
+
+	// examine DeletionTimestamp to determine if object is under deletion
+	if graph.ObjectMeta.DeletionTimestamp.IsZero() {
+		// The object is not being deleted, so if it does not have our finalizer,
+		// then lets add the finalizer.
+		if !utils.Includes(graph.ObjectMeta.Finalizers, constants.InferenceGraphFinalizerName) {
+			graph.ObjectMeta.Finalizers = append(graph.ObjectMeta.Finalizers, constants.InferenceGraphFinalizerName)
+			patchYaml := "metadata:\n  finalizers: [" + strings.Join(graph.ObjectMeta.Finalizers, ",") + "]"
```
spolti, line 162: Can't we move lines 162 to 166 to function? As this very same piece of code used twice here, it would be better, wdyt?

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
```diff
@@ -98,6 +104,20 @@ func createInferenceGraphPodSpec(graph *v1alpha1.InferenceGraph, config *RouterC
...
+	// * Add the --inferencegraph-name argument, so that the router is aware of its name
+	// * Enable auto-mount of the ServiceAccount, because it is required for validating tokens
+	// * Set a non-default ServiceAccount with enough privileges to verify auth
+	if graph.HasAuthEnabled() {
+		podSpec.Containers[0].Args = append(podSpec.Containers[0].Args, "--enable-auth")
+
+		podSpec.Containers[0].Args = append(podSpec.Containers[0].Args, "--inferencegraph-name")
+		podSpec.Containers[0].Args = append(podSpec.Containers[0].Args, graph.GetName())
+
+		podSpec.AutomountServiceAccountToken = proto.Bool(true)
```
spolti, line 118: I think I've already asked you about it, but is the auto mount required for IG?
If so, would you mind adding a comment here explaining why?

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
```diff
@@ -188,6 +208,183 @@ func handleInferenceGraphRawDeployment(ctx context.Context, cl client.Client, cl
 	return deployment[0], reconciler.URL, nil
 }
 
+func handleInferenceGraphRawAuthResources(ctx context.Context, clientset kubernetes.Interface, scheme *runtime.Scheme, graph *v1alpha1.InferenceGraph) error {
```
spolti, line 211: maybe add a comment on this and the next func, wdyt?

spolti: the next ones too :)

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
spolti: May we create an issue to now forget about this? :)


Is there a test case for custom SAs?

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
```diff
@@ -188,6 +208,183 @@ func handleInferenceGraphRawDeployment(ctx context.Context, cl client.Client, cl
...
+				return errors.Wrapf(err, "fails to bind privileges for auth verification to inference graph")
+			}
+		}
+	}
+
+	return nil
+}
+
+func removeAuthPrivilegesFromGraphServiceAccount(ctx context.Context, clientset kubernetes.Interface, graph *v1alpha1.InferenceGraph) error {
+	clusterRole, err := clientset.RbacV1().ClusterRoleBindings().Get(ctx, constants.InferenceGraphAuthCRBName, metav1.GetOptions{})
```
spolti, line 317: Is clusterRole mandatory for IG?

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
israel-hdez: > Should we check all owner refs as there could be more than one in some cases and ours is not necessarily the "first" one to make this check more robust?

@pierDipi my thinking is that when kserve-controller creates the SA inside `handleInferenceGraphRawAuthResources`, for the time being we would expect the InferenceGraph to be the only owner (until some other controller starts operating on the SA). Still, we can make it more robust if you think it is worth it.

BTW, once the upgrade to controller-runtime v0.20 can happen, we can remove our own implementation and the `HasOwnerReference` func would cover that case.

> Is there a test case for custom SAs?

Hmm...  @spolti  I think no. I'll add some.

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: > Shouldn't we also "watch" and requeue associated InferenceGraph(s) when the newly reconciled / owned resources change (eg ServiceAccount, ClusterRoleBinding, etc)?

I can agree that watching the ServiceAccount is desirable. I'll do this.

The ClusterRoleBinding is where I have doubts... I don't know if we can watch a cluster-scoped resource (I haven't done this). In case we can, given it is shared between all IGs, I wonder if it would lead to many reconciles? I'd assume a reconcile that modifies the CRB would cause enqueuing a reconcile of the InferenceGraphs.


> While I think it should happen under normal circumstances, maybe, when security is involved, allowing it would open a security breach to updating the roles and the associated SAs, wdyt?

@spolti I think I'm confused and I don't think I understand. Can you elaborate?

pkg/apis/serving/v1alpha1/inference_graph.go
israel-hdez: Yes, the second bullet is for auth-protected ones. I'll fix it and remove the `*not*`

pkg/apis/serving/v1alpha1/inference_graph.go
israel-hdez: I don't think there is a validation for this. It is probably user responsibility doing that, if they are customizing the SA.

pkg/controller/v1alpha1/inferencegraph/controller.go
israel-hdez: This makes sense. I'll do.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
israel-hdez: When auth is enabled, yes. Because the kserve-router will be doing calls to the cluster API. I'll add the comment.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
israel-hdez: Yes, I'll add comments.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
israel-hdez: For checking authentication, yes. Because the TokenReview and SubjectAccessReview APIs are cluster-scoped.

spolti: okay, thanks.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
spolti: Thanks.

pkg/controller/v1alpha1/inferencegraph/raw_ig.go
spolti: Sounds good, thanks.

pkg/apis/serving/v1alpha1/inference_graph.go
spolti: yeah, agreed.

pkg/controller/v1alpha1/inferencegraph/controller.go
spolti: I meant, by allowing the SA and CRB be reconciled, we would allow the user to use a SA with higher privileges. Maybe we could scope the needed privileges and don't accept anything greater than it.

pierDipi: @spolti wouldn't that be true even if we don't watch them?

> The ClusterRoleBinding is where I have doubts... I don't know if we can watch a cluster-scoped resource (I haven't done this). In case we can, given it is shared between all IGs, I wonder if it would lead to many reconciles? I'd assume a reconcile that modifies the CRB would cause enqueuing a reconcile of the InferenceGraphs.

It is possible to watch cluster-scoped resources, in this case, we can re-queue only one, however, generally we should ensure that "no op" reconciliations are "cheap" and cache-based without external requests, but that's something we can follow up.

Generally, the issue I'm trying to cover is when something else updates our own controlled resources and drift from the expected ones we would need to react and adjust ASAP

PR #4422, opened 2025-04-24
Add support for writing model request/response logs to object stores

### LINKED ISSUE #4380: Add support for writing model request/response logs to object stores

cjohannsen-cloudera:

Storage is defined using the StorageSpec.  The existing `v1beta1.StorageSpec` has been refactored to extract a base struct.  
```
// StorageSpec defines a spec for an object in an object store
type StorageSpec struct {
	// The path to the object in the storage. Note that this path is relative to the storage URI.
	// +optional
	Path *string `json:"path,omitempty"`
	// Parameters to override the default storage credentials and config.
	// +optional
	Parameters *map[string]string `json:"parameters,omitempty"`
	// The Storage Key in the secret for this object.
	// +optional
	StorageKey *string `json:"key,omitempty"`
}
```

The existing `StorageSpec` is now `ModelStorageSpec` and adds the `SchemaPath`
```
type ModelStorageSpec struct {
	StorageSpec `json:",inline"`
	// The path to the model schema file in the storage.
	// +optional
	SchemaPath *string `json:"schemaPath,omitempty"`
}
```

This is a pure refactor.

The LoggerConfig now contains a StorageSpec that defines the logger storage output location and a service account name used to fetch credentials.

```
  logger: |-
    {
        "image" : "kserve/agent:latest",
        "memoryRequest": "100Mi",
        "memoryLimit": "1Gi",
        "cpuRequest": "100m",
        "cpuLimit": "1",
        "defaultUrl": "http://default-broker",
        "storage": {
            "path": "",
            "parameters": {
              "type": "s3",
              "format": "json" 
            },
            "key": ""
            "serviceAccountName": "",
        }
    }
```

Three new command line parameters have been added to the agent to allow the storage configuration to be passed in when injecting the agent.
```
--log-store-path [path] - path within the storage location to write the logger output
--log-store-format - output format of the stored data, one of "json" or "yaml", default "json"
```

By default the logger service account is undefined, and the logger storage feature is disabled.

If the service account name is defined, the logger will attempt to look for a Service Account with the matching name and mount the secrets it contains.  When the agent is launched the secrets are available in the environment.

When an inference occurs the S3 client is constructed and the request and response are stored to the blob store path specified above.  The agent uses the existing client defined for the model downloader to simplify integration and configuration, enhanced to allow for uploads.

Which issue(s) this PR fixes

Fixes https://github.com/kserve/kserve/issues/4380

**Type of changes**

[x] Added additional test to logger/handler_test.go
[x] Added new test, logger/store_test.go
[x] Added new agent command argument to agent_injector_test.go

To test locally create a secret in the  `default` namespace
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    serving.kserve.io/s3-region: [YOUR_REGION]
  name: agent-logger-secret
  namespace: default
type: Opaque
data:
  AWS_ACCESS_KEY_ID: [YOUR_ACCESS_KEY_ID]
  AWS_DEFAULT_REGION: [YOUR_REGION]
  AWS_SECRET_ACCESS_KEY: [YOUR_SECRET_ACCESS_KEY]
```

Create a service account for the secret:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: logger-sa
  namespace: default
secrets:
- name: agent-logger-secret
```

To test the feature, deploy a model with inference logging: 
```
kind: InferenceService
metadata:
  name: sklearn-iris
spec:
  predictor:
    logger:
      mode: all
      url: s3://[YOUR_BUCKET]/
      storage:
        path: "/logger"
        parameters:
          type: s3
          format: json
        key: credentials
    model:
      modelFormat:
        name: sklearn
```

Send a prediction to the endpoint and the request and response are stored as json in the blob storage location `[YOUR_BUCKET]/logger/`

**Release note**:

```release-note

```

**Re-running failed tests**


spolti:
can you please rebase your pull request?

pkg/logger/store.go
```diff
@@ -0,0 +1,214 @@
...
+	if strings.HasPrefix(url, "http") {
+		return HttpStorage
+	}
+
+	// s3, s3a
+	if strings.HasPrefix(url, "s3") {
+		return S3Storage
+	}
+
+	return DefaultStorage
```
gavrissh: Better to have a switch case?
```suggestion
	switch {
	
	case strings.HasPrefix(url, "http"):  // http, https
		return HttpStorage
		
	case strings.HasPrefix(url, "s3"): // s3, s3a
		return S3Storage
		
	default:
		return DefaultStorage
	}
```

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: kserve/kserve-controller:latest
+      - image: 192.168.1.171:6000/kserve-controller:latest
```
gavrissh: Something used for testing purposes?

pkg/logger/types.go
```diff
@@ -36,3 +36,5 @@ type LogRequest struct {
 	CertName         string
 	TlsSkipVerify    bool
 }
+
+const DefaultStoreFormat string = "json"
```
gavrissh: Is this being used?

config/configmap/inferenceservice.yaml
```diff
@@ -664,7 +664,7 @@ data:
         "memoryLimit": "1Gi",
         "cpuRequest": "100m",
         "cpuLimit": "1",
-        "defaultUrl": "http://default-broker"
+        "defaultUrl": "http://default-broker",
```
sivanantha321:
```suggestion
        "defaultUrl": "http://default-broker"
```

hack/image_patch_dev.sh
```diff
@@ -77,7 +77,15 @@ data:
         "memoryRequest": "100Mi",
         "memoryLimit": "100Mi",
         "cpuRequest": "100m",
-        "cpuLimit": "100m"
+        "cpuLimit": "100m",
+        "storage": {
```
sivanantha321: for testing ?

pkg/apis/serving/v1beta1/component_test.go
```diff
@@ -61,53 +61,51 @@ func TestComponentExtensionSpec_Validate(t *testing.T) {
...
-			spec: &StorageSpec{
-				Parameters: &map[string]string{
-					"type": "gs",
-				},
-			},
+			spec:       &ModelStorageSpec{},
 			storageUri: proto.String("gs://test/model"),
 			matcher:    gomega.MatchError(fmt.Errorf(UnsupportedStorageURIFormatError, strings.Join(SupportedStorageSpecURIPrefixList, ", "), "gs://test/model")),
 		},
-		"InvalidStoragespec": {
```
sivanantha321, line 102: Any reason to remove this test case ?

pkg/logger/worker.go
```diff
@@ -23,15 +23,13 @@ import (
...
-	"net/http"
-	"os"
-	"path/filepath"
-
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
-	"go.uber.org/zap"
-
 	"github.com/kserve/kserve/pkg/constants"
+	"go.uber.org/zap"
```
sivanantha321: Probably need to run `make precommit`

pkg/logger/types.go
cjohannsen-cloudera: Removed

hack/image_patch_dev.sh
cjohannsen-cloudera: Removed

pkg/apis/serving/v1beta1/component_test.go
cjohannsen-cloudera: No purely accidental when cleaning up, adding back

pkg/logger/worker.go
cjohannsen-cloudera: 👍

config/default/manager_image_patch.yaml
cjohannsen-cloudera: Yeah, will roll back

PR #4429, opened 2025-04-25
InferenceGraph: Fix response code when condition step is not fulfilled

israel-hdez:

When a step in the InferenceGraph has a condition, if it is not fulfilled for a run, such step is skipped. Currently, when the step is skipped, the returned HTTP status code is 500. This is not compatible with KServe's python client which raises an exception when receiving the 500 status code.

This is changing the status code to 200 to let the python client to work correctly. It also makes more sense to return 200, since there were no errors processing the request.

**Type of changes**

- Setup the image classification example: https://kserve.github.io/website/latest/modelserving/inference_graph/image_pipeline/.
- Test inference using the cat sample. Observe a 200 (and no 500) status code is returned in HTTP headers.

**Checklist**:

**Release note**:

```release-note
Fix response code of InferenceGraph when a condition in a Sequence is not fulfilled.
```

**Re-running failed tests**

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -322,7 +322,7 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
 				}
 				// if the condition does not match for the step in the sequence we stop and return the response
 				if !gjson.GetBytes(responseBytes, step.Condition).Exists() {
-					return responseBytes, 500, nil
+					return responseBytes, 200, nil
 				}
 			}
 			if responseBytes, statusCode, err = executeStep(step, graph, request, headers); err != nil {
--- a/cmd/router/main_test.go
+++ b/cmd/router/main_test.go
@@ -451,6 +451,94 @@ func TestInferenceGraphWithCondition(t *testing.T) {
 	assert.Equal(t, expectedModel4Response, response["model4"])
 }
 
+func TestInferenceGraphSequenceWithUnmetCondition(t *testing.T) {
+	// Start a local HTTP server
+	model1 := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		_, err := io.ReadAll(req.Body)
+		if err != nil {
+			return
+		}
+		response := map[string]interface{}{
+			"predictions": []map[string]interface{}{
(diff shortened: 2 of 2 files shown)


cmd/router/main.go
```diff
@@ -322,7 +322,7 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
 				}
 				// if the condition does not match for the step in the sequence we stop and return the response
 				if !gjson.GetBytes(responseBytes, step.Condition).Exists() {
```
yuzisun, line 324: Add a test case?

cmd/router/main.go
```diff
@@ -322,7 +322,7 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
-					return responseBytes, 500, nil
+					return responseBytes, 200, nil
```
spolti, line 325: what assures that the response code is always 200?

israel-hdez: Within this same `main.go` file, the `routeStep` func is invoked in `graphHandler` which basically uses the returned stauts code from `routeStep` and writes it to the HTTP response here: https://github.com/kserve/kserve/blob/cb755147a7f1aa38bfd812ea85cf0c6beac3be9c/cmd/router/main.go#L380-L383

Observe the `w.WriteHeader(statusCode) ` line.

The `routeStep` can be called recursively in `executeStep`: https://github.com/kserve/kserve/blob/cb755147a7f1aa38bfd812ea85cf0c6beac3be9c/cmd/router/main.go#L356-L362

Still, the value is returned as is in the recursion. So, the returned `200` will bubble up.

cmd/router/main.go
israel-hdez: Test case is added.

israel-hdez:
CI issues are fixed in #4438

Jooho:
@israel-hdez the pr merged so please rebase your code and trigger the ci again.

spolti (approved):

israel-hdez:
@Jooho @spolti @yuzisun I rebased and now CI is passing.

spolti (approved):

PR #4440, opened 2025-05-02
fix: 4439 - update telepresence-setup.sh to adopt tooling changes

### LINKED ISSUE #4439: Telepresence debugging not operational
/kind bug

Running the script `hack/telepresence-setup.sh` errors out attempting to establish the intercept

The intercept is established and the manager can be run via telepresence

This is purely a development tooling issue, not a bug in kserve

cjohannsen-cloudera:

The telepresence debugging script is out-of-date and not currently functional.  This PR adopts changes required to enable telepresence again.

Fixes #4439

**Type of changes**

1. Run `hack/telepresence-setup.sh`
2. Run the manager locally
3. Issue an API request and observe that the local application intercepts the call

**Re-running failed tests**


hack/telepresence-setup.sh
```diff
@@ -25,6 +25,38 @@ if [[ "$1" == "uninstall" ]]; then
...
+            "op": "replace",
+            "path": "/spec/template/spec/containers/0/securityContext",
+            "value": {
+              "allowPrivilegeEscalation": false,
+              "capabilities": {
+                "drop": ["ALL"]
+              },
+              "privileged": false,
+              "readOnlyRootFilesystem": true,
+              "runAsNonRoot": true
```
yuzisun, line 43: These are configurable from the helm chart
https://github.com/kserve/kserve/blob/master/charts/kserve-resources/values.yaml#L101

cjohannsen-cloudera: @yuzisun What is the best way to set the values when in debugging mode so that they are not applied otherwise?

cjohannsen-cloudera: This is replacing the original settings that have to be modified to allow telepresense to run, back to the chart values that were present prior to starting debugging.  If there's a simpler/cleaner way please let me know.

sivanantha321:
@cjohannsen-cloudera Can you help rebase ?

PR #4441, opened 2025-05-03
Allow OCI for multi-node/multi-gpu

israel-hdez:

This modifies the InferenceService validation of the multi-node/multi-gpu case to allow OCI in the storageUri additionally to the already allowed PVC storage.

**Type of changes**

- [ ] Test A
- [ ] Test B

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


israel-hdez:
@Jooho May you review?

config/runtimes/kserve-huggingfaceserver-multinode.yaml
```diff
@@ -21,17 +21,22 @@ spec:
       - "bash"
       - "-c"
       - |
-        export MODEL=${MODEL_ID}
+        export MODEL_DIR_ARG=""
         if [[ ! -z ${MODEL_DIR} ]]
```
Jooho: It checks the same value(`MODEL_DIR`), is it right? should be `MODEL_ID`?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -45,8 +45,7 @@ const (
 	InvalidProtocol                                  = "invalid protocol %s. Must be one of [%s]"
 	MissingStorageURI                                = "the InferenceService %q is invalid: StorageURI must be set for multinode enabled"
 	InvalidAutoScalerError                           = "the InferenceService %q is invalid: Multinode only supports 'none' autoscaler(%s)"
-	InvalidNotSupportedStorageURIProtocolError       = "the InferenceService %q is invalid: Multinode only supports 'pvc' Storage Protocol(%s)"
-	InvalidCustomGPUTypesAnnotationFormatError       = "the InferenceService %q is invalid: invalid format for %s annotation: must be a valid JSON array"
```
Jooho, line 49: wonder why `InvalidCustomGPUTypesAnnotationFormatError` needs to be removed?

israel-hdez: Simple cleanup. The constant is unused.

config/runtimes/kserve-huggingfaceserver-multinode.yaml
israel-hdez: I fixed.

pkg/apis/serving/v1beta1/component.go
Jooho: sounds good

Jooho (commented):
/lgtm

config/runtimes/kserve-huggingfaceserver-multinode.yaml
```diff
@@ -21,17 +21,22 @@ spec:
       - "bash"
       - "-c"
       - |
-        export MODEL=${MODEL_ID}
+        export MODEL_DIR_ARG=""
+        if [[ ! -z ${MODEL_ID} ]]
+        then
+          export MODEL_DIR_ARG="--model_dir=${MODEL_ID}"
```
Jooho, line 27: I'm not sure if using model_id to dynamically download models from HuggingFace is the best approach, especially for multi-node/multi-GPU setups. In such cases, the models are typically large, and downloading them on each node can take significant time.

This concern is somewhat outside the main scope of this PR, so it might make sense to address it in a separate PR.
However, since this PR already touches that part of the code, it would be great if we could improve it here :)

Also, I suggest using a generic variable like `MODEL_ARG`, since `--model_id` and `--model_dir` are mutually exclusive
```
parser.add_argument(
    "--model_dir",
    required=False,
    default="/mnt/models",
    help="A URI pointer to the model binary",
)
parser.add_argument(
    "--model_id", required=False, default=None, help="Huggingface model id"
)
```
https://github.com/kserve/kserve/blob/master/python/huggingfaceserver/huggingfaceserver/__main__.py#L79C1-L87C2

Like the following 
```
  export MODEL_ARG=""
  if [[ ! -z ${MODEL_ID} ]]
        then
          export MODEL_ARG="--model_id=${MODEL_ID}"

   if [[ ! -z ${MODEL_DIR} ]]
        then
          export MODEL_ARG="--model_dir=${MODEL_DIR}"
        fi

python -m huggingfaceserver ${MODEL_ARG} --tensor-parallel-size=${TENSOR_PARALLEL_SIZE} --pipeline-parallel-size=${PIPELINE_PARALLEL_SIZE} $0 $@
```

israel-hdez: > This concern is somewhat outside the main scope of this PR, so it might make sense to address it in a separate PR.

With the changes in this PR, I basically added OCI support which (IIRC) would not pass `MODEL_ID` nor `MODEL_DIR` environment variables. So, I added that case while keeping existing behavior.

But I do agree with you comment. We have the validating webhook which won't allow configuring a `hf://` url. So, the HF case shouldn't happen and this code may not be relevant in the `ClusterServingRuntime`. I prefer we should do changes in a new PR, since I would need to better understand the difference between `--model_id` and `--model_dir`.

Jooho: sounds good

PR #3786, opened 2024-07-10
Fix issue with rolling update behavior

andyi2it:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3768 

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

**Feature/Issue validation/testing**:

Please describe the tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:

```release-note

```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

PR #4445, opened 2025-05-05
Fix raw deployment update 

andyi2it:

Fix the side effect introduced by https://github.com/kserve/kserve/pull/3786 that the strategic merge does not remove the fields during update, also simplify the fix from https://github.com/kserve/kserve/pull/4297.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
@@ -406,7 +406,6 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deploym
 			// To avoid the conflict between HPA and Deployment,
 			// we need to remove the Replicas field from the deployment spec
 			// For none autoscaler, it should not remove replicas
-			modDeployment := deployment.DeepCopy()
 			if modDeployment.Annotations[constants.AutoscalerClass] != string(constants.AutoscalerClassNone) {
 				modDeployment.Spec.Replicas = nil
 				curDeployment.Spec.Replicas = nil
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
@@ -387,9 +387,9 @@ func addGPUResourceToDeployment(deployment *appsv1.Deployment, targetContainerNa
 
 // Reconcile ...
 func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deployment, error) {
-	for _, deployment := range r.DeploymentList {
+	for _, desiredDep := range r.DeploymentList {
 		// Reconcile Deployment
-		checkResult, _, err := r.checkDeploymentExist(ctx, r.client, deployment)
+		checkResult, existingDeployment, err := r.checkDeploymentExist(ctx, r.client, desiredDep)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -398,33 +398,36 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deploym
...
+				return nil, err
 			}
 
 			modJson, err := json.Marshal(modDeployment)
 			if err != nil {
 				return nil, err
 			}
 			// Generate the strategic merge patch between the current and modified JSON
-			patchByte, err := strategicpatch.StrategicMergePatch(curJson, modJson, appsv1.Deployment{})
+			patchByte, err := strategicpatch.CreateTwoWayMergePatch(curJson, modJson, appsv1.Deployment{})
```
yuzisun, line 424: see https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment. 

How `strategicpatch.CreateTwoWayMergePatch` is different from `client.MergeFrom()` which returns the merge patch object ?

andyi2it: TwoWayMergePatch replaces a field in the original yaml if its an array whereas client.MergeFrom only adds new values to the field in the original because the server applies `strategicMergePatch`.

Eg:
```yaml
env:
 - name: original1
   value: original1value
 - name: original2
   value: original2value
```

If the modified yaml is like below.

```yaml
env:
  - name: original1
    value: original1value
  - name: new
    value: newval
```

then the 2 way merge will create

```yaml
env:
  - name: original1
    value: original1value
  - name: new
    value: newval
```
StrtegicMergePatch adds/updates in case of arrays whereas 2 way merge patch will replace entire array if there is change in any one value or field in array.

andyi2it: @yuzisun 
As a matter of fact there is no difference between the `client.StrategicMergeFrom` vs what we do with `CreateTwoWayMergePatch`, the only difference is we get hold of the generated patch and can be used for printing the diff and In future if we want to skip any specific field we have the capability to manipulate while when we directly merge using inbuilt functionality, we will not be.

yuzisun: Can you help test the case when fields are updated/removed, for example replacing tcp readiness probe kserve added by default to http readiness probe.

yuzisun: Cc @spolti to review

spolti: @andyi2it, does it work for removing envs or updating secrets?
Please take a look at this PR:  https://github.com/kserve/kserve/pull/4297

I guess this would fix the envs removal as well. 
The `CreateTwoWayMergePatch` treats lists as generic JSON objects.

Could you copy the unit test I added in the other PR here? That tests the addition and removal of envs.

andyi2it: @spolti 
Yes, I have added an e2e to test removal of envs but will add an unit test as well.

yuzisun (approved):

PR #4451, opened 2025-05-06
fix: Allow CA bundle path without config map

fabiendupont:

In the previous code, the AWS_CA_BUNDLE environment variable is only considered if the CA_BUNDLE_CONFIGMAP_NAME environment variable is set, while the doc doesn't explicitely tie they together. This change allows using AWS_CA_BUNDLE independently to leverage a CA bundle certificate that would be preexisting in the image or injected by Kubernetes in a known path.

One use case is leveraging OpenShift service serving certificates, which injects the service CA in the containers via a projected ConfigMap. The service CA is always available in
/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt.

The documentation already says that this variable can be used to set the path to the CA bundle, while the ConfigMap uses a static file name (cabundle.crt) located in the ConfigMap volume mount point.

**Type of changes**

### DIFF
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -227,8 +227,8 @@ class Storage(object):
                 global_ca_bundle_volume_mount_path = os.getenv(
                     "CA_BUNDLE_VOLUME_MOUNT_POINT"
                 )
-                ca_bundle_full_path = (
-                    global_ca_bundle_volume_mount_path + "/cabundle.crt"
+                ca_bundle_full_path = os.path.join(
+                    global_ca_bundle_volume_mount_path, "cabundle.crt"
                 )
             if ca_bundle_set:
                 if os.path.exists(ca_bundle_full_path):
--- a/python/kserve/kserve/storage/storage.py
+++ b/python/kserve/kserve/storage/storage.py
@@ -211,19 +211,26 @@ class Storage(object):
             verify_ssl = True
 
         # If verify_ssl is true, then check there is custom ca bundle cert
+        # The CA bundle can be any local file in the container under the path
+        # set in the AWS_CA_BUNDLE environment variable.
+        # It can also be coming from a ConfigMap, in which case the filename
+        # is cabundle.crt.
         if verify_ssl:
             global_ca_bundle_configmap = os.getenv("CA_BUNDLE_CONFIGMAP_NAME")
-            if global_ca_bundle_configmap:
-                isvc_aws_ca_bundle_path = os.getenv("AWS_CA_BUNDLE")
-                if isvc_aws_ca_bundle_path and isvc_aws_ca_bundle_path != "":
-                    ca_bundle_full_path = isvc_aws_ca_bundle_path
-                else:
-                    global_ca_bundle_volume_mount_path = os.getenv(
-                        "CA_BUNDLE_VOLUME_MOUNT_POINT"
-                    )
-                    ca_bundle_full_path = (
-                        global_ca_bundle_volume_mount_path + "/cabundle.crt"
-                    )
+            isvc_aws_ca_bundle_path = os.getenv("AWS_CA_BUNDLE")
+            ca_bundle_set = False
+            if isvc_aws_ca_bundle_path and isvc_aws_ca_bundle_path != "":
+                ca_bundle_set = True
+                ca_bundle_full_path = isvc_aws_ca_bundle_path
+            elif global_ca_bundle_configmap:
+                ca_bundle_set = True
+                global_ca_bundle_volume_mount_path = os.getenv(
+                    "CA_BUNDLE_VOLUME_MOUNT_POINT"
+                )
+                ca_bundle_full_path = (
+                    global_ca_bundle_volume_mount_path + "/cabundle.crt"
+                )
+            if ca_bundle_set:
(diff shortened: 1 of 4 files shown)


python/kserve/kserve/storage/storage.py
```diff
@@ -211,19 +211,25 @@ def _download_s3(uri, temp_dir: str) -> str:
...
-                if isvc_aws_ca_bundle_path and isvc_aws_ca_bundle_path != "":
-                    ca_bundle_full_path = isvc_aws_ca_bundle_path
-                else:
-                    global_ca_bundle_volume_mount_path = os.getenv(
-                        "CA_BUNDLE_VOLUME_MOUNT_POINT"
-                    )
-                    ca_bundle_full_path = (
-                        global_ca_bundle_volume_mount_path + "/cabundle.crt"
-                    )
+            isvc_aws_ca_bundle_path = os.getenv("AWS_CA_BUNDLE")
```
Jooho, line 220: The AWS_CA_BUNDLE environment variable is used to specify the path to the CA file, under the assumption that the CA is already present inside the pod. However, shouldn’t there be a logic to check whether this file actually exists?

Also, in the OpenShift use case, it's said that the service-ca.crt file is always present, but in reality, this is controlled by configuration. For example, in this [link](https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L721), if the value is set to false, the file is not attached to the pod. Our team discussed whether we really can remove this file from runtime. @VedantMahabaleshwarkar can you please share the result ?

fabiendupont: There is a check for the file existence lower. I am just splitting the logic between `AWS_CA_BUNDLE` and `CA_BUNDLE_VOLUME_MOUNT_POINT` environment variables. Then, the `if os.path.exists(ca_bundle_full_path):` block is checking the file.

Jooho: Yes I missed the part! then I am ok with this change :)

Jooho (dismissed):
/lgtm

sivanantha321:
@fabiendupont Can you add a unit test for this ?

fabiendupont:
> @fabiendupont Can you add a unit test for this ?

@sivanantha321, I didn't find existing tests for S3, so it would mean writing the whole test coverage for S3 and not just for this change. Sounds like something that would deserve its own pull request, no?

fabiendupont:
@sivanantha321, @Jooho, I added tests and the CI is green.

python/kserve/kserve/storage/test/test_s3_storage.py
Jooho, line 1: With [this PR](https://github.com/kserve/kserve/pull/4391), the storage was segregated. 
So I don't think we need this anymore in kserve folder. @spolti am I right?

python/kserve/kserve/storage/storage.py
```diff
@@ -211,19 +211,26 @@ def _download_s3(uri, temp_dir: str) -> str:
...
+            ca_bundle_set = False
+            if isvc_aws_ca_bundle_path and isvc_aws_ca_bundle_path != "":
```
Jooho, line 222:
```suggestion
            if isvc_aws_ca_bundle_path:
```

python/kserve/kserve/storage/storage.py
```diff
@@ -211,19 +211,26 @@ def _download_s3(uri, temp_dir: str) -> str:
...
+                ca_bundle_set = True
+                ca_bundle_full_path = isvc_aws_ca_bundle_path
+            elif global_ca_bundle_configmap:
+                ca_bundle_set = True
+                global_ca_bundle_volume_mount_path = os.getenv(
+                    "CA_BUNDLE_VOLUME_MOUNT_POINT"
+                )
+                ca_bundle_full_path = (
+                    global_ca_bundle_volume_mount_path + "/cabundle.crt"
+                )
```
Jooho:
```suggestion
                ca_bundle_full_path = os.path.join(global_ca_bundle_volume_mount_path, "cabundle.crt")
```

would be better to use `os.path`

Jooho (commented):
Thanks for your contribution again!.

@spolti is currently on holiday and unable to review my comment at the moment. Since having test cases in two places isn’t a major issue, I don’t consider this a blocker. If @spolti thinks this part is unnecessary, we can clean it up later.

I just left nit comments :) 
/lgtm

python/kserve/kserve/storage/test/test_s3_storage.py
fabiendupont: But as long as there are tests in kserve/kserve/storage, shouldn't we keep them in sync? I would expect the whole kserve/kserve/storage to be removed at once when all the code is using only kserve/storage, no?

python/kserve/kserve/storage/storage.py
fabiendupont: Thanks for the suggestion. I used it while rebasing.

python/kserve/kserve/storage/test/test_s3_storage.py
Jooho: right so I don't see any issues to have them now in both places :+1:

Jooho (dismissed):
/lgtm

sivanantha321:
Thanks for the contribution !
/lgtm

sivanantha321 (approved):

PR #4455, opened 2025-05-09
Stop and resume a model [Raw Deployment]

### LINKED ISSUE #4207: Support stop InferenceService

hdefazio:
This adds the ability to stop and resume services in raw deployment mode without permanently deleting configurations or incurring delays caused by gradual scaling processes

This PR also makes some minor changes to the work done in #4337 

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes https://github.com/kserve/kserve/issues/4207

**Type of changes**

cluster:
Client Version: v1.30.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.15+ff493be

Test setup:
```
$ kubectl create ns stop-resume-demo
$ kubectl config set-context --current --namespace stop-resume-demo 
$ kustomize build ./config/runtimes/| sed 's/ClusterServingRuntime/ServingRuntime/g' |kubectl create -n stop-resume-demo -f -

$ podman build -t kserve:stop-resume . 
$ podman push kserve:stop-resume  quay.io/hdefazio/kserve:stop-resume

Edited kserve-controller-manager to use quay.io/hdefazio/kserve:stop-resume
```

Test: 

1. Normal deployment --> add the annotation

create a isvc without the annotation 
```
kind: InferenceService
metadata:
  annotations:
    serving.kserve.io/deploymentMode: 'RawDeployment'
  name: "sklearndemo1"
  namespace: stop-resume-demo
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
EOF
```

observe that the inference service is Ready
```
$ kubectl get isvc -n stop-resume-demo
NAME           URL                                                                                         READY   PREV   LATEST   PREVROLLEDOUTREVISION   LATESTREADYREVISION            AGE
sklearndemo1   https://sklearndemo1-stop-resume-demo.apps.rosa.han-stop-resume.wtsd.p3.openshiftapps.com   True           100                              sklearndemo1-predictor-00002   2m8s
```

```
$ kubectl get pods -n stop-resume-demo
NAME                                                       READY   STATUS    RESTARTS   AGE
sklearndemo1-predictor-00002-deployment-679bd4754f-mwvsv   3/3     Running   0          4m50s
```

Stop the model
```
$ kubectl edit isvc/sklearndemo1 -n stop-resume-demo 
Add annotation `serving.kserve.io/stop: 'true'`
```

It should remove all objects except the isvc

Observe that the ISVC now has a Stopped condition
```
kubectl describe isvc/sklearndemo1 -n stop-resume-demo 
```

2. Stopped model--> set the annotation to false
Resume the model
```
$ kubectl edit isvc/sklearndemo1 -n stop-resume-demo 
Set annotation `serving.kserve.io/stop: 'false'`
```

Observe that the deployment, and pods are created and ready and that all conditions on the ISVC are Ready

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


johnugeorge:
@hdefazio Can you change the references to KServe issues?

hdefazio:
@johnugeorge Just updated!

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -194,43 +194,47 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+		isvc.Status.SetCondition(v1beta1.Stopped, stopped_condition)
+
+		return ctrl.Result{}, nil
+	} else {
+		resume_condition := &apis.Condition{
+			Type:   v1beta1.Stopped,
+			Status: corev1.ConditionFalse,
 		}
+		isvc.Status.SetCondition(v1beta1.Stopped, resume_condition)
 	}
```
pierDipi: I know this is in the existing branch but snakeCase for variable names is inconsistent with the camelCase used everywhere else

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -704,45 +725,65 @@ func (r *RawHTTPRouteReconciler) reconcileTopLevelHTTPRoute(ctx context.Context,
...
+			if existingHttpRoute.OwnerReferences[0].UID == isvc.UID && existingHttpRoute.GetDeletionTimestamp() == nil {
+				log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated top level http route")
+				if err := r.client.Delete(ctx, existingHttpRoute); err != nil {
+					return err
+				}
+			}
+		} else if !apierr.IsNotFound(getExistingErr) {
+			return getExistingErr
 		}
 	}
```
pierDipi: There is a deep level of `if / else` nesting which makes the logic hard to follow, can we restructure it to reduce nesting? (maybe with early return, separate methods/functions)

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -194,43 +194,47 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
-			stopped_condition := &apis.Condition{
-				Type:   v1beta1.Stopped,
-				Status: corev1.ConditionTrue,
-			}
-			isvc.Status.SetCondition(v1beta1.Stopped, stopped_condition)
+		// Set the ready condition
+		predictor_ready_condition := &apis.Condition{
+			Type:   v1beta1.PredictorReady,
+			Status: corev1.ConditionFalse,
+			Reason: "Stopped",
```
pierDipi: The `Stopped` Reason is used in different places, can we use a common constant?

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -215,10 +221,21 @@ func (r *DeploymentReconciler) checkDeploymentExist(ctx context.Context, client
 	}, existingDeployment)
 	if err != nil {
 		if apierr.IsNotFound(err) {
-			return constants.CheckResultCreate, nil, nil
+			if !forceStopRuntime {
+				return constants.CheckResultCreate, nil, nil
+			} else {
+				// Do nothing
+				return constants.CheckResultUnknown, nil, nil
+			}
```
pierDipi:
```suggestion
			if !forceStopRuntime {
				return constants.CheckResultCreate, nil, nil
			}
			return constants.CheckResultUnknown, nil, nil
```

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -428,6 +445,12 @@ func (r *DeploymentReconciler) Reconcile(ctx context.Context) ([]*appsv1.Deploym
 
 			// Patch the deployment object with the strategic merge patch
 			opErr = r.client.Patch(ctx, existingDep, kclient.RawPatch(types.StrategicMergePatchType, patchByte))
+
+		case constants.CheckResultDelete:
+			log.Info("Stopping deployment", "namespace", existingDep.Namespace, "name", existingDep.Name)
```
pierDipi: Should the message be `Deleting ...` ?

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -227,12 +227,32 @@ func shouldDeleteHPA(existing *autoscalingv2.HorizontalPodAutoscaler, desired *a
...
 func shouldCreateHPA(desired *autoscalingv2.HorizontalPodAutoscaler) bool {
+	// Skip creating the HPA if the stop annotation is true
+	forceStopRuntime := false
+	if val, exist := desired.Annotations[constants.StopAnnotationKey]; exist {
+		forceStopRuntime = strings.EqualFold(val, "true")
+	}
+
+	if forceStopRuntime {
+		return false
+	}
```
pierDipi: this logic is repeated in a few places [1] [2] [3], should we create a method/function that encapsulate it and we can just call it ? Even if the object is different we can use the interface `metav1.Object` that every Kind should implement `func ShouldForceStop(obj metav1.Object) bool`

[1] pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
[2] pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
[3] pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
pierDipi: This comment is generally applicable to other places (like `pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go`)

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -704,45 +725,65 @@ func (r *RawHTTPRouteReconciler) reconcileTopLevelHTTPRoute(ctx context.Context,
...
+					log.Error(err, "Failed to update toplevel HttpRoute", "name", isvc.Name)
+					return err
+				}
+			}
+		}
+	} else {
+		// Delete the http route
+		if getExistingErr == nil {
+			// Make sure that we only delete http route owned by the isvc
+			if existingHttpRoute.OwnerReferences[0].UID == isvc.UID && existingHttpRoute.GetDeletionTimestamp() == nil {
```
pierDipi: is it guaranteed in every case (even if some other components adds owner refs) that the first owner reference is "ours"? if it is, to make it more robust to future changes, should we loop through each owner ref?
don't we need to check the length before trying to access the first element?

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -289,6 +304,11 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context) ([]*corev1.Service, e
 			opErr = r.client.Create(ctx, svc)
 		case constants.CheckResultUpdate:
 			opErr = r.client.Update(ctx, svc)
+		case constants.CheckResultDelete:
+			log.Info("Stopping service", "namespace", svc.Namespace, "name", svc.Name)
```
pierDipi: Same here, should it be `Deleting ...` ?

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -145,19 +148,39 @@ func semanticOtelCollectorEquals(desired, existing *otelv1beta1.OpenTelemetryCol
...
+					log.Error(err, "Failed to create OTel Collector resource", "name", desired.Name)
+					return err
+				}
 			}
 		} else {
 			return err
 		}
 	} else {
+		if forceStopRuntime {
+			log.Info("Stopping OpenTelemetry Collector", "namespace", existing.Namespace, "name", existing.Name)
```
pierDipi: Same here, should it be `Deleting ...` ?

pierDipi (commented):
I would've structured the feature reconciliation a little bit differently but the Serverless mode was done this way, so I'd stay consistent, but was it considered to have the "stop" be a separate reconciliation path since it's mostly deleting child resources so that not every sub-reconciler needs to be aware of the stop feature (which would also reduce the level of branching each has)?

pierDipi:
In CI job [E2E Tests / test-path-based-routing (pull_request)](https://github.com/kserve/kserve/actions/runs/15122703557/job/42509483756?pr=4455) I see 
```
Loading image kserve-localmodel-controller-ac5980e00a94684e3d201e35e739a4b390a456cb
write /manager: no space left on device
```
in [Go / Build (pull_request)](https://github.com/kserve/kserve/actions/runs/15122703527/job/42508508399?pr=4455)
```
E0519 20:42:24.824750   34583 suite_test.go:52] unable to start control plane itself: failed to start the controlplane. retried 5 times: exec: "etcd": executable file not found in $PATHFailed to start testing panel
```

they look like unrelated errors

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: @israel-hdez Can you weigh in? I think I asked a similar question about needing to loop through the references when you suggested adding checking for the ownership before deleting, but I don't remember the logic that allowed `OwnerReferences[0]` to be okay

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: my only idea is to make new functions like createHttpRoute, updateHttpRoute, and deleteHttpRoute... which will accomplish a bit of better legibility 

I'm not seeing a way to reduce the amount of nesting. This is mostly the original logic, just bumped inwards one level since I wrapped the logic in a "if stopped {}" check basically

hdefazio: but I think creating the functions I mentioned will only eliminate one layer so I don't think it will improve it a ton

pierDipi: I've not checked if it's fully correct but something like this should give the idea:

- expedite force stop as soon as possible
- expedite desired == nil as soon as possible
- return early

```golang
func (r *RawHTTPRouteReconciler) reconcileTopLevelHTTPRoute(ctx context.Context, isvc *v1beta1.InferenceService) error {
	desired, err := createRawTopLevelHTTPRoute(ctx, isvc, r.ingressConfig, r.isvcConfig, r.client)
	if err != nil || desired == nil {
		return err
	}

	if err := controllerutil.SetControllerReference(isvc, desired, r.scheme); err != nil {
		log.Error(err, "Failed to set controller reference for top level HttpRoute", "name", isvc.Name)
		return err
	}

	existingHttpRoute := &gatewayapiv1.HTTPRoute{}
	getExistingErr := r.client.Get(ctx, types.NamespacedName{Namespace: isvc.Namespace, Name: isvc.Name}, existingHttpRoute)
	if getExistingErr != nil && !apierr.IsNotFound(getExistingErr) {
		return fmt.Errorf("failed to get existing http route: %w", getExistingErr)
	}
	if apierr.IsNotFound(getExistingErr) && isvc.GetForceStopRuntime() {
		return nil
	}
	if isvc.GetForceStopRuntime() {
		if ctrl := metav1.GetControllerOf(existingHttpRoute); ctrl != nil && ctrl.UID == isvc.UID {
			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated top level http route")
			if err := r.client.Delete(ctx, existingHttpRoute); err != nil {
				return fmt.Errorf("failed to delete HTTPRoute %s/%s: %w", desired.Namespace, desired.Name, err)
			}
		}
		return nil
	}

	if apierr.IsNotFound(getExistingErr) {
		log.Info("Creating top level HttpRoute resource", "name", isvc.Name)
		if err := r.client.Create(ctx, desired); err != nil {
			log.Error(err, "Failed to create top level HttpRoute", "name", desired.Name)
			return err
		}
	}

	// Set ResourceVersion which is required for update operation.
	desired.ResourceVersion = existingHttpRoute.ResourceVersion
	// Do a dry-run update to avoid diffs generated by default values.
	// This will populate our local httpRoute with any default values that are present on the remote version.
	if err := r.client.Update(ctx, desired, client.DryRunAll); err != nil {
		log.Error(err, "Failed to perform dry-run update on top level httpRoute", "name", desired.Name)
		return err
	}
	if !semanticHttpRouteEquals(desired, existingHttpRoute) {
		if err = r.client.Update(ctx, desired); err != nil {
			log.Error(err, "Failed to update toplevel HttpRoute", "name", isvc.Name)
			return err
		}
	}
	return nil
}
```

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
hdefazio: Updated, pretty sure I fixed all the places this was used including some from the previous pr

pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go
```diff
@@ -215,10 +217,19 @@ func (r *DeploymentReconciler) checkDeploymentExist(ctx context.Context, client
 	}, existingDeployment)
 	if err != nil {
 		if apierr.IsNotFound(err) {
-			return constants.CheckResultCreate, nil, nil
+			if !forceStopRuntime {
+				return constants.CheckResultCreate, nil, nil
+			}
+			return constants.CheckResultUnknown, nil, nil
```
israel-hdez: This was confusing to me. It make more sense to use `constants.CheckResultSkipped`.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -532,38 +532,60 @@ func (r *RawHTTPRouteReconciler) reconcilePredictorHTTPRoute(ctx context.Context
...
+	}
+
+	// ISVC is stopped, delete the httproute if it exists, otherwise, do nothing
+	forceStopRuntime := utils.GetForceStopRuntime(isvc)
+	if (getExistingErr != nil && httpRouteIsNotFound) && forceStopRuntime {
+		return nil
+	}
+	if forceStopRuntime {
+		if ctrl := metav1.GetControllerOf(existingHttpRoute); ctrl != nil && ctrl.UID == isvc.UID {
+			log.Info("The InferenceService ", httpRouteName, " is marked as stopped — delete its associated http route")
```
israel-hdez: The log looks malformed.

```suggestion
			log.Info("The InferenceService is marked as stopped — deleting its associated http route", "name", httpRouteName)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -655,43 +677,58 @@ func (r *RawHTTPRouteReconciler) reconcileTopLevelHTTPRoute(ctx context.Context,
...
+	}
+
+	// ISVC is stopped, delete the httproute if it exists, otherwise, do nothing
+	forceStopRuntime := utils.GetForceStopRuntime(isvc)
+	if (getExistingErr != nil && apierr.IsNotFound(getExistingErr)) && forceStopRuntime {
+		return nil
+	}
+	if forceStopRuntime {
+		if ctrl := metav1.GetControllerOf(existingHttpRoute); ctrl != nil && ctrl.UID == isvc.UID {
+			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated top level http route")
```
israel-hdez: Also here
```suggestion
			log.Info("The InferenceService is marked as stopped — deleting its associated top level http route", "name", isvc.Name)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
```diff
@@ -68,37 +68,64 @@ func (r *RawIngressReconciler) Reconcile(ctx context.Context, isvc *v1beta1.Infe
...
+	}
+
+	// ISVC is stopped, delete the ingress if it exists, otherwise, do nothing
+	forceStopRuntime := utils.GetForceStopRuntime(isvc)
+	if (getExistingErr != nil && ingressIsNotFound) && forceStopRuntime {
+		return nil
+	}
+	if forceStopRuntime {
+		if ctrl := metav1.GetControllerOf(existingIngress); ctrl != nil && ctrl.UID == isvc.UID {
+			log.Info("The InferenceService ", isvc.Name, " is marked as stopped — delete its associated ingress")
```
israel-hdez: Also here...

```suggestion
			log.Info("The InferenceService is marked as stopped — deleting its associated ingress", "name", isvc.Name)
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -532,38 +532,60 @@ func (r *RawHTTPRouteReconciler) reconcilePredictorHTTPRoute(ctx context.Context
...
+			log.Error(err, "Failed to create predictor HttpRoute", "name", desired.Name)
 			return err
 		}
-	} else {
-		// Set ResourceVersion which is required for update operation.
-		desired.ResourceVersion = existing.ResourceVersion
-		// Do a dry-run update to avoid diffs generated by default values.
-		// This will populate our local httpRoute with any default values that are present on the remote version.
-		if err := r.client.Update(ctx, desired, client.DryRunAll); err != nil {
-			log.Error(err, "Failed to perform dry-run update for predictor HttpRoute", "name", desired.Name)
```
israel-hdez, line 559: The removal of the dry-run, wouldn't it have any side effects? I think the cluster sets some defaults, so the controller may keep updating it constantly.

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -145,32 +147,49 @@ func semanticOtelCollectorEquals(desired, existing *otelv1beta1.OpenTelemetryCol
...
-		} else {
-			return err
 		}
-	} else {
-		// Set ResourceVersion which is required for update operation.
-		desired.ResourceVersion = existing.ResourceVersion
-		// Do a dry-run update to avoid diffs generated by default values.
-		// This will populate our local httpRoute with any default values that are present on the remote version.
-		if err := o.client.Update(ctx, desired, client.DryRunAll); err != nil {
-			log.Error(err, "Failed to perform dry-run update for OTel Collector", "name", desired.Name)
```
israel-hdez, line 166: Similar question here, about if the missing dry-run would lead to constantly updating the resource?

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -256,11 +258,19 @@ func (r *ServiceReconciler) checkServiceExist(ctx context.Context, client client
 	}, existingService)
 	if err != nil {
 		if apierr.IsNotFound(err) {
-			return constants.CheckResultCreate, nil, nil
+			if !forceStopRuntime {
+				return constants.CheckResultCreate, nil, nil
+			}
+			return constants.CheckResultUnknown, nil, nil
```
israel-hdez: Also here, it may make more sense to use `constants.CheckResultSkipped`

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler.go
```diff
@@ -289,6 +299,11 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context) ([]*corev1.Service, e
 			opErr = r.client.Create(ctx, svc)
 		case constants.CheckResultUpdate:
 			opErr = r.client.Update(ctx, svc)
+		case constants.CheckResultDelete:
+			log.Info("Deleting service", "namespace", svc.Namespace, "name", svc.Name)
+			if svc.GetDeletionTimestamp() == nil { // check if the service was already deleted
```
israel-hdez, line 307: No check for owner reference... Similarly, in the deployment_reconciler.
I can suggest to add the check in both places.

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -6101,6 +7169,11 @@ var _ = Describe("v1beta1 inference service controller", func() {
 							Status:   "True",
 							Severity: "Info",
 						},
+						{
```
israel-hdez: Units results suggests to change order in the array.

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -3585,6 +4638,11 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
```
israel-hdez: Also here, units results suggests to change order in the array.

israel-hdez:
The very large error has its root in a `panic` reported here: https://github.com/kserve/kserve/actions/runs/15434418706/job/43437934692?pr=4455#step:5:666 :

> panic: test timed out after 10m0s

This is `go test` timing out. I disabled the timeout and I get a much more readable report:

```
Summarizing 3 Failures:
  [FAIL] v1beta1 inference service controller When creating inference service with raw kube predictor and `serving.kserve.io/stop` [It] Should delete the httproute/service/deployment/hpa when the annotation is updated to true on an existing ISVC
  /home/ehernand/Projects/kserve/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go:2593
  [FAIL] v1beta1 inference service controller When creating inference service with raw kube predictor and transformer [It] Should have httproute/service/deployment/hpa created for transformer and predictor
  /home/ehernand/Projects/kserve/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go:4686
  [FAIL] v1beta1 inference service controller When creating inference service with raw kube path based routing predictor and transformer [It] Should have ingress/service/deployment/hpa created for transformer and predictor
  /home/ehernand/Projects/kserve/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go:7217

Ran 62 of 62 Specs in 1212.018 seconds
FAIL! -- 59 Passed | 3 Failed | 0 Pending | 0 Skipped
--- FAIL: TestV1beta1APIs (1212.02s)
```

Notice the last line `TestV1beta1APIs (1212.02s)` reports a full run is taking about 20minutes in my machine, while `go test` default timeout is 10minutes. [From docs](https://pkg.go.dev/cmd/go/internal/test):

```
	-timeout d
	    If a test binary runs longer than duration d, panic.
	    If d is 0, the timeout is disabled.
	    The default is 10 minutes (10m).
```

I don't think we should disable the timeout. IMO we should simply increase it to allow for the units to run properly. You can do it in the [Makefile](https://github.com/kserve/kserve/blob/5cb67351b7d0c88c48f291603046198d736274d6/Makefile#L164). I can suggest a 30min timeout given the results on my machine.

About the failing tests, the existing ones:
* rawkube_controller_test.go:4686
* rawkube_controller_test.go:7217

apparently are just failing because the check is out of order, and both give a diff like the following:

```
                              {
      -                                 Type:     "TransformerReady",
      +                                 Type:     "Stopped",
      -                                 Status:   "True",
      +                                 Status:   "False",
                                        Severity: "Info",
                                        ... // 1 ignored and 2 identical fields
                                },
                                {
      -                                 Type:     "Stopped",
      +                                 Type:     "TransformerReady",
      -                                 Status:   "False",
      +                                 Status:   "True",
                                        Severity: "Info",
                                        ... // 1 ignored and 2 identical fields
                                },

```

suggesting you just need to swap the array you updated here: https://github.com/kserve/kserve/pull/4455/files#r2127716685 and here: https://github.com/kserve/kserve/pull/4455/files#r2127718390. I have to say that the different order looks suspicious, although in practice order doesn't really matter.

Finally, the unit failing in rawkube_controller_test.go:2593, suggests that something is buggy. The error reported is that the Route is missing:

```
[FAILED] Timed out after 60.000s.
  Expected success, but got an error:
      <*errors.StatusError | 0xc00183ad20>: 
      HTTPRoute.gateway.networking.k8s.io "stop-edit-true-isvc" not found
```

I don't know if it the unit is buggy, or if the code is buggy. At the moment, I know that, apparently, the HTTPRoute reconciler is not running.

pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
```diff
@@ -1680,203 +1713,1213 @@ var _ = Describe("v1beta1 inference service controller", func() {
...
+			}`,
+		}
+
+		createInferenceServiceConfigMap := func() *corev1.ConfigMap {
+			configMap := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      constants.InferenceServiceConfigMapName,
+					Namespace: constants.KServeNamespace,
+				},
+				Data: configs,
```
israel-hdez: Fix units by importing `maps`, and then change this line with:

```suggestion
				Data: maps.Clone(configs),
```

israel-hdez:
The problematic unit rawkube_controller_test.go:2593 is failing because the Context is buggy.

Inside its containing `Context("When creating inference service with raw kube predictor and ``serving.kserve.io/stop``")` you have a [configs variable](https://github.com/kserve/kserve/blob/e5ea29017f37ebb6ef6ed09f8a16331a5657591e/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go#L1911-L1937):

```go
configs := map[string]string{
			"explainers": `[...snip...]`,
			"ingress": `[...snip...]`,
			"storageInitializer": `[...snip...]`,
		}
```

This variable is assigned [inside createInferenceServiceConfigMap function](https://github.com/kserve/kserve/blob/e5ea29017f37ebb6ef6ed09f8a16331a5657591e/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go#L1945), but the assignment is ordinary. Since maps are reference types, no copy is done.

Thus, units requiring to turn off Gateway API, like `It("Should keep the ingress/service/deployment/keda/otel created when gateway api is disabled and the annotation is set to false")` [here](https://github.com/kserve/kserve/blob/e5ea29017f37ebb6ef6ed09f8a16331a5657591e/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go#L2182-L2192) are mutating the global `configs` and the mutated config is used in following units.

To fix, see comment: https://github.com/kserve/kserve/pull/4455/files#r2130518090

It turns out that @pierDipi was right that Gateway API was configured as disabled.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: I changed semanticHttpRouteEquals to use DeepDerivative instead of  DeepEqual, which should accomplish determining the diff without needing a dry run. I can change it back if you're worried about it though

hdefazio: @pierDipi Suggested it to help clean up the functions

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: Resolved by changing to `if ctrl := metav1.GetControllerOf(existingHttpRoute); ctrl != nil && ctrl.UID == isvc.UID {` instead

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
israel-hdez: OK sounds good. I did see the change to DeepDerivative. I didn't relate this removal to that change.

israel-hdez (approved):
Great work @hdefazio

sivanantha321:
@hdefazio Documenting this feature would be nice !

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -509,7 +509,7 @@ func createRawTopLevelHTTPRoute(isvc *v1beta1.InferenceService, ingressConfig *v
 }
 
 func semanticHttpRouteEquals(desired, existing *gatewayapiv1.HTTPRoute) bool {
-	return equality.Semantic.DeepEqual(desired.Spec, existing.Spec)
+	return equality.Semantic.DeepDerivative(desired.Spec, existing.Spec)
```
sivanantha321, line 512: From the docs
> DeepDerivative is similar to DeepEqual except that unset fields in a1 are ignored (not compared). This allows us to focus on the fields that matter to the semantic comparison.

> The unset fields include a nil pointer and an empty string.

@hdefazio In the [backendobjectreference](https://github.com/kserve/kserve/blob/6530d1e0bc0427cfb198d1d020042481f9b025d5/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go#L133), the Group field is set to empty string (which is default). If it is accidently updated, does the controller pick up the change and update it to the desired spec i.e. set the Group field to an empty string ?

hdefazio:
@sivanantha321 Got it, I'll open a task on my side to write some documentation. I'll add it to the next sprint we have in ~2 weeks

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: Sorry for the late response, it will not set the Group field to an empty string (it should leave the field as-is) and we do not foresee any issues with this, as all tests are passing

sivanantha321: The test cases are not covering this edge case. The group field is not explicitly set because of the default value. Though it is a very rare scenario, IMO the controller should try to match the current state to the desired state as per best practices.

cc @pierDipi

PR #4471, opened 2025-05-14
Rework the order in which the knative autoscaler configmap is read during reconciliation

### LINKED ISSUE #4486: Configuration of by isvc causing breaking issue with rollout
/kind bug

We cannot default the value of init to the min scale. This will break the rollout and especially progressive rollout with knative.
https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/utils/utils.go#L89

initial-scale in knative serving is just an initial number of pods for deployment to scale up to, because the PA is able to take over the controller of the number of pods to autoscale.

Usually it is set to 1. As long as one pod is able to launch, knative believe this will be a a successful launch. PA then takes over to spin up with the rest number of the pods.

If this value is set to the minscale number, knative will directly set the number of replicas into minscale, which will break the progressive rollout, because progressive rollout definitely need to spin up with smaller substage.

We should keep `initial-scale` to 1.

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

brettmthompson:
Fixes #4486 
Addresses [comment](https://github.com/kserve/kserve/pull/4394#issuecomment-2869901540) in [PR 4394](https://github.com/kserve/kserve/pull/4394)

The changes in this PR encompass the following:

1. Moving the call to `CheckZeroInitialScaleAllowed` into the reconcile method of the `InferenceServiceReconciler` and `InferenceGraphReconciler` structs. This decouples the configmap retrieval logic from the ksvc reconciliation and eliminates the need to pass the k8s clientSet all the way down (now passing boolean `allowZeroInitialScale` value instead).
2. With the above change, there is no longer a need to return an error from the `createKnativeService` methods. All unnecessary error return values and handling logic are removed.
3. Reverting changes made to logic for setting min scale and max scale annotations in the `SetAutoscalingAnnotations` method.
4. There is support for overriding the default initial scale behavior by setting the `autoscaling.knative.dev/initial-scale` annotation in an ISVC or IG. Moving all logic that validates this set annotation to a new method `ValidateInitialScaleAnnotation`.

**Type of changes**

- [x] Unit Tests
- [ ] E2E Tests
- [x] Manual Integration Tests

**Test Strategy:**

Inference Service:
1. Install Knative Serving and Kserve in a cluster
2. Deploy sample ISVC with min replicas value of 0 and max replicas value of 1.
3. Observe that the default behavior is executed, one pod is created to initialize the ISVC, then scaled down to 0.
4. Add the `autoscaling.knative.dev/initial-scale`  annotation to the ISVC with a value of 0.
5. Observe that the annotation is ignored and default behavior is executed, ISVC is initialized with 1 pod and then scaled down to 0.
6. Update the KnativeServing operator and apply the following value under the config key. (Alternatively, if KnativeServing CR is not installed in the cluster, this value can be updated in the config-autoscaler configmap directly):
```    
autoscaler:
      allow-zero-initial-scale: "true"
````
7. Edit the kserve-controller-manager-deployment and update the image to be quay.io/bretthom/kserve-dev:autoscaling
8. Add the `autoscaling.knative.dev/initial-scale`  annotation to the ISVC with a value of 0.
9. Observe that no new pods are created for the new knative revision and the old pod is terminated.
10. Add the `autoscaling.knative.dev/inital-scale` annotation to the ISVC with a value of 2.
11. Observe that 2 new pods are created to initialize the new knative revision, then get scaled down to 0.
12. Add the `autoscaling.knative.dev/inital-scale` annotation to the ISVC with a non-integer value.
13. Observe that the annotation is ignored and default behavior is executed, ISVC is initialized with 1 pod and then scaled down to 0.

Inference Graph:
1. Install Knative Serving and Kserve in a cluster
2. Deploy sample IG with min replicas value of 0 and max replicas value of 1.
3. Observe that the default behavior is executed, one pod is created to initialize the IG, then scaled down to 0.
4. Add the `autoscaling.knative.dev/initial-scale`  annotation to the IG with a value of 0.
5. Observe that the annotation is ignored and default behavior is executed, IG is initialized with 1 pod and then scaled down to 0.
6. Update the KnativeServing operator and apply the following value under the config key. (Alternatively, if KnativeServing CR is not installed in the cluster, this value can be updated in the config-autoscaler configmap directly):
```    
autoscaler:
      allow-zero-initial-scale: "true"
````
7. Edit the kserve-controller-manager-deployment and update the image to be quay.io/bretthom/kserve-dev:autoscaling
8. Add the `autoscaling.knative.dev/initial-scale`  annotation to the IG with a value of 0.
9. Observe that no new pods are created for the new knative revision and the old pod is terminated.
10. Add the `autoscaling.knative.dev/inital-scale` annotation to the IG with a value of 2.
11. Observe that 2 new pods are created to initialize the new knative revision, then get scaled down to 0
12. Add the `autoscaling.knative.dev/inital-scale` annotation to the IG with a non-integer value
13. Observe that the annotation is ignored and default behavior is executed, IG is initialized with 1 pod and then scaled down to 0.

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


brettmthompson:
@yuzisun when you find time, could you please review to confirm if these changes align with what you outlined [here](https://github.com/kserve/kserve/pull/4394#issuecomment-2869901540).

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -137,28 +136,30 @@ func semanticEquals(desiredService, service *knservingv1.Service) bool {
...
-func createKnativeService(ctx context.Context, clientset kubernetes.Interface, componentMeta metav1.ObjectMeta, graph *v1alpha1.InferenceGraph, config *RouterConfig) (*knservingv1.Service, error) {
+func createKnativeService(
+	componentMeta metav1.ObjectMeta,
+	graph *v1alpha1.InferenceGraph,
+	config *RouterConfig,
+	allowZeroInitialScale bool,
+) *knservingv1.Service {
 	bytes, err := json.Marshal(graph.Spec)
 	if err != nil {
-		return nil, errors.Wrapf(err, "fails to marshal inference graph spec to json")
```
spolti, line 143: why removing the error returns from this func?

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler_test.go
```diff
@@ -363,25 +352,17 @@ func TestKsvcReconciler_Reconcile(t *testing.T) {
 				err := client.Create(t.Context(), tt.existingKsvc)
 				require.NoError(t, err)
 			}
-			clientset := fake.NewSimpleClientset(&corev1.ConfigMap{
```
spolti, line 366: @sivanantha321 isn't the clientset the preferred client to use?

spolti (commented):
minor comments, otherwise lgtm

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
brettmthompson: All error handling logic I am removing here was added by me in this [PR](https://github.com/kserve/kserve/pull/4394). Originally the methods used to create knative services and knative service reconcilers did not have an error return parameter. Those error returns were necessary if we read the knative serving config map within those methods, but since this PR moves that check to earlier in the reconciliation loop I am restoring these methods to a state where they no longer return an error.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler_test.go
brettmthompson: I believe it is preferred for certain scenario, which is why I was using it here @spolti 

This clientset was added by me in this [PR](https://github.com/kserve/kserve/pull/4394) for the purpose of using it to read the knative serving config map. Since we are no longer reading this configmap within the knative service related methods, the clientset is no longer needed as an input parameter. Prior to the changes in that PR, only the k8s client was passed to these methods.

spolti (approved):
/lgtm

pkg/controller/v1alpha1/utils/utils.go
```diff
@@ -93,28 +97,30 @@ func SetAutoScalingAnnotations(ctx context.Context,
...
-			autoscaling.InitialScaleAnnotationKey,
-			annotations[autoscaling.InitialScaleAnnotationKey],
-		)
-	}
-
-	// Retrieve the allow-zero-initial-scale value from the knative autoscaler configuration.
-	allowZeroInitialScale, err := CheckZeroInitialScaleAllowed(ctx, clientset)
-	if err != nil {
-		return errors.Wrapf(err, "failed to retrieve the knative autoscaler configuration")
+		// If the annotation is explicitly set, validate that it is set to an integer value.
```
yuzisun, line 100: @brettmthompson See https://github.com/kserve/kserve/issues/4486, on line 98 can we not default to MinScale as that's the changing the default Knative behavior.

pkg/controller/v1alpha1/utils/utils.go
```diff
@@ -70,8 +72,10 @@ func SetAutoScalingAnnotations(ctx context.Context,
 	// If a min replicas value is not set, use the default min replicas value.
 	if minReplicas == nil {
 		annotations[autoscaling.MinScaleAnnotationKey] = strconv.Itoa(int(constants.DefaultMinReplicas))
+		initialScale = constants.DefaultMinReplicas
 	} else {
 		annotations[autoscaling.MinScaleAnnotationKey] = strconv.Itoa(int(*minReplicas))
+		initialScale = *minReplicas
```
yuzisun: @brettmthompson See https://github.com/kserve/kserve/issues/4486, can we not default to MinScale as that's changing the default Knative behavior.

yuzisun (approved):

PR #4473, opened 2025-05-14
4380 - Inference logging to blob storage

### LINKED ISSUE #4380: Add support for writing model request/response logs to object stores

cjohannsen-cloudera:

Storage is defined using the StorageSpec.  The existing `v1beta1.StorageSpec` has been refactored to extract a base struct.  
```
// StorageSpec defines a spec for an object in an object store
type StorageSpec struct {
	// The path to the object in the storage. Note that this path is relative to the storage URI.
	// +optional
	Path *string `json:"path,omitempty"`
	// Parameters to override the default storage credentials and config.
	// +optional
	Parameters *map[string]string `json:"parameters,omitempty"`
	// The Storage Key in the secret for this object.
	// +optional
	StorageKey *string `json:"key,omitempty"`
}
```

The existing `StorageSpec` is now `ModelStorageSpec` and adds the `SchemaPath`
```
type ModelStorageSpec struct {
	StorageSpec `json:",inline"`
	// The path to the model schema file in the storage.
	// +optional
	SchemaPath *string `json:"schemaPath,omitempty"`
}
```

This is a pure refactor.

The LoggerConfig now contains a StorageSpec that defines the logger storage output location and a service account name used to fetch credentials.

```
  logger: |-
    {
        "image" : "kserve/agent:latest",
        "memoryRequest": "100Mi",
        "memoryLimit": "1Gi",
        "cpuRequest": "100m",
        "cpuLimit": "1",
        "defaultUrl": "http://default-broker",
        "storage": {
            "path": "",
            "parameters": {
              "type": "s3",
              "format": "json" 
            },
            "key": ""
            "serviceAccountName": "",
        }
    }
```

Three new command line parameters have been added to the agent to allow the storage configuration to be passed in when injecting the agent.
```
--log-store-path [path] - path within the storage location to write the logger output
--log-store-format - output format of the stored data, one of "json" or "yaml", default "json"
```

By default the logger service account is undefined, and the logger storage feature is disabled.

If the service account name is defined, the logger will attempt to look for a Service Account with the matching name and mount the secrets it contains.  When the agent is launched the secrets are available in the environment.

When an inference occurs the S3 client is constructed and the request and response are stored to the blob store path specified above.  The agent uses the existing client defined for the model downloader to simplify integration and configuration, enhanced to allow for uploads.

Which issue(s) this PR fixes

Fixes https://github.com/kserve/kserve/issues/4380

**Type of changes**

[x] Added additional test to logger/handler_test.go
[x] Added new test, logger/store_test.go
[x] Added new agent command argument to agent_injector_test.go

To test locally create a secret in the  `default` namespace
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    serving.kserve.io/s3-region: [YOUR_REGION]
  name: agent-logger-secret
  namespace: default
type: Opaque
data:
  AWS_ACCESS_KEY_ID: [YOUR_ACCESS_KEY_ID]
  AWS_DEFAULT_REGION: [YOUR_REGION]
  AWS_SECRET_ACCESS_KEY: [YOUR_SECRET_ACCESS_KEY]
```

Create a service account for the secret:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: logger-sa
  namespace: default
secrets:
- name: agent-logger-secret
```

To test the feature, deploy a model with inference logging: 
```
kind: InferenceService
metadata:
  name: sklearn-iris
spec:
  predictor:
    logger:
      mode: all
      url: s3://[YOUR_BUCKET]/
      storage:
        path: "/logger"
        parameters:
          type: s3
          format: json
        key: credentials
    model:
      modelFormat:
        name: sklearn
```

Send a prediction to the endpoint and the request and response are stored as json in the blob storage location `[YOUR_BUCKET]/logger/`

Documentation updated in [PR 481](https://github.com/kserve/website/pull/481)

**Release note**:

```release-note

```

**Re-running failed tests**


cmd/agent/main.go
```diff
@@ -300,8 +304,20 @@ func startLogger(workers int, logger *zap.SugaredLogger) *loggerArgs {
 		}
 	}
 
-	logger.Info("Starting the log dispatcher")
-	kfslogger.StartDispatcher(workers, logger)
+	var store kfslogger.Store
+	if kfslogger.GetStorageStrategy(*logUrl) != kfslogger.HttpStorage {
+		if logStoreFormat != nil && *logStoreFormat != "" {
```
sivanantha321: should we check for nil reference for logStorePath as well ?

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/onsi/gomega v1.36.2
 	github.com/open-telemetry/opentelemetry-operator v0.113.0
 	github.com/pkg/errors v0.9.1
+	github.com/scaleway/scaleway-sdk-go v1.0.0-beta.30
```
sivanantha321: Why do we need this dependency ?

pkg/logger/mock.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2023 The KServe Authors.
```
sivanantha321:
```suggestion
Copyright 2025 The KServe Authors.
```

pkg/logger/store_test.go
```diff
@@ -0,0 +1,78 @@
+/*
+Copyright 2023 The KServe Authors.
```
sivanantha321:
```suggestion
Copyright 2025 The KServe Authors.
```

sivanantha321:
@cjohannsen-cloudera Can you help update the documentation as well ?https://github.com/kserve/website/blob/main/docs/modelserving/logger/logger.md

go.mod
cjohannsen-cloudera: I have no idea!  I'll remove it, I'm not using it AFAIK.

cjohannsen-cloudera:
> @cjohannsen-cloudera Can you help update the documentation as well ?https://github.com/kserve/website/blob/main/docs/modelserving/logger/logger.md

@sivanantha321 yes I can do that

go.mod
cjohannsen-cloudera: @sivanantha321 tracked it down and removed it

cmd/agent/main.go
cjohannsen-cloudera: Fixed

cmd/agent/main.go
```diff
@@ -295,13 +297,25 @@ func startLogger(workers int, logger *zap.SugaredLogger) *loggerArgs {
...
+			log.Errorf("annotation does not adhere to desired format got key: %s value: %s", k, v)
 			os.Exit(-1)
 		}
 	}
 
-	logger.Info("Starting the log dispatcher")
-	kfslogger.StartDispatcher(workers, logger)
+	var store kfslogger.Store
+	if kfslogger.GetStorageStrategy(*logUrl) != kfslogger.HttpStorage {
+		if logStoreFormat != nil && *logStoreFormat != "" && logStorePath != nil && *logStorePath != "" {
```
sivanantha321, line 307: Maybe we can have a default for logFromat ?

pkg/credentials/service_account_credentials.go
```diff
@@ -97,10 +97,19 @@ func (c *CredentialBuilder) CreateStorageSpecSecretEnvs(namespace string, annota
 	if c.config.StorageSpecSecretName != "" {
 		storageSecretName = c.config.StorageSpecSecretName
 	}
-	// secret annotation takes precedence
-	if annotations != nil {
-		if secretName, ok := annotations[c.config.StorageSecretNameAnnotation]; ok {
-			storageSecretName = secretName
+
+	// if the secret name is provided use it
+	if secretName != nil {
```
sivanantha321, line 102: Can we have a test case for this ?

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -46,6 +48,7 @@ const (
 	LoggerArgumentTlsSkipVerify       = "--logger-tls-skip-verify"
 	LoggerArgumentMetadataHeaders     = "--metadata-headers"
 	LoggerArgumentMetadataAnnotations = "--metadata-annotations"
+	LoggerDefaultServiceAccountName   = "logger-sa"
```
sivanantha321, line 51: Is this default name configurable ?

pkg/webhook/admission/pod/agent_injector.go
```diff
@@ -56,16 +59,22 @@ type AgentConfig struct {
 	MemoryLimit   string `json:"memoryLimit"`
 }
 
+type LoggerStorageSpec struct {
```
sivanantha321: This type is already defined in the pkg/apis/serving/v1beta1/inference_service.go. We can reuse that.

cjohannsen-cloudera: Fixed

cmd/agent/main.go
cjohannsen-cloudera: defaulted to `json`

pkg/credentials/service_account_credentials.go
cjohannsen-cloudera: Done

pkg/webhook/admission/pod/agent_injector.go
cjohannsen-cloudera: No, this is the default value chosen if the logger config spec doesn't specify a service account.  This value matches the examples.

cjohannsen-cloudera: I could make this configurable if you'd like but I need a reference as to how.

cjohannsen-cloudera: FYI this can be specified when creating a iscv as it is part of the LoggerConfig

lizzzcai:
from the example, seems like it is using the SA to pass the `agent-logger-secret`. in this case, the storageUri and the logger will be using the same bucket? is it possible to use different buckets for them? like bucket-a for model artifact and bucket-b for logging.

cjohannsen-cloudera:
> from the example, seems like it is using the SA to pass the `agent-logger-secret`. in this case, the storageUri and the logger will be using the same bucket? is it possible to use different buckets for them? like bucket-a for model artifact and bucket-b for logging.

The logger URL in the logger config contains the full path to the bucket.  This can be fully independent of the model storage bucket.

sivanantha321 (dismissed):

sivanantha321:
@cjohannsen-cloudera Can you help resolve the conflicts ? Everything else looks good.

sivanantha321:
@cjohannsen-cloudera the precommit check is failing due to the difference in the autogenerated file. Can you help fix that ?

cjohannsen-cloudera:
@sivanantha321 The "Go / Build" target failed https://github.com/kserve/kserve/actions/runs/15907851660/job/44867391042?pr=4473 due to a timeout, can you rerun this target?  The commands for rerunning aren't working for me.

sivanantha321:
Wonderful !
/approve

sivanantha321 (approved):

PR #4492, opened 2025-05-24
feat: support remote storage URI injection for serving runtimes

### LINKED ISSUE #3428: Store original STORAGE_URI value in an environment variable
/kind feature

For custom predictors, when replacing the `STORAGE_URI` environment variable value with the mount path (`/mnt/models`) we should add an environment variable with the original value. Developers of custom predictors may want to access this value for logging/tracing. Something like `REMOTE_STORAGE_URI` works.

Relevant code: https://github.com/kserve/kserve/blob/e21b913fe10b377d72e67fbf92955dbcedd6f465/pkg/webhook/admission/pod/storage_initializer_injector.go#L291

de0725:
1. Injects REMOTE_STORAGE_URI= into the main user container.
2. Respects both storageSpec-based and PVC-based source URIs.
3. Adds corresponding unit tests to ensure correctness and backward compatibility.
[REMOTE_STORAGE_URI.Usage.and.Testing.Guide (1).docx](https://github.com/user-attachments/files/20422403/REMOTE_STORAGE_URI.Usage.and.Testing.Guide.1.docx)

Fixes #3428 

**Type of changes**

Added unit test: TestRemoteStorageUriEnvInjection
Verified behavior under both ConfigMap and Annotation-based configurations
Ran make test, all unit tests passed

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
Inject REMOTE_STORAGE_URI into user container when remote storage is enabled, allowing runtimes to access the original model source URI.
```

**Re-running failed tests**


de0725:
Hi @cmaddalozzo @spolti @yuzisun 
Would you have some time to review this PR ?
Thank you very much for your time and support!

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -868,3 +888,12 @@ func needCaBundleMount(caBundleConfigMapName string, initContainer *corev1.Conta
 	}
 	return result
 }
+
+func hasEnv(container *corev1.Container, key string) bool {
```
spolti: We have a similar function at `pkg/utils/utils.go` GetEnvVarValue.
If that is not suitable for this use case, can you please move this function to the package mentioned above?

pkg/constants/constants.go
```diff
@@ -69,6 +69,12 @@ var (
 	ModelConfigFileName = "models.json"
 )
 
+// Remote Storage URI
+const (
+	EnableRemoteStorageEnvAnnotation = "kserve.io/enable-remote-storage-env"
```
spolti: reading the related issue, adding a flag for it seems not to address the issue.
@cmaddalozzo can you help clarifying it?

pkg/webhook/admission/pod/storage_initializer_injector.go
de0725: Got it. I’ll review and move the function to the suggested package as soon as possible. Thank you for pointing it out!

pkg/constants/constants.go
de0725: Hi @spolti and @cmaddalozzo  ,

Just to double-check that I'm on the same page:  
When we overwrite the `STORAGE_URI` environment variable with `/mnt/models`, are we expected to:

1. **Always** inject a new environment variable (e.g., `REMOTE_STORAGE_URI`) with the original URI as a default behavior?

**or**

2. Make this behavior optional and controlled via an annotation flag like `kserve.io/enable-remote-storage-env`?

My initial understanding was option 2, but based on your comments it sounds like option 1 — making it the default behavior — might be preferred. 
Would love to confirm before proceeding with the update.

Thanks again for your guidance!

pkg/webhook/admission/pod/storage_initializer_injector.go
de0725: Hi @spolti , I’ve removed the custom hasEnv and replaced it with utils.GetEnvVarValue, added the necessary import. Please take another look and let me know if any further tweaks are needed. Thanks!

pkg/constants/constants.go
yuzisun: @de0725 I think we can always inject the `REMOTE_STORAGE_URI` environment variable if storageURI presents on YAML, we do not need a flag to enable this as a new environment variable should not affect anything.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -237,6 +251,13 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(pod *corev1.Pod)
 	transformerContainer := getContainerWithName(pod, constants.TransformerContainerName)
 	workerContainer := getContainerWithName(pod, constants.WorkerContainerName)
 
+	if enabledRemoteEnv {
+		if _, exists := utils.GetEnvVarValue(userContainer.Env, RemoteStorageEnvVarName); exists {
+			return fmt.Errorf("container %s already defines %s", userContainer.Name, RemoteStorageEnvVarName)
```
yuzisun: The webhook can be called multiple times so we need to make it idempotent, we do not need to return an error instead check if the environment does not exist we inject it.

pkg/constants/constants.go
de0725: Got it, I’ll update the logic to always inject REMOTE_STORAGE_URI and submit a PR soon. Thanks!

pkg/webhook/admission/pod/storage_initializer_injector.go
de0725: Good point, I’ll update it to make the injection idempotent as suggested. Thanks!

de0725:
Hi @yuzisun @spolti  , I’ve refactored the logic to always inject REMOTE_STORAGE_URI when the storageURI annotation is present, removed all related flag checks, and updated the unit tests accordingly. Please review and let me know if any further adjustments are required. Thanks!

de0725:
Hi @spolti  @yuzisun ,
Apologies for missing the pre-commit issues earlier. I've fixed the formatting and force-pushed the changes.
Appreciate your time and review. Thanks!

yuzisun (approved):
Looks good! Thanks @de0725 for the contribution !

PR #4518, opened 2025-06-11
fix: moved webhdfs to pyproject.toml file

LaurentGoderre:

Some of the dependencies for storage initializer live in the Dockerfile only. This moves them to the `pyproject.toml`

**Type of changes**


python/kserve/pyproject.toml
```diff
@@ -62,6 +62,9 @@ boto3 = { version = "^1.29.0", optional = true }
 huggingface-hub = { version = ">=0.32.0", extras = [
   "hf-transfer",
 ], optional = true }
+krbcontext = { version = "0.10", optional = true }
```
sivanantha321: Storage now has separate package. This should be moved there. https://github.com/kserve/kserve/blob/master/python/storage/pyproject.toml.

LaurentGoderre: Updated!

python/kserve/pyproject.toml
```diff
@@ -81,6 +81,9 @@ storage = [
   "azure-identity",
   "boto3",
   "huggingface-hub",
+  "krbcontext",
```
sivanantha321: @spolti Why do we need storage dependencies here ?

sivanantha321: This is not a blocker. If deemed unnecessary, it can be remove later.

python/storage/pyproject.toml
```diff
@@ -28,6 +28,9 @@ azure-storage-file-share = { version = "^12.16.0"}
 azure-identity = { version = "^1.15.0"}
 boto3 = { version = "^1.29.0"}
 huggingface-hub = { version = "^0.30.0", extras = ["hf-transfer"]}
+krbcontext = { version = "0.10", optional = true }
+hdfs = { version = "~2.6.0", optional = true }
+requests-kerberos = { version = "0.14.0", optional = true }
```
sivanantha321: @LaurentGoderre If this is marked as optional, how it will be installed ? Is the plan is to create an extra for hdfs ?

LaurentGoderre: Oh good catch, the optional was for when it was in the parent one!

LaurentGoderre:
@sivanantha321  refactored to the switch to uv

python/storage/pyproject.toml
```diff
@@ -24,3 +24,9 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Operating System :: OS Independent",
 ]
+[project.optional-dependencies]
+storage = [
```
sivanantha321: a better name would be `hdfs`

LaurentGoderre: Done!

sivanantha321 (approved):

sivanantha321:
Thanks for the contribution !
/lgtm
/approve

python/storage-initializer.Dockerfile
```diff
@@ -33,10 +33,8 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # Install Kerberos-related packages
-RUN uv pip install --no-cache \
-    krbcontext==0.10 \
-    hdfs~=2.6.0 \
-    requests-kerberos==0.14.0
+COPY storage/pyproject.toml storage/uv.lock storage/
+RUN cd storage && uv sync --extra hdfs --active --no-cache
```
sivanantha321, line 37: looks like this is removing the previously installed packages like kserve etc.

spolti: This PR needs a rebase, @LaurentGoderre, can you please rebase? The storage-init Dockerfile looks outdated; we don't install KServe deps anymore.

IIRC I tried something similar previously, I am failing to remember why I kept it this way (installing directly in the Dockerfile) for now.

LaurentGoderre: Done

Issue #4371, opened 2025-04-08
Support secure access to the metrics server with TriggerAuthentication for KEDA autoscaling


pierDipi:
/kind feature

**Describe the solution you'd like**
Currently, when using KEDA for autoscaling with Prometheus in KServe, there is no way to configure authentication parameters.

This would enable secure access to the metrics server configuring KEDA authentication parameters (using TriggerAuthentication).

Maybe with the following API changes:

```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: huggingface-fbopt
  annotations:
    serving.kserve.io/deploymentMode: "RawDeployment"
    serving.kserve.io/autoscalerClass: "keda"
    serving.kserve.io/enable-prometheus-scraping: "true"
    prometheus.io/scrape: "true"
    prometheus.io/path: "/metrics"
    prometheus.io/port: "8080"
    prometheus.io/scheme: "http"
spec:
  # ...
    autoScaling:
      metrics:
        - type: External
          external:
            metric:
              backend: "prometheus"
              serverAddress: "http://<prometheus-host>:9090"
              query: vllm:num_requests_running
            authenticationRef:
              name: keda-prom-creds
            target:
              type: Value
              value: "2"
---
# Just for reference ...
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: keda-prom-creds
spec:
  secretTargetRef:
    - parameter: username
      name: keda-prom-secret
      key: username
    - parameter: password
      name: keda-prom-secret
      key: password
```

**Anything else you would like to add:**

This feature is crucial for production environments where secure access to metrics is a requirement.

Refs:
- https://keda.sh/docs/2.17/scalers/prometheus/#authentication-parameters

PR #4319, opened 2025-03-19
Envoy AI Gateway integration

sivanantha321:
**What this PR does / why we need it**:
This diagram illustrates the architecture of AI Gateway integration with KServe InferenceService using Gateway API in RawDeployment mode.

```mermaid
graph TB
    subgraph "User Input"
        U[User] --> ISVC_YAML[InferenceService YAML<br/>- AI Gateway annotation set to true<br/>- RawDeployment mode<br/>- TrafficPolicy with RateLimit<br/>- Gateway API enabled]
    end

    subgraph "KServe Control Plane"
        ISVC_YAML --> ISVC_CONTROLLER[InferenceService Controller]
        ISVC_CONTROLLER --> ISVC_CRD[(InferenceService CRD)]
        
        subgraph "AI Gateway Reconcilers"
            ISVC_CONTROLLER --> ASB_RECONCILER[AIServiceBackend Reconciler]
            ISVC_CONTROLLER --> AGR_RECONCILER[AIGatewayRoute Reconciler]
            ISVC_CONTROLLER --> BTP_RECONCILER[BackendTrafficPolicy Reconciler]
        end
    end

    subgraph "Created Resources"
        ISVC_CONTROLLER --> PREDICTOR_DEPLOY[Predictor Deployment<br/>- Hugging Face Model<br/>- VLLM Backend<br/>- OpenAI Compatible API]
        PREDICTOR_DEPLOY --> K8S_SERVICE[Kubernetes Service<br/>- Predictor Service<br/>- ClusterIP/LoadBalancer]
        
        ASB_RECONCILER --> ASB[AIServiceBackend<br/>- Links to K8s Service<br/>- OpenAI Schema Config<br/>- Backend Reference]
        AGR_RECONCILER --> AGR[AIGatewayRoute<br/>- Unified AI API<br/>- Schema Definition<br/>- Route Configuration]
        BTP_RECONCILER --> BTP[BackendTrafficPolicy<br/>- Rate Limiting Rules<br/>- Token-based Tracking]
        
        %% Connect to respective management systems
        ASB --> ASB_MGMT
        AGR --> AGR_MGMT
        BTP --> BTP_MGMT
    end

    subgraph "Gateway Infrastructure"
        subgraph "Envoy AI Gateway"
            EAG[Envoy AI Gateway<br/>- External Processing<br/>- Model Routing<br/>- Content Translation]
            EAG --> ASB_MGMT[AIServiceBackend Management]
[...]

PR #4522, opened 2025-06-13
[API] Define LLMInferenceService and LLMInferenceServiceConfig types and CRDs

### LINKED ISSUE #4521: [API] Define LLMInferenceService and LLMInferenceServiceConfig types and CRDs
/kind feature

We need to define the Go types for the LLMInferenceService and LLMInferenceServiceConfig Custom Resources. These types will serve as the foundation for our new controller. Once the types are defined, we will generate the Custom Resource Definition (CRD) manifests.

## Acceptance Criteria

Go type definitions for LLMInferenceServiceSpec and LLMInferenceServiceStatus are created in pkg/apis/serving/v1alpha1/llminferenceservice_types.go.
Go type definition for LLMInferenceServiceConfig is created.
The CRD manifests are generated using controller-gen.
The API types should align with the proposal document, including fields for model, template, router, etc.

[Design document](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit)

pierDipi:

Currently, KServe InferenceService API offers some level of GenAI focused features, however, the API was designed before GenAI and transformer based models were popular, and for LLM use cases, there is a need for more specialized configurations (disaggregated serving, llm-aware routing, multi-node, and much more) that don’t match with the generic and runtime agnostic nature of the InferenceService API.

Define LLMInferenceService and LLMInferenceServiceConfig types and CRDs.

The main changes are in `pkg/apis/serving/v1alpha1/llm_inference_service_types.go`

Fixes #4521

**Type of changes**

**Checklist**:

**Release note**:

```release-note
Add LLMInferenceService and LLMInferenceServiceConfig types and CRDs.
```

**Re-running failed tests**


go.mod
```diff
@@ -171,3 +170,5 @@ require (
 // Instead of replacing the path directly, exclude the problematic package
 // This avoids "module used for two different module paths" error
 exclude google.golang.org/grpc/stats/opentelemetry v0.0.0-00010101000000-000000000000
+
+replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.19.7
```
pierDipi: KEDA requires 0.19 but GIE uses 0.21

bartoszmajsak: I would make a comment directly there, so we will have an explanation in one place

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// Template for the worker pod spec. In a multi-node case, this configures the "worker" pod.
+	// +optional
+	Template *corev1.PodSpec `json:"template,omitempty"`
+
+	// Worker configuration for multi-node deployments.
+	// +optional
+	Worker *corev1.PodSpec `json:"worker,omitempty"`
+}
+
```
Jooho, line 114: It seems like both Template and Worker are podSpecs and both are related to multi-node deployments. How are they different?

Jooho: btw, is the `template` in this spec conflicting with the `template` in `LLMInferenceService`?

go.mod
sivanantha321: @pierDipi you can update the keda version as done in [here](https://github.com/kserve/kserve/pull/4319/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R20)

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,262 @@
...
+// LLMModelSpec defines the model source.
+type LLMModelSpec struct {
```
Jooho, line 116: I noticed that `LLMModelSpec` doesn’t seem to be used anywhere at the moment. Just wondering — is that by design?

Jooho: Based on the design doc, it looks like `LLMModelSpec` is meant to be used for `spec.model`

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,262 @@
...
+	// +optional
+	Parallelism *ParallelismSpec `json:"parallelism,omitempty"`
+
+	// Router configuration for the service.
+	// +optional
+	Router *RouterSpec `json:"router,omitempty"`
+
+	// Prefill configuration for disaggregated serving.
+	// +optional
+	Prefill *WorkloadSpec `json:"prefill,omitempty"`
```
Jooho, line 80: It looks like Prefill also needs to have parallelism, but currently there’s no dedicated `WorkloadSpec` for it.
Do you think it makes sense to create a separate struct for Prefill that includes both `WorkloadSpec` and `ParallelismSpec`?

Since `WorkloadSpec` is also used for multi-node workers, and those don’t need parallelism, it feels cleaner to separate them.WDYT?

Jooho:
Is distributed KV cache mandatory? I didn’t find a specific API to configure it, and based on the design doc, it seems to rely on the cache API. Just checking if that’s the expected behavior.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Fixed

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: I've added `Parallelism *ParallelismSpec` to `WorkloadSpec`

pierDipi:
> Is distributed KV cache mandatory? I didn’t find a specific API to configure it, and based on the design doc, it seems to rely on the cache API. Just checking if that’s the expected behavior.

The distributed cache api and design will be a follow up to the initial phases

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Fixed

Jooho:
While testing, I noticed one minor inconvenience, and I’m not sure if it’s worth adding as an API feature.

When using the HF protocol, we need to add `HF_TOKEN` under template.env, but since model pulling happens in multiple places, I have to repeat this for each deployment.(head/worker, p/d)
It might be nice to have a way to define common environment variables that apply across all deployments in the `llmInferenceServiceConfig`

It’s not critical — just a convenience improvement, so it’s something that could be considered later if needed.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
johnugeorge: Same question:  What is the difference between `template` and `worker`? Can you explicitly add comments against the fields used in multi-node and P/D scenario ?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,258 @@
...
+	// +optional
+	Refs []corev1.LocalObjectReference `json:"refs,omitempty"`
+
+	// Spec custom spec of the HTTPRoute.
+	// +optional
+	Spec *gatewayapi.HTTPRouteSpec `json:"spec,omitempty"`
+}
+
+// GatewaySpec defines the configuration for Gateway.
+// Spec and Refs are mutually exclusive.
```
sivanantha321: The Spec field is missing. Is this intentional ?

terrytangyuan (commented):
Comments to be addressed:
1. A section to the doc with all of the planned configurations for DP, PP and TP and P/D scenarios
2. Our plan with LWS

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: I've added a few more comments
```golang
	// Template for the main pod spec.
	// In a multi-node deployment, this configures the "head" or "master" pod.
	// In a disaggregated deployment, this configures the "decode" pod if it's the top-level template,
	// or the "prefill" pod if it's within the Prefill block.
	// +optional
	Template *corev1.PodSpec `json:"template,omitempty"`

	// Worker configuration for multi-node deployments.
	// The presence of this field triggers the creation of a multi-node (distributed) setup.
	// This spec defines the configuration for the worker pods, while the main 'Template' field defines the head pod.
	// The controller is responsible for enabling discovery between head and worker pods.
	// +optional
	Worker *corev1.PodSpec `json:"worker,omitempty"`

```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+	// The controller is responsible for enabling discovery between head and worker pods.
+	// +optional
+	Worker *corev1.PodSpec `json:"worker,omitempty"`
+}
+
+// LLMModelSpec defines the model source and its characteristics.
+type LLMModelSpec struct {
+	// URI of the model, specifying its location, e.g., hf://meta-llama/Llama-4-Scout-17B-16E-Instruct
+	// The storage-initializer init container uses this URI to download the model.
+	URI apis.URL `json:"uri"`
```
johnugeorge, line 119: @yuzisun   Any reason to use storage uri as a string in the Inference service?

https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/predictor.go#L91

Jooho:
> A section to the doc with all of the planned configurations for DP, PP and TP and P/D scenarios

Looks like the combination would be (DP + EP +TP) and (TP + PP) for now. (refer to [DP pr](https://github.com/vllm-project/vllm/pull/19008) Do we need DP scenario for the first phase? or just doc? 

>Our plan with LWS

The first phase we decided to implement the single case, and then we will move to multinode, P/D. If we already decided to go with LWS for multi-node, then we need to add LWS example to the doc.  If we need something to API for LWS, it can be added later.

pierDipi:
> > A section to the doc with all of the planned configurations for DP, PP and TP and P/D scenarios
> 
> Looks like the combination would be (DP + EP +TP) and (TP + PP) for now. (refer to [DP pr](https://github.com/vllm-project/vllm/pull/19008) Do we need DP scenario for the first phase? or just doc?

DP is not part of the initial goals and so we can discuss it separately.

> 
> > Our plan with LWS
> 
> The first phase we decided to implement the single case, and then we will move to multinode, P/D. If we already decided to go with LWS for multi-node, then we need to add LWS example to the doc. If we need something to API for LWS, it can be added later.

I agree, for now, I'd continue to use the current multi node approach with Deployments and maybe later add LWS support, generally the API should be independent on how it's been deployed as "template" and "worker" are podSpec which are also part of the LWS API https://lws.sigs.k8s.io/docs/reference/leaderworkerset.v1/#leaderworkerset-x-k8s-io-v1-LeaderWorkerTemplate

johnugeorge:
In what scenario, will the `template` in `WorkerSpec` be empty ?   I see that Multi node, with worker pod spec overrides section in the doc ([Link](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit?tab=t.0#heading=h.r91h8k1md6y7))  has `template` empty but how does it work without a head pod?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec LLMInferenceServiceSpec `json:"spec,omitempty"`
+}
+
+// LLMInferenceServiceSpec defines the desired state of LLMInferenceService.
+type LLMInferenceServiceSpec struct {
+	// Type of the deployment, e.g., 'default' or 'llm-d'.
+	// 'default' uses standard Kubernetes Deployments. 'llm-d' would imply a specific disaggregated architecture.
+	Type string `json:"type"`
```
yuzisun: Can we infer the type instead of adding an explicit field? this is mainly yaml UX.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
yuzisun: I think it is confusing to `apis.URL` here as storage uri is not a HTTP url.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+
+	// Name is the name of the model as it will be set in the "model" parameter for an incoming request.
+	// If omitted, it will default to `metadata.name`. For LoRA adapters, this field is required.
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// Criticality defines how important it is to serve the model compared to other models.
+	// This is used by the Inference Gateway scheduler.
+	// +optional
+	Criticality *igwapi.Criticality `json:"criticality,omitempty"`
```
yuzisun, line 129: why not priority which is a more familiar concept.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+	Refs []UntypedObjectReference `json:"refs,omitempty"`
+}
+
+// SchedulerSpec defines the Inference Gateway extension configuration.
+type SchedulerSpec struct {
+	// Pool configuration for the InferencePool, which is part of the Inference Gateway extension.
+	// +optional
+	Pool *InferencePoolSpec `json:"pool,omitempty"`
+
+	// Template for the Inference Gateway Extension pod spec.
```
yuzisun, line 226: Is this the envoy ext proc server pod or scheduler pod, if I understand correctly: envoy -> extproc -> scheduler? My question is that if user needs to customize this pod.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+	// in addition to the main 'decode' deployment, inspired by the llm-d architecture.
+	// This allows for independent scaling and hardware allocation for prefill and decode steps.
+	// +optional
+	Prefill *WorkloadSpec `json:"prefill,omitempty"`
+
+	// BaseRefs allows inheriting and overriding configurations from one or more LLMInferenceServiceConfig instances.
+	// The controller merges these base configurations, with the current LLMInferenceService spec taking the highest precedence.
+	// When multiple baseRefs are provided, the last one in the list overrides previous ones.
+	// +optional
+	BaseRefs []corev1.LocalObjectReference `json:"baseRefs,omitempty"`
```
yuzisun, line 86: Is there a use case when you need multiple `LLMInferenceServiceConfig` ?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// WorkloadSpec defines the configuration for a deployment workload, such as replicas and pod specifications.
+type WorkloadSpec struct {
+	// Number of replicas for the deployment.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
yuzisun, line 93: For now, minimally we need to support autoscaling for single node deployment, so we need additional field for maxReplica.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,304 @@
...
+// StorageSpec is a copy of the v1beta1.StorageSpec. It is duplicated here to avoid
+// import cycles between the v1alpha1 and v1beta1 API packages.
+type StorageSpec struct {
+	// The path to the model object in the storage. It cannot co-exist
+	// with the storageURI.
+	// +optional
+	Path *string `json:"path,omitempty"`
+	// The path to the model schema file in the storage.
+	// +optional
+	SchemaPath *string `json:"schemaPath,omitempty"`
```
yuzisun: I think this model schema field is not relevant for LLM model, maybe chat completion template file?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Yes, we can drop it for now

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Inference Gateway uses this keyword, I'd keep the name consistent with their terminology to be consistent and leverage their documentation and future references.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: It might depend on the implementation, however, the Istio implementation deploys the Envoy instance based on the Gateway API resource (this is provided and we don't create them, so any change to that pod is out of our control). 
The Gateway pod (basically Envoy) is directly connected to the scheduler (epp) via `ExtProc`, the scheduler pod spec is what is provided here as `Template`.


`inference-gateway-istio-7677bff484-gg982` is the Gateway pod, `vllm-llama3-8b-instruct-epp-59c7d4547d-7ptkr` is the epp/extproc, `vllm-llama3-8b-instruct-77ff4f95f9-*` are the model pods.
```
$ k get pods -A
NAMESPACE            NAME                                           READY   STATUS    RESTARTS   AGE
default              inference-gateway-istio-7677bff484-gg982       1/1     Running   0          41m
default              vllm-llama3-8b-instruct-77ff4f95f9-85hkn       1/1     Running   0          50m
default              vllm-llama3-8b-instruct-77ff4f95f9-98j7w       1/1     Running   0          50m
default              vllm-llama3-8b-instruct-77ff4f95f9-pzzx2       1/1     Running   0          50m
default              vllm-llama3-8b-instruct-epp-59c7d4547d-7ptkr   1/1     Running   0          49m
istio-system         istio-ingressgateway-699446c6dc-zhhq2          1/1     Running   0          45m
istio-system         istiod-5d46f44bfc-z78bn                        1/1     Running   0          48m
...
```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Yes, we want to provide presents (partial schemas) for multiple use cases, having multiple of them here, allows composing them in very interesting ways.

For example, as a platform team you could offer common configs for `config-llm-decode-template-nvidia-small` [1], `config-llm-prefill-template-nvidia-medium` or `kserve-config-scheduler` [2] and when deploying the service, a developer can compose them here without repeating them:

```yaml
baseRefs:
- config-llm-decode-template-nvidia-small
- config-llm-prefill-template-nvidia-medium
- config-llm-scheduler-small
```

[1]
```yaml
metadata:
  name: config-llm-decode-template-nvidia-small
  namespace: kserve
spec:
  template:
    containers:
      - image: ghcr.io/llm-d/llm-d:0.0.8
        imagePullPolicy: IfNotPresent
        name: vllm
        ports:
          - containerPort: 5557
            protocol: TCP
          - containerPort: 80
            protocol: TCP
        args:
          - --gpu_memory_utilization=0.9
          - --max_model_len=65536
          - --served-model-name
          - "{{ .Spec.Model.Name }}"
          - --port
          - "8001"
        command:
          - vllm
          - serve
          - "{{ .Spec.Model.Name }}"
        env:
          - name: HOME
            value: /home
          - name: VLLM_LOGGING_LEVEL
            value: INFO
          - name: VLLM_NIXL_SIDE_CHANNEL_PORT
            value: "5557"
          - name: VLLM_NIXL_SIDE_CHANNEL_HOST
            valueFrom:
              fieldRef:
                apiVersion: v1
                fieldPath: status.podIP
          - name: UCX_TLS
            value: ^cuda_ipc
          - name: HF_HUB_CACHE
            value: /models
        resources:
          limits:
            nvidia.com/gpu: "1"
          requests:
            nvidia.com/gpu: "1"
        securityContext:
          capabilities:
            drop:
              - MKNOD
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: FallbackToLogsOnError
        volumeMounts:
          - mountPath: /home
            name: home
          - mountPath: /dev/shm
            name: dshm
          - mountPath: /models
            name: model-cache
    initContainers:
      - name: llm-d-routing-sidecar
        imagePullPolicy: IfNotPresent
        image: ghcr.io/llm-d/llm-d-routing-sidecar:0.0.6
        restartPolicy: Always
        ports:
          containerPort: 8000
          protocol: TCP
        resources: { }
        securityContext: { }
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: FallbackToLogsOnError
        args:
          - "--port=8080"
          - "--vllm-port=8081"
          - "--connector=nixlv2"

    terminationGracePeriodSeconds: 30
    tolerations:
      - effect: NoSchedule
        key: nvidia.com/gpu
        operator: Exists
    volumes:
      - emptyDir: { }
        name: home
      - emptyDir:
          medium: Memory
          sizeLimit: 1Gi
        name: dshm
      - emptyDir: { }
        name: model-cache
```

[2]
```yaml
metadata:
  name: config-llm-scheduler-small
  namespace: kserve
spec:
  router:
    scheduler:
      pool:
        spec:
          extensionRef:
            failureMode: FailOpen
      template:
        containers:
          - name: main
            image: …
            resources:
              requests:
                memory: 4Gi
                cpu: 2
              limits:
                memory: 4Gi
                cpu: 4
```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Good point, I think for the chat template, it should already be possible via vllm argument `--chat-template` + custom CM volume via the various `PodSpec` (maybe we can reconsider in the future?)

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: As an additional point, we want to enable web UIs to discover and leverage config-presents and there it's not easy to use things like Kustomize to overlay configs

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
Jooho: @pierDipi I also agreed with @johnugeorge @yuzisun so could you please follow up on this?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
Jooho: I think it's better to use the name from the inferenceGateway to prevent confusion.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
Jooho: In case of conflicting API configurations, does the latter override the former?

pierDipi: Yes, it's described in the field comment:

```
the last one in the list overrides previous ones.
```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Yes, we need this, I want to re-use isvc types for that to avoid copying them to eventually get the #4371 but the current import cycle issue v1alpha1 -> v1beta1 -> v1alpha1 requires a refactoring that is unrelated to this PR.

can we follow up wrt autoscaling API?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
yuzisun: Could you help update the description so it is more clear that the template is for configuring the scheduler(epp) pod possibly with a diagram like gateway(envoy) -> epp(extproc) -> model ?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: From a usage perspective it's a string (meaning users will provide a string) but it's nicer to have types with a parsed URL, while it's not HTTP it's still a URL in general:

```yaml
spec:
  model:
    uri: hf://...
```

as we can just access the parsed URL properties (ie scheme, etc) without parsing it over and over in the controller.

Here is how it's defined

```golang
// URL is an alias of url.URL.
// It has custom json marshal methods that enable it to be used in K8s CRDs
// such that the CRD resource will have the URL but operator code can can work with url.URL struct
// +kubebuilder:validation:Type=string
type URL url.URL
```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: https://github.com/kserve/kserve/pull/4522/commits/5c85a58645dc93c32f80cda1d7ae36d43aa4011b

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: Updated the comment, instead of embedding the Gateway Spec, we will just take refs of pre-existing Gateway(s)

pierDipi:
> In what scenario, will the `template` in `WorkerSpec` be empty ? I see that Multi node, with worker pod spec overrides section in the doc ([Link](https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit?tab=t.0#heading=h.r91h8k1md6y7)) has `template` empty but how does it work without a head pod?

Good question, the way we envision it is: there will be well-known set of `LLMInferenceServiceConfig` that are overlayed based on what's being opted into and these `LLMInferenceServiceConfig`(s) are provided as part of the Kserve installation, however, they can be further customized before hand by a platform team (or any other external automation/operator) to match and request the hardware available in the cluster.

For example, in the multi-node case, when `spec.worker` is non-nil pointer we would always overlay a (for example) `kserve-config-llm-worker` which configures both the `spec.worker` PodSpec and the `spec.template` PodSpec (with Ray configurations and commands), similar to the current "multi node serving runtime".

However, generally, I expect users to override their `spec.worker` and `spec.template` in the multi node case, either using the way described above or via additional explicit `baseRefs`, or directly in the `LLMInferenceService` resource.

terrytangyuan (approved):
/lgtm

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
lizzzcai: it seems like the lora adatpor is using `storageUri` but the base model is using `uri`? maybe should make it the same.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
lizzzcai: I looked at the `LLMInferenceServiceConfig`, it seems similar to `ServingRuntime`/`ClusterServingRuntime`, is it possible to reuse it in some way so we have less CRD.

lizzzcai:
Thanks for the PR, just saw this PR. I have a couple of comments regarding the approach to supporting LLM use cases.

Consider extending the existing `InferenceService` CRD instead of creating a new one. Refining the `Predictor` or adding a new component like an `LLMPredictor` (or a better name) would allow us to reuse features like `transformers` and `explainers` which are broadly applicable, and it would reduce our CRD count.

If a new CRD is deemed necessary, I suggest a more generic name like `GenerativeInferenceService` instead of `LLMInferenceService`. This would better accommodate future support for models across different modalities (image, video, voice) and avoid user confusion.

Happy to hear the views from other as well.

pierDipi:
@lizzzcai the design doc has more details on why we went this way and we had a few discussions on the new API name as the scope is really LLM models specific since other types of Generative models have very different serving needs and topology, see https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: @lizzzcai `*ServiceRuntime` CRDs assume a single pod template generally (or a subset of it) and has some fields which are not applicable to LLMInferenceService, we need more multiple sub-components that will be deployed as part of the same `LLMInferenceService` object.

Generally the goal with having a specific config object for `LLMInferenceService` is to have clearer semantic and schema.

srampal:
> @lizzzcai the design doc has more details on why we went this way and we had a few discussions on the new API name as the scope is really LLM models specific since other types of Generative models have very different serving needs and topology, see https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit

@pierDipi 
1) Please provide a reference to the section in the design doc that describes why this API name is LLM specific. Even the ModelService CRD from llm-d does not have the term LLM in its name.
2) Please add documentation on the motivation for this CRD in comparison with the llm-d ModelService CRD/ Helm chart. Is the goal for this feature to be the Operator version, with more active reconciliation but otherwise configure the same functions as deployed by the ModelService helm chart ? Or will this be a superset of the resources deployed by the ModelService helm chart ?  What are the best practice recommendations for users to deploy the ModelService helm chart vs this api ?

yuzisun:
> > @lizzzcai the design doc has more details on why we went this way and we had a few discussions on the new API name as the scope is really LLM models specific since other types of Generative models have very different serving needs and topology, see https://docs.google.com/document/d/11ZQJ2VhTc42S9K4yau2dMs3Q3f4jqWJL_7Sq14C3hzY/edit
> 
> @pierDipi
> 
> 1. Please provide a reference to the section in the design doc that describes why this API name is LLM specific. Even the ModelService CRD from llm-d does not have the term LLM in its name.


LLMInferenceService aims to provide a unified and scalable interface for running large language model inference across both single-node and multi-node deployments. It supports advanced features like disaggregated prefilling to optimize performance and resource efficiency, enabling high-throughput, low-latency inference at scale. Since P/D optimization is very specific for LLMs hence the prefix as it is not clear how it applies to other models like diffusions, however it can support MultiModal inputs. 

ModelService does not really convey much information and kserve already have InferenceService for predictive and simple generative inference use cases.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -0,0 +1,313 @@
...
+	// When multiple baseRefs are provided, the last one in the list overrides previous ones.
+	// +optional
+	BaseRefs []corev1.LocalObjectReference `json:"baseRefs,omitempty"`
+}
+
+// WorkloadSpec defines the configuration for a deployment workload, such as replicas and pod specifications.
+type WorkloadSpec struct {
+	// Number of replicas for the deployment.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
Yanghaihai1020, line 93: In the multi-machine inference scenario, does this "Replicas" refer to the number of Decode/Prefill instance groups? And the number of workers in each instance group is determined by "Parallelism"?

PR #4527, opened 2025-06-13
Resolve inference endpoint using runtime protocol when applicable

israel-hdez:

Typically, protocol v1 is assumed. However, there is a corner case: if the runtime name is specified in the InferenceService, then kserve-controller won't validate that the protocol in the InferenceService and the ServingRuntime matches (likely, assuming the user knows what is doing). More importantly, if the protocol is not specified in the InferenceService, the GetPredictorEndpoint function returns the V1 endpoint even if the ServingRuntime only supports V2 (and correctly specifies it). This has an impact on the InferenceGraph feature because the kserve-router will be misconfigured.

These changes cover this corner case to properly resolve the protocol.

**Type of changes**

1. Deploy an InferenceService using a runtime that only supports protocol v2 (e.g. Triton). Avoid auto-selection by specifying the `runtime` name in the model spec. Omit specifying the `protocolVersion` in the ISVC.
2. Deploy an InferenceGraph using the InferenceService.
3. Notice the graph workload is correctly configured with the v2 endpoint, instead of v1 endpoint.

Also, do sanity checks for other cases.

**Checklist**:

- N/A Have you made corresponding changes to the documentation?

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


Jooho (commented):
/lgtm

PR #4536, opened 2025-06-22
fix: hard-coded gateway class name in helm chart

vitush93:


yuzisun:
done in https://github.com/kserve/kserve/pull/4544/files

PR #4544, opened 2025-06-26
fix(helm): make gatewayClassName configurable

### LINKED ISSUE #4720: Configurable GatewayClass
/kind feature

Is there a requirement for envoy to be GatewayClass? It is hardcoded in https://github.com/kserve/kserve/blob/88cf57e5a069ee1213d4a133c8f070950dc2bbfa/charts/kserve-resources/templates/ingress_gateway.yaml#L8
If one wants to use Kgateway as GatewayClass for example, it should be configurable. Otherwise documentation should state that it is not allowed.

bartoszmajsak:

Previously hardcoded "envoy" value prevented from using other Gateway API implementations unless changed by hand after installation

- Template now uses `className` defined `values.yaml`
- Backwards compatibility with "envoy" default

Fixes #4720

**Type of changes**

**Checklist**:

**Release note**:

Not sure we need release notes for this? It's been exposed through the helm values but never applied.

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/charts/kserve-resources/files/kserve/ingress_gateway-patch.yaml
+++ b/charts/kserve-resources/files/kserve/ingress_gateway-patch.yaml
@@ -4,7 +4,7 @@ kind: Gateway
 metadata:
   name: kserve-ingress-gateway
 spec:
-  gatewayClassName: envoy
+  gatewayClassName: {{ .Values.kserve.controller.gateway.ingressGateway.className | default "envoy" }}
   listeners:
    - name: http
      port: 80


pierDipi (approved):
/lgtm

Ref in values https://github.com/kserve/kserve/blob/ce4304052c6867841eba068227e7c1b09f895003/charts/kserve-resources/values.yaml#L195

bartoszmajsak:
> /lgtm
> 
> Ref in values
> 
> https://github.com/kserve/kserve/blob/ce4304052c6867841eba068227e7c1b09f895003/charts/kserve-resources/values.yaml#L195

Right, I wonder if that wouldn't break anything, as it will de-facto be a new default now. Not sure how the manifests are used.

charts/kserve-resources/files/kserve/ingress_gateway-patch.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: kserve-ingress-gateway
   namespace: {{ .Release.Namespace }}
 spec:
-  gatewayClassName: envoy
+  gatewayClassName: {{ .Values.kserve.controller.gateway.ingressGateway.className | default "envoy" }}
```
Jooho, line 7: https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L654

default would be `istio`?

bartoszmajsak: That's correct. I think that was the intention but why is it hardcoded to `envoy` I do not know.

Jooho: so will we keep the envoy as a default here?

israel-hdez: Weirdly, I'm not seeing this resource under `config/`, which used to be the source of truth for install. I wonder if kustomize is somehow already deprecated?

About `className` it goes to the ConfigMap: https://github.com/kserve/kserve/blob/a070e2c7d9db74ec3dd52897c8f1ef96b003f402/charts/kserve-resources/templates/configmap.yaml#L630

It is for the case when GW API is disabled. Then kserve-controller will, instead, create `networking.k8s.io/v1/Ingress` resources configuring them with the specified class name. 

I would not have any objection on re-using that helm values field for GW API case, but see that its default value is set to `istio` rather than `envoy` :thinking: : https://github.com/kserve/kserve/blob/a070e2c7d9db74ec3dd52897c8f1ef96b003f402/charts/kserve-resources/values.yaml#L194-L195

Perhaps that note `If Gateway API is enabled, this will not affect the ingress routing.` is trying to tell that it is not applicable for GW API case? IDK, but I emphazise that I would not have any objection on re-using the field.

sivanantha321: As @israel-hdez pointed out, ingressClassName is used for k8s Ingress case if gateway api is not enabled. If gateway api is enabled, ingressClassName field is ignored. Note that, current implementation does not allow using k8s ingress and gateway api at the same time. We plan to integrate envoy ai gateway with kserve, that's why envoy is chosen as the default for gateway api.

The gateway resource is not part of kustomize because we don't have control over when it is installed unlike helm. in helm it will be only installed if gateway api is enabled and create gateway flag is set.

The intention for creating the gateway as part of installation is for quick setup of kserve for development.

sivanantha321, line 7: @bartoszmajsak maybe define the default value in the values.yaml ? so it will be documented in the README.

sivanantha321: Please, run the `make precommit` to auto update the README.

Jooho: I believe that having a gap between Helm and Kustomize can be confusing for users, regardless of the reasons behind it. Additionally, if the default values for the same field differ across environments, it might cause even more confusion. With that in mind, perhaps it would be better to separate them entirely. Although there has been a plan to split Helm into two charts for the new API, I think it’s important to provide more detailed guides and overlays specifically for users who rely on Kustomize.

By the way, which one is the source of truth? (Kustomize or Helm? I think Kustomize should be one.)

sivanantha321: @Jooho I believe both installations maintain consistency. even if you enable gateway api during installation with helm, it is not going to create gateway resource unless and until you explicitly set the createGateway flag. I am not sure which field you are referring to. In both installations gateway api is disabled by default. As for the k8s ingress both installations defaults to `istio`. btw, I believe kustomize installation is not a recommended practice. Usually the assets in the github release is used for installation as described in the [docs](https://kserve.github.io/website/latest/admin/kubernetes_deployment/#3-install-kserve). but i am okay with creating a overlay for gateway api with gateway resource.

israel-hdez: > but i am okay with creating a overlay for gateway api with gateway resource.

Perhaps, in can make sense to have a `config/optional` directory holding configs that don't fit on the base install. If end-users installation with kustomize is no longer recommended, I think it still makes sense at least for development for having something that we can setup quickly.

yuzisun: The Kustomize method is mostly consumed by kubeflow, for kserve we recommend using helm to install. @sivanantha321 we can explicitly document on the new site?

bartoszmajsak:
Sorry for being behind with this PR. I'll sort it out mid next week when I'm back from the longer weekend offline.

spolti (approved):
/lgtm

Jooho:
@bartoszmajsak It’s already applied to master, but the default value is istio. If you want to change it to envoy, please update it here:
https://github.com/kserve/kserve/blob/master/charts/_common/kserve-resources-specific.yaml#L58

Jooho (approved):
/lgtm
/approved

PR #4546, opened 2025-06-26
Stop and resume an explainer

hdefazio:
Adds the ability to stop and resume a explainer if it is present with the ISVC. This PR covers both serverless and raw deployment.

Follow up from https://github.com/kserve/kserve/issues/4207

**Type of changes**

**Re-running failed tests**


pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -657,38 +657,57 @@ func (r *RawHTTPRouteReconciler) reconcileExplainerHTTPRoute(ctx context.Context
 	if err != nil {
 		return err
 	}
+
+	// reconcile httproute
```
VedantMahabaleshwarkar, line 661: nit: this block is re-used multiple times. I'd prefer if it was extracted into a function

VedantMahabaleshwarkar (approved):
LGTM

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
hdefazio: I think this will be covered by the follow up to refactor these to use CheckResultType anyway

terrytangyuan (approved):

PR #4556, opened 2025-06-27
fix: unset clampMax and clampMin, since they are not for replicas

### LINKED ISSUE #4551: [Autoscaling KEDA] Leave clampMax and clampMin empty, since minReplicaCount and maxReplicaCount are enough
/kind bug

I found the issue of keda: minReplicaCount and maxReplicaCount are good to enough to set the min and max number of pods. The other fields like clampMax and clampMin should not be set, because these two fields set the lower and upper bound of the scale itself(if we check the vllm_num_requets_running, the min will be set to clampMin, the max will be set to clampMax. they do not actually set the bound for the number of pods :disappointed:), not the number of the replicas(the official doc could be wrong in this regard)

Let's simply remove the setting of clampMax and clampMin.
As long as minReplicaCount and maxReplicaCount are set, we can control the lower and upper bound of the number replicas.

houshengbo:

The fields like clampMax and clampMin should not be set, because these two fields set the lower and upper bound of the scale itself.

Fixes #4551 

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
@@ -144,8 +144,6 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec,
 				if triggerType == string(constants.AutoScalerMetricsSourceOpenTelemetry) {
 					trigger.Type = "external"
 					trigger.Metadata = map[string]string{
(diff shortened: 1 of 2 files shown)


houshengbo:
@yuzisun @andyi2it Please review this PR. The description is as in the issue.

yuzisun (approved):

PR #4564, opened 2025-07-01
feat(envtest): simplifies CRD lookup

bartoszmajsak:

Relying on relative paths for resolving CRD to be loaded when starting EnvTest is error prone.

This PR introduces a simple `GetProjectRoot()` lookup function that traverses project tree up until it finds first `go.mod` file and assumes this is the root of the project. That is sufficient to make EnvTest tests resilient.

Also... makes the code legal ;)

**Type of changes**

Changes are for EnvTest bootstrap logic so if those tests pass we should be good :)

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/testing/envtest_setup.go
+++ b/pkg/testing/envtest_setup.go
@@ -17,21 +17,16 @@ limitations under the License.
 package testing
 
 import (
-	"context"
-	"sync"
-
 	"google.golang.org/protobuf/proto"
 	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
 	netv1 "k8s.io/api/networking/v1"
 
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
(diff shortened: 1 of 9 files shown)


pkg/testing/envtest_setup.go
```diff
@@ -69,14 +62,3 @@ func SetupEnvTest(crdDirectoryPaths []string) *envtest.Environment {
...
-// StartTestManager adds recFn
-func StartTestManager(ctx context.Context, mgr manager.Manager, g *gomega.GomegaWithT) *sync.WaitGroup {
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		g.Expect(mgr.Start(ctx)).NotTo(gomega.HaveOccurred())
-	}()
-	return wg
-}
```
bartoszmajsak, line 82: Seems to be dead code.

sivanantha321 (approved):

PR #4566, opened 2025-07-02
feat: Add disable postprocessing option for raw logits

sivanantha321:
The Hugging Face server component had two issues that this change addresses:

BUG FIX: The --return_probabilities flag was not consistently applying softmax to convert logits to probabilities across all model types. For sequence classification, it was incorrectly returning raw logits instead of probabilities, while fill-mask models were properly returning probabilities.

ENHANCEMENT: Added a new --return_raw_logits option for users who specifically want to access raw logits from the model without any post-processing.

These changes provide consistent behavior for probability returns and additional flexibility for advanced use cases requiring access to raw model outputs:

- When return_raw_logits is set, the model returns unprocessed logits as a dictionary mapping indices to values
Supported for all classification tasks: sequence_classification, fill_mask, and token_classification

- Ensures consistent application of softmax when `return_probabilities=True` across all task types
Previously, softmax was properly applied only for token_classification but inconsistently in other tasks
Now all tasks use the same probability processing logic

- The two flags `return_probabilities` and `return_raw_logits` are implemented as mutually exclusive options in the CLI, allowing users to select between:

Fixes #

**Type of changes**

**Checklist**:

**Release note**:

```release-note
Add —return_raw_logits option
```

**Re-running failed tests**


python/huggingfaceserver/huggingfaceserver/__main__.py
```diff
@@ -138,10 +138,19 @@ def is_vllm_backend_enabled(
...
+# This group allows the user to choose between returning probabilities or disabling postprocessing.
+output_format_group = parser.add_mutually_exclusive_group()
+output_format_group.add_argument(
     "--return_probabilities",
     action="store_true",
-    help="Return all probabilities",
+    help="Return probabilities instead of logits for classification tasks such as token classification, text classification and fill-mask.",
+)
+output_format_group.add_argument(
+    "--disable_postprocess",
```
gavrissh: `--disable_postprocess` seems very broad?
Rather, make it specific `--return_raw_logits`

yuzisun (approved):

PR #4569, opened 2025-07-03
Added deployment name and namespace in otel collector

### LINKED ISSUE #4542: [Autoscaling KEDA] Differentiate the same metrics across different deployments under different namespaces
/kind feature

We leverage the OTEL-ADDON as the intermediate storage to save the metrics collected from the workloads, and serve them to KEDA via the PromQL queries.

If we have multiple services running under the same or even different namespaces, but all of them collect and scale based on the same metric, e.g. `vllm:num_requests_running`, the metrics collected will be all mixed together in the `OTEL-ADDON`.

It means if we scale the workloads with the query:
```
      - podmetric:
          metric:
            backend: opentelemetry
            metricNames:
            - vllm:num_requests_running
            operationOverTime: avg
            query: sum(vllm:num_requests_running)
``` 
we won't be able to differentiate the which metrics `vllm:num_requests_running` are coming from which deployment under which namespace.

Maybe I have missed some configuration options, but I am wondering if we support adding deployment name and namespace into the metrics, so that we can query with something like:
```
query: "sum(vllm:num_requests_waiting{namespace=<the namespace>, deployment=<the workload name>})"
```
to scale the workload, only based on the metrics coming from the specific workload.

I personally leverage the OpenTelemetry processors to add the namespace and the deployment name into the metrics collected from the workload like:
```
    processors:
      resourcedetection/env:
        detectors: [env]
        timeout: 2s
        override: false
      transform:
        metric_statements:
          - context: datapoint
            statements:
              - set(attributes["namespace"], resource.attributes["k8s.namespace.name"])
              - set(attributes["pod"], resource.attributes["k8s.pod.name"])
              - set(attributes["deployment"], resource.attributes["k8s.deployment.name"])
```
This is able to configure with the CR of OpenTelemetryCollector, but is glad to see if there is any other option to configure it as well.

andyi2it:
Adds support for differentiating metrics across namespace and workloads

Fixes #4542 

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/rawkube_controller_test.go
@@ -10313,6 +10313,18 @@ var _ = Describe("v1beta1 inference service controller", func() {
 													"targets": []interface{}{"localhost:8080"},
 												},
 											},
+											"relabel_configs": []interface{}{
+												map[string]interface{}{
+													"source_labels": []interface{}{},
+													"target_label":  "namespace",
+													"replacement":   predictorSObjectKey.Namespace,
+												},
+												map[string]interface{}{
+													"source_labels": []interface{}{},
+													"target_label":  "deployment",
+													"replacement":   predictorSObjectKey.Name,
+												},
+											},
 										},
 									},
 								},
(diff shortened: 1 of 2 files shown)


yuzisun:
done in https://github.com/kserve/kserve/pull/4593

PR #4570, opened 2025-07-03
Added advanced config for ScaleUp and ScaleDown

### LINKED ISSUE #4541: [Autoscaling KEDA] Make the stabilizationWindowSeconds for scaleUp and scaleDown configurable
/kind feature

In HPA, the default stabilizationWindowSeconds for scaleUp is 0, and the default stabilizationWindowSeconds for scaleDown is 5 minutes. The inference spec should be able to configure both of them with different values.

My suggestion is to able to add the following section into the scaledobject:

```
advanced:
  horizontalPodAutoscalerConfig:
    behavior:
      scaleDown:
        stabilizationWindowSeconds: 10
      scaleUp:
        stabilizationWindowSeconds: 10
```

Similar to configure the HPA section as below:
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#default-behavior

andyi2it:
Add AutoscalingBehavior spec from the k8s autoscaling API, this sets the KEDA's advancedConfig to control the autoscaling behavior.

Fixes #4541 

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -200,6 +212,25 @@ func createKedaScaledObject(componentMeta metav1.ObjectMeta,
 		},
 	}
 
+	if ScaleDownStabilizationWindowSeconds != nil || ScaleUpStabilizationWindowSeconds != nil {
```
houshengbo: Thanks for working on implementing this feature.
Just a suggestion here: if the user does not specify any StabilizationWindowSeconds, shall we provide some default values for both ScaleUp and ScaleDown? Right now, if both values are unset, the default ScaleUpStabilizationWindowSeconds is 0, and ScaleDownStabilizationWindowSeconds is 5 mins, if I remember correctly.

KServe, can offer configurations for this `Advanced` filed section, including the StabilizationWindowSeconds globally in the ConfigMap(In future, it may include more options). If users do not specify on the service level, we can still pick up these values from the ConfigMap.

What do you think?

pkg/apis/serving/v1beta1/component.go
```diff
@@ -137,6 +140,53 @@ type AutoScalingSpec struct {
 	Metrics []MetricsSpec `json:"metrics,omitempty"`
 }
 
+type AdvancedSpec struct {
```
yuzisun: this should be part of `AutoScalingSpec`

pkg/apis/serving/v1beta1/component.go
```diff
@@ -137,6 +140,53 @@ type AutoScalingSpec struct {
...
+	// HorizontalPodAutoscalerConfig contains the configuration for the Horizontal Pod Autoscaler.
+	HorizontalPodAutoscalerConfig *HorizontalPodAutoscalerConfig `json:"horizontalPodAutoscalerConfig,omitempty"`
+}
+
+// HorizontalPodAutoscalerConfig contains the configuration for the Horizontal Pod Autoscaler.
+type HorizontalPodAutoscalerConfig struct {
+	// Behavior contains the scaling behavior configuration for the Horizontal Pod Autoscaler.
+	// +optional
+	Behavior *HorizontalPodAutoscalerBehavior `json:"behavior,omitempty"`
```
yuzisun: can we reuse the type from kubernetes ?
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v2/#HorizontalPodAutoscalerSpec

pkg/apis/serving/v1beta1/component.go
houshengbo: This is trying to match what keda defined: https://github.com/kedacore/keda/blob/main/apis/keda/v1alpha1/scaledobject_types.go#L105.
KEDA leveraged `type AdvancedConfig struct {...}`.

pkg/apis/serving/v1beta1/component.go
houshengbo: https://github.com/kedacore/keda/blob/main/apis/keda/v1alpha1/scaledobject_types.go#L146
This is how keda used it.

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
andyi2it: Thanks @houshengbo 
Changes have been made to include in the config map.

pkg/apis/serving/v1beta1/component.go
andyi2it: @yuzisun 
As @houshengbo said, we are trying to match what Keda has defined. Should we match it with Kubernetes Spec.

pkg/apis/serving/v1beta1/component.go
andyi2it: @yuzisun Same as above, what do you think?

houshengbo: I think we can replace Line 152 with 
```
Behavior *autoscalingv2.HorizontalPodAutoscalerBehavior `json:"behavior,omitempty"`
```

We can use `import autoscalingv2 "k8s.io/api/autoscaling/v2"`. We can remove the definition for `HorizontalPodAutoscalerBehavior` along Line 156 and even `ScalingRules` along Line 166, since they are both defined in `k8s.io/api/autoscaling/v2`.

charts/kserve-resources/templates/configmap.yaml
```diff
@@ -697,4 +697,6 @@ data:
       "scrapeInterval": "5s",
       "metricReceiverEndpoint": "keda-otel-scaler.keda.svc:4317",
       "metricScalerEndpoint": "keda-otel-scaler.keda.svc:4318"
+      "scaleUpStabilizationWindowSeconds": "0",
+      "scaleDownStabilizationWindowSeconds": "300"
```
yuzisun, line 705: this section is for otel collector, these two settings can be moved out to the autoscaling section.

andyi2it: New section for autoscaling

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -215,6 +244,25 @@ func createKedaScaledObject(componentMeta metav1.ObjectMeta,
 		},
 	}
 
+	if scaleDownStabilizationWindowSeconds != nil || scaleUpStabilizationWindowSeconds != nil {
+		hpaBehavior := &autoscalingv2.HorizontalPodAutoscalerBehavior{}
```
yuzisun, line 248: Can we initialize with `componentExtension.AutoScaling.Behavior` and then override the ScaleDown, ScaleUp, otherwise what user specified on `componentExtension.AutoScaling.Behavior` is not taking effect in KEDA.

yuzisun (approved):

PR #4576, opened 2025-07-03
Add LLM InferenceService base configurations 

VedantMahabaleshwarkar:
Add comprehensive LLMInferenceServiceConfig templates:

- Decode template with vLLM server and routing sidecar
- Router route configuration with Gateway API integration
- Advanced scheduler with load balancing and prefix-aware scoring
- Base Presets for LWS taken from https://github.com/tlrmchlsmth/vllm-dp-lws/blob/main/lws.yaml 

Additional docs : https://docs.google.com/document/d/1mSYsWQEbp4Oq50ghFWUun7OgagbuUJivAC8Ds-pu-xU/edit?tab=t.0 

Inject GIE pool selector and EPP service account when unspecified.

**Type of changes**

**Checklist**:

- [n/a] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


VedantMahabaleshwarkar:
cc @pierDipi @bartoszmajsak @Jooho @yuzisun @sivanantha321 more follow ups for LLM-D

config/llmisvc/config-llm-decode-template.yaml
```diff
@@ -0,0 +1,85 @@
...
+          timeoutSeconds: 5
+          failureThreshold: 60
+        volumeMounts:
+          - mountPath: /home
+            name: home
+          - mountPath: /dev/shm
+            name: dshm
+          - mountPath: /models
+            name: model-cache
+    initContainers:
```
yuzisun, line 70: I thought this is a sidecar container not init container

config/llmisvc/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
yuzisun, line 2: Can we use helm instead of Kustomization ? KServe is going with the direction with helm as standard installation.

config/llmisvc/config-llm-decode-template.yaml
bartoszmajsak: It is a real Kubernetes sidecar container, as it's defined with `restartPolicy: Always` :)

Unless I'm missing something here https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

config/llmisvc/kustomization.yaml
Jooho: > KServe is going with the direction with helm as standard installation.

Just to clarify — does this mean we’ll only provide Helm for KServe and drop Kustomize support?
From my perspective, keeping Kustomize doesn’t really have any downside. I know many users still rely on Kustomize for their deployments, so I’d vote to keep it if possible

config/llmisvc/config-llm-decode-template.yaml
bartoszmajsak: @yuzisun There's one thing I wonder - as it's a fairly new feature - which versions of k8s are we targeting with the upcoming release? Should we care for clusters older than 1.29?

config/llmisvc/kustomization.yaml
VedantMahabaleshwarkar: agree to what @Jooho said above. @yuzisun is it fine to add the helm files in a follow up PR so as to keep the changes per PR small and easy to review?

pkg/controller/v1alpha1/llmisvc/config_presets_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"knative.dev/pkg/apis"
+	"sigs.k8s.io/yaml"
+
+	kservetesting "github.com/kserve/kserve/pkg/testing"
+
```
sivanantha321, line 37:
```suggestion
```

config/llmisvc/kustomization.yaml
VedantMahabaleshwarkar: @yuzisun this is the PR that adds the helm files https://github.com/kserve/kserve/pull/4583

terrytangyuan (commented):
@VedantMahabaleshwarkar Maybe rename the PR title to be LLM inference service instead of llm-d?

VedantMahabaleshwarkar:
I somehow messed up this branch while trying to cherrypick changes from https://github.com/kserve/kserve/pull/4596 and ended up with 10 commits 😅 
Replacing this PR with https://github.com/kserve/kserve/pull/4613

PR #4578, opened 2025-07-07
chore(utils): simplifies code using generics

bartoszmajsak:

In several of the webhook implementations there are duplicated functions to ensure that the received `client.Object` can be cast to the desired type. This can be uniformly handled through a single `Convert[T]` function.

In addition, `"k8s.io/utils/ptr"` package has been introduced that further simplifies the code base.

**Type of changes**

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


brettmthompson:
The use of `k8s.io/utils/ptr` allows for the removal of duplicate type specific pointer conversion utility functions and the addition of the type agnostic `Convert` utility function allows for the removal of duplicate type dependent object conversion utility functions. 

Both changes simplify the code and  lgtm

sivanantha321 (approved):

PR #4583, opened 2025-07-07
Helm charts for llmisvc configs created.

andresllh:
Add LLMInferenceServiceConfig templates helm chart

**Type of changes**

- [x] Run `cd charts/llmsivc-resources && helm install -f values.yaml llmisvc-resources . --namespace kserve --dry-run > test_install.yaml`
        Compare output with expected output  from [here](https://gist.github.com/VedantMahabaleshwarkar/9018342414a848ddc2d74e31bba4e07b)
- Logs

**Re-running failed tests**


bartoszmajsak:
@andresllh Thanks for spearheading this follow-up PR. Couple of things to consider from the top of my head:

These are well-known configurations, which means they have a structure that we have tested and know they can serve as reasonable defaults. With that many knobs exposed in helm templates, we allow users to do anything with this before they get applied on the cluster. As a consequence, they are no longer as "well known", and the result might not be the desired state we aimed for when creating and testing them. 

I think we should keep it simple. I would start with almost static content, like we have for other resources ([ingress gateway](https://github.com/kserve/kserve/blob/master/charts/kserve-resources/templates/ingress_gateway.yaml) being one example). What others think?

We should also decide where the source of truth is and automate publishing of the other "format". As we now have it, both as `kustomize` and `helm` manifests, we risk disconnect and potential bugs when things diverge.

andresllh:
Will be using the Makefile `manifests` target. @VedantMahabaleshwarkar will be making the update in his PR.

PR #4588, opened 2025-07-11
Stop and resume an inference graph

hdefazio:
Adds the ability to stop and resume an inference graph in serverless mode. This will only stop the inference graph and will not affect the associated inference services. 

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Follow up from https://github.com/kserve/kserve/issues/4207

**Type of changes**

create a serving runtime (ref: https://github.com/hdefazio/modelservermetrics_demo/blob/main/basic_demo/servingruntime.yaml)
`$ kubectl apply -f servingruntime.yaml `

create an inference service (ref: https://github.com/hdefazio/modelservermetrics_demo/blob/main/basic_demo/isvc.yaml )
`$ kubectl apply isvc.yaml`

create an inference graph:
```
kind: InferenceGraph
metadata:
  name: dog-pipeline
spec:
  nodes:
    root:
      routerType: Sequence
      steps:
        # Identify the dog breed
        - name: dog-breeds-classifier
          serviceName: dog-breeds-classifier
  resources:
    limits:
      cpu: 1
      memory: 1Gi
    requests:
      cpu: 100m
      memory: 256Mi
EOF
```

Observe that the isvc and ig are ready
`kubectl get isvc, ig`

Observe the inference graph's ksvc exists: 
`kubectl get ksvc`

Stop the inference graph: 
`$ kubectl edit ig/dog-pipeline `
Add annotation `serving.kserve.io/stop: "true"`

Observe that the ksvc (and related objects are removed and the inference graph is no longer ready)
Observe that the isvc is unaffected
`kubectl get isvc, ig, ksvc, pods, service`

Resume the inference graph: 
`$ kubectl edit ig/dog-pipeline `
Edit annotation `serving.kserve.io/stop: "false"`

Observe that the ig is ready and the ksvc (and related objects) were created. 
`kubectl get isvc, ig, ksvc, pods, service`

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -257,6 +260,40 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				LastTransitionTime: transition_time,
+				Type:               apis.ConditionReady,
+				Status:             corev1.ConditionFalse,
+				Reason:             v1beta1.StoppedISVCReason,
+			}
+			graph.Status.Conditions = []apis.Condition{stoppedCondition, readyCondition}
+
+			graph.Status.URL = nil
+		}
+	} else {
```
brettmthompson, line 296: Should we also check that the stopped condition is not already set to false here so we keep a correct record of when the service was resumed? With the current logic the LastTransitionTime of the stopped false condition will be overwritten with every reconciliation.

pkg/controller/v1alpha1/inferencegraph/controller_test.go
```diff
@@ -1041,6 +1059,256 @@ var _ = Describe("Inference Graph controller test", func() {
...
+		// Checks that any Kubernetes object to be not found.
+		expectResourceIsDeleted := func(ctx context.Context, obj client.Object, objKey types.NamespacedName) {
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, objKey, obj)
+				return apierr.IsNotFound(err)
+			}, time.Second*10, interval).Should(BeTrue(), "%T %s should not be created", obj, objKey.Name)
+		}
+
+		// Wait for any Kubernetes object to be not found.
+		expectResourceToBeDeleted := func(ctx context.Context, obj client.Object, objKey types.NamespacedName) {
```
brettmthompson, line 1106: This method is identical to the one above

pkg/controller/v1alpha1/inferencegraph/controller_test.go
```diff
@@ -1041,6 +1059,256 @@ var _ = Describe("Inference Graph controller test", func() {
...
+					if stopped_cond != nil && stopped_cond.Status == corev1.ConditionFalse {
+						return true
+					}
+				}
+				return false
+			}, timeout, interval).Should(BeTrue(), "The stopped condition should be set to false")
+		}
+
+		// Wait for the InferenceGraph's Stopped condition to be true.
+		expectIGTrueStoppedStatus := func(ctx context.Context, serviceKey types.NamespacedName) {
```
brettmthompson: Nit: this method could be made generic and reused for false condition as well.

pkg/controller/v1alpha1/inferencegraph/controller_test.go
```diff
@@ -1041,6 +1059,256 @@ var _ = Describe("Inference Graph controller test", func() {
...
+						Annotations: map[string]string{
+							"serving.kserve.io/deploymentMode": string(constants.Serverless),
+						},
+					},
+					Spec: v1alpha1.InferenceGraphSpec{
+						Nodes: map[string]v1alpha1.InferenceRouter{
+							v1alpha1.GraphRootNodeName: {
+								RouterType: v1alpha1.Sequence,
+								Steps: []v1alpha1.InferenceStep{
+									/*{
```
brettmthompson: Nit: remove comment

pkg/controller/v1alpha1/inferencegraph/controller_test.go
hdefazio: The only difference is a `Consistently` vs `Eventually` check... not sure if those can be used generically

pkg/controller/v1alpha1/inferencegraph/controller_test.go
```diff
@@ -1041,6 +1059,256 @@ var _ = Describe("Inference Graph controller test", func() {
...
+				},
+				Data: configs,
+			}
+			return configMap
+		}
+
+		// --- Reusable Check Functions ---
+		// Wait for the IG to exist.
+		expectIGToExist := func(ctx context.Context, serviceKey types.NamespacedName) v1alpha1.InferenceGraph {
+			// Check that the ISVC was updated
```
brettmthompson: the comment and variables related to 'update' here are misleading

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -257,6 +260,40 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	// Handle InferenceGraph status updates based on the force stop annotation.
+	// If true, transition the service to a stopped and unready state; otherwise, ensure it's not marked as stopped.
+	transition_time := apis.VolatileTime{Inner: metav1.Now()}
+	if forceStopRuntime {
+		// Exit early if we have already set the status to stopped
+		existingStoppedCondition := graph.Status.GetCondition(v1beta1.Stopped)
```
brettmthompson: Not related to the viability of this PR, but I think this stopped vs stopping logic should be reversed when the `Stopping` condition is added.
```
		if existingStoppingCondition != nil && existingStoppingCondition.Status == corev1.ConditionTrue {
			// check all resources associated with the inference service are stopped
                        // update status from stopping to stopped
		} else if existingStoppedCondition != nil && existingStoppedCondition.Status != corev1.ConditionTrue{
			// Add the stopping condition
			stoppingCondition := apis.Condition{
				LastTransitionTime: transition_time,
				Type:               v1beta1.Stopping,
				Status:             corev1.ConditionTrue,
			}
			readyCondition := apis.Condition{
				LastTransitionTime: transition_time,
				Type:               apis.ConditionReady,
				Status:             corev1.ConditionFalse,
				Reason:             v1beta1.StoppedISVCReason,
			}
			graph.Status.Conditions = []apis.Condition{stoppingCondition, readyCondition}

			graph.Status.URL = nil
		}
```

pkg/controller/v1alpha1/inferencegraph/controller_test.go
brettmthompson: Ah I missed that part :)

brettmthompson:
Changes LGTM

pkg/controller/v1alpha1/inferencegraph/controller.go
```diff
@@ -257,6 +260,43 @@ func (r *InferenceGraphReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	}
 
+	// Handle InferenceGraph status updates based on the force stop annotation.
+	// If true, transition the service to a stopped and unready state; otherwise, ensure it's not marked as stopped.
+	transition_time := apis.VolatileTime{Inner: metav1.Now()}
+	existingStoppedCondition := graph.Status.GetCondition(v1beta1.Stopped)
+	if forceStopRuntime {
+		// If the graph's stopped condition is not set or
+		// If the graph is currently running, update its status to signal that it should be stopped
+		if existingStoppedCondition == nil || (existingStoppedCondition != nil && existingStoppedCondition.Status == corev1.ConditionFalse) {
```
Jooho: The condition:
```
if existingStoppedCondition == nil || (existingStoppedCondition != nil && existingStoppedCondition.Status == corev1.ConditionFalse) {
```
contains a duplicated existingStoppedCondition != nil check.

Since the first clause (existingStoppedCondition == nil) already ensures safety, the second clause doesn’t need an explicit nil check.

It can be simplified to:
```
if existingStoppedCondition == nil || existingStoppedCondition.Status == corev1.ConditionFalse {
```
This improves readability without changing the logic.

pkg/controller/v1alpha1/inferencegraph/controller.go
Jooho: What we're trying to handle here is the transition where forceStopRuntime changes from true to false, right?
If that's the case, it makes more sense to update the status only when we are actually resuming from a previously stopped state.

It would be more appropriate to handle the case only when:
```
 else if existingStoppedCondition != nil && existingStoppedCondition.Status == corev1.ConditionTrue {
```

This way, we ensure the status update is meaningful and only occurs when we're truly transitioning out of the "Stopped" state.

pkg/controller/v1alpha1/inferencegraph/knative_reconciler.go
```diff
@@ -92,12 +92,28 @@ func (r *GraphKnativeServiceReconciler) Reconcile(ctx context.Context) (*knservi
...
 
+		if forceStopRuntime {
+			log.Info("Deleting inference graph knative service", "namespace", existing.Namespace, "name", existing.Name)
+			if existing.GetDeletionTimestamp() == nil { // check if the ksvc was already deleted
+				err := r.client.Delete(ctx, existing)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
```
Jooho, line 114: The current logic immediately returns even if the Knative service is still in the process of being deleted (DeletionTimestamp != nil).

This might lead to a premature exit, skipping further reconciliation until the next trigger. I think a better approach would be to requeue the request when deletion is in progress. Also, adding a log message when deletion is still pending would help with observability and debugging.

pkg/controller/v1alpha1/inferencegraph/controller.go
hdefazio: updated - the issue was I was trying to account for the stopped status not existing in the same block. I've separated it out so we set the stopped condition by default and then update it's value false <--> true as needed

Jooho (approved):
/lgtm

Issue #3419, opened 2024-02-08
Support OpenAI Schema for KServe LLM runtime


yuzisun:
/kind feature

**Describe the solution you'd like**
KServe recently added the support for [generate endpoint](https://github.com/kserve/open-inference-protocol/blob/main/specification/protocol/generate_rest.yaml) from open inference protocol and huggingface runtime API has already started to support the generate endpoint. KServe would like like to offer the OpenAI compatibility with generate endpoints. The benefits are that the API can be directly used with OpenAI's client libraries or third party tools like LangChain or LlamaIndex. It also can help people migrate from GPT4 to Mistral or Llama2. 

## Proposal:

KServe open inference protocol endpoints:
- v2/models/{model}/completion
- v2/models/{model}/chat/completion

OpenAI compatible endpoints:
- openai/v1/completion
- openai/v1/chat/completion

**Anything else you would like to add:**
References:
- OpenAI Schema: https://platform.openai.com/docs/api-reference/making-requests 
- OpenAI python types: https://github.com/openai/openai-python/tree/main/src/openai/types
- Azure OpenAI endpoints: https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#completions

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

johnugeorge:
+1 Must feature to provide seamless user experience. Ecosystem support is a huge value addition with OpenAI API compatibility

alexagriffith:
https://github.com/vllm-project/vllm/tree/main/vllm/entrypoints/openai

PR #4592, opened 2025-07-15
add openvino model server runtime

dtrawins:
Add OpenVINO Model Server runtime

**Type of changes**

Tested with inference service:
```
kind: InferenceService
metadata:
  name: openvino-model
spec:
  predictor:
    serviceAccountName: sa
    model:
      runtime: kserve-openvino
      modelFormat:
        name: openvino
      storageUri: "s3://ovms-models/model"
```
or
```
kind: InferenceService
metadata:
  name: qwen3-8b-int4-ov
spec:
  predictor:
    model:
      runtime: kserve-openvino
      modelFormat:
        name: huggingface
      args:
        - --source_model=OpenVINO/Qwen3-8B-int4-ov
        - --model_repository_path=/tmp
        - --task=text_generation
        - --target_device=CPU
```
Those parameters will skip storage initializer container and pull the model to /tmp (RW filesystem) by the serving container.

- [ ] Test A
- [ ] Test B

- Logs

New samples can be added in the PR if that would be recommended.

**Checklist**:

**Release note**:
```release-note
Included Serving Runtime configuration for [OpenVINO Model Server](https://github.com/openvinotoolkit/model_server). 
```

**Re-running failed tests**


config/runtimes/kserve-openvino.yaml
```diff
@@ -0,0 +1,58 @@
...
+      capabilities:
+        drop:
+        - ALL
+      privileged: false
+      runAsNonRoot: true
+      runAsUser: 5000
+  protocolVersions:
+  - v1
+  - v2
+  - v3
```
yuzisun: there is a v3 version ?

dtrawins: @yuzisun 
yes, openvino model server has 3 versions:
v1 with tensorflow serving compatible API
v2 with Kserve API 
v3 with generative API compatible with OpenAI and Cohear API

config/runtimes/kserve-openvino.yaml
```diff
@@ -0,0 +1,58 @@
...
+  annotations:
+    prometheus.kserve.io/path: /metrics
+    prometheus.kserve.io/port: "8080"
+  containers:
+  - args:
+    - --model_name={{.Name}}
+    - --model_path=/mnt/models
+    - --port=9000
+    - --rest_port=8080
+    - --file_system_poll_wait_seconds=0
```
gavrissh, line 20: Just curious what these parameters do
- diff between `rest_port` and `port`?
- Is this a mandatory one?`file_system_poll_wait_seconds`?

dtrawins: --port is setting grpc port
--rest_port is setting rest port as the name suggest
file_system_poll_wait_seconds is not mandatory but it disables checking if new model version got added to the model folder. In Kserve deployment, this content is static (copied during the initialization) so setting it to 0 is recommended. The default value is 1s wait time between checking for new version.

config/runtimes/kserve-openvino.yaml
yuzisun: It is a little confusing to call v3 though as we are separating predictive inference and generative inference.

Jooho: @yuzisun Yes, KServe uses v2 for the open inference protocol (https://github.com/kserve/kserve/issues/3419
), so calling it v3 could be confusing. Just to clarify, are you suggesting that we expose OpenVINO under v2 in KServe while internally using v3? Could you provide some guidance on how to approach this?

dtrawins: v3 version has been dropped to avoid confusion

juangon:
Hi @yuzisun @Jooho , do you have any estimate for reviewing and committing this? We think it's a great addition to kserve. Thanks!

config/runtimes/kserve-openvino.yaml
Jooho: @dtrawins I see another version now `grpc-v2`.  Kserve only has v1 and v2 so I think it would be a good idea dropping the `grpc-v2`

Jooho:
@juangon It needs a small fix. After it is updated and passes ci, then I will merge this pr.

config/runtimes/kserve-openvino.yaml
spolti: @Jooho the triton runtime uses grpc-v2, it might be ok to use it, wdyt?

Jooho: @dtrawins 
So just to confirm — does OpenVINO support REST v2 and gRPC v2? Since there are four constants defined([source](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go#L348-L352)), it seems fine to include gRPC v2 as well. My understanding is that OpenVINO supports KServe v1, v2, and gRPC-v2. Is that correct?

dtrawins: @Jooho @spolti  yes, that is correct. grpc_v2 is supported also in ovms. It is compatible with Triton server

config/runtimes/kserve-openvino.yaml
```diff
@@ -0,0 +1,57 @@
...
+      runAsUser: 5000
+  protocolVersions:
+  - v1
+  - v2
+  - grpc-v2
+  supportedModelFormats:
+  - autoSelect: true
+    name: openvino
+    priority: 1
+    version: "10"
```
Jooho, line 46: version 10 is right?

config/runtimes/kserve-openvino.yaml
```diff
@@ -0,0 +1,57 @@
...
+  annotations:
+    meta.helm.sh/release-name: kserve
+    meta.helm.sh/release-namespace: kserve
+  labels:
+    app.kubernetes.io/managed-by: Helm
+  name: kserve-openvino
+spec:
+  annotations:
+    prometheus.kserve.io/path: /metrics
+    prometheus.kserve.io/port: "8080"
```
Jooho, line 13: Could you please double-check the metrics port? 
looks 8000 is the one. 
https://docs.openvino.ai/2025/model-server/ovms_docs_metrics.html

config/runtimes/kserve-openvino.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: openvino
+    priority: 1
+    version: "10"
+  - autoSelect: true
+    name: onnx
+    priority: 2
+    version: "1"
+  - autoSelect: true
+    name: tensorflow
+    priority: 3
```
Jooho, line 53: `priority:3` is intentional? 

Priority defines the order among ClusterServingRuntimes for each modelFormat.
In other words, if multiple ClusterServingRuntimes share the same modelFormat,
the priority value determines which runtime will be selected first.

spolti:
@dtrawins maybe you can use this as example, https://github.com/opendatahub-io/odh-model-controller/blob/incubating/config/runtimes/ovms-kserve-template.yaml

We've been using it for a while and it seems to be very stable.

config/runtimes/kserve-openvino.yaml
dtrawins: @Jooho metrics port is the same like rest_port. Metrics and inference endpoints are exposed on the same port. Here, we set the --rest_port statically to 8080 so metrics are on the same.

config/runtimes/kserve-openvino.yaml
dtrawins: @Jooho  that was referring to OpenVINO model format version which is supported by the recent OVMS versions (staring from 2024) https://docs.openvino.ai/2025/documentation/openvino-ir-format.html The latest format is v11 which is also supported by the current versions 2025. Model format version shouldn't be older then v10. If that information is not useful for the deployments, it can be dropped from that yaml file.

config/runtimes/kserve-openvino.yaml
dtrawins: There were already created instances of ClusterServingRuntimes assigned with priorities 1 and 2 for Tensorlow. It didn't seam right to overwrite old settings. I guess users could still set explicitly the serving runtime regardless of the default priority.

juangon:
> @juangon It needs a small fix. After it is updated and passes ci, then I will merge this pr.

Thanks @Jooho ! Looks like all had been resolved.

juangon:
Hi @Jooho , can something be additionally done so this can be merged? Thanks very much!

Jooho (approved):
Thanks you for the contribution.

/lgtm

PR #4593, opened 2025-07-15
Fixed the issue of the same metrics across different deployments under different namespaces

### LINKED ISSUE #4542: [Autoscaling KEDA] Differentiate the same metrics across different deployments under different namespaces

houshengbo:
## What this PR does

- Fixes an issue where the OpenTelemetryCollector and KEDA ScaledObject did not properly isolate metrics by deployment and namespace.
- Now, the OTelCollector always injects `namespace` and `deployment` labels into metrics using the `resourcedetection/env` and `transform` processors.
- The KEDA ScaledObject’s PromQL query for pod metrics now always filters by `namespace` and `deployment`, ensuring correct autoscaling per workload.
- Unit tests for both reconcilers have been updated and now pass.

## Why is this needed?

- Without these changes, metrics from different deployments/namespaces could be mixed, causing incorrect autoscaling behavior.
- This resolves [issue #4542](https://github.com/kserve/kserve/issues/4542).

## How was this tested?

- All relevant unit tests were updated and pass (`go test ./pkg/controller/v1beta1/inferenceservice/reconcilers/...`).

## User impact

- No user-facing API changes.
- Autoscaling now works as expected when multiple deployments use the same metric name.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
@@ -67,7 +67,7 @@ func NewKedaReconciler(client client.Client,
 	}, nil
 }
 
-func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec, configMap *corev1.ConfigMap,
+func getKedaMetrics(componentMeta metav1.ObjectMeta, componentExt *v1beta1.ComponentExtensionSpec, configMap *corev1.ConfigMap,
 ) ([]kedav1alpha1.ScaleTriggers, error) {
 	var triggers []kedav1alpha1.ScaleTriggers
 
@@ -156,8 +156,12 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec, configMap *cor
 
 				if triggerType == string(constants.AutoScalerMetricsSourceOpenTelemetry) {
 					trigger.Type = "external"
+					// Inject namespace and deployment label selectors into the query for metric isolation.
+					// This ensures the metricQuery only selects metrics for the correct deployment and namespace.
+					// Example: sum(<query>{namespace="<namespace>", deployment="<deployment>"})
+					metricQuery := fmt.Sprintf("sum(%s{namespace=\"%s\", deployment=\"%s\"})", query, componentMeta.Namespace, componentMeta.Name)
 					trigger.Metadata = map[string]string{
-						"metricQuery":   query,
+						"metricQuery":   metricQuery,
 						"targetValue":   fmt.Sprintf("%f", targetValue),
 						"scalerAddress": MetricScalerEndpoint,
 					}
@@ -189,7 +193,7 @@ func createKedaScaledObject(componentMeta metav1.ObjectMeta,
 	if MaxReplicas < *MinReplicas {
 		MaxReplicas = *MinReplicas
 	}
-	triggers, err := getKedaMetrics(componentExtension, configMap)
+	triggers, err := getKedaMetrics(componentMeta, componentExtension, configMap)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
@@ -68,6 +68,41 @@ func createOtelCollector(componentMeta metav1.ObjectMeta,
 		port = "8080"
 	}
 
+	processors := map[string]interface{}{
+		"resourcedetection/env": map[string]interface{}{
+			"detectors": []interface{}{"env"},
+			"timeout":   "2s",
+			"override":  false,
+		},
+		"transform": map[string]interface{}{
+			"metric_statements": []interface{}{
+				map[string]interface{}{
+					"context": "datapoint",
+					"statements": []interface{}{
+						"set(attributes[\"namespace\"], resource.attributes[\"k8s.namespace.name\"])",
+						"set(attributes[\"deployment\"], resource.attributes[\"k8s.deployment.name\"])",
+						"set(attributes[\"pod\"], resource.attributes[\"k8s.pod.name\"])",
+					},
+				},
+			},
+		},
+	}
+
+	pipelineProcessors := []string{"resourcedetection/env", "transform"}
+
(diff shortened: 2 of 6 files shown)


houshengbo:
@andyi2it @yuzisun 
I have fixed the issue of making the metrics unique by adding the ns, deployment name and pod name to it.

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -156,8 +156,12 @@ func getKedaMetrics(componentExt *v1beta1.ComponentExtensionSpec, configMap *cor
 
 				if triggerType == string(constants.AutoScalerMetricsSourceOpenTelemetry) {
 					trigger.Type = "external"
+					// Inject namespace and deployment label selectors into the query for metric isolation.
+					// This ensures the metricQuery only selects metrics for the correct deployment and namespace.
+					// Example: sum(<query>{namespace="<namespace>", deployment="<deployment>"})
+					metricQuery := fmt.Sprintf("sum(%s{namespace=\"%s\", deployment=\"%s\"})", query, componentMeta.Namespace, componentMeta.Name)
```
yuzisun, line 162: The query itself can include the sum and labels already? not sure if we can inject this way and still ensure a valid query.

pkg/controller/v1beta1/inferenceservice/reconcilers/otel/otel_reconciler.go
```diff
@@ -35,7 +35,54 @@ import (
...
+	ProcessorFilterMetrics        = "filter/metrics"
+	JobNameOtelCollector          = "otel-collector"
+	PrometheusReceiver            = "prometheus"
+	OtlpExporter                  = "otlp"
+	ModeSidecar                   = "sidecar"
+
+	AnnotationPrometheusPort = "prometheus.kserve.io/port"
+	DefaultPrometheusPort    = "8080"
+
+	ResourcedetectionDetectorEnv = "env"
```
yuzisun, line 50:
```suggestion
	ResourceDetectionEnv = "env"
```

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
houshengbo: Yes, that's exactly the metricQuery I used for keda.

yuzisun: I mean the query field is exposed on the YAML, if users themselves specify the query with sum and namespace/deployment tag then kserve shouldn’t inject again ?

houshengbo: OK, it is possible that users specify only metrics with this parameter, or specify a full query with this parameter. How can we tell the differences?

yuzisun: on Autoscaling spec there are both `metricNames` and `query` field
https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/component.go#L305. I think if the user only specifies the metricsNames then we can default the query with your logic. If query field is specified we do not change it.

andyi2it:
@houshengbo 
Wouldn't the change here be enough?
https://github.com/kserve/kserve/pull/4569

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
houshengbo: It seems this `query` was taken as the metric in either the test and the example.
I will keep the `query` the same as before, but for metricNames, I will add the labels of deployment name and namespace.

yuzisun:
> @houshengbo Wouldn't the change here be enough? #4569

@andyi2it I think relabel might not work as we do not know the pod name before hand.

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
yuzisun: will create a separate PR to fix this

yuzisun (approved):

houshengbo:
@andyi2it I am not familiar with how "relabel_configs" works.
Do you have an example for what query will be used, if we add deployment name and namespace in "relabel_configs"? How to query certain metrics based on deployment name, namespace and the metric itself?

yuzisun:
> @andyi2it I am not familiar with how "relabel_configs" works. Do you have an example for what query will be used, if we add deployment name and namespace in "relabel_configs"? How to query certain metrics based on deployment name, namespace and the metric itself?

I think the issue is that we also need the pod label, relabel is more static and you need to know the value beforehand.

PR #4604, opened 2025-07-22
fix(deps): adds missing LLMInferenceService types for python imports

bartoszmajsak:

Adds missing LLMInferenceService types for python imports.

**Type of changes**

**Release note**:
```release-note
NONE
```

**Re-running failed tests**


pierDipi (approved):
Thanks!

PR #4605, opened 2025-07-22
chore(test-setup): replaces gatewayclass in-memory

bartoszmajsak:

This way we do not overwrite versioned file and risk pushing it by mistake.

**Type of changes**

**Release note**:
```release-note
NONE
```

**Re-running failed tests**


pierDipi (approved):
/lgtm

yuzisun (approved):

PR #4596, opened 2025-07-16
update CRDs for LWS based multi node support

VedantMahabaleshwarkar:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Follow up to previous LLM-D contributions. 
This PR includes the CRD changes to LLM InferenceService to add support for LWS based multi node 

**Type of changes**
Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [x] Have you made corresponding changes to the documentation?

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

PR #4614, opened 2025-07-24
Adding Llm-d scheduler reconcilier

andresllh:
Depends on: https://github.com/kserve/kserve/pull/4596
The llm-d scheduler allows for smart load balancing for llm workloads.

**Type of changes**

`make test`

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+
+	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Scheduler != nil && llmSvc.Spec.Router.Scheduler.Template != nil {
+		d.Spec.Template.Spec = *llmSvc.Spec.Router.Scheduler.Template.DeepCopy()
+		for i := range d.Spec.Template.Spec.Containers {
+			if d.Spec.Template.Spec.Containers[i].Name != "main" {
+				continue
+			}
+
+			if slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "--configText") ||
```
gavrissh, line 296: Here we would need a null check on `.Args`?

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+		for name, port := range actualPorts {
+			servicePorts = append(servicePorts, corev1.ServicePort{
+				Name:       name,
+				Port:       port.ContainerPort,
+				TargetPort: intstr.FromString(name),
+				Protocol:   port.Protocol,
+			})
+		}
+
+		sort.Slice(servicePorts, func(i, j int) bool {
```
gavrissh, line 201: For clarity, is sorting mandatory?

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: r.schedulerLabels(llmSvc),
+		},
+	}
+
+	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Scheduler != nil && llmSvc.Spec.Router.Scheduler.Template != nil {
```
gavrissh: used in multiple places, should we consider a helper?

```
func (r *LLMISVCReconciler) hasScheduler(llmSvc *v1alpha1.LLMInferenceService) bool {
    return llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Scheduler != nil
}
```

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+	log.FromContext(ctx).V(2).Info("Expected router scheduler deployment", "deployment", d)
+
+	return d
+}
+
+func schedulerConfigText(llmSvc *v1alpha1.LLMInferenceService) string {
+	switch {
+	case llmSvc.Spec.Prefill != nil:
+		return `
+apiVersion: inference.networking.x-k8s.io/v1alpha1
```
yuzisun, line 320: Is this a global config? can we read the resource instead of harding coding here

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/pkg/kmeta"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	igwapi "sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+)
+
+func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) error {
```
yuzisun, line 43: please help add the test files for the scheduler

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
```diff
@@ -18,8 +18,16 @@ package v1alpha1
 
 import (
 	"k8s.io/utils/ptr"
+	"knative.dev/pkg/kmeta"
 )
 
+func (r *RouterSpec) EPPServiceName(llmSvc *LLMInferenceService) string {
```
yuzisun, line 40: this is an internal function which we do not need to export

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
```diff
@@ -18,8 +18,16 @@ package v1alpha1
...
+	if r == nil || r.Route == nil || r.Scheduler == nil || r.Scheduler.Pool == nil || !r.Scheduler.Pool.HasRef() || r.Scheduler.Pool.Spec == nil || r.Scheduler.Pool.Spec.ExtensionRef == nil {
```
yuzisun, line 41: Can you add comment here why there are so many checks

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
```diff
@@ -18,8 +18,16 @@ package v1alpha1
 
 import (
 	"k8s.io/utils/ptr"
+	"knative.dev/pkg/kmeta"
```
johnugeorge, line 21: Not a blocker: But do you need knative import for this util function?  We will have more flexibility in future if we avoid any knative dependency in new LLM CRD code.

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
johnugeorge: +1 
Have a higher level function to check this. Not all checks can be nil independently

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-inference-pool"),
+			Namespace: llmSvc.GetNamespace(),
+			Labels:    labels,
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+		},
+	}
+	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Scheduler != nil && llmSvc.Spec.Router.Scheduler.Pool != nil && llmSvc.Spec.Router.Scheduler.Pool.Spec != nil {
```
johnugeorge, line 226: Same comment

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+		},
+		Spec: igwapi.InferenceModelSpec{
+			ModelName: ptr.Deref(llmSvc.Spec.Model.Name, llmSvc.GetName()),
+			PoolRef: igwapi.PoolObjectReference{
+				Group: "inference.networking.x-k8s.io",
+				Kind:  "InferencePool",
+				Name:  igwapi.ObjectName(kmeta.ChildName(llmSvc.GetName(), "-inference-pool")),
```
johnugeorge: better to centralise naming pattern for all  names which are referenced in multiple functions. Same for -inference-model

pkg/controller/v1alpha1/llmisvc/scheduler.go
johnugeorge: EndpointPickerConfig can be customized as per requirements. eg: You can give different weights to different scorers.  Why is it hardcoded ?

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-epp-role"),
+			Namespace: llmSvc.GetNamespace(),
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+			Labels: r.schedulerLabels(llmSvc),
+		},
+		Rules: []rbacv1.PolicyRule{
```
johnugeorge, line 403: F

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: I don't believe it is. Can you confirm @pierDipi ?

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: slices.Contains already handles nil checks gracefully, it will just return false

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: This is the default config. If it is not overwritten by the user, then this is what will be used.

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
andresllh: This file is part of the v1alpha1 package, while scheduler.go is part of the llmisvc package.

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
andresllh: @pierDipi @bartoszmajsak @Jooho

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/scheduler.go
johnugeorge: Can't we provide via ConfigMap?

andresllh: @pierDipi @bartoszmajsak any objection to moving this into a configmap?

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
andresllh: The only thing used from that package is the ChildName() function.

pkg/controller/v1alpha1/llmisvc/scheduler.go
pierDipi: I think it makes sense to have it injected, however, I'd like to us to make progress and track improvements like this one as follow up.

andresllh: https://github.com/kserve/kserve/issues/4628

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: will be added as a follow up https://github.com/kserve/kserve/issues/4629

pkg/apis/serving/v1alpha1/llm_inference_service_types_func.go
pierDipi: Essentially what this is trying to do is to figure out whether the EPPServiceName is overriden in the `LLMInferenceService`, all this checks ensure we use the "default EPP Service" if it's not overriden 

To override it, users can specify `r.Scheduler.Pool.Spec.ExtensionRef.Name` and all nested fields can be `nil`

pkg/controller/v1alpha1/llmisvc/scheduler.go
pierDipi: cc @bartoszmajsak

pierDipi: probably the idea is to reduce unnecessary updates, but I'll let Bartosz confirm

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -0,0 +1,61 @@
...
+		if cond.Type == appsv1.DeploymentAvailable {
+			if cond.Status == corev1.ConditionTrue {
+				ready()
+			} else {
+				notReady(cond.Reason, cond.Message)
+			}
+			return nil
+		}
+	}
+	logger.Info("Deployment processed")
```
yuzisun: remove the unnecessary debug logging

pkg/controller/v1alpha1/llmisvc/scheduler.go
yuzisun: I think some of these validations can be done at the crd level using CEL expressions. Pool should be mandatory if the scheduler is specified.

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+		sa.Name = llmSvc.Spec.Router.Scheduler.Template.ServiceAccountName
+	}
+
+	return sa
+}
+
+func (r *LLMISVCReconciler) expectedSchedulerRole(llmSvc *v1alpha1.LLMInferenceService) *rbacv1.Role {
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-epp-role"),
```
yuzisun, line 396: Why do we need to create a role per inference service? looks like it is the same set of rules.

pkg/controller/v1alpha1/llmisvc/scheduler.go
pierDipi: This function is also used to handle transitions from "scheduler enabled" to "scheduler disabled or external reference", where we need to delete the InferencePool object (see reconcileSchedulerInferencePool) and in that case any of this fields can be `nil` and we need to guard it

pierDipi: One more thought, the validation at the webhook or CEL level is generally tricky to do when also considering the presets, as those are external resources that can be created out of order, so we might not have the "full spec" until the reconciliation happen (see config_merge, in particular, the `combine...` function).

While it's technically still prevented for some `nil` fields with `combine..` logic, I'd still guard against potential panics, as a panic error means that a user will need to look into the logs (and they might not have access) to debug potential bugs (on our side) or mis-configuration

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
yuzisun:
```suggestion
```

yuzisun (dismissed):

yuzisun (approved):

pkg/controller/v1alpha1/llmisvc/scheduler.go
bartoszmajsak: Correct, my thinking here was that we should make it stable.

PR #4616, opened 2025-07-24
llm-d workload reconciliation

brettmthompson:
This PR creates the single and multi node workload reconcilers for LLMInferenceServices. All required dependencies for those reconcilers and tests are also included in this PR.

The core controller will be finalized through a separate PR.

This PR should land after https://github.com/kserve/kserve/pull/4614 since these reconcilers depend on the llmisvc scheduler.

**Type of changes**

- Finalized unit tests and e2e tests for the llmisvc feature will land through a separate PR.

**Checklist**:

**Release note**:

```release-note
llm-d workload reconciliation
```

**Re-running failed tests**


pkg/controller/v1alpha1/llmisvc/workload_single_node.go
pierDipi, line 1: @johnugeorge @yuzisun @sivanantha321 This is the core single node reconciliation

charts/llmisvc-resources/templates/clusterrole.yaml
```diff
@@ -4,6 +4,18 @@ kind: ClusterRole
 metadata:
   name: llmisvc-manager-role
 rules:
+- apiGroups:
+  - leaderworkerset.x-k8s.io
```
gavrissh, line 8: Is the `leaderworkset` change needed for the single-node approach?

gavrissh: Oh I see there are changes for both.

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -0,0 +1,191 @@
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+			},
+		},
+	}
+
+	if llmSvc.Spec.Template != nil {
```
gavrissh: Do we need this nil check again? as we already return an error if nil on the start

pkg/constants/constants.go
```diff
@@ -281,6 +297,7 @@ const (
 	CustomSpecMultiModelServerEnvVarKey               = "MULTI_MODEL_SERVER"
 	KServeContainerPrometheusMetricsPortEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_METRICS_PORT"
 	KServeContainerPrometheusMetricsPathEnvVarKey     = "KSERVE_CONTAINER_PROMETHEUS_METRICS_PATH"
+	ModelInitModeEnv                                  = "MODEL_INIT_MODE"
```
gavrissh: nit: ModelInitModeEnvKey?
Similar to other vars

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -0,0 +1,476 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+			},
+		}
+	}
+
+	if llmSvc.Spec.Prefill != nil && llmSvc.Spec.Prefill.Template != nil {
+		d.Spec.Template.Spec = *llmSvc.Spec.Prefill.Template.DeepCopy()
+	}
```
gavrissh:
```suggestion
	if llmSvc.Spec.Prefill != nil {
		d.Spec = appsv1.DeploymentSpec{
			Replicas: llmSvc.Spec.Prefill.Replicas,
			Selector: &metav1.LabelSelector{
				MatchLabels: labels,
			},
			Template: corev1.PodTemplateSpec{
				ObjectMeta: metav1.ObjectMeta{
					Labels: labels,
				},
			},
		}
		
		if  llmSvc.Spec.Prefill.Template != nil {
		        d.Spec.Template.Spec = *llmSvc.Spec.Prefill.Template.DeepCopy()
	        }
	}
```

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -0,0 +1,375 @@
...
+		if llmSvc.Spec.Prefill.Template != nil {
+			expected.Spec.LeaderWorkerTemplate.LeaderTemplate = &corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: leaderLabels,
+				},
+				Spec: *llmSvc.Spec.Prefill.Template.DeepCopy(),
+			}
+		}
+		if llmSvc.Spec.Prefill.Worker != nil {
+			expected.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec = *llmSvc.Spec.Worker.DeepCopy()
```
gavrissh:
```suggestion
			expected.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec = *llmSvc.Spec.Prefill.Worker.DeepCopy()
```

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -0,0 +1,375 @@
...
+			RolloutStrategy: lwsapi.RolloutStrategy{
+				Type: lwsapi.RollingUpdateStrategyType,
+			},
+			StartupPolicy: lwsapi.LeaderReadyStartupPolicy,
+		},
+	}
+
+	if llmSvc.Spec.Prefill != nil {
+		expected.Spec.Replicas = llmSvc.Spec.Prefill.Replicas
+		expected.Spec.LeaderWorkerTemplate.Size = llmSvc.Spec.Prefill.Parallelism.GetSize()
```
gavrissh, line 242: Can `Parallelism` be nil? do we need validation for that?

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,94 @@
+/*
+Copyright 2023 The KServe Authors.
```
gavrissh: nit: 2025 in some files

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return s
+}
+
+func hasRoutingSidecar(pod corev1.PodSpec) bool {
+	return routingSidecar(&pod) != nil
+}
+
+func routingSidecar(pod *corev1.PodSpec) *corev1.Container {
+	for i := range pod.InitContainers {
```
gavrissh: Nil check for the `pod` would be required?

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,94 @@
...
+	// TODO https://github.com/llm-d/llm-d-inference-scheduler/issues/220 and DP template
+
+	return s
+}
+
+func hasRoutingSidecar(pod corev1.PodSpec) bool {
+	return routingSidecar(&pod) != nil
+}
+
+func routingSidecar(pod *corev1.PodSpec) *corev1.Container {
```
gavrissh, line 77: Misleading function name: routingSidecar searches in InitContainers, but the name suggests it's looking for a sidecar container. Sidecars are typically regular containers, not init containers. Consider renaming to routingInitContainer or adding clarifying documentation..

pierDipi: Historically that's the case, however, now the official sidecars definition in k8s are initContainers with a RestartPolicy (in our case will be "Always") https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/ (it's a relatively new feature)

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
pierDipi: Yes, it can, however:
- GetSize handles `Parallelism = nil`
- webhook validation is coming in a separate PR and takes into account fields in `Parallelism`

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
pierDipi: Good catch!

pierDipi:
Thanks for the review @gavrissh!

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	kserveTypes "github.com/kserve/kserve/pkg/types"
+)
+
+const (
+	routingSidecarContainerName = "llm-d-routing-sidecar"
+)
+
+var sidecarSSRFProtectionRules = []rbacv1.PolicyRule{
```
johnugeorge, line 35: Fix the format

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -0,0 +1,375 @@
...
+	if err := r.reconcileMultiNodePrefillWorkload(ctx, llmSvc); err != nil {
+		return fmt.Errorf("failed to reconcile multi-node prefill workload: %w", err)
+	}
+	return nil
+}
+
+func (r *LLMInferenceServiceReconciler) reconcileMultiNodeMainWorkload(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) error {
+	expected := r.expectedMainMultiNodeLWS(ctx, llmSvc)
+	if llmSvc.Spec.Worker == nil {
+		if err := Delete(ctx, r, llmSvc, expected); err != nil {
```
johnugeorge, line 55: Is this tested? Delete assume that llmSvc is already present

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,94 @@
...
+	defer llmSvc.DetermineWorkloadReadiness()
+
+	if err := r.reconcileSelfSignedCertsSecret(ctx, llmSvc); err != nil {
+		return fmt.Errorf("failed to reconcile self-signed certificates secret: %w", err)
+	}
+
+	// We need to always reconcile every type of workload to handle transitions from P/D to another topology (meaning
+	// finalizing superfluous workloads).
+
+	if err := r.reconcileMultiNodeWorkload(ctx, llmSvc); err != nil {
```
johnugeorge, line 57: We should do multi node reconcile only if enabled in spec.  Please ensure that LWS is not required in single node workload

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
pierDipi: Delete internally handles `NotFound` and it will not raise errors when the resource is not found.

pkg/controller/v1alpha1/llmisvc/workload.go
brettmthompson: @pierDipi @bartoszmajsak @Jooho 
Can you please help to confirm?

pkg/controller/v1alpha1/llmisvc/workload.go
brettmthompson: @johnugeorge thank you for the review! Can you confirm what formatting fix you are looking for here?

pkg/controller/v1alpha1/llmisvc/workload.go
brettmthompson: added nil check

pkg/controller/v1alpha1/llmisvc/workload.go
brettmthompson: updated all new files to 2025

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
brettmthompson: applied in latest commit

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: applied in latest commit

pkg/constants/constants.go
brettmthompson: update to `ModelInitModeEnvKey`

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: removed redundant nil check

pkg/controller/v1alpha1/llmisvc/workload.go
pierDipi: Agree, the intention is to not have LWS required.

we still need to reconcile multi node to handle transitions from multi -> single (or the other way around) and clean up resources, however, generic CRUD functions are handling `IsNotFound` and `IsNoMatchError` (IsNoMatchError == LWS is not installed) when multi node is not used [1], so we should be good ?

[1]
```golang
	if err := c.Delete(ctx, expected); err != nil {
		if !apierrors.IsNotFound(err) && !meta.IsNoMatchError(err) {
			return fmt.Errorf("failed to delete %s %s/%s: %w", typeLogLine, expected.GetNamespace(), expected.GetName(), err)
		}
		return nil
	}
```

pierDipi (approved):
This is a good start, we can iterate as we go for improvements

/lgtm

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -61,6 +61,19 @@ func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1al
 	return nil
 }
 
+func (r *LLMISVCReconciler) reconcileSchedulerAuthDelegatorBinding(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService, sa *corev1.ServiceAccount) error {
```
yuzisun, line 66: Can you help add comment on what does this auth delegator rbac do ?

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -0,0 +1,375 @@
...
+func (r *LLMISVCReconciler) expectedMainMultiNodeLWS(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) *lwsapi.LeaderWorkerSet {
+	workerLabels := map[string]string{
+		"app.kubernetes.io/component": "llminferenceservice-workload-worker",
+		"app.kubernetes.io/name":      llmSvc.GetName(),
+		"app.kubernetes.io/part-of":   "llminferenceservice",
+	}
+	if llmSvc.Spec.Template == nil {
+		// When there is no leader template, workers become part of the InferencePool selector.
+		workerLabels["kserve.io/component"] = "workload"
+		workerLabels["llm-d.ai/role"] = "decode"
```
yuzisun, line 114: should we use the same group here as this is in kserve

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -0,0 +1,375 @@
...
+
+	if llmSvc.Spec.Template != nil {
+		expected.Spec.LeaderWorkerTemplate.LeaderTemplate = &corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: leaderLabels,
+			},
+			Spec: *llmSvc.Spec.Template.DeepCopy(),
+		}
+
+		if hasRoutingSidecar(expected.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec) {
```
yuzisun, line 159: is the routing sidecar required by LWS?

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -18,17 +18,176 @@ package llmisvc
...
+		role = "both"
+	}
+
+	labels := r.singleNodeLabels(llmSvc)
+	labels["kserve.io/component"] = "workload"
+	labels["llm-d.ai/role"] = role
+
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-kserve"),
```
yuzisun, line 86: do we need to append kserve here?

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -18,17 +18,176 @@ package llmisvc
...
+			Namespace: llmSvc.GetNamespace(),
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+			Labels: labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: llmSvc.Spec.Replicas,
```
yuzisun, line 94: We only support static replica for singe node deployment?

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -18,17 +18,176 @@ package llmisvc
...
+		"app.kubernetes.io/component": "llminferenceservice-workload-prefill",
+		"app.kubernetes.io/name":      llmSvc.GetName(),
+		"app.kubernetes.io/part-of":   "llminferenceservice",
+		"kserve.io/component":         "workload",
+		"llm-d.ai/role":               "prefill",
+	}
+
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-kserve-prefill"),
```
yuzisun, line 156: I think the kserve name here is redundant

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const (
+	certificateDuration                      = time.Hour * 24 * 365 * 10 // 10 years
+	certificateExpirationRenewBufferDuration = certificateDuration / 5
+
+	certificatesExpirationAnnotation = "certificates.kserve.io/expiration"
+)
+
+func (r *LLMISVCReconciler) reconcileSelfSignedCertsSecret(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) error {
```
yuzisun, line 50: Can you add comments on why we need this self signed certificate?

pkg/controller/v1alpha1/llmisvc/workload.go
yuzisun: Would appreciate if there are comments about the sidecar role here, i thought the routing can be handled by the inference gateway.

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -18,17 +18,176 @@ package llmisvc
...
+	}
+	return nil
+}
+
+func (r *LLMISVCReconciler) reconcileSingleNodeMainWorkload(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService, storageConfig *types.StorageInitializerConfig) error {
+	expected, err := r.expectedSingleNodeMainDeployment(ctx, llmSvc, storageConfig)
+	if err != nil {
+		return fmt.Errorf("failed to get expected main deployment: %w", err)
+	}
+	if llmSvc.Spec.Worker != nil {
```
yuzisun, line 61: why do we delete if the worker is not nil?

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -18,17 +18,176 @@ package llmisvc
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+			},
+		}
+
+		if llmSvc.Spec.Prefill.Template != nil {
+			d.Spec.Template.Spec = *llmSvc.Spec.Prefill.Template.DeepCopy()
```
yuzisun, line 179: Do we expect user to provide the full deployment template? I thought we have the preset configurations

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -46,6 +205,7 @@ func (r *LLMISVCReconciler) propagateDeploymentStatus(ctx context.Context, expec
 			return nil
 		}
 	}
+	logger.Info("Deployment processed")
```
yuzisun: remove the debugging log

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -55,3 +215,120 @@ func semanticDeploymentIsEqual(expected *appsv1.Deployment, curr *appsv1.Deploym
...
+		},
+	}
+	ro.Rules = append(ro.Rules, sidecarSSRFProtectionRules...)
+	return ro
+}
+
+func (r *LLMISVCReconciler) expectedSingleNodeRoleBinding(llmSvc *v1alpha1.LLMInferenceService, sa *corev1.ServiceAccount) *rbacv1.RoleBinding {
+	return &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      kmeta.ChildName(llmSvc.GetName(), "-kserve-rb"),
```
yuzisun, line 305: same here, I think we can remove the kserve name to keep the resource name short

pkg/utils/storage.go
```diff
@@ -0,0 +1,387 @@
...
+//
+// Parameters:
+//   - modelUri: The URI specifying the PVC and optional sub-path to mount.
+//   - containerName: The name of the container within the PodSpec to which the PVC should be mounted.
+//   - readOnly: Whether the mount should be read-only.
+//   - podSpec: PodSpec to modify.
+//
+// Returns:
+//   - error: An error if the modelUri is invalid or if any other issue occurs; otherwise, nil.
+func AddModelPvcMount(modelUri, containerName string, readOnly bool, podSpec *corev1.PodSpec) error {
```
yuzisun, line 79: Do we have the test coverage for these functions? it is fairly large refactoring

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
pierDipi: This is to reduce potential conflicts

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
pierDipi: This is required by llm-d scheduler until this is fixed https://github.com/llm-d/llm-d-inference-scheduler/issues/220

pkg/utils/storage.go
brettmthompson: If acceptable test coverage for the new storage utility functions will be added as a followup - https://github.com/kserve/kserve/issues/4639

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: If worker is not nil than this will fall under a multi node workload and should be handled by the multi node workload reconciler.

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: Currently only static replica is supported for single node deployment. We do not support a range like for standard inference services, i.e. utilizing `min` and `max` replicas.

@pierDipi do we plan to support autoscaling for single node workloads in the future?

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: This is to reduce potential conflicts and keep consistency with the naming convention of other created resources

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: This is to reduce potential conflicts and keep consistency with the naming convention of other created resources

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
brettmthompson: While not strictly required by LWS, we are using a routing sidecar here for enhancement of the core functionality.

pierDipi: The routing sidecar is used in llm-d for doing disaggregated serving (separate prefill and decode instances) is also used in the single node deployment topology when P/D is enabled, so generally the sidecar is not LWS specific, is more about P/D specific

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
pierDipi: for llmisvc, autoscaling API is not defined yet, once we have the initial implementation we will need to follow up wrt autoscaling

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
pierDipi: Yes, if worker is not nil, we will need to create an LWS (and remove the single node deployment if there is one)

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: Preset configurations will be added if an empty (`{}`) deployment template is provided by the user. Well known presets will be applied by the logic in `config_merge.go` file as the first step of reconciliation. 

https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/llmisvc/config_merge.go#L104

pierDipi: in the top-level case `spec.template`, that is the fallback case so even when `template = nil` and no other case is entered in the well-known presets merge logic:

https://github.com/kserve/kserve/blob/64d775bb99a4d7f24837fe01a4c5f0a10853de28/pkg/controller/v1alpha1/llmisvc/config_merge.go#L121-L122

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
brettmthompson: removed in latest commit

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed.go
brettmthompson: comment added in latest commit

pkg/controller/v1alpha1/llmisvc/scheduler.go
brettmthompson: comment added in latest commit

brettmthompson:
> Would appreciate if there are comments about the sidecar role here, i thought the routing can be handled by the inference gateway.

@yuzisun added comment on `routingSidecar` method to describe the role of the sidecar.

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
yuzisun: Since this is in the scope of KServe LLMInferenceService, I think this suffix is unnecessary but I see there is potential conflict with other k8s deployments in the cluster.

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
yuzisun: I thought the P/D routing can be done at the inference gateway side, is the sidecar required for P/D?

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed.go
yuzisun: Could we make this optional? I am not sure if this is the proper way to handle TLS workload

yuzisun: If you are intended to run service mesh mode then TLS should be handled by istio ?

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
brettmthompson: Routing sidecar is used in the initial architecture for llm-d 0.2, which is why we are using this pattern for now. This will be changed in the future as the architecture for llm-d moves forward.

pkg/controller/v1alpha1/llmisvc/workload.go
Jooho: @yuzisun The routing sidecar acts as a proxy between the prefill and decode components. The scheduler selects the best decode pod, and the routing sidecar decides whether to send the request to the prefill or not.

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed.go
brettmthompson: cert-manager is an option, but for now we have made an internal decision to use self signed certs here for the purpose of moving fast

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
Jooho: It is currently required for P/D. There are plans to remove the sidecar in the future, but for the targeted LLM-D 0.2 release, it is still needed.

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed.go
Jooho: >Could we make this optional? I am not sure if this is the proper way to handle TLS workload

I believe encrypting communication between each component is a good practice. The parts where requests are encrypted using TLS are as follows (note that encryption between vLLM instances is currently not supported):
```
gateway -> scheduler
scheduler -> routing sidecar
routing sidecar -> vLLM
```
>If you are intended to run service mesh mode then TLS should be handled by istio ?

Among these, the service-based access happens only from gateway to scheduler. Therefore, when accessing this part, TLS should be used and explicitly indicated. For this, TLS needs to be enabled using the GW API’s BackendTLSPolicy, but currently, we are using DestinationRule instead. This will be updated continuously in the future.

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func hasRoutingSidecar(pod corev1.PodSpec) bool {
+	return routingSidecar(&pod) != nil
+}
+
+// routingSidecar returns the routing sidecar container within a pod if present.
+// The routing sidecar is used for disaggregated serving (separate prefill and decode instances) for multi node workloads
+// and is used in the single node deployment topology when prefill/decode is enabled for single node workloads.
+func routingSidecar(pod *corev1.PodSpec) *corev1.Container {
```
sivanantha321, line 77: maybe rename it getRoutingSidecar ?

pkg/controller/v1alpha1/llmisvc/workload_storage.go
```diff
@@ -0,0 +1,167 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/credentials/s3"
+	"github.com/kserve/kserve/pkg/types"
+	"github.com/kserve/kserve/pkg/utils"
+)
+
+// attachModelArtifacts configures a PodSpec to fetch and use a model froma provided URI in the LLMInferenceService.
```
sivanantha321:
```suggestion
// attachModelArtifacts configures a PodSpec to fetch and use a model from a provided URI in the LLMInferenceService.
```

pkg/credentials/service_account_credentials.go
```diff
@@ -235,7 +221,7 @@ func (c *CredentialBuilder) CreateSecretVolumeAndEnv(namespace string, annotatio
 		if secretName, ok := annotations[c.config.StorageSecretNameAnnotation]; ok {
 			err := c.mountSecretCredential(secretName, namespace, container, volumes)
 			if err != nil {
-				log.Error(err, "Failed to mount the secret credentials", "secretName", secretName)
+				log.Error(err, "Failed to amount the secret credentials", "secretName", secretName)
```
sivanantha321, line 224: typo ?

pkg/credentials/service_account_credentials.go
```diff
@@ -97,19 +97,10 @@ func (c *CredentialBuilder) CreateStorageSpecSecretEnvs(namespace string, annota
 	if c.config.StorageSpecSecretName != "" {
 		storageSecretName = c.config.StorageSpecSecretName
 	}
-
-	// if the secret name is provided use it
-	if secretName != nil {
```
sivanantha321, line 102: any reason to remove this ?

pkg/credentials/service_account_credentials.go
brettmthompson: Spelling here looks to be correct. `Failed to mount` is logged since the method returning the error is `mountSecretCredential`

pkg/controller/v1alpha1/llmisvc/workload.go
Jooho: Effective Go and many Go projects, it’s common not to use Get when creating a getter. In other words, they prefer naming it X() instead of GetX().

So, I think the existing approach isn’t too bad.

pkg/credentials/service_account_credentials.go
Jooho: This is only called from one place, and as it always receives nil, retaining it is unnecessary.

https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L448

brettmthompson: The `secretName` and `storeConfigEnvVarName` input parameters are not utilized anywhere in code. See: https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L448

yuzisun (approved):

PR #4617, opened 2025-07-25
Llm d http route reconciler

andresllh:
Depends on https://github.com/kserve/kserve/pull/4614 and https://github.com/kserve/kserve/pull/4596
This PR adds the capability to reconcile routes for llm ISVC.

Fixes #

**Type of changes**

`make test`

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1alpha1/llm_inference_service_lifecycle.go
```diff
@@ -1,5 +1,9 @@
 /*
+<<<<<<< HEAD
```
hdefazio: merge conflict?

andresllh: Good catch

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (r *LLMISVCReconciler) collectReferencedRoutes(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) ([]*gwapiv1.HTTPRoute, error) {
+	if llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Route == nil || !llmSvc.Spec.Router.Route.HTTP.HasRefs() {
+		return nil, nil
+	}
+
+	referencedRoutes := make([]*gwapiv1.HTTPRoute, 0, len(llmSvc.Spec.Router.Route.HTTP.Refs))
+	for _, routeRef := range llmSvc.Spec.Router.Route.HTTP.Refs {
+		route := &gwapiv1.HTTPRoute{}
+		if err := r.Client.Get(ctx, types.NamespacedName{Namespace: routeRef.Name, Name: llmSvc.GetNamespace()}, route); err != nil {
```
gavrissh: Incorrect assignment?

```suggestion
		if err := r.Client.Get(ctx, types.NamespacedName{Namespace: llmSvc.GetNamespace(), Name: routeRef.Name}, route); err != nil {
```

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+type resolvedGateway struct {
+	gateway      *gwapiv1.Gateway
+	gatewayClass *gwapiv1.GatewayClass
+	parentRef    gwapiv1.ParentReference
+}
+
+func DiscoverGateways(ctx context.Context, c client.Client, route *gwapiv1.HTTPRoute) ([]resolvedGateway, error) {
+	gateways := make([]resolvedGateway, 0)
+	for _, parentRef := range route.Spec.ParentRefs {
+		ns := ptr.Deref((&parentRef).Namespace, gwapiv1.Namespace(route.Namespace))
```
gavrissh: Is parentRef a pointer? If not, this (&parentRef) is unnecessary and confusing?

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+func selectListener(gateway *gwapiv1.Gateway, sectionName *gwapiv1.SectionName) *gwapiv1.Listener {
+	if sectionName != nil {
+		for _, listener := range gateway.Spec.Listeners {
+			if listener.Name == *sectionName {
+				return &listener
+			}
+		}
+	}
+
+	return &gateway.Spec.Listeners[0]
```
gavrissh, line 128: Is this expected to fall back to the default if sectionName not found?

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type resolvedGateway struct {
+	gateway      *gwapiv1.Gateway
+	gatewayClass *gwapiv1.GatewayClass
+	parentRef    gwapiv1.ParentReference
+}
+
+func DiscoverGateways(ctx context.Context, c client.Client, route *gwapiv1.HTTPRoute) ([]resolvedGateway, error) {
```
yuzisun: this is an internal function we do not need to export

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+
+	log.FromContext(ctx).V(2).Info("Expected router scheduler deployment", "deployment", d)
+
+	return d
+}
+
+func schedulerConfigText(llmSvc *v1alpha1.LLMInferenceService) string {
+	switch {
+	case llmSvc.Spec.Prefill != nil:
+		return `
```
yuzisun: why do we return the string here?

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+				slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "-configText") ||
+				slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "--configFile") ||
+				slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "-configFile") {
+				// When the configuration is overridden, don't add/override it.
+				break
+			}
+
+			d.Spec.Template.Spec.Containers[i].Args = append(d.Spec.Template.Spec.Containers[i].Args,
+				"--configText",
+				schedulerConfigText(llmSvc),
```
yuzisun: Can we refer to the `EndpointPickerConfig` resource name instead of passing the text ?

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+
+	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Scheduler != nil && llmSvc.Spec.Router.Scheduler.Template != nil {
+		d.Spec.Template.Spec = *llmSvc.Spec.Router.Scheduler.Template.DeepCopy()
+		for i := range d.Spec.Template.Spec.Containers {
+			if d.Spec.Template.Spec.Containers[i].Name != "main" {
+				continue
+			}
+
+			if slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "--configText") ||
+				slices.Contains(d.Spec.Template.Spec.Containers[i].Args, "-configText") ||
```
yuzisun: why we use two different arg naming conventions here?

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	// We do not support Gateway's spec, when creating HTTPRoutes either the default gateway or those provided
+	// as refs are attached to reconciled routes
+	if err := r.reconcileHTTPRoutes(ctx, llmSvc); err != nil {
+		llmSvc.MarkRouterNotReady("HTTPRouteReconcileError", "Failed to reconcile HTTPRoute: %v", err.Error())
+		return fmt.Errorf("failed to reconcile HTTP routes: %w", err)
+	}
+
+	if err := r.reconcileIstioDestinationRules(ctx, llmSvc); err != nil {
```
yuzisun: This makes Istio as a hard dependency for llmisvc

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed_istio.go
```diff
@@ -0,0 +1,227 @@
...
+	if shadowSvc != nil {
+		dr.Spec.Host = network.GetServiceHostname(shadowSvc.GetName(), shadowSvc.GetNamespace())
+	}
+
+	log.FromContext(ctx).V(2).Info("Expected destination rule for workload", "destinationrule", dr)
+
+	return dr, nil
+}
+
+func (r *LLMISVCReconciler) getIstioShadowInferencePoolService(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) (*corev1.Service, error) {
```
yuzisun: Can you add comment to explain what's shadow InferencePool serivce?

yuzisun:
I have a bit concerns here that the implementation is tightly coupled with Istio API, ideally we should only use Gateway API so it is agnostic of the gateway implementation.

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -0,0 +1,481 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/pkg/kmeta"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	igwapi "sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+)
+
+func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) error {
```
yuzisun: Please add tests for these new files

terrytangyuan:
> I have a bit concerns here that the implementation is tightly coupled with Istio API, ideally we should only use Gateway API so it is agnostic of the gateway implementation.

Agreed. @andresllh Let's revisit the implementation in this PR to make sure to not depend on Istio.

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -50,3 +59,49 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	return &Config{
+		SystemNamespace:         constants.KServeNamespace,
+		IngressGatewayNamespace: igwNs,
+		IngressGatewayName:      igwName,
+		// TODO make it configurable
+		IstioGatewayControllerNames: []string{
+			"istio.io/gateway-controller",
+			"istio.io/unmanaged-gateway",
+			"openshift.io/gateway-controller",
```
terrytangyuan: These are hard-coded for Istio and OpenShift

pkg/controller/v1alpha1/llmisvc/scheduler.go
pierDipi: it's not a real CRD

pkg/controller/v1alpha1/llmisvc/scheduler.go
pierDipi: the config can be passed and overridden by users via configText or configFile flag (and in Go it can be passed with both single and double dash), so if it's overridden we don't use our default configuration

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router_discovery.go
andresllh: Spec and Listeners can't be empty, so it'll default to the first one in the list.

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router_discovery.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router_discovery.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: the string gets appended to container args, which is a string

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed_istio.go
andresllh: @pierDipi @bartoszmajsak @Jooho

pierDipi: Given the other comment on moving away from Istio specific APIs, do we still need to describe it?

Generally, this is an Istio implementation detail for the InferencePool API, it creates this service to add it to the "service registry"

pkg/controller/v1alpha1/llmisvc/scheduler.go
johnugeorge: Can't it be read into a type
https://github.com/kubeflow/katib/blob/master/pkg/util/v1beta1/katibconfig/config.go#L149-L160

pkg/controller/v1alpha1/llmisvc/controller.go
andresllh: I've removed any Istio dependencies from this PR. The logic will be modified and added as another PR later on.

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: the controller tests are rather large and there were a bunch of additions made to the testing package. Would it make more sense to add all tests as part of another PR?

pkg/controller/v1alpha1/llmisvc/workload_tls_self_signed_istio.go
andresllh: I've removed the Istio implementation from this PR, given that we want to move away from the hard dependency. I will rework it and add it as another PR later on.

andresllh:
Will be rebasing tomorrow morning @johnugeorge . For now it's ready for review once more. I've removed the Istio hard dependency from this PR so that it can be reworked and added in another PR later.

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: I've removed the istio hard dependency from this PR. It will be reworked and added as a separate PR.

pkg/controller/v1alpha1/llmisvc/scheduler.go
andresllh: https://github.com/kserve/kserve/issues/4629

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,209 @@
...
+			Namespace: llmSvc.GetNamespace(),
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha1.LLMInferenceServiceGVK),
+			},
+			Labels: RouterLabels(llmSvc),
+		},
+	}
+
+	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Route != nil && llmSvc.Spec.Router.Route.HTTP.Spec != nil {
+		httpRoute.Spec = *llmSvc.Spec.Router.Route.HTTP.Spec.DeepCopy()
```
yuzisun, line 133: I think we should create the HTTPRoute for user as kserve knows what routing rule to create instead of providing the spec

pkg/controller/v1alpha1/llmisvc/fixture/gwapi_builders.go
```diff
@@ -0,0 +1,364 @@
...
+
+type GatewayOption ObjectOption[*gwapiv1.Gateway]
+
+func Gateway(name string, opts ...GatewayOption) *gwapiv1.Gateway {
+	gw := &gwapiv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: gwapiv1.GatewaySpec{
+			GatewayClassName: defaultGatewayClass,
```
yuzisun, line 36: there is a gatewayclass in the configmap, we can read from there. 
https://github.com/kserve/kserve/blob/master/config/configmap/inferenceservice.yaml#L293

pkg/controller/v1alpha1/llmisvc/fixture/required_resources.go
```diff
@@ -0,0 +1,159 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: defaultGatewayClass,
+		},
+		Spec: gwapiv1.GatewayClassSpec{
+			ControllerName: "istio.io/gateway-controller",
+		},
+	}
+}
+
+func InferenceServiceCfgMap(ns string) *corev1.ConfigMap {
```
yuzisun, line 93: Is this used for testing?

pkg/controller/v1alpha1/llmisvc/fixture/gwapi_builders.go
andresllh: Even for test fixture?

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: A user can enable the route by just using `route: {}`  which will have the spec be injected using the presets, but it is still overridable if a user wants to override the preset.

Route preset:
https://github.com/kserve/kserve/blob/7794ff31bd39cab87e4d0664ee68df1af75ab395/pkg/controller/v1alpha1/llmisvc/config_merge.go#L101-L103

Jooho:
@yuzisun Could you please take another look when you have a chance?

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+			return nil, fmt.Errorf("failed to combine URLs for Gateway %s/%s: %w", g.gateway.Namespace, g.gateway.Name, err)
+		}
+
+		urls = append(urls, gatewayURLs...)
+	}
+
+	return urls, nil
+}
+
+func extractRoutePath(route *gwapiv1.HTTPRoute) string {
```
yuzisun, line 109: Can you comment on what is function doing?

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type resolvedGateway struct {
+	gateway      *gwapiv1.Gateway
+	gatewayClass *gwapiv1.GatewayClass
+	parentRef    gwapiv1.ParentReference
+}
+
+func discoverGateways(ctx context.Context, c client.Client, route *gwapiv1.HTTPRoute) ([]resolvedGateway, error) {
```
yuzisun, line 43: It would be helpful to add comments and provide the examples what these functions are doing

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return httpRoute
+}
+
+func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService, routes ...*gwapiv1.HTTPRoute) error {
+	logger := log.FromContext(ctx)
+
+	var urls []*apis.URL
+	for _, route := range routes {
+		discoverURL, err := DiscoverURLs(ctx, r.Client, route)
```
yuzisun: discoverURL is an array, so should be using plural

pkg/controller/v1alpha1/llmisvc/fixture/gwapi_builders.go
yuzisun: nvm, did not realize these are fixtures based on the name of the file.

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,209 @@
...
+	return nil
+}
+
+func (r *LLMISVCReconciler) reconcileHTTPRoutes(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService) error {
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling HTTPRoute")
+
+	expectedHTTPRoute := r.expectedHTTPRoute(ctx, llmSvc)
+
+	// TODO should we remove "llmSvc.Spec.Router.Route.HTTP == nil" from the condition below so that a non nil Route meeans "all type of routes are enabled"?
```
yuzisun:
```suggestion
	// TODO should we remove "llmSvc.Spec.Router.Route.HTTP == nil" from the condition below so that a non nil Route means "all type of routes are enabled"?
```

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+
+	slices.SortStableFunc(urls, func(a, b *apis.URL) int {
+		return cmp.Compare(a.String(), b.String())
+	})
+
+	externalURLs := FilterExternalURLs(urls)
+	if len(externalURLs) == 0 {
+		logger.Info("no public URL discovered")
```
yuzisun, line 172: should this be an error instead ? the llmisvc should always have a url

pkg/controller/v1alpha1/llmisvc/router_discovery.go
```diff
@@ -0,0 +1,203 @@
...
+			values = append(values, addr.Value)
+		}
+	}
+	return values
+}
+
+func combineIntoURLs(hostnames []string, scheme string, port gwapiv1.PortNumber, path string) ([]*apis.URL, error) {
+	urls := make([]*apis.URL, 0, len(hostnames))
+
+	sortedHostnames := make([]string, len(hostnames))
```
yuzisun, line 167: Should we explicitly configure the primary host instead of relying on the sorting ?

pkg/controller/v1alpha1/llmisvc/router_discovery.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: this will only expose the first available public url, the private url should always be there. I guess the question is, if there are no public urls, do we want to have the privateURL listed in the llmSvc.Status.URL or just keep it blank? All available URLs will still be listed under llmSvc.Status.Addresses both public and private.
Any thoughts? @yuzisun @pierDipi @bartoszmajsak @Jooho

pkg/controller/v1alpha1/llmisvc/router_discovery.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router_discovery.go
bartoszmajsak: Could you elaborate on the use case here? I am curious where that would help. 

What would be the primary hostname in this case - an ingress domain?

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,210 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/utils/ptr"
+
+	"k8s.io/apimachinery/pkg/api/equality"
```
sivanantha321, line 33: nit: reformat the imports

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,210 @@
...
+	"fmt"
+	"slices"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/utils/ptr"
```
sivanantha321, line 31: nit: reformat imports

pkg/controller/v1alpha1/llmisvc/router_discovery.go
yuzisun: Since we are only showing one URL in the status, should that be an explicitly configured domain instead of relying on sorting logic which is a bit arbitrary.

pkg/controller/v1alpha1/llmisvc/router.go
Jooho: I think externalURL should be treated as info rather than an error, because there are several situations where it may be unset either unintentionally or by design.
For example, if there is no external load balancer setup, or if it’s not configured properly, externalURL may be missing — and this could be intentional. Another case is when a user wants the service to remain internal and does not want it exposed.
Therefore, in these scenarios, logging it as informational rather than an error makes more sense.

yuzisun: Is there an error condition in the status to indicate that the load balancer is not configured properly?

Jooho: By default, the following logic treats the absence of any external addresses in the Gateway’s status as an error.
If the configuration is correct, this field should be properly populated with the expected address.

https://github.com/kserve/kserve/pull/4617/files/1e5446ab1a187c4d236a1a03e2006851c96ef4b6#diff-39e77924bad68b0a93b51989b48b2dac5e69810aa61ee13382540d77799163eaR77

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router.go
andresllh: Done

pkg/controller/v1alpha1/llmisvc/router.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return httpRoute
+}
+
+func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1alpha1.LLMInferenceService, routes ...*gwapiv1.HTTPRoute) error {
+	logger := log.FromContext(ctx)
+
+	var urls []*apis.URL
+	for _, route := range routes {
+		discoverURLs, err := DiscoverURLs(ctx, r.Client, route)
```
sivanantha321, line 156: Is there a possibility for duplicate urls ?

yuzisun (approved):

yuzisun (approved):

pkg/controller/v1alpha1/llmisvc/router_discovery.go
bartoszmajsak: We can address it as a follow-up. The idea is that this function sorts all addresses to have them stable. Then we find first external address and expose is as `.status.url`, everything goes into `.status.addresses`. Is it fine for you @yuzisun ?

https://github.com/kserve/kserve/blob/b5e3d533eb7e2f5b6a02b9d4c01a6baa00fc0eb0/pkg/controller/v1alpha1/llmisvc/router.go#L165-L182

bartoszmajsak: We don't need to sort here, as it's done in the caller, I will clean it up.

PR #4621, opened 2025-07-28
feat(build): runs flake8 on *.py tracked by git

bartoszmajsak:
Otherwise any ignored files or directories containing python files are scanned which might not be desired.

For example using worktree (e.g. in globally ignored `.worktree`  folder that is local to cloned repo) leads to false positives.

**Type of changes**

**Release note**:
```release-note
none
```

**Re-running failed tests**


bartoszmajsak:
I might be solving the wrong problem. Closing.

PR #4622, opened 2025-07-28
Remove storage spec from LLMIsvc

israel-hdez:
The Storage Spec was added to InferenceServices because this is the format that ModelMesh supports. For non-ModelMesh deployments, KServe added support for compatibility.

Given ModelMesh is virtually stalled, likely LLMIsvc CRD shouldn't carry storage spec.

N/A

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


johnugeorge:
LGTM. We will be primarily relying on URI field

Jooho:
LGTM to me too :)

PR #4608, opened 2025-07-23
refactor: 4602 - Refactor "RawDeployment" to "Standard" and "Serverless" to  "KNative" to clarify usage

cjohannsen-cloudera:
**What this PR does / why we need it**:

Refactor the names and values used for the deploy mode to clarify their meaning.  The current values are difficult to understand and the new values map directly to the implementations.

- RawDeployment has changed to Standard
- Serverless has changed to KNative
- Default is now Standard
- Legacy values preserved as deprecated fields.  
- Legacy values are upconverted to current values in config constructor

Fixes # 4602

**Type of changes**
Please delete options that are not relevant.

- [x] New feature 
- [x] Breaking change - new default value
- [x] This change requires a documentation update

**Feature/Issue validation/testing**:

All unit tests executed after refactor to ensure no breaking changes introduce.
DeployConfig will now map the deprecated values to supported values at construction

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x] Have you made corresponding changes to the documentation? https://github.com/kserve/website/pull/506

**Release note**:
```release-note
Update names for deployMode to reflect actual usage.
```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

PR #4623, opened 2025-07-29
Implement the progressive rollout for raw deployment

### LINKED ISSUE #4618: Strategy of the Rolling-out for raw deployment
/kind bug

The raw deployment rolling upgrade on kserve should have an option with the default strategy with ResourceAware and Availability mode.

With resource util mode, we prioritize the usage of compute resource. Old pods are terminated before the new pods are launched.
With high availability mode, we prioritize the the service uptime. Old pods are terminated after the new pods are launched.

We should have the option to configure the rollout strategy with the isvc yaml.

We plan to implement the progressive rollout for the raw deployment in Kserve.
There are two modes: one can be named Availability and the other can be named ResourceAware.
The Availability mode means the deployment will roll out by launching new pods first and terminating the old pods later. While the ResourceAware mode means the deployment will roll out
by terminating the old pods first and then launching the new pods.
 
We can leverage the following two fields to configure maxSurge and maxUnavailable. If it is set
to the Availability mode, maxUnavailable is set to 0 and maxSurge is set to either a integer or a percentage value(rollout-ratio); if it is set to the ResourceAware mode, maxSurge is set to 0 and maxUnavailable is set to either a integer or a percentage value.

There will be some changes to the inference service CRD. We need to add two fields to define the
rollout strategy: Availability or ResourceAware, and the integer or the percentage value to specify either maxSurge or maxUnavailable, based on the rollout strategy.

In the inference service CRD, there is a section for the predictor and a section for the transformer. All the above configurations should be made available to both of them.

If the user does not specify anything regarding the rollout strategy. There should be default
values available for the rollout strategy and the rollout-ratio. The default values should be available in the ConfigMap, so that we can configure globally for all inference services.

**Environment:**

- KServe Version: main branch

houshengbo:

Fixes #4618

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


docs/samples/rollout-strategy-example.yaml
```diff
@@ -0,0 +1,33 @@
...
+    custom:
+      container:
+        image: my-transformer:latest
+        env:
+          - name: MODEL_NAME
+            value: "my-model"
+    # Example 2: ResourceAware mode - terminates old pods first, then launches new pods
+    # This prioritizes resource efficiency over availability
+    rollout:
+      mode: "ResourceAware"
```
johnugeorge: Is there any advantage in bringing a wrapper than exposing the maxUnavailable and maxSurge directly?

houshengbo: It is more intuitive to the users to understand either Availability mode prioritizing service uptime or ResourceAware mode prioritizing resource usage. One launches new pods first and then terminate the old pods during the rollout alternatively, and vice versa for the other.

If we expose maxUnavailable and maxSurge, the users have to first understand what these params mean and then figure out what value combinations can lead to what kind of rollout strategies.

From inference service's perspective, users only need to select one rollout mode, or just use the default mode. It shields the process of figuring out the combination of maxUnavailable and maxSurge. Otherwise, when we explain this feature, we would always refer to the concept of maxUnavailable and maxSurge, and different values leading to different strategies.(It is not quite friendly to remember maxUnavailable and maxSurge). It simplifies our explanation of this feature as well.

docs/ROLLOUT_STRATEGY_GUIDE.md
```diff
@@ -0,0 +1,309 @@
...
+- **maxSurge**: `25%`
+
+### Special Case: Multinode Deployments
+
+For multinode deployments, KServe automatically overrides the rollout strategy to ensure high availability:
+
+- **maxUnavailable**: `0%` (no pods are taken down during rollout)
+- **maxSurge**: `100%` (can have up to double the number of pods during rollout)
+
+This ensures that original pods remain available until new pods are ready.
```
yuzisun: @Jooho can you help review if this is ok?

docs/ROLLOUT_STRATEGY_GUIDE.md
```diff
@@ -0,0 +1,309 @@
...
+## Migration from Previous Versions
+
+If you were using the previous rollout strategy configuration (individual `rolloutStrategy` and `rolloutRatio` fields), update your InferenceService specs to use the new nested `rollout` object:
+
+```yaml
+# Old format (deprecated)
+spec:
+  predictor:
+    rolloutStrategy: "Availability"
+    rolloutRatio: "25%"
```
yuzisun: I do not think kserve has supported this ?

docs/samples/rollout-strategy-example.yaml
yuzisun: I find the ratio field is still hard for user to understand what it means, we may still end up pointing user to the maxUnavailable/maxSurge doc.

docs/ROLLOUT_STRATEGY_GUIDE.md
Jooho: @houshengbo From my best memory, it would be the right value for multi-node. Have you tested?

docs/samples/rollout-strategy-example.yaml
Jooho: I share a similar view. Deployment rollout is a fundamental Kubernetes API, and most users familiar with Kubernetes would likely already understand it. Wrapping it might unintentionally make things a bit more confusing for them.

houshengbo: Thx for the feedback. I can replace the `strategy section` in the inference service CRD with:

```
...
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 2
      maxUnavailable: 1
...
```

The user can define the type and maxSurge & maxUnavailable in the rollingUpdate directly.
We can remove the configuration about the ratio, since the user can specify any combination of maxSurge & maxUnavailable.

If the user does not specify anything for the strategy, the raw deployment can just pick up the default, which is the same default strategy for the deployment resource.

What do you think?

Jooho: Since we are already reusing [Kubernetes DeploymentStrategy](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/component.go#L133) in the InferenceService, both int and % values are supported for maxSurge and maxUnavailable.

I think we already support configuring the deployment type and the rollingUpdate options. I tested it using the Recreate strategy with [Multi-node/Multi-gpu](https://kserve.github.io/website/docs/model-serving/generative-inference/multi-node). 

>We can remove the configuration about the ratio, since the user can specify any combination of maxSurge & maxUnavailable.If the user does not specify anything for the strategy, the raw deployment can just pick up the default, which is the same default strategy for the deployment resource.

This is what we currently have, unless I’m misunderstanding the code.

houshengbo: Thanks for pointing out the DeploymentStrategy. Then we don't really need to change anything in the inference service's CRD or implement anything on the inference service level.

Would it be valuable to add default section in the configMap? Like
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: inferenceservice-config
  namespace: kserve
data:
  deploy: |-
    {
      "defaultDeploymentMode": "Raw",
      "rawDeploymentRolloutStrategy": {
        "defaultRollout": {
          "mode": "Availability",
          "maxSurge": "25%"
          "maxUnavailable": "25%"
        }
      }
    }
```
This provides an option to configure the RollouStrategy globally for all inference services, if DeploymentStrategy is empty in the inference service.

If the mode is configured as `Availability`, and DeploymentStrategy is empty in the inference service, the deployment resource will set the maxUnavailable to 0, and the maxSurge to the value as defined in the configMap.

If the mode is configured as `ResourceAware`, and DeploymentStrategy is empty in the inference service, the deployment resource will set the  maxSurge to 0, and the maxUnavailable to the value as defined in the configMap.

docs/ROLLOUT_STRATEGY_GUIDE.md
houshengbo: https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go#L196 This is the logic of the code for multi-node. I just updated based the logic here.

config/configmap/inferenceservice.yaml
```diff
@@ -537,22 +537,41 @@ data:
...
-         # RawDeployment https://kserve.github.io/website/master/admin/kubernetes_deployment/
-         # ModelMesh https://kserve.github.io/website/master/admin/modelmesh/
-         "defaultDeploymentMode": "Serverless"
-       }
+    # ====================================== DEPLOYMENT CONFIGURATION ======================================
+    # Example
+    deploy: |-
+      {
+        "defaultDeploymentMode": "RawDeployment",
+        "rawDeploymentRolloutStrategy": {
```
yuzisun: We already changed rawDeployment in https://github.com/kserve/kserve/pull/4608/files, let's rename this to `deploymentRolloutStrategy` ?

docs/ROLLOUT_STRATEGY_GUIDE.md
```diff
@@ -0,0 +1,364 @@
...
+You can configure Kubernetes deployment strategy directly at the component level:
+
+```yaml
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
+  name: deployment-strategy-example
+  namespace: default
+  annotations:
+    serving.kserve.io/deploymentMode: "RawDeployment"
```
yuzisun: same here, let's change to `standard`. We also need to move this doc to the website https://github.com/kserve/website

test/crds/rollout-strategy.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: serving.kserve.io/v1beta1
```
yuzisun: this is the directory for crds not the actual examples

config/configmap/inferenceservice.yaml
```diff
@@ -537,22 +537,38 @@ data:
...
-         # Serverless https://kserve.github.io/website/master/admin/serverless/serverless/
-         # RawDeployment https://kserve.github.io/website/master/admin/kubernetes_deployment/
-         # ModelMesh https://kserve.github.io/website/master/admin/modelmesh/
-         "defaultDeploymentMode": "Serverless"
-       }
+    # ====================================== DEPLOYMENT CONFIGURATION ======================================
+    # Example
+    deploy: |-
+      {
+        "defaultDeploymentMode": "Standard",
```
yuzisun: let's keep the original default value for backwards compatibility.

docs/ROLLOUT_STRATEGY_GUIDE.md
```diff
@@ -0,0 +1,364 @@
+# Rollout Strategy Guide
```
yuzisun: Can you move this doc to the website repo?

config/configmap/inferenceservice.yaml
houshengbo: You mean to use Knative, not Serverless. Correct? Do we still support Serverless?

docs/ROLLOUT_STRATEGY_GUIDE.md
houshengbo: yes.

yuzisun (approved):

PR #4630, opened 2025-07-31
feat(webhook): introduces Validating Webhook for LLMInferenceServiceConfig

VedantMahabaleshwarkar:

Adds a webhook test suite and necessary deps and a validating webhook for LLMInferenceServiceConfig 

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


VedantMahabaleshwarkar:
PTAL @bartoszmajsak @pierDipi @johnugeorge @sivanantha321

Makefile
```diff
@@ -218,6 +225,8 @@ deploy: manifests
 
 
 deploy-dev: manifests
+	kubectl apply --server-side=true --force-conflicts -k config/crd
```
Jooho, line 234: nit: Could you please add a comment on why we need to create CRD twice?

brettmthompson: Added the following:
```
	# Given that llmisvc CRs and CRDs are packaged together, when using kustomize build a race condition will occur.
	# This is because before the CRD is registered to the api server, kustomize will attempt to create the CR.
	# The below kubectl apply and kubectl wait commands are necessary to avoid this race condition.

PR #4642, opened 2025-08-12
fix: Adds support for NVIDIA MIG GPU resource detection

### LINKED ISSUE #4640: Is default image of HuggingFace server non-GPU in 0.15?
/kind bug

We updated our KServe setup from 0.14 to latest 0.15.2 version and all our LLM models with HuggingFace serving runtime crashed because there is no support for GPU in new version by default.

Since two types of kserve/huggingfaceserver image have started from 0.15.* version (GPU and non-GPU) we expected that default values will still be GPU latest tag. 
Here we see that the default version is non-GPU: https://github.com/kserve/kserve/blob/master/charts/kserve-resources/values.yaml#L332

```
kind: InferenceService
metadata:
  name: huggingface-llama3
  annotations:
    sidecar.istio.io/inject: "false"
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      args:
        - --model_name=llm_model
        - --model_id=Qwen/Qwen2.5-14B-Instruct
        - --max-model-len=2048
      env:
        - name: HF_TOKEN
          valueFrom:
            secretKeyRef:
              name: hf-secret
              key: HF_TOKEN
              optional: false
      resources:
        limits:
          cpu: "6"
          memory: 24Gi
          nvidia.com/mig-7g.80gb: "1"
        requests:
          cpu: "6"
          memory: 24Gi
          nvidia.com/mig-7g.80gb: "1"
```

I'm confused because it's not compatible and it took several hours to find out that there is problem in the image, not the hardware.
Please help! Maybe we are doing smth wrong and somehow incorrectly updated KServe charts.

- KServe Version: 0.15.2
- Cloud Environment: Kubernetes

sivanantha321:
Extends GPU detection logic to recognize NVIDIA MIG resource names, enabling accurate identification of GPU-enabled workloads using MIG instances. Updates tests to cover MIG GPU scenarios for improved reliability.

Fixes #4640

**Type of changes**

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -55,7 +55,16 @@ func AppendVolumeIfNotExists(slice []corev1.Volume, volume corev1.Volume) []core
 
 func IsGPUEnabled(requirements corev1.ResourceRequirements) bool {
 	_, ok := requirements.Limits[constants.NvidiaGPUResourceType]
-	return ok
+	if ok {
+		return true
+	}
+	for resourceName := range requirements.Limits {
+		// Check if the resource name has the MIG prefix
+		if strings.HasPrefix(string(resourceName), constants.NvidiaMigGPUResourceTypePrefix) {
+			return true
+		}
+	}
+	return false
 }
 
 // FirstNonNilError returns the first non nil interface in the slice
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -259,10 +259,11 @@ var (
 
 // GPU Constants
 const (
-	NvidiaGPUResourceType = "nvidia.com/gpu"
-	AmdGPUResourceType    = "amd.com/gpu"
-	IntelGPUResourceType  = "intel.com/gpu"
-	GaudiGPUResourceType  = "habana.ai/gaudi"
+	NvidiaGPUResourceType          = "nvidia.com/gpu"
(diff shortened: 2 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -55,7 +55,17 @@ func AppendVolumeIfNotExists(slice []corev1.Volume, volume corev1.Volume) []core
...
 func IsGPUEnabled(requirements corev1.ResourceRequirements) bool {
 	_, ok := requirements.Limits[constants.NvidiaGPUResourceType]
-	return ok
+	if ok {
+		return true
+	}
+	for resourceName, _ := range requirements.Limits {
+
+	// Check if the resource name has the MIG prefix
+	if strings.HasPrefix(string(resourceName), constants.NvidiaMigGPUResourceTypePrefix) {
```
gavrissh: In `IsGPUEnabled()`, we check only for NvidiaTypes?

sivanantha321: We are only providing prebuilt  Nvidia cuda gpu images, right ?

gavrissh: Got it, Thanks!

yuzisun (approved):

PR #4663, opened 2025-08-25
[llmisvc] Improve config merge and update well-known presets

pierDipi:
## Update LLMInferenceServiceConfig well-known preset

This brings improvements and fully working well-known presets for P/D, single and multi node

Multi node DP+EP will use one pod per node for now (as the one pod per rank is not working well in vLLM or requires elevated permissions) 

## Improve config_merge logic to handle no scheduler cases

Improved config_merge and combineBaseRefsConfig logic to handle 
- no scheduler case
- External InfernecePool object

## Update LoRA types in llmisvc to use LLMModelSpec

The LLMInferneceService CRD LoRA API wasn't meant to use the existing ModelSpec but the LLMModelSpec type

This will be used and tested in https://github.com/kserve/kserve/pull/4666


charts/llmisvc-resources/templates/config-llm-decode-template.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   template:
     containers:
-      - image: ghcr.io/llm-d/llm-d:v0.2.0
+      - image: ghcr.io/llm-d/llm-d-dev:v0.2.2
```
spolti, line 8: is this expected?

pierDipi: yes, I wanted to pick up the latest to get some vllm fixes, the plan is to eventually (really soon) switch to 0.3

charts/llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
```diff
@@ -52,66 +62,39 @@ spec:
...
-                --ssl-certfile \
-                /etc/ssl/certs/tls.crt \
-                --ssl-keyfile \
-                /etc/ssl/certs/tls.key
-            fi
+            START_RANK=0
+            eval "vllm serve \
+              /mnt/models \
+              --served-model-name "{{ .Spec.Model.Name }}" \
+              --port 8001 \
```
spolti, line 80: Shouldn't the port be parameterized?

pkg/controller/v1alpha1/llmisvc/config_merge.go
```diff
@@ -170,6 +196,38 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
 		return llmSvcCfg, err
 	}
 
+	// Point HTTPRoute to a Service if there is no Scheduler or InferencePool, and the HTTPRoute uses the default
```
spolti, line 199: Is there a test for it?

pierDipi: we will have tests for all the pieces here https://github.com/kserve/kserve/pull/4666

charts/llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
pierDipi: No, why? most users won't need to change it (as this port is also referenced in the routing-sidecar)

PR #4666, opened 2025-08-26
[llmisvc] Update workload and router reconciliation

pierDipi:
- Update single and multi node workload reconciliation to have service accounts where we can attach additional roles
- Improve router-related status conditions and reporting with specific sub-conditions
- Add additional comments to every function and critical logic
- Add envtest-based tests for single, multi node and routing


charts/llmisvc-resources/templates/gateway-inference-extension.yaml
```diff
@@ -0,0 +1,540 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.1
+  name: inferencemodels.inference.networking.x-k8s.io
```
yuzisun, line 6: Do we want to install GIE as part of kserve?

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -248,8 +226,162 @@ func (r *LLMISVCReconciler) updateStatus(ctx context.Context, desired *v1alpha1.
...
+// The controller conditionally registers watchers based on CRD availability to avoid errors in environments where optional CRDs are not installed.
 func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+	logger := mgr.GetLogger().WithName("LLMInferenceService.SetupWithManager")
+
+	b := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.LLMInferenceService{}).
-		Complete(r)
+		Watches(&v1alpha1.LLMInferenceServiceConfig{}, r.enqueueOnLLMInferenceServiceConfigChange(logger)).
+		Owns(&netv1.Ingress{}, builder.WithPredicates(childResourcesPredicate)).
```
yuzisun, line 237: should also include HTTPRoute?

yuzisun:
@pierDipi Let me know when this is ready for review

charts/llmisvc-resources/templates/gateway-inference-extension.yaml
pierDipi: I think generally we want to have some control of the versioning of GIE (as Kserve is the primary controller of GIE and EPP), however, we could have a bring your own option

pkg/controller/v1alpha1/llmisvc/scheduler.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The KServe Authors.
+Copyright 2023 The KServe Authors.
```
terrytangyuan: This seems unintentional?

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -75,84 +88,40 @@ type Config struct {
...
+		// Resource is not being deleted, ensure finalizer is present
 		if controllerutil.AddFinalizer(original, finalizerName) {
 			if err := r.Update(ctx, original); err != nil {
 				return ctrl.Result{}, err
 			}
 		}
 	} else {
-		logf.Log.Info("Marked for deletion, finalizing resources")
+		// Resource is being deleted, perform cleanup
+		logger.Info("Marked for deletion, finalizing resources")
```
yuzisun, line 118: log the resource name

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -163,76 +132,85 @@ func (r *LLMISVCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	// reconciler.PreProcessReconcile(ctx, resource)
-	// reconciler.PostProcessReconcile(ctx, resource, original)
+	// Pre/post process hooks for status management
+	reconciler.PreProcessReconcile(ctx, resource)
 	reconcileErr := r.reconcile(ctx, resource)
+	reconciler.PostProcessReconcile(ctx, resource, original)
 
 	if reconcileErr != nil {
-		logf.Log.Error(reconcileErr, "Failed to reconcile LLMInferenceService")
+		logger.Error(reconcileErr, "Failed to reconcile LLMInferenceService")
```
yuzisun, line 145: same here for logging the resource name

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -163,76 +132,85 @@ func (r *LLMISVCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// Combine base configurations with service-specific overrides
+	// This includes default configs based on deployment pattern (single node, multi-node, etc.)
 	baseCfg, err := r.combineBaseRefsConfig(ctx, llmSvc, config)
 	if err != nil {
 		llmSvc.MarkPresetsCombinedNotReady("CombineBaseError", err.Error())
 		return fmt.Errorf("failed to combine base-configurations: %w", err)
 	}
 	llmSvc.MarkPresetsCombinedReady()
+
+	logger.Info("Reconciling with combined base configurations", "combined.spec", baseCfg.Spec, "original.spec", llmSvc.Spec)
```
yuzisun: this should be debug? spec might be large.

pkg/controller/v1alpha1/llmisvc/controller.go
```diff
@@ -248,8 +226,162 @@ func (r *LLMISVCReconciler) updateStatus(ctx context.Context, desired *v1alpha1.
...
+	}
+	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), lwsapi.GroupVersion.String(), "LeaderWorkerSet"); ok && err == nil {
+		b = b.Owns(&lwsapi.LeaderWorkerSet{}, builder.WithPredicates(childResourcesPredicate))
+	}
+
+	return b.Complete(r)
+}
+
+// enqueueOnGatewayChange creates an event handler that triggers reconciliation of LLMInferenceServices
+// when a Gateway they depend on changes. This is necessary because Gateway changes can affect routing.
```
yuzisun, line 274: what are the gateway changes what can affect LLMInferenceServices?

pkg/controller/v1alpha1/llmisvc/workload.go
```diff
@@ -22,51 +22,118 @@ import (
...
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Name:        "https",
+					Protocol:    corev1.ProtocolTCP,
+					AppProtocol: ptr.To("https"),
+					Port:        8000,
+					TargetPort: intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 8000,
```
yuzisun, line 119: why this is hard coded ? it depends on the workload ports defined on container spec

pkg/controller/v1alpha1/llmisvc/workload_multi_node.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The KServe Authors.
+Copyright 2023 The KServe Authors.
```
yuzisun:
```suggestion
Copyright 2025 The KServe Authors.
```

pkg/controller/v1alpha1/llmisvc/workload_single_node.go
```diff
@@ -203,24 +241,30 @@ func (r *LLMISVCReconciler) propagateDeploymentStatus(ctx context.Context, expec
 			return nil
 		}
 	}
-
+	logger.Info("Deployment processed")
```
yuzisun: this logging can be removed

pkg/controller/v1alpha1/llmisvc/controller.go
pierDipi: The Gateway(s) are fetched to set the llmisvc URL(s) in the status (see router_discovery.go `DiscoverURLs`)

pkg/controller/v1alpha1/llmisvc/workload.go
pierDipi: The rationale for not supporting changing the vLLM port is because it requires a lot of changes beyond this service (presets, routing sidecar flags, routing sidecar container spec , etc) or selecting the exact port is not trivial, for example, when there is P/D, the request goes through the routing sidecar, so here we would need to use the routing sidecar container ports.

The presets we bundle assume `8000` for the "main receiver" (being it vllm or P/D routing sidecar) and for now, I'd discourage that use case unless we have a use case that requires not assuming "main receiver" port.

pkg/controller/v1alpha1/llmisvc/controller.go
pierDipi: I will add this to the logger so that it's propagated and used in every log line during reconciliation

pkg/controller/v1alpha1/llmisvc/workload.go
yuzisun: Ok let’s add a comment then.

pierDipi: Added

PR #4693, opened 2025-09-16
Fixed the issue with the quantity for autoScale section

### LINKED ISSUE #4692: Fractional values are not correctly supported by isvc for the autoscaling section
/kind bug

The fractional value was not well supported for autoscaling.
For example, if we apply the yaml with the following section in the spec:
```
...
    autoScaling:
      metrics:
        - podmetric:
            metric:
              backend: opentelemetry
              metricNames:
                - vllm:gpu_cache_usage_perc
              operationOverTime: avg
              query: sum(vllm:gpu_cache_usage_perc)
            target:
              type: Value
              value: "0.6"
          type: PodMetric
...
```
The target value for the metric "vllm:gpu_cache_usage_perc" is between 0 and 1, and the fractional value like 0.6 is supported by the scaledobject.

However, when the yaml with the above content is applied, our isvc will automatically adjust the value of 0.6 into 60m, which will pass the value 60m to the scaledobject instead of 0.6. This will make the ScaledObject fail to recognize the metric "vllm:gpu_cache_usage_perc" as the metric to scale the workload, but falling back to the default cpu.

If user specify values like "0.6" in the spec, shall we keep it as it is, instead of changing it into 60m. Especially for the autoscaling section.

The Root Cause:
KServe uses *resource.Quantity → Kubernetes automatically normalizes "0.90" to "900m"
KEDA uses map[string]string → Expects plain string values like "0.9", not Kubernetes quantities
The impedance mismatch: KServe's quantity-based system vs KEDA's string-based system

We need to make sure keda pick up the original fractional value.
Change proposed: https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go#L151 
```
targetValue := metric.PodMetric.Target.Value.String()
```
can be changed into 	
```
targetValue := fmt.Sprintf("%g", metric.PodMetric.Target.Value.AsApproximateFloat64())
```

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

houshengbo:

Fixes #4692

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
@@ -93,11 +93,11 @@ func getKedaMetrics(componentMeta metav1.ObjectMeta, componentExt *v1beta1.Compo
 					}
 				case v1beta1.AverageValueMetricType:
 					if metric.Resource.Target.AverageValue != nil {
-						targetValue = metric.Resource.Target.AverageValue.String()
+						targetValue = *metric.Resource.Target.AverageValue
 					}
 				case v1beta1.ValueMetricType:
 					if metric.Resource.Target.Value != nil {
-						targetValue = metric.Resource.Target.Value.String()
+						targetValue = *metric.Resource.Target.Value
 					}
 				}
 				triggers = append(triggers, kedav1alpha1.ScaleTriggers{
@@ -115,7 +115,7 @@ func getKedaMetrics(componentMeta metav1.ObjectMeta, componentExt *v1beta1.Compo
 					Metadata: map[string]string{
 						"serverAddress": serverAddress,
 						"query":         query,
-						"threshold":     fmt.Sprintf("%f", metric.External.Target.Value.AsApproximateFloat64()),
+						"threshold":     *metric.External.Target.Value,
 					},
 				}
 
@@ -148,7 +148,7 @@ func getKedaMetrics(componentMeta metav1.ObjectMeta, componentExt *v1beta1.Compo
 
 				triggerType := string(metric.PodMetric.Metric.Backend)
 				query := metric.PodMetric.Metric.Query
-				targetValue := fmt.Sprintf("%g", metric.PodMetric.Target.Value.AsApproximateFloat64())
+				targetValue := *metric.PodMetric.Target.Value
 
 				trigger := kedav1alpha1.ScaleTriggers{
 					Metadata: map[string]string{},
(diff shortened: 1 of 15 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -148,7 +148,7 @@ func getKedaMetrics(componentMeta metav1.ObjectMeta, componentExt *v1beta1.Compo
 
 				triggerType := string(metric.PodMetric.Metric.Backend)
 				query := metric.PodMetric.Metric.Query
-				targetValue := metric.PodMetric.Target.Value.String()
+				targetValue := fmt.Sprintf("%g", metric.PodMetric.Target.Value.AsApproximateFloat64())
```
yuzisun: how does it work for the other pod metrics target like memory which is actually a string "50Gi" ?

houshengbo: It will become the Scientific notation like "5.36870912e+10".
KEDA does not support it, since it expects the decimal/integer strings.
Looks like we need to change the type for metric.PodMetric.Target.Value and metric.PodMetric.Target.AverageValue into string. They are now both resource.Quantity. 

These values are used either for keda or the validation purpose. I think it should be fine that we keep them as they are using string type.
But need more thoughts on it to confirm.

andyi2it:
@yuzisun @houshengbo 
The current implementation of PodMetricSourceType is designed primarily for external metrics, such as those used by vLLM, rather than for standard resource metrics like CPU or memory. 

Standard resource metrics, such as CPU and memory, are handled separately under the ResourceMetricSourceType case. For memory, the code uses the .String() method on the Quantity type, preserving the original string representation (e.g., "50Gi"), which is what KEDA expects for memory metrics.

houshengbo:
Do you mean this line targetValue := metric.PodMetric.Target.Value.String() preserve the original value? @andyi2it 
I tested it, it does not.
If I set "0.6" in the spec, metric.PodMetric.Target.Value.String() gives "600m".


This is part of the yaml content I applied:
```
...
    autoScaling:
      metrics:
        - type: PodMetric          # This is MetricsSpec.Type
          podmetric:               # This is MetricsSpec.PodMetric
            metric:                # This is PodMetricSource.Metric
              backend: opentelemetry
              metricNames:
                - vllm:gpu_cache_usage_perc
              query: vllm:gpu_cache_usage_perc
            target:                # This is PodMetricSource.Target
              type: Value
              value: "0.6"
...
```

After the creation of the isvc, I check the isvc manifest with "kubectl get isvc -o yaml". This is what I see in the cluster:
```
...
    autoScaling:
      metrics:
      - podmetric:
          metric:
            backend: opentelemetry
            metricNames:
            - vllm:gpu_cache_usage_perc
            query: vllm:gpu_cache_usage_perc
          target:
            type: Value
            value: 600m
        type: PodMetric
...
```

The value was converted into `value: 600m` from `value: "0.6"` automatically since we use resource.Quantity as the type. The user input 0.6 was lost, and cannot be passed to the ScaledObject. What keda needs is the value 0.6 as the original in this case.

houshengbo:
Refer to https://github.com/kserve/kserve/pull/4699 for new fix.

Issue #3413, opened 2024-02-06
Support multiple StorageUri in Inference Service


lizzzcai:
/kind feature

**Describe the solution you'd like**

Crrently KServe only supports one StorageUri, which fit most of the cases. However, in some scenarios like serving fine-tuned models like LoRA, base model and the LoRA adapters might come from different buckets (it is possible to put it into one but in some cases, it will not work if they belong to different owners). So allowing users to provide multiple storageUri will help in this case, similar to volumeMounts in K8s.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
As multiple storageUri are supported, user should be able to specify the destination of the mounted storage. Below is one of the example:
```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: "huggingface"
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      runtime: kserve-huggingface
      storageUri:
      - src: gs://kfserving-examples/models/huggingface
        dest: /mnt/models
      - src: gs://kfserving-examples/adapters/huggingface
        dest: /mnt/adapters
```

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

skuchipu:
@lizzzcai  do we have runtime kserve-huggingface that comes with Kserve deployment ?

lizzzcai:
Hi @skuchipu , sorry for the late reply. Yes, you can follow [this](https://github.com/kserve/kserve/issues/3393), it is already in `v0.12.0`.

example:

```
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: bloom
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      #runtime: kserve-huggingface
      args:
      - --model_id=bigscience/bloom-560m
      - --tensor_input_names=input_ids
      resources:
        limits:
          cpu: "2"
          memory: 4Gi
        requests:
          cpu: "1"
          memory: 2Gi
```

Stannislav:
[...]

PR #4702, opened 2025-09-19
Support Multiple Storage URIs for InferenceServices

anurags25:
This PR implements support for multiple storage URIs in KServe InferenceServices, enabling users to download artifacts from multiple storage locations with configurable destination paths.

  ## Motivation

  Currently KServe only supports one storage URI and downloads files to `/mnt/models`. This limitation prevents use cases such as:
  - Base LLM models with separate LoRA adapters from different storage locations
  - Preprocessing artifacts alongside model files for collocated transformers
  - Models with dependencies stored in separate repositories for better versioning and reuse

  ## Changes

  ### API Extensions
  - Added new `StorageUri` struct with `uri` and `path` fields
  - Added `storageUris` array field to `PredictorSpec`, `TransformerSpec`, and `ExplainerSpec`
  - Maintains full backwards compatibility with existing `storageUri` field

  ### Implementation
  - **Container injection**: Storage initializer container is added as init container during reconciliation (requires enabling [init container feature flag](https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-init-containers) for knative deployments)

  ### Backwards Compatibility
  - Existing `storageUri` field continues to work unchanged
  - `storageUri` and `storageUris` are mutually exclusive (validation enforced)
  - Default behavior preserved when using single storage URI

  ### Example Usage
  ```yaml
  kind: InferenceService
  metadata:
    name: multi-storage-example
  spec:
    predictor:
      model:
        modelFormat:
          name: huggingface
      storageUris:
        - uri: s3://bucket/awesome-model
          path: /mnt/models/awesome-model
        - uri: s3://bucket/another-awesome-model
          path: /mnt/models/another-awesome-model
 ```

**Type of changes**

- Unit Tests: Added comprehensive unit tests covering multiple storage URI scenarios including:
    - Mixed PVC and cloud storage combinations
    - Multiple PVC URIs with different mount paths
    - Error handling for unsupported storage containers
    - Backwards compatibility with single storage URI
    - Container selection logic (kserve-container, worker-container fallbacks)
  - Integration Tests: Verified storage initializer injection and volume mounting:
    - Init container creation with multiple storage URIs
    - Proper volume mount configuration for different container types
    - Storage credential injection for cloud providers
    - PVC mounting alongside cloud storage
  - Backwards Compatibility Tests: Confirmed existing functionality remains unchanged:
    - Single storageUri field continues to work as before
    - Default mount path /mnt/models preserved for legacy usage
    - Environment variable injection for legacy mode
  - Context Parameter Fix: Added proper context handling to resolve linter warnings:
    - Updated CreateStorageSpecSecretEnvs function to accept context parameter
    - Propagated context through storage initialization pipeline
    - Updated all callers including reconcilers and test files

  Test Configuration:
  - Go unit tests using testify and gomega frameworks
  - Kubernetes fake client for API operation simulation
  - Mock storage configurations for various cloud providers
  - Test coverage includes both successful scenarios and error conditions
 
 
1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes. NA

**Checklist**:

**Release note**:

```release-note
Users can now specify multiple storage sources with custom mount paths using the new `storageUris` field. This enables use cases like base models with separate adapters, collocated transformers, and models with dependencies from different storage locations. The existing `storageUri` field remains fully supported for backwards compatibility.

  Example:
  ```yaml
  spec:
    predictor:
      storageUris:
        - uri: s3://bucket/base-model
          path: /mnt/models/base
        - uri: gs://bucket/adapter
          path: /mnt/models/adapter

  Note: For Knative deployments, ensure the init container feature flag is enabled.
```

**Re-running failed tests**


pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -71,15 +74,34 @@ func (e *Explainer) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
...
+	if sourceURI != nil {
 		annotations[constants.StorageInitializerSourceUriInternalAnnotationKey] = *sourceURI
 		err := isvcutils.ValidateStorageURI(ctx, sourceURI, e.client)
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("StorageURI not supported: %w", err)
 		}
 	}
+
+	if storageURIs != nil {
+		isvcutils.ValidateMultipleStorageURIsSpec(storageURIs)
```
spolti: I think it is better to do these validations in the isvc webhook.

pkg/apis/serving/v1alpha1/storage_container_types.go
```diff
@@ -34,6 +34,9 @@ type StorageContainerSpec struct {
 	SupportedUriFormats []SupportedUriFormat `json:"supportedUriFormats" validate:"required"`
 	// +kubebuilder:default="initContainer"
 	WorkloadType WorkloadType `json:"workloadType,omitempty"`
+
+	// +kubebuilder:default=false
+	SupportsMultiModelDownload *bool `json:"supportsMultiModelDownload,omitempty"`
```
spolti, line 39: Is there a need to use a flag to enable the feature?
I mean, if you set multiple models, it means that you want to use the feature.

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -35,6 +35,9 @@ type ExplainerSpec struct {
 	PodSpec `json:",inline"`
 	// Component extension defines the deployment configurations for explainer
 	ComponentExtensionSpec `json:",inline"`
+
+	// Spec for multiple storage uris.
+	StorageUris []StorageUri `json:"storageUris,omitempty"`
```
spolti, line 42: May we also consider also making the StorageSpec support multiple models?

spolti:
@anurags25 thanks for the PR, could you also take a look here: https://github.com/kserve/kserve/issues/3413 ?

There are some useful information about the some requirements.

pkg/apis/serving/v1alpha1/storage_container_types.go
anurags25: This is more of a check for scenarios where a custom storage initializer image is being used that the users do not manage themselves (for example, one supplied by a cloud provider).

pkg/controller/v1beta1/inferenceservice/components/explainer.go
anurags25: Fixed.

pkg/apis/serving/v1beta1/explainer.go
anurags25: Are we talking about a StorageSpec per model/blob? I’m hesitant to support that unless there’s a compelling use case; The storage logic is getting pretty complex. I’d rather KServe lean on well-defined patterns like service account based auth.

anurags25:
@spolti All the tests are passing and the change is ready to go, can you review when you get a chance?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -493,3 +497,103 @@ func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService
...
+
+	return nil
+}
+
+// ValidateMultipleStorageURISpecs validates a list of storage URI specifications.
+// It ensures that:
+// - Each individual URI specification is valid (non-empty URI, absolute path)
+// - All non-PVC paths share a common parent directory (not root)
+// - PVC paths are unique across the list
+// - The total number doesn't exceed the configured maximum
```
greenmoon55: nit: what does it mean? total number of uris?

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -58,29 +62,42 @@ type KsvcReconciler struct {
...
+	storageInitializerConfig *kserveTypes.StorageInitializerConfig,
+	storageSpec *v1beta1.StorageSpec,
+	credentialBuilder *credentials.CredentialBuilder,
+	storageContainerSpec *v1alpha1.StorageContainerSpec,
 ) *KsvcReconciler {
 	return &KsvcReconciler{
 		client:          client,
 		scheme:          scheme,
-		Service:         createKnativeService(componentMeta, componentExt, podSpec, componentStatus, disallowedLabelList),
+		Service:         createKnativeService(ctx, client, componentMeta, componentExt, podSpec, componentStatus, disallowedLabelList, storageUrisSpec, storageInitializerConfig, storageSpec, credentialBuilder, storageContainerSpec),
```
greenmoon55, line 82: I might be missing something. Why do we update knative and raw kube reconciler? I think the the injection is done in the mutating webhook.

pkg/controller/v1alpha1/llmisvc/workload_storage.go
```diff
@@ -113,7 +113,20 @@ func (r *LLMISVCReconciler) attachOciModelArtifact(modelUri string, podSpec *cor
 //
 // TODO: For now, this supports only direct mount. Copying from PVC would come later (if it makes sense at all).
 func (r *LLMISVCReconciler) attachPVCModelArtifact(modelUri string, podSpec *corev1.PodSpec) error {
```
greenmoon55, line 115: Is this code change required or can we move it to a separate PR?

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
anurags25: The new logic adds storage initializer as init container directly during reconciliation, so there's no need for pod mutation anymore (for the new code path).

pkg/controller/v1alpha1/llmisvc/workload_storage.go
anurags25: Code that handled storage was getting very complex so I refactored it a bit, which affects this; Without this LLMIsvc tests fail.

pkg/apis/serving/v1beta1/inference_service_validation.go
anurags25: Great catch! Typo leftover from one of the earlier changes; Fixed.

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
greenmoon55: but then for `storage_uri` we still do it in storage_initializer_injector? I don't know if there are better alternatives but this feels pretty messy.

greenmoon55: Can we check if knative allows init containers by default? Otherwise we might want to ask users to tweak this knative flag?

anurags25: Yes, storage_uri maintains the same behaviour as today; Not sure if there are better alternatives, I'm open to suggestions from the community.

Meanwhile, I've refactored the code to share the storage related logic as much as possible, ideally storage_uri field would be marked as legacy and deprecated in future releases to simplify this code.

anurags25: knative doesn't support init containers by default, using this with knative deployments requires setting a feature flag. It's called out in the [proposed documentation](https://github.com/kserve/website/pull/529/files#diff-cdaf3f4fdb3c04c067c4f79bc995bebcc736a20da87a29541cf63c8e13ad9b71R426) for this feature.

pkg/apis/serving/v1beta1/explainer.go
yuzisun: Agreed, storage spec was originally introduced mainly for model mesh and I think we should consider reducing the complexity for the two different ways of specifying storage.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -68,6 +72,18 @@ type PredictorSpec struct {
...
+type StorageUri struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Uri string `json:"uri"`
+
+	// +kubebuilder:default="/mnt/models"
+	// +kubebuilder:validation:Pattern="^/.*"
+	// +kubebuilder:validation:MaxLength=255
+	// +optional
+	Path string `json:"path"`
```
yuzisun: I think this is really mount path ? Let’s add comment explaining this field and consider rename to mountPath.

pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -71,15 +74,24 @@ func (e *Explainer) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 	annotations := utils.Filter(isvc.Annotations, func(key string) bool {
 		return !utils.Includes(e.inferenceServiceConfig.ServiceAnnotationDisallowedList, key)
 	})
+
+	sourceURI := explainer.GetStorageUri()
+	storageURIs := isvc.Spec.Explainer.StorageUris
+
+	if sourceURI != nil && storageURIs != nil {
+		return ctrl.Result{}, errors.New("Setting both StorageURI and StorageURIs is not supported.")
```
yuzisun: This should be validated in the webhook so user gets error back.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -116,10 +118,20 @@ func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 
 	predictor := isvc.Spec.Predictor.GetImplementation()
 
+	sourceURI := predictor.GetStorageUri()
+	storageURIs := isvc.Spec.Predictor.StorageUris
+
+	if sourceURI != nil && storageURIs != nil {
```
yuzisun: Same here move the validation logic to webhook.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -102,213 +178,381 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
-		}
-		if pvcName, ok := pod.ObjectMeta.Annotations[constants.LocalModelPVCNameAnnotationKey]; ok {
-			srcURI = "pvc://" + pvcName + "/models/" + modelName + subPath
-		} else {
-			return fmt.Errorf("Annotation %s not found", constants.LocalModelPVCNameAnnotationKey)
+	// Handle multiple storage URIs (new functionality)
+	if len(params.StorageURIs) > 0 && !params.IsLegacyURI {
+		// Validate that the storage container supports multiple downloads
+		if params.StorageContainerSpec != nil && !*params.StorageContainerSpec.SupportsMultiModelDownload {
+			return fmt.Errorf("storage container %q does not support multi-model download; use the default storage initializer or a compatible storage container",
```
yuzisun: Default storage container means kserve provided storage container right ?

anurags25: Yes, updated the error to be more specific.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -54,6 +54,10 @@ type PredictorSpec struct {
 	// Model spec for any arbitrary framework.
 	Model *ModelSpec `json:"model,omitempty"`
 
+	// Spec for multiple storage uris.
+	// +listType=atomic
```
greenmoon55, line 58: nit:
```suggestion
	// +listType=atomic
    // +kubebuilder:validation:MinItems=1
```
to make sure users cannot set an empty list

pkg/apis/serving/v1beta1/explainer.go
```diff
@@ -35,6 +35,10 @@ type ExplainerSpec struct {
 	PodSpec `json:",inline"`
 	// Component extension defines the deployment configurations for explainer
 	ComponentExtensionSpec `json:",inline"`
+
+	// Spec for multiple storage uris.
+	// +listType=atomic
```
greenmoon55, line 40:
```suggestion
	// +listType=atomic
	// +kubebuilder:validation:MinItems=1
```

pkg/apis/serving/v1beta1/predictor.go
greenmoon55, line 58:
```suggestion
	// +listType=atomic
	// +kubebuilder:validation:MinItems=1
```

pkg/apis/serving/v1beta1/transformer.go
```diff
@@ -26,6 +26,10 @@ type TransformerSpec struct {
 	PodSpec `json:",inline"`
 	// Component extension defines the deployment configurations for a transformer
 	ComponentExtensionSpec `json:",inline"`
+
+	// Spec for multiple storage uris.
+	// +listType=atomic
```
greenmoon55, line 31:
```suggestion
	// +listType=atomic
	// +kubebuilder:validation:MinItems=1
```

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -50,7 +51,71 @@ type StorageInitializerInjector struct {
...
+//   - storageUri: The storage URI to find a compatible container for (e.g., "s3://bucket/path")
+//   - client: Kubernetes client for listing ClusterStorageContainer resources
+//
+// Returns:
+//   - *v1alpha1.StorageContainerSpec: The container specification that supports the URI, or nil if none found
+//   - error: Error if the Kubernetes API call fails or URI format checking fails
+//
+// The function iterates through ClusterStorageContainer resources and uses their
+// IsStorageUriSupported method to determine compatibility with the provided URI.
+func GetStorageSpecForUri(ctx context.Context, storageUri string, client client.Client) (*v1alpha1.StorageContainerSpec, error) {
```
greenmoon55: Storage spec is another way to specify storage uri. It's different from Storage Container Spec. Can we rename it to `GetStorageContainerSpec` or sth like that?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -102,213 +178,381 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
-		initContainer = utils.AddStorageInitializerContainer(&pod.Spec, transformerContainer.Name, srcURI, isvcReadonlyStringFlag, mi.config)
-	}
+			// Apply volume mount to all relevant containers
+			for _, containerName := range mountContainerNames {
+				if mountErr := utils.AddModelMount(storageMountParams, containerName, params.PodSpec); mountErr != nil {
+					return fmt.Errorf("failed to add volume mount for container %q: %w", containerName, mountErr)
+				}
+			}
+		}
+	} else if len(params.StorageURIs) == 1 {
```
greenmoon55, line 343: add a comment saying that this is for the old single storageuri only

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -102,213 +178,381 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
+//
+// The function supports:
+//   - Multiple storage URIs with custom mount paths for complex use cases (e.g., base model + LoRA adapters)
+//   - Backward compatibility with single storage URI scenarios
+//   - Mixed PVC and cloud storage URI handling
+//   - Automatic volume mounting and path management
+//   - Storage container specification for custom downloaders
+//
+// Parameters:
+//   - params: A StorageInitializerParams struct containing all necessary configuration including
```
greenmoon55, line 204: There are quite a few variables in `params`. Do we only update `PodSpec` in `params`? Can we add a comment about which ones could be updated?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -102,213 +178,381 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
+			initContainer.VolumeMounts = append(initContainer.VolumeMounts, caBundleVolumeMount)
 		}
 
-		caBundleVolumeMount := corev1.VolumeMount{
-			Name:      CaBundleVolumeName,
-			MountPath: caBundleVolumeMountPath,
-			ReadOnly:  true,
+		// Update initContainer (container spec) from a storage container CR if there is a match,
+		// otherwise initContainer is not updated.
+		// Priority: CR > configMap
```
greenmoon55:
```suggestion
```
Can remove this comment since matching storage container CR is moved to the controller

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -102,213 +178,381 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
+}
 
-		pod.Spec.Volumes = append(pod.Spec.Volumes, caBundleVolume)
-		initContainer.VolumeMounts = append(initContainer.VolumeMounts, caBundleVolumeMount)
+// InjectStorageInitializer injects an init container to provision model data
+// for the serving container in a unified way across storage tech by injecting
+// a provisioning INIT container. This is a workaround because Knative does not
+// support INIT containers: https://github.com/knative/serving/issues/4307
+func (mi *StorageInitializerInjector) InjectStorageInitializer(ctx context.Context, pod *corev1.Pod) error {
+	// Only inject if the required annotations are set
```
greenmoon55, line 504: nit: add a comment that this is only for the old single storage uri case and this is moved to the controller for `storageUris`

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -58,7 +59,13 @@
 
 class Storage(object):
     @staticmethod
-    def download(uri: str, out_dir: str = None) -> str:
+    def download_files(source_uris: list[str], out_dirs: list[str]) -> list[str]:
```
greenmoon55, line 62: Have you tested this? Like build a image and test it out? We add can an e2e test in the future.

pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -71,15 +74,19 @@ func (e *Explainer) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 	annotations := utils.Filter(isvc.Annotations, func(key string) bool {
 		return !utils.Includes(e.inferenceServiceConfig.ServiceAnnotationDisallowedList, key)
 	})
+
+	sourceURI := explainer.GetStorageUri()
+
 	// Knative does not support INIT containers or mounting, so we add annotations that trigger the
 	// StorageInitializer injector to mutate the underlying deployment to provision model data
```
greenmoon55, line 81: maybe update it to say this feature has to be enabled in knative
```suggestion
```

pkg/controller/v1beta1/inferenceservice/reconcilers/knative/ksvc_reconciler.go
```diff
@@ -150,6 +167,34 @@ func createKnativeService(
...
 	})
 
+	if storageUris != nil && len(*storageUris) > 0 {
+		isvcReadonlyStringFlag := true
+		isvcReadonlyString, ok := annotations[constants.StorageReadonlyAnnotationKey]
+		if ok {
+			if isvcReadonlyString == "false" {
+				isvcReadonlyStringFlag = false
+			}
+		}
```
greenmoon55: nit: It's simple but it's repeated in three places. would it be better to have a util function?

greenmoon55 (commented):
Overall LGTM. I don't know a lot about `llmsvc`.

python/storage/kserve_storage/kserve_storage.py
anurags25: Yes, built the image and tested this locally with kind.

yuzisun (approved):

PR #4707, opened 2025-09-23
Parallelize blob downloads from azure in storage-initializer

agoston:

Parallelize blob downloads from azure in storage-initializer.

- No more loading entire files into memory with readall()
- Multiple files download concurrently
- Large files are processed in chunks

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: N/A

**Type of changes**

Deployed kserve on azure. Checked the logs and the download times. Noted that download was successful, and for a specific model, download time went from 349s to 96s.

N/A

**Checklist**:

**Release note**:
```release-note
NONE
```

**Re-running failed tests**


python/storage/kserve_storage/kserve_storage.py
```diff
@@ -55,6 +56,13 @@
 _HDFS_SECRET_DIRECTORY = "/var/secrets/kserve-hdfscreds"
 _HDFS_FILE_SECRETS = ["KERBEROS_KEYTAB", "TLS_CERT", "TLS_KEY", "TLS_CA"]
 
+# Azure async download configuration
+_AZURE_MAX_FILE_CONCURRENCY = int(os.getenv("AZURE_MAX_FILE_CONCURRENCY", "4"))
```
lizzzcai, line 60: I assumed currently, there is no way to adjust these parameters like passing the env to storage initializer?

agoston: This is more of a helping hand than a full-blown solution.

My idea was that these values likely won't need to be changed. And since storage-initializer has no configuration, I did not want to add the complexity just for this.

Can you live with this?

agoston:
Sorry, had a big fight with DCO/sign-off, got into a nasty merge issue - closing this PR, opening a new one.

python/storage/kserve_storage/kserve_storage.py
lizzzcai: Hi @agoston , probably user can customize it via the [ClusterStorageContainer](https://kserve.github.io/website/docs/model-serving/storage/storage-containers). Overall it is good to me, thanks.

agoston: Thanks @lizzzcai, I didn't know about ClusterStorageContainer.
Can you approve https://github.com/kserve/kserve/pull/4709 then? I moved it over there because of a probably rushed decision when finding myself in a "github entanglement".

PR #4727, opened 2025-10-01
Add Support for Configuring S3 Storage via Secret Data

brettmthompson:
This PR aims to accomplish 3 goals:
1. Extend support for s3 configurations to allow values to be defined via annotations or in secret data
2. Standardize the way env variables are set for s3 storage

To extend support for setting s3 configurations via secret data, the logic has been updated to search for relevant s3 related values in the sa/secret annotations, the secret data (if present) and the s3 config defined in the `inferenceservice-config` configmap. The hierarchy of this lookup is annotations -> secret data -> s3 config. We will support any combination of the 3 for all values, but the hierarchical search will always take precedent. This means that a combination of values defined in the annotations, secret data, and s3 config is allowed. The keys within the secret data reflect the associated env variable name.

To standardize the way env variables are set for s3 storage, all env variables will now only be set in the container if their corresponding value is a non-empty string. Currently all values follow the same logic and do not support empty strings, except for `S3_ENDPOINT` and  `S3_USE_HTTPS. S3_ENDPOINT`. `S3_ENDPOINT` will currently be populated in the container env as an empty strings if its corresponding annotation is set with a value of an empty string. This will in turn cause the `AWS_ENDPOINT_URL` to be set to only the `http://` or `https://` prefix, with no actual endpoint. The same thing applies for `S3_USE_HTTPS`  when set as an annotation; if the value of the annotation is an empty string this will be set in the container env. On the other hand, if using the `S3_USE_HTTPS`  value defined in the s3 config it will only be set in the container env if the value is 0. 
The changes in this PR standardize the approach for all env variables. Regardless of where their value is set (annotation, secret data, or s3 config) they will always be set in the container env unless the value is an empty string. 

**Type of changes**

- Added additional Unit tests
- Manually tested s3 storage usng both annotations and secret data

**Checklist**:

**Release note**:

```release-note
Support for s3 storage configurations in secret data
```

**Re-running failed tests**


pierDipi (approved):
/lgtm

pkg/credentials/s3/utils.go
```diff
@@ -18,124 +18,138 @@ package s3
...
-		if s3UseHttps, ok := annotations[InferenceServiceS3SecretHttpsAnnotation]; ok {
-			if s3UseHttps == "0" {
-				s3EndpointUrl = "http://" + annotations[InferenceServiceS3SecretEndpointAnnotation]
-			}
-			envs = append(envs, corev1.EnvVar{
-				Name:  S3UseHttps,
-				Value: s3UseHttps,
-			})
-		}
+	s3UseHttps := getEnvValue(annotations, secretData, InferenceServiceS3SecretHttpsAnnotation, S3UseHttps, s3Config.S3UseHttps)
```
spolti, line 35: shouldn't we add a validation here?

pkg/credentials/s3/utils.go
```diff
@@ -18,124 +18,138 @@ package s3
...
 		})
 	}
 
-	// For each variable, prefer the value from the annotation, otherwise default to the value from the inferenceservice configmap if set.
-	verifySsl, ok := annotations[InferenceServiceS3SecretSSLAnnotation]
-	if !ok {
-		verifySsl = s3Config.S3VerifySSL
-	}
-	if verifySsl != "" {
+	s3VerifySSL := getEnvValue(annotations, secretData, InferenceServiceS3SecretSSLAnnotation, S3VerifySSL, s3Config.S3VerifySSL)
```
spolti, line 59: For boolean values like these, isn't it better to parse to bool?

pkg/credentials/s3/utils_test.go
```diff
@@ -99,9 +100,249 @@ func TestBuildS3EnvVars(t *testing.T) {
...
+					Value: "value",
+				},
+			},
+		},
+		"AnnotationsSecretAndS3Config": {
+			annotations: map[string]string{
+				InferenceServiceS3SecretEndpointAnnotation:    "s3.aws.com-annotation",
+				InferenceServiceS3SecretRegionAnnotation:      "us-east-2-annotation",
+				InferenceServiceS3SecretSSLAnnotation:         "0-annotation",
+				InferenceServiceS3UseVirtualBucketAnnotation:  "true-annotation",
```
spolti, line 218: IMHO, we shouldn't accept values like this for boolean-like configuration

brettmthompson:
@spolti I agree with you that adding validation regarding the values configured for s3 storage (whether in the annotation or secret data) would be good to have. Given that there is no validation in place currently and this PR is just to expand the allowed configuration to support secret data would it make sense to add this validation in a separate PR as a feature?

spolti (approved):
/lgtm

yuzisun (approved):

PR #4728, opened 2025-10-01
Injecting CA Bundle Into Storage Initializer Container for S3 Storage on LLMISVC Reconciliation

brettmthompson:
This PR injects the ca bundle into the storage-init container for llmisvcs utilizing s3 based storage connections. The ca bundle is injected based on the configuration defined in the service account, secret annotations, or in the inferenceservice-config configmap.

**Type of changes**

- [x] Added unit tests
- [x] Validated the presence of ca bundle in storage-init container via manual testing

- Logs

**Checklist**:

**Release note**:

```release-note
CA bundle injection in storage-init container for LLMISVCs with S3 storage
```

**Re-running failed tests**

### DIFF
--- a/pkg/controller/v1alpha1/llmisvc/workload_storage.go
+++ b/pkg/controller/v1alpha1/llmisvc/workload_storage.go
@@ -139,23 +139,18 @@ func (r *LLMISVCReconciler) attachPVCModelArtifact(modelUri string, podSpec *cor
 // Returns:
 //
 //	An error if the configuration fails, otherwise nil.
-func (r *LLMISVCReconciler) attachS3ModelArtifact(ctx context.Context, serviceAccount *corev1.ServiceAccount, llmSvc *v1alpha1.LLMInferenceService, modelUri string, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig, credentialConfig *credentials.CredentialConfig) (s3ArtifactError error) {
+func (r *LLMISVCReconciler) attachS3ModelArtifact(ctx context.Context, serviceAccount *corev1.ServiceAccount, llmSvc *v1alpha1.LLMInferenceService, modelUri string, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig, credentialConfig *credentials.CredentialConfig) error {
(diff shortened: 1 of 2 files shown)


pkg/controller/v1alpha1/llmisvc/workload_storage.go
```diff
@@ -135,11 +139,17 @@ func (r *LLMISVCReconciler) attachPVCModelArtifact(modelUri string, podSpec *cor
...
 	if err := r.attachStorageInitializer(modelUri, podSpec, storageConfig); err != nil {
 		return err
 	}
 	if initContainer := utils.GetInitContainerWithName(podSpec, constants.StorageInitializerContainerName); initContainer != nil {
+		// Inject CA bundle if applicable
+		defer func() {
+			if s3ArtifactError == nil {
+				injectCaBundle(llmSvc.Namespace, podSpec, initContainer, storageConfig)
+			}
+		}()
```
pierDipi: I didn't get why this a defer block vs being a normal call as part of the normal flow

brettmthompson: This injection needs to be performed in 2 places. Prior to exiting if the default service account does not exist in the namespace [here](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/llmisvc/workload_storage.go#L149) or after the secret volume and env are created with no error [here](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/llmisvc/workload_storage.go#L162).

This defer block handles both those scenarios. If it would make for more readable code I can separate this into two distinct function calls.

brettmthompson: The reason for using the defer block was to make this behavior explicit. If an init container exists, then the ca bundle should always be injected if the parent method does not return an error.

brettmthompson: Removed the defer in the latest commit. Now explicitly calling `injectCaBundle` where required

pierDipi (approved):
/lgtm
/approve

spolti (approved):
/lgtm

yuzisun (approved):

Issue #3248, opened 2023-11-14
Inference Graph Sequence Req/Res Problem


satyajitghana:
https://github.com/kserve/kserve/blob/bd5a5d9dfeb58daf76e87b79785a5bf68966c6f9/cmd/router/main.go#L211C2-L213

I think there's a problem in this line of code

```
if step.Data == "$response" && i > 0 {
				request = responseBytes
			}
```

so basically what it is trying to do is chain response and send it as a request

but the thing is,

the inputs request looks like `{"instances": [ stuff here ]}` and the output is `{"predictions": [outputs here]}`

now what's happening is `request = responseBytes` simply copies response and sends it as a request

so the request now is `{"predictions": [outputs here]}`

and i think the KServe Requesst Envelope will simply discard this key, and the preproces function will get nothing as input and nothing works. sad :(

satyajitghana:
```
{
    "detail": [
        {
            "loc": [
                "body",
                "inputs"
            ],
            "msg": "field required",
            "type": "value_error.missing"
        }
    ]
}
```

I tried on V1 and V2, both are not working, thinking maybe the code handles V2 differently, but no.

I don't see Sequence working at all, there's no use case, if req and res cannot be chained. The Cat Dog Example simply Takes the `$request` again, which is okay, but you would want to reuse outputs in Sequence. Person Detection -> Crops -> Face Detection -> Face Recognition, this kind of chain is not possible as of now.

satyajitghana:
One option is to modify KServe's Router, or another option is to fix KServe Request Envelope, https://github.com/pytorch/serve/blob/master/ts/torch_handler/request_envelope/kservev2.py , but i think the former is better.

rachitchauhan43:
> i think the KServe Requesst Envelope will simply discard this key

@satyajitghana : This can be an issue with V2 as V2 is strict in terms of input format. What issue did you see with V1 protocol on your inference service side ?

satyajitghana:
In V1, the output is `{"predictions": [outputs here]}` which is passed on to the next step in `Sequence`, and in KServe V1 Envelope

```
def parse_input(self, data):
        self._data_list = [row.get("data") or row.get("body") for row in data]
        # selecting the first input from the list torchserve creates
        logger.debug("Parse input data_list %s", self._data_list)
        data = self._data_list[0]

        # If the KF Transformer and Explainer sends in data as bytesarray
        if isinstance(data, (bytes, bytearray)):

            data = data.decode()
[...]

PR #4738, opened 2025-10-08
Allow mapping predictions to instances in sequential inference graphs

tmbochenski:

This PR addresses [issue #3248](https://github.com/kserve/kserve/issues/3248), which highlights a limitation when chaining `InferenceServices` that use the V1 Inference Protocol. Specifically, the output of an upstream service is returned under the `"predictions"` key, while a downstream service expects input under the `"instances"` key. This mismatch prevents straightforward chaining of V1-based services in an inference graph.

To resolve this, the PR introduces a new optional property:

### `MapPredictionsToInstances` (boolean)

When set to `true`, the router will:
- Extract the value from the `"predictions"` field of the upstream response.
- Map it to the `"instances"` field in the request body sent to the downstream service.

Fixes #

[issue #3248](https://github.com/kserve/kserve/issues/3248)

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -316,6 +316,13 @@ type InferenceStep struct {
 	// +optional
 	Data string `json:"data,omitempty"`
 
+	// If true, maps the 'predictions' field from the previous step's response
+	// to the 'instances' field of this step's request.
+	// Useful in sequential inference graphs where one step's output
+	// becomes the input for the next.
+	// +optional
+	MapPredictionsToInstances bool `json:"mapPredictionsToInstances,omitempty"`
```
yuzisun, line 324: Did you try the `Data` field with `$response.predictions` ? that might achieve the same.

cmd/router/main.go
```diff
@@ -323,6 +328,19 @@ func routeStep(nodeName string, graph v1alpha1.InferenceGraphSpec, input []byte,
...
+			if step.MapPredictionsToInstances {
+				decoded := sequenceReqRes{}
+				err = json.Unmarshal(request, &decoded)
+
+				// If unmarshaling succeeds and Predictions is non-empty,
+				// move Predictions to Instances and re-marshal the request.
+				if err == nil && len(decoded.Predictions) > 0 {
+					decoded.Instances = decoded.Predictions
+					decoded.Predictions = []interface{}{}
+					request, _ = json.Marshal(decoded) // TODO check if you need err handling for Marshalling
```
yuzisun, line 340: can you handle the error ? The predictions key should also be removed.

tmbochenski: I didn’t handle it on purpose, to stay consistent with the rest of the `json.Marshal` calls in this file, see for example:

```go
stepResponse, _ := json.Marshal(ensembleStepOutput.StepResponse) // TODO check if you need err handling for marshalling
```

Regarding the `predictions` key - it’s already removed. The line:

```go
decoded.Predictions = []interface{}{}
```

clears the property, and because of:

```go
Predictions []interface{} `json:"predictions,omitempty"`
```

empty fields are omitted when encoding to JSON.

pkg/apis/serving/v1alpha1/inference_graph.go
tmbochenski: I checked the KServe source code, and `$response.predictions` isn’t actually supported.  
The `data` field is treated as a plain string, and only `$request` is handled explicitly.  

The only place the `data` field is checked is here:

```go
request := input
if step.Data == "$response" && i > 0 {
    request = responseBytes
}
```

If `data` is anything other than `$response` (e.g., `$response.predictions`), the `$request` data are actually passed to the next step.  
So `$response.predictions` wouldn’t work as expected, the full `$response` can be passed, but not sub-fields.

yuzisun (approved):

PR #4765, opened 2025-10-22
Fix the issue of Invalid value for env[1].valueFrom

### LINKED ISSUE #4761: spec.initContainers[0].env[1].valueFrom: Invalid value: ""
/kind bug

When I create the inference service, it failed to launch the pod. The deployment is there, but not the pod.
The deployment resource gave me the following error:
```
[spec.initContainers[0].env[1].valueFrom: Invalid value: "": may not be specified
      when `value` is not empty, spec.initContainers[0].env[2].valueFrom: Invalid
      value: "": may not be specified when `value` is not empty]
```

This error seems to indicate conflicting value and valueFrom fields set in the container env.
The error was introduced by this PR: https://github.com/kserve/kserve/pull/4726

If we remove the section 
```
			Env: []corev1.EnvVar{
				{
					Name:  "HF_HUB_ENABLE_HF_TRANSFER",
					Value: "1",
				},
				{
					Name:  "HF_XET_HIGH_PERFORMANCE",
					Value: "1",
				},
				{
					Name:  "HF_XET_NUM_CONCURRENT_RANGE_GETS",
					Value: "8",
				},
			},
```
from the file pkg/utils/storage.go, the issue is gone.

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

houshengbo:

Fixes #4761

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -488,6 +488,11 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
 				return err
 			}
 		}
+
+		// Add default HuggingFace optimization environment variables if they don't already exist.
+		// This is done after merging to avoid conflicts with user-defined environment variables.
+		// See https://github.com/kserve/kserve/issues/4761
+		utils.AddDefaultHuggingFaceEnvVars(initContainer)
 	}
 
 	return nil
--- a/pkg/webhook/admission/pod/storage_initializer_injector.go
+++ b/pkg/webhook/admission/pod/storage_initializer_injector.go
@@ -430,13 +430,18 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
 		}
 
 		// Inject CA bundle configMap if caBundleConfigMapName or constants.DefaultGlobalCaBundleConfigMapName annotation is set
-		caBundleConfigMapName := params.Config.CaBundleConfigMapName
-		if ok := needCaBundleMount(caBundleConfigMapName, initContainer); ok {
+		// Store the CA bundle configuration to be applied after merge to avoid conflicts
+		var caBundleConfigMapName string
+		var caBundleVolumeMountPath string
+		var needsCABundle bool
+		if ok := needCaBundleMount(params.Config.CaBundleConfigMapName, initContainer); ok {
+			needsCABundle = true
+			caBundleConfigMapName = params.Config.CaBundleConfigMapName
 			if params.Namespace != constants.KServeNamespace {
 				caBundleConfigMapName = constants.DefaultGlobalCaBundleConfigMapName
 			}
 
-			caBundleVolumeMountPath := params.Config.CaBundleVolumeMountPath
+			caBundleVolumeMountPath = params.Config.CaBundleVolumeMountPath
 			if caBundleVolumeMountPath == "" {
 				caBundleVolumeMountPath = constants.DefaultCaBundleVolumeMountPath
 			}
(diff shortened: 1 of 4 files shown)


pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -488,6 +488,11 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
 				return err
 			}
 		}
+
+		// Add default HuggingFace optimization environment variables if they don't already exist.
+		// This is done after merging to avoid conflicts with user-defined environment variables.
```
brettmthompson, line 518: So root cause for https://github.com/kserve/kserve/issues/4761 is when non-default user-defined env variables are configured for `HF_HUB_ENABLE_HF_TRANSFER` , `HF_XET_HIGH_PERFORMANCE`, or `HF_XET_NUM_CONCURRENT_RANGE_GETS`. 

This is a good catch. I think this vulnerability would have been present even when only `HF_TRANSFER` was set in [hf_secret.go](https://github.com/kserve/kserve/blob/release-0.15/pkg/credentials/hf/hf_secret.go#L38).

This raises the question - should we also apply the logic to setting env variables for other storage types as well, for example when setting env variables for [s3 storage](https://github.com/kserve/kserve/blob/release-0.15/pkg/credentials/s3/utils.go#L32)?

houshengbo: I think pkg/webhook/admission/pod/storage_initializer_injector.go has general functions to call by all type of storage. Or does each type of storage have its implementation?

brettmthompson: If a storage spec is utilized then a [generic function](https://github.com/kserve/kserve/blob/master/pkg/credentials/service_account_credentials.go#L101) is used for all storage types to create the container env.

On the other, if a storage spec is not utilized then the container envs are created for each storage type via a different [implementation.](https://github.com/kserve/kserve/blob/master/pkg/credentials/service_account_credentials.go#L271) . These differing implementations can be found in the subfolders in this [directory](https://github.com/kserve/kserve/tree/master/pkg/credentials).

houshengbo: @brettmthompson I investigated this thoroughly and here's what I found:

### S3 and Other Storage Credentials Are Safe 

The good news is that S3, GCS, Azure, and other storage credential environment variables do **not** have the same vulnerability. Here's why:

1. **Credential env vars use `ValueFrom`**: 
   - Variables like `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` are injected using `ValueFrom.SecretKeyRef`
   - These are added by `CreateSecretVolumeAndEnv()` in `service_account_credentials.go` (lines 404-430 of storage_initializer_injector.go)
   - Since they use `ValueFrom`, even if a user tries to override them, the strategic merge patch won't create a Value/ValueFrom conflict.

2. **S3 configuration env vars are conditionally added**:
   - Variables like `S3_ENDPOINT`, `S3_USE_HTTPS`, `S3_VERIFY_SSL`, etc. are built by `BuildS3EnvVars()` in `pkg/credentials/s3/utils.go`
   - Each var is only added **if it has a non-empty value** (see lines 36-115 in s3/utils.go)
   - They all use `Value` (not `ValueFrom`), so if a user tries to set the same var with `ValueFrom`, it would fail - but this is unlikely since users typically use storage secrets

3. **These env vars are appended after init container creation**:
   - They're added **before** the `mergeContainerSpecs` call 
   - But since credentials use `ValueFrom` and configs use `Value`, there's no same-field conflict

### The HF Env Vars Were a Unique Case

The HF optimization env vars (`HF_HUB_ENABLE_HF_TRANSFER`, `HF_XET_HIGH_PERFORMANCE`, `HF_XET_NUM_CONCURRENT_RANGE_GETS`) were unique because they:
- Were **hardcoded with `Value = "1"`** inside `CreateInitContainerWithConfig()`
- Were meant to be **generic optimizations** (not credentials)
- Could reasonably be customized by users (e.g., disabling transfer acceleration, using different values from ConfigMaps)

### Potential Edge Case: CA Bundle Env Vars ⚠️

There is **one other place** that could theoretically have the same issue - the CA bundle env vars at lines 453-461:

```go
initContainer.Env = append(initContainer.Env, corev1.EnvVar{
    Name:  constants.CaBundleConfigMapNameEnvVarKey,
    Value: caBundleConfigMapName,
})
initContainer.Env = append(initContainer.Env, corev1.EnvVar{
    Name:  constants.CaBundleVolumeMountPathEnvVarKey,
    Value: caBundleVolumeMountPath,
})
```


If we want to be extra cautious, we could also apply the same defensive pattern to the CA bundle env vars, but it's not as critical since those are rarely customized by users.

I can apply the same pattern to CA bundle env vars as a preventive measure.

brettmthompson: Thank you for the thorough investigation @houshengbo

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -481,13 +514,10 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
...
-		if params.StorageContainerSpec != nil {
-			err := mergeContainerSpecs(initContainer, &params.StorageContainerSpec.Container)
-			if err != nil {
-				return err
-			}
-		}
+		// Add default HuggingFace optimization environment variables if they don't already exist.
+		// This is done after merging to avoid conflicts with user-defined environment variables.
+		// See https://github.com/kserve/kserve/issues/4761
+		utils.AddDefaultHuggingFaceEnvVars(initContainer)
```
brettmthompson, line 520: Would it be beneficial to only add the hugging face related values to the container env if the storage uris set in the  `StorageInitializerParams` input param contain a hugging face uri?

brettmthompson:
@houshengbo the [changes](https://github.com/kserve/kserve/pull/4765/commits/fb51bf4d8ba03f8b90193970095df51f5b1941e6) for ca bundle env vars would also apply to the logic used for llmisvcs [here](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha1/llmisvc/workload_storage.go#L264). As I don't want to put more work on your plate, I can make the changes for llmisvcs in a separate PR.

pkg/webhook/admission/pod/storage_initializer_injector.go
houshengbo: I think keeping them unconditional is better because: 
(1) they're harmless when unused.
(2) HF libraries only activate for hf:// URIs anyway.
(3) it keeps the code simpler. If needed, users can override them through the storage container spec.

houshengbo:
@brettmthompson You're right that the CA bundle env var logic in `pkg/controller/v1alpha1/llmisvc/workload_storage.go` would benefit from the same defensive pattern.

Thanks for the thorough review!

houshengbo:
https://github.com/kserve/kserve/pull/4777/files i opened another pr as well.

brettmthompson:
LGTM

yuzisun (approved):

PR #4769, opened 2025-10-22
feat: add vLLM as a supported runtime

yashaswipiplani:
Added vLLM as a separate supported runtime. This is the first PR in separating vLLM and HuggingFace transformers runtimes. Will raise a follow up PR for removing vLLM code from HuggingFaceServer. Created issue to track the same: https://github.com/kserve/kserve/issues/4770

**Type of changes**

**Single Node GPU Runtime**
ISVC applied:
```
kubectl apply -n kserve-test -f - <<EOF
kind: "InferenceService"
metadata:
  name: "qwen-llm"
  namespace: kserve-test
spec:
  predictor:
    model:
      modelFormat:
        name: vllm
      args:
        - --served-model-name=qwen
      storageUri: "hf://Qwen/Qwen2.5-0.5B-Instruct"
      resources:
        limits:
          cpu: "2"
          memory: 6Gi
          nvidia.com/gpu: "1"
        requests:
          cpu: "1"
          memory: 4Gi
          nvidia.com/gpu: "1"
EOF
```
**Single Node CPU Runtime**
ISVC applied:
```
kubectl apply -n kserve-test -f - <<EOF
kind: "InferenceService"
metadata:
  name: "qwen-llm"
  namespace: kserve-test
spec:
  predictor:
    model:
      modelFormat:
        name: vllm
      args:
        - --served-model-name=qwen
        - --max-model-len=4096
      storageUri: "hf://Qwen/Qwen2.5-0.5B-Instruct"
      resources:
        limits:
          cpu: "16"
          memory: 20Gi
        requests:
          cpu: "16"
          memory: 20Gi
EOF
```

**Inference tested for both:**
Screenshot 2025-10-22 at 1 59 37 PM
Screenshot 2025-10-22 at 1 59 49 PM

**Multi node inferencing**
ISVC Applied:
```
kubectl apply -n kserve-test -f - <<EOF
kind: "InferenceService"
metadata:
  name: "qwen-llm"
  namespace: kserve-test
  annotations:
    serving.kserve.io/autoscalerClass: "none"
spec:
  predictor:
    model:
      modelFormat:
        name: vllm
      args:
        - --served-model-name=qwen
      storageUri: "pvc://qwen-model-pvc/Qwen2.5-0.5B-Instruct"
      resources:
        limits:
          cpu: "2"
          memory: 6Gi
          nvidia.com/gpu: "1"
        requests:
          cpu: "1"
          memory: 4Gi
          nvidia.com/gpu: "1"
    workerSpec:
      tensorParallelSize: 1
      pipelineParallelSize: 2
      resources:
        limits:
          cpu: "2"
          memory: 6Gi
          nvidia.com/gpu: "1"
        requests:
          cpu: "1"
          memory: 4Gi
          nvidia.com/gpu: "1"
EOF
```

**Inference testing:**
Screenshot 2025-10-19 at 1 15 15 PM
Screenshot 2025-10-19 at 1 15 36 PM

**Checklist**:

**Release note**:
```release-note

```

**Re-running failed tests**


gavrissh:
sign-off your PR with these steps - https://github.com/kserve/kserve/pull/4769/checks?check_run_id=53384248700

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,323 @@ spec:
...
+  annotations:
+    prometheus.kserve.io/port: '8000'
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
+      version: "1"
+      autoSelect: true
+      priority: 1
+  protocolVersions:
+    - v1
```
gavrissh: vLLM runtime will not support v1 protocol

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,323 @@ spec:
...
+          then
+            export MODEL=${MODEL_ID}
+          fi
+          
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@
+        - "bash"
+      env:
+        - name: LMCACHE_USE_EXPERIMENTAL
+          value: "{{ .Values.kserve.servingruntime.vllmserver.lmcacheUseExperimental }}"
+        - name: VLLM_GPU_IMAGE
```
gavrissh, line 758: Could you describe what this change does in the description

test/e2e/predictor/test_vllm.py
```diff
@@ -0,0 +1,745 @@
...
+            args=[
+                "--served-model-name",
+                "vllm-qwen-chat",
+                "--max-model-len",
+                "512",
+                "--dtype",
+                "bfloat16",
+            ],
+            resources=V1ResourceRequirements(
+                requests={"cpu": "2", "memory": "8Gi", "nvidia.com/gpu": "1"},
```
gavrissh: There are no GPUs in GitHub workers; we can run only on CPU tests

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: Removed

test/e2e/predictor/test_vllm.py
yashaswipiplani: Removed GPU

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: Added the description, please review now. Thanks

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -378,6 +378,11 @@ func UpdateImageTag(container *corev1.Container, runtimeVersion *string, serving
 				if !strings.HasSuffix(container.Image, "-gpu") {
 					container.Image = image + "-gpu"
 				}
+			} else if servingRuntime != nil && *servingRuntime == constants.VLLMServer {
+				// For vLLM server, use the GPU image from environment variable if GPU Enabled
+				if image, exists := utils.GetEnvVarValue(container.Env, "VLLM_GPU_IMAGE"); exists {
+					container.Image = image
```
spolti, line 384: I am not sure if it is a good approach; ideally, we need to have VLLM CPU and GPU-specific runtimes, since the parameters and configurations aren't the same.

test/e2e/predictor/test_vllm.py
```diff
@@ -0,0 +1,745 @@
...
+    predictor = V1beta1PredictorSpec(
+        min_replicas=1,
+        model=V1beta1ModelSpec(
+            model_format=V1beta1ModelFormat(
+                name="vllm",
+            ),
+            env=[
+                client.V1EnvVar(
+                    name="MODEL_ID",
+                    value="Qwen/Qwen2-0.5B-Instruct",
```
spolti, line 59: This model size is around 1.2 GB. Having to download it for every test is not a good thing. Can we use the tiniest model available for VLLM?

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - name: vllm
+      version: "1"
+      autoSelect: true
+      priority: 1
+  protocolVersions:
+    - v2
+  containers:
+    - name: kserve-container
+      image: vllmserver:replace
+      command:
```
spolti, line 19: Where are the args?
You can use this as an example https://github.com/opendatahub-io/odh-model-controller/blob/incubating/config/runtimes/vllm-cpu-template.yaml

And provide gpu and cpu runtimes independently.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
yashaswipiplani: I was thinking about a constant vLLM GPU image which we can use to override image in controller if GPU is enabled. This seems like a much simpler approach and close to what we already have for HuggingFaceServer. Do you think we should separate out vLLM CPU and GPU runtimes? Because, the controller can't select GPU runtime if GPU is enabled as of now. 
Also, the parameters and configuration remains almost the same for single node GPU and CPU runtime, right? The difference is only in the case of multi-node

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: `vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@`
Arguments that are provided in isvc are appended to the vLLM serve command itself. I was thinking we do not add any default arguments and let the users add whatever arguments they want?

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,321 @@ spec:
...
+              exit 1
+            fi
+            echo "Waiting for Ray head node to be ready..."
+            sleep 2
+            elapsed=$((elapsed + 2))
+          done
+          echo "Ray head node is ready"
+
+          # Start vLLM Server with multi-node configuration
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 --tensor-parallel-size ${TENSOR_PARALLEL_SIZE:-1} --pipeline-parallel-size ${PIPELINE_PARALLEL_SIZE:-1} $@
```
cjackal: Don't we need to wait for the worker pods to be up and joined? huggingface runtime has [this check](https://github.com/kserve/kserve/blob/00e88c51ff62455d5a22fca53cdeadc3e14a53d6/python/huggingfaceserver/health_check.py#L117-L152) for example.

yashaswipiplani: Added the check in the pod. Please check now. Thanks

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,347 @@ spec:
...
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-vllmserver
+spec:
+  disabled: {{ .Values.kserve.servingruntime.vllmserver.disabled }}
+  annotations:
+    prometheus.kserve.io/port: '8000'
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
```
sivanantha321, line 734: Not sure if it is right to name the model format as vllm.

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,347 @@ spec:
...
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.kserve.servingruntime.vllmserver.hostIPC.enabled }}
+  hostIPC: true
+  {{- end }}
+---
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-vllmserver-multinode
```
sivanantha321: isn't it better to use llminferenceservice for multi node ?

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: Can you please suggest alternate appropriate name for the same. Thanks

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -708,3 +708,347 @@ spec:
...
+              exit 1
+            fi
+            
+            echo "Waiting for nodes to join... (${NODE_COUNT}/${EXPECTED_NODES} ready)"
+            sleep 5
+            elapsed=$((elapsed + 5))
+          done
+
+          # Start vLLM Server with multi-node configuration
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 --tensor-parallel-size ${TENSOR_PARALLEL_SIZE:-1} --pipeline-parallel-size ${PIPELINE_PARALLEL_SIZE:-1} $@
```
Datta0: Not really in favor of having TP and PP sizes in the args here. Shouldn't we let the user decide what config they want? Also, iirc $@ passes all the args entered at the start of the command. So not sure what happens if someone specifies --tensor-parallel-size in the runtime args again?

charts/kserve-resources/values.yaml
```diff
@@ -361,6 +361,36 @@ kserve:
       shm:
         enabled: true
         sizeLimit: "3Gi"
+    vllmserver:
+      disabled: false
```
Datta0, line 373: Is there a plan to bake in other vLLM CPU images like with AMX/open vino optimisations? Or is this supposed to be just a basic vllm cpu image with no updates scoped ?

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -244,8 +244,8 @@ func (isvc *InferenceService) setPredictorModelDefaults() {
 	}
 
 	if isvc.Spec.Predictor.Model != nil && isvc.Spec.Predictor.Model.ProtocolVersion == nil {
-		if isvc.Spec.Predictor.Model.ModelFormat.Name == constants.SupportedModelTriton {
-			// set 'v2' as default protocol version for triton server
+		if isvc.Spec.Predictor.Model.ModelFormat.Name == constants.SupportedModelTriton || isvc.Spec.Predictor.Model.ModelFormat.Name == constants.SupportedModelVLLM {
```
Datta0, line 247: Can you describe/explain this change?

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: Removed from clusterservingruntimes

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: This is removed from clusterservingruntimes altogether now. We will use llminferenceservice for multi node vLLM

charts/kserve-resources/values.yaml
yashaswipiplani: Not sure about this. Probably @gavrissh or @johnugeorge can answer this

pkg/apis/serving/v1beta1/inference_service_defaults.go
yashaswipiplani: Yes, so vLLM runtime will not support V1 protocol. Now when we apply an ISVC, the controller tries to use the V1 protocol and fails since it is not supported in runtime. Defaulting it to use V2 protocol, similar to Triton server. Without this change, we will need to specify to use V2 protocol in ISVC

charts/kserve-resources/values.yaml
gavrissh: The current CPU image here is an AMX-enabled one. 
OpenVino is deprecated in vLLM

charts/kserve-resources/templates/clusterservingruntimes.yaml
yashaswipiplani: Can you please review the changes again. Thanks

sivanantha321:
@yuzisun please provide your thoughts on this

terrytangyuan (approved):
+1 for a separate runtime. Although I am not sure if we should change the existing HF runtime since it may impact existing users. 

cc @Jooho @yuzisun

spolti:
+1 to keep it separate.

@yashaswipiplani can you please help with fixing the conflicts?

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-vllmserver
+spec:
+  disabled: {{ .Values.kserve.servingruntime.vllmserver.disabled }}
+  annotations:
+    prometheus.kserve.io/port: '8000'
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
```
spolti, line 734:
```suggestion
    - name: vLLM
```

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+      version: "1"
+      autoSelect: true
+      priority: 1
+  protocolVersions:
```
spolti, line 738: I think we can omit it.

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+    - v2
+  containers:
+    - name: kserve-container
+      image: "{{ .Values.kserve.servingruntime.vllmserver.image }}:{{ .Values.kserve.servingruntime.vllmserver.tag }}"
+      command:
```
spolti, line 743: the command can be as simple as:

```yaml
          command:
            - python
            - -m
            - vllm.entrypoints.openai.api_server
          args:
            - "--port=8080"
            - "--model=/mnt/models"
            - "--served-model-name={{.Name}}"
```

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+          then
+            export MODEL=${MODEL_ID}
+          fi
+          
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@
+        - "bash"
+      env:
+        - name: LMCACHE_USE_EXPERIMENTAL
+          value: "{{ .Values.kserve.servingruntime.vllmserver.lmcacheUseExperimental }}"
+        - name: VLLM_GPU_IMAGE
```
spolti, line 758: What is the purpose of it?

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+        - |
+          export MODEL=${MODEL_DIR}
+          if [[ -z ${MODEL_DIR} ]] && [[ ! -z ${MODEL_ID} ]]
+          then
+            export MODEL=${MODEL_ID}
+          fi
+          
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@
+        - "bash"
+      env:
```
spolti, line 755: We would also need to set HF_HOME to some place where the storageInitializer has write permissions on it to ensure it will run fine on more restricted environments.

config/runtimes/kserve-vllmserver.yaml
spolti: having default is needed so user does not need to worry about it every time.
also, please see my comments [here](https://github.com/kserve/kserve/pull/4769/changes#diff-9ff2b26b73b2bedec0ae977c62090223372b509bed0ed577adc91088b1f8fa4b) and apply as needed.

pkg/controller/v1beta1/inferenceservice/utils/utils.go
spolti: For vLLM I think we should treat it differently, and not rely on the env as you did.
Essentially a similar approach to multi-node to detect GPUs, if no GPU is available, fallback to the cpu vllm container.

spolti (changes requested):

spolti:
also, I guess you need a full rebase, there are other commits on this PR.

charts/kserve-resources/templates/clusterservingruntimes.yaml
Jooho: That’s correct—since we’re using the OpenAI API with the vLLM model, protocolVersion doesn’t seem relevant in this context.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
...
+  annotations:
+    prometheus.kserve.io/port: "8000"
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
+      version: "1"
+      autoSelect: true
+      priority: 1
+  protocolVersions:
+    - v2
```
Jooho, line 15: since we’re using the OpenAI API with the vLLM model, protocolVersion doesn’t seem relevant in this context.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-vllmserver
+spec:
+  annotations:
+    prometheus.kserve.io/port: "8000"
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
```
Jooho, line 10: I’m fine with vllm, but since vLLM is the most commonly used form in the documentation, I think vLLM would be preferable.

charts/kserve-resources/templates/clusterservingruntimes.yaml
Jooho, line 1: This YAML file will be generated automatically based on kustomize yaml file so do not update any in this file. Instead, please update the changes on kustomize yaml.

config/runtimes/kserve-vllmserver.yaml
Jooho, line 1: If we have a ServingRuntime for CPU, shouldn’t we also have ServingRuntimes for other accelerators such as GPU/rocm? If so we need to change this file/sr name to `kserve-vllmserver-cpu.yaml`

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - "bash"
+      ports:
+        - containerPort: 8000
+          name: http
+          protocol: TCP
+      env:
+        - name: LMCACHE_USE_EXPERIMENTAL
+          value: "True"
+        - name: VLLM_GPU_IMAGE
+          value: vllmserver-gpu:replace
```
Jooho, line 39: These env are required?  how this gpu image used? I thought this servingruntime is for vllm cpu.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
...
+        httpGet:
+          path: /health
+          port: 8000
+        failureThreshold: 2
+        periodSeconds: 5
+        successThreshold: 1
+        timeoutSeconds: 15
+      readinessProbe:
+        httpGet:
+          path: /health
```
Jooho, line 63: This does not aware that model is loaded or not.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
...
+      resources:
+        requests:
+          cpu: "1"
+          memory: 2Gi
+        limits:
+          cpu: "1"
+          memory: 2Gi
+      livenessProbe:
+        httpGet:
+          path: /health
```
Jooho, line 55: I think we may need a more refined approach. For liveness, it might be worth requesting a single token, since that would provide a more accurate signal.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,85 @@
...
+        timeoutSeconds: 15
+      readinessProbe:
+        httpGet:
+          path: /health
+          port: 8000
+        failureThreshold: 2
+        periodSeconds: 5
+        successThreshold: 1
+        timeoutSeconds: 15
+      startupProbe:
```
Jooho, line 69: If this is for waiting for loading a model, `/health` would not be the right way.

config/runtimes/kustomization.yaml
```diff
@@ -58,4 +59,12 @@ images:
...
   newName: kserve/huggingfaceserver
-  newTag: latest-gpu
\ No newline at end of file
+  newTag: latest-gpu
+  
+- name: vllmserver
+  newName: public.ecr.aws/q9t5s3a7/vllm-cpu-release-repo
+  newTag: latest
+
+- name: vllmserver-gpu
```
Jooho, line 68: do we have SR for this ?

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+kind: ClusterServingRuntime
+metadata:
+  name: kserve-vllmserver
+spec:
+  disabled: {{ .Values.kserve.servingruntime.vllmserver.disabled }}
+  annotations:
+    prometheus.kserve.io/port: '8000'
+    prometheus.kserve.io/path: "/metrics"
+  supportedModelFormats:
+    - name: vllm
```
yashaswipiplani: Done

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+      version: "1"
+      autoSelect: true
+      priority: 1
+  protocolVersions:
```
yashaswipiplani: Removed

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+    - v2
+  containers:
+    - name: kserve-container
+      image: "{{ .Values.kserve.servingruntime.vllmserver.image }}:{{ .Values.kserve.servingruntime.vllmserver.tag }}"
+      command:
```
yashaswipiplani: Done

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+          then
+            export MODEL=${MODEL_ID}
+          fi
+          
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@
+        - "bash"
+      env:
+        - name: LMCACHE_USE_EXPERIMENTAL
+          value: "{{ .Values.kserve.servingruntime.vllmserver.lmcacheUseExperimental }}"
+        - name: VLLM_GPU_IMAGE
```
yashaswipiplani: Removed. vLLM now pushes CPU images to `vllm/vllm-openai-cpu`. Using that now and if GPU is not enabled, using CPU image

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Removed

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Done

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
yashaswipiplani: Reverted and generated

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Followed similar approach as `TFServing` and `TorchServe` where only the image is different. Using CPU image for CPU

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Not needed now. Used proper logic

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: That is correct. Used httpGet /v1/models, which will only return 200 when model is active. Thanks for this catch

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: How can we request token, since we don't know what kind of model it will be like: text-gen, embedding, reranker or so on? I think since this is only single node vLLM support, livenessProbe is not needed following similar to HuggingFace. Please let me know your thoughts

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: That is true. Used httpGet /v1/models

config/runtimes/kustomization.yaml
yashaswipiplani: No, used a similar approach to TFServing and TorchServe where only the image is different. Using CPU image for CPU

pkg/controller/v1beta1/inferenceservice/utils/utils.go
```diff
@@ -374,16 +374,27 @@ func UpdateImageTag(container *corev1.Container, runtimeVersion *string, serving
...
+			// with a "-gpu" tag suffix; append it when runtimeVersion is not specified.
+			if (*servingRuntime == constants.TFServing || *servingRuntime == constants.TorchServe || *servingRuntime == constants.HuggingFaceServer) &&
+				!strings.HasSuffix(image, "-gpu") {
+				container.Image = image + "-gpu"
+			}
+		} else {
+			// For vLLM the CPU build lives in a sibling repository named "<image>-cpu"
+			// (e.g. vllm/vllm-openai -> vllm/vllm-openai-cpu), so insert "-cpu" immediately
+			// before the tag separator instead of appending it to the tag.
+			if *servingRuntime == constants.VLLMServer && !strings.HasSuffix(imageWithoutTag, "-cpu") {
```
Jooho: I understand you used the existing way but the way has some issues so for the new runtime, I would be great to use a new way.

`UpdateImageTag` checks the runtime by name (*servingRuntime == constants.VLLMServer), so the CPU image rewrite only works for the well-known `kserve-vllmserver` name. If someone creates a custom ServingRuntime with a different name but the same vLLM image (e.g. just tweaking args), the CPU fallback won't kick in and the GPU image gets deployed on a CPU-only node.

There's already a pattern for this in SetRuntimeDefaults — it checks the serving.kserve.io/server-type annotation first, then falls back to  runtime name. UpdateImageTag should do the same. Would also need a ServerTypeVLLMServer constant added alongside the others.

https://github.com/yashaswipiplani/kserve/blob/f914f427ad14372f424b141a3d0df57ff8656fdc/pkg/apis/serving/v1beta1/inference_service_defaults.go#L394-L399

pkg/constants/constants.go
Jooho, line 615: need to add vllmserver.

config/runtimes/kserve-vllmserver.yaml
```diff
@@ -0,0 +1,76 @@
...
+      env:
+        - name: LMCACHE_USE_EXPERIMENTAL
+          value: "True"
+        - name: HF_HOME
+          value: /tmp
+        - name: VLLM_CONFIG_ROOT
+          value: /tmp
+      securityContext:
+        allowPrivilegeEscalation: false
+        privileged: false
```
Jooho, line 39: `runAsNonRoot: true` 
Do you remove this part intentionally?

test/e2e/common/utils.py
```diff
@@ -420,6 +422,68 @@ def rerank(
...
+            data = {"input": [text]}
+
+        url, host = _get_vllm_endpoint_and_host(service_name, url_suffix, version)
+        headers = {"Host": host, "Content-Type": "application/json"}
+
+        logger.info("Sending Header = %s", headers)
+        logger.info("Sending url = %s", url)
+        logger.info("Sending request data: %s", data)
+        # temporary sleep until this is fixed https://github.com/kserve/kserve/issues/604
+        time.sleep(10)
```
Jooho: Looks like it is already fixed. Do we need this?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
yashaswipiplani: I wasn't aware of this pattern. Updated `UpdateImageTag` function to follow similar pattern. Thanks for this

pkg/constants/constants.go
yashaswipiplani: Done

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Added

test/e2e/common/utils.py
yashaswipiplani: Removed

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: The upstream vllm/vllm-openai{,-cpu}:latest image has no USER directive (runs as root) and isn't designed for arbitrary-UID operation: torch's _dynamo cache init calls getpass.getuser() at import time, which fails with KeyError: getpwuid(): uid not found: 1000 when the chosen UID isn't in /etc/passwd. Workarounds (runAsUser + USER/HOME env vars) are brittle and likely to break on every vLLM/torch release. The other CSRTs that carry runAsNonRoot: true either ship a kserve-built image with USER 1000 baked in (sklearn/xgb/huggingface/…) or wrap an upstream image that explicitly documents non-root operation (triton/tf/torchserve). vLLM does neither today. We retain the other hardenings (allowPrivilegeEscalation: false, capabilities: drop ALL, privileged: false, hostIPC: false); deployers who need PSS-Restricted compliance can layer runAsNonRoot: true on the InferenceService or via namespace PSS labels. If/when upstream vLLM adopts a non-root user — or KServe ships its own vLLM image — this can be revisited.

Jooho: Got it. That makes sense. One thing I’m wondering, though, is whether we should harden the vLLM image. Using an image that runs as root in production is often not acceptable, which means users would either need to harden it themselves or spend time finding an alternative image.

This is probably outside the scope of this PR, but it might be worth discussing separately in the future.

yashaswipiplani: Makes sense. We can discuss it

charts/kserve-runtime-configs/files/runtimes/clusterservingruntimes-patch.yaml
```diff
@@ -720,3 +720,101 @@ spec:
...
+        - |
+          export MODEL=${MODEL_DIR}
+          if [[ -z ${MODEL_DIR} ]] && [[ ! -z ${MODEL_ID} ]]
+          then
+            export MODEL=${MODEL_ID}
+          fi
+          
+          vllm serve ${MODEL} --host 0.0.0.0 --port 8000 $@
+        - "bash"
+      env:
```
yashaswipiplani: Added and set it to /tmp

config/runtimes/kserve-vllmserver.yaml
yashaswipiplani: Added as suggested. Please let me know, if we should add any more defaults. Thanks

test/e2e/predictor/test_vllm.py
yashaswipiplani: I used this model as it was being used in test_huggingface_vllm e2e tests already but makes sense to use a smaller model. 
Should this be taken separately for all e2e tests or with this PR only?

pkg/controller/v1beta1/inferenceservice/utils/utils.go
yashaswipiplani: Yes, followed the similar approach as other SRs

config/runtimes/kserve-vllmserver.yaml
cjohannsen-cloudera: We should definitely harden the image or build off a hardened base image.  @yashaswipiplani can you please file a followup issue for that?

yashaswipiplani: Created the issue: https://github.com/kserve/kserve/issues/5666. Would love to work on it. Just wanted to understand do we want to request this change in upstream vLLM or are we more inclined to build the image in kserve

PR #4787, opened 2025-10-30
Refactor existing script to use centralized installation scripts

Jooho:
This change aligns with the centralized installation approach and improves maintainability by reusing existing management scripts.

- Replace inline KServe deployment with manage.kserve-kustomize.sh
- Simplify deployment mode handling (Standard vs Knative)
- Extract gateway class name dynamically from NETWORK_LAYER variable
- Delegate gateway creation to manage.kserve-gateway.sh
- Remove redundant gateway creation logic for envoy/istio
- Apply cluster resources using server-side apply

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


hack/setup/common.sh
```diff
@@ -226,6 +240,25 @@ wait_for_crd() {
     local timeout="${2:-60s}"
 
     log_info "Waiting for CRD '$crd_name' to be established..."
+
+    # Add small delay to allow CRD status to be initialized
+    sleep 2
```
spolti, line 245: isn't better to increase the max_retries below?

hack/setup/common.sh
```diff
@@ -242,6 +275,92 @@ wait_for_crds() {
     log_success "All CRDs are established!"
 }
 
+# Update multiple fields in KServe inferenceservice-config ConfigMap
+# Usage: update_isvc_config "ingress.enableGatewayApi=true" "deploy.defaultDeploymentMode=Standard"
+# Example:
+#   update_isvc_config "ingress.enableGatewayApi=true"
+#   update_isvc_config "ingress.enableGatewayApi=true" "ingress.className=\"envoy\""
+update_isvc_config() {
```
spolti, line 283: I am not sure about this function; a simple `kubectl json patch` would address it.

This function seems very hard to debug if an issue happens later.

hack/setup/global-vars.env
```diff
@@ -38,3 +38,14 @@ ISTIO_NAMESPACE="${ISTIO_NAMESPACE:-istio-system}"
 
 # Gateway (for KServe ingress)
 GATEWAY_NAMESPACE="${GATEWAY_NAMESPACE:-kserve}"
+
+# KServe Configuration
+DEPLOYMENT_MODE="${DEPLOYMENT_MODE:-Knative}"
```
spolti, line 43: imho, for deployment mode and parameter inputs, it is best to use all lowercase.

hack/setup/infra/external-lb/templates/metallb-config.yaml.tmpl
spolti, line 1: is there a need to update permissions?

.github/workflows/e2e-test.yml
```diff
@@ -573,16 +571,13 @@ jobs:
 
       - name: Install KServe
         run: |
+          export KSERVE_CUSTOM_ISVC_CONFIGS='ingress.pathTemplate=/serving/{{ .Namespace }}/{{ .Name }}|ingress.ingressDomain=mydomain.com'
```
spolti, line 574: How are these configs being handled?

I am not sure, but patching the isvc sounds easier to debug and maintain. (if this conifg is related with the comment I left on another script that configures the isvc cm).

.github/workflows/e2e-test.yml
```diff
@@ -798,7 +793,7 @@ jobs:
 
       - name: Install KServe
         run: |
-          ./test/scripts/gh-actions/setup-kserve.sh "raw" ${{ matrix.network-layer }}
+          ./test/scripts/gh-actions/setup-kserve.sh "Standard" ${{ matrix.network-layer }}
```
spolti, line 796: all lowercase, maybe?

.github/workflows/e2e-test.yml
```diff
@@ -1110,7 +1096,7 @@ jobs:
         run: ./test/scripts/gh-actions/setup-uv.sh
 
       - name: Install KServe
-        run: |
+        run: |
```
spolti, line 1099: extra space

hack/setup/infra/knative/templates/knative-serving-istio.yaml
spolti, line 1: permissions update, is it needed?

hack/setup/infra/manage.kserve-helm.sh
```diff
@@ -21,16 +21,28 @@
 #
 # Environment variables:
 #   USE_LOCAL_CHARTS      - Use local charts instead of OCI registry (default: false)
-#   KSERVE_VERSION        - KServe version to install (default: from kserve-deps.env)
+#   SET_KSERVE_VERSION    - KServe version to install (default: from kserve-deps.env)
```
spolti, line 24: not sure about it, KSERVE_VERSION sounds good to me, any specific reason to update it?

test/crds/serving.kserve.io_all_crds.yaml
spolti, line 1: Is it needed?
Why can't we use the ones from config/crd?

hack/setup/common.sh
Jooho: 6 secs is kind enough to register CRD but I can increase 30 secs.
I will update this with the following PR soon

.github/workflows/e2e-test.yml
Jooho: No, this has actually been one of the recurring issues we’ve faced.
In overlays/test, we had multiple ConfigMap YAML files for different patches, which meant we had to update all of them each time. If we missed even one, it would cause problems.

The new approach is that overlays/test now copies the original YAML file from the configmap folder each time, and only updates the necessary parts before deploying KServe.
For debugging purposes, I’ve added some log statements, and since the issue happens during KServe setup — before the tests even start — I don’t think debugging will be too difficult.

.github/workflows/e2e-test.yml
Jooho: This change was intentional as well.
Instead of converting everything to lowercase and then changing it back internally, I thought it would be better to just use the values as they are.
After all, setup-kserve.sh is mainly for CI purposes and is only used by community members, so I don’t think it’ll cause any major issues.

.github/workflows/e2e-test.yml
Jooho: will update this part with the following up PR.

hack/setup/infra/external-lb/templates/metallb-config.yaml.tmpl
Jooho: It’s probably a change that happened while updating the .sh script files. I will update this part with the following up PR.

hack/setup/infra/knative/templates/knative-serving-istio.yaml
Jooho: It’s probably a change that happened while updating the .sh script files.
I will update this part with the following up PR.

hack/setup/infra/manage.kserve-helm.sh
Jooho: KSERVE_VERSION is already set globally across the repository through kserve-deps.env.

Since the scripts load that file using source, the KSERVE_VERSION itself cannot be changed directly.
This design is intentional — it’s required by the overall script architecture.

hack/setup/common.sh
Jooho: You’re right — this part is quite complex. I plan to refactor it later to make the logic more efficient and maintainable.
For now, the main priority is that the function works correctly, and since all the CI checks have passed, it seems to be performing well.
It would make sense to handle the refactoring in a separate PR later, so we can keep this one focused on functionality.

hack/setup/global-vars.env
Jooho: As I mentioned elsewhere, converting parameters to lowercase and then changing them back internally doesn’t seem very efficient to me.
Also, since DeployMode is a fundamental concept in KServe, I don’t think it will cause any issues.
For example, even though `RawDeployment` is quite complex, people have been using it without any problems.

test/crds/serving.kserve.io_all_crds.yaml
Jooho: This wasn’t something I changed. It’s generated by the pre-commit hook from what was already in place.
However, it seems to be intended for gathering test-related files within the test directory.

Jooho (commented):
ChatGPT said:

There are about four small parts that need to be changed, and I’ll address them in the upcoming PR.

brettmthompson:
Given all the comments documenting changes to be added in a follow up PR, I am ok with merging this in its current state.

LGTM

yuzisun (approved):

.github/actions/kserve-dep-setup/action.yml
```diff
@@ -13,21 +13,20 @@ inputs:
     description: 'Enable KEDA for autoscaling'
     required: false
     default: 'false'
+  deploy-llmisvc:
```
sivanantha321, line 16: where is this used ?

config/overlays/test/configmap/inferenceservice.yaml
```diff
@@ -52,36 +686,86 @@ data:
...
+        "memoryLimit": "1Gi",
         "cpuRequest": "100m",
-        "cpuLimit": "300m"
+        "cpuLimit": "1",
+        "imagePullPolicy": "IfNotPresent"
     }
+
+  deploy: |-
+    {
+      "defaultDeploymentMode": "Serverless"
```
sivanantha321, line 722: Should be Knative ?

PR #4795, opened 2025-11-05
ci: add retry request for e2e tests to reduce transient failures

sivanantha321:
Adds retry request using httpx-retires library for e2e tests to reduce transient failures


spolti (approved):
/lgtm

PR #4803, opened 2025-11-06
Improved CA Bundle Management For LLM Inference Services

brettmthompson:
This PR improves the logic to fully support all ca bundle configurations for model storage with llmiscvcs.

**Note**: llmisvcs do not support the use of a storage spec

The following ca bundle configurations are now fully supported for llmisvcs:
1. If using aws IAM role set the `serving.kserve.io/s3-cabundle-configmap` annotation on the service account
2. Set the `serving.kserve.io/s3-cabundle-configmap` annotation on the secret
3. Set the `AWS_CA_BUNDLE_CONFIGMAP` key in the secret data
4. Set `s3CABundleConfigMap` in the s3 config
5. Set ca bundle configmap name in storage config - this config map must exist in the same namespace as the kserve-controller-manager pod. A "global-ca-bundle" configmap will then be created in every llmisvc namespace with the same content. This generated configmap will be mounted as a volume in the init container.

To support 5 above the ca bundle config map reconciler has been updated to be resource agnostic. It now accepts a namespace rather than an isvc object as input. An unused `scheme` field in the `CaBundleConfigMapReconciler` has also been removed. Logic has been added to the llmisvc reconciler to reconcile the ca bundle configmap during llmisvc reconciliation.

This PR also updates the `inferenceservice-config` config map to include the `s3CABundleConfigMap` value, set to `""` by default (this has always been supported, but was not included in the config map). It also updates the description of the `s3CABundle` value in this config map to correctly describe both of its use cases:
1. When used with a configured CA bundle config map, specifies the mounted path for the config map.
2. When used absent of a configured CA bundle config map, specifies the path to a certificate bundle to use for HTTPS certificate validation.

Lastly unit tests have been added to cover llmisvcs utilzing s3 storage configured via the s3 config. Additional tests have been added to cover llmisvc multi node with ca bundles as well.

**Type of changes**

1.  Configure private s3 storage + ca bundles in any of the 5 ways described above.
2. Deploy an inference services that uses the s3 storage.
4. The storage initializer pod should successfully communicate with the s3 storage using the configured ca bundle over `http` and the model should be successfully downloaded.

**Checklist**:

**Release note**:

```release-note
Full support for CA bundle management with LLM inference services
```

**Re-running failed tests**


brettmthompson:
Looks like filenames have been changed for llmisvc charts to add the `kserve-` prefix to their names. This is causing `LLMInferenceService E2E Tests / test-llmisvc` to fail. 

I will create another PR to address this bug.

pierDipi (approved):
Thanks!

/lgtm

pkg/controller/v1alpha1/llmisvc/controller_workload_storage_int_test.go
```diff
@@ -1491,14 +1762,259 @@ var _ = Describe("LLMInferenceService Controller - Storage configuration", func(
...
+					Namespace: nsName,
+				}, expectedPrefillLWS)
+			}).WithContext(ctx).Should(Succeed())
+
+			// validate the storage initializer configuration in the leader worker sets
+			validateStorageInitializerIsConfiguredForLWS(expectedMainLWS, "s3://user-id/repo-id:tag")
+			validateStorageInitializerIsConfiguredForLWS(expectedPrefillLWS, "s3://user-id/repo-id:tag")
+
+			// validate the storage initializer credentials are properly set
+			expectedEnvVars := []corev1.EnvVar{
```
spolti: this one seems to be very similar to the previous tests, can this be reused for both tests?

pkg/controller/v1alpha1/llmisvc/controller_workload_storage_int_test.go
```diff
@@ -1618,12 +2134,72 @@ var _ = Describe("LLMInferenceService Controller - Storage configuration", func(
 					Name:  s3.AWSRegion,
 					Value: s3Region,
 				},
+				{
```
spolti: This env too.

spolti (commented):
Minor improvement that could be done, otherwise, looks good :)
Thanks.

pkg/controller/v1alpha1/llmisvc/controller_workload_storage_int_test.go
brettmthompson: yes I can try to consolidate the expected env variables into a single variable to be reused across tests

brettmthompson: added a utility function to create the expected env vars as needed

pkg/controller/v1alpha1/llmisvc/controller_workload_storage_int_test.go
brettmthompson: added a utility function to create the expected env vars as needed

Issue #4823, opened 2025-11-13
Enhancement: Upgrade KServe LLMISVC to Gateway API Inference Extension v1


atalhens:
## Background
KServe currently depends on `sigs.k8s.io/gateway-api-inference-extension` `v0.3.0`. Upstream has moved to v1 with several breaking API updates: `InferenceObjective` replaces `InferenceModel`, `InferencePool` schemas changed, and request prioritisation moved to numeric `Priority`. These updates are not yet reflected in KServe’s controller, configs, or tests, leaving the project pinned to an outdated GIE release.

## Problem

- Not in sync with the latest changes that are provided in 1.1. Also current version utilizes the InferenceModel spec that has been removed from the newest release. Moving forward, there is a need to upgrade the same.
- Missing upstream bug fixes and features.

## Requested Work
- Audit all KServe code paths that consume the GIE API (LLMISVC controller, fixtures, configs, CRDs).
- Migrate from `InferenceModel` to `InferenceObjective` and adopt the v1 schema.
- Update documentation and tests to reflect the new API.
- Resolve tooling or Go-module issues encountered during the upgrade (e.g., ensuring a Go 1.24-compatible `controller-gen`).

## Additional Context
Upstream release notes: https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases

PR #4825, opened 2025-11-13
Migrate LLMISVC controller to Gateway API InferenceObjective v1

### LINKED ISSUE #4824: [Bug] `make generate` fails with controller-gen v0.16.2 under Go 1.24
## Background
After upgrading the repository to Go 1.24.1, the `make generate` target (used to refresh deepcopy code, CRDs, etc.) began failing on macOS and Linux environments. The failure blocks regeneration of API artifacts, which we need for the Gateway API inference-extension migration and other schema changes.

Note:
The Go version upgrade was necessary to update GIE's version 1.1 packages in Kserve.

## Environment
- Go: 1.24.1
- Current controller-gen version in `kserve-deps.env`: `CONTROLLER_TOOLS_VERSION=v0.16.2`

## Reproduction Steps

```
git clone https://github.com/kserve/kserve.git
cd kserve
make generate## Expected Result
`make generate` completes successfully and updates generated files.
```

The error that I face while trying to run make generate while having the controller-gen dependency.

```
.../internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)

```
# Root Cause
`controller-gen` v0.16.2 relies on code that no longer compiles with the Go 1.24 toolchain (`cmd/cover` in x/tools). `make generate` installs that version using the `CONTROLLER_TOOLS_VERSION` specified in `kserve-deps.env`, causing the build to fail whenever Go ≥1.24 is used.

## Proposed Fix
- Bump `CONTROLLER_TOOLS_VERSION` to a Go 1.24-compatible release (`v0.17.1` or later) in `kserve-deps.env`.
- Run `make controller-gen` once to install the new binary (`$(call go-install-tool, ...)` will pick up the updated version).
- Re-run `make generate` to confirm it completes successfully.

## Additional Notes
- This bug blocks regeneration of `zz_generated.deepcopy.go`, CRDs, and any updates that rely on controller-gen.

atalhens:
- Migrates the LLM Inference Service controller from the deprecated `InferenceModel` resource to the new `InferenceObjective` API (Gateway API inference extension v1).
- Preserves existing functionality for model naming, pool selection, and request prioritisation by mapping the old `Criticality` semantics to the new numeric `Priority`.
- Updates supporting configs, tests, and fixtures to the v1 schema (`endpointPickerRef`, `targetPorts`, `LabelSelector.matchLabels`, etc.).
- Fixes build and tooling issues introduced by the dependency and Go toolchain upgrades (e.g. controller-gen version mismatch).

Fixes : #4824 #4823

**Type of changes**

**Checklist**:

**Release note**:

```release-note
Migrated the LLM Inference Service controller to the Gateway API InferenceObjective v1 schema and aligned configs/tests with the updated prioritisation model.
```

**Re-running failed tests**


pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -126,10 +126,13 @@ type LLMModelSpec struct {
...
-	// Criticality defines how important it is to serve the model compared to other models.
-	// This is used by the Inference Gateway scheduler.
+	// Priority defines how important it is to serve this model compared to other models.
+	// This is used by the Inference Gateway scheduler for request prioritization.
+	// Higher values indicate higher priority. Negative values are allowed.
+	// If unset, defaults to 0. This maps to the Priority field in InferenceObjective.
+	// For example: 10 (high priority), 0 (default), -10 (low priority/sheddable).
 	// +optional
-	Criticality *igwapi.Criticality `json:"criticality,omitempty"`
+	Priority *int `json:"priority,omitempty"`
```
pierDipi, line 135: since this is a breaking change, to be able to upgrade objects from one version to another, we'll have to create a v1alpha2 API and add a conversion webhook for v1alpha1 <-> v1alpha2

pierDipi: (this is not the only one, the embedded InferencePoolSpec has changed too)

atalhens: That makes sense; I will create a v1aplah2 reference for the LLMSVC changes, including a conversion webhook.

PR #4832, opened 2025-11-16
fix: Fix various bugs in quickstart scripts

### LINKED ISSUE #4831: Quickstart guide installation fails with "No such file or directory" errors
/kind bug

Following the [KServe quickstart guide](https://kserve.github.io/website/docs/getting-started/quickstart-guide), I ran the installation command:

```bash
curl -s "https://raw.githubusercontent.com/kserve/kserve/master/hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh" | bash
```

The installation fails with the following error:

```
mv: cannot move '/tmp/tmp.Hbxk19H93N/kustomize' to '\033[93m[WARNING]\033[0m Could not find git repository root, using current directory:
/home/jonah/Projects/github.com/kserve\n/home/jonah/Projects/github.com/kserve/bin/kustomize': No such file or directory
```

The installation script should successfully install KServe in standard mode as documented in the quickstart guide.

```
N/A - Installation fails before KServe is deployed.
```

Root causes identified:

1. The find_repo_root() function outputs warnings to stdout instead of stderr. When the function's output is captured via command substitution (e.g., BIN_DIR="$(find_repo_root)"), the warning messages get included in the variable value, corrupting the path.
2. The script assumes the bin/ directory exists but never creates it. When tools like kustomize, helm, and yq are downloaded, the mv command fails because the target directory doesn't exist.

Impact:
- Users following the official quickstart guide cannot install KServe
- Affects fresh installations where the bin/ directory doesn't already exist
- Related to PR #4813 which introduced the find_repo_root refactoring

Workaround:
- Run the script from a git repository.
- Manually create the bin/ directory before running the script:
  ```
  mkdir -p bin
  curl -s "https://raw.githubusercontent.com/kserve/kserve/master/hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh" | bash
  ```

**Environment:**

- Istio Version: N/A (not yet installed)
- Knative Version: N/A (not yet installed)
- KServe Version: master branch (attempting to install)
- Kubeflow version: N/A
- Cloud Environment: local
- Minikube/Kind version: N/A (installation script fails before cluster setup)
- Kubernetes version: N/A (installation script fails before cluster interaction)
- OS (e.g. from /etc/os-release): Linux, Fedora 42

### LINKED ISSUE #4843: Quickstart scripts exit early and have incorrect installation function names
/kind bug

I was deploying kserve to my Kubernetes cluster with the quickstart scripts and noticed that the [verify installation commands](https://kserve.github.io/website/docs/getting-started/quickstart-guide#verify-installation) were not finding the pods, CRD, etc... I noticed that the logs for the getting started script abruptly stopped and many things weren't getting installed.

The installation script should successfully install KServe in standard mode as documented in the quickstart guide.

N/A - Installation fails before KServe is deployed.

Root causes identified:

1. The installation scripts inside `hack/setup/infra` exit early due to some extraneous `exit 0`s inside the install functions. This happens, for example, during external LB setup when using Openshift or Kubernetes:
   https://github.com/kserve/kserve/blob/6b72be4cfbbc1a6c1d6ae8247bfe640df31ebea7/hack/setup/quick-install/kserve-standard-mode-full-install-helm.sh#L949-L953
2. Additionally, when these functions are pasted into the various getting started scripts inside [generator.py](https://github.com/kserve/kserve/blob/master/hack/setup/scripts/install-script-generator/generator.py), the functions themselves are renamed (e.g. `install` -> `install_istio_ingress_class`) but the calls are not. Compare the following:
   https://github.com/kserve/kserve/blob/6b72be4cfbbc1a6c1d6ae8247bfe640df31ebea7/hack/setup/infra/manage.cert-manager-helm.sh#L39-L57
   
   https://github.com/kserve/kserve/blob/6b72be4cfbbc1a6c1d6ae8247bfe640df31ebea7/hack/setup/quick-install/kserve-standard-mode-full-install-helm.sh#L730-L747
   
   The latter will fail to uninstall inside `install_cert_manager` as `uninstall` is not renamed to `uninstall_cert_manager`.
   

Impact:
- Users following the official quickstart guide cannot install KServe

**Environment:**

- Istio Version: N/A (not yet installed)
- Knative Version: N/A (not yet installed)
- KServe Version: master branch (attempting to install)
- Kubeflow version: N/A
- Cloud Environment: local
- Minikube/Kind version: N/A (installation script fails before cluster setup)
- Kubernetes version: N/A (installation script fails before cluster interaction)
- OS (e.g. from /etc/os-release): Linux, Fedora 42

JonahSussman:

This PR fixes installation failures when following the [quickstart guide](https://kserve.github.io/website/docs/getting-started/quickstart-guide) by fixing some issues along the same lines as #4813:

1. **BIN_DIR creation**: The `bin` directory is now created if it doesn't exist, preventing failures when tools like `kustomize`, `helm`, and `yq` are downloaded
2. **stderr redirection**: Warnings from `find_repo_root()` are now properly redirected to stderr, preventing them from polluting the `REPO_ROOT` variable when captured via command substitution
3. **Cleanup logic**: Added trap-based cleanup to remove `BIN_DIR` if created by the script, keeping the environment clean
4. **Early installation exit**: Removed `exit 0`s inside installation functions causing the install scripts to not finish
5. **Documentation fix**: Fixed broken contributor guide link in template

**Problem:**
Running the quickstart command:
```bash
curl -s "https://raw.githubusercontent.com/kserve/kserve/master/hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh" | bash
```

Results in error:
```
mv: cannot move '/tmp/tmp.Hbxk19H93N/kustomize' to '\033[93m[WARNING]\033[0m Could not find git repository root, using current directory:
/home/jonah/Projects/github.com/kserve\n/home/jonah/Projects/github.com/kserve/bin/kustomize': No such file or directory
```

Additionally, after fixing these issues, the installation functions exit early, failing to install all the required components. 

Root causes:
1. Warning messages were sent to stdout instead of stderr, causing them to be captured in BIN_DIR variable
2. The bin/ directory was assumed to exist but was never created
3. The infra installation scripts had extra `exit 0`s inside them
4. The calls for the `uninstall` and `install`s do not get renamed.

**Which issue(s) this PR fixes:**
Fixes #4831
Fixes #4843

**Type of changes**

**Feature/Issue validation/testing:**

Validated the fix by:

- Ran hack/setup/scripts/validate-install-scripts.py - all checks pass
- Tested installation in local environment - bin directory created and cleaned up properly
- Verified warning messages appear on stderr without polluting variables
- Confirmed tools (helm, kustomize, yq) install successfully to BIN_DIR

Test configuration:
- Environment: Fedora Linux 42, Bash 5.2.37
- Installation method: Direct script execution and curl piping
- Verified both with and without pre-existing bin/ directory

**Special notes for your reviewer:**

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes. (N/A - This PR only modifies installation script logic, no image versions changed)
2. Changes are in the template file (templates/generated-script.template) so all generated scripts will benefit from these fixes
5. The cleanup logic uses a flag (`BIN_DIR_CREATED_BY_SCRIPT`) to only remove directories created by the script, preserving any pre-existing bin/ directories
6. Also fixed the broken contributor guide link

**Checklist:**

  - Validated using existing test suite (validate-install-scripts.py)
  - Added comments explaining BIN_DIR tracking and cleanup logic
  - No documentation changes needed (fixed existing broken link)

**Release note:**

```release-note
Fixed quickstart installation script to properly create bin directory and redirect warnings to stderr
```

### DIFF
--- a/hack/setup/common.sh
+++ b/hack/setup/common.sh
@@ -36,11 +36,11 @@ find_repo_root() {
 
     # Git repository not found
     if [[ "$skip" == "true" ]]; then
-        log_warning "Could not find git repository root, using current directory: $PWD" >&2
+        log_warning "Could not find git repository root, using current directory: $PWD"
         echo "$PWD"
         return 0
     else
-        echo "Error: Could not find git repository root" >&2
+        log_error "Could not find git repository root"
         exit 1
     fi
 }
@@ -60,7 +60,7 @@ detect_os() {
     case "$(uname -s)" in
         Linux*)  os="linux" ;;
         Darwin*) os="darwin" ;;
-        *)       echo "Unsupported OS" >&2; exit 1 ;;
+        *)       log_error "Unsupported OS detected: $(uname -s)" ; exit 1 ;;
     esac
     echo "$os"
 }
@@ -70,7 +70,7 @@ detect_arch() {
     case "$(uname -m)" in
         x86_64)  arch="amd64" ;;
         aarch64|arm64) arch="arm64" ;;
-        *)       echo "Unsupported architecture" >&2; exit 1 ;;
+        *)       log_error "Unsupported architecture detected: $(uname -m)" ; exit 1 ;;
     esac
     echo "$arch"
 }
@@ -107,7 +107,7 @@ else
 fi
 
 log_info() {
-    echo -e "${BLUE}[INFO]${RESET} $*"
+    echo -e "${BLUE}[INFO]${RESET} $*" >&2
 }
 
 log_error() {
@@ -115,11 +115,11 @@ log_error() {
 }
 
 log_success() {
-    echo -e "${GREEN}[SUCCESS]${RESET} $*"
+    echo -e "${GREEN}[SUCCESS]${RESET} $*" >&2
 }
 
 log_warning() {
-    echo -e "${YELLOW}[WARNING]${RESET} $*"
+    echo -e "${YELLOW}[WARNING]${RESET} $*" >&2
 }
 
 
--- a/hack/setup/common.sh
+++ b/hack/setup/common.sh
@@ -36,7 +36,7 @@ find_repo_root() {
 
     # Git repository not found
     if [[ "$skip" == "true" ]]; then
-        log_warning "Could not find git repository root, using current directory: $PWD"
+        log_warning "Could not find git repository root, using current directory: $PWD" >&2
         echo "$PWD"
         return 0
     else
--- a/hack/setup/common.sh
+++ b/hack/setup/common.sh
@@ -435,8 +435,15 @@ version_gte() {
 if [[ -z "${REPO_ROOT:-}" ]]; then
     REPO_ROOT="$(find_repo_root "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
     export REPO_ROOT
-    export BIN_DIR="${REPO_ROOT}/bin"
-    ensure_dir "${BIN_DIR}"
+
+    # Set up BIN_DIR - use repo bin if it exists, otherwise use temp directory
+    if [[ -d "${REPO_ROOT}/bin" ]]; then
+        export BIN_DIR="${REPO_ROOT}/bin"
+    else
+        export BIN_DIR="$(mktemp -d)"
+        log_info "Using temp BIN_DIR: ${BIN_DIR}"
+    fi
+
     export PATH="${BIN_DIR}:${PATH}"
 fi
 
--- a/hack/setup/common.sh
+++ b/hack/setup/common.sh
@@ -75,6 +75,22 @@ detect_arch() {
     echo "$arch"
 }
 
+cleanup_bin_dir() {
+    # Remove BIN_DIR if it was created by this script
+    if [[ "${BIN_DIR_CREATED_BY_SCRIPT:-false}" == "true" ]] && [[ -d "${BIN_DIR:-}" ]]; then
+        log_info "Cleaning up BIN_DIR: ${BIN_DIR}"
+        rm -rf "${BIN_DIR}"
+    fi
+}
+
+cleanup() {
+    # Call all cleanup functions
+    cleanup_bin_dir
+}
+
+# Set up trap to run cleanup on exit
+trap cleanup EXIT
+
 # Color codes (disable if NO_COLOR is set or not a terminal)
 if [[ -z "${NO_COLOR:-}" ]] && [[ -t 1 ]]; then
     BLUE='\033[94m'
--- a/hack/setup/quick-install/kserve-knative-mode-dependency-install.sh
+++ b/hack/setup/quick-install/kserve-knative-mode-dependency-install.sh
@@ -478,16 +478,15 @@ set_env_with_priority() {
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-.}")" && pwd)"
 REPO_ROOT="$(find_repo_root "${SCRIPT_DIR}" "true")"
 export REPO_ROOT
-export BIN_DIR="${REPO_ROOT}/bin"
 
-# Track if we created BIN_DIR so we can clean it up later
-BIN_DIR_CREATED_BY_SCRIPT=false
-if [[ ! -d "${BIN_DIR}" ]]; then
-    log_info "Creating BIN_DIR: ${BIN_DIR}"
-    BIN_DIR_CREATED_BY_SCRIPT=true
+# Set up BIN_DIR - use repo bin if it exists, otherwise use temp directory
+if [[ -d "${REPO_ROOT}/bin" ]]; then
+    export BIN_DIR="${REPO_ROOT}/bin"
+else
+    export BIN_DIR="$(mktemp -d)"
+    log_info "Using temp BIN_DIR: ${BIN_DIR}"
 fi
 
-ensure_dir "${BIN_DIR}"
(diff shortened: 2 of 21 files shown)


sivanantha321 (approved):

hack/setup/common.sh
Jooho, line 1: You should add the new function in here.  Then you can execute `make generate-quick-install-scripts`. It will add the function into each quick install script.

hack/setup/scripts/install-script-generator/templates/generated-script.template
Jooho, line 1: It does not need to add here

JonahSussman:
Hey @Jooho , let me know if the latest revision implements what you were talking about

hack/setup/common.sh
```diff
@@ -75,6 +75,22 @@ detect_arch() {
...
+    fi
+}
+
+cleanup() {
+    # Call all cleanup functions
+    cleanup_bin_dir
+}
+
+# Set up trap to run cleanup on exit
+trap cleanup EXIT
```
Jooho, line 92: Should we clean up the directory every time? If the first attempt failed, it will download the files again.

hack/setup/quick-install/kserve-standard-mode-full-install-helm.sh
```diff
@@ -49,7 +49,7 @@ find_repo_root() {
 
     # Git repository not found
     if [[ "$skip" == "true" ]]; then
-        log_warning "Could not find git repository root, using current directory: $PWD"
+        log_warning "Could not find git repository root, using current directory: $PWD" >&2
```
Jooho: Please add `>&2` to [these methods](https://github.com/kserve/kserve/blob/master/hack/setup/common.sh#L93-L107)

Jooho (commented):
The root cause was that the correct value wasn’t being passed into the path, and it is resolved by using >&2. However, I’m not sure whether we also need to remove the existing bin folder every time. Could you clarify why that step is necessary?

JonahSussman:
Thanks for the review, @Jooho and noted, I also cleaned up the `>&2` redirection.

To your question: we don’t remove an existing bin directory. The cleanup only runs when:

1. `bin/` did not exist before the script ran, and
2. The script created it specifically for installing the tools.

That’s what the `BIN_DIR_CREATED_BY_SCRIPT` flag tracks. If a user already has a `bin/` directory, we leave it completely untouched.

The main goal was just to avoid leaving stray bin/ folders behind when users run `curl | bash` from arbitrary directories. 

An alternative would be to install into a directory under `/tmp` instead, then we wouldn’t need cleanup logic since the OS eventually clears temp dirs. Happy to switch to that approach if you’d prefer.

Let me know what you think

spolti (approved):

JonahSussman:
Apologies for the re-review @sivanantha321 and @spolti , I modified some of the Python files to fix the script generation as well.

Jooho:
@JonahSussman thanks for the reply.

KServe doesn’t have a bin directory by default, so it always gets created whenever the scripts run. It’s also included in .gitignore, so it never gets committed. Because of that, my view is that there’s no real need to delete the bin folder at all — it will always be created anyway.

For the quick-install script specifically, I’m fine with creating a temporary bin directory under /tmp. But when the repository is cloned, I think it makes more sense to place the binaries under kserve/bin.

If we want this behavior, it seems like we only need to update the quick-install template file. What do you think?

Jooho:
Thanks for the update. I triggered the test ci.

JonahSussman:
@Jooho Thanks for the feedback! I think you're right - this makes more sense. I've made some changes that should address your concerns.

I added a bin/.gitkeep file so the bin/ directory will now be committed to the repo (but stay empty). I also updated .gitignore to ignore everything in
bin/ except the .gitkeep:

I also updated the logic in both common.sh and the generated-script template to check if the bin/ directory exists - if it does, it uses it; otherwise it
falls back to a temp directory:

```sh
# Set up BIN_DIR - use repo bin if it exists, otherwise use temp directory
if [[ -d "${REPO_ROOT}/bin" ]]; then
    export BIN_DIR="${REPO_ROOT}/bin"
else
    export BIN_DIR="$(mktemp -d)"
    log_info "Using temp BIN_DIR: ${BIN_DIR}"
fi
```

Now when someone clones the repo, bin/ will already exist and binaries will go there. For the quick-install curl | bash scenario where there's no repo,
it'll use /tmp instead. Since we're using temp directories for that case, there's no cleanup needed.

Does this address what you were thinking?

Jooho:
please do `make precommit`

JonahSussman:
@Jooho My apologies, I re-ran `make precommit`

Jooho (approved):
/lgtm
/approve

hack/setup/quick-install/kserve-knative-mode-dependency-install.sh
```diff
@@ -83,11 +83,27 @@ detect_arch() {
     case "$(uname -m)" in
         x86_64)  arch="amd64" ;;
         aarch64|arm64) arch="arm64" ;;
-        *)       echo "Unsupported architecture" >&2; exit 1 ;;
+        *)       log_error "Unsupported architecture detected: $(uname -m)" ; exit 1 ;;
     esac
     echo "$arch"
 }
 
+cleanup_bin_dir() {
```
sivanantha321, line 91: is this still necessary ?

PR #4837, opened 2025-11-18
fix: remove cert-manager annotation in inferenceservice crd

juhyeon-cha:
**What this PR does / why we need it**: this pr fix inferenceservice crd's annotation

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note
NONE
```

**Re-running failed tests**


sivanantha321 (approved):

spolti (approved):
/lgtm

charts/kserve-crd/templates/serving.kserve.io_inferenceservices.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: kserve/serving-cert
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/serving-cert
```
Jooho: kustomize does not have this annotation so I wonder if we can remove this annotation.

Can you share the exact log or issue in detail? I don't see it in the description. 

@andresllh could you please double check this using the converting tool?

Jooho (commented):
/hold

charts/kserve-crd/templates/serving.kserve.io_inferenceservices.yaml
juhyeon-cha: @Jooho I’m not a cert-manager or K8s expert, so I’m not certain what the precise effect of that annotation is.
However: since other resources in the chart use that annotation with `{{ .Release.Namespace }}`, I updated it accordingly for consistency.

Jooho: Thanks for your explanation!

From my understanding, since this is a CRD, it doesn’t seem necessary for it to have a cert-manager annotation.
Moreover, if the converting tool does not add this, we don't need it. @andresllh can confirm this. 
@sivanantha321 wdyt?

sivanantha321: Yeah. Only the webhook need this annotation. this is for automatically injecting cert to the webhook.

sivanantha321: Probably we can remove it.

juhyeon-cha: I removed it.

sivanantha321 (approved):

PR #4840, opened 2025-11-19
add startup probe and reduce liveness delay

strangiato:

Currently when starting a model the model server has a fixed delay of 120 seconds (300 in some of the workers) before the liveness probe starts in order to allow the model server to start.

This means that a smaller model that can start in less than 120 seconds will never be made ready before that, and a larger models which require longer than 120 seconds will likely fail.

This PR adds a `startupProbe` that can have a larger failure threshold to allow larger models to startup quicker, and still allow smaller models to be marked as ready faster.

Fixes #

**Type of changes**

Manually changed configs on deployed cluster and tested intelligent routing and prefill/decode desegregation scenarios.

- Logs

**Checklist**:

**Release note**:

```release-note
Updated model servers to utilize startupProbe to improve startup speed and reliability
```

**Re-running failed tests**


config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -45,12 +45,21 @@ spec:
...
+          periodSeconds: 10
+          timeoutSeconds: 10
+          failureThreshold: 60
         livenessProbe:
           httpGet:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 120
+          initialDelaySeconds: 10
```
spolti: IMHO, this delay here should be higher, as it should wait for a few time, the startup checks are done by the readinessProbe.

Isn't it better to tune the readinessPprobe instead of adding a new one?

strangiato: The `startupProbe` feature was created specifically to handle containers with large startup times (ideal for vLLM).

The idea is that a StartupProbe will delay the execution of the liveness/readiness probes.  This allows the StartupProbe to have a high failure rate while keeping the failure rates of the liveness/readiness probes low.  We want the liveness/readiness probes to fail fast when we see real problems.  Essentially a StartupProbe acts like an intelligent `initialDelay` for the readiness/liveness probe.

Additionally, you generally also want your readiness probe to be lower than your liveness probe, which can allow a container to possibly recover without completely restarting it, especially for containers with large startup times.

I also, didn't realize when I updated this that the readinessProbe failureThreshold is 60, which isn't great.  This means that the container will only ever get restarted if it has a problem.  If your readinessProbe is lower than your liveness probe, it is possible for the container to be pulled out of the Service pool and give it time to recover before attempting to restart.  This can be useful if the container is overwhelmed with requests and and stops responding, but may just need a few seconds to clear its current workload to recover without a full restart.

Currently we have a long delay in the livenessProbe and we **hope** that the model loads in that time.  I started running into issues with gpt-oss-20b so any model larger than this will almost guaranteed have problems.

Here is a comparison of a few difference scenarios...

With the current setup, if I am loading a model that takes 60 seconds to load, here is what will happen:
* [0s] Container starts and the model begins loading
* [10s] Readiness probe kicks off after the initial 10 seconds, starts failing every ten seconds, failure threshold is 60 so it must complete in 600 seconds, or until the liveness probe reaches its failure threshold
* [60s] Model finishes loading, the readiness probe succeeds on its next execution, the pod is marked as read
* [120s] The liveness probe kicks off its first execution and succeed

With the current setup, if I am loading a model that takes 300 seconds to load, here is what will happen:
* [0s] Container starts and the model begins loading
* [10s] Readiness probe kicks off after the initial 10 seconds, starts failing every ten seconds, failure threshold is 60 so it must complete in 600 seconds, or until the liveness probe reaches its failure threshold
* [120s] The liveness probe kicks off its first execution and fails
* [150s] The liveness probe kicks off its third attempt and fails, reaching the threshold and restarting the container, repeat until the container startup is set to BackOff

With the startupProbe setup, if I am loading a model that takes 60 seconds to load, here is what will happen:
* [0s] Container starts and the model begins loading
* [10s] Startup probe kicks off after the initial 10 seconds, starts failing every ten seconds, failure threshold is 60 so it must complete in 600 seconds
* [60s] Model finishes loading, the startup probe succeeds on its next execution
* [70s] Liveness probe and Readiness probe start, container is marked as ready

With the startupProbe setup, if I am loading a model that takes 300 seconds to load, here is what will happen:
* [0s] Container starts and the model begins loading
* [10s] Startup probe kicks off after the initial 10 seconds, starts failing every ten seconds, failure threshold is 60 so it must complete in 600 seconds
* [300s] Model finishes loading, the startup probe succeeds
* [310s] Liveness probe and Readiness probe start, container is marked as ready

Here is how the container would behave if it stops responding after it has been running a while:
* [0s] Liveness probe and readiness probe fail for the first time
* [30s] Liveness probe reaches failure threshold and restarts container, readiness probe failure threshold is irrelevant

However, this behavior changes when issues start to occur immediately after the container has started:
* [60s] Model load completes and readiness probe marks container as ready
* [80s] Model server stops responding and readiness probe starts failing, threshold is 60 and is irrelevant 
* [120s] Liveness probe kicks off and starts to fail
* [150s] Liveness probe reaches failure threshold and restarts the container

Here is how a container would behave if we had a readiness/liveness probe like this when the container has been running a while:
```
readinessProbe:
  initialDelay: 10
  periodSeconds: 5
  timeoutSeconds: 10
  failureThreshold: 3
livenessProbe:
  initialDelay: 10
  periodSeconds: 10
  timeoutSeconds: 2
  failureThreshold: 3
```
* [0s] Liveness probe and readiness probe fail for the first time
* [15s] Readiness probe reaches failure threshold and removes the container from the service pool, giving the container a chance to recover
* [30s] Liveness probe reaches failure threshold and restarts container

Here is how the container would behave with the same configuration with a startupProbe when the container starts failing right after the model 
* [60s] Model finishes loading, the startup probe succeeds on its next execution
* [70s] Liveness probe and Readiness probe start, container is marked as ready
* [80s] Model server stops responding and readiness and liveness probe starts failing
* [95s] Readiness probe reaches failure threshold and removes the container from the service pool, giving the container a chance to recover
* [110s] Liveness probe threshold reached and the container is restarted

The initialDelay could actually be removed entirely from this scenario since we already know that the /health endpoint has successfully responded to the Startup Probe and the reaction time of k8s responding to these scenarios could be even faster, without risking problems with the model startup time.

Having large initial delays creates all kinds of weird behavior in different scenarios that can be very inconsistent.  StartupProbes were specifically designed to help solve these exact problems.

Feel free to reach out via slack if you want to discuss in real time!

strangiato: Here is what I would probably recommend with some updates to the readiness probe:
```
        startupProbe:
          httpGet:
            path: /health
            port: 8001
            scheme: HTTP
          initialDelaySeconds: 10
          periodSeconds: 10
          timeoutSeconds: 10
          failureThreshold: 60
        livenessProbe:
          httpGet:
            path: /health
            port: 8001
            scheme: HTTP
          initialDelaySeconds: 0
          periodSeconds: 10
          timeoutSeconds: 10
          failureThreshold: 3
        readinessProbe:
          httpGet:
            path: /health
            port: 8001
            scheme: HTTP
          initialDelaySeconds: 0
          periodSeconds: 5
          timeoutSeconds: 5
          failureThreshold: 3
```

config/llmisvcconfig/config-llm-decode-worker-data-parallel.yaml
```diff
@@ -237,12 +237,20 @@ spec:
...
         terminationMessagePolicy: FallbackToLogsOnError
+        startupProbe:
+          httpGet:
+            path: /health
+            port: 8001
+            scheme: HTTP
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          timeoutSeconds: 10
+          failureThreshold: 60
```
pierDipi, line 247: There is another pod spec up in the same file where we need to add this too

config/llmisvcconfig/config-llm-prefill-worker-data-parallel.yaml
```diff
@@ -180,12 +180,20 @@ spec:
...
           terminationMessagePolicy: FallbackToLogsOnError
+          startupProbe:
+            httpGet:
+              path: /health
+              port: 8000
+              scheme: HTTP
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            timeoutSeconds: 10
+            failureThreshold: 60
```
pierDipi, line 190: There is another pod spec down in the same file where we need to add this too

config/llmisvcconfig/config-llm-worker-data-parallel.yaml
```diff
@@ -179,12 +179,20 @@ spec:
...
         terminationMessagePolicy: FallbackToLogsOnError
+        startupProbe:
+          httpGet:
+            path: /health
+            port: 8000
+            scheme: HTTP
+          initialDelaySeconds: 10
+          periodSeconds: 10
+          timeoutSeconds: 10
+          failureThreshold: 60
```
pierDipi, line 189: There is another pod spec down in the same file where we need to add this too

config/llmisvcconfig/config-llm-decode-template.yaml
spolti: thanks for the explanation.

spolti:
I don't have permissions to start the tests yet, @Jooho may yoy trigger it?

strangiato:
@pierDipi I added probes to the workers in the files you mentioned where there were not any probes at all. The port seems

I also ran the `make precommit` process locally which looks like it generated some updates to a few scripts.  I assume I should be committing those changes back as well but let me know if that is not correct.

strangiato:
Any suggestions on the failed tests?

As far as I can tell they are not related to the changes that I have made, and I see some similar error in other runs.

Jooho:
I rebased and retrigger the tests.

spolti:
Hi @strangiato can you please rebase and fix the conflicts?

sivanantha321:
closing in favor of #5063

PR #4856, opened 2025-11-25
feat: allow llmisvc scheduler config using a CM or inline

### LINKED ISSUE #4628: Provide default EndpointPickerConfig using a configmap for llmisvc scheduler
Can't we provide via ConfigMap?

_Originally posted by @johnugeorge in https://github.com/kserve/kserve/pull/4614#discussion_r2243541513_

pierDipi:
Simplify setting custom EndpointPickerConfig 

## Inline

```yaml
kind: LLMInferenceService
metadata:
  name: qwen2-7b-kv-cache-routing
spec:
  # ...
  router:
    # ...
    scheduler:
      config:
        inline:
          kind: EndpointPickerConfig
          plugins:
            - type: prefix-cache-scorer
              parameters:
                mode: cache_tracking  # Real-time KV cache tracking via ZMQ events
                indexerConfig:
                  tokenProcessorConfig:
                    blockSize: 64       # Must match vLLM --block-size (default is 16)
                    hashSeed: "42"      # Must match PYTHONHASHSEED in vLLM pods
                  kvBlockIndexConfig:
                    enableMetrics: true                   # Enable Prometheus metrics for KV cache index
                    metricsLoggingInterval: 60000000000   # Log metrics every 60 seconds (in nanoseconds)
                  tokenizersPoolConfig:
                    tokenizersCacheDir: /mnt/tokenizers
            - type: queue-scorer
            - type: kv-cache-utilization-scorer
            - type: max-score-picker
          schedulingProfiles:
            - name: default
              plugins:
                - pluginRef: queue-scorer
                  weight: 2
                - pluginRef: kv-cache-utilization-scorer
                  weight: 2
                - pluginRef: prefix-cache-scorer
                  weight: 3
                - pluginRef: max-score-picker
```

## ConfigMap reference

```yaml
kind: LLMInferenceService
metadata:
  name: qwen2-7b-kv-cache-routing
spec:
  # ...
  router:
    # ...
    scheduler:
      config:
        ref:
          name: my-scheduler-config
          key: kv-queue-prefix-default
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-scheduler-configmap
data:
  kv-queue-prefix-default: |-
    kind: EndpointPickerConfig
    plugins:
       # ...
```

Fixes #4628


pkg/controller/v1alpha2/llmisvc/validation_int_test.go
```diff
@@ -402,6 +403,159 @@ var _ = Describe("LLMInferenceService API validation", func() {
...
+			llmSvc := fixture.LLMInferenceService("test-scheduler-config-valid-ref",
+				fixture.InNamespace[*v1alpha2.LLMInferenceService](nsName),
+				fixture.WithModelURI("hf://facebook/opt-125m"),
+			)
+			// Manually set scheduler config with valid ref
+			llmSvc.Spec.Router = &v1alpha2.RouterSpec{
+				Scheduler: &v1alpha2.SchedulerSpec{
+					Config: &v1alpha2.SchedulerConfigSpec{
+						Ref: &corev1.ConfigMapKeySelector{
+							LocalObjectReference: corev1.LocalObjectReference{Name: "test-config"},
```
spolti, line 536: Do we have a validation to fail beforehand if the configMap does not exist?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -245,6 +246,17 @@ type SchedulerSpec struct {
 	// This configures the Endpoint Picker (EPP) Deployment.
 	// +optional
 	Template *corev1.PodSpec `json:"template,omitempty"`
+
+	// Config is the configuration for the EndpointPicker.
+	Config *SchedulerConfigSpec `json:"config,omitempty"`
+}
+
+type SchedulerConfigSpec struct {
```
spolti, line 254: Shouldn't it be mutual exclusivity?
What happens if both are configured together?

cmd/llmisvc/main.go
```diff
@@ -143,36 +147,71 @@ func main() {
 		metricsServerOptions.FilterProvider = filters.WithAuthenticationAndAuthorization
 	}
 
+	llmSvcCacheSelector, _ := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app.kubernetes.io/part-of": "llminferenceservice",
```
spolti, line 152: Is this label automatically added by all the cached resources?
Also, would be nice to move it to a constant, wdyt?

hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
```diff
@@ -51697,6 +51735,25 @@ spec:
...
+                            type: object
+                            x-kubernetes-preserve-unknown-fields: true
+                          ref:
+                            properties:
+                              key:
+                                type: string
+                              name:
+                                default: ""
+                                type: string
+                              optional:
```
spolti, line 51776: What does this optional parameter mean here? Is it to define if the config can be skipped if the configMap does not exist?

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -296,6 +297,54 @@ func (l *LLMInferenceServiceValidator) validateImmutableParallelism(basePath *fi
...
+
+	if svc.Spec.Router.Scheduler.Config.Ref == nil && svc.Spec.Router.Scheduler.Config.Inline == nil {
+		allErrs = append(allErrs, field.Invalid(
+			configPath,
+			svc.Spec.Router.Scheduler.Config,
+			"either inline or ref is required",
+		))
+	}
+
+	if svc.Spec.Router.Scheduler.Config.Inline != nil && svc.Spec.Router.Scheduler.Config.Ref != nil {
```
spolti, line 319: can't this be moved to kubebuilder validation?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
spolti: ok, just saw the validator.

pkg/controller/v1alpha2/llmisvc/validation_int_test.go
pierDipi: No, resources can be created out of order and we don't want to reject/not support such cases

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: See the additional validation, it's mutually exclusive

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -405,3 +408,56 @@ func (r *LLMISVCReconciler) enqueueOnLLMInferenceServiceConfigChange(logger logr
...
+
+		for _, llmSvc := range llmSvcList.Items {
+			// Use WithSkipClearSchedulerConfigRef to preserve the Ref for matching
+			resolved, err := r.combineBaseRefsConfig(ctx, &llmSvc, cfg, WithSkipClearSchedulerConfigRef())
+			if err != nil {
+				logger.Error(err, "Failed to combine baseRefs config", "namespace", llmSvc.Namespace, "name", llmSvc.Name)
+				continue
+			}
+
+			if resolved.Spec.Router == nil ||
```
spolti, line 510: it seems similar ifs are used, can't we have a reusable predicate or function to match it?

hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
pierDipi: We inherit this from the core types, it's unused/ignored

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
pierDipi: Since we need additional validation logic (see below) I think concentrating the full logic here makes debugging/reasoning/troubleshooting easier

pkg/controller/v1alpha2/llmisvc/controller_int_scheduler_config_test.go
```diff
@@ -0,0 +1,859 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
+	. "github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc/fixture"
+)
+
+var _ = Describe("LLMInferenceService Scheduler Config", func() {
+	Context("Inline scheduler config", func() {
+		It("should use inline scheduler config in the scheduler deployment", func(ctx SpecContext) {
```
spolti, line 41: is there a test trying to set inline and ref parameters?

pierDipi: That type of test is in `pkg/controller/v1alpha2/llmisvc/validation_int_test.go` https://github.com/kserve/kserve/pull/4856/changes#diff-185ef7cd554cc0aa9f67d76f601663d226d9fa93cafacd30859e83da91acbb8f

pkg/controller/v1alpha2/llmisvc/controller.go
pierDipi: Maybe as a follow up to not touch unrelated code?

cmd/llmisvc/main.go
pierDipi: The expectation for the external (user-provided) ConfigMap(s) is to have this label if they want the controller to watch/react to them, controller created resources are using this label

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
```diff
@@ -177,3 +178,49 @@ func logLineForObject(obj client.Object) string {
...
+// Returns the object (either from cache or fallback) and any error.
+func Get[T client.Object](ctx context.Context, c client.Client, key client.ObjectKey, obj T, opts ...GetOption[T]) (T, error) {
+	options := &getOptions[T]{}
+	for _, opt := range opts {
+		opt(options)
+	}
+
+	if err := c.Get(ctx, key, obj); err != nil {
+		if !apierrors.IsNotFound(err) || options.fallback == nil {
+			return obj, fmt.Errorf("failed to get ConfigMap %s/%s from cached client: %w", key.Namespace, key.Name, err)
```
Jooho: This is a generic function, but the error message currently hardcodes "ConfigMap".

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
```diff
@@ -177,3 +178,49 @@ func logLineForObject(obj client.Object) string {
...
+		}
+		// Resource not in cache - try fallback.
+		r, err := options.fallback(ctx, key.Namespace, key.Name)
+		if err != nil {
+			return obj, fmt.Errorf("failed to get ConfigMap %s/%s from API Server: %w", key.Namespace, key.Name, err)
```
Jooho: This is a generic function, but the error message currently hardcodes "ConfigMap".

cmd/llmisvc/main.go
Jooho: I agree with this. That said, users will need to be aware of this label requirement, so it would be good to add it to the documentation.

Jooho (commented):
Not a big problem, just a minor comment.

cmd/llmisvc/main.go
spolti: or a mutating webhook

hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
spolti: ok

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
spolti: okay

pkg/controller/v1alpha2/llmisvc/controller.go
spolti: sounds good

spolti (approved):

cmd/llmisvc/main.go
pierDipi: touching user resources is not a good idea, this is to ensure memory usage doesn't grow with cluster side, let's keep it as is, it will still be used but we just don't react to changes

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
pierDipi: Fixed

cmd/llmisvc/main.go
bartoszmajsak: As a follow-up maybe we can move some labels to consts.

cmd/llmisvc/main.go
```diff
@@ -143,36 +147,68 @@ func main() {
...
+					Label: llmSvcCacheSelector,
+				},
+				&appsv1.Deployment{}: {
+					Label: llmSvcCacheSelector,
+				},
+				&corev1.Pod{}: {
+					Label: llmSvcCacheSelector,
+				},
+			},
+		},
```
bartoszmajsak, line 178: Good optimization! I wonder if we should cache other resources that we own.

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
bartoszmajsak, line 1: Very nice!

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -296,6 +297,54 @@ func (l *LLMInferenceServiceValidator) validateImmutableParallelism(basePath *fi
...
+		allErrs = append(allErrs, field.Invalid(
+			configPath,
+			svc.Spec.Router.Scheduler.Config,
+			"both inline and ref are set, either specify inline or ref",
+		))
+	}
+
+	if svc.Spec.Router.Scheduler.Config.Inline != nil && len(svc.Spec.Router.Scheduler.Config.Inline.Raw) < 3 /* we expect at least a few characters '{...}' */ {
```
bartoszmajsak, line 327: Maybe a silly question - we don't want to convert it to EPPConfig at this point to at least make sure what we passed is matching the expected kind?

bartoszmajsak (approved):
That's a quality of life improvement! 

/lgtm

But I left some comments for my better understanding

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
pierDipi: For now, the idea is to keep it open to any other format/version (including kind/group changes)

cmd/llmisvc/main.go
pierDipi: isn't that the default if unspecified?

spolti (approved):
lgtm

pierDipi:
@Jooho can we merge this one pls? there are constant conflicts to fix

terrytangyuan (approved):

Issue #4844, opened 2025-11-20
KServe quick install script error with Helm v4.0.0


changminbark:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
I tried to install kserve for local development and I am using kind (Ubuntu 24.04.3 LTS  ). I am following the instructions on the website but keep on running into this error. 

<img width="720" height="310" alt="Image" src="https://github.com/user-attachments/assets/9c5a9419-b364-4b9d-ba39-1fa035e558e3" />

I think that this is because the webhook server takes some time to initialize. After waiting about a minute, I ran the following command and it successfully installed the kserve resources:

```
helm upgrade kserve oci://ghcr.io/kserve/charts/kserve --version v0.16.0   --namespace kserve   --set-string kserve.controller.deploymentMode="RawDeployment"   --wait
```

On another note: the quick_install.sh -u also seems to be buggy as running quick_install.sh -r afterwards does not allow me to run the install again.

<img width="720" height="115" alt="Image" src="https://github.com/user-attachments/assets/a583730c-f690-4d3a-9276-7c5d654acd70" />

**What did you expect to happen:**
I expected running the script to install the main kserve helm (and therefore the ClusterServingRuntime resources).

**What's the InferenceService yaml:**
[To help us debug please run `kubectl get isvc $name -n $namespace -oyaml` and paste the output]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Istio Version:
- Knative Version: v1.16.0
- KServe Version:
- Kubeflow version:
- Cloud Environment:[k8s_istio/istio_dex/gcp_basic_auth/gcp_iap/aws/aws_cognito/ibm]
- Minikube/Kind version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

changminbark:
@Jooho This is the issue regarding the quick_install.sh script for creating a development environment.

Jooho:
@sivanantha321 will you push llmisvc chart with `kserve-llmisvc-crd` and `kserve-llmisvc-resource` ?

We are using `kserve-crd` and `kserve` so I think we should follow the same manner
- `kserve-llmisvc-crd`
- `kserve-llmisvc`

What do you think? @yuzisun

sivanantha321:
@Jooho We need to backport the renamed chart to 0.16 branch.

Jooho:
[...]

PR #4880, opened 2025-12-05
fix: the quick install scripts temporarily.

Jooho:
This is only a temporary workaround because the root issue is that the Helm chart creates `ClusterServingRuntime` and `LLMIsvcConfig` resources, which require the controller to be ready — but the controller itself is also managed by Helm, so we can’t guarantee the ordering.

To properly fix this, we should split those objects out of the `kserve-resources` / `kserve-llmisvc-resources` charts into separate charts. We should discuss this approach later.

With this PR, I addressed the following issues:

* **Helm CRD conflict (Helm v4):**
  Helm v3.16.3 had no issue applying the Gateway API/GIE CRDs before installing Envoy Gateway/Envoy AI Gateway. However, Helm v4.0.1 fails because the Envoy Gateway chart also includes the CRDs, causing a conflict.
  → I separated the Gateway API CRD and GIE CRD, and now only the GIE CRD is applied since the Envoy AI Gateway does not include it.

* **Unified install script:**
  Updated `quick_install.sh` and `llm_quick_install.sh` to use the new `hack/setup/infra/manage.kserve-helm.sh` script.
  → All installation scripts are now consolidated into a single script.

* **External LB script update:**
  Modified the external-lb setup script to use an absolute path (from the repository root: `kserve`) when calling the new quick-install script.

* **Retry mechanism for KServe installation:**
  Added a retry loop for installing KServe when `helm upgrade` fails.

 
**Verified platforms and scenarios**
- kind
  - `quick_install.sh -r` , `quick_install.sh -s`
  - `llm_quick_install.sh`
  - `llmisvc-full-install-with-manifests.sh`, `kserve-standard-mode-full-install-helm.sh`
- minikube
  - `quick_install.sh -r` , `quick_install.sh -s`
  - `llm_quick_install.sh`
  - `llmisvc-full-install-with-manifests.sh`, `kserve-standard-mode-full-install-helm.sh`

Fixes #https://github.com/kserve/kserve/issues/4844

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/hack/llmisvc_quick_install.sh
+++ b/hack/llmisvc_quick_install.sh
@@ -15,15 +15,14 @@ Help() {
    echo
 }
 
-export GATEWAY_API_VERSION=v1.2.1
-export KSERVE_VERSION=v0.16.0
-export LLMISVC_VERSION=v0.16.0
-export LWS_VERSION=0.7.0
-export ENVOY_GATEWAY_VERSION=v1.5.0
-export ENVOY_AI_GATEWAY_VERSION=v0.3.0
 SCRIPT_DIR="$(dirname -- "${BASH_SOURCE[0]}")"
-export CERT_MANAGER_VERSION=v1.16.1
 export SCRIPT_DIR
+
+source "${SCRIPT_DIR}/setup/common.sh"
+REPO_ROOT="$(find_repo_root "${SCRIPT_DIR}")"
+
+source "${REPO_ROOT}/kserve-deps.env"
+installKserve=true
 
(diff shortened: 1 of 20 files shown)


changminbark:
Tested the PR for llmisvc-full-install-with-manifests.sh and it works.

image

terrytangyuan (approved):

PR #4881, opened 2025-12-06
fix: make deploy-dev for development env

Jooho:
- Remove self-signed certificate part for deploy-dev target.
  - Since we already have a script that installs cert-manager, using it doesn’t add much cost. We can also use the new script when installing KServe, so keeping the self-signed certificate option doesn’t seem necessary. That’s why I think removing this option is the better choice.
- Add a new Makefile target `redeploy-dev-image` to build new images and redeploy them
- Add new scripsts
  - Install Kind cli(install-kind.sh)
    - Add kind cli version to kserve-deps.env
  - Deploy kind cluster with local registry using docker. (manage.kind-with-registry.sh)

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**


hack/setup/quick-install/llmisvc-dependency-install.sh
```diff
@@ -520,6 +520,7 @@ BLACK_FMT_VERSION=24.3
 FLAKE8_LINT_VERSION=7.1
 POETRY_VERSION=1.8.3
 UV_VERSION=0.7.8
+KIND_VERSION=v0.30.0
```
spolti, line 511: Can't we centralize these envs in the kserve-deps.env and load them where it is needed?

Jooho: Everything I’ve enhanced so far has been about centralizing things. The reason is that many scripts were using different versions of the same packages, which caused issues. So I believe it’s meaningful to consolidate all dependencies into a single file.

PR #4886, opened 2025-12-09
Bump Gateway API Inference Extension (GIE) to v1.2.0

### LINKED ISSUE #4823: Enhancement: Upgrade KServe LLMISVC to Gateway API Inference Extension v1

### LINKED ISSUE #4859: LLMIsvc: scheduler uses legacy flags (configFile, configText) — incompatible with llm-d ≥ 0.3.0
/kind bug

`LLMInferenceService` generates `llm-d-inference-scheduler` arguments using the old camelCase flags:

- `--configFile`
- `--configText`

Starting from `llm-d >= 0.3`, after syncing with IGW release `1.0.0-rc` in this PR:

- https://github.com/llm-d/llm-d-inference-scheduler/pull/320

the scheduler flags were switched to **kebab-case** (see the `kebab-case flags` commit in that PR):

- `--config-file`
- `--config-text`

As a result, the scheduler deployment generated by KServe LLMIsvc is incompatible with current `llm-d` releases:  
`llm-d-inference-scheduler` fails to start because it receives unknown/invalid flags.

`LLMInferenceService` should generate valid scheduler arguments so that  
`llm-d-inference-scheduler` (for llm-d >= 0.3) starts successfully using the kebab-case flags (`--config-file`, `--config-text`), and the InferenceService becomes Ready without manual patching.

The incompatible flags are constructed here:

- https://github.com/kserve/kserve/blob/1cb39eba0f1a18580a1c6cdaf106453a79cd4b7a/pkg/controller/v1alpha1/llmisvc/scheduler.go#L335-L346

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version:
- Kubeflow version:
- Minikube/Kind version:

sbekkerm:
## What this PR does / why we need it:

This PR upgrades Gateway API Inference Extension (GIE) to v1.2.0 and migrates the LLMInferenceService API from `serving.kserve.io/v1alpha1` to `serving.kserve.io/v1alpha2`.

### Background

The Gateway API Inference Extension (GIE) project released v1.2.0, which graduates their API from `api/v1alpha2` to `api/v1`. This is a breaking change in the upstream dependency that affects KServe's LLMInferenceService, which embeds GIE types like `InferencePoolSpec`.

Key upstream changes in GIE v1.2.0:
- API package moved from `api/v1alpha2` → `api/v1`
- API group graduated from `inference.networking.x-k8s.io` → `inference.networking.k8s.io`
- Resource renamed from `InferenceModel` → `InferenceObjective`

### Changes in this PR

**GIE Dependency Upgrade**
- Updated `gateway-api-inference-extension` from v0.3.0 to v1.2.0
- Updated import paths from `api/v1alpha2` to `api/v1`

**LLMInferenceService API Version Bump (v1alpha1 → v1alpha2)**
- The KServe API version bump reflects the breaking upstream GIE dependency change
- Migrated all controllers, clients, informers, and listers to v1alpha2
- Regenerated CRDs and OpenAPI schemas
- Added conversion webhooks for v1alpha1 ↔ v1alpha2 backward compatibility

**API Group Alignment**
- Updated HTTPRoute backendRef to use graduated API group `inference.networking.k8s.io`
- Updated all references from `inference.networking.x-k8s.io` to `inference.networking.k8s.io`

**RBAC Updates**
- Added permissions for both API groups (`k8s.io` and `x-k8s.io`) for backward compatibility
- Updated resource name from `inferencemodels` to `inferenceobjectives`

**InferencePool Readiness Check Improvement**
- Modified `IsInferencePoolReady()` to handle cases where no Gateway controller populates `status.Parents`
- The pool is now considered ready if it exists with a valid selector and target ports
- This fixes an issue where InferencePools were incorrectly marked as "not ready" when using Envoy Gateway with AI Gateway extension

**Workload Pod Label Fix**
- Added missing `app` label to workload pods for InferencePool selector matching
- The InferencePool selector expects pods with label `app: <name>-predictor`, but workload pods were previously missing this label

**llm-d Image Updates**
- Updated llm-d-cuda to v0.4.0
- Updated llm-d-inference-scheduler to v0.4.0
- Updated llm-d-routing-sidecar to v0.4.0

---

## Which issue(s) this PR fixes:

Fixes #4859
Fixes #4823

---

## Type of changes

### ⚠️ Breaking Change Notice

This PR changes the LLMInferenceService API version from `v1alpha1` to `v1alpha2`.

**Migration required:**

Before:
```yaml
kind: LLMInferenceService
```

After:
```yaml
kind: LLMInferenceService
```

**Backward Compatibility:** Conversion webhooks are included to automatically convert between v1alpha1 and v1alpha2. Existing v1alpha1 resources will continue to work, but users are encouraged to migrate to v1alpha2.

---

## Feature/Issue validation/testing:

- [x] Deploy LLMInferenceService with the updated API version (v1alpha2)
- [x] Verify HTTPRoute is accepted by Envoy Gateway
- [x] Integration tests updated and passing
- [x] `make precommit` passes

---

## Special notes for your reviewer:

**Why v1alpha2?** The LLMInferenceService API embeds types from GIE. When GIE graduated their API from `api/v1alpha2` to `api/v1`, it became a breaking change for KServe. The version bump to v1alpha2 signals this incompatibility to users.

**API Group Support:** Both `inference.networking.k8s.io` (graduated) and `inference.networking.x-k8s.io` (experimental) are supported in RBAC for environments that may still have the old CRDs.

**Conversion Webhooks:** This PR includes conversion webhooks that enable backward compatibility between v1alpha1 and v1alpha2. The v1alpha2 API is the hub (storage) version.

**InferencePool Readiness:** The `IsInferencePoolReady` fix addresses a real-world issue where Envoy Gateway + AI Gateway extension doesn't populate `status.Parents` on InferencePools, but the pool is still functional when the HTTPRoute resolves successfully.

**Pod Label Fix:** The addition of the `app` label to workload pods fixes InferencePool selector matching, ensuring traffic can be properly routed to predictor pods.

**Large diff warning:** The diff is large due to:
- Package rename from v1alpha1 to v1alpha2
- Regenerated client code, CRDs, and install manifests
- Updated GIE CRDs embedded in install scripts
- Conversion webhook code for backward compatibility

---

## Checklist:

---

## Release note

```release-note
Upgrade Gateway API Inference Extension to v1.2.0 and migrate LLMInferenceService API from v1alpha1 to v1alpha2. This is a breaking change - update your LLMInferenceService manifests to use `apiVersion: serving.kserve.io/v1alpha2`. Conversion webhooks provide backward compatibility for existing v1alpha1 resources. Also includes llm-d images update to v0.4.0 and fixes for InferencePool readiness checks and pod label selector matching.
```


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -219,8 +214,8 @@ type IngressSpec struct {
 //
 // The Scheduler is only effective when having multiple inference pod replicas.
 //
-// Step 1: Gateway (Envoy) &lt;-- ExtProc --&gt; EPP (select the optimal replica to handle the request)
-// Step 2: Gateway (Envoy) &lt;-- forward request --&gt; Inference Pod X
+// Step 1: Gateway (Envoy) <-- ExtProc --> EPP (select the optimal replica to handle the request)
```
sivanantha321: the comment were using html representation to avoid errors when auto generating api docs for the website.

sbekkerm: Thanks, addressed this in the latest commits

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -235,10 +248,6 @@ type SchedulerSpec struct {
 // InferencePoolSpec defines the configuration for an InferencePool.
 // 'Spec' and 'Ref' are mutually exclusive.
 type InferencePoolSpec struct {
-	// Spec defines an inline InferencePool specification.
-	// +optional
-	Spec *igwapi.InferencePoolSpec `json:"spec,omitempty"`
```
pierDipi, line 240: This is a breaking API change, we need to keep v1alpha1 with the same API structure as before in order to upgrade safely

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -35,23 +35,23 @@ import (
...
 
 	if err := r.reconcileSchedulerServiceAccount(ctx, llmSvc); err != nil {
 		return err
 	}
-	if err := r.reconcileSchedulerInferenceModel(ctx, llmSvc); err != nil {
-		return err
-	}
+	// if err := r.reconcileSchedulerInferenceModel(ctx, llmSvc); err != nil {
+	// 	return err
+	// }
```
pierDipi: This can be removed

pierDipi: Same for the related commented functions down

pkg/controller/v1alpha2/llmisvc/validation/config_validator.go
```diff
@@ -47,14 +47,14 @@ var _ webhook.CustomValidator = &LLMInferenceServiceConfigValidator{}
 
 func (l *LLMInferenceServiceConfigValidator) SetupWithManager(mgr ctrl.Manager) error {
```
pierDipi, line 51: we need to have/keep validators for each version, maybe a good idea to move the validators to apis/serving/v1alpha1/validation and apis/serving/v1alpha2/validation packages (separate packages from API types but versioned) ?

pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
```diff
@@ -239,18 +241,20 @@ func (r *LLMISVCReconciler) expectedMainMultiNodeLWS(ctx context.Context, llmSvc
...
 		"app.kubernetes.io/part-of":   "llminferenceservice",
 	}
 	if llmSvc.Spec.Prefill != nil && llmSvc.Spec.Prefill.Template == nil {
 		// When there is no leader template, workers become part of the InferencePool selector.
+		workerLabels["app"] = kmeta.ChildName(llmSvc.GetName(), "-predictor")
 		workerLabels["kserve.io/component"] = "workload"
 		workerLabels["llm-d.ai/role"] = "prefill"
 	}
 	leaderLabels := map[string]string{
+		"app":                         kmeta.ChildName(llmSvc.GetName(), "-predictor"),
```
pierDipi: These seems unrelated changes, maybe we can have a separate PR/follow up for this? 
Generally for LLMs "predictor" is a term we probably want to avoid

charts/kserve-llmisvc-resources/templates/webhookconfiguration.yaml
```diff
@@ -10,7 +10,7 @@ webhooks:
       service:
         name: llmisvc-webhook-server-service
         namespace: {{ .Release.Namespace }}
-        path: /validate-serving-kserve-io-v1alpha1-llminferenceservice
+        path: /validate-serving-kserve-io-v1alpha2-llminferenceservice
```
pierDipi, line 33: see comment on validating both versions

charts/kserve-llmisvc-resources/templates/webhookconfiguration.yaml
```diff
@@ -38,7 +39,7 @@ webhooks:
       service:
         name: llmisvc-webhook-server-service
         namespace: {{ .Release.Namespace }}
-        path: /validate-serving-kserve-io-v1alpha1-llminferenceserviceconfig
+        path: /validate-serving-kserve-io-v1alpha2-llminferenceserviceconfig
```
pierDipi, line 83: see comment on validating both versions

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	if src.Scheduler != nil {
+		dst.Scheduler = &SchedulerSpec{
+			Template: src.Scheduler.Template,
+		}
+		if src.Scheduler.Pool != nil {
+			dst.Scheduler.Pool = &InferencePoolSpec{
+				Ref: src.Scheduler.Pool.Ref,
+				// Note: v1alpha2 InferencePoolSpec.Spec is not converted (lossy)
```
pierDipi: See comment on API breaking change, we need lossless conversion

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.2/apix/v1alpha2/inferencepool_types.go#L51-L52

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
pierDipi: There is a conversion method here we could try to re-use https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.2/apix/v1alpha2/inferencepool_conversion.go#L29

pierDipi (commented):
Thanks! I did a first pass and left a few comments

For testing, we might want to also augment the testing matrix to test both versions end to end https://github.com/kserve/kserve/blob/f15792d42f31c499cbedc089b9b9b7f8fbcdf6e5/test/e2e/llmisvc/test_llm_inference_service.py#L63-L83

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -130,7 +130,7 @@ type LLMModelSpec struct {
 	// Criticality defines how important it is to serve the model compared to other models.
 	// This is used by the Inference Gateway scheduler.
 	// +optional
-	Criticality *igwapi.Criticality `json:"criticality,omitempty"`
+	Criticality *string `json:"criticality,omitempty"`
```
bartoszmajsak: Probably we want to keep the enum validation rules - reusing existing type sounds like a good idea here:

https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-0.5/api/v1alpha2/inferencemodel_types.go#L132

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	return dst
+}
+
+func convertModelSpecToV1Alpha2(src *LLMModelSpec) v1alpha2.LLMModelSpec {
+	dst := v1alpha2.LLMModelSpec{
+		URI:  src.URI,
+		Name: src.Name,
+		// Note: Criticality field is dropped in v1alpha2
```
bartoszmajsak: We are loosing this value when serving `v1alpha1`. One option we discussed in the past was to carry it using annotation so it will not get lost.

bartoszmajsak (commented):
Few minor things to follow-up.

It would be great to expand tests as [@pierDipi suggested](https://github.com/kserve/kserve/pull/4886#pullrequestreview-3570985094).

pkg/controller/v1alpha2/llmisvc/scheduler.go
natifridman: removed

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
natifridman: I pushed a proposal for using annotation in this [commit](https://github.com/kserve/kserve/pull/4886/commits/c62e0be1826ac5564c4a191db46e7c6127b1f0a7)

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
natifridman: It was dropped in [v1](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/release-1.2/api/v1), do we still want to use the enum?

bartoszmajsak: We do have validating webhook for v1alpha1 so probably won't harm to keep the contract. Maybe we can just copy this type internally.

natifridman: Done. I just copied it here

cmd/llmisvc/main.go
```diff
@@ -167,17 +167,29 @@ func main() {
...
-	llmConfigValidator := &llmisvcvalidation.LLMInferenceServiceConfigValidator{
-		ClientSet: clientSet,
+	// Register v1alpha1 validators
+	v1alpha1LLMValidator := &v1alpha1.LLMInferenceServiceValidator{}
+	if err = v1alpha1LLMValidator.SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "llminferenceservice-v1alpha1")
+		os.Exit(1)
+	}
+
+	v1alpha1ConfigValidator := &v1alpha1.LLMInferenceServiceConfigValidator{}
```
bartoszmajsak: Aren't we missing the validation logic from `pkg/controller/v1alpha1/llmisvc/validation/`? We should bring it back and update it for the new version. Up to you if you want to pass it as a closure or extract it as its own func. Let's also restore the tests to improve our safety net.

The duplication in `pkg/controller/v1alpha1/llmisvc/webhook/` should stay deleted. Not sure how that happened in the first place :)

bartoszmajsak (changes requested):
Thanks for the hard work, @natifridman and @sbekkerm. We're almost there!

It would also be good to re-run `kube:builder` generation to make sure our manifests reflect latest changes.

bartoszmajsak:
@yuzisun @terrytangyuan, as it's quite a significant piece of work, I would very much appreciate the review.

I also wonder if we can cut the release with those changes? it would help with the adoption, as we are behind with GIE/llm-d. If that's an option do we need release notes and docs update?

terrytangyuan:
> I also wonder if we can cut the release with those changes? it would help with the adoption, as we are behind with GIE/llm-d. If that's an option do we need release notes and docs update?

Yeah we can probably cut a release and will need some docs update. cc @yuzisun

cabrinha:
I've deployed this with the help of @andresllh and it seems to be working well.

Couple things I had to change in the Kustomize manifests:

- ValidatingWebhookConfigurations should only contain 1 item:
  - The old `name: llminferenceservice.kserve-webhook-server.v1alpha1.validator` should be removed entirely
  - The existing `ca-injection-patches` need to be updated to reflect the new name: https://github.com/kserve/kserve/blob/master/config/llmisvc/llmisvc_validatingwebhook_cainjection_patch.yaml#L8 (`llminferenceserviceconfig.kserve-webhook-server.v1alpha2.validator`, etc.)

terrytangyuan:
Thanks! @cabrinha 

@sbekkerm Could you address the comments and incorporate the additional changes mentioned by @cabrinha above?

cmd/llmisvc/main.go
natifridman: Thanks, I brought it back

natifridman:
Thanks @bartoszmajsak and @cabrinha for the comments!
Please let me know if I made the changes as expected

natifridman:
@terrytangyuan do we need to make the docs update as part of this PR?
If so, please let me know where and I'll try to make the changes

yuzisun:
> @terrytangyuan do we need to make the docs update as part of this PR? If so, please let me know where and I'll try to make the changes

Can you help update the dependency here https://kserve.github.io/website/docs/admin-guide/kubernetes-deployment-llmisvc?

yuzisun:
> > I also wonder if we can cut the release with those changes? it would help with the adoption, as we are behind with GIE/llm-d. If that's an option do we need release notes and docs update?
> 
> Yeah we can probably cut a release and will need some docs update. cc @yuzisun

Let's do a Christmas release for 0.17.0 :santa:

terrytangyuan:
Great idea

natifridman:
> Can you help update the dependency here https://kserve.github.io/website/docs/admin-guide/kubernetes-deployment-llmisvc?

@yuzisun sure. I created [this](https://github.com/kserve/website/pull/572) PR in the website repo

gavrissh:
> Let's do a Christmas release for 0.17.0 🎅

Agree, it is a good idea to get this in a release.

install/v0.16.0/kserve-cluster-resources.yaml
```diff
@@ -1592,7 +1592,7 @@ spec:
           - "9002"
           - --grpcHealthPort
           - "9003"
-          image: ghcr.io/llm-d/llm-d-inference-scheduler:v0.2.0
```
sivanantha321, line 1595: We don't need to modify the manifests under install directory. These are auto generated when creating a new release.

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -48,22 +56,22 @@ spec:
               successThreshold: 1
               timeoutSeconds: 1
             args:
-              - --poolName
+              - --pool-name
```
sivanantha321, line 59: This renaming should also be done in helm chart as well ?

install/v0.16.0/kserve-cluster-resources.yaml
natifridman: Thanks, reverted this change

config/llmisvcconfig/config-llm-scheduler.yaml
natifridman: Done

spolti:
Guess we also need to update this file https://github.com/kserve/kserve/blob/master/kserve-deps.env

kserve-deps.env
```diff
@@ -35,6 +35,6 @@ OPENTELEMETRY_OPERATOR_VERSION=0.113.0
 
 # LLMISvc dependencies
 LWS_VERSION=v0.7.0
-GATEWAY_API_VERSION=v1.2.1
-GIE_VERSION=v0.3.0
+GATEWAY_API_VERSION=v1.3.1-0.20251106052652-079e4774d76b
+GIE_VERSION=v1.2.0
```
natifridman, line 40: @spolti I think it was updated here. Is there anything else needs to be updated?

spolti: No, the file filter in the PR didn't show it for me, thanks for confirming :)

bartoszmajsak (commented):
I wonder if `v1alpha1` validation is working or is it bypassed? The `config/webhook/llmisvc/manifests.yaml` only lists `apiVersions: v1alpha2`. Either we need CRD conversion webhook enabled (so `v1alpha1` gets converted before validation), or separate webhook entries for `v1alpha1`. In the `main.go` we actually wire both webhooks so I am confused and maybe missing something here.

Did we test this path, e.g., by trying to create an invalid `v1alpha1` resource (e.g. wrong enum values as a simple unhappy path)?

natifridman:
I tested it after bringing back the webhook
```
kubectl apply -n nati-kserve-demo -f - <<EOF
kind: LLMInferenceService
metadata:
  name: invalid-parallelism
spec:
  model:
    uri: hf://facebook/opt-125m
    name: facebook/opt-125m
  replicas: 1
  parallelism:
    pipeline: 4
    data: 4
    dataLocal: 2
  router:
    route: {}
    gateway: {}
EOF
The LLMInferenceService "invalid-parallelism" is invalid: spec.parallelism: Invalid value: {"pipeline":4,"data":4,"dataLocal":2}: cannot set both pipeline parallelism and data parallelism (data or dataLocal) simultaneously
```

bartoszmajsak:
We could probably simplify the validation wiring in the main.go further but overall LGTM. 

I don't see e2e tests, but maybe we can do it as a follow-up?

WDYT @terrytangyuan @yuzisun ?

terrytangyuan:
Sounds good

go.mod
```diff
@@ -20,67 +20,80 @@ require (
...
+	k8s.io/kube-openapi v0.0.0-20250814151709-d7b6acb124c3
+	k8s.io/utils v0.0.0-20250820121507-0af2bda4dd1d
 	knative.dev/networking v0.0.0-20250117155906-67d1c274ba6a
 	knative.dev/pkg v0.0.0-20250117084104-c43477f0052b
 	knative.dev/serving v0.44.0
-	sigs.k8s.io/controller-runtime v0.21.0
-	sigs.k8s.io/gateway-api v1.2.1
-	sigs.k8s.io/gateway-api-inference-extension v0.3.0
+	sigs.k8s.io/controller-runtime v0.22.4
+	sigs.k8s.io/gateway-api v1.3.1-0.20251106052652-079e4774d76b
```
terrytangyuan, line 51: CI is failing because this tag is missing https://github.com/kubernetes-sigs/gateway-api/tags

terrytangyuan: > [INFO] Installing Gateway API CRDs v1.3.1-0.20251106052652-079e4774d76b... error: unable to read URL "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.3.1-0.20251106052652-079e4774d76b/standard-install.yaml", server reported 404 Not Found, status code=404

natifridman: @terrytangyuan This version is required for GIE https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/52ed3abfd4aa2374cc0f4ab5abe94675eb35311c/go.mod#L39

While the script is looking for Github releases

natifridman: Looks like there is a [PR](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/1998) to bump this dependency, but it's breaking some tests over there.
 I tried to use v1.4.1 in the go.mod, but I got errors that this [PR](https://github.com/kubernetes-sigs/gateway-api/pull/4357) in the gateway-api should fix

natifridman: In the meantime I can modify the `manage.gateway-api-crd.sh` script to find the next release (1.4.0) if it can't find the release in Github.
@terrytangyuan WDYT?

terrytangyuan: > In the meantime I can modify the manage.gateway-api-crd.sh script to find the next release (1.4.0) if it can't find the release in Github.

Sounds good

natifridman: I added this change, now it should fallback to the next release

charts/kserve-llmisvc-resources/templates/webhookconfiguration.yaml
```diff
@@ -12,7 +12,8 @@ webhooks:
         namespace: {{ .Release.Namespace }}
         path: /validate-serving-kserve-io-v1alpha1-llminferenceservice
     failurePolicy: Fail
-    name: llminferenceservice.kserve-webhook-server.validator
+    matchPolicy: Exact
+    name: llminferenceservice.kserve-webhook-server.v1alpha1.validator
```
pierDipi: by changing the name, an existing installation when upgrading will continue to have the webhook `llminferenceservice.kserve-webhook-server.validator` as a "ghost" (without match policy set), I think we need to try to keep the name for v1alpha1 validation(s) the same (for both llmisvc and llmisvc config down below)

sbekkerm: Good catch - fixed in the last commit

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
KillianGolds, line 1: Looks good overall. Nice clean standard Hub Spoke conversion pattern. My only minor concern is the silent error fallback on InferencePoolSpec conversion, but the comment acknowledges it shouldn't happen in practice so maybe just a nit.

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	// Use the built-in conversion by wrapping in temporary InferencePool objects
+	srcPool := &igwapiv1alpha2.InferencePool{Spec: *src}
+	dstPool := &igwapiv1.InferencePool{}
+
+	if err := srcPool.ConvertTo(dstPool); err != nil {
+		// Fallback: return empty spec on error (should not happen in practice)
+		return &igwapiv1.InferencePoolSpec{}
+	}
```
KillianGolds, line 329: Could the silent fallback hide potential issues ? Would logging/propagating it be worth it?

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
```diff
@@ -0,0 +1,471 @@
...
+
+	// Use the built-in conversion by wrapping in temporary InferencePool objects
+	srcPool := &igwapiv1.InferencePool{Spec: *src}
+	dstPool := &igwapiv1alpha2.InferencePool{}
+
+	if err := dstPool.ConvertFrom(srcPool); err != nil {
+		// Fallback: return empty spec on error (should not happen in practice)
+		return &igwapiv1alpha2.InferencePoolSpec{}
+	}
```
KillianGolds, line 400: Same as previous silent fallback comment.

pkg/controller/v1alpha2/llmisvc/scheduler.go
KillianGolds, line 1: Looking good. The changes are straightforward, moving from GIE v1alpha2 to v1, removing the InferenceModel (which GIE v1 doesn't use in the same way), and updating RBAC. The dual API group in RBAC is good for backward compatibility.

KillianGolds: ### Follow-up: Zero-Downtime Migration
This PR does a clean break to GIE v1. For production environments, a zero-downtime migration approach (dual-pool creation with weighted HTTPRoute backendRefs) would be a valuable follow-up contribution.

pkg/controller/v1alpha2/llmisvc/router_discovery.go
```diff
@@ -350,10 +351,17 @@ func nonReadyHTTPRouteTopLevelCondition(route *gwapiv1.HTTPRoute) (*metav1.Condi
...
+	if pool == nil {
 		return false
 	}
 
+	// If no parents have been set, consider the pool ready if it exists and has a valid spec
+	// This handles cases where no Gateway controller is populating the status
+	if len(pool.Status.Parents) == 0 {
+		// Pool is ready if it exists with a valid selector and target ports
+		return len(pool.Spec.Selector.MatchLabels) > 0 && len(pool.Spec.TargetPorts) > 0
+	}
```
KillianGolds, line 363: This seems like a reasonable fix for the InferencePool readiness check. If I'm understanding correctly, the alternative would be blocking users whose gateway controller doesn't populate `Status.Parents`, which seems worse.

One thing I noticed, with this fallback, a pool could briefly be considered "ready" before the gateway has actually processed it (since we're only checking the spec exists, not that the routing is set up). Probably a short window and acceptable tradeoff, but wanted to flag it. Let me know if I'm misunderstanding the scenario.

pkg/controller/v1alpha2/llmisvc/router.go
KillianGolds, line 1: Nice addition, validation failure shortcircuit in `EvaluateGatewayConditions` and `EvaluateHTTPRouteConditions` prevents overwriting validation errors.

KillianGolds (commented):
LGTM overall. The implementation is clean and aligns with similar work we did downstream. The conversion webhooks, RBAC changes, and InferencePool readiness fallback all look reasonable.

### Follow-up: Zero-Downtime Migration

This PR does a clean break to GIE v1 which is the right approach for simplicity. For production environments upgrading from existing v1alpha2 deployments, we might follow up with a zero-downtime migration strategy like I was working on downstream:

- Dual-pool creation (both v1 and v1alpha2 InferencePools)
- Readiness Check for Both Pools
- Weighted HTTPRoute backendRefs to gradually shift traffic
- One-way migration annotation to prevent flapping

Happy to contribute this as a follow-up PR

terrytangyuan (approved):

PR #4887, opened 2025-12-10
modelcache: add Namespace Scoped ModelCache 

VedantMahabaleshwarkar:
# What this PR does / why we need it : `Local Model Cache Enhancements`

### Proposal
The goal of this change is to simplify the process of caching models that require authentication (e.g., private HuggingFace models or S3 buckets) by adding an optional storage specification directly to the `LocalModelCache` CRD.

Previously, to cache a model with credentials, users were required to:
1. Create a `ClusterStorageContainer` CRD with `workloadType: localModelDownloadJob`.
2. Manually embed credentials (e.g., environment variables) into this container specification.
3. Manage secrets in a dedicated namespace (`kserve-localmodel-jobs`).

**Pros of the new approach:**
*   **Eliminates the need for `ClusterStorageContainer`:** Users no longer need to define a separate cluster-level resource just to handle download credentials.
*   **Consistency:** The configuration now mirrors the `InferenceService` storage configuration, providing a unified experience for defining storage access.
*   **Simplified Operations:** Credentials can be managed via `ServiceAccountName` or direct `Storage` parameters within the cache definition itself, reducing operational overhead.

### Explanation of Changes
*   **Updated `LocalModelCache` CRD:** Added `serviceAccountName` and `storage` fields to the `LocalModelCacheSpec`.
    *   `serviceAccountName`: Allows referencing a ServiceAccount (and its attached secrets) for credential lookup.
    *   `storage`: Allows specifying a storage `key` and `parameters` (e.g., endpoint, region) to override default configurations.
*   **Dynamic Download Job:** The model download job now dynamically configures its container based on the `storageInitializer` configuration found in the `inferenceservice-config` ConfigMap. It injects the necessary environment variables and volume mounts derived from the specified credentials.
*   **Backward Compatibility:** Existing `ClusterStorageContainer` configurations continue to function and take precedence if they exist.

### Proposal
The `LocalModelCache` resource is Cluster-scoped, making cached models available globally across the cluster. To support multi-tenancy and better isolation, we propose the `LocalModelNamespaceCache` CRD.

**Key Benefits:**
*   **Isolation:** Allows defining model caches that are restricted to a specific namespace.
*   **Security:** Ensures that sensitive models cached in a namespace are only usable by InferenceServices within that same namespace.
*   **Multi-tenancy:** Different teams (namespaces) can manage their own cache policies and models without affecting the cluster-wide state.

### Explanation of Changes
*   **New CRD `LocalModelNamespaceCache`:** Introduced a new namespace-scoped Custom Resource Definition.
    *   It shares the same specification structure as `LocalModelCache` (`SourceModelUri`, `ModelSize`, `NodeGroups`, `ServiceAccountName`, `Storage`).
*   **Controller Implementation:** Added a new reconciler to handle the lifecycle of `LocalModelNamespaceCache` resources. The controller manages the underlying `LocalModelNode` resources to ensure models are downloaded to the specified node groups.
*   **InferenceService Integration:** Updated the InferenceService logic to support resolving models from `LocalModelNamespaceCache`. When an InferenceService requests a model, the system checks for a matching cache in the local namespace.
*   **API Types:** Added corresponding Go types (`LocalModelNamespaceCache`, `LocalModelNamespaceCacheList`) and generated client code.

### Workflow Differences: Cluster vs. Namespace Cache
The workflow for `LocalModelNamespaceCache` differs from `LocalModelCache` in several key ways to support isolation:

1.  **Job Execution Context:**
    *   **LocalModelCache:** Download jobs run in a centralized namespace (default `kserve-localmodel-jobs`).
    *   **LocalModelNamespaceCache:** Download jobs run **in the same namespace** where the CR is defined. This ensures that the job uses the namespace's quota and service accounts.

2.  **Resolution Scope:**
    *   **LocalModelCache:** Available to any InferenceService in the cluster.
    *   **LocalModelNamespaceCache:** Only available to InferenceServices within the **same namespace**. The webhook logic prioritizes checking the local namespace cache before falling back to the cluster cache.

3.  **Credential Access:**
    *   **LocalModelCache:** Credentials (Secrets/ServiceAccounts) must exist in the centralized job namespace.
    *   **LocalModelNamespaceCache:** Credentials must exist in the user's namespace, allowing teams to manage their own access tokens securely.

### LocalModelNode CR and Status Updates
The `LocalModelNode` CR, which tracks the models present on a specific node, has been updated to handle both cluster-scoped and namespace-scoped models.

1.  **Spec Updates:**
    *   The `LocalModelInfo` struct in `LocalModelNodeSpec` now includes a `Namespace` field.
    *   If `Namespace` is empty, it represents a cluster-scoped `LocalModelCache`.
    *   If `Namespace` is set, it represents a `LocalModelNamespaceCache`.
    *   Credential fields (`ServiceAccountName`, `Storage`) are also passed down to the `LocalModelNode` to ensure the node agent can launch the download job with correct permissions.

2.  **Status Key Changes:**
    *   To prevent name collisions between models with the same name in different namespaces, the status map keys have been updated:
        *   **Cluster-scoped models:** Key remains `modelName`.
        *   **Namespace-scoped models:** Key is now `namespace/modelName`.
    *   This allows the node agent to track the status of `my-model` (cluster) and `my-ns/my-model` (namespace) independently.

3.  **Storage Deduplication & Status:**
    *   The node agent uses a **hash of the `SourceModelUri`** to determine the physical storage folder on the node.
    *   **Deduplication:** If multiple caches (e.g., in different namespaces) point to the exact same `SourceModelUri`, they will share the same physical folder on the disk.
    *   **Smart Status Updates:** The agent tracks which storage keys have been processed. If a download for a specific URI is already in progress or completed (initiated by another CR), the agent reuses that status for the current model. This prevents redundant download jobs while correctly updating the status of all referencing `LocalModelNode` entries.

-------------------------------
**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:
```release-note
Add namespace scoped LocalModelNamespaceCache CRD for caching models in a namespaced scope
```

**Re-running failed tests**


terrytangyuan:
I wonder if we should use these names instead:
* ModelCache (namespaced)
* ClusterModelCache (cluster scoped)

Similar to what we are doing in Argo Workfows, e.g. WorkflowTemplate vs ClusterWorkflowTemplate

Although the renaming might be a breaking change.

If not, LocalModelNamespacedCache might be better than LocalModelNamespaceCache.

terrytangyuan:
cc @yuzisun @greenmoon55 @johnugeorge @cjohannsen-cloudera Could you take a look at this proposal?

yuzisun:
@VedantMahabaleshwarkar If the intent is to cache the model on each node in the cluster, how do you control the access to the model cache to only allow a specific namespace.

VedantMahabaleshwarkar:
@yuzisun 
> If the intent is to cache the model on each node in the cluster, how do you control the access to the model cache to only allow a specific namespace. 

The [existing](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_defaults.go#L468) LocalModelCache logic checks for a StorageUri match with all existing LocalModelCache CRs (which are clusterwide), and adds a ModelCache label (`internal.serving.kserve.io/localmodel: my-model`) if a match is found, which triggers the logic to inject the cache PV/PVC to the deployment.  

For the LocalModelNamespaceCache, the StorageUri matching logic is [modified](https://github.com/kserve/kserve/pull/4887/changes#diff-8e905aab2189585cddda5b3cf02ff241118450135c912dce14b93ecefde4c68fR478) as follows : 
The ModelCache labels are added only if a matching StorageURI is found in either 
- A LocalModelNamespaceCache CR in the same namespace as the ISVC (First precedence, check for namespaced cache) 
  - This adds an additional `internal.serving.kserve.io/localmodel-namespace: <namespace>` label which is checked by the LocalModelNamespaceCache Reconciler while injecting the PV and PVC.  
- A LocalModelCache CR (second priority, check for global cache if no namespaced cache is found)

docs/samples/localmodelcache/README.md
```diff
@@ -0,0 +1,250 @@
...
+  persistentVolumeClaimSpec:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 100Gi
+    storageClassName: local-storage
+  persistentVolumeSpec:
+    accessModes:
+      - ReadWriteOnce
```
spolti, line 204: Shouldn't it be RWX?

spolti:
Wouldn't it be better move the discussion to a GH issue and attach the architecture proposal?

greenmoon55:
Looks great! We thought about something similar. One follow up could be validating disk quota to make sure disk space is not used up by some namespaces.

spolti:
@VedantMahabaleshwarkar can you please rebase the PR and fix the conflicts?

pkg/apis/serving/v1alpha1/local_model_namespace_cache_types.go
```diff
@@ -0,0 +1,91 @@
...
+	Items           []LocalModelNamespaceCache `json:"items" validate:"required"`
+}
+
+func init() {
+	SchemeBuilder.Register(&LocalModelNamespaceCache{}, &LocalModelNamespaceCacheList{})
+}
+
+// MatchStorageURI checks if the storageUri from inference service matches the sourceModelUri
+// of the LocalModelNamespaceCache or is a subdirectory of the sourceModelUri
+func (spec *LocalModelNamespaceCacheSpec) MatchStorageURI(storageUri string) bool {
```
spolti, line 71: This seems to be identical with [local_model_cache_types.go](https://github.com/kserve/kserve/pull/4887/changes#diff-162f91c8d288931027b8295011dc6762a955b97572017bc98bc05e945a22ddcd) `LocalModelCacheSpec.MatchStorageURI`

let's extract to a common function and reuse as necessary.

pkg/apis/serving/v1beta1/inference_service_defaults_test.go
```diff
@@ -343,7 +343,7 @@ func TestInferenceServiceDefaults(t *testing.T) {
 	for _, scenario := range scenarios {
 		resources := corev1.ResourceRequirements{Requests: defaultResource, Limits: defaultResource}
 		scenario.isvc.Spec.DeepCopy()
-		scenario.isvc.DefaultInferenceService(scenario.config, scenario.deployConfig, nil, nil)
+		scenario.isvc.DefaultInferenceService(scenario.config, scenario.deployConfig, nil, nil, nil)
```
spolti, line 346: Does it make sense to add namespace-scoped label logic tests?

cmd/localmodel/main.go
```diff
@@ -151,6 +151,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup LocalModelNamespaceCache controller
+	setupLog.Info("Setting up v1alpha1 LocalModelNamespaceCache controller")
+	if err = (&localmodelcontroller.LocalModelNamespaceCacheReconciler{
+		Client:    mgr.GetClient(),
+		Clientset: clientSet,
+		Log:       ctrl.Log.WithName("v1alpha1Controllers").WithName("LocalModelNamespaceCache"),
+		Scheme:    mgr.GetScheme(),
```
spolti, line 148: shouldn't we register the `LocalModelNamespaceCacheValidator` here as well?

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelcache_reconciler.go
```diff
@@ -0,0 +1,326 @@
...
+			return true
+		},
+	}
+
+	nodePredicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Only reconciles the local model crs when the node becomes ready from not ready
+			// Todo: add tests
+			oldNode := e.ObjectNew.(*corev1.Node)
+			newNode := e.ObjectNew.(*corev1.Node)
```
spolti: Both use ObjectNew, so readiness transition check is always false. Should be oldNode := e.ObjectOld.

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
```diff
@@ -0,0 +1,326 @@
...
+				return false
+			}
+			return true
+		},
+	}
+
+	nodePredicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldNode := e.ObjectNew.(*corev1.Node)
+			newNode := e.ObjectNew.(*corev1.Node)
```
spolti: same here.

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
```diff
@@ -0,0 +1,326 @@
...
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	controllerBuilder := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.LocalModelNamespaceCache{}).
+		Owns(&corev1.PersistentVolume{}).
```
spolti: is it needed? I guess namespaced PVs do not have owner ref.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -160,18 +205,79 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, localModelNode
...
+		Name:                     DownloadContainerName,
+		Image:                    image,
+		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+	}
+
+	// Set resource limits/requests from config if available
+	if config != nil {
+		container.Resources = corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse(config.CpuLimit),
```
spolti: Will panic on invalid config values. Could use ParseQuantity with error handling.

test/e2e/modelcache/test_localmodelnamespacecache.py
```diff
@@ -0,0 +1,188 @@
...
+        "minikube-m02",
+    )
+    worker_node_2_cache = k8s_client.get_cluster_custom_object(
+        constants.KSERVE_GROUP,
+        constants.KSERVE_V1ALPHA1_VERSION,
+        constants.KSERVE_PLURAL_LOCALMODELNODE,
+        "minikube-m03",
+    )
+    assert (
+        worker_node_1_cache["status"]["modelStatus"][model_cache.metadata.name]
```
spolti: Is this correct? wondering if we shouldn't use the `namespace/modelName` due https://github.com/kserve/kserve/pull/4887/changes#diff-cdf83e8ce1e5b3c7d3f22de6cb1d59a2db900cef34c302e9618eb14d051ad1c5R40

spolti (changes requested):

VedantMahabaleshwarkar:
@yuzisun hey, checking in here to see if your [question](https://github.com/kserve/kserve/pull/4887#issuecomment-3647265387) was answered [here](https://github.com/kserve/kserve/pull/4887#issuecomment-3647265387)

yuzisun (approved):

spolti (approved):
/lgtm

terrytangyuan (approved):

Issue #4798, opened 2025-11-05
Broken patch and delete with new release


balazsprehoda:
/kind bug

**What steps did you take and what happened:**
We upgraded kserve from **v0.15.2** to **v0.16.0**. Inference services that were created with the old version cannot be patched or deleted. These had the deployment mode annotation and the status set to `'Serverless'`.
Logs from the controller:
```
"error":"admission webhook \"inferenceservice.kserve-webhook-server.validator\" denied the request: update rejected: deploymentMode cannot be changed from 'Knative' to 'Serverless'","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.1/pkg/internal/controller/controller.go:224"
```

**What did you expect to happen:**
The patch and the delete should work. Existing deployment mode annotations and status fields should be updated to use the new naming (`Standard` or `Knative`).

**Environment:**

- Knative Version: 1.19.0
- KServe Version: 0.16.0
- Cloud Environment: AWS EKS
- Kubernetes version: (use `kubectl version`): v1.33.5-eks-113cf36

balazsprehoda:
I think this issue is fixed by #5025

PR #4800, opened 2025-11-06
fix: broken patch and delete isvc

andyi2it:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** **:
Fixes #4798

**Type of changes**
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

**Feature/Issue validation/testing**:

- Logs

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

**Release note**:
```release-note
NONE
```

**Re-running failed tests**

- `/rerun-all` - rerun all failed workflows.
- `/rerun-workflow <workflow name>` - rerun a specific failed workflow. Only one workflow name can be specified. Multiple /rerun-workflow commands are allowed per comment.

PR #4896, opened 2025-12-13
fix: isvc deletion using old label

jokestax:

When an InferenceService resource is created using the deprecated deploymentMode values RawDeployment or Serverless, deleting the resource can fail. This happens because the InferenceService webhook validator compares the deployment mode stored in annotations with the value in status. While status is updated to use the new names (Standard and Knative), the annotations may still contain the old values, causing the validation to fail during deletion.

This PR fixes the issue by normalizing deprecated deployment mode values to their corresponding new names before performing the comparison. As a result, resources created with legacy deployment modes can now be validated and deleted correctly.

**Type of changes**

- [ ] Test A
Steps to Reproduce

Create an InferenceService using the following manifest, which specifies the deprecated deployment mode RawDeployment via annotations:

```
kind: InferenceService
metadata:
  annotations:
  name: qwen-llm4
  namespace: kserve
spec:
  predictor:
    runtimeClassName: nvidia
    model:
      modelFormat:
        name: huggingface
      storageUri: hf://Qwen/Qwen2.5-0.5B-Instruct
      args:
        - --served-model-name=qwen
      env:
        - name: VLLM_LOGGING_LEVEL
          value: DEBUG
      resources:
        limits:
          cpu: "2"
          memory: 6Gi
          nvidia.com/gpu: "1"
        requests:
          cpu: "1"
          memory: 4Gi
          nvidia.com/gpu: "1"
```
and try to delete using cmd `kubectl delete isvc qwen-llm4 -n kserve`

we can bserve that the deletion is denied by the validating webhook. The `kserve-controller-manager` logs show the following error:

`"error":"admission webhook \"inferenceservice.kserve-webhook-server.validator\" denied the request: update rejected: deploymentMode cannot be changed from 'Standard' to 'RawDeployment'"`

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -495,6 +495,12 @@ func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService
 	if len(statusDeploymentMode) != 0 {
 		annotations := newIsvc.Annotations
 		annotationDeploymentMode, ok := annotations[constants.DeploymentMode]
+		if annotationDeploymentMode == string(constants.LegacyRawDeployment) {
+			annotationDeploymentMode = string(constants.Standard)
+		}
+		if annotationDeploymentMode == string(constants.LegacyServerless) {
+			annotationDeploymentMode = string(constants.Knative)
+		}
 		if ok && annotationDeploymentMode != statusDeploymentMode {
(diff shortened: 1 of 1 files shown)


balazsprehoda:
This is related to #4800, which aims to fix #4798. Since that issue is a blocking the kserve upgrade on our end, I'm really looking forward to a fix.

sivanantha321:
addressed in https://github.com/kserve/kserve/pull/5025. Thanks!

PR #4910, opened 2025-12-25
fix: watch pod init container status changes

jlost:

Init container errors (e.g., invalid S3 credentials) take 10 minutes to propagate to the InferenceService status. This is because the controller watches the Deployment, but Deployment status doesn't update until `progressDeadlineSeconds` expires (default 600s).

This PR adds a pod watch to the InferenceService controller to react immediately when init container status changes, dramatically reducing the time to surface common errors such as invalid S3 credentials to users, tests, and other components.

**Changes:**
- Add `podInitContainersFunc` mapper to route pod events to the owning InferenceService based on the `serving.kserve.io/inferenceservice` label
- Add `podInitContainersPredicate` to filter pod updates to only process init container status changes, preventing event storms from unrelated pod status updates (e.g., main container readiness changes)
- Add `.Watches` on pods with the above mapper and filter
- Refactor `servingRuntimesPredicate` and `clusterServingRuntimesPredicate` from inline variables to standalone functions for consistency

**Additional Fixes**
While testing, I discovered and fixed two bugs, which are now covered by the test suite:
* If the InitContainer of a Predictor failed while in Gateway API mode, the `ISVC.status` would never become populated (the isvc would become "stuck").
* If a ClusterServingRuntime were updated, reconciliation would be triggered for ALL isvcs in the cluster (with `DisableAutoUpdateAnnotatationKey == false`), rather than only the ones that use that runtime.

**Design Considerations**
An alternate approach would be to use `ReconcileAfter` if pods are the isvc is starting -- essentially polling-based -- simpler, but with more noop reconciliations and/or slower updates.

**Performance**
The cache is filtered and the predicate is tight. Cache size is O(n) where n is number of ISVC pods (not all pods in cluster). Benchmarks show the filter takes ~1.8µs per event. Events are triggered every time an ISVC pod updates in the cluster.

**Type of changes**

- [x] Unit tests for `podInitContainersFunc` mapper function
- [x] Unit tests for `podInitContainersPredicate` predicate function  
- [x] Tests verifying event storm prevention (pod updates only reconcile the specific owning InferenceService, not unrelated ones)
- [x] e2e tests to verify overall functionality in Standard and Serverless modes and all network_layers

**Checklist**:

```release-note
Fix slow propagation of init container errors (e.g., storage-initializer failures) to InferenceService status by watching pod init container status changes directly
Fix `ISVC.status` never becoming populated if the InitContainer of a Predictor failed while in Gateway API mode
Fix ClusterServingRuntime changes triggering reconciliation of all InferenceServices in the cluster, rather than only for that runtime.
```


pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -478,16 +482,94 @@ func (r *InferenceServiceReconciler) clusterServingRuntimeFunc(ctx context.Conte
 				continue
 			}
 		}
-		requests = append(requests, reconcile.Request{
-			NamespacedName: types.NamespacedName{
-				Namespace: isvc.Namespace,
-				Name:      isvc.Name,
-			},
-		})
+		if isvc.Status.ClusterServingRuntimeName == clusterServingRuntimeObj.Name {
```
spolti, line 485: Why enforcing ClusterServingRuntime type?

jlost: This mapper will only be called for ClusterServingRuntimes [(see watch)](https://github.com/kserve/kserve/pull/4910/changes/BASE..785d14cdd8df165650dcc933999dac0e1c546efa#diff-f14605793ed521dabc5c4794584c2961365d2d8324ad41f009aa44b1513d50feR670)
For ServingRuntimes, we have a different mapper.

jlost: I added this line because if a ClusterServingRuntime were updated, reconciliation would be triggered for ALL isvcs in the cluster (with `DisableAutoUpdateAnnotatationKey == false`), rather than only the ones that use that runtime.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -478,16 +482,94 @@ func (r *InferenceServiceReconciler) clusterServingRuntimeFunc(ctx context.Conte
...
+func podInitContainersPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Only process pods managed by InferenceServices
+			newPod, ok := e.ObjectNew.(*corev1.Pod)
+			if !ok || newPod == nil {
+				return false
+			}
+			// Check if pod has the InferenceService label
+			_, found := newPod.Labels[constants.InferenceServicePodLabelKey]
```
spolti: can be resumed to:
```go
  if _, found := newPod.Labels[constants.InferenceServicePodLabelKey]; !found {
          return false
      }
```

jlost: Fixed

Jooho:
This PR looks good to me. One note: this change appears to cover only InferenceService (ISVC). I believe we will also need corresponding updates for LLMISVC. That can be addressed in a follow-up PR.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -478,16 +482,93 @@ func (r *InferenceServiceReconciler) clusterServingRuntimeFunc(ctx context.Conte
...
+		},
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}
+}
+
+// clusterServingRuntimesPredicate returns a predicate that filters ClusterServingRuntime updates
+// to only include those where the Spec has changed.
+// TODO: Find a way to distinguish if the ServingRuntime is a ClusterServingRuntime or not
```
brettmthompson: Are these not two distinct resource types?

Nevertheless, would you be able to tell by the namespace, i.e. ServingRuntimes are namespaced resources whereas ClusterServingRuntimes are not

brettmthompson: I see this is just an existing comment you copied, so my comment here may not be relevant to your work

jlost: This code block was moved from one location to another. The comment was copied too. But yes, I think the comment is cruft.

jlost: It looks like I'll need to update the branch and rerun the tests anyway, so I'll remove that comment... :)

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -478,16 +482,93 @@ func (r *InferenceServiceReconciler) clusterServingRuntimeFunc(ctx context.Conte
...
+			if !ok || newPod == nil {
+				return false
+			}
+			// Check if pod has the InferenceService label
+			if _, found := newPod.Labels[constants.InferenceServicePodLabelKey]; !found {
+				return false
+			}
+			// Only watch pod status changes, not spec changes
+			oldPod := e.ObjectOld.(*corev1.Pod)
+			return !equality.Semantic.DeepEqual(oldPod.Status.InitContainerStatuses, newPod.Status.InitContainerStatuses)
```
brettmthompson, line 563: Are we certain that `InitContainerStatuses` is the only status field that has a delay in propagation to the isvc? I am just wondering if there would be any benefit to checking the entire `.Status` object instead of isolating the check to only `.Status.InitContainerStatuses`? 

Other side of the coin would be that checking the entire `.Status` object could lead to excessive/unneeded reconciliation of isvcs.

Either way we can easily extend this solution to other fields if needed in the future.

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -478,16 +482,93 @@ func (r *InferenceServiceReconciler) clusterServingRuntimeFunc(ctx context.Conte
...
+func podInitContainersPredicate() predicate.Funcs {
+	return predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Only process pods managed by InferenceServices
+			newPod, ok := e.ObjectNew.(*corev1.Pod)
+			if !ok || newPod == nil {
+				return false
+			}
+			// Check if pod has the InferenceService label
+			if _, found := newPod.Labels[constants.InferenceServicePodLabelKey]; !found {
```
brettmthompson: nit: since we also check that `isvcName != ""` in `podInitContainersFunc` (and exit early if it does), adding that same check here could potentially save some unnecessary calls to `podInitContainersFunc`.

pkg/controller/v1beta1/inferenceservice/controller.go
jlost: There could be more situations in which we'd want a fast update. I was being abundantly conservative regarding performance, bearing in mind that we could easily expand the scope later.

pkg/controller/v1beta1/inferenceservice/controller.go
jlost: Done!

Jooho (approved):
/lgtm
/approved

PR #4923, opened 2026-01-02
fix: install script fixes for git worktree support

jlost:

Fixes issues in the install scripts:

1. **Git worktree support**: The `find_repo_root` function used `-d` to check for `.git` directory, but in git worktrees `.git` is a file (not a directory) containing a reference to the main repository. Changed to `-e` (exists) to support both cases.

**Type of changes**

- Tested install scripts from a git worktree

**Checklist**:

```release-note
Fix install scripts to work in git worktrees
```


hack/setup/common.sh
```diff
@@ -27,7 +27,7 @@ find_repo_root() {
     local skip="${2:-false}"
 
     while [[ "$current_dir" != "/" ]]; do
-        if [[ -d "${current_dir}/.git" ]]; then
+        if [[ -e "${current_dir}/.git" ]]; then
```
spolti, line 30: Interesting, why checking for dir, file, symlinks, etc?

Jooho: This part is used to locate the top-level directory in KServe. I’m using the presence of the .git folder to determine that we’ve reached the repo root, using the -d test.  For context, I’ve always assumed a standard git clone of the KServe repository as the baseline environment.

I think I heard the reason why you want to change this, but could you please state here?

jlost: When using git worktrees, .git is a file (not a directory) containing a reference to the main repository. 
Example: 
```
❯ cat .git   
gitdir: /home/jostrand/projects/kserve/.git/worktrees/kserve-RHOAIENG-41816
```

Git worktrees are used by AI agents for parallel work, but they're also generally useful to keep your work organized. I'm experimenting with it for workflow automation. In my environment, 1 worktree == 1 JIRA. The script as-is prevents me from bootstrapping an environment outside of the main repo.

Jooho (approved):
/lgtm
/approved

Jooho:
@jlost can you please rebase ?

Issue #4251, opened 2025-02-15
Enhancing KServe and MLflow Integration for Improved Usability


anencore94:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

This project aims to improve the usability and integration between KServe and MLflow, making it easier for users to serve and manage machine learning models efficiently who uses MLflow. The proposed enhancements focus on:

1. Direct MLflow Model Download in KServe
- Expanding the existing KServe interface to support direct model downloads from the MLflow Tracking Server via the mlflow.download() API.
- Similar to the Hugging Face Hub, users should be able to specify tracking_uri and run_id to seamlessly fetch models, rather than manually using an S3 path URL from the MLflow artifact store.
- Reference: [Issue #3555](https://github.com/kserve/kserve/issues/3555)

2. Automating Conda Packaging & Unpacking for the KServe-MLflow Serving Runtime
- The MLServer runtime in KServe relies on Conda for environment management.
- Automating the packaging and unpacking process will simplify deployment and reduce manual overhead for users.
- References: [Discussion #4100](https://github.com/kserve/kserve/discussions/4100), [Issue #4122](https://github.com/kserve/kserve/issues/4122), [Issue #830](https://github.com/kserve/kserve/issues/830)

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

I would be happy to work on this feature natively in KServe if the community finds it valuable.

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

frosk1:
I would add a change for the way how to bring in models from mlflow Backend. It should be possible to use model namen with version or alias to download it. Not only tracking_Uri and run_id. But it might be contradicting with kube flow strategy, since the project is trying to bring in its own model registry in the mix.

mariusdanciu:
This would be a good feature.

lizzzcai:
There is a [MLflow integration into KubeFlow](https://github.com/kubeflow/community/pull/892).
[...]

PR #4925, opened 2026-01-04
feat: add support for MLFlow in kserve_storage

moulin-louis:

Working torward fixing #4251

**Type of changes**:

     - This PR does not change any image versions, it only adds MLflow storage support.

 This is an initial implementation.
 My goal is to better integrate MLFlow to KServe by allowing usage of MLFlow Registered Models (https://mlflow.org/docs/latest/ml/model-registry/#concepts) and MLFlow webhooks (https://mlflow.org/docs/latest/ml/webhooks/) for automatic redeployment.
 
 Maybe usage 

  **Release note**:
  ```release-note
  Add MLflow storage support for loading models from MLflow Model Registry using mlflow:// URIs
```

**Re-running failed tests**

### CODE REVIEW

**Comment by moulin-louis** [comment]
Just push the changes from 'make precommit'

File: python/storage/kserve_storage/kserve_storage.py [hunk 1]
```diff
@@ -67,6 +67,8 @@
 _AZURE_MAX_FILE_CONCURRENCY = int(os.getenv("AZURE_MAX_FILE_CONCURRENCY", "4"))
 _AZURE_MAX_CHUNK_CONCURRENCY = int(os.getenv("AZURE_MAX_CHUNK_CONCURRENCY", "4"))
 
+_MLFLOW_PREFIX = "mlflow://"
```
spolti: Could you please move together with the other prefixes?

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -932,3 +953,33 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
             ) from e
         os.remove(file_path)
         return target_dir
+
+    @staticmethod
+    def _download_mlflow(uri, out_dir: str) -> str:
```
spolti, line 1286: can you please add some tests?

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -170,6 +181,16 @@ def _update_with_storage_spec():
                     f.write(value)
                     f.flush()
 
+        if storage_secret_json.get("type", "") == "mlflow":
+            for env_var, key in (
+                ("MLFLOW_TRACKING_URI", "tracking_uri"),
```
spolti, line 287: Also, can you please document these envs in the README here in this repo, plus add documentation for this in the official docs?

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -932,3 +953,33 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
...
+            raise ValueError(
+                "Tracking Token cannot be set with Username/Password combo"
+            )
+
+        models = uri[len(_MLFLOW_PREFIX) :]
+        logger.info(f"Downloading {models} from {mlflow_tracking_uri}")
+        if not models:
+            raise ValueError("Model uri cannot be empty")
+        try:
+            mlflow.artifacts.download_artifacts(artifact_uri=models, dst_path=out_dir)
```
spolti: what happens when large models are downloaded, will it put everything in memory? Or download it in chunks and persist it on disk to save memory?
Considering the storage-initializer should use a very small amount of memory.

python/storage/kserve_storage/kserve_storage.py
moulin-louis: Done !

python/storage/kserve_storage/kserve_storage.py
moulin-louis: MLFlow handle downloading artifacts chunk by chunk when possible.

You can see the implementation here for HTTP Artifact Store: https://github.com/mlflow/mlflow/blob/master/mlflow/store/artifact/http_artifact_repo.py#L102.

For S3 Artifact store they use boto3 S3 Transfer which stream the artifacts as well: https://github.com/mlflow/mlflow/blob/master/mlflow/store/artifact/s3_artifact_repo.py, https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.S3Transfer.download_file

python/storage/kserve_storage/kserve_storage.py
moulin-louis: Done for the README, is the official docs located on this repo or this one https://github.com/kserve/website ?

python/storage/kserve_storage/kserve_storage.py
moulin-louis: Just pushed `test_mlflow_storage.py` in the test folder.

pkg/credentials/mlflow/mlflow_secret.go
```diff
@@ -0,0 +1,56 @@
...
+*/
+
+package mlflow
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	MLFlowTrackingUri      = "MLFLOW_TRACKING_URI"      // #nosec G101
```
spolti: Are this envs? If so, append `Env` at the end of the const name.
e.g. `MLFlowTrackingUriEnv`

moulin-louis: Good catch,  just push the update with the `Env` sufix.

spolti:
Hi, can you please review the test failures?

@Jooho @sivanantha321  do you mind taking a look?

moulin-louis:
Should be fixed now @spolti .

moulin-louis:
Hi, sorry to ping you but can you launch the CI please ? @spolti

spolti:
@Jooho do you mind taking a look?

moulin-louis:
i've pushed a fix for `make precommit` as well

Erratum: it was my system python version that was too old (3.9), i've reverted my commit

python/storage/test/test_mlflow_storage.py
```diff
@@ -0,0 +1,161 @@
...
+from kserve_storage import Storage
+
+
+class TestDownloadMlflow:
+    """Tests for Storage._download_mlflow() method."""
+
+    @mock.patch.dict(os.environ, {}, clear=True)
+    def test_missing_tracking_uri_raises_error(self, tmp_path):
+        """Should raise ValueError when MLFLOW_TRACKING_URI is not set."""
+        uri = "mlflow://models:/my-model/1"
```
lizzzcai: @moulin-louis thanks for your PR. for the storage initalizer format here, how about `mlflow://models/my-model/1`? (remove `:` after models/runs.)

lizzzcai:
@moulin-louis Thanks for the contribution — this is exactly the feature we need, and the overall design looks solid.

  One thing worth discussing: the storage URI format. The current approach passes the MLflow artifact URI verbatim after mlflow://, giving:

Current (PR):
  mlflow://models:/wine-quality/2
  mlflow://models:/wine-quality@champion
  mlflow://runs:/abc123/iris_model

  An alternative would be to drop the inner colon and treat models/runs as the path root:

 Alternative:
  mlflow://models/wine-quality/2
  mlflow://models/wine-quality@champion
  mlflow://runs/abc123/iris_model

The alternative reads more like a standard URL. The trade-off is a small reconstruction step internally (models + :/ + rest). No strong preference on our end — happy to go with whichever you think is clearer to users.

Thanks again for driving this!

moulin-louis:
Hi @lizzzcai !

Thanks for the feedback !
The goal was to mimic the models URI format from MLFlow so that users can just append ```mlflow://``` and it works.
 
Happy to ear about other thoughts on this.

I'll rebases the PR tonigh, hoping we can merge it before ```0.19.0```.

moulin-louis:
Should be good now. Can you take a look @spolti or @Jooho maybe ? Thanks !
Btw: the E2E tests are failing due to a network error? I can't retry it easily.

moulin-louis:
Fixed the conflicts from master.

lizzzcai:
> Hi @lizzzcai !
> 
> Thanks for the feedback ! The goal was to mimic the models URI format from MLFlow so that users can just append `mlflow://` and it works.
> 
> Happy to ear about other thoughts on this.
> 
> I'll rebases the PR tonigh, hoping we can merge it before `0.19.0`.

I am not sure is it ok to have both `://` and `:/` in the storage uri. maybe @kserve/kserve-leads @kserve/kserve-contributors can provide feedback here as this is a new storage support.

terrytangyuan:
Having both seems inconsistent

moulin-louis:
Yes, i agree. I think the best path is to follow the other format by having "mlflow://" followed by the standard Mlflow Uri. (https://mlflow.org/docs/latest/ml/model-registry/#concepts)

This is my current implementation in this PR. @lizzzcai @terrytangyuan

adrianAzoitei:
Dropping in 👀. What's the status on this issue? Is feedback from maintainers needed / no changes to the PR need to happen?

Happy to help the PR progress by contributing!

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -1261,3 +1282,34 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
...
+        model_uri = uri[len(_MLFLOW_PREFIX) :]
+        logger.info(f"Downloading {model_uri} from {mlflow_tracking_uri}")
+        if not model_uri:
+            raise ValueError("Model uri cannot be empty")
+        try:
+            mlflow.artifacts.download_artifacts(
+                artifact_uri=model_uri, dst_path=out_dir
+            )
+        except MlflowException:
+            raise RuntimeError("Failed to download model from MLFlow")
```
lizzzcai: this discarding the original error message, maybe can be `raise RuntimeError(f"Failed to download model from MLFlow: {e}")`

lizzzcai:
> Having both seems inconsistent

@terrytangyuan thanks for chiming in! Just to make sure we address the right thing —
when you say "having both seems inconsistent", do you mean the `://` and `:/` appearing
within the same URI string in the current implementation (e.g. `mlflow://models:/wine-quality/2`),
or something else? Wanting to make sure we are aligned before deciding on the format.

lizzzcai:
> Yes, i agree. I think the best path is to follow the other format by having "mlflow://" followed by the standard Mlflow Uri. (https://mlflow.org/docs/latest/ml/model-registry/#concepts)
> 
> This is my current implementation in this PR. @lizzzcai @terrytangyuan


@moulin-louis thanks again for driving this.

From my side, I lean towards the alternative format I suggested earlier — dropping the
inner `:` so the URI is consistent with all other KServe storage prefixes (`s3://`, `gs://`, `hf://`):

| Current (this PR) | Proposed |
|---|---|
| `mlflow://models:/wine-quality/2` | `mlflow://models/wine-quality/2` |
| `mlflow://models:/wine-quality/Production` | `mlflow://models/wine-quality/Production` |
| `mlflow://models:/wine-quality@champion` | `mlflow://models/wine-quality@champion` |
| `mlflow://runs:/abc123def456/iris_model` | `mlflow://runs/abc123def456/iris_model` |

The reconstruction internally is a one-liner (`models/` → `models:/`), and it covers all
the common patterns:

```
mlflow://models/wine-quality/2                        # by version
mlflow://models/wine-quality/Production               # by stage
mlflow://models/wine-quality@champion                 # by alias
mlflow://models/wine-quality/2/data/model.pkl         # sub-artifact
mlflow://runs/abc123def456/iris_model                 # run artifact
mlflow://mlflow-artifacts/0/abc123/artifacts/model    # artifact proxy
```

Happy to hear other thoughts on this.

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -1261,3 +1282,34 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
...
+            raise ValueError("Cannot find MLFlow tracking Uri")
+        mlflow.set_tracking_uri(mlflow_tracking_uri)
+
+        if (
+            os.getenv("MLFLOW_TRACKING_USERNAME") is not None
+            and os.getenv("MLFLOW_TRACKING_PASSWORD") is not None
+            and os.getenv("MLFLOW_TRACKING_TOKEN") is not None
+        ):
+            raise ValueError(
+                "Tracking Token cannot be set with Username/Password combo"
```
lizzzcai: The credential conflict check here is stricter than MLflow itself. When both token and username/password are set, MLflow does not error — it silently picks username/password (basic auth is checked first in [`mlflow/utils/rest_utils.py#L255-L259`](https://github.com/mlflow/mlflow/blob/master/mlflow/utils/rest_utils.py#L255-L259))

Also, the current implementation seems like a token combined with username alone, or a token combined with password alone is silently passes through.

terrytangyuan:
Yes ://

python/storage/kserve_storage/kserve_storage.py
moulin-louis: Yeah your right, Its fixed now.

python/storage/kserve_storage/kserve_storage.py
moulin-louis: Lets MLFlow handle auth in their side then.

moulin-louis: Fix pushed

moulin-louis:
I've pushed your implementation @lizzzcai. I still think that following MLFlow implementation would be clearer for the end users but if consistency is more important, I'm down for it.

lizzzcai:
@moulin-louis two checks seems failed and two latest comments seems show unverified, can you check, thanks.

moulin-louis:
Hello, yes I can take a look this weekend 👍

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -1351,3 +1372,29 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
...
+            raise ValueError("Cannot find MLFlow tracking Uri")
+        mlflow.set_tracking_uri(mlflow_tracking_uri)
+
+        model_uri = uri[len(_MLFLOW_PREFIX) :]  # Stripping mlflow prefix
+        if not model_uri:
+            raise ValueError("Model uri cannot be empty")
+        prefix, rest = model_uri.split("/", 1)  # Spliting on the first `/`
+        model_uri = f"{prefix}:/{rest}"  # reconstruing mlflow model_uri from `model/` to `model:/`
+        logger.info(f"Downloading {model_uri} from {mlflow_tracking_uri}")
+        logger.error(f"model_uri = {model_uri}")
```
lizzzcai, line 1392: this `logger.error` should be removed I assumed. @moulin-louis

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -1351,3 +1372,29 @@ def _unpack_archive_file(file_path, mimetype, target_dir=None) -> str:
...
+
+        mlflow_tracking_uri = os.getenv("MLFLOW_TRACKING_URI")
+        if mlflow_tracking_uri is None:
+            raise ValueError("Cannot find MLFlow tracking Uri")
+        mlflow.set_tracking_uri(mlflow_tracking_uri)
+
+        model_uri = uri[len(_MLFLOW_PREFIX) :]  # Stripping mlflow prefix
+        if not model_uri:
+            raise ValueError("Model uri cannot be empty")
+        prefix, rest = model_uri.split("/", 1)  # Spliting on the first `/`
```
lizzzcai, line 1389: malformed input will crash, maybe
```python
  parts = model_uri.split("/", 1)
  if len(parts) < 2:
      raise ValueError(f"Invalid mlflow URI format: {uri}")
  prefix, rest = parts
```

python/storage/README.md
```diff
@@ -99,6 +100,15 @@ model_dir = Storage.download("hf://org-name/model-name")
 model_dir = Storage.download("hf://org-name/model-name:revision")
 ```
 
+### MLFlow
+
+```python
+# Using model version alias
+model_dir = Storage.download("mlflow://models:/model@dev")
```
lizzzcai, line 107: both examples should be updated for the new format.

pkg/client/clientset/versioned/typed/serving/v1alpha1/fake/fake_inferencegraph.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The KServe Authors.
+Copyright 2026 The KServe Authors.
```
lizzzcai, line 2: this shouldn't be updated for existing files, only the new added one. Can you check again on the modified files.

lizzzcai:
@moulin-louis I manually tested your PR, and it is working as expected for the mlflow storage. Few comments added, can you help to fix it.

PR #4932, opened 2026-01-06
Fix: opentelemetry helm installation script

Jooho:
Fix: opentelemetry helm installation script

- add it to quick install script
- remove set value in setup-deps.sh for ci (duplicated part)
- update the helm install script to use the specific Helm version used locally (where the issue does not occur) and aligned CI to use the same version.

Fixes #

**Type of changes**

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:

```release-note

```

**Re-running failed tests**

### DIFF
--- a/hack/setup/cli/install-helm.sh
+++ b/hack/setup/cli/install-helm.sh
@@ -27,13 +27,14 @@ install() {
 
     log_info "Installing Helm ${HELM_VERSION} for ${os}/${arch}..."
 
-    if command -v helm &>/dev/null; then
-        local current_version=$(helm version --template='{{.Version}}' 2>/dev/null)
-        if [[ -n "$current_version" ]] && version_gte "$current_version" "$HELM_VERSION"; then
-            log_info "Helm ${current_version} is already installed (>= ${HELM_VERSION})"
+    # Check if helm is already installed in BIN_DIR with the exact required version
+    if [[ -f "${BIN_DIR}/helm" ]]; then
+        local current_version=$("${BIN_DIR}/helm" version --template='{{.Version}}' 2>/dev/null)
+        if [[ "$current_version" == "$HELM_VERSION" ]]; then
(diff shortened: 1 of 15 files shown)


hack/setup/scripts/generate-versions-from-gomod.py
```diff
@@ -0,0 +1,197 @@
...
+HELM_REPOS = {
+    "istio": "https://istio-release.storage.googleapis.com/charts",
+    "kedacore": "https://kedacore.github.io/charts",
+    "open-telemetry": "https://open-telemetry.github.io/opentelemetry-helm-charts",
+}
+
+
+def run(cmd):
+    """Run command and return stdout"""
+    return subprocess.run(cmd, shell=True, capture_output=True, text=True, check=True).stdout
```
jlost, line 35: If a command fails, it will fail silently, possibly with no stdout (returning an empty string).
Adding `check=True` could help ensure errors are surfaced to the user if and when they are encountered.
In that case, you'd also want to remove the `try / except: pass` patterns e.g. in `ensure_helm_repo`.

jlost (commented):
Looks good -- love to see it in Python rather than a shell script!
I left a suggestion to help avoid swallowing exception information.
I lack write access to approve, but /lgtm!

hack/setup/scripts/generate-versions-from-gomod.py
Jooho: Yes true. `check=True` is already there so I just removed the `try/except` part.

Jooho:
I encountered a weird CI issue, so I decided to split this PR into two to release other PRs. This PR will resolve the OpenTelemetry helm script issue and https://github.com/kserve/kserve/pull/4935, addressing the original issue.

Jooho:
The root cause was Helm v4, which triggers the JSON schema conflict. We updated the helm install script to use the specific Helm version used locally (where the issue does not occur) and aligned CI to use the same version.

PR #4948, opened 2026-01-12
Fix: use correct image tag in LLMISvc E2E workflow

pierDipi:
The workflow was patching the deployment to use an image tag that wasn't loaded into minikube. The image is loaded as 'local-test' but the patch was using the github.sha tag.

Also change from `kubectl wait --for=condition=available` to `kubectl rollout status` to ensure the new pod is ready, not just the old one during rolling update.

**Re-running failed tests**

### DIFF
--- a/.github/workflows/e2e-test-llmisvc.yaml
+++ b/.github/workflows/e2e-test-llmisvc.yaml
@@ -107,17 +107,17 @@ jobs:
 
           # Set environment variable to use local charts
           export USE_LOCAL_CHARTS=true
-          LLMISVC_CONTROLLER_IMG_TAG="${DOCKER_REPO}/${LLMISVC_CONTROLLER_IMG}:${{ github.sha }}"
 
           # Use the quick install script to install all dependencies and LLMISvc
           ./hack/llmisvc_quick_install.sh
 
           # Update the deployment to use our built image with Never pull policy
+          # Use 'local-test' tag since that's what was loaded into minikube
           kubectl patch deployment kserve-llmisvc-controller-manager -n kserve \
-            -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"'${LLMISVC_CONTROLLER_IMG_TAG}'","imagePullPolicy":"Never"}]}}}}'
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -107,17 +107,17 @@ jobs:
...
-          LLMISVC_CONTROLLER_IMG_TAG="${DOCKER_REPO}/${LLMISVC_CONTROLLER_IMG}:${{ github.sha }}"
 
           # Use the quick install script to install all dependencies and LLMISvc
           ./hack/llmisvc_quick_install.sh
 
           # Update the deployment to use our built image with Never pull policy
+          # Use 'local-test' tag since that's what was loaded into minikube
           kubectl patch deployment kserve-llmisvc-controller-manager -n kserve \
-            -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"'${LLMISVC_CONTROLLER_IMG_TAG}'","imagePullPolicy":"Never"}]}}}}'
+            -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"kserve/llmisvc-controller:local-test","imagePullPolicy":"Never"}]}}}}'
```
spolti, line 117: is it fine to hard-code it?
in this case, we don't need this env `LLMISVC_CONTROLLER_IMG_TAG` anymore, right?

pierDipi: yes, it's hardcoded in a previous step

```
          echo "🐳 Tagging and loading LLMISvc controller image into minikube..."
          # Load the image from the tar file
          docker load -i ./tmp/llmisvc-controller-${{ github.sha }}
          # Tag the loaded image as kserve/llmisvc-controller:local-test
          IMAGE_ID=$(docker images --format "{{.Repository}}:{{.Tag}} {{.ID}}" | grep llmisvc-controller | awk '{print $2}' | head -n1)
          docker tag $IMAGE_ID kserve/llmisvc-controller:local-test
          # Load the tagged image into minikube
          minikube image load kserve/llmisvc-controller:local-test
```

spolti (approved):
It will be updated to use the env in another PR.
Approving for now.

/lgtm

Jooho (approved):
/lgtm
/approve

PR #4951, opened 2026-01-12
fix: watch pod init container status changes for faster error propagation (RequeueAfter)

jlost:

Init container errors (e.g., invalid S3 credentials) take 10 minutes to propagate to the InferenceService status. This is because the controller watches the Deployment, but Deployment status doesn't update until `progressDeadlineSeconds` expires (default 600s).

This PR fixes the issue by polling pod status while the init container is running. When `PropagateModelStatus` detects the storage-initializer init container is in the `Running` or `Waiting` state, it returns `false` to trigger a `RequeueAfter: 10 * time.Second`, ensuring failures are detected within 10 seconds.

**Related PR**: This is an alternative to #4910 that solves the same problem using `RequeueAfter` polling instead of a pod watch.

**Changes:**
- Continue polling while init container is running (not just waiting) to detect failures quickly
- Fix `ISVC.status` never becoming populated if the InitContainer of a Predictor failed while in Gateway API mode
- Fix ClusterServingRuntime changes triggering reconciliation of all InferenceServices in the cluster, rather than only the ones using that runtime
- Refactor `servingRuntimesPredicate` and `clusterServingRuntimesPredicate` from inline variables to standalone functions for consistency

**Tradeoffs: Polling (this PR) vs Pod Watch (#4910)**

| Aspect | Polling (RequeueAfter) | Pod Watch |
|--------|------------------------|-----------|
| **Code impact** | ~10 lines changed  | ~100 lines added  |
| **Update latency** | 0-10 seconds | Near-instant |
| **Resource overhead** | Periodic reconciliations during init | Pod cache + event handlers |
| **Idle cost** | None once loaded | Minimal (events filtered tightly) |

The polling approach is a minimal change to existing logic, trading slightly slower error detection (~10s max vs near-instant) for a smaller code footprint.

**Type of changes**

- [x] Unit tests for `PropagateModelStatus` verifying requeue behavior for init container states
- [x] Unit tests for `clusterServingRuntimeFunc` mapper verifying it only reconciles ISVCs using the specific runtime
- [x] Unit tests for predicate functions

**Checklist**:

```release-note
Fix slow propagation of init container errors (e.g., storage-initializer failures) to InferenceService status by polling during model loading
Fix `ISVC.status` never becoming populated if the InitContainer of a Predictor failed while in Gateway API mode
Fix ClusterServingRuntime changes triggering reconciliation of all InferenceServices in the cluster, rather than only for that runtime
```


jlost:
Closing - we settled on the approach in #4910

PR #4955, opened 2026-01-13
feat: add enqueueOnHttpRouteChange handler

vivekk16:

1. Adds `enqueueOnHttpRouteChange` event handler to trigger `LLMInferenceService` reconciliation when referenced HTTPRoutes change (including those referenced via `baseRefs`)
2. Registers the HTTPRoute watch in `SetupWithManager`
3. Improves clarity of comments on all three event handlers (`enqueueOnGatewayChange`, `enqueueOnHttpRouteChange`, `enqueueOnLLMInferenceServiceConfigChange`)

**Type of changes**

1. Create an `LLMInferenceServiceConfig` with an HTTPRoute reference
2. Create an `LLMInferenceService` that references the config via `baseRefs`
3. Create/update the HTTPRoute
4. Verify that the `LLMInferenceService` reconciles when the HTTPRoute changes

**Checklist**:

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/controller.go
+++ b/pkg/controller/v1alpha2/llmisvc/controller.go
@@ -255,7 +255,8 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("failed to add GIE APIs to scheme: %w", err)
 	}
 	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), gwapiv1.GroupVersion.String(), "HTTPRoute"); ok && err == nil {
-		b = b.Owns(&gwapiv1.HTTPRoute{}, builder.WithPredicates(childResourcesPredicate))
+		b = b.Owns(&gwapiv1.HTTPRoute{}, builder.WithPredicates(childResourcesPredicate)).
+			Watches(&gwapiv1.HTTPRoute{}, r.enqueueOnHttpRouteChange(logger))
 	}
 	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), gwapiv1.GroupVersion.String(), "Gateway"); ok && err == nil {
 		b = b.Watches(&gwapiv1.Gateway{}, r.enqueueOnGatewayChange(logger))
@@ -282,7 +283,7 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // enqueueOnGatewayChange creates an event handler that triggers reconciliation of LLMInferenceServices
-// when a Gateway they depend on changes. This is necessary because Gateway changes can affect routing.
+// when a referenced Gateway changes. This ensures routing is updated when Gateway status changes.
 func (r *LLMISVCReconciler) enqueueOnGatewayChange(logger logr.Logger) handler.EventHandler {
 	logger = logger.WithName("enqueueOnGatewayChange")
 	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, object client.Object) []reconcile.Request {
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm
/approve

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -350,8 +351,70 @@ func (r *LLMISVCReconciler) enqueueOnGatewayChange(logger logr.Logger) handler.E
 	})
 }
 
+// enqueueOnHttpRouteChange creates an event handler that triggers reconciliation of LLMInferenceServices
+// when a referenced HTTPRoute changes. This ensures routing is updated when HTTPRoute status changes.
+func (r *LLMISVCReconciler) enqueueOnHttpRouteChange(logger logr.Logger) handler.EventHandler {
+	logger = logger.WithName("enqueueOnHttpRouteChange")
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, object client.Object) []reconcile.Request {
+		sub := object.(*gwapiv1.HTTPRoute)
```
Jooho, line 360: The variable name feels a bit ambiguous. I had to trace through the code to understand what sub actually represents, and given that this is related to an HTTPRoute, it was somewhat unexpected. It might improve readability to use a more descriptive name that reflects its role more explicitly.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -254,7 +254,8 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("failed to add GIE APIs to scheme: %w", err)
 	}
 	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), gwapiv1.GroupVersion.String(), "HTTPRoute"); ok && err == nil {
-		b = b.Owns(&gwapiv1.HTTPRoute{}, builder.WithPredicates(childResourcesPredicate))
+		b = b.Owns(&gwapiv1.HTTPRoute{}, builder.WithPredicates(childResourcesPredicate)).
```
Jooho, line 258: Using both Owns() and Watches() for HTTPRoute itself seems fine. My only concern is that, in scenarios where an HTTPRoute is both owned by the LLMInferenceService and also referenced (e.g. via baseRefs), the same update could trigger reconciliation twice.

I didn’t see an explicit guard or validation to prevent or account for this, so I wanted to flag it as something worth double-checking. This may be acceptable in practice, but the intended behavior wasn’t immediately obvious to me from the code.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -350,8 +351,70 @@ func (r *LLMISVCReconciler) enqueueOnGatewayChange(logger logr.Logger) handler.E
...
+					continue
+				}
+
+				// Skip services that don't use HTTPRoute refs
+				if combinedCfg.Spec.Router == nil || combinedCfg.Spec.Router.Route == nil || !combinedCfg.Spec.Router.Route.HTTP.HasRefs() {
+					continue
+				}
+
+				// Check if service explicitly references this HTTPRoute
+				for _, ref := range combinedCfg.Spec.Router.Route.HTTP.Refs {
```
Jooho, line 396: A minor concern is that this could also enqueue a request for the LLMInferenceService that already owns the HTTPRoute. Since Owns() automatically triggers reconciliation for owned resources, this could not only result in duplicate reconciles, but may also cause setController errors when the controller tries to take ownership of an already-owned object.

Jooho (commented):
I’ve left some comments, and I strongly recommend adding test cases for these changes.

pkg/controller/v1alpha2/llmisvc/controller.go
pierDipi: in every handler sub is used as the subject of the change, let's keep it for consistency or follow up to change it everywhere

pkg/controller/v1alpha2/llmisvc/controller.go
pierDipi: if it's owned (managed), doesn't really make sense to reference it (in practice)

pkg/controller/v1alpha2/llmisvc/controller.go
pierDipi: if it's owned (managed), doesn't really make sense to reference it (in practice), double reconciliation is fine if it ever happens

spolti (approved):
/lgtm

pkg/controller/v1alpha2/llmisvc/controller.go
Jooho: I think Kuberentes operator is smarter than I thought
```
The controller-runtime uses a WorkQueue that automatically deduplicates multiple requests for the same resource (namespace/name). If multiple events (from both Own and Watch) arrive simultaneously, they are collapsed into a single Reconcile call.
```

pkg/controller/v1alpha2/llmisvc/controller.go
Jooho: OK

Jooho (approved):
/approved
/lgtm

PR #4964, opened 2026-01-15
fix: add the (default) zmq port 5557 to scheduler

pierDipi:
add the (default) zmq port 5557 to scheduler for precise prefix scorer events

vLLM uses the epp service (created by Kserve) to connect to the scheduler

**Release note**:

```release-note
Add the (default) zmq port 5557  to scheduler for precise prefix scorer
```

### DIFF
--- a/charts/kserve-llmisvc-resources/templates/config-llm-scheduler.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-scheduler.yaml
@@ -35,6 +35,9 @@ spec:
               - containerPort: 9090
                 name: metrics
                 protocol: TCP
+              - containerPort: 5557
+                name: zmq
+                protocol: TCP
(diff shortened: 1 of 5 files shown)


Jooho:
@pierDipi 

From what I know, ZMQ was already being used internally back then, and exposing port 5557 wasn’t strictly necessary.

Could you clarify what changed that made this PR needed? Specifically:

- Why do we now need to expose port 5557 in the default scheduler ConfigMap / Service?

- Was there a change in how cache_tracking interacts with the scheduler that requires the port to be reachable externally, or something similar?

By the way, source change looks good to me

pierDipi:
This is necessary because vLLM uses the epp service (created by Kserve) to connect to the scheduler

spolti (approved):
/lgtm

terrytangyuan (approved):

PR #4966, opened 2026-01-15
feat: added git download support to storage initializer

agoston:
Added git support to storage initializer.

To use it, simply pass a download URL of any of the below format:
- https://github.com/kserve/kserve.git

HTTPS transport is supported.

**Type of changes**

- [x] Unit tests added
- [x] Manual tests with the built docker container

- Logs

None.

**Checklist**:

**Release note**:
```release-note
Added git support to storage initializer.

To use it, simply pass a download URL of any of the below format:
- https://github.com/kserve/kserve.git

```


python/storage-initializer.Dockerfile
```diff
@@ -54,6 +54,10 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
 RUN useradd kserve -m -u 1000 -d /home/kserve
 
+# Install system dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends git openssh-client && apt-get clean && \
```
spolti: can't we rely only on https?
We don't need to use git features in this case.

agoston: Good one. In general, ssh has quite some advantages over https:
- private key generally offers much higher security;
- the possibility to include ssh host keys can effectively disable man-in-the-middle attacks;
- in general, ssh transport layer offers higher security.

You're right that it adds complexity and I've been contemplating this myself. At Cloudera, we use this to download models from github via ssh, hence I am leaning towards this feature.

Jooho:
From my perspective, while Git access can be scoped to read-only (for example, via deploy keys), using Git over SSH still relies on long-lived private keys, which are generally harder to rotate, revoke, and audit. In a Kubernetes environment—where secrets are mounted into ephemeral pods—this can increase operational complexity and the potential blast radius if a key is ever exposed.

Additionally, requiring the runtime image to include Git tooling expands the container’s attack surface and works against the goal of keeping runtime images minimal. For a strictly read-only model download use case, HTTPS- or identity-based approaches often provide equivalent access control with shorter-lived credentials, simpler lifecycle management, and a smaller runtime footprint.

Overall, this is just my perspective rather than a hard objection to Git itself, with the main goal being to minimize long-lived secrets and unnecessary tooling when viable alternatives exist.

agoston:
> From my perspective, while Git access can be scoped to read-only (for example, via deploy keys), using Git over SSH still relies on long-lived private keys, which are generally harder to rotate, revoke, and audit. In a Kubernetes environment—where secrets are mounted into ephemeral pods—this can increase operational complexity and the potential blast radius if a key is ever exposed.
> 
> Additionally, requiring the runtime image to include Git tooling expands the container’s attack surface and works against the goal of keeping runtime images minimal. For a strictly read-only model download use case, HTTPS- or identity-based approaches often provide equivalent access control with shorter-lived credentials, simpler lifecycle management, and a smaller runtime footprint.
> 
> Overall, this is just my perspective rather than a hard objection to Git itself, with the main goal being to minimize long-lived secrets and unnecessary tooling when viable alternatives exist.

Thanks @Jooho ! I was contemplating the same. There is a [pure python library](https://pypi.org/project/dulwich/) that would handle git downloads perfectly. That would be much more along the lines you expect. Would you agree merging that solution?

Jooho:
That sounds good to me. Using a pure Python library like Dulwich aligns well with what I had in mind, especially since it allows us to avoid pulling Git and SSH tooling into the runtime image, which is a net win from an image-hardening and security perspective.

From my point of view, the key benefit here would be pairing Dulwich with HTTPS-based access and scoped, read-only tokens, rather than SSH keys. Avoiding long-lived SSH private keys altogether would significantly simplify secret management in Kubernetes and reduce the overall blast radius. Even if a token ends up being long-lived initially, keeping it tightly scoped and easy to rotate would already be a meaningful improvement.

If the implementation follows that model (no SSH keys, minimal runtime dependencies, read-only access), I’d be supportive of moving forward and happy to review the details in the PR.

agoston:
@Jooho Updated as requested. Please re-review.

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -55,6 +55,7 @@
 _HEADERS_SUFFIX = "-headers"
 _PVC_PREFIX = "/mnt/pvc"
 _HF_PREFIX = "hf://"
+_GIT_RE = r"https://.+\.git"
```
lizzzcai, line 58: for other storage types, we have `hf://`, `s3://`, `gcs://`, `hdfs://`. can we have smiliar patter nfor git like `git://`

agoston: I was thinking the same. But it turns out, `git://` [refers to something else](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#The-Git-Protocol).

It would also be misleading, as it's not quite clear that e.g. we don't support ssh or git protocol itself, only git-over-https.

As @Jooho put it above, let's not overcomplicate this -- is a k8s native environment, 98% of use cases would/should use https anyway.

lizzzcai: @agoston , there is an exsting [prefix for http(s)](https://github.com/agoston/kserve/blob/dc0546e94c52a6909a759120362f1c76dced2e31/python/storage/kserve_storage/kserve_storage.py#L53-L54) as well.

Does the current git support also handle fetching from a non-default branch and a subdirectory within the repo? For example:
Repo: https://github.com/kserve/kserve
Branch: release-0.16
Subpath: charts/kserve-crd-minimal
URL example: https://github.com/kserve/kserve/tree/release-0.16/charts/kserve-crd-minimal

These are common scenarios. Not to overcomplicate it but just raise it up in case some scenario is missing, in case we need to support it later.

agoston: True that there is a prefix for https -- I had to rearrange the order of checks to leave that last, as a sort of 'catch-all'. There are great many protocols implemented over https after all, git is really just one of them.

As for branch support, we really just pass the url on to `dulwich`. Dulwich does support branches, however, the standard git URLs *do not*. The example you gave above is github-specific, e.g. `pip` supports the `git+https://github.com/user/repo.git@branch-name` format, whereas `go` supports the `import "example.com/repo.git/branch-name"` format. Gitlab uses the `https://gitlab.com/user/repo/-/tree/branch-name` format.

Having `.git` at the end of the https:// URL however is standard, hence the regex pattern above.

At this stage, there is no git support storage-initializer at all. I really want to focus on adding a basic support that covers the needs of 90%+ of users. Note that up to this point only the absolute standard git features are supported, so we are not tying ourselves to some vendor-specific extension (nor do I think that's a good idea anyway). So it is wide open for extension later on, if there's consensus.

Jooho (approved):
/lgtm

PR #4970, opened 2026-01-17
feat: make LLMInferenceService storageInitializer optional

### LINKED ISSUE #4899: Storage Initializer should be optional when deploying LLMInferenceService
/kind feature

storage-initializer should be optional, with a simple flag to disable it from being created as an init-container, defaulting to true.

With modules like [runai-streamer](https://docs.vllm.ai/en/stable/models/extensions/runai_model_streamer/?h=), and when running in a cluster that doesn't have a robust PVC storage solution with `ReadWriteMany`, components like storage-initializer only slow down the start time of vllm containers by forcing the download of models to an EmptyDir before vllm can start.

cabrinha:
**What this PR does / why we need it**: Makes Storage Initializer optional in LLMInferenceService. With [RunAI Streamer](docs.vllm.ai/en/stable/models/extensions/runai_model_streamer), models are streamed directly to the GPU's VRAM and don't need to be written to disk beforehand, making the storage-initializer initContainer an unnecessary step during startup.

**Which issue(s) this PR fixes** : https://github.com/kserve/kserve/issues/4899
Fixes #4899

- [x] Test A: Updated `controller_workload_storage_int_test.go` to include a test for checking that storage-initializer initContainer is not present when `enabled=false`

- Logs

**Checklist**:

**Release note**:
```release-note
Introduce a new optional field to LLMInferenceService spec `storageInitializer.enabled` to allow users to disable the storage-initializer initContainer. This defaults to `true`, to leave storage-initializer enabled by default.
```


pkg/controller/v1alpha2/llmisvc/workload_storage.go
pierDipi, line 1: qq: I see the checks only for s3 and hf but the API `StorageInitializer.Enabled` is generic and doesn't convey this behavior, should it be skipped in every case (and assume that it won't be disabled when it doesn't need to be disabled for OCI, etc) ?

cabrinha: S3 and HF were the only two I thought of, I also see `pvc://` referenced but I didn't consider `oci://` either. Are those four protocols all that are supported? If using `pvc://` or `oci://` does storage-initializer need to run at all?

My thinking is that if `StorageInitializer.Enabled=false`, then it should be skipped in every case. By setting to `false`, the user is conveying that model download and storage will be taken care of elsewhere and the init container is not needed.

Let me know what else I should change to better convey this behavior.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -153,6 +153,14 @@ func (r *LLMISVCReconciler) attachPVCModelArtifact(modelUri string, podSpec *cor
 //
 //	An error if the configuration fails, otherwise nil.
 func (r *LLMISVCReconciler) attachS3ModelArtifact(ctx context.Context, serviceAccount *corev1.ServiceAccount, llmSvc *v1alpha2.LLMInferenceService, modelUri string, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig, credentialConfig *credentials.CredentialConfig) error {
+	// Check if storage initializer is disabled
+	if llmSvc.Spec.StorageInitializer != nil &&
```
spolti: Maybe move it to a function, as this same if seems to be used in more places.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
spolti: wondering if this logic shouldn't be moved to the storageInitializer injector logic, there are some scenarios for any isvc or llmisvc that the SI could be avoided, like relying on the HF runtime, which fetches it.

I don't remember exactly what is the use case from your side, and where and how you are injecting the model. Could you please describe the details? 
thx

cabrinha: The issue I opened references RunAI Streamer: https://github.com/kserve/kserve/issues/4899

This streams the model from the remote storage directly to the GPUs VRAM, without writing it to disk first. If this logic needs to be moved, I'm open to that, just let me know.

pierDipi: I'd skip it in every case at the attachModelArtifacts level, the assumption is that when disabled model download/fetching/etc is managed fully externally

cabrinha: ok, going to give that a try

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -61,6 +61,11 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
 		return fmt.Errorf("invalid model URI: %s", modelUri)
 	}
 
+	// Check if storage-initializer is explicitly disabled
+	storageInitializerDisabled := llmSvc.Spec.StorageInitializer != nil &&
+		llmSvc.Spec.StorageInitializer.Enabled != nil &&
+		!*llmSvc.Spec.StorageInitializer.Enabled
```
pierDipi, line 67: I'd add the check right below this line and remove the per storage check

cabrinha: I hope I got it right here :D https://github.com/kserve/kserve/pull/4970/commits/9975f81842aaf9326cd88fba0ce94d2cb6ba8489

pierDipi:
@cabrinha pre-commit check doesn't like the changes in `openapi_generated.go` https://github.com/kserve/kserve/actions/runs/21407966979/job/61871365113?pr=4970

The other test seems like a flake

cabrinha:
> @cabrinha pre-commit check doesn't like the changes in `openapi_generated.go` https://github.com/kserve/kserve/actions/runs/21407966979/job/61871365113?pr=4970
> 
> The other test seems like a flake

Super odd because `make precommit` works on my end:
```
======================================================================
Generation Summary
======================================================================
Total files:  9
✅ Success:   9
❌ Failed:    0
```

I see some whitespace was removed from that file, maybe it's good to go again?

Luew2:
Just adding -- hope this gets in soon, trying to add dual support for gguf since vllm has it now but due to the storage initializer i need to bypass it and pass --model directly to vllm, wish storage initializer was optional or supported gguf directly from hugging face

pierDipi (approved):
/lgtm

(I'm not an approver, just giving my LGTM)

terrytangyuan (approved):
Thanks!

PR #4983, opened 2026-01-21
feat: add default node affinity injector for model pods

### LINKED ISSUE #4982: Add default node affinity injection for model server pods via environment variable
/kind feature

## Describe the solution you'd like

Allow cluster administrators to configure a **default node affinity** for all KServe model server pods through an environment variable (`KSERVE_DEFAULT_NODEPOOL`). Pods without explicit affinity will automatically target the specified node pool.

## Related to #730

**Why this is needed:** #730 requires explicit config on every InferenceService. This provides a baseline default—user-specified affinity from #730 still takes precedence.

## Proposed Implementation

Add `InjectAffinity` mutator to the pod webhook:
- Reads `KSERVE_DEFAULT_NODEPOOL` env var
- Injects `RequiredDuringSchedulingIgnoredDuringExecution` affinity using `cloud.google.com/gke-accelerator` label
- Skips if env var unset or pod already has affinity

## Example

```yaml
env:
  - name: KSERVE_DEFAULT_NODEPOOL
    value: "nvidia-tesla-t4"

### PR DESCRIPTION
**Pull request by neo99zhang**
Adds a webhook mutator that injects default node affinity to KServe pods without explicit affinity configured. This allows cluster admins to route inference workloads to specific node pools by default.

Configuration via environment variables:
- `KSERVE_DEFAULT_NODEPOOL`: target node pool value (required to enable)
- `KSERVE_DEFAULT_NODEPOOL_LABEL_KEY`: label key to match (default: `cloud.google.com/gke-nodepool`)

Skips injection if env var is not set or pod already has affinity configured.

Fixes #4982

- [x] Unit tests for affinity injection when pod has no affinity
- [x] Unit tests for skipping injection when pod already has affinity
- [x] Unit tests for skipping when env var not set
- [x] Unit tests for custom label key configuration

This complements #730 (per-InferenceService affinity) by providing a cluster-wide default. User-specified affinity always takes precedence.

**Checklist**:

**Release note**:
```release-note
Add support for default node affinity injection via KSERVE_DEFAULT_NODEPOOL environment variable

### DIFF
--- a/pkg/webhook/admission/pod/affinity_injector.go
+++ b/pkg/webhook/admission/pod/affinity_injector.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2021 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package pod
+
+import (
+	"os"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	DefaultNodepoolEnvVar   = "KSERVE_DEFAULT_NODEPOOL"
+	DefaultNodepoolLabelKey = "KSERVE_DEFAULT_NODEPOOL_LABEL_KEY"
+	defaultNodepoolLabelKey = "cloud.google.com/gke-nodepool"
+)
+
+// InjectAffinity injects a default affinity config (configurable through environment variables) to each kserve pod
+// if it is created without affinity. This allows cluster administrators to route inference workloads to specific
+// node pools by default.
+func InjectAffinity(pod *v1.Pod) error {
+	// If the default nodepool is not set in environment variable, skip injection.
+	pool := os.Getenv(DefaultNodepoolEnvVar)
+	if len(pool) == 0 {
+		return nil
+	}
+
+	// If the pod already has affinity configured, skip injection.
+	if pod.Spec.Affinity != nil {
+		return nil
+	}
+
+	// Get the label key from environment variable, or use default.
+	labelKey := os.Getenv(DefaultNodepoolLabelKey)
+	if len(labelKey) == 0 {
+		labelKey = defaultNodepoolLabelKey
+	}
+
+	// Add default affinity config to the pod.
+	pod.Spec.Affinity = defaultAffinity(labelKey, pool)
+	return nil
+}
+
+func defaultAffinity(labelKey, pool string) *v1.Affinity {
+	return &v1.Affinity{
+		NodeAffinity: &v1.NodeAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+				NodeSelectorTerms: []v1.NodeSelectorTerm{
+					{
+						MatchExpressions: []v1.NodeSelectorRequirement{
+							{
+								Key:      labelKey,
+								Operator: v1.NodeSelectorOpIn,
+								Values:   []string{pool},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by cjohannsen-cloudera — APPROVED** [review]

File: pkg/webhook/admission/pod/affinity_injector.go [hunk 1]
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright 2021 The KServe Authors.
```
spolti: can you please update to 2026?

pkg/webhook/admission/pod/affinity_injector_test.go
```diff
@@ -0,0 +1,144 @@
+/*
+Copyright 2021 The KServe Authors.
```
spolti: same

pkg/webhook/admission/pod/affinity_injector.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+
+	// If the pod already has affinity configured, skip injection.
+	if pod.Spec.Affinity != nil {
+		return nil
+	}
+
+	// Get the label key from environment variable, or use default.
+	labelKey := os.Getenv(DefaultNodepoolLabelKey)
```
spolti, line 45: imho, we should not enforce any default value here.

pkg/webhook/admission/pod/affinity_injector.go
```diff
@@ -0,0 +1,74 @@
...
+	DefaultNodepoolLabelKey = "KSERVE_DEFAULT_NODEPOOL_LABEL_KEY"
+	defaultNodepoolLabelKey = "cloud.google.com/gke-nodepool"
+)
+
+// InjectAffinity injects a default affinity config (configurable through environment variables) to each kserve pod
+// if it is created without affinity. This allows cluster administrators to route inference workloads to specific
+// node pools by default.
+func InjectAffinity(pod *v1.Pod) error {
+	// If the default nodepool is not set in environment variable, skip injection.
+	pool := os.Getenv(DefaultNodepoolEnvVar)
```
spolti, line 34: Is there a reason to require it to come from envs? Wondering if it wouldn't be a good idea read it from the `inferenceservice-config` ConfigMap.

pkg/webhook/admission/pod/affinity_injector.go
spolti: with that said, if the cluster does not have the default node pool you are using as default, it could lead to all isvc to not being scheduled because the pool does not exist, and you have used `RequiredDuringSchedulingIgnoredDuringExecution` instead of`preferredDuringSchedulingIgnoredDuringExecution`, which implicates into a hard affinity.

pkg/webhook/admission/pod/affinity_injector.go
neo99zhang: Thanks for the suggestion! We chose environment variables for simplicity since this is cluster-wide default, not per-namespace/service config. Env vars are set once and don't require `ConfigMap` parsing and also follows the same pattern as other webhook level configs

neo99zhang: Happy to refactor to `ConfigMap` if you feel strongly about consistency with other KServe configs.

pkg/webhook/admission/pod/affinity_injector.go
neo99zhang: Thanks for the review! I just removed default and make it a soft affinity.

pkg/webhook/admission/pod/affinity_injector.go
neo99zhang: Just updated

pkg/webhook/admission/pod/affinity_injector_test.go
neo99zhang: Updated as above

neo99zhang:
@spolti Hi Filippe, kindly following up on this PR. thanks!

pkg/webhook/admission/pod/affinity_injector_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package pod
+
+import (
```
spolti, line 18: Thanks for the updates. Can you also add a test in the controllers to make sure things work as expected during the reconciliation loops?

neo99zhang: Hi thanks for the reply! I just added a webhook handle pipeline with affinity injection to test the response is expected. Please let me know what else is desired.

Issue #4950, opened 2026-01-12
Add E2E tests for LLMInferenceService v1alpha2 API


KillianGolds:
### **Background**
PR #4886 introduced the v1alpha2 API for LLMInferenceService with conversion webhooks from v1alpha1. E2E test coverage for the new API version would help ensure stability.

### **Scope**
- E2E tests exercising the v1alpha2 API directly
- Conversion webhook testing (v1alpha1 ↔ v1alpha2 round-trip)
- Verify criticality field preservation via annotations during conversion

### **Related**
PR #4886 - GIE v1.2.0 upgrade and v1alpha2 API migration

KillianGolds:
Currently working on!

PR #4988, opened 2026-01-21
feat(llmisvc): zero-downtime InferencePool migration

### LINKED ISSUE #4949: Add zero-downtime migration strategy for GIE v1alpha2 to v1 InferencePool upgrade
### **Background**
PR #4886 upgraded KServe to use GIE v1 InferencePool. This is a clean break approach as it only creates v1 InferencePools.

For production environments upgrading from existing v1alpha2 deployments, a zero-downtime migration strategy would be valuable.

### **Proposal**
Implement dual-pool creation with gradual traffic migration:

- Dual-pool creation - Controller creates both v1 and v1alpha2 InferencePools simultaneously
- Readiness check for both pools - Report ready if either pool is ready (prioritizing v1)
- Weighted HTTPRoute backendRefs - Gradually shift traffic from v1alpha2 to v1
- One-way migration annotation - Once migrated to v1, prevent flapping back to v1alpha2

### **Benefits**
- Zero traffic disruption during GIE CRD upgrade
- Graceful fallback if v1 pool isn't ready yet
- Clear migration state tracking

KillianGolds:

Implements zero-downtime migration from GIE InferencePool v1alpha2 to v1 for the LLMInferenceService controller. This enables seamless upgrades when Istio transitions from v1alpha2-only (Istio 1.27) to v1 support (Istio 1.28+).

This is a follow-up feature to #4886 (GIE integration) and relates to #4950 (v1 API support).

Key features:
- **Dual-pool creation**: Controller creates both v1 and v1alpha2 InferencePools simultaneously (same name, different API groups)
- **CRD discovery**: Adapts behavior based on which CRDs are installed (v1-only, v1alpha2-only, or both)
- **Auto-migration**: When v1 pool becomes ready, automatically swaps HTTPRoute backendRef from v1alpha2 to v1
- **Migration lock**: One-way annotation (`serving.kserve.io/inferencepool-migrated: v1`) prevents flapping
- **Fallback retained**: v1alpha2 pool kept after migration as fallback

Fixes #4949

**Related**:
- #4886 - GIE integration for LLMInferenceService
- #4950 - Support v1 InferencePool API

- [x] Unit tests for migration helpers (`v1migration_test.go`)
- [x] Integration tests for migration scenarios (`controller_int_migration_test.go`)
- [x] Manual E2E testing on OpenShift cluster

**E2E Test Results (OpenShift + Istio):**

| Test Scenario | Result |
|---------------|--------|
| v1alpha2-only (Istio 1.27) | ✅ Only v1alpha2 pool created, HTTPRoute → v1alpha2 |
| Dual-pool + migration (Istio 1.28) | ✅ Both pools created, auto-migrates to v1 when ready |
| v1-only (future Istio) | ✅ Only v1 pool created, HTTPRoute → v1 |
| Existing deployment upgrade | ✅ Zero-downtime migration, no orphaned resources |

**Cluster setup:**
- OpenShift on AWS
- Istio 1.27.1 → 1.28.0 upgrade tested
- GIE CRDs v1.2.0 (provides both v1 and v1alpha2)

1. Both InferencePools use the same name (`{llmisvc-name}-inference-pool`) since they're different CRDs (different API groups can coexist)
2. Migration is transparent to users - no opt-in required, controller handles everything automatically
3. The migration annotation is set by the controller, not by users

**Checklist**:

**Release note**:
```release-note
Add zero-downtime migration support for GIE InferencePool v1alpha2 to v1. The LLMInferenceService controller now creates dual InferencePools and automatically migrates traffic to v1 when ready, enabling seamless Istio upgrades from 1.27 to 1.28+.

### CODE REVIEW

**Comment by KillianGolds** [comment]
cc; @pierDipi @bartoszmajsak @Jooho @spolti

File: pkg/controller/v1alpha2/llmisvc/router_discovery.go [hunk 1]
```diff
@@ -394,3 +395,49 @@ func nonReadyInferencePoolTopLevelCondition(pool *igwapi.InferencePool) (*metav1
...
+	if pool == nil {
+		return false
+	}
+
+	// If no parents have been set, consider the pool ready if it exists and has a valid spec
+	// This handles cases where no Gateway controller is populating the status
+	if len(pool.Status.Parents) == 0 {
+		// Pool is ready if it exists with a valid selector and target port
+		return len(pool.Spec.Selector) > 0 && pool.Spec.TargetPortNumber > 0
+	}
```
pierDipi: Is this accurate? if no GW controller is setting status, it means it's not programmed/configured and won't work as expected

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -154,15 +155,42 @@ func (r *LLMISVCReconciler) reconcileSchedulerDeployment(ctx context.Context, ll
...
+		if shouldDelete {
+			if err := Delete(ctx, r, llmSvc, expectedV1Alpha2); err != nil {
+				return err
+			}
+		} else {
+			if err := Reconcile(ctx, r, llmSvc, &igwapiv1alpha2.InferencePool{}, expectedV1Alpha2, semanticInferencePoolV1Alpha2IsEqual); err != nil {
+				return err
+			}
+		}
 	}
```
pierDipi, line 172: Can we structure this logic in "exit early way" with less nested if/else that easier to read and reason about like

```golang
expectedV1Alpha2 := r.expectedSchedulerInferencePoolV1Alpha2(ctx, llmSvc)
expectedV1 := r.expectedSchedulerInferencePool(ctx, llmSvc)

if  shouldDelete {
   // ... here we don't need to check `InferencePoolV1Available` (Delete handles no API)
   if err := Delete(ctx, r, llmSvc, expectedV1Alpha2); err != nil { 
        return err
   }
   // ... same for the other
}

if r.InferencePoolV1Available {
    if err := Reconcile(ctx, r, llmSvc, &igwapi.InferencePool{}, expectedV1, semanticInferencePoolIsEqual); err != nil {
        return err
    }
}

// ... same for the other
```

pkg/controller/v1alpha2/llmisvc/router_discovery.go
KillianGolds: Good catch! I originally emulated the v1 `IsInferencePoolReady` logic from PR #4886 which had the same behavior. I've now fixed both v1 and v1alpha2 versions in commit ccb39b59a to return false when Status.Parents is empty, since the pool hasn't been programmed by a Gateway controller and won't route traffic correctly.

pkg/controller/v1alpha2/llmisvc/scheduler.go
KillianGolds: Done! I've refactored this in commit 4f339a20d using an "exit early" pattern with helper functions. 

I kept the CRD availability check because the `expected*` functions access `llmSvc.Spec.Router.Scheduler.Pool.Spec` which could be nil possibly in edge cases, and it felt cleaner to skip the whole function when the CRD isn't available. 

Let me know if you'd prefer the structure you suggested instead.

pierDipi:
Ref https://github.com/kserve/kserve/pull/5007

PR #4996, opened 2026-01-26
feat: llmisvc preserves externally managed replicas

pierDipi:
This enables external controllers like HPA to manage Deployment and LeaderWorkerSet replicas without being overwritten by the reconciler.

Add UpdateOption pattern with AfterDryRun hook to allow preserving fields from the current resource state when the owner doesn't explicitly set them.

Changes:
- Add UpdateOption[T] and AfterDryRun hook to lifecycle_crud.go
- Add PreserveDeploymentReplicas() helper for Deployments
- Add PreserveLWSReplicas() helper for LeaderWorkerSets
- Apply options to all Deployment and LWS Reconcile calls
- Add integration tests for both preserve and override scenarios

**Release note**:

```release-note
feat(llmisvc): preserve externally managed replicas when user does not specify them.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/controller_int_multi_node_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/controller_int_multi_node_test.go
@@ -489,5 +489,185 @@ var _ = Describe("LLMInferenceService Multi-Node Controller", func() {
 
 			Expect(expectedLWS.Spec.LeaderWorkerTemplate.WorkerTemplate.Annotations).To(HaveKeyWithValue(PreemptionReclaimAnnotationKey, preemptPriority))
 		})
+
+		It("should preserve externally set replicas when owner does not specify replicas", func(ctx SpecContext) {
+			// given
+			svcName := "test-llm-mn-preserve-replicas"
+			nsName := kmeta.ChildName(svcName, "-test")
+			namespace := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: nsName,
+				},
+			}
+
+			Expect(envTest.Client.Create(ctx, namespace)).To(Succeed())
+			Expect(envTest.Client.Create(ctx, IstioShadowService(svcName, nsName))).To(Succeed())
+			defer func() {
+				envTest.DeleteAll(namespace)
+			}()
+
+			// Create LLMInferenceService WITHOUT specifying replicas
+			llmSvc := LLMInferenceService(svcName,
+				InNamespace[*v1alpha2.LLMInferenceService](nsName),
+				WithModelURI("hf://facebook/opt-125m"),
+				WithParallelism(ParallelismSpec(
+					WithDataParallelism(2),
+					WithDataLocalParallelism(1),
+				)),
+				WithWorker(&corev1.PodSpec{}),
+				WithManagedRoute(),
+				WithManagedGateway(),
+				// Note: Not using WithReplicas() - replicas should be nil
+			)
+
+			// Verify replicas is nil in the spec
+			Expect(llmSvc.Spec.Replicas).To(BeNil())
+
+			// when
+			Expect(envTest.Create(ctx, llmSvc)).To(Succeed())
+			defer func() {
+				Expect(envTest.Delete(ctx, llmSvc)).To(Succeed())
+			}()
+
+			// then - LeaderWorkerSet should be created with default replicas (1)
+			expectedLWS := &lwsapi.LeaderWorkerSet{}
+			Eventually(func(g Gomega, ctx context.Context) error {
+				return envTest.Get(ctx, types.NamespacedName{
+					Name:      svcName + "-kserve-mn",
+					Namespace: nsName,
+				}, expectedLWS)
+			}).WithContext(ctx).Should(Succeed())
+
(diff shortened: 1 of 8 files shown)


pkg/controller/v1alpha2/llmisvc/controller_int_multi_node_test.go
```diff
@@ -489,5 +489,185 @@ var _ = Describe("LLMInferenceService Multi-Node Controller", func() {
...
+			}).WithContext(ctx).Should(Succeed())
+
+			Expect(expectedLWS.Spec.Replicas).To(Equal(ptr.To[int32](1)))
+
+			// Simulate external scaling (e.g., HPA scaling to 3 replicas)
+			errRetry := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				lws := &lwsapi.LeaderWorkerSet{}
+				if err := envTest.Get(ctx, types.NamespacedName{
+					Name:      svcName + "-kserve-mn",
+					Namespace: nsName,
```
bartoszmajsak: nit, but maybe worth extracting as test fixture? We have repeated 3 times in this test, maybe name it e.g. `lwsNsName`.

pkg/controller/v1alpha2/llmisvc/controller_int_multi_node_test.go
```diff
@@ -489,5 +489,185 @@ var _ = Describe("LLMInferenceService Multi-Node Controller", func() {
...
+			})
+			Expect(errRetry).ToNot(HaveOccurred())
+
+			// Trigger a reconciliation by updating the LLMInferenceService
+			errRetry = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				_, errUpdate := ctrl.CreateOrUpdate(ctx, envTest.Client, llmSvc, func() error {
+					if llmSvc.Annotations == nil {
+						llmSvc.Annotations = make(map[string]string)
+					}
+					llmSvc.Annotations["reconcile-trigger"] = "true"
```
bartoszmajsak: Is this made-up annotation for test purposes? Probably not really relevant right now, but it's not rare to use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#GenerationChangedPredicate where only `.spec` field triggers reconcile as an optimization technique. I wonder if we could make some changes in the spec itself to future-proof it?

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
bartoszmajsak, line 172: I just wonder if we can't rely on `MutateFn` here. Looks somewhat similar?

https://github.com/kubernetes-sigs/controller-runtime/blob/18e7485d3a47b26d95ec47caf2393cbb159f7a4f/pkg/controller/controllerutil/controllerutil.go#L305-L321

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
```diff
@@ -179,6 +191,28 @@ func logLineForObject(obj client.Object) string {
 // This allows custom comparison logic beyond simple equality checks
 type SemanticEqual[T client.Object] func(expected T, curr T) bool
 
+// UpdateOption is a functional option for the Update and Reconcile functions
+type UpdateOption[T client.Object] func(*updateOptions[T])
+
+type updateOptions[T client.Object] struct {
+	afterDryRun func(expected, expectedGiven, curr T)
```
bartoszmajsak: Maybe for simplicity make an alias type for `func(expected, expectedGiven, curr T)` so we can also use it below?

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
```diff
@@ -116,12 +116,17 @@ func Reconcile[O client.Object, T client.Object](ctx context.Context, c clientWi
...
 }
 
 // Update modifies an existing Kubernetes resource to match the expected state
 // It validates ownership and only updates if the resource has actually changed
-func Update[O client.Object, T client.Object](ctx context.Context, c clientWithRecorder, owner O, curr, expected T, isEqual SemanticEqual[T]) error {
+func Update[O client.Object, T client.Object](ctx context.Context, c clientWithRecorder, owner O, curr, expected T, isEqual SemanticEqual[T], opts ...UpdateOption[T]) error {
+	options := &updateOptions[T]{}
+	for _, opt := range opts {
+		opt(options)
+	}
```
bartoszmajsak, line 128: Wouldn't it mean only last "afterDryRun" gets applied? Is it what we want?

bartoszmajsak (commented):
Few considerations which we might want to iron out early.

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
pierDipi: maybe we can take a refactoring as a follow up? MutateFn is a function without args and I like the self-contained / pure function for dry run option more

pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go
pierDipi: I've added an array now

spolti (approved):
/lgtm

bartoszmajsak (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #4997, opened 2026-01-26
fix: removed GIE crd from llmisvc kustomzation/Makefile

Jooho:
- removed GIE crd from llmisvc kustomzation/Makefile because it is dependency that should be installed in advance
- remove llmisvcconfig from llmisvc kustomzation.yaml 
  - It should be installed separately.
Fixes #

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:
```release-note

```

### DIFF
--- a/config/llmisvc/kustomization.yaml
+++ b/config/llmisvc/kustomization.yaml
@@ -5,11 +5,9 @@ namespace: kserve
 
 resources:
   - service.yaml
-  - manager.yaml
-  - gateway-inference-extension.yaml
+  - manager.yaml  
   - ../certmanager/llmisvc
   - ../rbac/llmisvc
-  - ../llmisvcconfig
   - ../webhook/llmisvc
 
 replacements:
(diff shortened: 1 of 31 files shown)


config/llmisvc/kustomization.yaml
```diff
@@ -5,11 +5,9 @@ namespace: kserve
 
 resources:
   - service.yaml
-  - manager.yaml
-  - gateway-inference-extension.yaml
+  - manager.yaml  
   - ../certmanager/llmisvc
   - ../rbac/llmisvc
-  - ../llmisvcconfig
```
pierDipi, line 12: just GIE or llmisvcconfig too?

Jooho: Both. When I submitted the PR for the component-based Kustomize manifests, the llmisvcconfig was mistakenly not removed. :)  If it needs, I can send a separate PR only for that

PR #5006, opened 2026-01-28
fix(llmisvc): inject CSC envs in storage-initializer

### LINKED ISSUE #4987: LLMInferenceService controller doesn't respect URI prefix to configure storage-initializer
Kserve version: 0.16

I have the following llmisvc from docs:
```
kind: LLMInferenceService
metadata:
  name: llama-3-8b
  namespace: kserve-test
spec:
  model:
    uri: hf://meta-llama/Llama-3.1-8B-Instruct
    name: meta-llama/Llama-3.1-8B-Instruct
  replicas: 2

  template:
    containers:
      - name: main
        image: img-proxy.io/dhub/vllm/vllm-openai:latest
        resources:
          limits:
            nvidia.com/mig-1g.10gb: '1'
            cpu: "2"
            memory: 8Gi

  router:
    gateway: {}     # Managed gateway
    route: {}       # Managed HTTPRoute
    scheduler: {}   # Managed scheduler
```

And I have ClusterStorageContainer config with some env variables:
```
kind: ClusterStorageContainer
metadata:
  name: default
spec:
  container:
    env:
      - name: HTTP_PROXY
        value: http://stage-staging-webproxy.example.com:3128
    image: img-proxy.io/dhub/kserve/storage-initializer:v0.16.0
    name: storage-initializer
    resources:
      limits:
        cpu: '1'
        memory: 1Gi
      requests:
        cpu: 100m
        memory: 100Mi
    securityContext:
      capabilities:
        drop:
          - ALL
      privileged: false
      runAsNonRoot: true
  supportedUriFormats:
    - prefix: gs://
    - prefix: s3://
    - prefix: hdfs://
    - prefix: hf://
    - prefix: webhdfs://
    - regex: https://(.+?).blob.core.windows.net/(.+)
    - regex: https://(.+?).file.core.windows.net/(.+)
    - regex: https?://(.+)/(.+)
  supportsMultiModelDownload: true
  workloadType: initContainer
```

So LLMInferenceService controller should respect `hf://` prefix from LLMInferenceService and create storage-initializer init container with envs described in the ClusterStorageContainer. But I have storage-initializer without any env variables. Is it intended behavior or a bug? If that's intended the please tell me how to pass custom environment variables to storage-initializer.

EdgeN8v:
**Description**
This PR fixes an issue where environment variables defined in `ClusterStorageContainer` were not being injected into the `storage-initializer` initContainer for `LLMInferenceService`.

Previously, the controller did not apply the `ClusterStorageContainer` configuration for `LLMInferenceService`, which caused proxy-related environment variables (e.g., `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`) to be missing during model download.

**Link to Issue**
Fixes #4987

**Type of change**
- Bug fix (non-breaking change which fixes an issue)

* Added the required RBAC markers (`//+kubebuilder:rbac...`) and manually updated `config/rbac/llmisvc/role.yaml` to match.
* I verified the behavior locally with Kind.
* I ran the unit tests for the modified package (`pkg/controller/v1alpha2/llmisvc`), and they passed.
* Note: I could not run the full `make test` suite locally due to environment limitations (WSL resource constraints causing timeouts), but the relevant controller tests passed successfully.

**Verification**:
* Deployed the patched controller in a Kind cluster and verified that `HTTP_PROXY` defined in a `ClusterStorageContainer` is correctly propagated to the `storage-initializer` container.
* Ran unit tests for the modified controller:
  `go test ./pkg/controller/v1alpha2/llmisvc/... -count=1` -> **PASS**


pierDipi (approved):
/cc @brettmthompson

pierDipi:
@EdgeN8v FYI tests are failing currently which seems related to the change

```
kserve-ci-e2e-test     2m20s       Warning   Error               llminferenceservice/llmisvc-model-fb-opt-125m-route-ddce928e               Reconciliation failed: failed to reconcile workload: failed to reconcile multi node workload: failed to reconcile multi-node main workload: failed to build the expected main LWS: failed to attach model artifacts to leader template: no matches for kind "ClusterStorageContainer" in version "serving.kserve.io/v1alpha1"
```

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -251,6 +252,28 @@ func (r *LLMISVCReconciler) attachStorageInitializer(modelUri string, podSpec *c
 		ReadOnly:   false,
 	}
 	initContainer := utils.CreateInitContainerWithConfig(storageConfig, containerArgs)
+
+	// List ClusterStorageContainers to find a match for the model URI
+	cscList := &v1alpha1.ClusterStorageContainerList{}
+	if err := r.List(ctx, cscList); err != nil {
```
pierDipi: given the error in CI, the llmisvc chart is independent from the existing Kserve core chart, I think we might want to check for `NoKindMatchError`

pkg/controller/v1alpha2/llmisvc/workload_storage.go
pierDipi, line 1: can we add tests in `pkg/controller/llmisvc/controller_workload_storage_int_test.go`

brettmthompson:
We should be able to simplify this approach by utilizing already exposed methods [GetStorageContainerSpec](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L118) and [MergeContainerSpecs](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/storage_initializer_injector.go#L734) (requires export).

In [attachStorageInitializer](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha2/llmisvc/workload_storage.go#L243) the approach would be as follows:
```

func (r *LLMISVCReconciler) attachStorageInitializer(ctx context.Context, modelUri string, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig) error {
	containerArgs := []string{
		modelUri,
		constants.DefaultModelLocalMountPath,
	}
	storageMountParams := utils.StorageMountParams{
		MountPath:  constants.DefaultModelLocalMountPath,
		VolumeName: constants.StorageInitializerVolumeName,
		ReadOnly:   false,
	}
	// Create default storage initializer container
	initContainer := utils.CreateInitContainerWithConfig(storageConfig, containerArgs)

	// Check if a custom ClusterStorageContainer exists for this URI
	storageContainerSpec, err := pod.GetStorageContainerSpec(ctx, modelUri, r.Client)
	if err != nil {
		return fmt.Errorf("failed to get storage container spec for URI %s: %w", modelUri, err)
	}

	// If custom storage container spec exists, merge it with the default container
	if storageContainerSpec != nil {
		if err := pod.MergeContainerSpecs(initContainer, &storageContainerSpec.Container); err != nil {
			return fmt.Errorf("failed to merge custom storage container spec: %w", err)
		}
	}

	// Append the init container to the pod spec
	podSpec.InitContainers = append(podSpec.InitContainers, *initContainer)

	utils.AddModelMount(storageMountParams, initContainer.Name, podSpec)

	storageMountParams.ReadOnly = true
	utils.AddModelMount(storageMountParams, "main", podSpec)

	return nil
}
```

brettmthompson:
> Added the required RBAC markers (//+kubebuilder:rbac...) and manually updated config/rbac/llmisvc/role.yaml to match.

`make precommit` will auto update the manifests based on the added marker. Please ensure this is run as it looks like some file updates were missed, specifically `kserve/charts/kserve-llmisvc-resources/templates/clusterrole.yaml`

pkg/controller/v1alpha2/llmisvc/workload_storage.go
brettmthompson: Since llmisvc will now support cluster storage containers, shouldn't we ensure that crd is installed as a part of llmisvc e2e tests setup so that we can properly test that functionality?

EdgeN8v:
@Brettmthompson @pierDipi I refactored the CSC logic to use the shared helper (`GetStorageContainerSpec`), exported `MergeContainerSpecs`, and added `meta.IsNoMatchError` so we don’t fail when the CSC CRD isn’t installed.
Ran `go test ./pkg/controller/v1alpha2/llmisvc/... -count=1` locally — looks good. Re-review when you have time.

EdgeN8v:
@pierDipi Just a friendly ping. Could you please help approve the CI workflows? I'd like to verify that the checks pass so we can move forward. Thanks!

spolti (approved):
Thanks!

Approving, can be merged once the tests are added as suggested by @pierDipi

brettmthompson:
@EdgeN8v I think you still need to run `make precommit` to update the quick install scripts and charts. Other than that changes look good.

@pierDipi @spolti we should determine if we want the `ClusterStorageContainer` CRD to be installed as a part of llmisvc installation since we are now supporting it. It is not treated as optional for isvcs, i.e. no `IsNoMatchError` check

brettmthompson:
Based on the kubebuilder comment added `precommit` should only make the following update to these files:
- `charts/kserve-llmisvc-resources/templates/clusterrole.yaml`
- `config/rbac/llmisvc/role.yaml`
- `hack/setup/quick-install/kserve-knative-mode-full-install-with-manifests.sh`
- `hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh`
- `/hack/setup/quick-install/llmisvc-full-install-with-manifests.sh`

```
- apiGroups:
  - serving.kserve.io
  resources:
  - clusterstoragecontainers
  verbs:
  - get
  - list
  - watch
  ```
  
@EdgeN8v  thanks for running the command, but it looks like a lot of other files were affected which makes review difficult. Could you please check?

EdgeN8v:
@brettmthompson Thanks, I've updated the PR to only include the RBAC changes and the controller logic.

@pierDipi @spolti Before I add the tests, could you clarify the direction on the `ClusterStorageContainer` CRD? Should it be mandatory or optional?
Once decided, I'll update the code and add the corresponding tests. Thanks!

EdgeN8v:
@pierDipi @spolti @brettmthompson I've just pushed the integration tests. I went ahead with treating the CSC as optional (gracefully skip if not found) for now to keep things moving. If you'd rather have it mandatory, let me know and I'll update the logic. Ready for review! 🚀

EdgeN8v:
@pierDipi @spolti @brettmthompson Conflicts resolved. Added the requested integration tests and integrated the latest master changes. Ready for a final look and CI run. Thanks!

brettmthompson:
@EdgeN8v could you please resolve the conflicts. I will request the CI checks/tests to be triggered once resolved.

PR #5007, opened 2026-01-29
feat(llmisvc): inferencePool auto-migration

### LINKED ISSUE #4949: Add zero-downtime migration strategy for GIE v1alpha2 to v1 InferencePool upgrade

KillianGolds:

Implements automatic migration from GIE InferencePool v1alpha2 to v1 for the LLMInferenceService controller. This enables seamless upgrades when Gateway controllers transition from v1alpha2 to v1 support (e.g., Istio 1.27 → 1.28+, Envoy AI Gateway).

This is a follow-up feature to #4886 (GIE integration) and relates to #4950 (v1 API support).

**Why not dual backendRefs?**

The [GIE GA Migration Guide](https://gateway-api-inference-extension.sigs.k8s.io/guides/ga-migration/#zero-downtime-migration) documents zero-downtime migration using weighted dual backendRefs (50/50 split between v1alpha2 and v1 pools). However, this requires Gateway controllers to support both API groups simultaneously—which none currently do (Envoy AI Gateway supports v1 only, Istio 1.28+ removed v1alpha2). Since dual backendRefs cannot work in practice, we pivoted to runtime detection.

**Key features:**
- **Dual-pool creation**: Controller creates both v1 and v1alpha2 InferencePools simultaneously (same name, different API groups)
- **CRD discovery**: Adapts behavior based on which CRDs are installed (v1-only, v1alpha2-only, or both)
- **Runtime detection**: HTTPRoute initially points to v1alpha2; if Gateway rejects it (`ResolvedRefs=False`, `Reason=InvalidKind`), automatically swaps to v1 on next reconcile
- **One-way migration lock**: Annotation `serving.kserve.io/inference-pool-migrated: v1` on HTTPRoute prevents flapping back to v1alpha2

**Note**: When Gateway rejects v1alpha2, there may be a brief service interruption while the controller detects the rejection and swaps to v1 on the next reconciliation.

Fixes #4949

**Related**:
- #4886 - GIE integration for LLMInferenceService
- #4950 - Support v1 InferencePool API

- [x] Unit tests for runtime detection (`router_discovery_test.go`)
- [x] Integration tests for migration scenarios (`controller_int_migration_test.go`)
- [x] E2E tests passed (16/16 in CI)

**Test Coverage:**

| Test Scenario | Test Type | File |
|---------------|-----------|------|
| Both CRDs available → dual pool creation | Integration | `controller_int_migration_test.go` |
| HTTPRoute → v1alpha2 backendRef initially | Integration | `controller_int_migration_test.go` |
| Pre-migrated (HTTPRoute annotation set) → stays on v1 | Integration | `controller_int_migration_test.go` |
| Gateway rejects v1alpha2 → swap to v1, set annotation | Integration | `controller_int_migration_test.go` |
| Migration lock (annotation persists across reconciles) | Integration | `controller_int_migration_test.go` |
| `HasUnsupportedBackendRefError` detection | Unit | `router_discovery_test.go` |

1. Both InferencePools use the same name (`{llmisvc-name}-inference-pool`) since they're different CRDs (different API groups can coexist)
2. Runtime detection relies on Gateway API standard condition: `ResolvedRefs=False` with `Reason=InvalidKind` (used by both Envoy AI Gateway and Istio)
3. Migration annotation is stored on the **managed HTTPRoute** - this is the resource that needs migration tracking
4. Migration only applies to managed routes; user-provided routes (baseRefs) are the user's responsibility

**Checklist**:

**Release note**:
```release-note
Add automatic migration support for GIE InferencePool v1alpha2 to v1. The LLMInferenceService controller creates dual InferencePools and uses runtime detection to swap HTTPRoute backendRefs from v1alpha2 to v1 when the Gateway rejects v1alpha2, enabling seamless Gateway controller upgrades.

### CODE REVIEW

File: pkg/controller/v1alpha2/llmisvc/controller.go [hunk 1]
```diff
@@ -159,6 +165,14 @@ func (r *LLMISVCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		r.Eventf(original, corev1.EventTypeWarning, "Error", "Reconciliation failed: %v", reconcileErr.Error())
 	}
 
+	// Update annotations if they changed (e.g., migration annotation)
+	if !equality.Semantic.DeepEqual(original.Annotations, resource.Annotations) {
+		if err := r.updateAnnotations(ctx, resource); err != nil {
+			logger.Error(err, "Failed to update annotations for LLMInferenceService")
+			return ctrl.Result{}, err
+		}
+	}
```
pierDipi: we can't touch spec or metadata of user-provided resources

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -239,6 +256,31 @@ func (r *LLMISVCReconciler) updateStatus(ctx context.Context, desired *v1alpha2.
...
+
+		latest.Annotations = desired.Annotations
+
+		if err := r.Client.Update(ctx, latest); err != nil {
+			return fmt.Errorf("failed to update annotations for LLMInferenceService: %w", err)
+		}
+
+		return nil
+	})
+}
```
pierDipi: see above

pierDipi (commented):
Some initial feedback

pkg/controller/v1alpha2/llmisvc/controller.go
KillianGolds: Thanks for the feedback, woops you're right, I've reworked the approach to use stateless runtime detection instead. The idea is that the HTTPRoute status already tells us whether the Gateway rejected the v1alpha2 backendRef, so we can just check that each reconcile rather than persisting state ourselves.

**Changes:**
- Removed `updateAnnotations()` and the annotation update call
- Added `detectV1Alpha2Rejected()` that reads HTTPRoute status
- Removed the functions that were writing to the LLMInferenceService
- Kept `isMigratedToV1()` since it only reads a user-provided opt-in annotation

Let me know if this approach works or if you'd prefer something different!.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -105,7 +111,7 @@ func (r *LLMISVCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		WithValues("Namespace", req.Namespace, "Name", req.Name)
 	ctx = log.IntoContext(ctx, logger)
 
-	logger.Info("Starting reconciliation")
+	logger.Info("DEBUG: Starting reconciliation (check what triggered this)")
```
pierDipi:
```suggestion
	logger.Info("Starting reconciliation")
```

maybe orthogonal we could log what's triggering reconciliations in the various handlers as separate PR

KillianGolds: Understood and addressed!

pierDipi (changes requested):
I've tested this with Istio 1.27.1 and I'm getting the pool migrated to v1 but that Istio version doesn't support it

```
[pierdipi] e2e (gie-migration-runtime-detection-upstream) (k8s:kind-kind)
-> $ k get httproutes.gateway.networking.k8s.io -oyaml
apiVersion: v1
items:
- apiVersion: gateway.networking.k8s.io/v1
  kind: HTTPRoute
  metadata:
    annotations:
      serving.kserve.io/inference-pool-migrated: v1
    creationTimestamp: "2026-02-02T11:41:36Z"
    generation: 1
    labels:
      app.kubernetes.io/component: llminferenceservice-router
      app.kubernetes.io/name: facebook-opt-125m-single
      app.kubernetes.io/part-of: llminferenceservice
    name: facebook-opt-125m-single-kserve-route
    namespace: default
    ownerReferences:
    - apiVersion: serving.kserve.io/v1alpha2
      blockOwnerDeletion: true
      controller: true
      kind: LLMInferenceService
      uid: 4fe184d4-57eb-4f01-aaab-2c607ec596cd
    resourceVersion: "2629"
    uid: e67cbe8e-f785-4fec-b39c-752f04c8a7c6
  spec:
    parentRefs:
      kind: Gateway
      name: kserve-ingress-gateway
      namespace: kserve
    rules:
    - backendRefs:
      - group: inference.networking.k8s.io
        kind: InferencePool
        name: facebook-opt-125m-single-inference-pool
        port: 8000
        weight: 1
      filters:
      - type: URLRewrite
        urlRewrite:
          path:
            replacePrefixMatch: /
            type: ReplacePrefixMatch
      matches:
      - path:
          type: PathPrefix
          value: /default/facebook-opt-125m-single
      timeouts:
        backendRequest: 0s
        request: 0s
  status:
    parents:
    - conditions:
      - lastTransitionTime: "2026-02-02T11:41:36Z"
        message: Route was valid
        observedGeneration: 1
        reason: Accepted
        status: "True"
        type: Accepted
      - lastTransitionTime: "2026-02-02T11:41:36Z"
        message: 'referencing unsupported backendRef: group "inference.networking.k8s.io"
          kind "InferencePool"'
        observedGeneration: 1
        reason: InvalidKind
        status: "False"
        type: ResolvedRefs
      controllerName: istio.io/gateway-controller
      parentRef:
        kind: Gateway
        name: kserve-ingress-gateway
        namespace: kserve
kind: List
metadata:
  resourceVersion: ""
[pierdipi] e2e (gie-migration-runtime-detection-upstream) (k8s:kind-kind)
-> $ k get inferencepools.inference.networking.x-k8s.io -oyaml
apiVersion: v1
items:
- apiVersion: inference.networking.x-k8s.io/v1alpha2
  kind: InferencePool
  metadata:
    creationTimestamp: "2026-02-02T11:41:36Z"
    generation: 1
    labels:
      app.kubernetes.io/component: llminferenceservice-router-scheduler
      app.kubernetes.io/name: facebook-opt-125m-single
      app.kubernetes.io/part-of: llminferenceservice
    name: facebook-opt-125m-single-inference-pool
    namespace: default
    ownerReferences:
    - apiVersion: serving.kserve.io/v1alpha2
      blockOwnerDeletion: true
      controller: true
      kind: LLMInferenceService
      uid: 4fe184d4-57eb-4f01-aaab-2c607ec596cd
    resourceVersion: "2604"
    uid: 6a13266c-1faa-4cb4-a4d2-7d4d46d4c651
  spec:
    extensionRef:
      failureMode: FailOpen
      group: ""
      kind: Service
      name: facebook-opt-125m-single-epp-service
      portNumber: 9002
    selector:
      app.kubernetes.io/name: facebook-opt-125m-single
      app.kubernetes.io/part-of: llminferenceservice
      kserve.io/component: workload
    targetPortNumber: 8000
  status:
    parent:
    - conditions:
      - lastTransitionTime: "1970-01-01T00:00:00Z"
        message: Waiting for controller
        reason: Pending
        status: Unknown
        type: Accepted
      parentRef:
        kind: Status
        name: default
kind: List
metadata:
  resourceVersion: ""
[pierdipi] e2e (gie-migration-runtime-detection-upstream) (k8s:kind-kind)
-> $ k get inferencepools.inference.networking.k8s.io -oyaml
apiVersion: v1
items:
- apiVersion: inference.networking.k8s.io/v1
  kind: InferencePool
  metadata:
    creationTimestamp: "2026-02-02T11:41:36Z"
    generation: 1
    labels:
      app.kubernetes.io/component: llminferenceservice-router-scheduler
      app.kubernetes.io/name: facebook-opt-125m-single
      app.kubernetes.io/part-of: llminferenceservice
    name: facebook-opt-125m-single-inference-pool
    namespace: default
    ownerReferences:
    - apiVersion: serving.kserve.io/v1alpha2
      blockOwnerDeletion: true
      controller: true
      kind: LLMInferenceService
      uid: 4fe184d4-57eb-4f01-aaab-2c607ec596cd
    resourceVersion: "2599"
    uid: bb52fa62-bec2-40dc-acbd-5dc49c3d0f05
  spec:
    endpointPickerRef:
      failureMode: FailOpen
      group: ""
      kind: Service
      name: facebook-opt-125m-single-epp-service
      port:
        number: 9002
    selector:
      matchLabels:
        app.kubernetes.io/name: facebook-opt-125m-single
        app.kubernetes.io/part-of: llminferenceservice
        kserve.io/component: workload
    targetPorts:
    - number: 8000
kind: List
metadata:
  resourceVersion: ""
```

pierDipi:
I've tested the additional diff here https://github.com/KillianGolds/kserve-fork/pull/1 with:
- starting with istio 1.27.1 (controller settles on using InferencePool v1alpha2)
- Upgraded to Istio 1.28.3 (controller upgrades to using InferencePool v1)

## Installation

```bash
  # Install Istio
  echo "Installing Istio as the Gateway implementation..."

  helm repo add istio https://istio-release.storage.googleapis.com/charts --force-update

  echo "Installing istio-base ${ISTIO_VERSION}..."
  helm upgrade --install istio-base istio/base \
    --namespace istio-system \
    --create-namespace \
    --version "${ISTIO_VERSION}" \
    --set defaultRevision=default \
    --wait

  echo "Installing istiod ${ISTIO_VERSION}..."
  helm upgrade --install istiod istio/istiod \
    --namespace istio-system \
    --version "${ISTIO_VERSION}" \
    --set proxy.autoInject=disabled \
    --set pilot.env.ENABLE_GATEWAY_API_INFERENCE_EXTENSION="true" \
    --set pilot.env.PILOT_ENABLE_GATEWAY_API="true" \
    --set-string pilot.podAnnotations."cluster-autoscaler\.kubernetes\.io/safe-to-evict"=true \
    --wait

  echo "Installing istio-ingressgateway ${ISTIO_VERSION}..."
  helm upgrade --install istio-ingressgateway istio/gateway \
    --namespace istio-system \
    --version "${ISTIO_VERSION}" \
    --set-string podAnnotations."cluster-autoscaler\.kubernetes\.io/safe-to-evict"=true

  kubectl wait --timeout=5m -n istio-system deployment/istiod --for=condition=Available
  kubectl wait --timeout=5m -n istio-system deployment/istio-ingressgateway --for=condition=Available
```

## Upgrade

```bash
helm upgrade istio-base istio/base --namespace istio-system --version "1.28.3" --set defaultRevision=default --wait

helm upgrade istiod istio/istiod --namespace istio-system --version "1.28.3" --set proxy.autoInject=disabled --set pilot.env.ENABLE_GATEWAY_API_INFERENCE_EXTENSION="true" --set pilot.env.PILOT_ENABLE_GATEWAY_API="true" --set-string pilot.podAnnotations."cluster-autoscaler\.kubernetes\.io/safe-to-evict"=true --wait

helm upgrade istio-ingressgateway istio/gateway --namespace istio-system --version "1.28.3" --set-string podAnnotations."cluster-autoscaler\.kubernetes\.io/safe-to-evict"=true
```

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -184,6 +192,78 @@ func (r *LLMISVCReconciler) expectedHTTPRoute(ctx context.Context, llmSvc *v1alp
...
+	// Migration logic: check if we should switch from v1alpha2 to v1 InferencePool
+	// Only applies to managed routes with a scheduler (not using external pool refs)
+	if llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Scheduler == nil ||
+		llmSvc.Spec.Router.Scheduler.Pool.HasRef() {
+		return httpRoute
+	}
+
+	logger := log.FromContext(ctx).WithName("migration").
+		WithValues("InferencePoolV1Alpha2Available", r.InferencePoolV1Alpha2Available,
+			"InferencePoolV1Available", r.InferencePoolV1Available)
```
pierDipi: since we're adding these 2 values in `reconcileRouter` we probably don't need to re-add them we could add an additional field without renaming the logger with something like this:

```suggestion
	logger := log.FromContext(ctx).WithValues("migration", "InferencePool")
```

KillianGolds: Addressed thank you! https://github.com/kserve/kserve/pull/5007/commits/92e65d9559a9a667019c64b635f016e5bd85903a

KillianGolds:
## Istio E2E Testing Overview

### Scenario A: Istio 1.27.1 (No Migration)

| Check | Result |
|-------|--------|
| HTTPRoute created with v1alpha2 | ✅ `inference.networking.x-k8s.io` |
| Istio accepts v1alpha2 | ✅ `ResolvedRefs=True` |
| No migration annotation | ✅ None added |
| Route stays on v1alpha2 | ✅ Stable |

**Key log**: `infPoolV1Alpha2Support=True`

---

### Scenario B: Istio 1.28.3 (Migration Triggered)

| Step | Result |
|------|--------|
| Initial HTTPRoute | v1alpha2 (`inference.networking.x-k8s.io`) |
| Istio 1.28.3 rejection | `ResolvedRefs=False`, `InvalidKind` |
| Controller detection | `infPoolV1Alpha2Support=False` |
| Migration to v1 | `inference.networking.k8s.io` |
| Migration annotation | `serving.kserve.io/inference-pool-migrated: v1` |
| Istio accepts v1 | `ResolvedRefs=True` |

**Key logs**:
**Before migration** `infPoolV1Alpha2Support:False` and `isMigrated:false`

**After migration** `infPoolV1Support:True` and `isMigrated:true`

---

### Three-State Logic Verified

| State | Condition | Action |
|-------|-----------|--------|
| Unknown | First reconcile, no status yet | Use v1alpha2 |
| Supported | `ResolvedRefs=True` on v1alpha2 | Stay on v1alpha2 |
| Not Supported | `ResolvedRefs=False`, `InvalidKind` | Migrate to v1 |

pierDipi (approved):
/lgtm
/approve

Jooho (approved):
/lgtm
/approve

PR #5020, opened 2026-02-01
feat(modelserver): add asyncmodel for sync predict

### LINKED ISSUE #5019: Feature request: AsyncModel to avoid event-loop blocking for sync predict()
## Why
A lot of KServe custom predictors are implemented with a synchronous `predict()` method, which is natural for many MLEs and aligns with most existing examples. **But when `predict()` contains blocking I/O or long-running inference, it runs on the asyncio event loop and can block the entire server, causing stalls and timeouts.** And this happens a lot.

Yes, KServe supports async `predict()` (see discussion in #3950), but in practice:
- Many MLEs are not familiar with async refactoring.
- Existing models already contain blocking I/O deep inside libraries without async alternatives.
- Converting all blocking code to non‑blocking is risky and often unrealistic for production systems.

So teams are left with a hard choice: rewrite everything to async or accept degraded concurrency. We need a safer, ergonomic middle path.

## What
Introduce an `AsyncModel` base class that automatically offloads a sync `predict()` to a threadpool executor so the event loop stays responsive.

## How (high level)
- Detect sync `predict()` on subclass and wrap it with `run_in_executor(...)`
- Use the loop’s default executor (configured by `ModelServer --max_asyncio_workers`)
- Optionally allow an env override for executor size

## References
- https://github.com/kserve/kserve/issues/3950
- https://github.com/kserve/kserve/discussions/3592

## Proposal update
- PR opened with a concrete implementation suggestion: https://github.com/kserve/kserve/pull/5020

anencore94:
Adds an `AsyncModel` base class that wraps sync `predict()` in `run_in_executor`, preventing event-loop blocking for sync predictors while keeping existing code largely unchanged.

Fixes #5019

- [x] `uv run -m pytest test/test_async_model.py`

- Logs
  - 2 passed

- `AsyncModel.predict` must be an instance method; static/class methods are rejected.

**Checklist**:

**Release note**:
```release-note
Add AsyncModel to offload sync predict() to a threadpool executor to avoid blocking the event loop.
```

### DIFF
--- a/python/kserve/kserve/async_model.py
+++ b/python/kserve/kserve/async_model.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import atexit
 import asyncio
 import functools
 import inspect
@@ -54,6 +55,13 @@ class AsyncModel(Model):
 
     _shared_executor: Optional[ThreadPoolExecutor] = None
     _max_workers: Optional[int] = None
+    _cleanup_registered: bool = False
+
+    @classmethod
+    def _cleanup_executor(cls):
+        if cls._shared_executor is not None:
+            cls._shared_executor.shutdown(wait=False, cancel_futures=True)
+            cls._shared_executor = None
 
     @classmethod
     def _get_executor(cls) -> Optional[ThreadPoolExecutor]:
@@ -66,10 +74,13 @@ class AsyncModel(Model):
                 try:
                     cls._max_workers = int(env_workers)
                 except ValueError:
-                    cls._max_workers = None
+                    cls._max_workers = -1
 
         if cls._max_workers and cls._max_workers > 0:
             cls._shared_executor = ThreadPoolExecutor(max_workers=cls._max_workers)
+            if not cls._cleanup_registered:
+                atexit.register(cls._cleanup_executor)
+                cls._cleanup_registered = True
             return cls._shared_executor
 
         return None
--- a/python/kserve/kserve/async_model.py
+++ b/python/kserve/kserve/async_model.py
@@ -1,4 +1,4 @@
-# Copyright 2024 The KServe Authors.
+# Copyright 2026 The KServe Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/python/kserve/kserve/async_model.py
+++ b/python/kserve/kserve/async_model.py
@@ -1,0 +1,98 @@
+# Copyright 2024 The KServe Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import functools
+import inspect
+import os
+from concurrent.futures import ThreadPoolExecutor
+from typing import Optional
+
+from .model import Model
+
+
+class AsyncModel(Model):
+    """Model base class that offloads sync predict() to a thread executor.
+
+    Example
+    -------
+    from kserve import AsyncModel
+
+    class MyModel(AsyncModel):
+        def predict(self, payload, headers=None):
+            return slow_inference(payload)
+
+    Why
+    Many custom predictors are written with sync predict() and include blocking
+    I/O or long-running inference. When KServe's async __call__() executes a sync
+    predict() directly, the event loop can be blocked, reducing concurrency and
+    causing timeouts. Rewriting the full stack to async is often impractical, so
+    AsyncModel provides a safe default that keeps the event loop responsive while
+    preserving existing sync code.
+
+    How it works
+    1) Subclass defines sync predict() as usual.
+    2) __init_subclass__ wraps it in async + run_in_executor.
+    3) KServe's __call__ sees async predict and awaits it.
+    4) Sync code runs in a thread pool (non-blocking).
+
+    Configuration
+    - ASYNC_MODEL_WORKERS env var for max workers, or ModelServer --max_asyncio_workers.
+    - Default (when using ModelServer): min(32, cpu_count + 4).
+    """
+
+    _shared_executor: Optional[ThreadPoolExecutor] = None
+    _max_workers: Optional[int] = None
+
+    @classmethod
+    def _get_executor(cls) -> Optional[ThreadPoolExecutor]:
+        if cls._shared_executor is not None:
+            return cls._shared_executor
+
+        if cls._max_workers is None:
+            env_workers = os.getenv("ASYNC_MODEL_WORKERS")
+            if env_workers:
+                try:
+                    cls._max_workers = int(env_workers)
+                except ValueError:
+                    cls._max_workers = None
+
+        if cls._max_workers and cls._max_workers > 0:
+            cls._shared_executor = ThreadPoolExecutor(max_workers=cls._max_workers)
+            return cls._shared_executor
+
+        return None
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        predict_attr = cls.__dict__.get("predict")
+        if predict_attr is None:
+            return
+
+        if isinstance(predict_attr, (staticmethod, classmethod)):
+            raise TypeError("AsyncModel.predict must be an instance method.")
+
+        original_predict = predict_attr
+
+        if inspect.iscoroutinefunction(original_predict):
+            return
+
+        @functools.wraps(original_predict)
+        async def async_predict(self, *args, **kwargs):
+            loop = asyncio.get_running_loop()
+            executor = cls._get_executor()
+            bound = functools.partial(original_predict, self, *args, **kwargs)
+            return await loop.run_in_executor(executor, bound)
+
+        setattr(cls, "predict", async_predict)
--- a/python/kserve/kserve/async_model.py
+++ b/python/kserve/kserve/async_model.py
@@ -106,4 +106,4 @@ class AsyncModel(Model):
             bound = functools.partial(original_predict, self, *args, **kwargs)
             return await loop.run_in_executor(executor, bound)
 
-        setattr(cls, "predict", async_predict)
+        cls.predict = async_predict
--- a/python/kserve/test/test_async_model.py
+++ b/python/kserve/test/test_async_model.py
@@ -29,6 +29,19 @@ class SyncPredictModel(AsyncModel):
         return threading.get_ident()
 
 
+@pytest.fixture(autouse=True)
+def reset_async_model_state():
+    SyncPredictModel._cleanup_executor()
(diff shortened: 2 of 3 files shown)


python/kserve/kserve/async_model.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2024 The KServe Authors.
```
spolti:
```suggestion
# Copyright 2026 The KServe Authors.
```

python/kserve/kserve/async_model.py
```diff
@@ -0,0 +1,98 @@
...
+        if cls._shared_executor is not None:
+            return cls._shared_executor
+
+        if cls._max_workers is None:
+            env_workers = os.getenv("ASYNC_MODEL_WORKERS")
+            if env_workers:
+                try:
+                    cls._max_workers = int(env_workers)
+                except ValueError:
+                    cls._max_workers = None
```
spolti: Shouldn't it be -1?

python/kserve/test/test_async_model.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2024 The KServe Authors.
```
spolti:
```suggestion
# Copyright 2026 The KServe Authors.
```

python/kserve/kserve/async_model.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if cls._max_workers and cls._max_workers > 0:
+            cls._shared_executor = ThreadPoolExecutor(max_workers=cls._max_workers)
+            return cls._shared_executor
+
+        return None
+
+    def __init_subclass__(cls, **kwargs):
```
spolti, line 88: The _shared_executor is never cleaned up, shouldn't

python/kserve/kserve/async_model.py
anencore94: Good catch, changed this to use `-1` as the invalid-value sentinel.

`None` is also used for "not initialized yet", so keeping invalid parse results as `None` caused us to retry env parsing on every call. With `-1`, we cache the invalid state once and still fall back to `run_in_executor(None, ...)` (loop default executor).

python/kserve/kserve/async_model.py
anencore94: Agreed. I added explicit cleanup for the shared executor via `_cleanup_executor()` and register it once with `atexit`, so `ThreadPoolExecutor.shutdown(...)` runs on process exit. Also added test coverage for cleanup behavior.

anencore94:
Fixed the precommit-check CI failure (ruff B010 lint violation). Could you please re-review when you get a chance? @spolti

spolti:
@sivanantha321 hey, do you mind taking a look?

python/kserve/kserve/async_model.py
```diff
@@ -0,0 +1,109 @@
...
+import inspect
+import os
+from concurrent.futures import ThreadPoolExecutor
+from typing import Optional
+
+from .model import Model
+
+
+class AsyncModel(Model):
+    """Model base class that offloads sync predict() to a thread executor.
```
sivanantha321, line 27: Do we need a separate class for this ? Why not modify the existing Model Class ? Is there any restriction ?

anencore94: Hi, Sorry for the late check. @sivanantha321 

Thanks for raising this! :)
I've also considered in that way. (Modifying the current Model Class).
Actually, there isn't any hard technical restriction that prevents us from modifying `Model Class` directly. However, the reason why I kept this as a separate `AsyncModel` is to make the behavior **opt-in**, because it could be a unexpected change for specific usecases.

If we changed `Model` to offload every sync `predict()` automatically, existing users could visit the breaking changes; their sync code would start running in a worker thread, requests concurrency would increase, and models that rely on thread affinity or are not thread-safe could be affected.

I thought this changes could affect some users unexpected way. So, it could be dangerous. I think if we want to modifying the existing `Model` class, there should be an graceful deprecation notification before applying this.

anencore94:
Gentle ping for the updates :) @sivanantha321 @spolti

python/kserve/kserve/async_model.py
spolti: As the configuration to enable it must be present to opt-in, what are the difficulties you have find to keep it under the same class?

python/kserve/test/test_async_model.py
```diff
@@ -0,0 +1,73 @@
...
+
+import threading
+from concurrent.futures import ThreadPoolExecutor
+
+import pytest
+
+from kserve import AsyncModel
+
+
+class SyncPredictModel(AsyncModel):
```
spolti, line 25: We could add a end to end test to
Verify if the long prediction is no blocking new ones. Wdyt?

anencore94: Added in 542a45629 — two tests:

- `test_long_predict_does_not_block_new_requests`: deterministic (gated by `threading.Event`); a new request completes while a long predict is still blocked in its worker thread.
- `test_async_model_runs_predicts_concurrently`: N concurrent predicts run in parallel (distinct worker threads, ~delay instead of N*delay), not serialized.

Kept them at the `predict` level since `Model.__call__` just awaits `predict`, so this covers the real awaited path deterministically without a flaky full HTTP e2e. Both fail if predictions serialize (workers=1). Thanks for the suggestion!

python/kserve/kserve/async_model.py
anencore94: You're right. The main reason I kept it as a separate class was to leave `Model.__call__` untouched: `__init_subclass__` rewrites the sync `predict` into a coroutine, so the existing `await self.predict(...)` branch picks it up with zero changes to `Model`. But that's not a hard blocker.

With an explicit opt-in, there's no real reason it can't live under the same class — e.g. `class MyModel(Model, offload_sync_predict=True)`, defaulting to the current sync behavior so nothing changes for existing models. That would also let the built-in predictors (sklearn, xgboost, …) opt in without switching base class, and `AsyncModel` could stay as a thin alias for back-compat.

Would it look better to locate this inside the existing `Model` class as an opt-in? Happy to refactor it that way. WDYT? @spolti

anencore94:
Gentle ping @spolti — no rush at all, but whenever you have a moment I'd value your take on the opt-in approach above (keeping the sync-offload inside `Model` behind an opt-in flag rather than a separate `AsyncModel` class). Just want to keep the PR moving whenever it's convenient for you. Thanks!

spolti:
@Jooho @sivanantha321 can you please help with it while I am out of office?

anencore94:
Thanks @spolti, enjoy your time off! @Jooho @sivanantha321 — whenever you have a chance, your review would be much appreciated. The main open question is whether to keep this as a separate `AsyncModel` class or fold the sync-offload into `Model` behind an opt-in flag (see the thread above); I'm happy to go either way. Thank you!

anencore94:
Gentle follow-up @Jooho @sivanantha321 — no rush, but whenever you have some bandwidth I'd really appreciate a review here. The one open design question is whether to keep this as a separate `AsyncModel` class or fold the sync-offload into `Model` behind an opt-in flag — I'm happy to implement either direction. Thanks a lot!

anencore94:
Gentle ping @Jooho @sivanantha321 🙏 — following up on this one. Since @spolti delegated the review before going OOO, could either of you take a look when you get a chance?

Happy to align on the open design question first: keep `AsyncModel` as a separate class, or fold the sync-`predict` offload into the existing `Model` class behind an opt-in flag? I'll adjust the implementation to whichever direction you prefer. The requested e2e test (verifying a long-running prediction doesn't block new requests) is already in place.

Thanks!

anencore94:
Gentle ping @Jooho @sivanantha321 (and welcome back @spolti 🙂) — this PR has been waiting on a review for a while, so I'd love to get it moving again.

To make it easier to decide, here's the one open question restated with a concrete suggestion:

- **Option A (current):** keep `AsyncModel` as a separate class.
- **Option B:** fold the sync-`predict` offload into the existing `Model` class behind an explicit opt-in flag, and drop `AsyncModel`.

Both @spolti and @sivanantha321 leaned toward B in the threads above, and since the offload only activates when the user opts in, B wouldn't change behavior for existing models. I'm happy to go either way — a quick 👍 on A or B is all I need to proceed.

The earlier request for an e2e test (a long prediction must not block new requests) is already addressed in 542a45629.

Thanks a lot!

anencore94:
Rebased onto the latest `master` — no functional changes, and fresh CI is running now.

Two things that should make this cheaper to review:

**1. The previously red CI was infrastructure, not this PR.** On the last push, `build (3.11)` failed while `build (3.10)` and `build (3.12)` passed on the exact same commit — and this PR is version-agnostic pure-stdlib Python, so there's no plausible 3.11-only path. `test-kourier (kustomize)` failed while all 13 sibling e2e lanes passed, and this PR touches zero Go/e2e/kustomize files (that lane's flakiness was addressed separately on master in 3a87283e). `enforce-all-checks` was just the aggregator, and master has since replaced it entirely (#5792). Locally, the full `python/kserve` suite passes on 3.11 (272 passed), including the 6 tests this PR adds.

**2. A default for the one open design question,** so it stops blocking the PR. Unless someone prefers Option B by **Sep 5**, I'll keep **Option A (a separate `AsyncModel` class, as currently implemented)** and consider it settled. I'm genuinely happy to implement Option B instead (fold the sync-`predict` offload into `Model` behind an opt-in flag) — I just don't want the PR to sit indefinitely on a question I can decide either way.

@spolti @sivanantha321 — recap of the problem: a synchronous `predict()` runs inline on the event loop, so a slow sync model blocks the entire server, including the readiness path. The probe keeps reporting healthy while no new request can be accepted. This adds an opt-in offload path for sync models with no behavior change for existing ones.

anencore94:
Following up on the deadline I set in my last comment: it has passed with no objection, so I'm treating the design question as settled on **Option A — `AsyncModel` stays a separate class**, which is what the current diff implements. If a reviewer prefers Option B (folding the sync-`predict` offload into `Model` behind an opt-in flag) I will still gladly switch; I just did not want an unanswered either/or to keep blocking the PR indefinitely.

Status after the rebase: **CI is fully green — 46 passed, 0 failed**, which also settles the earlier red run. Those failures were infrastructure, as suspected: `build (3.11)` had failed while 3.10 and 3.12 passed on the identical commit, the `test-kourier` lane failed while all 13 sibling e2e lanes passed, and `enforce-all-checks` has since been removed from master entirely.

@jinchihe @yuzisun — this PR only touches files under `python/kserve/`, and you are the approvers listed in `python/kserve/OWNERS`, so I think it needs your review rather than the root-level list I had been pinging. It is three files: a new `async_model.py`, a one-line export, and tests (+272/-0, no existing lines modified).

Why it matters: a synchronous `predict()` runs inline on the event loop, so one slow sync model blocks the entire server — including the readiness path. The probe keeps reporting healthy while the server can no longer accept work, so an orchestrator sees a healthy pod that serves nothing. This adds an opt-in offload path for sync models, with no behavior change for anything that exists today.

Happy to rebase again or split it further if that would help it move.

anencore94:
Hi @jinchihe @yuzisun — following up on my September 7 review request. The current branch has no merge conflicts, and the reported CI checks are passing.

Could you review the opt-in sync-predict offload, particularly whether the separate `AsyncModel` class is acceptable? That is the current implementation; if you prefer an opt-in flag on `Model`, please let me know so I can adapt it. An explicit direction on that point would help move this forward.

If someone else owns this part of the Python SDK, a reviewer referral would also help. Thank you!

PR #5022, opened 2026-02-02
fix: llmisvc ClusterRoleBinding to point to new ClusterRole

pierDipi:
ClusterRole is called `kserve-llmisvc-manager-role`


Jooho (approved):
/lgtm
/approve

PR #5025, opened 2026-02-02
fix: normalize deprec. deploymentMode annos; allow deletion

jlost:

Fixes InferenceService stuck in Terminating state when the `deploymentMode` annotation uses a deprecated value (`RawDeployment` or `Serverless`).

**Root Cause:**

The controller normalizes deprecated annotation values when setting `status.DeploymentMode`:
- `RawDeployment` -> `Standard`
- `Serverless` -> `Knative`

However, the validation webhook compares the raw annotation value against the status, causing a mismatch that blocks all updates - including finalizer removal during deletion.

**This fix:**

1. **Defaulting webhook**: Normalizes deprecated annotation values at admission to prevent future mismatches
2. **Validation webhook**: Skips `deploymentMode` validation during deletion (when `DeletionTimestamp` is set) to allow finalizer cleanup for existing resources

**Type of changes**

- [x] Unit tests added for defaulting webhook normalization
- [x] Unit tests added for validation bypass during deletion

**Checklist**:

```release-note
Fix InferenceService stuck in Terminating when using deprecated deploymentMode annotation values (RawDeployment/Serverless)
```


spolti (approved):
Nice, thanks!

/lgtm

spolti:
@Jooho @sivanantha321 hey, can you please take a look?

jlost:
Note that there is an existing PR addressing the same problem: https://github.com/kserve/kserve/pull/4800/changes
The basic difference between that PR and this one is that #4800 makes validation accept RawDeployment/Serverless as if they were Standard/Knative, while this one actually updates them to Standard/Knative.
This one cleans up deprecated values, #4800 handles the deprecated values so they can stay around forever.
Thoughts @terrytangyuan ?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -88,9 +88,11 @@ func (v *InferenceServiceValidator) ValidateUpdate(ctx context.Context, oldObj,
 		validatorLogger.Error(err, "Unable to convert object to InferenceService")
 	}
 	validatorLogger.Info("validate update", "name", isvc.Name)
-	err = validateDeploymentMode(isvc, oldIsvc)
-	if err != nil {
-		return nil, err
+	if isvc.DeletionTimestamp == nil {
```
bartoszmajsak: nit, but most of the times I see `obj.GetDeletionTimestamp().IsZero()`

jlost: Addressed

Jooho (approved):
/lgtm
/approved

PR #5033, opened 2026-02-03
feat: add LocalModelCacheDeployment CRD

### LINKED ISSUE #5030: LocalModelCache versioning for model updates
/kind feature

Add a `Version` field to `LocalModelCache` to support multiple versions of the same model coexisting in the cluster.

Currently, only one `LocalModelCache` can exist per `StorageURI`. This makes it impossible to update cached models without downtime:
1. Delete existing `LocalModelCache`
2. Create new `LocalModelCache` with updated model
3. Recreate `InferenceServices`

With versioning support:
1. Create new `LocalModelCache` with same `StorageURI` but higher version (e.g., v2)
2. While waiting for the model to be downloaded, v1 remains active. If needed, v2 LocalModelCache can be 
tested with a separate InferenceService without production traffic. This means LocalModelCache v1 and v2 
coexist simultaneously.
3. Rollout to the new version of LocalModelCache by updating the InferenceService
4. Delete old version when no longer needed

bluayer:

LocalModelCacheDeployment CRD manages LocalModelCache resources as revisions, enabling cached model updates with minimal downtime.

Currently, updating a cached model requires deleting the existing LocalModelCache and recreating it, causing downtime. 

With this PR:
1. Create LocalModelCacheDeployment with model spec
2. Controller creates LocalModelCache revision (e.g., my-model-v1)
3. Update LocalModelCacheDeployment spec → new revision created (my-model-v2)
4. Both versions coexist; You can pin to specific revision via annotation at isvc
5. (Optional and Manually, Not Automatically in this PR) Delete old revision when no longer needed

Key changes:
- Added LocalModelCacheDeployment CRD type and controller
- Controller creates LocalModelCache resources named {name}-v{generation}
- InferenceService webhook resolves deployment to specific revision when serving.kserve.io/localmodelcachedeployment label is present
- Existing LocalModelCache users are unaffected (backward compatible)

Fixes #5030 

- [x] Unit tests for LocalModelCacheDeployment controller
- [x] E2E test on EKS cluster

Tested scenarios:
1. Create LocalModelCacheDeployment → LocalModelCache revision created
2. Update spec → new revision created, old revision preserved
3. InferenceService with deployment label → correct PVC mounted
4. Update InferenceService revision annotation → switches to different revision
5. Delete deployment → all revisions cascade deleted

- Logs
```
$ kubectl get localmodelcache
NAME                 AGE
model-deployment-v1   35s
model-deployment-v2   12s

$ kubectl get isvc model-isvc -o jsonpath='{.metadata.annotations}'
{
  "internal.serving.kserve.io/localmodel-pvc-name": "model-deployment-v1-test-nodes",
  ...
}
```

This takes a different approach than the original issue proposal (adding Version field to LocalModelCache). 

The original approach has some concerns:
- Adding Version to spec is inflexible - users must manually manage version numbers and label names
- Version conflicts can occur when multiple users create caches with the same version
- Extending or managing LocalModelCache becomes more complex with version logic embedded in the resource

Instead, it introduces a parent resource (LocalModelCacheDeployment) that manages versioned LocalModelCache resources. This approach follows Kubernetes Deployment/ReplicaSet pattern.

**Checklist**:

**Release note**:
```release-note
Added LocalModelCacheDeployment CRD for managing LocalModelCache revisions, enabling cached model updates with minimal downtime
```


bluayer:
Hi, @spolti. I know you're busy, so I apologize for the mention. When you have a moment, could you let me know if there's anything else I should address to help move the PR review forward?

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	existingLmc := &v1alpha1.LocalModelCache{}
+	err := r.Get(ctx, client.ObjectKey{Name: lmcName}, existingLmc)
+
+	if errors.IsNotFound(err) {
+		// Create new LocalModelCache
+		lmc := &v1alpha1.LocalModelCache{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: lmcName,
+				Labels: map[string]string{
+					"serving.kserve.io/localmodelcachedeployment": localModelDeployment.Name,
```
spolti: Please, use constants for these labels.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	if err := r.List(ctx, lmcList, client.MatchingLabels{
+		"serving.kserve.io/localmodelcachedeployment": localModelDeployment.Name,
+	}); err != nil {
+		return
+	}
+
+	revisions := []v1alpha1.LocalModelCacheDeploymentRevision{}
+	for _, lmc := range lmcList.Items {
+		var revNum int32
+		if revLabel, ok := lmc.Labels["serving.kserve.io/revision"]; ok {
```
spolti: constant

cmd/localmodel/main.go
```diff
@@ -151,6 +151,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Setup LocalModelCacheDeployment controller
+	setupLog.Info("Setting up v1alpha1 LocalModelCacheDeployment controller")
```
spolti, line 155: can't we have both contollers in the localModel controller?

config/rbac/localmodel/role.yaml
```diff
@@ -49,6 +49,7 @@ rules:
 - apiGroups:
   - serving.kserve.io
   resources:
+  - localmodelcachedeployments
```
spolti, line 52: Wondering if we need a new type; we could make this the default behavior as it seems more reliable.

pkg/apis/serving/v1alpha1/local_model_cache_deployment_types.go
```diff
@@ -0,0 +1,85 @@
...
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LocalModelCacheDeploymentSpec defines the desired state of LocalModelCacheDeployment
+type LocalModelCacheDeploymentSpec struct {
```
spolti, line 25: How this will work if user has both CRs deployed (LocalModel and this proposed one?

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	localModelDeployment := &v1alpha1.LocalModelCacheDeployment{}
+	if err := r.Get(ctx, req.NamespacedName, localModelDeployment); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Calculate revision number from generation
+	revision := localModelDeployment.Generation
```
spolti: Can you please confirm what it returns? Is it something like {name}-v1?
If so, can we have a more unique name, like a hash or something? It potentially could collide with orphaned LocalModelCache.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
spolti: please use constants.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		// Create new LocalModelCache
+		lmc := &v1alpha1.LocalModelCache{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: lmcName,
+				Labels: map[string]string{
+					"serving.kserve.io/localmodelcachedeployment": localModelDeployment.Name,
+					"serving.kserve.io/revision":                  fmt.Sprintf("%d", revision),
+				},
+				OwnerReferences: []metav1.OwnerReference{
+					*metav1.NewControllerRef(localModelDeployment, v1alpha1.SchemeGroupVersion.WithKind("LocalModelCacheDeployment")),
```
spolti: this also should be moved to a constant

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *LocalModelCacheDeploymentReconciler) updateRevisionList(ctx context.Context, localModelDeployment *v1alpha1.LocalModelCacheDeployment) {
+	// List all LocalModelCaches owned by this LocalModelCacheDeployment
+	lmcList := &v1alpha1.LocalModelCacheList{}
+	if err := r.List(ctx, lmcList, client.MatchingLabels{
+		"serving.kserve.io/localmodelcachedeployment": localModelDeployment.Name,
+	}); err != nil {
+		return
```
spolti: is it fine to ignore this error?

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *LocalModelCacheDeploymentReconciler) updateRevisionList(ctx context.Context, localModelDeployment *v1alpha1.LocalModelCacheDeployment) {
+	// List all LocalModelCaches owned by this LocalModelCacheDeployment
+	lmcList := &v1alpha1.LocalModelCacheList{}
+	if err := r.List(ctx, lmcList, client.MatchingLabels{
+		"serving.kserve.io/localmodelcachedeployment": localModelDeployment.Name,
```
spolti: constant.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -147,14 +148,23 @@ func (d *InferenceServiceDefaulter) Default(ctx context.Context, obj runtime.Obj
 			mutatorLogger.Error(err, "Cannot List local models")
 			return err
 		}
+
+		// Check if InferenceService references a LocalModelCacheDeployment
+		if deploymentName, ok := isvc.Labels[constants.LocalModelCacheDeploymentLabel]; ok {
```
spolti, line 153: constant.

spolti, line 153: If the user sets an invalid value for the LocalModelCacheDeployment label,  the webhook silently sets it to nil and continues. I guess we should make sure at least some feedback is returned to the user to avoid misconfiguration.

pkg/apis/serving/v1beta1/inference_service_defaults_test.go
```diff
@@ -343,7 +343,7 @@ func TestInferenceServiceDefaults(t *testing.T) {
 	for _, scenario := range scenarios {
 		resources := corev1.ResourceRequirements{Requests: defaultResource, Limits: defaultResource}
 		scenario.isvc.Spec.DeepCopy()
-		scenario.isvc.DefaultInferenceService(scenario.config, scenario.deployConfig, nil, nil)
+		scenario.isvc.DefaultInferenceService(scenario.config, scenario.deployConfig, nil, nil, nil)
```
spolti, line 346: PLease add new tests to cover the new feature.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
spolti: please use constant.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	}); err != nil {
+		return
+	}
+
+	revisions := []v1alpha1.LocalModelCacheDeploymentRevision{}
+	for _, lmc := range lmcList.Items {
+		var revNum int32
+		if revLabel, ok := lmc.Labels["serving.kserve.io/revision"]; ok {
+			fmt.Sscanf(revLabel, "%d", &revNum)
```
spolti: The return value is ignored. Malformed labels silently default revNum to 0, producing a confusing status.

pkg/apis/serving/v1alpha1/local_model_cache_deployment_types.go
```diff
@@ -0,0 +1,85 @@
...
+type LocalModelCacheDeploymentStatus struct {
+	// Current active revision name
+	// +optional
+	CurrentRevision string `json:"currentRevision,omitempty"`
+	// ObservedGeneration is the last observed generation
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// List of all revisions
+	// +optional
+	Revisions []LocalModelCacheDeploymentRevision `json:"revisions,omitempty"`
```
spolti, line 59: Shouldn't we have the RevisionHistoryLimit to control how long it can grow?

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
```diff
@@ -0,0 +1,138 @@
...
+				NodeGroups:     localModelDeployment.Spec.NodeGroups,
+			},
+		}
+
+		if err := r.Create(ctx, lmc); err != nil {
+			log.Error(err, "Failed to create LocalModelCache", "name", lmcName)
+			return ctrl.Result{}, err
+		}
+		log.Info("Created LocalModelCache", "name", lmcName)
+		existingLmc = lmc
```
spolti: nit, assigned but never used.

pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -473,12 +483,80 @@ func deleteLocalModelMetadata(isvc *InferenceService) {
 	if isvc.Annotations != nil {
 		delete(isvc.Annotations, constants.LocalModelSourceUriAnnotationKey)
 		delete(isvc.Annotations, constants.LocalModelPVCNameAnnotationKey)
+		delete(isvc.Annotations, constants.LocalModelRevisionAnnotationKey)
 	}
 }
 
+// setLocalModelCacheFromDeployment resolves LocalModelCacheDeployment to LocalModelCache
+// Uses existing revision annotation if present, otherwise uses currentRevision from status
+func (isvc *InferenceService) setLocalModelCacheFromDeployment(localModel *v1alpha1.LocalModelCacheDeployment, localModelCaches *v1alpha1.LocalModelCacheList) {
```
spolti, line 504: nit: `setLocalModelCacheDeploymentLabel` name implies it only handles deployments, but it also contains legacy `LocalModelCache` fallback logic. `resolveLocalModel` would be more accurate.

spolti (changes requested):

pkg/apis/serving/v1beta1/inference_service_defaults.go
bluayer: The constant is already in use

config/rbac/localmodel/role.yaml
bluayer: Thank you for all of the comments. As a user, what I really want is an easier way to manage LocalModelCache — especially when updating/changing cached models in production. The current LocalModelCache represents a single model snapshot: SourceModelUri is immutable, and the webhook enforces unique StorageURI. So updating a cached model requires manual orchestration:

1. Create a new LocalModelCache with the new sourceUri and name
2. Wait for download to complete on all nodes
3. Switch ISVCs to the new LocalModelCache
4. Delete the old LocalModelCache

I introduced LocalModelCacheDeployment to automate this workflow and provide version management. It has a different responsibility from LocalModelCache:

- **LocalModelCache**: represents a single cached model on nodes — users create it, and the system handles downloading the model to the specified node groups.
- **LocalModelCacheDeployment**: manages version transitions — when users update the model spec, the system automatically creates a new LocalModelCache revision while keeping the old one serving, then handles cleanup.

If you have a different approach in mind to make LocalModelCache management easier, I'd be happy to discuss.

bluayer: Btw, I really appreciate your time and detailed feedback for reviewing this PR. I'm not yet familiar with Go and KServe conventions, so I apologize for any inconvenience during the review.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
bluayer: It's not fine. I fixed it to return an error.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
bluayer: Changed to hash-based naming ({name}-{specHash}). Same spec 
produces the same hash, which also prevents duplicate revisions.

pkg/apis/serving/v1alpha1/local_model_cache_deployment_types.go
bluayer: Added RevisionHistoryLimit field to the spec (default: 10). The 
controller cleans up old revisions exceeding the limit, protecting 
the current revision from deletion.

pkg/apis/serving/v1beta1/inference_service_defaults.go
bluayer: I agree.  Changed to return err.

pkg/controller/v1alpha1/localmodel/localmodelcachedeployment_controller.go
bluayer: Fixed. Now checks the return value and returns an error if the 
revision label is malformed

PR #5034, opened 2026-02-03
feat(llmisvc): watch pods for init container failures

jlost:

Adds a watch on pods with LLMInferenceService labels to detect init container failures quickly. This enables faster error propagation when storage-initializer or other init containers fail, improving the user experience when deployments have configuration issues.

**Related PR**: This applies the same pod watch pattern from #4910 (InferenceService) to LLMInferenceService.

Fixes #

- [x] Unit tests for PodInitContainersFunc mapper
- [x] Unit tests for PodInitContainersPredicate (UpdateFunc, CreateFunc, DeleteFunc, GenericFunc)
- [x] E2E test: verifies init container failure triggers fast reconciliation
- [x] E2E test: verifies failure in one LLMISVC doesn't affect unrelated LLMISVCs

The implementation filters pods at three levels:
1. **Cache level** (`cmd/llmisvc/main.go`): Only pods with `app.kubernetes.io/part-of: llminferenceservice` are cached
2. **Predicate level**: Only UpdateFunc passes through, and only when InitContainerStatuses changed
3. **Map function level**: Double-checks labels before enqueuing reconcile requests

This ensures we don't cache all cluster pods or trigger unnecessary reconciliations.

**Checklist**:

```release-note
Add pod watch for LLMInferenceService to enable faster init container failure detection
```


pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -524,3 +539,58 @@ func (r *LLMISVCReconciler) enqueueOnConfigMapChange(logger logr.Logger) handler
...
+func (r *LLMISVCReconciler) PodInitContainersFunc(ctx context.Context, obj client.Object) []reconcile.Request {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok || pod == nil {
+		return nil
+	}
+	// Check if pod belongs to an LLMInferenceService by verifying the part-of label
+	partOf := pod.Labels[PodLabelPartOf]
+	if partOf != PodLabelPartOfValue {
+		return nil
+	}
```
bartoszmajsak: I think this is already covered by the cache https://github.com/kserve/kserve/blob/47885225a6c854cc2308fc5490bcdba6521667e8/cmd/llmisvc/main.go#L150-L154

jlost: It's also defensively repeated [in the filter predicate](https://github.com/kserve/kserve/pull/5034/changes/BASE..b159654b5525ae8db74232e88a110caa30041974#diff-a8b8866369d9b70375c024df95597d68465558e8e52f359e3e7510cf55a3fde4R578-R580).  
Technically the filter predicate's check isn't strictly necessary either. Or you could argue that the filter logic should be correct independently of the cache...
Would you like me to pare this down?

jlost: I removed the redundant filtering and moved the .Watches to the top of the function.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -65,6 +66,15 @@ var childResourcesPredicate, _ = predicate.LabelSelectorPredicate(metav1.LabelSe
 	},
 })
 
+const (
+	// PodLabelName is the label key used to identify pods by the name of their owning LLMInferenceService
+	PodLabelName = "app.kubernetes.io/name"
```
spolti: Wondering if we already have another label or annotations with these same values.

jlost: I moved them to a shared constants file, following the same pattern as the isvc controller. Since I followed through completely, there are quite a few more changes, but they are surface-level and low risk. Take a look -- I can back out the commit if we prefer it the other way.

pkg/controller/v1alpha2/llmisvc/pod_watch_test.go
bartoszmajsak, line 1: I wonder if we can't capture most of those cases in tests that run against real cluster. Not really sure we gain much for such low level tests. WDYT?

bartoszmajsak (commented):
It's great that we are bubbling up errors to the owning resource. What are your thoughts about other failure states of a pod? Should we consider watching those too as a follow-up?

pkg/controller/v1alpha2/llmisvc/pod_watch_test.go
jlost: re: watching other failure states -- Good question! 
The basic behavior (without this patch) is to watch the Deployment. This works fine for most pod failures because deployment.status elements like readyReplicas are responsive to the pod states. It _doesn't_ work well for init containers because nothing on the Deployment updates until the progressDeadlineSeconds (default: 600) has expired, and there is high utility in learning about those errors early. (In fact, it looks like a LeaderWorkerSet will **never** update, which would be even more problematic -- but I have not specifically repro'd that case)
In short, if we identified a reason to watch other fields we could easily do so, but watching the higher-level resource (deployment/lws) is actually preferable because they are less noisy in terms of updates (triggering unnecessary reconciliations) and more "architecturally expedient" (fewer skipped layers).

re: removing unit tests -- Yes, I removed a few that had very low utility. I would be inclined to keep the rest because watch logic is mostly hidden from users, and while much of it is also covered by [e2e tests](test/e2e/llmisvc/test_pod_watch.py), those are more expensive to set up and run.

spolti (approved):
/lgtm

spolti:
@Jooho @sivanantha321 fyi

pkg/constants/constants.go
```diff
@@ -384,6 +384,24 @@ const (
 	InferenceServiceLabel       = "serving.kserve.io/inferenceservice"
 )
 
+// LLMInferenceService label constants (Kubernetes recommended labels: app.kubernetes.io/*)
+const (
+	LLMInferenceServicePodNameLabelKey   = "app.kubernetes.io/name"
```
Jooho: Since app.kubernetes.io/name is a Kubernetes recommended label and not specific to LLMInferenceService, would it make sense to keep it as a generic shared constant instead of mapping it to LLMInferenceServicePodNameLabelKey?

Jooho: There are 2 more

jlost: Good point - updated accordingly

Jooho (approved):
/lgtm
/approved

PR #5041, opened 2026-02-05
[WIP] feat: detect Gateway for auto-migration to v1 InferencePool

pierDipi:
This helps when Gateway(s) or HTTPRoute(s) have other policies attached to them resulting in multiple parent refs and controllers.

Resovles a TODO.

Related to #5007

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router.go
+++ b/pkg/controller/v1alpha2/llmisvc/router.go
@@ -211,8 +211,8 @@ func (r *LLMISVCReconciler) expectedHTTPRoute(ctx context.Context, llmSvc *v1alp
 	if routeExists {
 		migrationValue, hasMigrationAnnotation := curr.Annotations[AnnotationInferencePoolMigrated]
 		isMigrated = hasMigrationAnnotation && migrationValue == v1MigrationValue
-		infPoolV1Alpha2Support = IsInferencePoolV1Alpha2Supported(curr)
-		infPoolV1Support = IsInferencePoolV1Supported(curr)
+		infPoolV1Alpha2Support = IsInferencePoolV1Alpha2Supported(ctx, r.Client, curr)
+		infPoolV1Support = IsInferencePoolV1Supported(ctx, r.Client, curr)
 	}
 
 	// Switch to v1 if:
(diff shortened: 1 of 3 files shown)


KillianGolds:
LGTM overall! The controller name matching approach is clean and the test coverage for multi-parent scenarios with policy controllers looks solid 👍

One question, I was tracing through the fallback logic and wanted to check my understanding:

If `DiscoverGateways` fails or returns empty [L435-438], `gatewayControllerName` would stay as `""` [L463] and wouldn't match any parent's `ControllerName` [L475]. In that case, we'd fall back to the first Gateway parent with `ResolvedRefs` [L457-461]

Could there be a scenario where a policy controller (like Kuadrant) sets `ResolvedRefs=True` on its parent status, appears before the Gateway controller parent in the list, and gets selected as the fallback? e.g.:

1. Kuadrant parent: `ResolvedRefs=True` (first in list)
2. Istio parent: `ResolvedRefs=False/InvalidKind`

With working gateway discovery → correctly returns `False`
With failed gateway discovery → would return `True`?

I might be misreading the code though, happy to be corrected if I'm missing something!

Jooho (approved):
/lgtm
/approved

pierDipi:
@Jooho why? it's WIP

pierDipi:
@Jooho we need to revert https://github.com/kserve/kserve/pull/5316

romilbhardwaj:
@pierDipi thanks for the work on this. I see that #5316 reverted this PR before v0.18 cut and the auto-migration logic in v0.18 still falls back to the v1alpha2 `inference.networking.x-k8s.io` group on the first reconcile of a fresh LLMInferenceService.

The route only switches to v1 on a subsequent reconcile after the Gateway controller (kgateway in our case) stamps ResolvedRefs=False with InvalidKind on the backendRef.

This works on healthy clusters but takes minutes on some of our backed-up Gateway controllers, leaving fresh endpoints returning BackendNotFound in the meantime. Which we're currently working around with a post-apply HTTPRoute patch.

Is there a follow-up planned to this PR, so the controller writes the v1 group on the first reconcile? Happy to help test if useful. Thanks!

bartoszmajsak:
@romilbhardwaj Thanks for your feedback! I am afraid this got buried with the merge. Would you mind opening an issue and tag us there if that's still relevant? We can definitely resume work to improve this. I am also curious about the details how you worked it around.

PR #5045, opened 2026-02-09
fix(build): remove GOPATH dependency from OpenAPI script

bartoszmajsak:
The `update-openapigen.sh` script was dependent on the legacy GOPATH workspace structure. This caused failures in modern Go module environments where projects can be located anywhere on the filesystem.

This commit removes the rigid GOPATH dependency and refactors the script to dynamically locate the project root. This ensures that the OpenAPI generation runs reliably, independent of the project's location or the directory from which it is executed.

Example failure (different project path than assumed GOPATH):

```
✦ ❯ make precommit
hack/update-openapigen.sh
You are not in the target directory: /home/bartek/go/src/github.com/kserve/kserve
Creating a symbolic link for the target directory ...
ln: failed to create symbolic link '/home/bartek/go/src/github.com/kserve/kserve': File exists
make: *** [Makefile:153: generate] Error 1
```

### DIFF
--- a/hack/update-openapigen.sh
+++ b/hack/update-openapigen.sh
@@ -18,36 +18,36 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+# find_project_root [start_dir] [marker]
+#   start_dir : directory to begin the search (defaults to this script’s dir)
+#   marker    : filename or directory name to look for (defaults to "go.mod")
+#
+# Prints the first dir containing the marker, or exits 1 if none found.
+find_project_root() {
+  local start_dir="${1:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
+  local marker="${2:-go.mod}"
+  local dir="$start_dir"
+
+  while [[ "$dir" != "/" && ! -e "$dir/$marker" ]]; do
+    dir="$(dirname "$dir")"
+  done
+
+  if [[ -e "$dir/$marker" ]]; then
+    printf '%s\n' "$dir"
+  else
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
> [PR Style Check / Title (pull_request_target)](https://github.com/kserve/kserve/actions/runs/21817618609/job/62942710702?pr=5045)

~Complains about the PR title which I cannot edit myself. Can someone help?~

**EDIT**: It's possible by using `gh pr edit` CLI

bartoszmajsak:
Precommit check failing b/c of the bug fixed in #5046

hack/update-openapigen.sh
```diff
@@ -18,36 +18,36 @@ set -o errexit
...
 
-GOPATH=$(go env GOPATH)
-if [[ -z $GOPATH ]]
-then
-    echo "Error: GOPATH is not set. Please configure your GOPATH environment variable."
-    exit 1
-fi    
-TARGET_DIR="$GOPATH/src/github.com/kserve/kserve"
-CURRENT_DIR=$(pwd)
+  while [[ "$dir" != "/" && ! -e "$dir/$marker" ]]; do
```
israel-hdez, line 31: I have used this short command:

```sh
git rev-parse --show-toplevel
```

Although you need `git` command, and the sources need to be properly initialized repo.
It looks safer to me, since the script won't traverse to the very root of the FS... so, I wonder if you want compatibility with bare sources?

Current solution is fine. Thus, see this more like a nit.

israel-hdez:
Despite my comment, this looks good to me.
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5046, opened 2026-02-09
fix(build): enforce local yq installation

bartoszmajsak:
The `install-yq.sh` script used `command -v yq` to check for an existing installation, which could find a system-wide yq and skip the local download. This left `bin/yq` missing, causing `make manifests` (and `make precommit`) to fail with "No such file or directory".

The version check now targets `${BIN_DIR}/yq` exclusively, ensuring the binary is always placed in the project-local bin directory. The Makefile target is also restructured to guard against redundant downloads and properly propagate errors.

It's also present in CI https://github.com/kserve/kserve/actions/runs/21817619326/job/62942712668?pr=5045#step:6:16

### DIFF
--- a/Makefile.tools.mk
+++ b/Makefile.tools.mk
@@ -37,9 +37,11 @@ $(ENVTEST): $(LOCALBIN)
 ## Download yq locally if necessary.
 .PHONY: yq
 yq: $(YQ)
-$(YQ): $(LOCALBIN)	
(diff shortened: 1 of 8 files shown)


spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5063, opened 2026-02-12
feat(llmisvc): add startupProbe to vLLM containers

### LINKED ISSUE #5062: Add startupProbe to vLLM main containers and remove initialDelaySeconds from liveness/readiness probes
/kind feature

Replace the `initialDelaySeconds` on liveness and readiness probes with a Kubernetes `startupProbe` on all main vLLM containers in LLMInferenceService config templates.

Currently, vLLM main containers use `initialDelaySeconds` on liveness probes (120s for standard templates, 300s for data-parallel) to account for model loading time. This causes two problems:

1. **Small models** that start in under 120s are blocked from becoming ready, adding unnecessary delay before serving traffic.
2. **Large models** that need more than 120s (or 300s) get killed and restarted by the liveness probe before they finish loading.

A `startupProbe` solves both issues by gating liveness and readiness probes until the model is actually running:

- Models get up to **10 minutes** to start (`failureThreshold: 60` x `periodSeconds: 10`)
- Fast models proceed **immediately** once `/health` returns success

The proposed changes are:

1. Add `startupProbe` (`httpGet /health`, `failureThreshold: 60`, `periodSeconds: 10`) to every main vLLM container across all 6 config templates and their corresponding Helm chart templates
2. Set `initialDelaySeconds: 0` on all main container liveness probes (currently 120 or 300)
3. Set `initialDelaySeconds: 0` on all main container readiness probes (currently 10 or 200)
4. No changes to routing sidecar probes or worker container probes
5. Update unit tests to validate the new probe configuration

- The `startupProbe` has been GA since Kubernetes 1.20, so there are no compatibility concerns.
- Sidecar probes (e.g. `llm-d-routing-sidecar` in decode templates) should remain unchanged - they use `initialDelaySeconds: 10` and don't need a startup probe.
- Both config templates (`config/llmisvcconfig/`) and Helm chart templates (`charts/kserve-llmisvc-resources/templates/`) need to be updated and kept in sync.

N/A - this is a straightforward configuration change with no architectural implications.

KillianGolds:

Adds a Kubernetes `startupProbe` to all main vLLM containers in the LLMInferenceService config templates, and removes the now-redundant `initialDelaySeconds` from liveness and readiness probes.

**Problem:** The current liveness probes use `initialDelaySeconds: 120` (standard templates) or `initialDelaySeconds: 300` (data-parallel templates) as a workaround for slow model loading. This has two issues:
- Small models that start in seconds still wait the full initial delay before becoming ready
- Large models that need longer than the initial delay get killed and enter CrashLoopBackOff

**Solution:** A `startupProbe` with `failureThreshold: 60` and `periodSeconds: 10` provides up to 10 minutes for model startup. Once the startup probe succeeds, liveness and readiness probes take over immediately. This lets fast models become ready in seconds while giving large models ample time to load.

Changes:
- Add `startupProbe` (httpGet `/health`, `failureThreshold: 60`, `periodSeconds: 10`) to main containers in all 6 config templates
- Set `initialDelaySeconds: 0` on liveness probes (was 120 or 300)
- Set `initialDelaySeconds: 0` on readiness probes (was 10 or 200)
- Mirror all changes in the corresponding 6 Helm chart templates
- Sidecar and worker container probes are unchanged

Fixes #5062

- [x] Unit tests - `go test ./pkg/controller/v1alpha2/llmisvc/...` (101 specs pass)
- [x] `make precommit` passes cleanly
- [x] Kind cluster validation - deployed controller to a local Kind cluster, created a test `LLMInferenceService` CR, and inspected the generated pod spec to confirm:
  - `startupProbe` present on main container with correct values
  - `livenessProbe.initialDelaySeconds` is `0`
  - `readinessProbe.initialDelaySeconds` is `0`
  - Sidecar probes unchanged

- Logs

```
kubectl get pod -n test-llmisvc -o yaml | grep -A 15 startupProbe
        startupProbe:
          failureThreshold: 60
          httpGet:
            path: /health
            port: 8000
            scheme: HTTP
          periodSeconds: 10
```
### Behavioural validation (mock `/health` on Kind cluster)

Deployed pods with a Python HTTP server that returns `503` for a configurable startup delay, then `200` - simulating vLLM model loading. Compared old probe config (`initialDelaySeconds: 120`, no `startupProbe`) against new probe config (`startupProbe`, `initialDelaySeconds: 0`).

<details>
<summary>Small model (5s startup delay)</summary>

| Pod | Config | Time to Ready | Restarts |
|-----|--------|--------------|----------|
| `new-probes-small` | startupProbe, initialDelay: 0 | **11s** | 0 |
| `old-probes-small` | no startupProbe, initialDelay: 120 | 18s | 0 |

Both pods survive. The new config becomes Ready slightly faster since readiness probing begins immediately after the startup probe succeeds, rather than waiting for `initialDelaySeconds`.

</details>

<details>
<summary>Large model (150s startup delay) - the critical scenario</summary>

| Pod | Config | Time to Ready | Restarts |
|-----|--------|--------------|----------|
| `new-probes-large` | startupProbe, initialDelay: 0 | **160s** | **0** |
| `old-probes-large` | no startupProbe, initialDelay: 120 | 171s | **1** |

The old config's liveness probe starts at `t=120s`, fails 3 consecutive checks (`t=120s`, `t=130s`, `t=140s`), and **kills the container at ~150s** - right before the model would have been ready. The container restarts and must reload the model from scratch.

The new config's startup probe keeps polling every 10s until `/health` returns `200` at ~150s, then hands off to liveness/readiness - zero restarts, no wasted work.

```
# old-probes-large events
Warning  Unhealthy  spec.containers{main}: Liveness probe failed: HTTP probe failed with statuscode: 503
Normal   Killing    spec.containers{main}: Container main failed liveness probe, will be restarted
```

```
# new-probes-large events (startup probe absorbs the 503s gracefully)
Warning  Unhealthy  spec.containers{main}: Startup probe failed: HTTP probe failed with statuscode: 503
```

</details>

- Routing sidecar probes (in decode templates) are intentionally **not** changed - they have short `initialDelaySeconds: 10` and don't face the same slow-startup issue
- Worker containers have no probes today and are also unchanged
- The test file covers 2 of 8 configs with full `DeepDerivative` assertions (single-node template and decode-worker-data-parallel - the simplest and most complex cases). All 8 configs are structurally validated (valid YAML, correct Kind/APIVersion, registered in `WellKnownDefaultConfigs`, successful `ReplaceVariables` rendering)
- Config templates and Helm chart templates are kept in sync

**Checklist**:

**Release note**:
```release-note
Add startupProbe to vLLM main containers in LLMInferenceService templates, replacing initialDelaySeconds-based startup handling. Models now become ready as soon as they finish loading (up to 10 minutes) instead of waiting a fixed delay.
```

### DIFF
--- a/charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
@@ -48,7 +48,6 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 10
           failureThreshold: 3
@@ -57,7 +56,6 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 5
           failureThreshold: 60
--- a/charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
@@ -48,7 +48,7 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 120
+          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 10
           failureThreshold: 3
@@ -57,10 +57,17 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 10
+          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 5
           failureThreshold: 60
+        startupProbe:
+          httpGet:
+            path: /health
+            port: 8001
+            scheme: HTTP
+          failureThreshold: 60
+          periodSeconds: 10
         volumeMounts:
           - mountPath: /home
             name: home
--- a/charts/kserve-llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
@@ -244,7 +244,7 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 300
+          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 10
           failureThreshold: 3
@@ -253,10 +253,17 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 200
+          initialDelaySeconds: 0
           periodSeconds: 30
           timeoutSeconds: 5
           failureThreshold: 60
+        startupProbe:
+          httpGet:
+            path: /health
+            port: 8001
+            scheme: HTTP
+          failureThreshold: 60
+          periodSeconds: 10
         volumeMounts:
           - mountPath: /home
             name: home
--- a/charts/kserve-llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
@@ -244,7 +244,6 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 10
           failureThreshold: 3
@@ -253,7 +252,6 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 0
           periodSeconds: 30
           timeoutSeconds: 5
           failureThreshold: 60
--- a/charts/kserve-llmisvc-resources/templates/config-llm-prefill-template.yaml
+++ b/charts/kserve-llmisvc-resources/templates/config-llm-prefill-template.yaml
@@ -49,7 +49,7 @@ spec:
               path: /health
               port: 8000
               scheme: HTTP
-            initialDelaySeconds: 120
+            initialDelaySeconds: 0
             periodSeconds: 10
             timeoutSeconds: 10
             failureThreshold: 3
@@ -58,10 +58,17 @@ spec:
               path: /health
               port: 8000
               scheme: HTTP
-            initialDelaySeconds: 10
+            initialDelaySeconds: 0
             periodSeconds: 10
             timeoutSeconds: 5
             failureThreshold: 60
+          startupProbe:
+            httpGet:
+              path: /health
+              port: 8000
+              scheme: HTTP
+            failureThreshold: 60
+            periodSeconds: 10
           volumeMounts:
             - mountPath: /home
               name: home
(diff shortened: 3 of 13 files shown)


charts/kserve-llmisvc-resources/templates/config-llm-decode-worker-data-parallel.yaml
```diff
@@ -242,7 +242,7 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 300
+          initialDelaySeconds: 0
```
Jooho: Wasn’t this originally set to a longer value because of the model loading time?

Is there a specific reason we’re trying to change it to 0?

Since the default is already 0, if the intention is to use 0, it might be cleaner to just remove the field entirely rather than explicitly setting it to 0.

Jooho: Since the default is already 0, if the intention is to use 0, it might be cleaner to just remove the field entirely rather than explicitly setting it to 0.

KillianGolds: Thanks for the feedback and Yes, the original initialDelaySeconds: 120 (and 300 for data-parallel) was a workaround to give the model time to load before liveness/readiness probes started checking. The problem is that it's a fixed delay - small models that start in seconds still wait the full 120s/300s, and large models that need longer get killed when the delay expires.

The startupProbe replaces that workaround: it gates liveness and readiness until /health returns 200, so probes only start after the model is actually ready. That's why initialDelaySeconds on liveness/readiness is no longer needed.

Good call on removing the field entirely rather than explicitly setting it to 0! Was an oversight on my part I'll update to just drop the initialDelaySeconds lines since 0 is the default.

KillianGolds: Addressed in: [b11556f](https://github.com/kserve/kserve/pull/5063/commits/b11556f3cf70442f225780da1e96eb99864b6e77)

KillianGolds: Addressed in: [b11556f](https://github.com/kserve/kserve/pull/5063/commits/b11556f3cf70442f225780da1e96eb99864b6e77)

charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
```diff
@@ -48,7 +48,7 @@ spec:
             path: /health
             port: 8001
             scheme: HTTP
-          initialDelaySeconds: 120
+
```
pierDipi: nit: do we need all these blank lines?

pierDipi (approved):
/lgtm

charts/kserve-llmisvc-resources/templates/config-llm-decode-template.yaml
KillianGolds: Absolutely not 😅, addressing now!

KillianGolds: Addressed in [f4c5c87](https://github.com/kserve/kserve/pull/5063/commits/f4c5c874d2c02ce742b895a50c16e35887a08471)

andresllh (commented):
lgtm

spolti:
Seems to be pretty similar to https://github.com/kserve/kserve/pull/4840

pierDipi (approved):
/lgtm
/approve

sivanantha321 (approved):
/lgtm
/approve

sivanantha321:
/hold
Can you help resolve the precommit check failure ?

sivanantha321 (approved):

PR #5064, opened 2026-02-12
feat: add annotation-based runtime defaults for MLServer

Jooho:
Previously, runtime defaults were applied by hardcoded name matching (e.g., "kserve-mlserver"), preventing custom MLServer-based runtimes from receiving proper defaults. Now uses serving.kserve.io/serverType annotation to determine server type, enabling custom runtimes to inherit appropriate settings like protocol version and model implementation

Fixes #

```
kubectl create -k ./config/runtimes

kubectl create ns test
kubectl config set-context --current --namespace test

cat <<EOF|oc create -f -
kind: InferenceService
metadata:
  name: mlserver-xgboost-example-rest
spec:
  predictor:
    model:
      modelFormat:
        name: xgboost
        version: "1"
      name: ""
      resources: {}
      runtime: kserve-mlserver
      storageUri: gs://kfserving-examples/models/xgboost/iris
EOF
```
- [ ] Test A
- [ ] Test B

- Logs

The default protocolVersion is v1 now so auto-selector fails for mlserver for now unless you set` protocolVersion: v2`
So I added `runtime: kserve-mlserver`  for the test isvc.

**Checklist**:

**Release note**:
```release-note

```


pkg/apis/serving/v1beta1/inference_service_defaults.go
```diff
@@ -370,18 +370,23 @@ func (isvc *InferenceService) assignPaddleRuntime() {
...
+	if runtimeAnnotations == nil {
+		return
 	}
-	// add torchserve specific default values
-	if *isvc.Spec.Predictor.Model.Runtime == constants.TorchServe {
-		isvc.SetTorchServeDefaults()
+
+	serverType, exists := runtimeAnnotations[constants.ServerTypeAnnotationKey]
+	if !exists {
+		return
```
brettmthompson: should we fallback to using the `isvc.Spec.Predictor.Model.Runtime` if the annotation is not present for backwards compatibility?

Jooho: good idea. added

charts/kserve-resources/templates/clusterservingruntimes.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: serving.kserve.io/v1alpha1
 kind: ClusterServingRuntime
 metadata:
   name: kserve-lgbserver
+  annotations:
+    serving.kserve.io/serverType: lgbserver
```
spolti: imho `serving.kserve.io/runtime` is clearer in this case.

bartoszmajsak: also camelCase is not that common for annotations AFAICT

jlost:
Looks good!

charts/kserve-resources/templates/clusterservingruntimes.yaml
Jooho: >imho serving.kserve.io/runtime is clearer in this case.

`serverType` is correct to distinguish the server implementation (mlserver, triton, etc.) from the user-defined runtime name. Using `runtime` would conflict with `spec.predictor.model.runtime`.

>also camelCase is not that common for annotations AFAICT

Agreed I updated it to `server-type`

brettmthompson:
lgtm

spolti (approved):
/lgtm

PR #5065, opened 2026-02-13
test: harden patch-update E2E tests

bartoszmajsak:

Centralize generation-aware readiness in KServeClient by adding expected_generation support to wait_isvc_ready, and use the patched resource generation in autoscaling and canary tests to avoid stale post-patch assertions.

Also keep autoscaling env-update semantics correct by using separate initial and updated predictor specs so patch applies a real env change.

**Release note**:
```release-note
NONE
```


spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

sivanantha321 (approved):

PR #5066, opened 2026-02-13
chore(pytest): replaces deprecated get_event_loop

bartoszmajsak:

Replaces `asyncio.get_event_loop()` with `asyncio.new_event_loop()` in the session-scoped event_loop fixture. `get_event_loop()` is deprecated since Python 3.10 and emits DeprecationWarning. 

Using `new_event_loop` ensures a fresh loop is created and properly closed after the session.

**Release note**:
```release-note
NONE
```


Jooho (approved):
/approved
/lgtm

PR #5075, opened 2026-02-16
fix(llmisvc): allow custom gateway with managed route

bartoszmajsak:

Users should be able to attach custom Gateway while still relying on managed HTTPRoute. While the controller already has the logic to support it, the webhook was previously rejecting it as invalid, expecting a "bring your own" config to be an all-or-nothing scenario. This is obviously not desired when users want to attach their `LLMInferenceService` to another gateway.

Webhook validation now permits such configuration, and only rejects when HTTPRoute Spec of LLMInferenceService is also set, treating it as invalid configuration and suggesting removal of parentRef in favor of the global Gateway defined as `.spec.router.gateway.refs` instead

It also fixes the case when a custom gateway is specified via Gateway.Refs, the managed HTTPRoute was still referencing the default gateway from presets during validation, causing a RefsInvalid condition when the default gateway was absent. Move the gateway reference update from expectedHTTPRoute() to the config merge phase so the spec is correctly updated before reconciliation.

Port of opendatahub-io/kserve#911 and opendatahub-io/kserve#1028

**Release note**:
```release-note
NONE
```


bartoszmajsak:
@pierDipi biggest change comparing with midstream is due to #4856 - I added a commit specifically handling this 14232d55016bb2015fefb6840ff0160152bdbf36

pierDipi (approved):
Thank you!

/lgtm
/approve

terrytangyuan (approved):

PR #5076, opened 2026-02-16
fix(llmisvc): move gateway ref update to config merge phase

bartoszmajsak:

When a custom gateway is specified via Gateway.Refs, the managed HTTPRoute was still referencing the default gateway from presets during validation, causing a RefsInvalid condition when the default gateway was absent. Move the gateway reference update from expectedHTTPRoute() to the config merge phase so the spec is correctly updated before reconciliation.

Port of opendatahub-io/kserve#1028.

**Release note**:
```release-note
NONE
```


bartoszmajsak:
Consolidated into #5075

PR #5083, opened 2026-02-17
fix: skip validation for empty modelcar fields

bartoszmajsak:

When cpuModelcar and memoryModelcar fields are missing or empty in the inferenceservice-config ConfigMap, resource.ParseQuantity("") fails with a regex validation error, causing the admission webhook to reject pod creation. Make these fields truly optional by only validating them when non-empty.

Porf of opendatahub-io/kserve#930

**Release note**:
```release-note
NONE
```


spolti (approved):
/lgtm

sivanantha321 (approved):

PR #5085, opened 2026-02-17
fix: avoid storage-initializer rollout on operator upgrade

bartoszmajsak:

Preserve the existing storage-initializer container image when reconciling deployments and LeaderWorkerSets. During operator upgrades, the new storage-initializer image from the operator config would cause all pods to restart even when the model hasn't changed.

The fix fetches the current deployment/LWS before reconciliation and reuses the existing storage-init container image, avoiding unnecessary pod rollouts.

Port of opendatahub-io/kserve#1105

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm
/approve

terrytangyuan (approved):

PR #5009, opened 2026-01-29
feat(llmisvc): enable label/annotation propagation

khushiiagrawal:
**What this PR does / why we need it**:

This PR enables the propagation of custom labels and annotations from `LLMInferenceService` to the underlying model Deployment and Pods, addressing a key gap in metadata management for LLM workloads.

Specifically, it:
1.  **Workload Propagation**: Adds `Labels` and `Annotations` fields to `WorkloadSpec` (v1alpha1 & v1alpha2), allowing users to explicitly define metadata for the generated Pods.
2.  **Safety Filtering**: Implements a robust filtering mechanism for metadata inherited from the `LLMInferenceService` object:
    -   **Disallowed List**: Filters out keys present in `ServiceAnnotationDisallowedList` and `ServiceLabelDisallowedList` (loaded from KServe config).
    -   **Internal Annotation Protection**: Explicitly blocks propagation of internal KServe annotations (starting with `internal.serving.kserve.io/`) to prevent leaking internal state or conflicting with sidecar injection logic.
3.  **Config Integration**: Updates the controller to adhere to the global KServe configuration for disallowed lists.

Fixes #5005

**Feature/Issue validation/testing**:

- [x] **Unit Testing**: Added `TestPropagateDeploymentMetadata` in `pkg/controller/v1alpha2/llmisvc/label_propagation_test.go` to verify:
    -   Restricted labels/annotations (from the disallowed list) are correctly stripped.
    -   **Internal KServe annotations (`internal.serving.kserve.io/*`) are completely blocked.**
    -   Explicit `WorkloadSpec` labels correctly override inherited values on the Pod.
- [x] **Integration Testing**: Updated `TestLLMInferenceServiceController` in `pkg/controller/v1alpha2/llmisvc/controller_int_test.go` to verify end-to-end propagation.
- [x] **Verification**:
    -   Ran `make test` (PASSED).
    -   Ran `make precommit` (PASSED). **Note**: This updated installation scripts in `hack/setup/quick-install/` to match the new CRD schema.

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note
feat: Enable label and annotation propagation from LLMInferenceService to Pods with disallowed list and internal annotation filtering
```

PR #5086, opened 2026-02-17
feat(llmisvc): propagate Prometheus annotations to workloads

bartoszmajsak:
Propagate `prometheus.io/*` annotations (scrape, port, path, scheme) from the `LLMInferenceService` to the pod template of single-node deployments. This allows Prometheus to discover and scrape metrics from workload pods when annotations are set on the `LLMInferenceService` resource.

Port of opendatahub-io/kserve#962

**Release note**:
```release-note
NONE
```


pkg/controller/v1alpha2/llmisvc/workload_single_node.go
```diff
@@ -85,6 +85,20 @@ func (r *LLMISVCReconciler) expectedSingleNodeMainDeployment(ctx context.Context
 	labels[constants.KServeComponentLabelKey] = constants.KServeComponentWorkload
 	labels[constants.LLMDRoleLabelKey] = role
 
+	annotationsToPass := []string{
```
sivanantha321, line 88: I wonder if we need a config for passing annotations just like isvc.

sivanantha321 (approved):

pkg/controller/v1alpha2/llmisvc/workload_single_node.go
bartoszmajsak: Thanks for this suggestion @sivanantha321!

I think it makes sense to unify the approach. IIUC you are referring to "disallowed-list"? Do we want to use single source of truth for both kinds or have them separate? Overall sounds like a good follow-up.

sivanantha321: I think having a separate config feels better to me since, llmisvc is considered standalone. but, llmisvc already has dependency on `inferenceservice-config`, so not sure.

sivanantha321: looks like there is already a pr https://github.com/kserve/kserve/pull/5009

PR #5087, opened 2026-02-17
feat: route only completions through InferencePool

bartoszmajsak:

Split the single catch-all HTTPRoute rule into three specific rules:
- /v1/completions -> InferencePool (with URL rewrite)
- /v1/chat/completions -> InferencePool (with URL rewrite)
- catch-all base path -> Service (direct backend)

This ensures only completion endpoints are routed through the InferencePool for intelligent load balancing, while other traffic (health checks, model info, etc.) goes directly to the Service.

The extractRoutePath function is updated to prefer Service-backed rule paths for URL construction, falling back to the shortest path from any rule when no Service backend exists.

Port of opendatahub-io/kserve#938

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/llmisvcconfig/config-llm-router-route.yaml
+++ b/config/llmisvcconfig/config-llm-router-route.yaml
@@ -26,6 +26,47 @@ spec:
                 - path:
                     type: PathPrefix
                     value: |-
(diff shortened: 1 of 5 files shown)


pierDipi (approved):
/lgtm
/approve

terrytangyuan (approved):

PR #5088, opened 2026-02-17
fix(llmisvc): use server state for Delete owner verification

bartoszmajsak:

Rewrites the Delete function to fetch current server state via a Get call before performing deletion, replacing the previous approach that used the caller-provided expected object for ownership and deletion-timestamp checks. The old approach could use stale or incomplete data, causing confusing errors during namespace deletion and finalization.

Also adds NoMatchError handling to the Get operation, which occurs when the CRD for the resource is not installed on the cluster (e.g., LeaderWorkerSet). This prevents LLMInferenceService from reporting reconciliation failures on clusters that do not have optional CRDs installed.

Includes unit tests covering: CRD not installed (NoMatchError), resource not found (NotFound), and a characterization test documenting the difference between NoMatchError and NotFound error types.

Port of opendatahub-io/kserve#1031 and opendatahub-io/kserve#1056
**Release note**:
```release-note
NONE
```


spolti (approved):
/lgtm

sivanantha321 (approved):

PR #5090, opened 2026-02-18
feat: support path based routing in raw deployment mode

### LINKED ISSUE #4888: Support path-based routing for Standard Deployment mode
/kind feature

Knative deployment mode ingress with istio support path-based routing using a path template.  This allows multiple isvcs to deploy to the same domain and route to the individual endpoints using the URL path instead of the domain name.

Standard deployment mode only supports domain template routing.  Feature parity between the two deployment modes should be maintained such that Standard deployments can also use path routing templates when configuring ingress.

For example:
If `IngressDomain` is example.com,
`Namespace` is `default`,
`Name` is `TestModelEndpoint`,
Then setting the DomainTemplate to `{{ .IngressDomain }}` and the PathTemplate to `/namespaces/{{ .Namespace }}/endpoints//{{ .Name }}/` should result in an endpoint URL `http://example.com/namespaces/default/endpoints/TestModelEndpoint` that can be used to invoke the endpoint for inference.

copy-on-meow:
Adds the ability for the person deploying KServe to set up path based routing via the `ingressPathTemplate` field in the `inferenceservice-config` configmap:
```yaml
ingress: |-
    {   
      "ingressPathTemplate": "/namespaces/{{ .Namespace }}/endpoints/{{ .Name }}"
    }
```
This allows users to put all inference services behind a single domain and increases feature parity between the RawDeployment and Serverless mode.

Fixes #4888 

`make test` passed without issues

- In the original issue (#4888) the author suggested to use `pathTemplate` to configure the path template for RawDeployment mode, this however would result in a breaking change for RawDeployment users since the default is already set to a non empty value which would change the ingress prefixes in those deployments of KServe.
- I'm not familiar with the GatewayAPI that's why path based routing hasn't been implemented for RawDeployment's with GatewayAPI enabled

**Checklist**:

**Release note**:
```release-note
Added path based routing to `InferenceServices` deployed in `RawDeployment` mode. Users are able to configure the path template via `ingressPathTemplate` under the `ingress` key of the `inferenceservice-config` config map.
```

### DIFF
--- a/config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
+++ b/config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
@@ -15,5 +15,5 @@ data:
         "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
         "urlScheme": "http",
         "ingressPathTemplate": "/endpoints/{{ .Namespace }}/{{ .Name }}/",
-        "ingressDomain": "mydomain.com",
+        "ingressDomain": "mydomain.com"
     }
--- a/config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
+++ b/config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: inferenceservice-config
+  namespace: kserve
+data:
+  ingress: |-
+    {   
+        "enableGatewayApi": false,
+        "kserveIngressGateway": "kserve/kserve-ingress-gateway",
+        "ingressGateway": "knative-serving/knative-ingress-gateway",
+        "localGateway": "knative-serving/knative-local-gateway",
+        "localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local",
+        "ingressClassName": "istio",
+        "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
+        "urlScheme": "http",
+        "ingressPathTemplate": "/endpoints/{{ .Namespace }}/{{ .Name }}/",
+        "ingressDomain": "mydomain.com",
+    }
--- a/config/overlays/test/configmap/inferenceservice.yaml
+++ b/config/overlays/test/configmap/inferenceservice.yaml
@@ -365,11 +365,11 @@ data:
            # NOTE: This configuration only applicable for raw deployment.
            "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
 
-           # ingressPathTemplate to expose all InferenceServices in a cluster behind a single domain by combining the following variables:
+           # ingressPathTemplate specifies the template for generating the path of url's for each inference service by combining variables from:
            # Name of the inference service  ( {{ .Name}} )
            # Namespace of the inference service ( {{ .Namespace }} )
-           # If not set or empty: the root path is used
-           # NOTE: This configuration only applicable for raw deployment.
+           # If not set or empty: the root path is used.
+           # NOTE: This configuration only applicable for raw deployment with Gateway API disabled.
            "ingressPathTemplate": "/namespaces/{{ .Namespace }}/endpoints/{{ .Name }}",
 
            # urlScheme specifies the url scheme to use for inference service and inference graph.
--- a/config/overlays/test/configmap/inferenceservice.yaml
+++ b/config/overlays/test/configmap/inferenceservice.yaml
@@ -94,13 +94,13 @@ data:
         "resource": {
            # cpuLimit is the limits.cpu to set for the inference service.
            "cpuLimit": "1",
-    
+
            # memoryLimit is the limits.memory to set for the inference service.
            "memoryLimit": "2Gi",
-    
+
            # cpuRequest is the requests.cpu to set for the inference service.
            "cpuRequest": "1",
-    
+
            # memoryRequest is the requests.memory to set for the inference service.
            "memoryRequest": "2Gi"
         }
@@ -364,7 +364,14 @@ data:
            # If domain template is empty the default template {{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }} is used.
            # NOTE: This configuration only applicable for raw deployment.
            "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
-     
+
+           # ingressPathTemplate to expose all InferenceServices in a cluster behind a single domain by combining the following variables:
+           # Name of the inference service  ( {{ .Name}} )
+           # Namespace of the inference service ( {{ .Namespace }} )
+           # If not set or empty: the root path is used
+           # NOTE: This configuration only applicable for raw deployment.
+           "ingressPathTemplate": "/namespaces/{{ .Namespace }}/endpoints/{{ .Name }}",
+
            # urlScheme specifies the url scheme to use for inference service and inference graph.
            # If urlScheme is empty then by default http is used.
            "urlScheme": "http",
--- a/pkg/apis/serving/v1beta1/configmap.go
+++ b/pkg/apis/serving/v1beta1/configmap.go
@@ -127,10 +127,10 @@ type IngressConfig struct {
 	// This field has been added to prevent enabling path based routing by default in RawDeployment without GatewayAPI
 	//
 	// If using any other mode, use IngressConfig.PathTemplate
-	IngressPathTemplate     string `json:"ingressPathTemplate,omitempty"`
-	UrlScheme               string `json:"urlScheme,omitempty"`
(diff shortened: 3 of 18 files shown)


sivanantha321:
@copy-on-meow Can you add an end to end test for this ?

copy-on-meow:
@sivanantha321 Maybe I'm not seeing something but to me it seems like all the e2e tests are run against the default values in the controller configmap (minus the options for setting the deployment mode and networking layer), the feature that this PR introduces is disabled by deafult for backwards compatibility so it's currently not really possible to write e2e tests for it. 

To write tests for this feature we'd need a way to run e2e tests against different configurations in the `inferenceservice-config` config map, if there is already such a way I'd appreciate some pointers, if not I could look into introducing it but that's probably beyond the scope of this PR.

sivanantha321:
Please refer this https://github.com/kserve/kserve/blob/c2517a578f6d7ffd17bb39e70551fe2d756c6c44/.github/workflows/e2e-test.yml#L501

copy-on-meow:
@sivanantha321 Could you check the latest changes? I added the E2E test case

config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
```diff
@@ -0,0 +1,19 @@
...
+        "enableGatewayApi": false,
+        "kserveIngressGateway": "kserve/kserve-ingress-gateway",
+        "ingressGateway": "knative-serving/knative-ingress-gateway",
+        "localGateway": "knative-serving/knative-local-gateway",
+        "localGatewayService": "knative-local-gateway.istio-system.svc.cluster.local",
+        "ingressClassName": "istio",
+        "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
+        "urlScheme": "http",
+        "ingressPathTemplate": "/endpoints/{{ .Namespace }}/{{ .Name }}/",
+        "ingressDomain": "mydomain.com",
```
spolti: This comma probably needs to be removed.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
```diff
@@ -233,6 +233,29 @@ func generateIngressHost(ingressConfig *v1beta1.IngressConfig,
 	}
 }
 
+// generateIngressHost return the path to be used when an configmap.IngressPathTemplate is set
```
spolti:
```suggestion
// generateIngressPath return the path to be used when an configmap.IngressPathTemplate is set
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -81,6 +81,10 @@ func createRawURL(isvc *v1beta1.InferenceService,
 	if err != nil {
 		return nil, err
 	}
+	url.Path, err = GenerateIngressPath(isvc.Name, isvc.ObjectMeta, ingressConfig)
```
spolti: what about the Gateway API routing still uses host-based matching?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler_test.go
```diff
@@ -0,0 +1,505 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			shouldFail: false,
+		},
+		"Transformer and Explainer{ ready": {
```
spolti:
```suggestion
		"Transformer and Explainer ready": {
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler_test.go
```diff
@@ -0,0 +1,505 @@
...
+							},
+						},
+					},
+				},
+			},
+			shouldFail: false,
+		},
+	}
+
+	// tests := []struct {
```
spolti: left over?

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
```diff
@@ -66,6 +67,35 @@ func GenerateDomainName(name string, obj metav1.ObjectMeta, ingressConfig *v1bet
...
+		Namespace: obj.Namespace,
+	}
+
+	tpl, err := template.New("ingress-path-template").Parse(ingressConfig.IngressPathTemplate)
+	if err != nil {
+		return "", err
+	}
+
+	buf := bytes.Buffer{}
+	if err := tpl.Execute(&buf, values); err != nil {
```
spolti, line 86: url.ParseRequestURI accepts absolute URIs (e.g., http://evil.com/path), so a template like http://test.com/{{ .Name }} would silently produce /model. Consider validating the raw rendered
    string starts with /.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -121,6 +122,7 @@ type IngressConfig struct {
 	IngressClassName           *string   `json:"ingressClassName,omitempty"`
 	AdditionalIngressDomains   *[]string `json:"additionalIngressDomains,omitempty"`
 	DomainTemplate             string    `json:"domainTemplate,omitempty"`
+	IngressPathTemplate        string    `json:"ingressPathTemplate,omitempty"`
```
spolti: Please consider adding documentation about this new field and how to use it. If possible, list the differences between this and the PathTemplate field.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
copy-on-meow: ~~I don't think it does but this specific change is a case of a misplaced function, `createRawURL` is defined in `httproute_reconciler.go`, tested in `httproute_reconciler_test.go` but only used in `kube_ingress_reconciler.go`, I'll move it to `kube_ingress_reconciler.go`.~~

Both HTTPRoute and Ingress reconcilers use the `createRawURL` function to set the `status.url` field on the `InferenceService` objects, I think I'll revert this change since it would set the wrong url when Gateway API is enabled.

config/overlays/test/configmap/inferenceservice-ingress-path-template.yaml
copy-on-meow: Removed it

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/kube_ingress_reconciler.go
copy-on-meow: Comment fixed

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/domain.go
copy-on-meow: Switched it to strings.HasPrefix

pkg/apis/serving/v1beta1/configmap.go
copy-on-meow: I also added documentation to the `PathTemplate` field with hints when to use those this or the other field.

config/overlays/test/configmap/inferenceservice.yaml
```diff
@@ -364,7 +364,14 @@ data:
            # If domain template is empty the default template {{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }} is used.
            # NOTE: This configuration only applicable for raw deployment.
            "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
-     
+
+           # ingressPathTemplate to expose all InferenceServices in a cluster behind a single domain by combining the following variables:
```
brettmthompson: Any reason for the description and note here to be different than in `config/configmap/inferenceservice.yaml`?

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -111,21 +112,30 @@ type MultiNodeConfig struct {
...
+	KnativeLocalGatewayService string    `json:"knativeLocalGatewayService,omitempty"`
+	LocalGateway               string    `json:"localGateway,omitempty"`
+	LocalGatewayServiceName    string    `json:"localGatewayService,omitempty"`
+	IngressDomain              string    `json:"ingressDomain,omitempty"`
+	IngressClassName           *string   `json:"ingressClassName,omitempty"`
+	AdditionalIngressDomains   *[]string `json:"additionalIngressDomains,omitempty"`
+	DomainTemplate             string    `json:"domainTemplate,omitempty"`
+	// Specifies the template for generating path based url's in the RawDeployment mode with GatewayAPI disabled.
+	//
+	// This field has been added to prevent enabling path based routing by default in RawDeployment without GatewayAPI
```
sivanantha321, line 127: How is path based routing enabled by default for ingress ?

PR #5092, opened 2026-02-18
ci: add workflow to check if tests are passed before merge

sivanantha321:
This resolves the issue where PRs are blocked from merging due to skipped required checks. There are two ways to handle this.
1. Use [wait-for-status-checks](https://github.com/poseidon/wait-for-status-checks) GitHub Action that polls the GitHub Check Runs on a Pull Request and waits until all runs succeed (or are skipped) or one of them fails.
2. Instead of using conditional GitHub Workflows, try to make each job conditional since skipped jobs are considered successes. This can be achieved using [dorny/paths-filter](https://github.com/dorny/paths-filter).

This PR makes use of the first approach as it requires less changes. This will block the PR merge only if any job is failed. skips will be considered as success.

We can try with this approach for now. If there are any issues, we can go with the second. 

**Notes to Owners**
- Add this job as the required check and remove other jobs.


spolti:
isn't easier define some mandatory tests?

yuzisun (approved):

bartoszmajsak:
@sivanantha321 more often than not this job is blocking merges - see #5220 

Not quite sure why it's stale and errors with complaints about PR title where PR-title check job is green. What are we missing?

PR #5094, opened 2026-02-19
feat: implement configmap caching

### LINKED ISSUE #5021: [Feature] Watch on ConfigMap with predicate and cache tuning to trigger reconcile when it changes
/kind feature

### Problem Statement

Both the LLM Inference Service Controller (v1alpha2) and InferenceService Controller (v1beta1) currently fetch the KServe ConfigMap (`inferenceservice-config`) directly from the API server on **every reconciliation loop**. This creates performance overhead through redundant API calls, increased etcd load, and higher reconciliation latency.

**Evidence from Codebase:**

1. **v1alpha2 LLM Controller** (`pkg/controller/v1alpha2/llmisvc/controller.go:177-180`):
```go
// TODO(ctrl): add watch on CfgMap with predicate and cache tuning to trigger reconcile when it changes
config, configErr := LoadConfig(ctx, r.Clientset)  // Direct API call every reconciliation
```

2. **LoadConfig Implementation** (`pkg/controller/v1alpha2/llmisvc/config_loader.go:69-74`):
```go
func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, error) {
    // Fetches directly from API server - NO CACHING
    isvcConfigMap, errCfgMap := v1beta1.GetInferenceServiceConfigMap(ctx, clientset)
```

The code already acknowledges this issue with a TODO comment at line 177 in the LLM controller.

### SOLUTION

Implement a **shared, thread-safe ConfigMap cache** with **watch-based automatic invalidation** that both controllers can use.

#### High-Level Architecture

```
┌─────────────────────────────────────────────────────────┐
│  Controller Manager Startup                             │
│  ┌────────────────────────────────────────────────────┐ │
│  │  ConfigMap Watcher                                  │ │
│  │  - Watch kserve/inferenceservice-config            │ │
│  │  - Update in-memory cache on changes               │ │
│  │  - Trigger reconciliation for affected resources   │ │
│  └────────────────────────────────────────────────────┘ │
└─────────────────────────────────────────────────────────┘
                    ↓ (cache updates)
┌─────────────────────────────────────────────────────────┐
│  Reconciliation Loops (v1beta1 & v1alpha2)              │
│  ┌────────────────────────────────────────────────────┐ │
│  │  Read from in-memory cache                          │ │
│  │  - Zero API calls                                   │ │
│  │  - Sub-millisecond access                           │ │
│  │  - Consistent snapshot per reconciliation           │ │
│  └────────────────────────────────────────────────────┘ │
└─────────────────────────────────────────────────────────┘
```

### PROPOSED IMPL

**1. Create Shared ConfigMap Cache Package**

- Create `pkg/controller/configcache/cache.go`:

**2. Update Controllers to Use Cache**

- v1alpha2 LLM Controller Changes
- Update event handlers that currently call `LoadConfig`
- v1beta1 InferenceService Controller Changes

**3. Add ConfigMap Watch with Predicate**

- Update `SetupWithManager` for both controllers

#### Risks & Mitigations

| Risk | Mitigation Strategy |
|------|---------------------|
| Stale cache on watch failure | Implement watch reconnection with exponential backoff + periodic refresh (every 5 min) |
| Initial cache miss during startup | Block first reconciliation until ConfigMap loaded; add readiness check |
| Memory overhead from caching | ConfigMap is ~50-100KB; negligible compared to controller memory footprint |
| Watch event loss (network partition) | Add generation tracking and periodic sync from API server |
| Race condition during updates | Use RWMutex for thread-safety; atomic cache replacement |

hfjxjjd123:
### **What this PR does / why we need it**:

Currently, every reconciliation loop reads the `inference-service-config` ConfigMap directly from the Kubernetes API server. This results in repeated API calls on each reconcile cycle across all controllers, adding unnecessary load and latency.

This PR introduces an **in-memory ConfigMap cache** (`pkg/controller/configcache`) that:
- Loads the ConfigMap once at startup via the API reader
- Keeps it up-to-date using an informer-based watch (via the controller-runtime manager cache)
- Exposes typed, thread-safe getters that return deep copies (e.g. `GetIngressConfig()`, `GetDeployConfig()`, `GetStorageInitializerConfig()`, etc.)
- Is registered as a `manager.Runnable` so its lifecycle is managed by the controller manager

All controllers (`InferenceService`, `InferenceGraph`, `LocalModel`, `LocalModelNode`, `LLMInferenceService`) are updated to receive the `ConfigCache` via constructor injection and use it in place of direct API reads.

### **Which issue(s) this PR fixes**:
Fixes #5021

- [x] `make precommit` passes (go vet, golangci-lint, codegen)
- [x] `make test` passes — unit/integration tests for all packages including the `configcache` package (479 lines of new tests) and all controller packages
- [x] Fixed pre-existing test suite hang in `pkg/controller/v1beta1/inferenceservice` caused by nil `ConfigCache` in test setup

- Logs: `make test` — 108/108 Ginkgo specs pass in `pkg/controller/v1beta1/inferenceservice`

- `SetupConfigCache` must be called before controllers are registered with the manager. See `cmd/manager/main.go` for the wiring pattern.
- On startup, if the ConfigMap does not yet exist, the cache initializes lazily via the informer `OnAdd` event — no error is returned.
- The `Get(ctx)` method returning the raw ConfigMap is a temporary measure until `pkg/credentials` is refactored to accept `CredentialConfig` directly (tracked as a TODO in `cache.go`).
- The `OnDelete` handler intentionally retains the last known config rather than zeroing it out, to avoid disrupting in-flight reconciliations.

**Checklist**:

**Release note**:
```release-note
Introduced in-memory caching for the `inference-service-config` ConfigMap to reduce repeated API server reads across all controllers. The cache is backed by a controller-runtime informer and exposes typed, thread-safe getters for all config subsections.
```

---

Does this look good, or would you like to adjust any part of it?


pierDipi:
/cc @VedantMahabaleshwarkar 

Vedant you might context on ConfigMap caching as we've run into issue in the past

cmd/llmisvc/main.go
bartoszmajsak, line 177: Any reason for removing this label selector?

pkg/controller/configcache/cache.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// ConfigCache provides thread-safe access to parsed ConfigMap configurations.
+// All typed getters return deep copies — callers may safely mutate returned values.
+//
+// Lifecycle methods (Start, WaitForCacheSync) are intentionally excluded from this
+// interface: Start satisfies manager.Runnable and is called by the manager internally;
+// WaitForCacheSync is not needed by controllers since SetupConfigCache performs the
+// initial load synchronously before returning.
+type ConfigCache interface {
```
bartoszmajsak: Is there a need for an interface?

bartoszmajsak (commented):
I think it's a desired improvement as we are bypassing the cache entirely at the moment. Thanks for bringing this up!

I wonder if we could replace existing `clientset` calls with `r.Client.Get()` and add  scoped `.Watches(&ConfigMap{}, ...)` for re-reconciliation of depending resources. Probably with deliberate no-op updates (`predicate.ResourceVersionChangedPredicate` comes to mind maybe combined with detection of relevant config parts).

cmd/llmisvc/main.go
hfjxjjd123: The label selector was moved to line 176 within namespace-specific filtering.

https://github.com/kserve/kserve/pull/5094/changes#diff-ff3ea38f8a47b9f0417d519ef56d3cc9386ac3862872721f5619a16d6b5df7f2R174-R179

Previously: All ConfigMaps filtered by label → excluded `inferenceservice-config`
Now: Namespace-specific filtering:
 - AllNamespaces: filter by label (llmisvc-owned ConfigMaps)
 - KServeNamespace: no filter (includes `inferenceservice-config`)

This is needed because the ConfigCache uses the manager's informer to watch
for changes to `inferenceservice-config`. The informer requires the ConfigMap
to be in the manager's cache, but `inferenceservice-config` doesn't have the
required label.

hfjxjjd123:
Thanks for the suggestion! You're right that using controller-runtime patterns(r.Client + scoped watches) would be better than centralized ConfigCache package.

I was concerned about the scope of changes, but your approach is cleaner:
- [x] Uses existing r.Client.Get() (which reads from manager's cache)
- [x] Each controller adds scoped .Watches(&corev1.ConfigMap{}, ...) for re-reconciliation
- [x] Add predicate.ResourceVersionChangedPredicate for re-reconciliation
=> Avoids introducing a new package and interface

I'll refactor to this approach. Thanks for the guidance!

VedantMahabaleshwarkar (approved):
@pierDipi I took a look at the changes. Looks good to me

bartoszmajsak:
@hfjxjjd123 I would like to review the new attempt, could you update your branch?

hfjxjjd123:
@bartoszmajsak Branch is updated and rebased onto latest master. Ready for review.

Key changes:
- Use r.Client.Get() (shared informer cache) instead of direct API calls for ConfigMap reads
- Each controller has scoped .Watches(&corev1.ConfigMap{}, ...) with `ResourceVersionChangedPredicate` for re-reconciliation
- Shared InferenceServiceConfigPredicate and GetInferenceServiceConfigMap helpers in pkg/utils/
- Unit tests added for all new utilities and handler functions

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -167,3 +169,30 @@ func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, e
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse StorageInitializerConfig: %w", err)
+	}
+
+	credentialConfig, err := credentials.GetCredentialConfig(configMap)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse CredentialConfig: %w", err)
+	}
+
+	return NewConfig(ingressConfig, storageInitializerConfig, &credentialConfig), nil
```
spolti: Did you run `make tests`  locally?
I suspect this will fail to compile.

pkg/apis/serving/v1beta1/configmap_deepcopy.go
```diff
@@ -0,0 +1,170 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+// DeepCopyInto copies all fields of IngressConfig into out. in must be non-nil.
+func (in *IngressConfig) DeepCopyInto(out *IngressConfig) {
```
spolti: Was it handwritten?
If so we should avoid it, since it is easy to lose track of it if anything new is added.

hfjxjjd123:
@spolti Thank you for catching issues. There were some mistakes while I updated my branch.
  
- I fixed issues and run `make test`  
- Also removed unused files from the previous approach.

hfjxjjd123:
@spolti @VedantMahabaleshwarkar @bartoszmajsak 
> PTAL whenever you have time — no rush

---

It's been a while since the last activity on this PR — apologies for the delay. A few things to flag before re-review:
On the first commit, I squashed an earlier series of commits into a single commit before pushing, but in the process I missed some things I should have double-checked. Some of the diff in that commit wasn't fully signed off in the original incremental review, so it deserves a fresh look rather than being treated as already-approved.

---

make precommit, make check, and make test all pass locally on the current state. Updates since the last review are listed below.  

**Main Updates**

- a9ef25085 fix: rbac changes for list, watch  
  - Adds list;watch verbs on configmaps, required by the native informer introduced in #5094.
  - Regenerated config/rbac/role.yaml

- c09d98c21 fix: switch remaining clientset reads to c.Client
  - Routes the three remaining LocalModelNamespaceCache reads through the cached client, with `mgr.GetAPIReader()`

- b4580fe21 feat: check data by DeepEqual for No-op update
  - Considering deliberate no-op updates, double check with DeepEqual and `predicate.ResourceVersionChangedPredicate`

- 86a6dd5f7 perf(cache): scope ConfigMap to KServe namespace via ByObject
  - Limits the manager cache to ConfigMaps in constants
  - KServeNamespace, refering the pattern in cmd/llmisvc/main.go

---

Happy to address feedback. If you'd prefer a tighter scope or a smaller diff, let me know and I'll split it.Is

bartoszmajsak:
@hfjxjjd123 thanks for this work. I'm back from holidays on Monday and will have a look.

hfjxjjd123:
For cleaning up the branch, closing this PR for now. Since #5573 has already landed and established a lighter-weight, controller-scoped pattern (client cache + scoped Watches + predicate) for llmisvc, I think it makes more sense to build on top of that pattern rather than introduce a separate centralized configcache package across all controllers.

Issue #4771, opened 2025-10-23
using KServe LocalModel with dynamic cluster autoscaling


using-system:
/kind bug

**What steps did you take and what happened:**

When using KServe LocalModel with dynamic cluster autoscaling, new `LocalModelNode` resources are not automatically created when new nodes join the cluster.

Steps to reproduce:
1. Deploy KServe with LocalModel enabled
2. Create a `LocalModelNodeGroup` with nodeAffinity
3. Create a `LocalModelCache` referencing the node group
4. Wait for cluster autoscaler to add a new node that matches the nodeAffinity
5. Check LocalModelNode resources: `kubectl get localmodelnodes.serving.kserve.io`

Result: The new node does not have a corresponding LocalModelNode resource.

The LocalModelNode agent on the new node logs:
```
{"level":"info","msg":"Skipping LocalModelNode because it is not for current node","name":"<other-node>","current node":"<new-node>"}
```

**What did you expect to happen:**

A `LocalModelNode` resource should be automatically created for the new node when it joins the cluster and matches the `LocalModelNodeGroup` nodeAffinity specification.

**What's the InferenceService yaml:**

Not related to InferenceService. The issue is at the LocalModel infrastructure level with LocalModelNodeGroup and LocalModelCache.

**Anything else you would like to add:**

**Root cause:** LocalModelNode resources are created by the LocalModelReconciler controller during LocalModelCache reconciliation. When new nodes join dynamically, the reconciliation is not always triggered reliably.

**Workaround:**
```bash
kubectl annotate localmodelcache <name> reconcile="$(date +%s)" --overwrite
```

This forces reconciliation and creates the missing LocalModelNode.

**Impact:** Affects environments with:
- Cluster autoscaling
- Spot/preemptible instances
- Dynamic node scaling

**Environment:**

- Istio Version: N/A
- Knative Version: N/A (RawDeployment mode)
- KServe Version: v0.16.0-rc0
- Kubeflow version: N/A
- Cloud Environment: Azure AKS
- Kubernetes version: v1.33.2
- OS: Ubuntu 22.04

johnthompson-ybor:
i saw this myself and noticed that it only comes up if i restart the kserve pods (version 0.15)

using-system:
> i saw this myself and noticed that it only comes up if i restart the kserve pods (version 0.15)

Yes same for me (version 0.16)

changminbark:
Hi, can I take this issue? Can I ask for more context?

[...]

PR #5095, opened 2026-02-19
fix(localmodel): use ObjectOld in node update predicate

tmvfb:
- Fixes a copy-paste bug in `nodePredicates.UpdateFunc` where `e.ObjectNew` was used for both `oldNode` and `newNode`, causing the NotReady→Ready transition check to always return `false`
- This prevented `LocalModelCache` reconciliation when nodes dynamically join the cluster (e.g., via cluster autoscaler), so `LocalModelNode` resources were never created for new nodes

**Which issue(s) this PR fixes**: #4771

**Details**:

In `pkg/controller/v1alpha1/localmodel/controller.go`, the node update predicate was:
```go
oldNode := e.ObjectNew.(*corev1.Node) // BUG: should be e.ObjectOld
newNode := e.ObjectNew.(*corev1.Node)
return !isNodeReady(*oldNode) && isNodeReady(*newNode)
```

Since both variables reference the same (new) object, `!isNodeReady(X) && isNodeReady(X)` is always `false`. The fix changes `oldNode` to use `e.ObjectOld`.

- The fix compiles successfully
- Existing unit/webhook tests pass
- The integration tests require envtest infrastructure (etcd/kubebuilder) and are unaffected by this change

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha1/localmodel/controller.go
+++ b/pkg/controller/v1alpha1/localmodel/controller.go
@@ -445,22 +445,7 @@ func (c *LocalModelReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
-	nodePredicates := predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			// Only reconciles the local model crs when the node becomes ready from not ready
-			// Todo: add tests
-			oldNode := e.ObjectNew.(*corev1.Node)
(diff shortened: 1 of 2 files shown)


sivanantha321 (approved):

sivanantha321:
Good find!

/lgtm
/approve

PR #5096, opened 2026-02-19
feat(llmisvc): add versioned well-known config resolution

bartoszmajsak:

Pin well-known config names into `Status.Annotations` at first reconciliation so that future config prefix changes don't affect existing services.

Port of opendatahub-io/kserve#999

**Release note**:
```release-note
Stabilizes config resolution for `LLMInferenceService` by binding well-known config references at creation time. Services are no longer affected by operator-level config naming changes and therefore enabling safe rolling upgrades.
```


Jooho (approved):
/lgtm
/approved

PR #5099, opened 2026-02-19
feat(llmisvc): add DNS/IP SAN & renewal to self-signed certs

bartoszmajsak:

Embeds DNS names and IP addresses as Subject Alternative Names in self-signed TLS certificates and only regenerates them when the current SANs no longer cover expected endpoints.

Port of opendatahub-io/kserve#957

**Release note**:
```release-note
NONE
```


Jooho (approved):
/lgtm 
/approved

PR #5100, opened 2026-02-19
chore: change 'owners' to 'project-leads' in OWNERS file

terrytangyuan:
Addressing feedback from CNCF:
> Term Inconsistency: The owners file uses owners:, members doc calls them Leads, and the Charter calls them Project Leads. It would be good to sort out all terms across doc doc while changing the charter.


spolti (approved):
/lgtm

PR #5104, opened 2026-02-20
feat(llmisvc): enhance Gateway API URL discovery

bartoszmajsak:

Rewrite URL discovery to support multiple Gateway listeners (HTTP/HTTPS), preferred URL scheme selection, wildcard hostname expansion, and internal cluster-local service discovery via gateway-backing Services. Add AddressTypeName classification for Addressable entries and rename error types for clarity (NoURLsDiscoveredError).

Port of opendatahub-io/kserve#1046

Additions on top of downstream:

#### guard against nil match.Path in `extractRoutePath`

HTTPRoute rules can contain matches with only header conditions and no path specification. The existing code unconditionally dereferenced match.Path.Value, causing a nil pointer panic for header-only matches.

Adds a nil check to skip matches without a path and a corresponding unit test exercising the header-only match scenario.

#### Additional hardening

Deterministic service selection when multiple services match the gateway label:

- sorts alphabetically and logs a warning
- guards against empty parent status in HTTPRoute readiness checks
- fixes namespace comparison in parentRef matching by threading the
route's own namespace as default
- synthesizes a condition when ResolvedRefs is missing from gateway
controller status
-f ixes IPv6 address handling in URL construction where
standard-port paths bypassed bracket wrapping
- adds TLS protocol listener test coverage

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_loader.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader.go
@@ -35,6 +35,7 @@ type Config struct {
 	SystemNamespace         string `json:"systemNamespace,omitempty"`
 	IngressGatewayName      string `json:"ingressGatewayName,omitempty"`
 	IngressGatewayNamespace string `json:"ingressGatewayNamespace,omitempty"`
+	UrlScheme               string `json:"urlScheme,omitempty"`
 
 	// Storage and credential configs are excluded from JSON serialization
 	// as they contain sensitive information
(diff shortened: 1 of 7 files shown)


pierDipi (approved):
/lgtm

sivanantha321:
@bartoszmajsak Can you resolve the conflicts ?

Jooho (approved):
/lgtm
/approved

PR #5105, opened 2026-02-20
fix(llmisvc): return error from selectListener on mismatch 

bartoszmajsak:

Changes listener selection logic to return an error instead of silently falling through to `Listeners[0]` when a `sectionName` does not match any listener, or panicking when the listeners slice is empty. 

Before this change, a `ParentReference` with a `sectionName` that didn't match any Gateway listener would silently select the first listener, potentially using the wrong scheme (HTTP vs HTTPS) and port for URL construction. An empty listeners slice caused an index-out-of-bounds panic.

While the Gateway API CRD currently   enforces MinItems=1, the spec notes this requirement MAY be dropped in a future release to support XListenerSets [1]. This  future-proofs the controller against that change.

[1] https://github.com/kubernetes-sigs/gateway-api/blob/58af9661c0f6d99cd34d1723f335af410d0c4756/apis/v1/gateway_types.go#L230

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router_discovery.go
+++ b/pkg/controller/v1alpha2/llmisvc/router_discovery.go
@@ -88,7 +88,10 @@ func DiscoverURLs(ctx context.Context, c client.Client, route *gwapiv1.HTTPRoute
 
 	// Extract URLs from each gateway based on its listeners and addresses
 	for _, g := range gateways {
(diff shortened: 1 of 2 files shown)


spolti (approved):
/lgtm

pierDipi (approved):
/lgtm
/approve

sivanantha321 (approved):
/lgtm
/approve

PR #5106, opened 2026-02-20
fix(llmisvc): unblocks certificate renewal

bartoszmajsak:
`semanticCertificateSecretIsEqual` returned true (equal) when the current certificate was expired, which told the generic Reconcile function that no update was needed. This blocked the freshly generated certificate from being persisted to the Secret, making the renewal path in `reconcileSelfSignedCertsSecret` dead code.

The fix changes the early return from `true` to `false` when the current certificate is expired, so that the update proceeds and the new certificate is written to the cluster.

The full renewal flow is:
1. `ShouldRecreateCertificate` detects expiry - `certFunc` generates new cert
2. `expectedSelfSignedCertsSecret` builds Secret with fresh cert data
3. Reconcile - Update calls `semanticCertificateSecretIsEqual`
4. (before) returned true - update skipped - expired cert persists
5. (after)  returns false - update proceeds - new cert is persisted

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm

sivanantha321 (approved):

PR #5109, opened 2026-02-21
fix(llmisvc): handle JSON marshal error in ReplaceVariables

bartoszmajsak:

`ReplaceVariables` discarded the `json.Marshal(llmSvcCfg)` error on line 351. While marshaling a well-typed Kubernetes struct practically never fails, discarding the error means any failure would surface as a confusing "failed to unmarshal config from template" error from the downstream `json.Unmarshal` call, making debugging harder than necessary.

**Release note**:
```release-note
NONE
```


sivanantha321 (approved):

pierDipi (approved):
/lgtm

PR #5111, opened 2026-02-21
fix(llmisvc): restricts GlobalConfig fields for templates

bartoszmajsak:

- Replaces the full `Config` pointer in the template data context with a restricted `templateGlobalConfig` struct exposing only `SystemNamespace`, `IngressGatewayName`, and `IngressGatewayNamespace`
- Previously the entire `Config` struct (including `CredentialConfig` and `StorageConfig`) was reachable through Go template expressions despite `json:"-"` tags, since `text/template` traverses struct fields directly
- In practice the severity is low — `CredentialConfig` holds metadata (secret names, S3 endpoints) not actual credentials, and users who can create `LLMInferenceService` CRs can already run arbitrary containers. This is a "defense-in-depth" fix

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_merge.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_merge.go
@@ -367,17 +367,34 @@ func ToParentRefs(gatewayRefs []v1alpha2.UntypedObjectReference) []gwapiv1.Paren
 	return parentRefs
 }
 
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm
/approve

sivanantha321 (approved):
/lgtm
/approve

Issue #4254, opened 2025-02-17
Support SGLang Runtime for LLMs


lizzzcai:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

[SGLang](https://github.com/sgl-project/sglang) is another LLM runtime that offers significant [performance](https://lmsys.org/blog/2024-09-04-sglang-v0-3/) improvements, and it actively developed to support new models (for example [DeepSeek](https://github.com/sgl-project/sglang/tree/main/benchmark/deepseek_v3#deepseek-v3-support)). It enhance KServe's capabilities for serving generative AI workloads and provides more runtime options to users.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

darwich6:
Any update on this issue?

haitwang-cloud:
Hey, folks. I am working on this topic in the PR https://github.com/kserve/kserve/pull/5124

haitwang-cloud:
Sorry for the late response.  

I’ve added the Sglang runtime to kserve and finally got this PR completed. Most of the code was written by Claude‑opus4.7. I’ve tested it on our Kubernetes cluster and everything works fine. I also pasted the test results into the description, so you can review them.

PR #5124, opened 2026-02-25
feat: add SGLang as an alternative LLM inference runtime

haitwang-cloud:

This PR adds support for SGLang as an alternative LLM inference runtime alongside vLLM, allowing users to choose between the two runtimes via the spec.runtime field.

Changes:
- Add LLMRuntime type with 'vllm' and 'sglang' enum values to API types
- Add Runtime field to LLMInferenceServiceSpec (defaults to 'vllm')
- Create config-llm-sglang-template.yaml for SGLang container configuration
- Update combineBaseRefsConfig to select appropriate template based on runtime
- Add unit tests for runtime selection in MergeSpecs
- Add sample YAML demonstrating SGLang usage

Usage:
  spec: runtime: sglang  # or vllm (default) model: uri: hf://meta-llama/Llama-3.2-3B-Instruct

Fixes # https://github.com/kserve/kserve/issues/4254

- [ ] Test A
- [ ] Test B

- Logs


pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -190,8 +194,12 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
...
 		switch {
 		case resolvedSpec.Worker == nil:
-			// single-node
-			refs = append(refs, corev1.LocalObjectReference{Name: wr.Resolve(llmSvc, configTemplateName)})
+			// single-node: select template based on runtime (vllm or sglang)
+			templateName := configTemplateName
+			if llmSvc.Spec.Runtime == v1alpha2.LLMRuntimeSGLang {
+				templateName = configSGLangTemplateName
+			}
+			refs = append(refs, corev1.LocalObjectReference{Name: wr.Resolve(llmSvc, templateName)})
```
pierDipi, line 202: maybe to avoid the constant ifs in each branch we could organize well know configs in a map per runtime and it would allow easier error handling for potentially unsupported settings for a given runtime

docs/samples/llm-sglang-example.yaml
pierDipi, line 1: there is an llmisvc directory of samples maybe we could place it in `docs/samples/llmisvc/sglang`

docs/samples/llm-sglang-example.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: llama-sglang
+spec:
+  runtime: sglang   # <-- Use SGLang runtime (default: vllm)
+  model:
+    uri: "hf://meta-llama/Llama-2-7b-hf"
+    name: llama2
+  replicas: 1
+  baseRefs:
+    - name: kserve-config-llm-sglang-template
```
pierDipi, line 13: since this is auto-injected based on the runtime, it can be removed (same for the example down below)

terrytangyuan (commented):
One question is whether we should reuse ServingRuntime instead of adding a new struct LLMRuntime.

haitwang-cloud:
```
⏺ E2E Test Results

  Controller image: built from feature/add-sglang-runtime-support branch

  vLLM Runtime (regression test)

  - Created LLMInferenceService with facebook/opt-125m (no runtime field, defaults to vLLM)
  - Controller reconciled successfully, all pods Running & Ready
  NAME                                                              READY   STATUS
  facebook-opt-125m-single-kserve-6d8df68b5c-fc4j5                  1/1     Running
  facebook-opt-125m-single-kserve-router-scheduler-64575845-h4gfc   2/2     Running
  - LLMInferenceService status:
  NAME                       READY
  facebook-opt-125m-single   True
  - Inference verified:
  curl localhost:8000/v1/completions -d '{"model":"facebook/opt-125m","prompt":"The meaning of life
  is","max_tokens":10}'
  → {"choices":[{"text":" never to abandon the moral law of man. -"}]}
  - Result: PASS — existing vLLM behavior unaffected by SGLang changes

  SGLang Runtime (new feature)

  - Created LLMInferenceService with runtime: sglang and facebook/opt-125m
  - Controller correctly merged kserve-config-sglang-template preset:
    - Image: lmsysorg/sglang:v0.5.11 (injected from template)
    - Command: python3 -m sglang.launch_server --model-path /mnt/models --served-model-name "facebook/opt-125m"
   --port 8000 --host 0.0.0.0
  - All pods Running & Ready:
  NAME                                                              READY   STATUS
  sglang-opt-125m-single-kserve-657cc7cbc5-wf7b6                    1/1     Running
  sglang-opt-125m-single-kserve-router-scheduler-776849d9f5-nfwtc   2/2     Running
  - LLMInferenceService status:
  NAME                       READY
  sglang-opt-125m-single     True
  - Inference verified:
  curl localhost:8000/v1/completions -d '{"model":"facebook/opt-125m","prompt":"The meaning of life
  is","max_tokens":10}'
  → {"choices":[{"text":" never to be like it used to be. By"}]}
  - Result: PASS — SGLang runtime preset merge, template rendering, and inference all working correctly

```

haitwang-cloud:
@terrytangyuan @pierDipi PTAL,tks

MohanKumar21:
Hi @haitwang-cloud. Thanks for this.  @terrytangyuan @pierDipi. However, instead of adding a runtime enum to the inferenceService CR, can be make the dependency to be on ServingRuntime, so we can integrate other runtime environments in the future as well. Let me know what you think. 
Happy to collaborate, contribute, or help with anything else needed to move this forward.

lizzzcai:
Thanks for the PR. Did an independent e2e validation building the controller from this branch (on CPU) — `runtime: sglang` works end-to-end with inference responses through the gateway. vLLM default behavior is unaffected.

A couple of things to address before merge:

**Helm chart not updated** — `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml` wasn't regenerated to include the SGLang config, so `kserve-config-sglang-template` won't be deployed for anyone using the default Helm install path. All `runtime: sglang` ISVCs will fail reconciliation silently.

**Image tag unpinned** — `lmsysorg/sglang:latest` should be pinned to a specific version, consistent with the other templates.

**Suggestion** — `lmsysorg/sglang:latest` is a 12.85 GB CUDA image but does auto-detect CPU when no GPU is available, so it works on CPU too. For CPU deployments `lmsysorg/sglang:v0.5.13-xeon` (1.3 GB) is a better fit. Worth adding a CPU sample similar to `docs/samples/llmisvc/opt-125m-cpu/`.

lizzzcai:
On the `ServingRuntime` topic, I agreed with @terrytangyuan and @MohanKumar21 — it gives open extensibility (no code change for new engines), operator-managed image versioning, and namespace-level overrides, while staying consistent with how `InferenceService` already works. Here's what I'd propose for this PR and what follows.

**Why not a `LLMRuntime` enum**

A closed enum requires a code change for every new engine. `ServingRuntime`/`ClusterServingRuntime` is already the operator-managed engine catalog in KServe — reusing it gives platform teams a single place to register and version inference engines for both `InferenceService` and `LLMInferenceService`.

**What `ServingRuntime` contributes to `LLMInferenceService`**

For now, `ServingRuntime` contributes the **image** — the part that needs operator-managed versioning. `LLMInferenceServiceConfig` owns the infrastructure layer (launch command template, probes, volumes, security context) and must stay for now because it uses `LLMInferenceService`-specific template functions (`{{ .Spec.Model.Name }}`, `{{ kvTransferConfig }}`, etc.) that `ServingRuntime` has no access to today.

Longer term, the launch command should move to `ServingRuntime` as well. Once the well-known `LLMInferenceServiceConfig`s are made engine-agnostic — aligning probes, volumes, and security context across engines — `kserve-config-llm-template` becomes a single generic infrastructure base and engine-specific configs like `kserve-config-sglang-template` are no longer needed. At that point `ServingRuntime` is the only place an operator touches to register a new engine: image, launch command, and engine-specific args all live there.

Merge order:
```
kserve-llm-sglang ClusterServingRuntime  (base: image)
  + kserve-config-sglang-template       (command, probes, volumes — no image)
    + spec.baseRefs
      + spec.template                   (highest priority)
```

`kserve-config-sglang-template` has no `image` field so the ServingRuntime's image survives the strategic merge untouched. `spec.template` can still override the image if the user explicitly sets one.

**Proposed changes**

1. Change `LLMRuntime` enum → `*string` for `spec.runtime`. Same field name, same user-facing value (`runtime: kserve-llm-sglang`), but now open to any `ServingRuntime`/`ClusterServingRuntime` name.

2. In `combineBaseRefsConfig`, resolve `spec.runtime` as a `ServingRuntime`/`ClusterServingRuntime` and prepend its container spec as the lowest-priority layer before `MergeSpecs`. Falls through silently if not found.

3. Add `ServingRuntime` and `ClusterServingRuntime` watches to `SetupWithManager` to re-reconcile on image updates.

4. Ship `kserve-llm-sglang` as a `ClusterServingRuntime` with a pinned image. Remove the image from `kserve-config-sglang-template`.

**Example**

```yaml
kind: ClusterServingRuntime
metadata:
  name: kserve-llm-sglang
spec:
  supportedModelFormats:
    - name: sglang
      version: "1"
      autoSelect: false
  containers:
    - name: main
      image: lmsysorg/sglang:v0.5.11
```

```yaml
kind: LLMInferenceService
metadata:
  name: opt-125m-sglang
spec:
  runtime: kserve-llm-sglang
  model:
    uri: hf://facebook/opt-125m
    name: facebook/opt-125m
  replicas: 1
  router:
    gateway: {}
    route: {}
    scheduler: {}
  template:
    containers:
      - name: main
        resources:
          limits:
            nvidia.com/gpu: "1"
            memory: 16Gi
```

**Follow-up**

Remove the default image from `kserve-config-llm-template` and ship `kserve-llm-vllm` as a `ClusterServingRuntime`, making the well-known configs fully engine-agnostic. At that point `spec.runtime` becomes the canonical engine selector across all `LLMInferenceService` topologies, and `kserve-config-sglang-template` is no longer needed.


Feel free to comment and share your thoughts

config/llmisvcconfig/config-sglang-template.yaml
```diff
@@ -0,0 +1,91 @@
...
+        env:
+          - name: HOME
+            value: /home
+          - name: HF_HUB_CACHE
+            value: /models
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: false
+          runAsUser: 0
```
lizzzcai: can it be non-root?

haitwang-cloud:
Tested this branch locally with the SGLang GPU sample.

Applied:
`docs/samples/llmisvc/single-node-gpu/llm-inference-service-sglang-opt-125m-gpu.yaml`

Verified pods:

| Pod | Status |
| --- | --- |
| `sglang-opt-125m-single-kserve-55b9d6b749-z5tfq` | `1/1 Running` |
| `sglang-opt-125m-single-kserve-router-scheduler-5cf6d4ffbb-x2qnn` | `1/1 Running` |

Verified the workload deployment used:

```text
image: lmsysorg/sglang:v0.5.14
command: python3 -m sglang.launch_server --model-path /mnt/models --served-model-name facebook/opt-125m --port 8000 --host 0.0.0.0
```
SGLang logs showed the server was healthy:
```
GET /health HTTP/1.1" 200 OK
Prefill batch, #new-seq: 1
```
A direct OpenAI-compatible /v1/completions request to the workload service returned a valid completion for facebook/opt-125m.

terrytangyuan (commented):
Maybe in a follow-up PR: could you help add some docs and content to the website to give some visibility to SGLang coverage?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -88,6 +88,27 @@ type LLMInferenceServiceSpec struct {
...
+	//
+	// +optional
+	Runtime *string `json:"runtime,omitempty"`
+
+	// TrustRemoteCode allows the inference runtime to execute custom model code bundled
+	// with model weights (e.g. HuggingFace models with custom architectures).
+	// Enable only when loading models from trusted sources — this executes arbitrary Python
+	// at model load time.
+	// +optional
+	TrustRemoteCode bool `json:"trustRemoteCode,omitempty"`
```
pierDipi, line 110: should this be in the LLMModelSpec closer to `spec.model` ?

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -114,6 +118,28 @@ const (
...
 
+// SGLangServingRuntimeName is the well-known ClusterServingRuntime name shipped
+// with KServe that supplies the SGLang container image. When spec.runtime is
+// set to this name, the controller selects the SGLang infrastructure template
+// (kserve-config-sglang-template).
+//
+// TODO: Longer term we want to make kserve-config-llm-template engine-agnostic
+// (no image, aligned probes / volumes / security context across engines) so
+// that the ServingRuntime alone drives engine selection. At that point this
+// name-based mapping and kserve-config-sglang-template can be removed.
```
pierDipi, line 133: why not doing it right from the beginning?

haitwang-cloud:
> Maybe in a follow-up PR: could you help add some docs and content to the website to give some visibility to SGLang coverage?

Sure, good idea.

pkg/controller/v1alpha2/llmisvc/config_merge.go
haitwang-cloud: > why not doing it right from the beginning?

If we proceed this way, the PR will turn into a giant one with countless functions. I suggest handling that in the next PR.

terrytangyuan:
Could you resolve the conflicts so we can get this merged soon?

haitwang-cloud:
@terrytangyuan PTAL again,thanks

terrytangyuan (approved):

terrytangyuan:
Merged for now. We can iterate in follow-up PRs if needed

lizzzcai:
@haitwang-cloud  can you help to create a [doc](https://github.com/kserve/website) on this support as well.

haitwang-cloud:
@terrytangyuan PTAL this follow up PR https://github.com/kserve/website/pull/748

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi: I had a comment, top-level `trustRemoteCode` doesn't make sense to me

bartoszmajsak:
> Merged for now. We can iterate in follow-up PRs if needed

@haitwang-cloud @terrytangyuan can we identify follow-up issues? I am afraid a lot of user-facing settings in `LLMInferneceService`, e.g. kv-cache, are not covered yet and these are serious gaps worth addressing.

I am also not sure we settled on the API shape https://github.com/kserve/kserve/pull/5124#discussion_r4034620284, and as it shipped already that's not ideal :(

bartoszmajsak:
@haitwang-cloud @terrytangyuan I went through what actually happens when `runtime: kserve-llm-sglang` meets the rest of the spec (rendered the workloads through the reconciler in envtest, checked flags against SGLang v0.5.14 `server_args.py`). 

TL;DR: Only plain single-node seems to work. The rest either crash-loops or quietly does something else:

**Crash-loops** - controller appends vLLM flags SGLang's argparse rejects
- `spec.model.lora` -> `--lora-modules` (and `--max-cpu-loras`)
- `spec.tracing` -> `--collect-detailed-traces all`

**Silently runs vLLM**
- `spec.prefill` (P/D) and `spec.worker` (multi-node) - presets hardcode `llm-d-cuda` + `vllm serve`

**Silently ignored**
- `kvCacheOffloading` (secondary-tier volume still gets mounted), `parallelism.data/expert/pipeline`
- TLS - no certs or ssl flags, Service still says `https`
- nit: any (unrecognized) runtime name other than `kserve-llm-sglang` gets vLLM by default.

Follow-ups  I would suggest, in order:
1. Reject unsupported combinations at admission, so users get an error instead of a surprise
2. Mark SGLang experimental in the website docs (kserve/website#748)
3. e2e for the single-node path (should really be part of this PR or stack of PRs for this feature)
4. Settle where `trustRemoteCode` lives before a release freezes it /cc @cjohannsen-cloudera

haitwang-cloud:
@bartoszmajsak,

Thanks for the thorough investigation — your findings are accurate. The current KServe + SGLang integration is intentionally just a first step, with only single-node mode supported for now. I'll continue working on the advanced features you listed and will add admission validation to surface unsupported combinations as clear errors.

I've also started addressing the documentation gap here: https://github.com/kserve/website/pull/748 — feedback and comments are very welcome.

Issue #5126, opened 2026-02-25
test `predictor/test_pod_watch.py::test_event_storm_prevention_init_container_isolation` is flaky


pierDipi:
/kind bug

**What steps did you take and what happened:**
test `predictor/test_pod_watch.py::test_event_storm_prevention_init_container_isolation` is flaky

```
FAILED predictor/test_pod_watch.py::test_event_storm_prevention_init_container_isolation - AssertionError: Primary ISVC 'isvc-primary-28fc2d' was modified during secondary ISVC failure. ResourceVersion changed from 8924 to 8932. This indicates potential event storm - init container status changes on secondary ISVC may have triggered modification of unrelated primary ISVC.
assert '8924' == '8932'
  - 8932
  ?   -
  + 8924
  ?    +
```

**What did you expect to happen:**
Test sucesss 

**What's the InferenceService yaml:**
[To help us debug please run `kubectl get isvc $name -n $namespace -oyaml` and paste the output]

**Anything else you would like to add:**
- Failed in https://github.com/kserve/kserve/pull/5110

**Environment:**

- Istio Version:
- Knative Version:
- KServe Version: main
- Kubeflow version:
- Cloud Environment:[k8s_istio/istio_dex/gcp_basic_auth/gcp_iap/aws/aws_cognito/ibm]
- Minikube/Kind version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

Siva-Sainath:
I would like to work on this @pierDipi

pierDipi:
Thanks @Siva-Sainath

sivanantha321:
fixed in #5129

PR #5129, opened 2026-02-25
fix(isvc): preserve reason/message on true conditions

Siva-Sainath:
Fixes a flaky e2e test ([test_event_storm_prevention_init_container_isolation]  caused by the controller performing spurious status updates on unrelated InferenceService resources.

The issue originates in InferenceServiceStatus.SetCondition,  when a component reconciler propagated a True condition (e.g., PredictorReady) with a Reason/Message from a Deployment, then SetCondition calls Knative’s MarkTrue() helper. 

MarkTrue() intentionally omits the reason/message, resulting in the new condition differing from the existing one. 

Knative then replaced the condition with an updated LastTransitionTime, causing every reconciliation of a stable InferenceService to update its status and bump its resourceVersion.

Secondary ISVC failures therefore triggered repeated writes to the primary ISVC, making the event-storm isolation test flaky.

This PR changes SetCondition to invoke MarkTrueWithReason() when the status is True, preserving the reason and message. No other behavior is changed. The fix stops unnecessary status writes and restores deterministic test results.

Which issue(s) this PR fixes:
predictor/test_pod_watch.py::test_event_storm_prevention_init_container_isolation is flaky #5126

No new tests added since existing e2e covers the scenario; fix is minimal.


Jooho (approved):
/lgtm
/approved

Issue #5012, opened 2026-01-29
[Feature] Integrate Workload Variant Autoscaler (WVA) with LLMInferenceService


vivekk16:
/kind feature

**Describe the solution you'd like**
I would like to extend the LLMInferenceService API to integrate with the Workload Variant Autoscaler (WVA) to enable intelligent autoscaling for single-node LLM inference deployments. Unlike traditional autoscaling based on CPU or memory, this solution scales based on saturation metrics such as KV cache utilization and queue depth.

**Anything else you would like to add:**
The implementation involves a WVA controller that gathers server metrics from Prometheus via a Collector and uses a Saturation Analyzer to determine optimal replica counts based on defined thresholds, such as 80% KV cache usage or a queue depth of 10 requests. To ensure configuration integrity, the admission webhook will enforce validation rules where static replicas and scaling.wva are mutually exclusive, and users must explicitly opt-in by specifying an hpa: {} or keda: {} block. The controller will automatically manage the lifecycle of the VariantAutoscaling CR, Service Monitors for Prometheus scraping, and the respective scaling actuator (HPA or KEDA). This design currently focuses on single-node deployments where each replica runs a complete vLLM server with full model weights.

Proposal: https://docs.google.com/document/d/1hHYyWq9utKIdutqkHFQGzOhoW6vVqrpL68prW3u9-Ts/

PR #5130, opened 2026-02-25
feat(llmisvc): add autoscaling API with WVA support

vivekk16:

Adds an autoscaling API to `LLMInferenceService` (both `v1alpha1` and `v1alpha2`) via a new `ScalingSpec` field on `WorkloadSpec`. The scaling spec supports a Workload Variant Autoscaler (WVA) backend with two actuator options: HPA (Horizontal Pod Autoscaler) and KEDA (ScaledObject). This enables users to configure autoscaled `LLMInferenceService` deployments for both the decode and prefill workloads independently.

Key additions:
- `ScalingSpec` with `minReplicas`, `maxReplicas`, and `wva` fields
- `WVASpec` with mutually exclusive `hpa` (`HPAScalingSpec`) and `keda` (`KEDAScalingSpec`) actuators, plus `variantCost`
- `HPAScalingSpec` supports custom `behavior` (scale-up/scale-down policies)
- `KEDAScalingSpec` supports `idleReplicaCount` to set a lower replica floor when no triggers are active (≥ 1 and < `minReplicas`; scale-to-zero is not supported)
- Validation, conversion between `v1alpha1` ↔ `v1alpha2`, updated CRD YAMLs, Helm charts, and quick-install scripts

Partially addresses #5012

- [x] Unit tests for `validateWorkloadScaling` covering all validation rules (replicas/scaling mutual exclusivity, `maxReplicas` required, `minReplicas` ≤ `maxReplicas`, WVA required, HPA/KEDA mutual exclusivity, `variantCost` format, `idleReplicaCount` < `minReplicas`)
- [x] Unit tests for `validateScaling` covering decode + prefill workload path reporting and correct field paths
- [x] Unit tests for `v1alpha1` ↔ `v1alpha2` round-trip conversion of `ScalingSpec` (HPA, KEDA, nil scaling, prefill scaling, mixed decode/prefill scaling)

- This PR covers only the **API layer** — types, validation, and conversion. Controller changes to manage the WVA CR, ServiceMonitor, and HPA/KEDA actuator lifecycles will follow in a subsequent PR.
- `scaling` and `replicas` are mutually exclusive — `scaling` is for autoscaled deployments, `replicas` for static ones.
- `wva` is required when `scaling` is configured; within `wva`, exactly one of `hpa` or `keda` must be set.
- `idleReplicaCount` (KEDA-only) must be ≥ 1 and strictly less than `minReplicas`. **Scale-to-zero is not supported.**
- `variantCost` accepts non-negative numeric strings (e.g. `"10"`, `"0.5"`).

**Checklist**:

**Release note**:
```release-note
Add autoscaling API to LLMInferenceService via ScalingSpec on WorkloadSpec, supporting WVA (Workload Variant Autoscaler) with HPA and KEDA backends. Scaling can be configured independently on decode and prefill workloads. Scale-to-zero is not supported. Controller-side WVA integration will follow in a subsequent PR. Includes validation, v1alpha1<->v1alpha2 conversion, and updated CRDs.
```


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -283,6 +296,100 @@ type InferencePoolSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="!(has(self.hpa) && has(self.keda))",message="hpa and keda are mutually exclusive; choose one actuator backend"
+// +kubebuilder:validation:XValidation:rule="has(self.hpa) || has(self.keda)",message="either hpa or keda must be specified as the actuator backend"
+type WVASpec struct {
+	// VariantCost specifies the cost per replica for this variant (used in saturation analysis).
+	// Must be a non-negative numeric string (e.g., "10", "10.0", "0.5").
+	// Defaults to "10.0" if not specified.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +kubebuilder:default="10.0"
+	VariantCost string `json:"variantCost,omitempty"`
```
pierDipi, line 356: should we inline the VASpec directly so that it's easier to add fields as they are added in llm-d?

vivekk16: Good point. I can see the argument for inlining as it keeps the API flatter as llm-d evolves. But keeping wva as a sub-key also leaves room to add other scaling mechanisms alongside it (standalone HPA/KEDA) in the future without a breaking change. Happy to inline if the consensus is that WVA is the only mechanism we'll support here. What do you think?

pierDipi: I mean it will still be under `scaling.wva.variantCost`, it is just the type that instead of having individual field will have `Scaling.WVASpec.VariantAutoscalingSpec ...,inline`

vivekk16: @pierDipi 

I looked into inlining `VariantAutoscalingSpec` from the llm-d WVA module, but it doesn't quite work well. The upstream type has `scaleTargetRef` and `modelID` marked as required fields (with `+kubebuilder:validation:Required` and not `omitempty`), so inlining would pull those constraints into our CRD and force users to fill in fields that the controller should be managing. There is no clean way to suppress or override those markers from the embedding side, adding `CEL` rules to block them just creates a contradiction where `Required` says "must set" and `CEL` says "must not set."

andresllh: You bring up a great point @vivekk16 and I agree with your point that inlining won't be possible here. Your current approach is the correct way of handling these types of situations.

pierDipi: Sounds good to me too!

charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
```diff
@@ -123,6 +123,225 @@ spec:
                     format: int32
                     minimum: 0
                     type: integer
+                  scaling:
+                    properties:
+                      maxReplicas:
```
lionelvillard, line 144: out of curiosity, why `maxReplicas` and `minReplicas` `scaling` properties and not `hpa` properties? (and `horizontalPodAutoscalerConfig`) properties for KEDA.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -283,6 +296,100 @@ type InferencePoolSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas is the maximum number of replicas for the deployment.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// WVA configures the Workload Variant Autoscaler (WVA) for saturation-based scaling.
```
asm582: Saturation-based scaling is one approach; post v0.6, we are on the path to include SLO-based scaling. Can we make this more general, please?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -283,6 +296,100 @@ type InferencePoolSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="!has(self.minReplicas) || !has(self.maxReplicas) || self.minReplicas <= self.maxReplicas",message="minReplicas cannot exceed maxReplicas"
+// +kubebuilder:validation:XValidation:rule="!has(self.wva) || !has(self.wva.keda) || !has(self.wva.keda.idleReplicaCount) || has(self.minReplicas)",message="minReplicas is required when idleReplicaCount is set; idleReplicaCount must be less than minReplicas"
+// +kubebuilder:validation:XValidation:rule="!has(self.wva) || !has(self.wva.keda) || !has(self.wva.keda.idleReplicaCount) || !has(self.minReplicas) || self.wva.keda.idleReplicaCount < self.minReplicas",message="idleReplicaCount must be less than minReplicas; idleReplicaCount defines the replica floor when no triggers are active"
+type ScalingSpec struct {
+	// MinReplicas is the minimum number of replicas for the deployment during active scaling.
+	// This is the scaling floor when triggers are active.
+	// For idle scale-down, use KEDA's idleReplicaCount instead.
+	// Defaults to 1 if not specified.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
asm582, line 333: How do we support scaling to 0 when minReplicas is 1?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -97,12 +99,23 @@ type LLMInferenceServiceSpec struct {
...
 	// +kubebuilder:validation:Minimum=0
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// Scaling configuration for autoscaling the workload deployment.
+	// When specified, the controller creates and manages autoscaling resources
+	// (VariantAutoscaling CR, ServiceMonitor, and the selected actuator — HPA or KEDA ScaledObject)
+	// targeting this workload's Deployment.
+	// Mutually exclusive with the static 'replicas' field.
+	// In a disaggregated setup, each workload (decode and prefill) can have its own independent scaling configuration,
+	// resulting in separate VA CRs per deployment.
```
asm582: ScalingSpec only supports deployment as per the doc?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: Scale-to-zero will not be supported in the initial implementation due to the current absence of flow control. Validation logic will be updated in a later phase when the scale-to-zero feature is officially planned.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: Sure. I'll be happy to discuss more about this in our sync and make this spec more generic.

charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
vivekk16: `minReplicas` and `maxReplicas` are intentionally actuator-agnostic. They define the replica bounds at the `ScalingSpec` level so they are meaningful regardless of which actuator the user chooses (HPA or KEDA). The naming happens to match HPA's field names, but the semantic intent is generic ("this workload should scale between N and M replicas"). The controller is responsible for projecting these values into the actuator-specific fields (spec.minReplicas for HPA, spec.minReplicaCount/spec.maxReplicaCount for KEDA's ScaledObject). This avoids users having to re-specify the same bounds inside the actuator block just because they switched backends.

vivekk16: As far as `HorizontalPodAutoscalerConfig` for KEDA is concerned, it is taken care of through the advanced field available in `KEDAScalingSpec` (`keda.advanced.horizontalPodAutoscalerConfig`). KEDA internally manages an HPA as its actuator, and this field allows tuning its behavior. It is intentionally distinct from wva.hpa, which is for a standalone HPA without KEDA.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: Yes, only `Deployment` is supported as `LLMInferenceService` is stateless so `Deployment` is the correct Kubernetes primitive here. `StatefulSet` is not in scope for this proposal. The comment is describing what the controller creates, not a limitation of ScalingSpec itself. The wording could be clearer, I'll be happy to reword if "Deployment" is causing confusion.

sivanantha321:
lgtm

sivanantha321 (approved):

XiaoHanChina:
Thank you for the autoscaling API and WVA support in this PR — it’s exactly what we need for LLMInferenceService.

Could you share when this is planned to land in a release (e.g. 0.17 or a specific milestone)?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
pierDipi: @vivekk16 this is for the description
```suggestion
	// WVA configures the Workload Variant Autoscaler (WVA) for scaling.
```

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -283,6 +296,100 @@ type InferencePoolSpec struct {
...
+	// +kubebuilder:validation:Minimum=1
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas is the maximum number of replicas for the deployment.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// WVA configures the Workload Variant Autoscaler (WVA) for saturation-based scaling.
+	// WVA scales based on GPU saturation metrics (KV cache utilization, queue depth)
```
pierDipi:
```suggestion
	// WVA scales based on a variety of inference metrics (KV cache utilization, queue depth, etc)
```
Related to https://github.com/kserve/kserve/pull/5130/files#r2860988587

pierDipi: (to be updated also in v1alpha1 types)

pierDipi:
> Thank you for the autoscaling API and WVA support in this PR — it’s exactly what we need for LLMInferenceService.
> 
> Could you share when this is planned to land in a release (e.g. 0.17 or a specific milestone)?

@XiaoHanChina we're planning to have this implemented in the next few weeks, I'm not sure if it will make it in 0.17.0  but we're trying

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
pierDipi: I think we can also follow up on the description update to unblock the implementation

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
pierDipi: @vivekk16 let's reword this and stay generic

lionelvillard: Does the controller support multinode P/D (LWS)?

vivekk16: @lionelvillard Multinode scaling is out of scope for this initial implementation. It will be addressed in the future.

pierDipi (approved):
/lgtm

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -283,6 +296,123 @@ type InferencePoolSpec struct {
...
+	// MaxReplicas is the maximum number of replicas for the deployment.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
+	// WVA configures the Workload Variant Autoscaler (WVA) for scaling.
+	// WVA scales based on a variety of inference metrics (KV cache utilization, queue depth, etc.)
+	// rather than traditional CPU/memory metrics.
+	// +optional
+	WVA *WVASpec `json:"wva,omitempty"`
```
yuzisun, line 345: is the plan to support WVA as the only way to scale llmisvc ?

vivekk16: For now that’s the plan for implementation leaving enough flexibility to add HPA and KEDA standalone if needed at later point of time.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
```
yuzisun, line 17: The general question here is that if we need to keep updating v1alpha1 as we are transiting to v1alpha2, I already see code duplication here for the validations.

pkg/apis/serving/v1alpha1/llm_inference_service_validation.go
```diff
@@ -337,6 +342,140 @@ func (l *LLMInferenceServiceValidator) validateSchedulerConfig(svc *LLMInference
 	return allErrs
 }
 
+func (l *LLMInferenceServiceValidator) validateScaling(llmSvc *LLMInferenceService) field.ErrorList {
```
yuzisun, line 342: validations are copy-pasted identically between v1alpha1 and v1alpha2

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
vivekk16: @pierDipi what do you think?

pierDipi: we need to update both and support converting objects between API versions, that is why we need to have fields in both APIs

yuzisun: make sense, let's reuse the same validation code then we are good to merge.

ev-shindin:
@vivekk16 @pierDipi Is it possible to align this with ongoing API changes as stated here: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/806? What is your timeline?

pierDipi:
> @vivekk16 @pierDipi Is it possible to align this with ongoing API changes as stated here: [llm-d/llm-d-workload-variant-autoscaler#806](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/806)? What is your timeline?

We will target WVA 0.5.1 for this integration

sivanantha321 (approved):

sivanantha321:
@vivekk16 @pierDipi Are we targeting release 0.17 ? If yes, I can merge this, so we can focus on the Impl ASAP.

pierDipi:
yes

Issue #5142, opened 2026-02-28
manage.kserve-kustomize.sh: script is failing with kustomize: command not found error


smothiki:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
I ran `make deploy-dev` and the command failed with below error 
```
[INFO] Installing KServe CRDs...
[INFO]   - Installing CRDs from /Users/smothiki/goworld/src/github.com/kserve/kserve/config/crd/full...
hack/setup/infra/manage.kserve-kustomize.sh: line 168: kustomize: command not found
error: no objects passed to apply
make: *** [deploy-dev] Error 1
```

**What did you expect to happen:**
instead of using `kustomize` command script should be referring to `kubectl kustomize` that does the same job
Also in the development guide kustomize is not called as a requirement as kubectl already has kustomize command integrated.

**What's the InferenceService yaml:**
[To help us debug please run `kubectl get isvc $name -n $namespace -oyaml` and paste the output]
N/A
**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
full log can be found here https://gist.github.com/smothiki/ed58f2df65dbef2a73cfb4f7b6689222

**Environment:** 

- Istio Version: 1.27.1
- Knative Version: any
- KServe Version: 0.16
- Kubeflow version: any
- Cloud Environment:[k8s_istio/istio_dex/gcp_basic_auth/gcp_iap/aws/aws_cognito/ibm]
- Minikube/Kind version: 0.31.0
- Kubernetes version: (use `kubectl version`): v1.35.2
- OS (e.g. from `/etc/os-release`): macOS SEQUOIA  M1 ARM

Jooho:
It looks like the decision was made to use Kustomize, so I’ll go ahead and close this issue. If you need any further help, please feel free to reopen it.

PR #5145, opened 2026-02-28
fix: bug replace kustomize command with kubectl kustomize

smothiki:
 `make deploy-dev` command is failing with the below error
```
[INFO] Installing KServe CRDs...
[INFO]   - Installing CRDs from /Users/smothiki/goworld/src/github.com/kserve/kserve/config/crd/full...
hack/setup/infra/manage.kserve-kustomize.sh: line 168: kustomize: command not found
error: no objects passed to apply
make: *** [deploy-dev] Error 1
```

The PR addresses the issue by replacing kustomize command with kubectl kustomize command

Fixes #  https://github.com/kserve/kserve/issues/5142

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:
```release-note

```

### DIFF
--- a/hack/setup/infra/manage.kserve-kustomize.sh
+++ b/hack/setup/infra/manage.kserve-kustomize.sh
@@ -329,7 +329,7 @@ install() {
         for i in "${!TARGET_CRD_DIRS[@]}"; do
             crd_dir="${TARGET_CRD_DIRS[$i]}"
             log_info "  - Installing CRDs from ${crd_dir}..."
-            kustomize build "${crd_dir}" | kubectl apply --server-side --force-conflicts -f -
+            kubectl kustomize "${crd_dir}" | kubectl apply --server-side --force-conflicts -f -
 
             # Collect CRDs to verify
             crds="${TARGET_CRDS_TO_VERIFY[$i]}"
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-# The Go and Python based tools are defined in Makefile.tools.mk.
+safafa# The Go and Python based tools are defined in Makefile.tools.mk.
```
pierDipi:
```suggestion
# The Go and Python based tools are defined in Makefile.tools.mk.
```

smothiki: ACk Updated Makefile. Apologies some testing might have left the odd characters there

cjohannsen-cloudera (approved):

bartoszmajsak (commented):
FWIW `kubectl kustomize` is usually behind the standalone version, so we might be facing some old bugs depending on which version of `kubectl` is installed. Wouldn't it be better to ensure that kustomize is installed with the proper version?

WDYT @Jooho ?

smothiki:
> FWIW `kubectl kustomize` is usually behind the standalone version, so we might be facing some old bugs depending on which version of `kubectl` is installed. Wouldn't it be better to ensure that kustomize is installed with the proper version?
> 

I don't see any backward compatibility  changes in the release notes skimmed through release notes of kustomize for the last 7 months. But again I'm no expert will wait for other's opinion

Jooho:
@bartoszmajsak yes I prefer to use `kustomize`. We already created the centralized script to download the cli and also managed the version by `kserve-deps.env` . Using this way, every environment use the same version of kustomize and will provide the same result.

smothiki:
Closing the PR as kustomize is still used by many devs to try out things and this PR might abruptly cause kustomize obsolete .

PR #5152, opened 2026-03-02
fix: generate OpenAPI and Python SDK models for v1alpha2 LLMISVC

bartoszmajsak:

The v1alpha2 Go types (introduced in PR https://github.com/kserve/kserve/pull/4886) were never added to the OpenAPI generation pipeline, leaving the Python SDK without any `v1alpha2` model support.

Replaces the hardcoded list of serving API packages with auto-discovery that scans `pkg/apis/serving/v*/` for `+k8s:openapi-gen=true` markers. New API versions are now picked up automatically.

Adds a post-generation fixup to `hack/python-sdk/client-gen.sh` that corrects malformed model references in generated test files, a known bug in openapi-generator 4.3.1 where dotted swagger definition names produce broken Python division expressions.

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm

sivanantha321 (approved):

PR #5153, opened 2026-03-03
fix(llmisvc): add Gateway API v1.5.0 and listenersets RBAC support

thuanpham582002:
## Summary

- Update Gateway API from v1.4.1 to v1.5.0
- Use `experimental-install.yaml` instead of `standard-install.yaml` to include ListenerSet CRD
- Add RBAC patch for `listenersets` permission in Envoy Gateway ClusterRole
- Wait for `listenersets` CRD during installation

## Problem

Envoy Gateway v1.6.x requires Gateway API experimental CRDs (including `ListenerSet`) and RBAC permissions for the listenersets resource.

Without this fix, Envoy Gateway crashes with:
```
no matches for kind "ListenerSet" in version "gateway.networking.k8s.io/v1"
```

And even after installing the CRD:
```
listenersets.gateway.networking.k8s.io is forbidden: User
system:serviceaccount:envoy-gateway-system:envoy-gateway cannot list
resource "listenersets"
```

## Test Plan

- [ ] Verify Envoy Gateway starts without ListenerSet errors
- [ ] Verify HTTPRoute with InferencePool backend resolves correctly
- [ ] Run e2e tests for llmisvc


pierDipi (commented):
I think it's fine if we want to install experimental CRDs/manifests of GW API but I'm not sure about the Envoy GW cluster role patch done here

pierDipi:
is there any reference to an Envoy issue about this?

PR #5154, opened 2026-03-03
feat(llmisvc): add --disable-uvicorn-access-log to vllm configs

pierDipi:
Reduce excessive logging caused by EPP's aggressive metrics scraping (every 200ms) by disabling uvicorn access logs in all LLMInferenceServiceConfig templates.

**Release note**:

```release-note
Reduce excessive logging caused by EPP's aggressive metrics scraping (every 200ms) by disabling uvicorn access logs in all LLMInferenceServiceConfig templates.
```


charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       - '{{ .Spec.Model.Name }}'
       - --port
       - "8001"
+      - --disable-uvicorn-access-log
```
sivanantha321, line 17: There is a better flag available for this use case. 
https://docs.vllm.ai/en/latest/cli/serve/#-disable-access-log-for-endpoints

pierDipi: Yes, however, it would break when using previous versions of vLLM that don't support it (it's a very recent feature), that's why I'd avoid it for now in the llmisvcconfig

sivanantha321 (approved):

bartoszmajsak (approved):
/lgtm

anishasthana (commented):
@pierDipi Looking at https://docs.vllm.ai/en/latest/cli/serve/#-disable-uvicorn-access-log-no-disable-uvicorn-access-log, I'm a _little_ confused.

Is this what the new behaviour will be?

* It not specified, `--disable-uvicorn-access-log` will be added as a default
* If a user wants to enable logs, they should specify `--no-disable-uvicorn-access-log`

spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5157, opened 2026-03-03
fix(llmisvc): change TLS cert mount path to avoid CA bundle conflict

bartoszmajsak:

Mounting TLS certificates at `/etc/ssl/certs` can shadow the system CA bundle directory, breaking outbound HTTPS connections from containers. 

It can also break vLLM container: `OSError: Could not find a suitable TLS CA certificate bundle, invalid path: /etc/pki/tls/certs/ca-bundle.crt ` because `/etc/ssl/certs` can be a symlink.

Change the mount path to `/var/run/kserve/tls` across all config templates, tests, and e2e fixtures.

Port of opendatahub-io/kserve#963

**Release note**:
```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ Single platform that unifies Generative and Predictive AI inference on Kubernete
 ### Features
 
 **Generative AI**
-  * 🧠 **LLM-Optimized**: OpenAI-compatible inference protocol for seamless integration with large language models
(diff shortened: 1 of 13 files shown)


pierDipi (approved):
/lgtm
/approve

Jooho (approved):
/lgtm
/approved

sivanantha321 (approved):

PR #5164, opened 2026-03-04
build(make): add optional overrides include to Makefile

bartoszmajsak:

Adds `-include Makefile.overrides.mk` at the end of the Makefile, enabling downstream forks and local workflows to inject additional Make targets without modifying the main Makefile and reducing potential merge conflicts.

The `-` prefix causes `make` to silently skip the include when the file is absent, making this a no-op for upstream.

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm

spolti (approved):

PR #5173, opened 2026-03-04
fix(pytest): check readiness before sleeping

bartoszmajsak:

Reorder the polling loop to check readiness first, then sleep. This avoids an unnecessary initial delay when the resource is already ready, reducing test execution time by one polling interval per wait call.

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm

PR #5178, opened 2026-03-05
fix: resolve `go test -cover` failure

bartoszmajsak:

When the system Go version is older than what `go.mod` requires (e.g. Go 1.24 with `go 1.25.4` in go.mod), `GOTOOLCHAIN=auto` downloads a minimal toolchain to the module cache that is missing the `covdata` tool. This causes `go test -cover` to fail with `no such tool "covdata"` (exit code 2) even though all tests pass.

This PR:
- Splits the `go` and `toolchain` directives in `go.mod` / `qpext/go.mod` (`go 1.25.0` + `toolchain go1.25.4`) so the toolchain version survives `go mod tidy`
- Forces `GOTOOLCHAIN` in the Makefile to the exact version from `go.mod`, bypassing the broken auto-download path

Upstream Go bug: https://go.dev/issue/75031

- [x] Reproduced the failure with Go 1.24.13 (`MISE_GO_VERSION=1.24.13 mise exec -- make test-qpext` → exit code 2, `no such tool "covdata"`)
- [x] Verified the fix with Go 1.24.13 (same command → exit code 0)
- [x] Verified the fix with Go 1.26.0 (`make test-qpext` → exit code 0)

The same workaround is used by [etcd-operator](https://github.com/etcd-io/etcd-operator/pull/196) and other projects affected by the upstream Go bug.

`FORCE_HOST_GO=1` can be set to skip the override and use the local Go toolchain as-is.

**Checklist**:

```release-note
NONE
```


Jooho (approved):
/lgtm
/approved

PR #5068, opened 2026-02-13
test: automatic envtest cleanup and suite alignment

bartoszmajsak:
**What this PR does / why we need it**:

`envtest` child processes (`etcd`, `kube-apiserver`) leak when tests exit abnormally because cleanup relies entirely on callers remembering to register `DeferCleanup`. Moving context cancellation and environment teardown into `Config.Start()` makes cleanup automatic.

Migrates all Ginkgo-based suites from verbose, duplicated setup patterns (manual scheme registration, manager creation, goroutine launch) to the consolidated `NewEnvTest()` builder.

Eliminates duplicated API scheme wiring across managers and test environments.

This keeps behavior consistent between production runtime and `envtest` execution.

**Release note**:
```release-note
NONE
```

PR #5179, opened 2026-03-05
refactor(test): migrate pod webhook suite to centralized envtest setup

bartoszmajsak:

Migrates the pod webhook test suite (`pkg/webhook/admission/pod/suite_test.go`) from the
legacy `SetupEnvTest` helper to the centralized `NewEnvTest` builder introduced in #5068.

This eliminates manual CRD path construction, per-suite scheme registration, and global
`scheme.Scheme` mutation - replacing them with the shared configuration that already handles
all KServe and Kubernetes schemes.

Also introduces `BuildEnvironment()` on `Config` so that `TestMain`-based suites (which
cannot use Ginkgo's `DeferCleanup`) can obtain a configured `envtest.Environment` without
starting a manager.

Follow-up to #5068

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

### DIFF
--- a/pkg/testing/config.go
+++ b/pkg/testing/config.go
@@ -103,17 +103,10 @@ func (e *Config) WithManagerOptions(opts ...ManagerOption) *Config {
 	return e
 }
 
-// Start wires controller-runtime manager with controllers which are subject of the tests
-// and starts Kubernetes EnvTest to verify their behavior.
-func (e *Config) Start(ctx context.Context) *Client {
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
I have follow-up PR that splits unit/integration tests into _int_test.go and converts to Ginkgo for uniformity. I don't want to scope creep this one, so I stacked it.

sivanantha321 (approved):
/lgtm
/approve

PR #5180, opened 2026-03-05
fix(build): fall back to existing version when URL verification fails

bartoszmajsak:

When the GitHub API rate limit is hit during `make precommit` / `make check` in CI, `generate-versions-from-gomod.py` cannot verify release URLs and falls back to the stripped pseudo-version from `go.mod` (e.g. `v1.4.2`), even when no such release exists. This writes an unverified version to `kserve-deps.env` and the install
scripts, causing a dirty tree diff that fails CI with:

```
Please ensure that you have run 'make precommit' and committed the changes.
```

This fix reads existing values from `kserve-deps.env` before generation and uses them as fallback when URL verification fails, instead of writing the unverified pseudo-version.

Fixes flaky `precommit-check` CI failures caused by GitHub API rate limits.

- [x] Verified locally: script resolves `GATEWAY_API_VERSION` to `v1.4.1` (correct)
- [x] Simulated rate-limit scenario: fallback correctly keeps existing `v1.4.1` instead of writing unverified `v1.4.2`
- [x] When no existing value exists (fresh checkout), behavior is unchanged — uses the stripped pseudo-version

The root cause: `go.mod` references `sigs.k8s.io/gateway-api v1.4.2-0.20260116062110-...` (a pseudo-version for an unreleased commit). `strip_pseudo_version()` resolves this to `v1.4.2`, but no such GitHub release exists. Locally, the script finds `v1.4.1` via the GitHub releases API. In CI, the API is rate-limited (HTTP 403), so it falls back to the raw `v1.4.2`, creating a diff against the checked-in `v1.4.1`.

**Checklist**:

```release-note
NONE
```

### DIFF
--- a/hack/setup/scripts/generate-versions-from-gomod.py
+++ b/hack/setup/scripts/generate-versions-from-gomod.py
@@ -193,12 +193,26 @@ def ensure_helm_repo(name, url):
     run(f"helm repo add {name} {url}")
 
 
+def parse_existing_versions(env_file):
+    """Parse existing VAR=value pairs from kserve-deps.env"""
+    existing = {}
+    if env_file.exists():
+        for line in env_file.read_text().splitlines():
+            line = line.strip()
+            if line and not line.startswith("#") and "=" in line:
+                key, value = line.split("=", 1)
+                existing[key] = value
+    return existing
+
+
 def main():
     repo_root = Path(__file__).resolve().parent.parent.parent.parent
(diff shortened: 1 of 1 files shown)


Jooho (approved):
/lgtm
/approved

Jooho (approved):
/lgtm
/approved

Jooho (approved):
/lgtm
/approved

PR #5093, opened 2026-02-18
refactor(e2e/llmisvc): flexible endpoint testing

bartoszmajsak:
**What this PR does / why we need it**:

Replace hardcoded assertion functions with a configurable `create_response_assertion()` factory. 

Add payload formatter pattern (completions_payload, chat_completions_payload) and per-test endpoint field, enabling `GET` and `POST` dispatch in wait_for_model_response().

Remove `after_test` cleanup hooks in favor of idempotent `create_or_update_*` resource management. 

Drop `post_with_retry` in favor of direct requests calls.

Port of opendatahub-io/kserve#938
Port of opendatahub-io/kserve#935
Port of opendatahub-io/kserve#1062

**Release note**:
```release-note
NONE
```

PR #5187, opened 2026-03-06
feat(e2e/llmisvc): configurable test environment and new endpoint coverage

bartoszmajsak:

Follow-up to #5093 (next part of port).

Adds configurability and new endpoint coverage to the LLMInferenceService e2e test suite:

- **`GATEWAY_CLASS_NAME`** env var in `test_resources.py` (default `envoy`) so downstream deployments can override the gateway class without forking
- **`SKIP_DELETION_ON_FAILURE`** env var and shared `maybe_delete_llmisvc()` helper for debugging failed tests (keeps resources around for inspection)
- **`custom_gateway`** marker on router tests that create their own Gateway resources
- New **chat-completions** test case exercising `POST /v1/chat/completions` with `model-qwen2.5-0.5b`
- New **models-endpoint** test case exercising `GET /v1/models`
- Additional pytest markers in `pytest.ini`: `cluster_multi_node`, `cluster_gpu`, `no_scheduler`, `custom_gateway`, `auth`

- [x] Existing e2e test suite passes (no behavioral changes to existing tests)
- [x] New test cases validated with llm-d simulator

Builds on the flexible endpoint testing framework from #5093. The `payload_formatter` and `endpoint` fields introduced there are now exercised by the new chat-completions and models-endpoint test cases.

Port of opendatahub-io/kserve#938, opendatahub-io/kserve#1062, opendatahub-io/kserve#832, opendatahub-io/kserve#935.

**Checklist**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/llmisvc/test_llm_inference_service.py
+++ b/test/e2e/llmisvc/test_llm_inference_service.py
@@ -39,8 +39,8 @@ from .test_resources import (
     ROUTER_GATEWAYS,
     ROUTER_ROUTES,
 )
-from .logging import log_execution
-from ..common.http_retry import post_with_retry
+from .logging import log_execution, logger
+from ..common.http_retry import get_with_retry, post_with_retry
 
 KSERVE_PLURAL_LLMINFERENCESERVICE = "llminferenceservices"
 
@@ -88,7 +88,7 @@ class TestCase:
 
     __test__ = False  # So pytest will not try to execute it.
     base_refs: List[str]
-    prompt: str
+    prompt: Optional[str] = None
     service_name: Optional[str] = None
     endpoint: str = "/v1/completions"
     max_tokens: int = 100
@@ -148,6 +148,7 @@ def chat_completions_payload(test_case: TestCase) -> Dict[str, Any]:
                 pytest.mark.cluster_cpu,
                 pytest.mark.cluster_single_node,
                 pytest.mark.llmd_simulator,
+                pytest.mark.custom_gateway,
             ],
         ),
         pytest.param(
@@ -193,7 +194,11 @@ def chat_completions_payload(test_case: TestCase) -> Dict[str, Any]:
                     )
                 ],
             ),
-            marks=[pytest.mark.cluster_cpu, pytest.mark.cluster_single_node],
+            marks=[
+                pytest.mark.cluster_cpu,
+                pytest.mark.cluster_single_node,
+                pytest.mark.custom_gateway,
+            ],
         ),
         pytest.param(
             TestCase(
@@ -241,7 +246,11 @@ def chat_completions_payload(test_case: TestCase) -> Dict[str, Any]:
                     )
                 ],
             ),
-            marks=[pytest.mark.cluster_cpu, pytest.mark.cluster_single_node],
+            marks=[
+                pytest.mark.cluster_cpu,
+                pytest.mark.cluster_single_node,
+                pytest.mark.custom_gateway,
+            ],
         ),
         pytest.param(
             TestCase(
@@ -332,6 +341,25 @@ def chat_completions_payload(test_case: TestCase) -> Dict[str, Any]:
             ),
             marks=[pytest.mark.cluster_cpu, pytest.mark.cluster_single_node],
         ),
+        # Chat completions endpoint coverage
+        pytest.param(
+            TestCase(
+                base_refs=[
+                    "router-managed",
+                    "workload-llmd-simulator",
+                    "model-qwen2.5-0.5b",
+                ],
+                endpoint="/v1/chat/completions",
+                prompt="What is KServe?",
+                payload_formatter=chat_completions_payload,
+                response_assertion=create_response_assertion(with_field="choices"),
+            ),
+            marks=[
+                pytest.mark.cluster_cpu,
+                pytest.mark.cluster_single_node,
+                pytest.mark.llmd_simulator,
+            ],
+        ),
         pytest.param(
             TestCase(
                 base_refs=[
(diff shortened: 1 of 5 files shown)


test/e2e/pytest.ini
```diff
@@ -27,4 +27,9 @@ markers =
     cluster_nvidia: test targeting cluster with NVIDIA
     cluster_nvidia_roce: test targeting cluster with NVIDIA ROCe
     cluster_single_node: test targeting single node cluster
+    cluster_multi_node: test targeting multi node cluster
+    cluster_gpu: test targeting cluster with GPU
+    no_scheduler: test without scheduler
+    custom_gateway: test creating custom gateway resources
+    auth: test requiring authentication setup
```
bartoszmajsak, line 35: It's a dummy marker upstream - we can probably remove it.

test/e2e/llmisvc/test_llm_inference_service.py
```diff
@@ -429,6 +467,41 @@ def delete_llmisvc(kserve_client: KServeClient, llm_isvc: V1alpha1LLMInferenceSe
...
+    """Delete LLMInferenceService unless env vars instruct otherwise.
+
+    Respects SKIP_RESOURCE_DELETION (skip always) and
+    SKIP_DELETION_ON_FAILURE (skip only when test_failed is True).
+    """
+    service_name = llm_isvc.metadata.name
+    try:
+        skip_all = os.getenv(
+            "SKIP_RESOURCE_DELETION", "False"
+        ).lower() in ("true", "1", "t")
```
KillianGolds: Nit: I could be missing context here, but I noticed the previous inline logic in this file checked for `("false", "0", "f")` (delete unless explicitly falsy), whereas this now checks for `("true", "1", "t")` (skip only if explicitly truthy). The stop test already used the truthy-check style so this aligns them nicely, but it does subtly change the behavior here doesnt it, any unrecognized value (e.g. `"yes"`, a typo) used to trigger deletion and now won't.

Probably totally fine in practice, just wanted to flag it in case it wasn't intentional - feel free to ignore if you've already considered this!

test/e2e/llmisvc/test_llm_inference_service.py
```diff
@@ -467,23 +540,32 @@ def assert_model_responds():
...
+                    headers={"Content-Type": "application/json"},
+                    json_data=test_payload,
+                    timeout=test_case.response_timeout,
+                )
+            else:
+                response = get_with_retry(
+                    model_url,
+                    headers={"Content-Type": "application/json"},
+                    timeout=test_case.response_timeout,
+                )
```
KillianGolds, line 593: Very minor and totally nitpicky: this sends `Content-Type: application/json` on a GET with no body. Harmless in practice, but wondering if it might be slightly cleaner as `Accept: application/json` or just omitted for the GET path? Happy to defer to your judgement on this one.

test/e2e/llmisvc/test_llm_inference_service.py
```diff
@@ -84,7 +84,7 @@ class TestCase:
     """Test case configuration for LLM inference service tests."""
     __test__ = False  # So pytest will not try to execute it.
     base_refs: List[str]
-    prompt: str
+    prompt: str = ""
```
KillianGolds: Small thought but with `prompt` defaulting to `""` (falsy), the GET-vs-POST dispatch in `wait_for_model_response` becomes implicit. I was wondering if a test case that accidentally omits `prompt` (and has no `payload_formatter`) could end up silently sending a GET to `/v1/completions` instead of a POST.

Would something like `Optional[str] = None` or an explicit `method` field on `TestCase` help make the intent clearer? That said, this might be typical me over-engineering it for the current use cases haha - still curious what you think.

KillianGolds (commented):
Nice work, the `maybe_delete_llmisvc()` extraction is a solid cleanup and the `GATEWAY_CLASS_NAME` configurability is a clean approach without adding complexity.

Left a few minor observations inline, nothing blocking - mostly just flagging things I noticed in case they're useful. Overall this looks great to me 👍

test/e2e/llmisvc/test_llm_inference_service.py
bartoszmajsak: I was thinking a typo shouldn't trigger/skip the deletion and hence little cleanup here. What do we prefer? I was thinking - delete by default unless you really need to keep resources hanging around :)

test/e2e/llmisvc/test_llm_inference_service.py
bartoszmajsak: Copy-pastie :) Fixed.

test/e2e/llmisvc/test_llm_inference_service.py
bartoszmajsak: Not sure, I would assume `/v1/completions` without a prompt fails loudly. Implicit method is me being lazy but also what happens when someone sets method="GET"  but also provides a `payload_formatter`? :)

`Optional[str]`  and explicit dispatch is a good idea!

test/e2e/llmisvc/test_llm_inference_service.py
KillianGolds: Agreed, delete by default is the safer stance, a typo leaking resources in CI is worse than a typo cleaning them up. 👍

test/e2e/llmisvc/test_llm_inference_service.py
KillianGolds: Didn't think of that edge case 🫢 but I reckon with the current dispatch keyed off `payload_formatter`/`prompt` presence rather than an explicit `method` field, that conflict can't happen today. If a `method` field gets added later, a validation check in `__post_init__` could catch the mismatch but no need to over-engineer it now, typical me over engineering I think _(whos checking the checks that checks the check)_ haha dont worry looks solid.

spolti (approved):

PR #5194, opened 2026-03-07
feat(llmisvc): implement autoscaling for WVA and KEDA/HPA integration

### LINKED ISSUE #5012: [Feature] Integrate Workload Variant Autoscaler (WVA) with LLMInferenceService

vivekk16:
## Summary

Closes #5012

This PR adds autoscaling support to the llmisvc controller, enabling it to reconcile scaling resources based on the `ScalingSpec` defined in `LLMInferenceService`. The controller creates and manages:

- **VariantAutoscaling (WVA)** — a `VariantAutoscaling` CR is created per LLMInferenceService to drive workload-variant-aware scaling via the llm-d WVA controller.
- **HPA or KEDA ScaledObject** — the user explicitly chooses one actuator backend in the spec. HPA reads the `wva_desired_replicas` metric via the Kubernetes Metrics API (requires Prometheus Adapter), while KEDA queries Prometheus directly (no Prometheus Adapter needed).

### Changes

- **`scaling.go` (new)**: Core scaling reconciler that handles the full lifecycle of VariantAutoscaling, HPA, and KEDA ScaledObject resources — including creation, updates, ownership, cleanup of stale resources when scaling config changes, and blocking reconciliation when required CRDs are missing.
- **`scaling_test.go` (new)**: Unit tests covering HPA construction, KEDA ScaledObject construction (including Prometheus auth), VariantAutoscaling construction, DeepEqual-based semantic equality checks, KEDA finalizer/label preservation, accelerator label resolution, and naming helpers.
- **`controller.go`**: Registers WVA and KEDA API schemes, checks CRD availability at startup, adds HPA to owned resources, and adds kubebuilder RBAC markers for `horizontalpodautoscalers`, `variantautoscalings`, and `scaledobjects`.
- **`config_loader.go`**: Adds `WVAAutoscalingConfig` with nested `PrometheusConfig` (URL, TLS settings, bearer token auth) loaded from the `wva-autoscaling` key in the `inferenceservice-config` ConfigMap. Surfaces JSON parse errors instead of masking them.
- **`workload.go`**: Calls `reconcileScaling()` as part of workload reconciliation and blocks reconciliation when required scaling CRDs are missing.
- **`workload_single_node.go`**: Deployment name helpers moved here from scaling code; actuator reconciliation refactored to be self-managing.
- **`register.go`**: Centralized LWS, KEDA, and WVA scheme registration.
- **`cmd/llmisvc/main.go`**: Adds `HorizontalPodAutoscaler` to the controller's informer cache.
- **`config/rbac/llmisvc/role.yaml`**: Adds RBAC permissions for HPA, KEDA ScaledObject, and WVA VariantAutoscaling resources.
- **`go.mod`**: Adds `llm-d-workload-variant-autoscaler v0.5.1` dependency.

### Behavior

- When `spec.scaling` is set on an LLMInferenceService, the controller creates a `VariantAutoscaling` CR and either an HPA or KEDA ScaledObject, depending on which actuator the user specified (`spec.scaling.wva.hpa` or `spec.scaling.wva.keda`). HPA and KEDA are mutually exclusive, enforced by CEL validation.
- When autoscaling is removed from the spec, stale scaling resources are cleaned up.
- If the required CRD for the selected scaling backend is not installed, reconciliation is blocked with a clear error.
- VA accelerator label is resolved from `WorkloadSpec.Labels`, giving users explicit control.

## Testing

- [x] Unit tests for scaling reconciler (HPA, ScaledObject, VariantAutoscaling construction, semantic equality, naming helpers)


pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+	if slashIdx <= 0 {
+		return ""
+	}
+	domain := resourceName[:slashIdx]
+	dotIdx := strings.Index(domain, ".")
+	if dotIdx <= 0 {
+		return ""
+	}
+	return domain[:dotIdx]
+}
```
pierDipi: This is not a very reliable way of detecting the accelerator, if you look at the RDMA RoCE examples, the slash is used for other devices and map loop doesn't have a guarateed ordering 

https://github.com/kserve/kserve/blob/be70224872d12d36ec7caa248bec612089b7b128/docs/samples/llmisvc/dp-ep/deepseek-r1-gpu-rdma-roce/llm-inference-service-dp-ep-deepseek-r1-gpu-deepep-ht.yaml#L97-L98

@lionelvillard is `inference.optimization/acceleratorName` on the VA necessary?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+// podSpec is the pod template for the specific workload (decode or prefill) so that
+// accelerator detection inspects the correct containers.
+func expectedVA(llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.ScalingSpec, deploymentName, vaName string, podSpec *corev1.PodSpec) *wvav1alpha1.VariantAutoscaling {
+	labels := scalingLabels(llmSvc)
+	labels[acceleratorNameLabelKey] = resolveAcceleratorName(llmSvc, podSpec)
+
+	modelID := llmSvc.Spec.Model.URI.String()
+	if llmSvc.Spec.Model.Name != nil {
+		modelID = *llmSvc.Spec.Model.Name
+	}
```
pierDipi, line 377: should model id include llmisvc namespace and name?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+					Type: "prometheus",
+					Name: "wva-desired-replicas",
+					Metadata: map[string]string{
+						"serverAddress": config.AutoscalingConfig.PrometheusURL,
+						"query":         query,
+						"threshold":     "1",
+						"unsafeSsl":     strconv.FormatBool(config.AutoscalingConfig.PrometheusTLSInsecureSkipVerify),
+					},
+				},
+			},
```
pierDipi, line 299: We need to support authentication and authorization against the Prometheus instance

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -289,6 +303,29 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		b = b.Owns(&wvav1alpha1.VariantAutoscaling{}, builder.WithPredicates(childResourcesPredicate))
+	}
+	logger.Info("VariantAutoscaling CRD availability", "available", r.VariantAutoscalingAvailable)
+
+	// Install KEDA API and check availability
+	if err := kedav1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("failed to add KEDA APIs to scheme: %w", err)
+	}
+	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), kedav1alpha1.SchemeGroupVersion.String(), "ScaledObject"); ok && err == nil {
+		r.ScaledObjectAvailable = true
```
pierDipi: I suggest removing CRD gating from the reconciler setup (`ScaledObjectAvailable` and `VariantAutoscalingAvailable`). This allows the controller to recover gracefully if CRDs are deployed later, accepting that watches will simply initialize if the controller is restarted.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+func mainDeploymentName(llmSvc *v1alpha2.LLMInferenceService) string {
+	if llmSvc.Spec.Worker != nil {
+		return kmeta.ChildName(llmSvc.GetName(), "-kserve-mn")
+	}
+	return kmeta.ChildName(llmSvc.GetName(), "-kserve")
+}
+
+func prefillDeploymentName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return kmeta.ChildName(llmSvc.GetName(), "-kserve-prefill")
+}
```
pierDipi: nit: instead of adding this utils method here, can we add them to the workload_single_node file and refactor existing usages to use them?

Also multi node is not really supported in WVA, so don't need to handle multi node name here, are we validating that when autoscaling and multi node is configured the resource is rejected ?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+// The VA tells the WVA controller to compute wva_desired_replicas for this deployment.
+// podSpec is the pod template for the specific workload so accelerator detection is accurate.
+func (r *LLMISVCReconciler) reconcileVA(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.ScalingSpec, deploymentName, vaName string, podSpec *corev1.PodSpec) error {
+	isStopped := utils.GetForceStopRuntime(llmSvc)
+
+	if scaling == nil || scaling.WVA == nil || isStopped {
+		return r.deleteVAIfExists(ctx, llmSvc, vaName)
+	}
+
+	if !r.VariantAutoscalingAvailable {
```
pierDipi: see controller setup comment

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	return nil
+}
+
+// --- KEDA ScaledObject reconciliation ---
+
+// reconcileScaledObject creates, updates, or deletes a KEDA ScaledObject based on the scaling configuration.
+func (r *LLMISVCReconciler) reconcileScaledObject(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.ScalingSpec, config *Config, deploymentName, vaName, scaledObjectName string) error {
+	if !r.ScaledObjectAvailable {
```
pierDipi: see controller setup comment

pkg/controller/v1alpha2/llmisvc/scaling_test.go
pierDipi, line 1: This test is testing implementation details, I'd refactor it to be envtest based with a full reconciliation to also cover integration points with other pieces.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+		}
+		return r.reconcileScaledObject(ctx, llmSvc, scaling, config, deploymentName, vaName, scaledObjectName)
+	}
+
+	if scaling.WVA.HPA != nil {
+		if err := r.deleteScaledObjectIfExists(ctx, llmSvc, scaledObjectName); err != nil {
+			return err
+		}
+		return r.reconcileHPA(ctx, llmSvc, scaling, deploymentName, vaName, hpaName)
+	}
```
pierDipi: the logic assumes the 2 actuators, I think we need to structure it in a way that is extensible

```golang
func reconcileActuator( ... ) {
  reconcileScaledObject( ... )
  reconcileHPA ( ... )
}
```

and have individual functions handle deletions of their own objects 

```golang
func reconcileHPA( ... ) {

   if scaling == nil || scaling.WVA == nil || isStopped || scaling.WVA.HPA == nil {
     Delete( ... )
   }
   // ...
}

func reconcileKEDAScaledObject( ... ) {

   if scaling == nil || scaling.WVA == nil || isStopped || scaling.WVA.KEDA == nil {
     Delete( ... )
   }

   // ...
}
```

This follows existing patterns in the controller.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,493 @@
...
+							Name: wvaDesiredReplicasMetricName,
+							Selector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									variantNameLabelKey: vaName,
+								},
+							},
+						},
+						Target: autoscalingv2.MetricTarget{
+							Type:         autoscalingv2.AverageValueMetricType,
+							AverageValue: resource.NewQuantity(1, resource.DecimalSI),
```
pierDipi: don't we want Value rather than AverageValue?

cc @lionelvillard

asm582:
Reading the description of the PR, I am not sure if this is useful:

```
Auto-detected — if not set, the controller inspects extended resource requests/limits on the pod spec (e.g., nvidia.com/gpu → nvidia, amd.com/gpu → amd) to infer the vendor name.
```
Detecting the vendor name is completely different from detecting the vendor GPU, isn't it?

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: No. `ModelID` identifies the model being served (ex: "unsloth/Meta-Llama-3.1-8B"), not the LLMISVC instance. WVA groups VAs by `ModelID` to optimize across variants serving the same model on different hardware, so including namespace/name would break that grouping. Resource level uniqueness is already ensured by the VA's own metadata.name/namespace and the owner reference.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: @pierDipi @asm582 

I agree it is not a very sensible way to detect the accelerator. But the VA skips metrics emission if this label is not available. PTAL - https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/855.

I would strongly prefer to get the value using an API field (`acceleratorName`) directly from the user. So I have this as a temporary stop gap solution for the initial implementation to not leave the label empty. I will create an issue to make the changes to the API and remove this auto detection logic in the future.

vivekk16:
> Reading the description of the PR, I am not sure if this is useful:
> 
> ```
> Auto-detected — if not set, the controller inspects extended resource requests/limits on the pod spec (e.g., nvidia.com/gpu → nvidia, amd.com/gpu → amd) to infer the vendor name.
> ```
> 
> Detecting the vendor name is completely different from detecting the vendor GPU, isn't it?

Please refer https://github.com/kserve/kserve/pull/5194#discussion_r2902687043

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Good catch. `wva_desired_replicas` is an absolute replica count, not a per-pod load, so `Value` is the semantically correct target type. Will update.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Makes sense. Will update.

pkg/controller/v1alpha2/llmisvc/scaling.go
pierDipi: can we start by having a set of well-known accelerators that are known to work with llm-d?

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: We are currently not validating autoscaling x multi node configuration. I'll create a separate PR to update the changes.
I've updated the suggested nit.

cmd/llmisvc/main.go
```diff
@@ -169,6 +170,9 @@ func main() {
 				&corev1.Pod{}: {
 					Label: llmSvcCacheSelector,
 				},
+				&autoscalingv2.HorizontalPodAutoscaler{}: {
+					Label: llmSvcCacheSelector,
+				},
```
Gregory-Pereira, line 175: Should we call this something else like `AutoscalingActuator` or `Actuator`? I think that would make it more clear, because other actuator types are support aside from HPA

config/rbac/llmisvc/role.yaml
Gregory-Pereira, line 19: I know with autoscaling there are also layered configmaps that we have, do we need to add more operations for create, update, delete, etc.? My guess is yes

hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
Gregory-Pereira, line 1: If I am right above where we want RBAC to include edit and deletion of confimaps we need to add that too

charts/kserve-llmisvc-resources/files/llmisvc/resources.yaml
Gregory-Pereira, line 1455: Do we need to expand this? See comment below

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -97,21 +97,36 @@ func NewSchedulerConfig(isvcConfigMap *corev1.ConfigMap) (*SchedulerConfig, erro
...
+// AutoscalingConfig holds cluster-wide autoscaling settings loaded from the "autoscaling"
+// key in the inferenceservice-config ConfigMap. These are shared across all LLMISVC instances.
+type AutoscalingConfig struct {
+	// PrometheusURL is the URL of the Prometheus server (used by KEDA to query wva_desired_replicas).
+	PrometheusURL string `json:"prometheusURL"`
+	// PrometheusTLSInsecureSkipVerify disables TLS certificate verification for the Prometheus connection.
+	PrometheusTLSInsecureSkipVerify bool `json:"prometheusTLSInsecureSkipVerify"`
+}
+
+const autoscalingConfigName = "autoscaling"
```
Gregory-Pereira: How can we make this more descriptive, what about `autoscaling-controller-config`? Or something that more aureately depicts what this is:

```suggestion
const autoscalingConfigName = "autoscaling-controller-config"
```

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -107,6 +111,9 @@ type LLMISVCReconciler struct {
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch;update
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
```
Gregory-Pereira, line 109: Potentially update based on the above

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -264,6 +271,7 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(childResourcesPredicate)).
 		Owns(&corev1.Secret{}, builder.WithPredicates(childResourcesPredicate)).
 		Owns(&corev1.Service{}, builder.WithPredicates(childResourcesPredicate)).
+		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(childResourcesPredicate)).
```
Gregory-Pereira, line 271: Again if we change this struct name this would need to be updated. Not 100% certain we want to do that Im just calling it out

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -289,6 +297,25 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	logger.Info("InferencePool CRD availability", "v1", r.InferencePoolV1Available, "v1alpha2", r.InferencePoolV1Alpha2Available)
 
+	// Install WVA API and check availability
+	if err := wvav1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("failed to add WVA APIs to scheme: %w", err)
+	}
```
Gregory-Pereira: Is there a case where this fails where CRDs are already available? If so we want it to gracefully handle case where CRDs already exist

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -289,6 +297,25 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), wvav1alpha1.GroupVersion.String(), "VariantAutoscaling"); ok && err == nil {
+		b = b.Owns(&wvav1alpha1.VariantAutoscaling{}, builder.WithPredicates(childResourcesPredicate))
+	}
+
+	// Install KEDA API and check availability
+	if err := kedav1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("failed to add KEDA APIs to scheme: %w", err)
+	}
```
Gregory-Pereira: Same as above

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -289,6 +297,25 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if ok, err := utils.IsCrdAvailable(mgr.GetConfig(), kedav1alpha1.SchemeGroupVersion.String(), "ScaledObject"); ok && err == nil {
+		b = b.Owns(&kedav1alpha1.ScaledObject{}, builder.WithPredicates(childResourcesPredicate))
+	}
+
+	if err := lwsapi.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("failed to add LeaderWorkerSet APIs to scheme: %w", err)
+	}
```
Gregory-Pereira: How does LWS relate to autoscaling, shouldn't this be out of scope here? Or is this just throwing this change it (because it is needed just unrelated) while here?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,447 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	wvav1alpha1 "github.com/llm-d/llm-d-workload-variant-autoscaler/api/v1alpha1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
```
Gregory-Pereira, line 27: Do we need both packages? Is this just used for backwards compat with v1?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,447 @@
...
+			return nil
+		}
+		return fmt.Errorf("failed to reconcile prefill workload scaling: %w", err)
+	}
+
+	return nil
+}
+
+// reconcileMainWorkloadScaling handles scaling for the main (decode) workload.
+func (r *LLMISVCReconciler) reconcileMainWorkloadScaling(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, config *Config) error {
```
Gregory-Pereira, line 70: Is this a consistent naming pattern in kserve where decode = main but prefill = prefill? If not should we use prefill and decode?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,447 @@
...
+	return map[string]string{
+		constants.KubernetesComponentLabelKey: constants.LLMComponentWorkload,
+		constants.KubernetesAppNameLabelKey:   llmSvc.GetName(),
+		constants.KubernetesPartOfLabelKey:    constants.LLMInferenceServicePartOfValue,
+	}
+}
+
+func mainHPAName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return kmeta.ChildName(llmSvc.GetName(), "-kserve-hpa")
+}
```
Gregory-Pereira, line 454: again maybe something like mainActuatorName

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,447 @@
...
+
+func prefillHPAName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return kmeta.ChildName(llmSvc.GetName(), "-kserve-prefill-hpa")
+}
```
Gregory-Pereira, line 458: Same as above

Gregory-Pereira (commented):
I have some comments, most of it is naming conventions that I think can be more clear, but also some RBAC points. Will try to go through the scaling with a more fine tooth comb later

pkg/controller/v1alpha2/llmisvc/controller.go
vivekk16: Good catch! this `lwsapi.AddToScheme` call is redundant. It was removed in `ea21e9c6c `as part of centralizing scheme registration into pkg/scheme/register.go, and was accidentally re-introduced here, probably during a merge conflict resolution while rebasing. I'll remove it.

pkg/controller/v1alpha2/llmisvc/scaling.go
pierDipi: yes, "main" is the default instance that is used in non P/D and decode for P/D

pkg/controller/v1alpha2/llmisvc/config_loader.go
pierDipi: I would go further to use a CM key that is wva specific like `autoscaling-wva-controller-config`

```suggestion
const autoscalingConfigName = "autoscaling-wva-controller-config"
```

cmd/llmisvc/main.go
pierDipi: This is the resource watch for HPA resources, we have to use `HorizontalPodAutoscaler` here

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -97,21 +97,36 @@ func NewSchedulerConfig(isvcConfigMap *corev1.ConfigMap) (*SchedulerConfig, erro
...
 	// Storage and credential configs are excluded from JSON serialization
 	// as they contain sensitive information
 	StorageConfig    *types.StorageInitializerConfig `json:"-"`
 	CredentialConfig *credentials.CredentialConfig   `json:"-"`
 	SchedulerConfig  *SchedulerConfig                `json:"-"`
 }
 
+// AutoscalingConfig holds cluster-wide autoscaling settings loaded from the "autoscaling"
+// key in the inferenceservice-config ConfigMap. These are shared across all LLMISVC instances.
+type AutoscalingConfig struct {
```
pierDipi: we probably want to rename this to be WVA specific

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: `autoscalingv1` is used solely because `wvav1alpha1.VariantAutoscalingSpec.ScaleTargetRef `requires `autoscalingv1.CrossVersionObjectReference`, that's a type imposed by the WVA library. `autoscalingv2` is used for the HPA actuator itself, which needs the full v2 API for external metrics and scaling behavior configuration.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Validation updated - https://github.com/kserve/kserve/pull/5212

cmd/llmisvc/main.go
Gregory-Pereira: Then we can resolve this

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: WVA upstream will have `autoscalingv2` for `CrossVersionObjectReference` in the next release. We will update it alongside to stay consistent with upstream.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: These are just helper functions to name the resources. I will not try to over complicate this with a common parameterized function. Planning to leave this helper functions as it is.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Refer https://github.com/kserve/kserve/pull/5194/changes#r2918286078

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: The `inference.optimization/acceleratorName` label is resolved from `WorkloadSpec.Labels` rather than auto-detected. In expectedVA(), the value is propagated from the workload labels to the VariantAutoscaling resource, defaulting to `"unknown"` if absent or empty. Users must set this label explicitly in their workload spec.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+			},
+		},
+		Spec: autoscalingv2.HorizontalPodAutoscalerSpec{
+			ScaleTargetRef: autoscalingv2.CrossVersionObjectReference{
+				APIVersion: "apps/v1",
+				Kind:       "Deployment",
+				Name:       deploymentName,
+			},
+			MinReplicas: minReplicas,
+			MaxReplicas: *scaling.MaxReplicas,
```
spolti: this can lead to nilPointer as there is no nil check before.

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -165,5 +199,14 @@ func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, e
 		return nil, fmt.Errorf("failed to parse scheduler config: %w", errConvert)
 	}
 
-	return NewConfig(ingressConfig, storageInitializerConfig, &credentialConfig, schedulerConfig), nil
+	config := NewConfig(ingressConfig, storageInitializerConfig, &credentialConfig, schedulerConfig)
+
+	if autoscalingData, ok := isvcConfigMap.Data[autoscalingConfigName]; ok {
+		asCfg := &WVAAutoscalingConfig{}
+		if err := json.Unmarshal([]byte(autoscalingData), asCfg); err == nil {
```
spolti: error is being masked.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+				"Scaling disabled: required CRD not found: %v", err)
+			return nil
+		}
+		return fmt.Errorf("failed to reconcile main workload scaling: %w", err)
+	}
+
+	if err := r.reconcilePrefillWorkloadScaling(ctx, llmSvc, config); err != nil {
+		if meta.IsNoMatchError(err) {
+			r.Eventf(llmSvc, corev1.EventTypeWarning, "ScalingCRDNotFound",
+				"Scaling disabled: required CRD not found: %v", err)
```
spolti: The two events are duplicated, what about emit the event just once?
If reconcileMainWorkloadScaling fails with no CRD, here it will fail as well.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+			Name:      vaName,
+			Namespace: llmSvc.GetNamespace(),
+		},
+	}
+	return Delete(ctx, r, llmSvc, va)
+}
+
+// expectedVA constructs the desired VariantAutoscaling resource from the LLMISVC spec.
+// workloadLabels are the labels from the WorkloadSpec for the specific workload (decode or prefill).
+// If the workload labels contain inference.optimization/acceleratorName, that value is propagated
```
spolti, line 364: who is in charge of setting `inference.optimization/acceleratorName`?

Where is the auto-detection logic as described in the PR description?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+func (r *LLMISVCReconciler) reconcileScaling(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, config *Config) error {
+	logger := log.FromContext(ctx).WithName("reconcileScaling")
+	ctx = log.IntoContext(ctx, logger)
+
+	if err := r.reconcileMainWorkloadScaling(ctx, llmSvc, config); err != nil {
+		if meta.IsNoMatchError(err) {
+			r.Eventf(llmSvc, corev1.EventTypeWarning, "ScalingCRDNotFound",
+				"Scaling disabled: required CRD not found: %v", err)
+			return nil
+		}
```
pierDipi: I don't think we want to silently continue for misconfigurations
```suggestion
```

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+		return fmt.Errorf("failed to reconcile main workload scaling: %w", err)
+	}
+
+	if err := r.reconcilePrefillWorkloadScaling(ctx, llmSvc, config); err != nil {
+		if meta.IsNoMatchError(err) {
+			r.Eventf(llmSvc, corev1.EventTypeWarning, "ScalingCRDNotFound",
+				"Scaling disabled: required CRD not found: %v", err)
+			return nil
+		}
```
pierDipi:
```suggestion
```

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+	wvaDesiredReplicasMetricName = "wva_desired_replicas"
+	variantNameLabelKey          = "variant_name"
+	acceleratorNameLabelKey      = "inference.optimization/acceleratorName"
+)
+
+// reconcileScaling manages the autoscaling resources (VariantAutoscaling + HPA or KEDA ScaledObject)
+// for the LLM workload. When scaling is configured, it creates a VariantAutoscaling CR for WVA to
+// compute desired replicas and an actuator (HPA or KEDA ScaledObject) to enforce them.
+// When scaling is removed (or the workload is stopped), it cleans up any existing autoscaling resources.
+func (r *LLMISVCReconciler) reconcileScaling(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, config *Config) error {
```
spolti, line 54: WVA does not support multi-node deployments, right? What happens if the user set auto-scaling with multi-node feature? I didn't find a validation to prevent this config.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+func expectedScaledObject(llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.ScalingSpec, config *Config, deploymentName, vaName, scaledObjectName string) *kedav1alpha1.ScaledObject {
+	labels := scalingLabels(llmSvc)
+	keda := scaling.WVA.KEDA
+
+	minReplicas := ptr.To(int32(1))
+	if scaling.MinReplicas != nil {
+		minReplicas = scaling.MinReplicas
+	}
+
+	query := fmt.Sprintf(`wva_desired_replicas{variant_name="%s",exported_namespace="%s"}`, vaName, llmSvc.GetNamespace())
```
spolti, line 272: Is it safe to direct inject the envs in the PromQL query?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -0,0 +1,436 @@
...
+		},
+	}
+
+	return va
+}
+
+func semanticVAIsEqual(expected, curr *wvav1alpha1.VariantAutoscaling) bool {
+	return equality.Semantic.DeepDerivative(expected.Spec, curr.Spec) &&
+		equality.Semantic.DeepDerivative(expected.Labels, curr.Labels) &&
+		equality.Semantic.DeepDerivative(expected.Annotations, curr.Annotations)
```
spolti: do you know if DeepDerivative could potentially ignore nil or removed fields?

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Auto detection was not preferred. So the current logic follows this - https://github.com/kserve/kserve/pull/5194#discussion_r2918451482.  I've update the PR description now.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Currently, the controller emits an event and proceeds with deployment if the autoscaling CRD is missing. Should we transition to a hard dependency where a missing scaling CRD results in a reconciliation error, effectively blocking the LLMInferenceService deployment?

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: The validation is part of this PR - https://github.com/kserve/kserve/pull/5212

pkg/controller/v1alpha2/llmisvc/scaling.go
pierDipi: yes, llmisvc is misconfigured so we need to report and block reconciliation of other resources

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: The `vaName` is safe here. It's fully constructed by the controller itself via `kmeta.ChildName(llmSvc.GetName(), "-kserve-va")`, where the suffix is a hardcoded string and `llmSvc.GetName()` is a k8s resource name validated by the API server at admission time to be a DNS subdomain (only lowercase alphanumeric characters and hyphens). The same applies to `llmSvc.GetNamespace()`. Neither can ever contain PromQL-special characters like `", \, {, }, or =`, so the direct interpolation is safe.

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -97,21 +97,55 @@ func NewSchedulerConfig(isvcConfigMap *corev1.ConfigMap) (*SchedulerConfig, erro
...
 	CredentialConfig *credentials.CredentialConfig   `json:"-"`
 	SchedulerConfig  *SchedulerConfig                `json:"-"`
 }
 
+// WVAAutoscalingConfig holds cluster-wide WVA autoscaling settings loaded from the
+// "autoscaling-wva-controller-config" key in the inferenceservice-config ConfigMap.
+// These are shared across all LLMISVC instances.
+type WVAAutoscalingConfig struct {
+	// PrometheusURL is the URL of the Prometheus server (used by KEDA to query wva_desired_replicas).
+	PrometheusURL string `json:"prometheusURL"`
```
yuzisun: could we add PrometheusConfig struct so we do not need to repeat the names in all the fields?

config/rbac/llmisvc/role.yaml
vivekk16: Resolving this conversation as it is out of scope for this initial implementation.

hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
vivekk16: Resolving this conversation as it is out of scope for this initial implementation.

charts/kserve-llmisvc-resources/files/llmisvc/resources.yaml
vivekk16: Resolving this conversation as it is out of scope for this initial implementation.

pkg/controller/v1alpha2/llmisvc/config_loader.go
vivekk16: Sure. Will update.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Switched all three scaling equality functions to DeepEqual, which detects field removals. To avoid reconciliation loops from controller-injected metadata, added an AfterDryRun hook for KEDA ScaledObjects.

pkg/controller/v1alpha2/llmisvc/controller.go
vivekk16: Resolving since this not relevant for initial implementation.

pkg/controller/v1alpha2/llmisvc/controller.go
vivekk16: Resolving since this is not relevant to initial implementation.

pkg/controller/v1alpha2/llmisvc/scaling_test.go
vivekk16: Will be addressed in a separate PR.

cmd/llmisvc/main.go
KillianGolds, line 212: ### 1. Possible missing `os.Exit(1)` after controller setup failure

In [`cmd/llmisvc/main.go:210-212`](https://github.com/kserve/kserve/pull/5194/files#diff-cmd/llmisvc/main.go):

```go
}).SetupWithManager(mgr); err != nil {
    setupLog.Error(err, "unable to create controller", "controller", "LLMInferenceService")
}
```

It looks like the other `SetupWithManager` / webhook setup blocks in this file all call `os.Exit(1)` on error (e.g. [`main.go:218-220`](https://github.com/kserve/kserve/pull/5194/files#diff-cmd/llmisvc/main.go) and [`main.go:227-229`](https://github.com/kserve/kserve/pull/5194/files#diff-cmd/llmisvc/main.go)), but this one only logs and continues. If I'm reading it right, this would mean the manager starts up without the core LLMInferenceService controller registered - so it'd accept traffic but silently not reconcile anything. Could be a subtle one to debug in production. I think just adding `os.Exit(1)` after the log line would do it, but wanted to check if there was an intentional reason for leaving it out.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
KillianGolds, line 434: ### 2. `IdleReplicaCount` validation might make the field unusable

In [`pkg/apis/serving/v1alpha2/llm_inference_service_types.go:434`](https://github.com/kserve/kserve/pull/5194/files#diff-pkg/apis/serving/v1alpha2/llm_inference_service_types.go):
```go
// +kubebuilder:validation:Minimum=1
IdleReplicaCount *int32 `json:"idleReplicaCount,omitempty"`
```

combined with the XValidation rule at [`llm_inference_service_types.go:327`](https://github.com/kserve/kserve/pull/5194/files#diff-pkg/apis/serving/v1alpha2/llm_inference_service_types.go):
```
rule="...self.wva.keda.idleReplicaCount < self.minReplicas"
```

and `minReplicas` at [`llm_inference_service_types.go:334-336`](https://github.com/kserve/kserve/pull/5194/files#diff-pkg/apis/serving/v1alpha2/llm_inference_service_types.go) having:
```go
// +kubebuilder:default=1
// +kubebuilder:validation:Minimum=1
MinReplicas *int32 `json:"minReplicas,omitempty"`
```

Unless I'm misreading the CEL evaluation order, this creates a situation where no valid value can be set - `idleReplicaCount` must be `>= 1` (from its own minimum) but also `< minReplicas` (which is itself `>= 1`). So even with `minReplicas=1` (the default), the only value that satisfies `Minimum=1` is `1`, but that fails `< 1`. And with `minReplicas=2`, you could set `idleReplicaCount=1`… but then you can't do scale-to-zero, which i beleive is one of the main reasons to pick KEDA over HPA.

I think changing `Minimum=1` to `Minimum=0` on `IdleReplicaCount` would fix it and unlock KEDA's scale-to-zero capability. But again, I might be missing something about how the defaulting interacts with CEL here - again could be missing context myself so disregard/resolve if im off base!

KillianGolds (commented):
Nice work @vivekk16, the API design with mutual exclusivity between HPA and KEDA is clean, the owner references and cleanup logic look well structured, and the test coverage on the construction helpers is thorough.

I don't have full context on all the design decisions here so it's entirely possible I'm misreading something - just wanted to flag the two items below that jumped out at me in case they weren't intentional. Totally happy to be corrected on either if there's context I'm missing as its been a long day haha!

Thanks for putting this together - happy to take another look whenever.

cmd/llmisvc/main.go
vivekk16: Thanks for flagging this! 
I'll leave it as-is in this PR to keep the scope focused, and open a separate issue to track adding the os.Exit(1).

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: Scale-to-zero is not supported as part of this initial implementation. This validation was intentionally added keeping that in mind.

cmd/llmisvc/main.go
vivekk16: PR for this - https://github.com/kserve/kserve/pull/5242

Gregory-Pereira (commented):
/lgtm

pierDipi (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5201, opened 2026-03-09
fix(llmisvc): handle missing LWS CRD gracefully

bartoszmajsak:

When the LeaderWorkerSet CRD is not installed on the cluster, `r.Get()` returns a `meta.NoKindMatchError`, which is distinct from `apierrors.IsNotFound`. Previously, only `IsNotFound` was checked, causing reconciliation to fail with "no matches for kind LeaderWorkerSet" even for single-node deployments.

Treat `meta.IsNoMatchError` the same as `IsNotFound` in the Get calls within `expectedMainMultiNodeLWS` and `expectedPrefillMultiNodeLWS` so the reconciler can proceed gracefully when the CRD is absent.

Port of opendatahub-io/kserve#1158

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm
/approve

spolti (approved):
/lgtm

PR #5202, opened 2026-03-09
feat(llmisvc): evaluate both InferencePool readiness

bartoszmajsak:

During the `v1alpha2` to `v1` `InferencePool` migration, the controller creates both pool versions. However, `EvaluateInferencePoolConditions` only checked the v1 pool - if the gateway only accepts `v1alpha2`, this reports "not ready" even though the `v1alpha2` pool IS ready.

For managed pools (not referenced), check both v1 and v1alpha2 InferencePool readiness and consider the pool ready if at least one version is accepted by the gateway. This prevents false "not ready" status during migration.

Uses typed clients for both pool versions (upstream already has igwapi v1 and igwapiv1alpha2 typed imports).

Additionally, rework `IsInferencePoolReady` (`v1` and `v1alpha2`) to no longer treat pools with empty `Status.Parents` as ready. A pool without any gateway claiming it is not ready for traffic - the `Owns()` watch ensures re-reconciliation when a gateway populates the status, and the `WaitingForGateway` reason provides clear observability.

Port of opendatahub-io/kserve#1099, opendatahub-io/kserve#1090

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router.go
+++ b/pkg/controller/v1alpha2/llmisvc/router.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	"knative.dev/pkg/network"
@@ -35,6 +36,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/kmeta"
 	igwapi "sigs.k8s.io/gateway-api-inference-extension/api/v1"
+	igwapiv1alpha2 "sigs.k8s.io/gateway-api-inference-extension/apix/v1alpha2"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -505,7 +507,9 @@ func (r *LLMISVCReconciler) EvaluateHTTPRouteConditions(ctx context.Context, llm
 }
 
 // EvaluateInferencePoolConditions evaluates the readiness of all Inference Pools in the LLMInferenceService
-// and updates the InferencePoolReady condition accordingly
+// and updates the InferencePoolReady condition accordingly.
+// During the v1alpha2 to v1 migration, it checks both pool versions for managed pools
+// and considers the pool ready if at least one is ready.
 func (r *LLMISVCReconciler) EvaluateInferencePoolConditions(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
 	logger := log.FromContext(ctx).WithName("EvaluateInferencePoolConditions")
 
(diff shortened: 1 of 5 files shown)


pkg/controller/v1alpha2/llmisvc/router.go
KillianGolds, line 518: ### `pkg/controller/v1alpha2/llmisvc/router.go`

**Small nit: logger re-creation in extracted method**

I noticed the parent `EvaluateInferencePoolConditions` creates a logger at [line 520](https://github.com/kserve/kserve/pull/5202/files#diff-router.go-L520), but the extracted `evaluateSingleInferencePoolCondition` re-derives it from context at [line 648](https://github.com/kserve/kserve/pull/5202/files#diff-router.go-L648):

```go
log.FromContext(ctx).WithName("EvaluateInferencePoolConditions").V(2).Info("Inference Pool is ready", "pool", curr)
```

Totally minor - might be slightly cleaner to pass the logger through or skip re-adding the name, but not a big deal at all.

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -523,27 +527,104 @@ func (r *LLMISVCReconciler) EvaluateInferencePoolConditions(ctx context.Context,
 		return nil
 	}
 
-	curr := &igwapi.InferencePool{}
-
+	// For referenced pools (external), only check that pool
```
KillianGolds, line 528: ### `pkg/controller/v1alpha2/llmisvc/router.go` - thought on test coverage

The new dual-version orchestration logic in `EvaluateInferencePoolConditions` ([lines 530–625](https://github.com/kserve/kserve/pull/5202/files#diff-router.go-L530-L625)) handles quite a few scenarios:

- v1 ready, v1alpha2 not (and vice versa)
- Both fetched but neither ready (prefer v1 for status reporting)
- v1alpha2 CRD not installed (`meta.IsNoMatchError`)
- Both not found vs transient error distinction
- `WaitingForGateway` when `Status.Parents` is empty

The unit tests for `IsInferencePoolReady`, `IsInferencePoolV1Alpha2Ready`, and `IsInferencePoolV1Supported` are thorough - just wondering if it might be worth adding a table-driven test for `EvaluateInferencePoolConditions` itself with a fake client, to cover the branching/fallback logic end-to-end? Might already be planned or covered elsewhere though, so feel free to ignore.

KillianGolds (commented):
### Overall

Really nice changes - cleanly solves a real problem during the v1alpha2 → v1 migration. The readiness functions and fixture refactoring look solid. Just a couple of minor thoughts below, happy to be corrected if I'm missing context.

pkg/controller/v1alpha2/llmisvc/router.go
bartoszmajsak: Good for follow-up I guess?

pkg/controller/v1alpha2/llmisvc/router.go
bartoszmajsak: Addressed in the amended commit.

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -523,27 +527,104 @@ func (r *LLMISVCReconciler) EvaluateInferencePoolConditions(ctx context.Context,
...
+			))
+		} else {
+			llmSvc.MarkInferencePoolNotReady("InferencePoolNotReady", fmt.Sprintf("The inference pool %s/%s is not ready", poolNamespace, poolName))
+		}
+		return nil
 	}
 
+	// Both failed to fetch - distinguish between "not found" and transient errors
+	if (apierrors.IsNotFound(v1Err) || meta.IsNoMatchError(v1Err)) &&
+		(apierrors.IsNotFound(v1alpha2Err) || meta.IsNoMatchError(v1alpha2Err) || !v1alpha2Available) {
```
spolti: isn't `meta.IsNoMatchError(v1alpha2Err) || !v1alpha2Available` essentially the same thing?
since we have v1alpha2Available := v1alpha2Err == nil || !meta.IsNoMatchError(v1alpha2Err)

bartoszmajsak: Good catch, thank you! Fixed in amended commit

pierDipi (approved):
/lgtm

PR #5203, opened 2026-03-09
fix(keda): delete legacy KServe-owned HPA before creating ScaledObject

### LINKED ISSUE #5196: [enhance] Switching InferenceService autoscalerClass from `hpa` to `keda` causes KEDA webhook error
/kind feature

## Summary

When a `Standard`(`RawDeployment`) mode `InferenceService` has its `autoscalerClass` annotation changed directly
from `hpa` to `keda` (without stopping the service first), the controller fails to reconcile with
the following error:

```
admission webhook "vscaledobject.kb.io" denied the request:
the workload 'keda-standard-predictor' of type 'apps/v1.Deployment'
is already managed by the hpa 'keda-standard-predictor'
```

The service is stuck in a reconcile error loop until the conflicting HPA is manually deleted.

## Steps to Reproduce

1. Create an `InferenceService` with `RawDeployment` mode using the default HPA autoscaler:
   ```yaml
   metadata:
     annotations:
       serving.kserve.io/deploymentMode: Standard
       # serving.kserve.io/autoscalerClass: hpa  ← default
   ```
2. Verify the HPA resource `<isvc-name>-predictor` is created.
3. Update the annotation **without stopping the service**:
   ```yaml
   metadata:
     annotations:
       serving.kserve.io/autoscalerClass: keda
   ```
4. Observe the controller log — reconcile fails immediately.

## Root Cause

When `autoscalerClass` is changed to `keda`, the `autoscaler_reconciler.go` creates a
`KedaReconciler` instead of an `HPAReconciler`. The old `HPAReconciler` is no longer invoked,
so the pre-existing HPA resource is never cleaned up.

On the next reconcile, `KedaReconciler.Reconcile()` attempts to `Create` a `ScaledObject`
targeting the same `Deployment`. KEDA's admission webhook (`vscaledobject.kb.io`) rejects this
because the `Deployment` is already managed by the legacy HPA.

Relevant call path:

```
InferenceServiceReconciler.Reconcile()
  → Predictor.reconcileRawDeployment()   [predictor.go:734]
    → RawKubeReconciler.Reconcile()
      → KedaReconciler.Reconcile()
        → client.Create(ScaledObject)    ← webhook denies here
```

## Workaround

Manually stop the isvc first before modification

## Proposed Fix

Before creating or updating the `ScaledObject`, `KedaReconciler.Reconcile()` should check for a
legacy KServe-owned HPA with the same name and delete it. Only HPAs with an `InferenceService`
owner reference are touched to avoid affecting user-managed resources.

fyuan1316:
Prevents KEDA webhook conflict when switching autoscalerClass from hpa to keda on a live InferenceService

Related to the issue I filed above. Since there's been no discussion yet, I went ahead and implemented a fix based on my understanding of the root cause — happy to iterate on the approach if the community has other thoughts.

What this does: 
Adds `cleanupLegacyHPA()` in KedaReconciler.Reconcile() to delete any KServe-owned HPA with the same name before creating the `ScaledObject`. This prevents the KEDA admission webhook from rejecting the new `ScaledObject` when autoscalerClass is switched from hpa to keda on a live InferenceService.

Fixes #5196 

Added unit tests in [keda_reconciler_test.go](cci:7://file:///Users/yuan/Dev/alauda/aml/kserve.io/kserve/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler_test.go:0:0-0:0) covering the new [cleanupLegacyHPA](cci:1://file:///Users/yuan/Dev/alauda/aml/kserve.io/kserve/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go:284:0-316:1) logic:
- `TestCleanupLegacyHPA/no_HPA_exists`: no-op when there is no pre-existing HPA
- `TestCleanupLegacyHPA/HPA_exists_but_not_owned_by_KServe`: HPA created by the user (no `InferenceService` owner reference) is not touched
- `TestCleanupLegacyHPA/HPA_exists_and_is_KServe-owned`: a KServe-managed HPA is deleted before the ScaledObject is created
- [TestReconcile_MigrateFromHPAToKeda](cci:1://file:///Users/yuan/Dev/alauda/aml/kserve.io/kserve/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler_test.go:1101:0-1143:1): end-to-end migration path — [Reconcile()](cci:1://file:///Users/yuan/Dev/alauda/aml/kserve.io/kserve/pkg/controller/v1beta1/inferenceservice/components/predictor.go:82:0-258:1) with a pre-existing KServe HPA succeeds, the HPA is deleted, and the ScaledObject is created
All tests pass:
```text
go test ./pkg/controller/v1beta1/inferenceservice/reconcilers/keda/... -run 'TestCleanupLegacyHPA|TestReconcile_MigrateFromHPAToKeda' -v
--- PASS: TestCleanupLegacyHPA (0.00s)
--- PASS: TestReconcile_MigrateFromHPAToKeda (0.00s)
```

- Logs

**Checklist**:

**Release note**:

NONE

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -281,9 +282,50 @@ func semanticScaledObjectEquals(desired, existing *kedav1alpha1.ScaledObject) bo
...
+// never be cleaned up. The old HPA must be removed before KEDA's admission webhook will accept the
+// new ScaledObject targeting the same Deployment.
+func (r *KedaReconciler) cleanupLegacyHPA(ctx context.Context, name, namespace string) error {
+	existingHPA := &autoscalingv2.HorizontalPodAutoscaler{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, existingHPA)
+	if apierr.IsNotFound(err) {
+		return nil // nothing to clean up
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get legacy HPA %s/%s: %w", namespace, name, err)
```
spolti, line 297: isn't better to just print a warning and return no error?

fyuan1316: I’d suggest keeping the return err here.

In the controller-runtime, returning an error ensures the requeue/retry mechanism kicks in. If we return nil but the HPA wasn't actually cleaned up (due to a transient API error), the subsequent KEDA ScaledObject creation will likely be blocked by its admission webhook, potentially leaving the migration in a 'stuck' state without any automatic recovery.

PR #5208, opened 2026-03-09
feat(llmisvc): add Prometheus metrics collection for LLMInferenceService

israel-hdez:

The controller now creates and manages per-namespace Prometheus Operator resources to collect metrics from LLMInferenceService components:

- A PodMonitor that scrapes vLLM engine pods over HTTPS, relabeling metrics with a kserve_ prefix and pod-level labels (isvc name, role, component).
- A ServiceMonitor that scrapes the llm-d scheduler's /metrics endpoint, using bearer-token authorization via a dedicated ServiceAccount and Secret.
- A ClusterRoleBinding that grants the metrics-reader ServiceAccount access to the kserve-metrics-reader-cluster-role ClusterRole.

Monitoring resources are shared across all LLMInferenceService instances in a namespace and are cleaned up only when the last one is deleted. Creation can be globally disabled via the LLMISVC_MONITORING_DISABLED environment variable, which is useful on clusters without the Prometheus Operator installed.

Fixes #

- [ ] Test A
- [ ] Test B

- Logs

**Checklist**:

**Release note**:
```release-note
Metrics collection of LLMInferenceServices using Prometheus Operator is implemented.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/monitoring.go
+++ b/pkg/controller/v1alpha2/llmisvc/monitoring.go
@@ -1,0 +1,417 @@
+/*
+Copyright 2025 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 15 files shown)


pkg/controller/v1alpha2/llmisvc/monitoring.go
```diff
@@ -0,0 +1,417 @@
...
+							"llminferenceservice-workload-leader-prefill",
+							"llminferenceservice-workload-worker-prefill",
+						},
+					},
+				},
+			},
+			PodMetricsEndpoints: []monitoringv1.PodMetricsEndpoint{
+				{
+					TargetPort: &metricsPort,
+					Scheme:     "https",
```
israel-hdez, line 239: Hmm... Likely, we don't have HTTPS.

PR #5212, opened 2026-03-10
fix(llmisvc): add validations for unsupported scaling configurations

### LINKED ISSUE #5211: add webhook and CEL validation for unsupported scaling configurations in LLMInferenceService
## Problem

Two invalid `LLMInferenceService` scaling configurations were silently accepted by the API server with no error, leading to confusing runtime behaviour:

1. **Autoscaling + multi-node**: When `spec.workloadSpec.scaling` is set alongside `spec.workloadSpec.worker`, the controller creates a `LeaderWorkerSet` (LWS) but WVA only knows how to reconcile against a `Deployment`. The `scaling` block is silently ignored and the deployment gets no autoscaling, with no indication to the user.

2. **Mismatched prefill/decode actuator backends**: When a disaggregated deployment configures HPA for decode and KEDA for prefill (or vice versa), both actuators operate independently with separate metric pipelines, producing uncoordinated scaling decisions across the two sides.

vivekk16:
Closes #5211 

## Summary

Two invalid `LLMInferenceService` scaling configurations were silently accepted by the API server with no error, leading to confusing and undetectable runtime behaviour:

1. **Autoscaling + multi-node**: Setting `scaling` alongside `worker` causes the controller to create a `LeaderWorkerSet`, but WVA only reconciles against a `Deployment`. The `scaling` block is silently ignored — the deployment gets no autoscaling with no indication to the user.

2. **Mismatched prefill/decode actuator backends**: Configuring HPA for decode and KEDA for prefill (or vice versa) creates two independent metric pipelines that produce uncoordinated scaling decisions across the disaggregated deployment.

## Changes

- **`ValidateWorkloadScaling`** (v1alpha2 webhook): Rejects any `WorkloadSpec` that sets both `scaling` and `worker`, with an error message directing the user to use `replicas` for a fixed count.
- **CEL `XValidation` marker** on `WorkloadSpec`: Enforces the same `worker + scaling` constraint at the CRD level so the API server catches it before the webhook runs.
- **`ValidateActuatorConsistency`** (exported, v1alpha2): Rejects configurations where decode and prefill use different actuator backends (HPA vs KEDA), explaining that this forces two separate metric pipelines and leads to unsynchronised scaling decisions.
- **v1alpha1 coverage**: `validateScaling` in v1alpha1 now delegates to `ValidateActuatorConsistency` via type conversion, so both API versions are protected.
- **CRD YAMLs regenerated** via `make manifests` to include the new CEL rule across all CRD variants and the Helm chart templates.
- **Tests added/updated** in both v1alpha1 and v1alpha2 covering all new valid and invalid scenarios.

## Notes

LWS *can* technically be targeted by an HPA directly, but WVA currently only supports `Deployment` as a scaling target. `TODO` comments mark both the CEL marker and the webhook guard for removal once WVA gains LWS support.

### DIFF
--- a/charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
+++ b/charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
@@ -7936,6 +7936,10 @@ spec:
                 - message: replicas and scaling are mutually exclusive; use scaling
                     for autoscaled deployments or replicas for static deployments
                   rule: '!(has(self.replicas) && has(self.scaling))'
+                - message: autoscaling (scaling) is not supported for multi-node deployments
+                    (worker is set); remove scaling and use replicas instead to set
+                    a fixed replica count
+                  rule: '!(has(self.worker) && has(self.scaling))'
               replicas:
                 format: int32
                 minimum: 0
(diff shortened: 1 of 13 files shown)


pierDipi (approved):
/lgtm
/approve

Jooho (approved):
/lgtm
/approved

Jooho (approved):
/lgtm
/approved

PR #5214, opened 2026-03-10
feat: include GIE CRDs as part of Kserve bundle

pierDipi:
we need Kserve to control the version of such CRDs putting it as prerequisite means that there might conflicts on API versions


Jooho (approved):
/lgtm
/approved

PR #5217, opened 2026-03-10
feat(llmisvc): add build hooks for distribution-specific logic

bartoszmajsak:

Adds compile-time hook points to the LLMInferenceService controller using Go build tags (`distro`/`!distro`), following the pattern used by projects like HashiCorp Vault CE/Enterprise.

Upstream gets zero-cost no-op stubs. Distributions can provide real implementations behind the `distro` tag without forking core reconciliation files - reducing merge conflicts and making upstream rebases clean.

Hook points:
- **`ensureGatewayPreconditions`** - guards HTTPRoute reconciliation; returns `ErrPreconditionNotMet` for non-retryable failures (e.g. missing CRD)
- **`reconcileWorkloadPlatformPermissions`** - platform-specific workload permissions before pod creation (e.g. SecurityContextConstraints on OpenShift)
- **`additionalEnvTestOptions`** - test fixture extensibility for distribution-specific CRDs

Also adds:
- `GOTAGS` variable to Makefile and Dockerfile so distributions can activate build tags through the standard build pipeline (`make docker-build-llmisvc GOTAGS=distro`)
- `rest.Config` field on the reconciler (needed for runtime CRD availability checks)

- [x] Upstream builds compile and run tests without the `distro` tag (no-op stubs)
- [x] Distribution builds compile with `-tags distro` and real implementations
- [x] Verified no behavioral change for upstream - all stubs return nil

This pattern is already used in production by distributions like OpenDataHub/RHOAI (see [opendatahub-io/kserve#1157](https://github.com/opendatahub-io/kserve/pull/1157)) for SCC role bindings and Kuadrant AuthPolicy preconditions. Upstreaming the hook infrastructure means distributions can maintain only their `_distro.go` files in `Makefile.overrides.mk` - the Makefile and core controller files stay identical to upstream.

**Checklist**:

```release-note
Add compile-time hook points to LLMInferenceService controller for distribution-specific logic using Go build tags. Distributions can now inject platform-specific behavior (gateway preconditions, workload permissions) without forking core controller files.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,13 @@ $(shell perl -pi -e 's/memory:.*/memory: $(KSERVE_CONTROLLER_MEMORY_LIMIT)/' con
 
 export GOFLAGS=-mod=mod
 
+# Go build tags (e.g. "distro" for distribution-specific code).
+# Passed to Docker image builds via --build-arg and to all go commands via GOFLAGS.
+GOTAGS ?=
+ifdef GOTAGS
+export GOFLAGS += -tags=$(GOTAGS)
+endif
+
 all: test manager agent router
 
 .PHONY: setup-envtest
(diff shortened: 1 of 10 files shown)


Jooho (approved):
/lgtm
/approved

PR #5219, opened 2026-03-10
fix: use defensive parameter expansion for KO_DOCKER_REPO

bartoszmajsak:

Prevents 'unbound variable' errors when `kserve-images.sh` is sourced by scripts that use 'set -o nounset' without KO_DOCKER_REPO being set.

This pattern is already used in opendatahub-io/kserve where CI scripts source kserve-images.sh under strict shell settings. Adopting it upstream makes the script safe by default and eliminates a recurring carry patch during rebases.

**Release note**:
```release-note
NONE
```


bartoszmajsak:
Probably irrelevant.

PR #5220, opened 2026-03-10
ci: improvements for setup-uv and build-graph image

bartoszmajsak:

- Add `--clear` flag to uv venv to ensure a clean virtual environment even when a stale .venv directory already exists
- Add `BUILDER` variable to support podman as an alternative to docker, with automatic detection when only podman is available

Both changes align upstream with opendatahub-io/kserve CI practices. The --clear flag avoids flaky CI runs caused by leftover venvs, and podman support enables building images in environments where docker is not installed - a common setup in Red Hat CI infrastructure.

**Release note**:
```release-note
NONE
```


Jooho (approved):
/lgtm
/approved

spolti (approved):
/lgtm

PR #5237, opened 2026-03-12
fix(llmisvc): add default for MinReplicas & make MaxReplicas required int32

vivekk16:
## Summary

- **MaxReplicas** was typed as `*int32` (optional pointer) despite being semantically required whenever `ScalingSpec` is present. This caused a mismatch between the CRD schema (which showed it as optional) and the actual validation (which rejected requests without it via an XValidation rule). Changed to `int32` so the API server enforces it as a required field at the schema level.

- **MinReplicas** documented "defaults to 1" but lacked a `+kubebuilder:default=1` marker, so the default only lived in controller code and was invisible in the CRD schema and `kubectl explain` output. Added the marker so the API server injects the default into stored objects.

## Changes

- Change `MaxReplicas` from `*int32` to `int32` in `ScalingSpec` (v1alpha1 and v1alpha2)
- Add `+kubebuilder:default=1` to `MinReplicas` (v1alpha1 and v1alpha2)
- Remove redundant `has(self.maxReplicas)` XValidation rule
- Simplify `minReplicas <= maxReplicas` XValidation rule (drop `!has(self.maxReplicas)` guard)
- Update Go validation code to reflect non-pointer `MaxReplicas`
- Update deepcopy to remove `MaxReplicas` pointer copy logic
- Update validation and conversion tests in both API versions

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./pkg/apis/serving/v1alpha1/... ./pkg/apis/serving/v1alpha2/...` passes
- [x] CRD schema shows `maxReplicas` in `required` list and `minReplicas` has `default: 1`

### DIFF
--- a/charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
+++ b/charts/kserve-llmisvc-crd-minimal/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
@@ -146,6 +146,7 @@ spec:
                         minimum: 1
                         type: integer
                       minReplicas:
+                        default: 1
                         format: int32
                         minimum: 1
                         type: integer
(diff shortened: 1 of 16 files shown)


pierDipi (approved):
/lgtm
/approve

pierDipi:
```

# [github.com/kserve/kserve/pkg/apis/serving/v1alpha2]
Error: vet: pkg/apis/serving/v1alpha2/llm_inference_service_validation_test.go:631:19: cannot use ptr.To(int32(5)) (value of type *int32) as int32 value in struct literal
```

Jooho (approved):
/lgtm
/approved

PR #5238, opened 2026-03-12
feat: allow configuring GatewayClass for llmisvc e2e tests

pierDipi:
Allow configuring GW class in e2e test with GATEWAY_CLASS_NAME env variable


Jooho (approved):
/lgtm
/approved

PR #5240, opened 2026-03-12
feat(llmisvc): add platform hooks for networking and manager cfg

bartoszmajsak:

Follow-up to #5217. Adds more build-tag extension points so distribution builds can plug in platform-specific behavior without touching upstream files.

New hooks:
- `createWorkloadCertificate` - lets distributions provide CA-signed TLS certs instead of self-signed ones, using a `certBundle` struct that cleanly separates key, cert, and optional CA material
- `reconcileRouterPlatformNetworking` - platform-specific networking setup after HTTP route reconciliation
- `extendControllerSetup` - platform-specific scheme registration and ownership watches during controller init
- `customizeManagerOptions` - distributions can extend the manager's cache configuration without modifying `main.go`

Also adds a `manifests-distro` Makefile hook for distribution-specific manifest generation.

Each hook follows the established `_default.go` (`//go:build !distro`) companion file pattern. Upstream files are no-ops; distribution builds provide the real implementations behind `//go:build distro`.

This PR only adds upstream no-op hooks - the actual platform implementations live behind `//go:build distro` in distribution forks.

**Checklist**:

```release-note
Add build-tag extension points for distribution-specific TLS certificate creation, router networking, controller setup, and manager options configuration in the LLMInferenceService controller.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/workload_tls_cert_default.go
+++ b/pkg/controller/v1alpha2/llmisvc/workload_tls_cert_default.go
@@ -1,0 +1,34 @@
+//go:build !distro
+
+/*
+Copyright 2025 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package llmisvc
+
+import "context"
+
+// createWorkloadCertificate returns a createCertFunc that generates a self-signed TLS certificate.
+// Because the certificate signs itself (issuer = subject, same key pair), it is its own CA -
+// ca.crt is therefore identical to tls.crt.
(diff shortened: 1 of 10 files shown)


bartoszmajsak:
Gentle ping for review.

pkg/controller/v1alpha2/llmisvc/workload_tls_cert_default.go
```diff
@@ -0,0 +1,34 @@
...
+// createWorkloadCertificate returns a createCertFunc that generates a self-signed TLS certificate.
+// Because the certificate signs itself (issuer = subject, same key pair), it is its own CA -
+// ca.crt is therefore identical to tls.crt.
+func (r *LLMISVCReconciler) createWorkloadCertificate(_ context.Context, dnsNames []string, ips []string) createCertFunc {
+	return func() (*certBundle, error) {
+		keyBytes, certBytes, err := createSelfSignedTLSCertificate(dnsNames, ips)
+		if err != nil {
+			return nil, err
+		}
+		return &certBundle{Key: keyBytes, Cert: certBytes}, nil
```
Jooho: Quick question — in the self-signed flow, CACert is nil so ca.crt is not set in the Secret.
Is that intentional, or would it make sense to always populate ca.crt (e.g., same as tls.crt) for consistency with the reuse path?

Jooho (commented):
I didn’t find any issues in the code. I left a small comment, but overall it looks good to me.
/lgtm

pkg/controller/v1alpha2/llmisvc/workload_tls_cert_default.go
bartoszmajsak: Good find, I added it.

Jooho (approved):
/lgtm
/approved

PR #5242, opened 2026-03-12
fix(llmisvc): add missing os.Exit(1) after controller setup failure

### LINKED ISSUE #5241: add missing os.Exit(1) after controller setup failure
In cmd/llmisvc/main.go, the SetupWithManager call for LLMISVCReconciler logs the error on failure but does not call os.Exit(1), unlike every other setup block in the file. This means if setup fails, the pod starts and passes health checks but silently never reconciles LLMInferenceService resources.

vivekk16:
Closes #5241 

Added missing `os.Exit(1)` after `LLMISVCReconciler.SetupWithManager` failure in `cmd/llmisvc/main.go`, consistent with all other setup blocks in the file.

### DIFF
--- a/cmd/llmisvc/main.go
+++ b/cmd/llmisvc/main.go
@@ -205,6 +205,7 @@ func main() {
 		},
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "LLMInferenceService")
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm
/approve

Jooho (approved):
/lgtm

PR #5249, opened 2026-03-13
feat(llmisvc): update LLMInferenceServiceConfig and add TLS flag

pierDipi:
- Replace direct vllm command arrays with bash script entrypoints containing RoCE auto-discovery, `${VLLM_ADDITIONAL_ARGS}`, and `$@` passthrough across all templates
- Add DP_ADDRESS retry loop (30 attempts + fallback) to all data-parallel containers, replacing exit-on-failure
- Fix NVSHMEM_HCA_LIST to use ${ucx_hcas} (with :1 port suffix)
- Add cat /proc/driver/nvidia/params for RoCE debugging
- Introduce {{ .GlobalConfig.EnableTLS }} Go template conditionals for TLS flags (vllm SSL, probe schemes, sidecar/scheduler secure-serving)
- Wire EnableTLS from IngressConfig through Config to templates
- Update sidecar/scheduler from args to command with explicit binary, add --pool-group, --enable-ssrf-protection, SSL_CERT_DIR


config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -12,21 +12,140 @@ spec:
...
+              fi
+            fi
+
+            eval "vllm serve /mnt/models \
+              --served-model-name "{{ .Spec.Model.Name }}" \
+              --port 8001 \
+              --disable-uvicorn-access-log \
+              {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }} \
+              {{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt{{- end }} \
+              {{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key{{- end }} \
```
bartoszmajsak, line 144: Why not:
```suggestion
              {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh \ {{- end }}
              {{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt \{{- end }}
              {{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key \{{- end }} 
```

config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -136,6 +253,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: SSL_CERT_DIR
+            value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
```
bartoszmajsak, line 257: Should this be also flagged or harmless?

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -71,13 +72,12 @@ spec:
...
-              # - "https"
-              # - --model-server-metrics-https-insecure-skip-verify
-              # - --cert-path
-              # - "/var/run/kserve/tls"
+              - '{{ if .GlobalConfig.EnableTLS }}--secure-serving=true{{else}}--secure-serving=false{{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--model-server-metrics-scheme=https{{else}}--model-server-metrics-scheme=http{{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- end }}'
+            env:
+              - name: SSL_CERT_DIR
+                value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
```
bartoszmajsak, line 80: Same as above - do we want to gate it too?

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -122,6 +122,7 @@ type IngressConfig struct {
 	AdditionalIngressDomains   *[]string `json:"additionalIngressDomains,omitempty"`
 	DomainTemplate             string    `json:"domainTemplate,omitempty"`
 	UrlScheme                  string    `json:"urlScheme,omitempty"`
+	EnableTLS                  bool      `json:"enableTLS,omitempty"`
```
bartoszmajsak: Do we want to scope it to LLMInferenceService path to not overpromise?

bartoszmajsak (commented):
Minor comments

config/llmisvcconfig/config-llm-decode-template.yaml
bartoszmajsak: Or maybe even:

```suggestion
{{- if .GlobalConfig.EnableTLS }}
              --enable-ssl-refresh \
              --ssl-certfile /var/run/kserve/tls/tls.crt \
              --ssl-keyfile /var/run/kserve/tls/tls.key \
{{- end }}
```

config/llmisvcconfig/config-llm-decode-template.yaml
pierDipi: it's harmless

pkg/apis/serving/v1beta1/configmap.go
pierDipi: Done

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -111,20 +111,21 @@ type MultiNodeConfig struct {
...
+	IngressGateway               string    `json:"ingressGateway,omitempty"`
+	KnativeLocalGatewayService   string    `json:"knativeLocalGatewayService,omitempty"`
+	LocalGateway                 string    `json:"localGateway,omitempty"`
+	LocalGatewayServiceName      string    `json:"localGatewayService,omitempty"`
+	IngressDomain                string    `json:"ingressDomain,omitempty"`
+	IngressClassName             *string   `json:"ingressClassName,omitempty"`
+	AdditionalIngressDomains     *[]string `json:"additionalIngressDomains,omitempty"`
+	DomainTemplate               string    `json:"domainTemplate,omitempty"`
+	UrlScheme                    string    `json:"urlScheme,omitempty"`
+	EnableLLMInferenceServiceTLS bool      `json:"enableLLMInferenceServiceTLS,omitempty"`
```
pierDipi, line 125: Only added this flag, rest is reformatting

KillianGolds (commented):
### Question: is `eval` needed for the vllm entrypoint?

Apologies if I'm missing context on why this is done this way as i could be way off here but just wanted to flag something I noticed in case it's helpful.

The vllm serve command across the templates is wrapped in `eval`:

- [`config-llm-template.yaml` (L142)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-template.yamlL142)
- [`config-llm-decode-template.yaml` (L138)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-decode-template.yamlL138)
- [`config-llm-prefill-template.yaml` (L139)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-prefill-template.yamlL139)
- [`config-llm-decode-worker-data-parallel.yaml` (L218, L460)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-decode-worker-data-parallel.yamlL218)
- [`config-llm-worker-data-parallel.yaml` (L159, L401)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-worker-data-parallel.yamlL159)
- [`config-llm-prefill-worker-data-parallel.yaml` (L160, L402)](https://github.com/kserve/kserve/pull/5249/files#diff-config/llmisvcconfig/config-llm-prefill-worker-data-parallel.yamlL160)

I might be wrong here, but I think the inner double quotes around the model name might not behave as expected with `eval` - e.g. at `config-llm-template.yaml` L142-143:

```
eval "vllm serve /mnt/models \
  --served-model-name "{{ .Spec.Model.Name }}" \
```

I believe `eval` re-parses the string, so those quotes end up ending and re-opening the outer double-quoted string rather than protecting the model name value right? If a model name ever had spaces or special characters it could get messy maybe.

Similarly `${VLLM_ADDITIONAL_ARGS}` would get full shell expansion inside `eval`, which might be intentional for word splitting but could surprise operators if the value contains metacharacters.

Since `"--"` is already passed as a separate container arg to enable `$@` passthrough, would it be possible to drop `eval` and just run vllm directly? Something roughly like:

```bash
vllm serve /mnt/models \
  --served-model-name '{{ .Spec.Model.Name }}' \
  --port 8000 \
  --disable-uvicorn-access-log \
  {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh \
  --ssl-certfile /var/run/kserve/tls/tls.crt \
  --ssl-keyfile /var/run/kserve/tls/tls.key \{{- end }}
  ${VLLM_ADDITIONAL_ARGS} \
  "$@"
```

If `eval` is intentionally needed for `VLLM_ADDITIONAL_ARGS`, totally fair - maybe just worth escaping the model name quotes (`\"{{ .Spec.Model.Name }}\"`) so they survive the re-parse?

Happy to be corrected if I'm misreading this!

pierDipi:
@KillianGolds eval + quoting is already used in llmisvcconfig, we can consider that separately 
https://github.com/kserve/kserve/blob/master/config/llmisvcconfig/config-llm-decode-worker-data-parallel.yaml#L203-L205

config/llmisvcconfig/config-llm-template.yaml
```diff
@@ -12,21 +12,144 @@ spec:
...
-          # - --ssl-certfile
-          # - /var/run/kserve/tls/tls.crt
-          # - --ssl-keyfile
-          # - /var/run/kserve/tls/tls.key
+          - "/bin/bash"
+          - "-c"
+          - |-
+            if [ -f /etc/profile.d/ibm-aiu-setup.sh ]; then
+              source /etc/profile.d/ibm-aiu-setup.sh
+            fi
```
Jooho: I thought this script was deleted, no?

config/llmisvcconfig/config-llm-prefill-worker-data-parallel.yaml
```diff
@@ -294,7 +318,7 @@ spec:
                     hcas=$(IFS=,; echo "${active_hcas[*]}")
                     echo "[Infer RoCE] Setting active HCAs: ${hcas}"
                     export NCCL_IB_HCA=${NCCL_IB_HCA:-${hcas}}
-                    export NVSHMEM_HCA_LIST=${NVSHMEM_HCA_LIST:-${hcas}}
+                    export NVSHMEM_HCA_LIST=${NVSHMEM_HCA_LIST:-${ucx_hcas}}
```
Jooho, line 321: was it a bug to use `hcas` before?

Jooho (commented):
Not directly related to this PR, but I noticed the RoCE configuration block is duplicated across multiple YAML files. This could become a maintenance issue—any change would require updating several files, which risks inconsistencies.

Would it make sense to extract the RoCE part into a shared script so it only needs to be updated in one place?

config/llmisvcconfig/config-llm-prefill-worker-data-parallel.yaml
pierDipi: yes, HCA_LIST requires a port (default being 1) and ucx_hcas has the `device:port`

config/llmisvcconfig/config-llm-template.yaml
pierDipi: Fixed

bartoszmajsak (approved):
/lgtm
/approved

Jooho (approved):
/lgtm
/approved

PR #5250, opened 2026-03-13
fix: replace lookup with values toggle for ClusterStorageContainer CRD

### LINKED ISSUE #5210: Oscillating install/deletion of ClusterStorageContainer CRD when installing kserve-crd with helm (v0.17.0-rc0)
/kind bug

Was looking at the `v0.17.0-rc0` release of KServe for awareness of what's to come and noticed interesting/unexpected behavior when creating subsequent releases of the `kserve-crd` chart. Initial install of the crds chart works as expected. However, on creating a new release, the `ClusterStorageContainer` CRD gets deleted. Creating another release of the CRD chart then causes the `ClusterStorageContainer` CRD to get installed again. Long story short, it's a oscillating cycle of this particular CRD being installed/deleted on each helm release.

Installing/upgrading the `kserve-crd` helm chart causes oscillating install/uninstall of the `ClusterStorageContainer` CRD. For example, performing an initial install like so:

`helm install kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version v0.17.0-rc0 -n kserve --create-namespace`

Will result in the following CRDs being installed:

```
clusterservingruntimes.serving.kserve.io              2026-03-09T20:48:43Z
clusterstoragecontainers.serving.kserve.io            2026-03-09T20:48:43Z
inferencegraphs.serving.kserve.io                     2026-03-09T20:48:43Z
inferenceservices.serving.kserve.io                   2026-03-09T20:48:43Z
servingruntimes.serving.kserve.io                     2026-03-09T20:48:43Z
trainedmodels.serving.kserve.io                       2026-03-09T20:48:43Z
```

Then performing an upgrade like so:

`helm upgrade kserve-crd oci://ghcr.io/kserve/charts/kserve-crd --version v0.17.0-rc0 -n kserve`

Will result in the `ClusterStorageContainer` CRD being deleted - the following CRDs remain post-upgrade:

```
clusterservingruntimes.serving.kserve.io              2026-03-09T20:48:43Z
inferencegraphs.serving.kserve.io                     2026-03-09T20:48:43Z
inferenceservices.serving.kserve.io                   2026-03-09T20:48:43Z
servingruntimes.serving.kserve.io                     2026-03-09T20:48:43Z
trainedmodels.serving.kserve.io                       2026-03-09T20:48:43Z
```

Seems it's related to this [helm lookup](https://github.com/kserve/kserve/blob/v0.17.0-rc0/charts/kserve-crd/templates/serving.kserve.io_clusterstoragecontainers.yaml) in the templating. I assume this is likely present to prevent ownership conflicts when creating helm releases of the various KServe CRD charts separately. 

Mainly was just curious to know if this behavior will persist long term or if it'd be possible to add a boolean option to the values to bypass the lookup in the templates or maybe a `helm.sh/resource-policy: keep` annotation to prevent its deletion. See below in the `Anything else` section for how I ran into this. The way we're installing KServe is definitely not typical so I completely understand if we'll have to get creative on our end to handle the upcoming `v0.17.0` release. Also if contributing would be an acceptable path, am happy to set aside some time for this! :)

Wasn't expecting the `ClusterStorageContainer` CRD to get deleted on subsequent helm releases.

N/A

We're using a product from the company Replicated called Embedded Cluster. In a nutshell, you define their [Kots HelmChart CRs](https://docs.replicated.com/reference/custom-resource-helmchart-v2) in a bundle and they package it with a slightly modified K0s binary. They also help handle upgrades which is the reason I ran into this. I have a HelmChart CR defined for the `kserve-crd` chart which causes it to create a new helm release of the `kserve-crd` each time our app gets upgraded. That's how I ran into the oscillating install/deletion of the `ClusterStorageContainer` CRD.

**Environment:**

- Istio Version: n/a
- Knative Version: n/a
- KServe Version: `v0.17.0-rc0`
- Kubeflow version: n/a
- Cloud Environment: Self-hosted k0s (Replicated Embedded Cluster)
- Minikube/Kind version: n/a
- Kubernetes version: (use `kubectl version`): 1.33

wojciech-wais:
The ClusterStorageContainer CRD template used a Helm `lookup` to skip rendering when the CRD already existed, preventing ownership conflicts between kserve-crd and kserve-llmisvc-crd charts. However, this caused the CRD to oscillate between installed and deleted on every helm upgrade because Helm deletes resources it owns that are no longer in the rendered manifest.

Replace the lookup guard with a `clusterStorageContainerCRD.install` values toggle (default true). Users installing both CRD charts can set the flag to false on one chart to avoid conflicts.

Fixes #5210

Fixes #

- [ ] Validated the syntax of the template files.

- Logs

**Checklist**:

**Release note**:
```release-note

```


aslafy-z:
Any chance this gets merged before 0.19 release? It would ease kserve setup with helmcharts on gitops managed platforms. cc @Jooho

wojciech-wais:
I just merged latest master into this branch and resolved the conflict in the 4 `ClusterStorageContainer` CRD templates. The conflict exists because #5539 ("prevent ClusterStorageContainer CRD deletion on helm upgrade") touched the exact same template guard. I want to clarify that the two changes are **not** redundant:

- **#5539** keeps the Helm `lookup` and adds release-ownership detection (`meta.helm.sh/release-name`). This fixes deletion-on-`helm upgrade` for a live cluster.
- **#5539 does not fix the GitOps case**, because `lookup` is itself the root cause there. Tools like Argo CD / Flux render manifests via `helm template` / server-side dry-run **without cluster read access**, so `lookup` returns empty and the CRD oscillates between rendered/absent across reconciles — exactly the behavior reported in #5210.

This PR removes `lookup` entirely and gates rendering on a `.Values.clusterStorageContainerCRD.install` toggle (default `true`), which renders deterministically regardless of cluster state.

I resolved the merge by keeping the toggle approach (dropping the reintroduced `lookup`), since combining the two would put `lookup` back and defeat the purpose of this fix.

wojciech-wais:
@Jooho 
Can you please check it?

aslafy-z:
This PR is still desired. Can you take a look @Jooho @vivekk16 @bartoszmajsak?

bartoszmajsak:
@wojciech-wais thanks for this PR. I couldn't get the e2e jobs to trigger, so I ran it on kind.

Both CRD charts now default to `install: true`, so installing both breaks:

```
ENABLE_KSERVE=true ENABLE_LLMISVC=true hack/setup/infra/manage.kserve-helm.sh
```

kserve-llmisvc-crd fails with invalid ownership metadata; key "meta.helm.sh/release-name" must equal "kserve-llmisvc-crd": current value is "kserve-crd". 

helm upgrade fails the same way on existing clusters that have both charts.

PR #5252, opened 2026-03-13
fix(controllers): gracefully handle resource deletion

bartoszmajsak:

When a resource is deleted while a reconcile is in flight, the updateStatus re-fetch returns NotFound and triggers error-driven backoff requeues. These requeues are pointless - the next reconcile just finds the resource gone and returns nil - but they delay processing of the deletion event, preventing finalizer cleanup from running promptly.

Treats a missing resource in updateStatus as a no-op instead of an error, consistent with how the main Reconcile functions already handle NotFound at their entry points.

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm
/approve

sivanantha321 (approved):
/lgtm
/approve

sivanantha321:
@bartoszmajsak looks like this needs a rebase ?

bartoszmajsak:
> @bartoszmajsak looks like this needs a rebase ?

Yes, I'm going through all the PRs that has been hanging in review limbo, they should get updates soon.

PR #5263, opened 2026-03-17
feat(localmodel): add build hooks for distribution-specific logic

VedantMahabaleshwarkar:

Adds compile-time hook points to the LocalModelNode agent controller using Go build tags (`distro`/`!distro`), following the pattern established in #5217 for LLMInferenceService.

Upstream gets zero-cost no-op stubs. Distributions can provide real implementations behind the `distro` tag without forking core reconciliation files — reducing merge conflicts and making upstream rebases clean.

Hook points:
- **`enhanceDownloadJob`** — platform-specific modifications to download jobs before creation (e.g., clearing SubPath, setting SELinux context, ServiceAccount)
- **`ensureVolumePermissions`** — guards model download by checking volume writability and launching permission fix jobs if needed
- **`EnsurePermissions`** — startup-time permission check before the controller manager starts

Also adds:
- `GOTAGS` variable to `localmodel-agent.Dockerfile` so distributions can activate build tags through the build pipeline
- `isWritable()` method on `FileSystemInterface` for volume permission detection
- Improved `getModelFolders()` to return `nil` instead of error on missing directory
- Replaced `panic()` with error return in `fsHelper` initialization

Depends on #5262

- [x] Upstream builds compile and run tests without the `distro` tag (no-op stubs)
- [x] Distribution builds compile with `-tags distro`
- [x] Verified no behavioral change for upstream — all stubs return nil
- [x] `TestFileSystemHelper_isWritable` — verifies writable and read-only directory detection

This follows the same pattern established in #5217 for `llmisvc`. The `platform_default.go` file uses `//go:build !distro` to provide upstream no-op stubs. A corresponding platform_go files can be created downstream for platform specific implementation 

**Checklist**:

```release-note
Add compile-time hook points to LocalModelNode agent controller for distribution-specific logic using Go build tags. Distributions can now inject platform-specific behavior (download job customization, volume permissions) without forking core controller files.
```

### DIFF
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -67,7 +67,6 @@ type LocalModelNodeReconciler struct {
 }
 
 const (
-	MountPath             = "/mnt/models" // Volume mount path for models, must be the same as the value in the DaemonSet spec
 	DownloadContainerName = "kserve-localmodel-download"
 	PvcSourceMountName    = "kserve-pvc-source"
 )
@@ -204,6 +203,7 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, localModelNode
 			},
 		},
 	}
+	enhanceDownloadJob(job, storageKey)
 	if err := controllerutil.SetControllerReference(&localModelNode, job, c.Scheme); err != nil {
 		c.Log.Error(err, "Failed to set controller reference", "name", modelInfo.ModelName)
 		return nil, err
@@ -526,10 +526,8 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// fsHelper is a global variable to allow mocking in tests
 	if fsHelper == nil {
 		fsHelper = NewFileSystemHelper(modelsRootFolder)
-		// TODO we need a way to ensure that the local path on persistent volume is the same as the local path of the node agent DaemonSet.
-		err := fsHelper.ensureModelRootFolderExists()
-		if err != nil {
-			panic("Failed to ensure model root folder exists: " + err.Error())
+		if err := fsHelper.ensureModelRootFolderExists(); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to ensure model root folder: %w", err)
 		}
 	}
 
@@ -575,6 +573,11 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	if c.CredentialBuilder == nil {
 		c.CredentialBuilder = credentials.NewCredentialBuilder(c.Client, c.Clientset, isvcConfigMap)
+	}
+
+	result, cont, err := ensureVolumePermissions(ctx, c, localModelConfig)
+	if !cont || err != nil {
+		return result, err
 	}
 
 	if err := c.downloadModels(ctx, &localModelNode); err != nil {
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -64,6 +64,7 @@ type LocalModelNodeReconciler struct {
 	Log               logr.Logger
 	Scheme            *runtime.Scheme
 	CredentialBuilder *credentials.CredentialBuilder
+	IsvcConfigMap     *corev1.ConfigMap
 }
 
 const (
@@ -523,34 +524,13 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	c.Log.Info("Agent reconciling LocalModelNode", "name", req.Name, "node", nodeName)
 
-	// fsHelper is a global variable to allow mocking in tests
-	if fsHelper == nil {
-		fsHelper = NewFileSystemHelper(modelsRootFolder)
-		if err := fsHelper.ensureModelRootFolderExists(); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to ensure model root folder: %w", err)
-		}
-	}
-
-	// Create Jobs to download models if the model is not present locally.
-	// 1. Check if LocalModelNode CR is for current node
-	localModelNode := v1alpha1.LocalModelNode{}
-	if err := c.Get(ctx, req.NamespacedName, &localModelNode); err != nil {
-		c.Log.Error(err, "Error getting LocalModelNode", "name", req.Name)
-		return reconcile.Result{}, client.IgnoreNotFound(err)
-	}
-
-	// 2. Cleanup jobs for models that are not in the spec
-	if err := c.cleanupJobs(ctx, localModelNode); err != nil {
-		c.Log.Error(err, "Job cleanup err")
-		return reconcile.Result{}, err
-	}
-
-	// 3. Kick off download jobs for all models in spec
+	// 1. Load config
 	isvcConfigMap, err := v1beta1.GetInferenceServiceConfigMap(ctx, c.Clientset)
 	if err != nil {
 		c.Log.Error(err, "unable to get configmap", "name", constants.InferenceServiceConfigMapName, "namespace", constants.KServeNamespace)
 		return reconcile.Result{}, err
 	}
+	c.IsvcConfigMap = isvcConfigMap
 	localModelConfig, err := v1beta1.NewLocalModelConfig(isvcConfigMap)
 	if err != nil {
 		c.Log.Error(err, "Failed to get local model config")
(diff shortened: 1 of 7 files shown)


pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -49,6 +49,7 @@ const (
 	OtelCollectorConfigName            = "opentelemetryCollector"
 	StorageInitializerConfigMapKeyName = "storageInitializer"
 	AutoscalerConfigName               = "autoscaler"
+	OpenShiftConfigName                = "openshiftConfig"
```
bartoszmajsak: Does it belong upstream?

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -411,6 +412,21 @@ func NewServiceConfig(isvcConfigMap *corev1.ConfigMap) (*ServiceConfig, error) {
 	return serviceConfig, nil
 }
 
+// +kubebuilder:object:generate=false
```
spolti: shouldn't this be added in a separate file with distro tag? 
otherwise we contradict the goal of having the distribution-specific logic behind build tags

pkg/controller/v1alpha1/localmodelnode/file_utils.go
```diff
@@ -71,3 +81,14 @@ func (f *FileSystemHelper) ensureModelRootFolderExists() error {
...
+
+func (f *FileSystemHelper) isWritable() bool {
+	file, err := os.CreateTemp(f.modelsRootFolder, ".write-test-*")
+	if err != nil {
+		return false
+	}
+	name := file.Name()
+	_ = file.Close()
+	_ = os.Remove(name)
+	return true
```
bartoszmajsak: nit, but maybe:

```suggestion
	defer os.Remove(file.Name())
	defer file.Close()
	return true
```
Or if we don't care about Windows:

```go
func (f *FileSystemHelper) isWritable() bool {
      return unix.Access(f.modelsRootFolder, unix.W_OK) == nil
}
```

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -411,6 +412,21 @@ func NewServiceConfig(isvcConfigMap *corev1.ConfigMap) (*ServiceConfig, error) {
...
+type OpenShiftConfig struct {
+	ModelcachePermissionFixImage string `json:"modelcachePermissionFixImage,omitempty"`
+}
+
+func NewOpenShiftConfig(isvcConfigMap *corev1.ConfigMap) (*OpenShiftConfig, error) {
```
spolti: where is this being used?

go.mod
```diff
@@ -147,6 +147,7 @@ require (
 	github.com/oasdiff/yaml v0.0.0-20250309154309-f31be36b4037 // indirect
 	github.com/oasdiff/yaml3 v0.0.0-20250309153720-d2182401db90 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
+	github.com/openshift/api v0.0.0-20240124164020-e2ce40831f2e // indirect
```
spolti: this dependency is quite old, is it still maintained?

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
```diff
@@ -121,14 +119,13 @@ func (c *LocalModelNamespaceCacheReconciler) Reconcile(ctx context.Context, req
 
 		pvc := corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: localModel.Name + "-" + nodeGroup.Name + "-download",
+				Name: localModel.Name + "-" + nodeGroup.Name + "-" + localModel.Namespace + "-download",
```
spolti, line 122: I guess with this name change the existing PVCs from previous deployments using the old naming scheme in the old namespace will become orphaned, no?

python/kserve/kserve/models/v1beta1_open_shift_config.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2023 The KServe Authors.
```
spolti: 2026

python/kserve/test/test_v1beta1_open_shift_config.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2023 The KServe Authors.
```
spolti: 2026

pkg/apis/serving/v1beta1/configmap.go
spolti: btw, shouldn't this be only on downstream?

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
spolti: Guess it will break upgrades as well.

VedantMahabaleshwarkar: This is only for the namespace cache which was just introduced last week and isn't in the 0.17 release so I don't think we need to worry about updates here

pkg/apis/serving/v1beta1/configmap.go
VedantMahabaleshwarkar: Yeah this shouldn't be upstream, I'll change it

pkg/apis/serving/v1beta1/configmap.go
VedantMahabaleshwarkar: will remove this from upstream, it isn't needed

pkg/apis/serving/v1beta1/configmap.go
bartoszmajsak: I think we don't even need a tag - just keep a new file `_ocp.go` in our repo and it should be good enough

pkg/controller/v1alpha1/localmodelnode/file_utils.go
VedantMahabaleshwarkar: used defer

go.mod
VedantMahabaleshwarkar: was wrongly brought in, removed

python/kserve/test/test_v1beta1_open_shift_config.py
VedantMahabaleshwarkar: no longer generated since the config was removed

python/kserve/kserve/models/v1beta1_open_shift_config.py
VedantMahabaleshwarkar: no longer generated since the config was removed

pkg/apis/serving/v1beta1/configmap.go
VedantMahabaleshwarkar: removed upstream

bartoszmajsak (approved):
/lgtm
/approve

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -570,24 +548,47 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return reconcile.Result{}, err
 	}
 
-	// 4. Delete models that are not in the spec. This function does not modify the resource.
+	// 6. Delete models that are not in the spec
 	if err := c.deleteModels(localModelNode); err != nil {
 		c.Log.Error(err, "Model deletion err")
 		return reconcile.Result{}, err
 	}
-	// Requeue to check local folders periodically
```
israel-hdez, line 590: May you bring back this comment?

For me, it makes clearer the motivation of the requeue.

pkg/controller/v1alpha1/localmodelnode/file_utils_test.go
```diff
@@ -149,3 +149,23 @@ func TestFileSystemHelper_ensureModelRootFolderExists(t *testing.T) {
...
+
+func TestFileSystemHelper_isWritable(t *testing.T) {
+	// Case 1: Writable directory returns true
+	tempDir := t.TempDir()
+	helper := NewFileSystemHelper(tempDir)
+	if !helper.isWritable() {
+		t.Errorf("expected writable temp dir to return true")
+	}
+
+	// Case 2: Read-only directory returns false
```
israel-hdez: nit: split to two different `funcs`. IMO tests should focus on a single check.

pkg/controller/v1alpha1/localmodelnode/platform_default.go
```diff
@@ -0,0 +1,43 @@
...
+
+	batchv1 "k8s.io/api/batch/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+)
+
+const MountPath = "/mnt/models"
+
+func enhanceDownloadJob(_ *batchv1.Job, _ string) {}
```
israel-hdez: would be worth to add `error` as return value?

pkg/controller/v1alpha1/localmodelnode/platform_default.go
```diff
@@ -0,0 +1,43 @@
...
+
+//nolint:unparam // signature must match distro variant which returns non-zero Result
```
israel-hdez: I see better if we add a type `ensureModelRootFolderResult` that can be augmented in distro-specific src, to prevent the linter error.

By adding artificial params we are cluttering the code here, and the function signature can be subject to changes later, depending on distro-specific implementations.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -523,36 +524,13 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	c.Log.Info("Agent reconciling LocalModelNode", "name", req.Name, "node", nodeName)
 
-	// fsHelper is a global variable to allow mocking in tests
-	if fsHelper == nil {
-		fsHelper = NewFileSystemHelper(modelsRootFolder)
-		// TODO we need a way to ensure that the local path on persistent volume is the same as the local path of the node agent DaemonSet.
```
israel-hdez, line 529: This `TODO` still applies, no?

I think we should bring it back in the `_default` file.

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -570,24 +548,47 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// 4. Cleanup jobs for models that are not in the spec
+	if err := c.cleanupJobs(ctx, localModelNode); err != nil {
+		c.Log.Error(err, "Job cleanup err")
+		return reconcile.Result{}, err
+	}
+
+	// 5. Download models
```
israel-hdez: IIUC, a more accurate comment: 

```suggestion
	// 5. Download models not present locally
```

(or correct me if I'm wrong)

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -570,24 +548,47 @@ func (c *LocalModelNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	storageInitializerConfig, err = v1beta1.GetStorageInitializerConfigs(isvcConfigMap)
 	if err != nil {
 		c.Log.Error(err, "Failed to get storage initializer config, using defaults")
-		// Don't fail, just use defaults
```
israel-hdez, line 573: I think this comment makes it clear that omitting `return nil, err` is intended.

pkg/controller/v1alpha1/localmodelnode/file_utils.go
```diff
@@ -71,3 +81,13 @@ func (f *FileSystemHelper) ensureModelRootFolderExists() error {
 	}
 	return nil
 }
+
+func (f *FileSystemHelper) isWritable() bool {
```
israel-hdez: Hmm... unused here?

VedantMahabaleshwarkar: removed

israel-hdez (approved):
/lgtm

terrytangyuan (approved):

PR #5264, opened 2026-03-17
feat: replace vllm-cpu image with upstream built image

pierDipi:
Currently, we're using my image as it wasn't available the cpu image previously, now it's on ECR and we can use it.

https://gallery.ecr.aws/q9t5s3a7/vllm-cpu-release-repo


bartoszmajsak (approved):
/lgtm
/approved

Jooho (approved):
/lgtm
/approved

PR #5265, opened 2026-03-17
feat: [WIP] Improve GH action skip filters

pierDipi:
Currently, CI runs 50+ jobs on small sub-component scoped changes which don't necessarily need to be tested, which causes a lot of reruns and flake failures on unrelated changes


pierDipi:
https://github.com/kserve/kserve/pull/5267

.github/workflows/go.yml
```diff
@@ -45,6 +45,7 @@ jobs:
               uses: actions/setup-go@v5
               with:
                go-version-file: go.mod
+               cache: true
```
bartoszmajsak: Isn't it default behaviour? https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

.github/workflows/e2e-test.yml
```diff
@@ -52,6 +64,48 @@ jobs:
...
+              - 'cmd/localmodel*/**'
+              - 'test/e2e/modelcache/**'
+            python_runtimes:
+              - 'python/**'
+            explainer:
+              - 'python/artexplainer/**'
+              - 'test/e2e/explainer/**'
+            huggingface:
+              - 'python/huggingfaceserver/**'
+              - 'test/e2e/predictor/test_huggingface*'
```
bartoszmajsak, line 122: I wonder if we shouldn't also act on Dockerfile changes:

```suggestion
            huggingface:
              - 'python/huggingfaceserver/**'
              - 'python/huggingface_server.Dockerfile'
              - 'python/huggingface_server_cpu.Dockerfile'
              - 'test/e2e/predictor/test_huggingface*'
```

.github/workflows/e2e-test.yml
```diff
@@ -52,6 +64,48 @@ jobs:
...
+            graph:
+              - 'pkg/apis/**/inference_graph*'
+              - 'pkg/controller/**/inference_graph*'
+              - 'test/e2e/graph/**'
+            qpext:
+              - 'qpext/**'
+              - 'test/e2e/qpext/**'
+            router:
+              - 'cmd/router/**'
+              - 'pkg/router/**'
```
bartoszmajsak, line 103: Shouldn't this include:


```suggestion
            router:
              - 'cmd/router/**'
              - 'pkg/router/**'
              - `test/e2e/graph`
```

bartoszmajsak (commented):
One thing I wonder, but I haven't looked deeply yet:

When a build job is skipped via `if:`, does GitHub Actions treat it as "succeeded" for `needs` purposes? If a downstream test job *isn't* skipped and tries to download artifacts from a skipped build - what happens then? AFAICT we are relying on GitHub commit hashes to pin exact image versions.

.github/workflows/e2e-test.yml
pierDipi: There is a separate graph key

Jooho (dismissed):
/lgtm
/approved

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -3,14 +3,18 @@ name: LLMInferenceService E2E Tests
 on:
   pull_request:
     paths:
-      - "pkg/apis/serving/v1alpha1/llm*"
-      - "pkg/apis/serving/v1alpha2/llm*"
-      - "pkg/controller/v1alpha1/llmisvc/**"
-      - "pkg/controller/v1alpha2/llmisvc/**"
-      - "charts/kserve-llmisvc-resources/**"
+      - "pkg/apis/serving/**/llm*"
```
sivanantha321, line 6: Configuring path filters in the workflow level will block the PR merge if it is marked as required. But at job level the skipped jobs are considered successful. So i am thinking if we can create a separate job to identify if the test jobs should be run based on the file changes. We can achieve this using `dorny/paths-filter@v3`

pierDipi:
Less fragile alternative https://github.com/kserve/kserve/pull/5941

PR #5266, opened 2026-03-17
fix(e2e): retry predict on transient gateway errors

pierDipi:
The InferenceGraph e2e tests are flaky because the ingress gateway sometimes returns 502 Bad Gateway before being fully ready, causing tests that expect application-level errors (e.g. 404) to fail with a JSONDecodeError on the non-JSON "EOF" response body.

Add retry logic to predict that retries on transient HTTP errors (502, 503, 504) with exponential backoff, while immediately re-raising non-transient errors.

### DIFF
--- a/test/e2e/common/utils.py
+++ b/test/e2e/common/utils.py
@@ -23,6 +23,8 @@ import portforward
 from kubernetes import client as k8s_client
 from orjson import orjson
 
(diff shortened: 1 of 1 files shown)


Jooho (approved):
/lgtm
/approved

PR #5267, opened 2026-03-17
fix(ci): add readiness to seaweedfs deployment

pierDipi:
The seaweedfs deployment had no readiness probe, so Kubernetes
marked the pod as ready before S3 was actually accepting connections,
causing the s3-init job to fail intermittently.

The s3-init job could start before seaweedfs was ready, causing
a timeout waiting for the job to complete. Add a rollout status
check to ensure seaweedfs is serving before initializing S3 storage.


bartoszmajsak (approved):
/lgtm
/approved

Thanks!

Jooho (approved):
/lgtm
/approved

PR #5268, opened 2026-03-17
feat: adding version overrides helm chart dependencies

andresllh:
Adding version overrides to generate-versions-from-gomody.py. This is needed to fix a CVE in KEDA that is solved in 2.17.3, but that helm chart is not available yet.

- [ ] Verify that kserve dependency files have KEDA_VERSION set to 2.17.3 and not 2.17.1

- Logs

**Checklist**:

- [x] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note

```

### DIFF
--- a/hack/setup/scripts/generate-versions-from-gomod.py
+++ b/hack/setup/scripts/generate-versions-from-gomod.py
@@ -54,12 +54,6 @@ HELM_REPOS = {
     "open-telemetry": "https://open-telemetry.github.io/opentelemetry-helm-charts",
 }
 
-# Manual version overrides for security patches where Helm charts lag behind go.mod
-# These take precedence over auto-generated versions from Helm chart lookups
-VERSION_OVERRIDES = {
-    "KEDA_VERSION": "2.17.3",  # CVE-2025-68476 fix (CVSS 8.2, HIGH) - Helm chart only has 2.17.1
-}
-
 
 def run(cmd):
     return subprocess.run(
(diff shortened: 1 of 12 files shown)


spolti (approved):
/lgtm

hack/setup/scripts/generate-versions-from-gomod.py
```diff
@@ -54,6 +54,12 @@
     "open-telemetry": "https://open-telemetry.github.io/opentelemetry-helm-charts",
 }
 
+# Manual version overrides for security patches where Helm charts lag behind go.mod
+# These take precedence over auto-generated versions from Helm chart lookups
+VERSION_OVERRIDES = {
```
Jooho: I’d prefer to add this information to kserve-deps.env so that all versions are managed in a single file.

Jooho (changes requested):

kserve-deps.env
```diff
@@ -1,3 +1,5 @@
+# Overrides, these are manual overrides to versions. Will take presedence over other checks.
+OVERRIDE_KEDA_VERSION=2.17.3 # CVE-2025-68476 fix (CVSS 8.2, HIGH) - Helm chart only has 2.17.1
```
Jooho: We already used 2.17.3 so it can be commented out but I would like to keep this line for future reference.

Jooho (approved):
/lgtm
/approved

PR #5274, opened 2026-03-17
build: optimize Python Dockerfile caching and build context

bartoszmajsak:

Every Python Dockerfile copied the full `storage/` directory as its first COPY instruction, invalidating all downstream layers whenever any storage file changed - even for images that only need storage metadata for editable dependency resolution.

This PR switches 16 Dockerfiles to a metadata-first pattern that copies only `storage/pyproject.toml` and `storage/uv.lock` early, deferring the full source copy until actually needed. For images that install storage as a package (sklearn, lgb, xgb, paddle, pmml, huggingface), the redundant two-phase storage install is collapsed into a single step.

Removes duplicate `COPY third_party` in prod stages where it was followed by a `COPY --from=builder` that merged over it (Docker COPY merges directory contents rather than replacing them, so the build-time `pip-licenses.py` script was leaking into runtime images).

The `huggingface_server_cpu` Dockerfile gets the largest restructure, applying metadata-first to both kserve and huggingfaceserver dependencies - preventing a ~2GB torch re-download on source-only changes.

Adds `python/.dockerignore` to exclude dev artifacts (`.venv/`, `__pycache__/`, `.egg-info/`, test/docs directories). In a clean checkout the context is already small (~18 MB), but in dev environments with virtualenvs and caches this prevents unnecessary bloat from reaching the daemon.

- [x] Built sklearn image: `KO_DOCKER_REPO=kserve make docker-build-sklearn`
- [x] Built custom_model image: `KO_DOCKER_REPO=kserve make docker-build-custom-model`
- [x] Built aiffairness image: `KO_DOCKER_REPO=kserve make docker-build-aif`
- [x] Built artexplainer image: `KO_DOCKER_REPO=kserve make docker-build-art`
- [x] Built success_200_isvc image: `KO_DOCKER_REPO=kserve make docker-build-success-200-isvc`
- [x] Context size measured: 18.3 MB without `.dockerignore`, 16.7 MB with (clean checkout - dev environments with `.venv/` would see a larger delta)

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/python/artexplainer.Dockerfile
+++ b/python/artexplainer.Dockerfile
@@ -18,8 +18,8 @@ ENV VIRTUAL_ENV=${VENV_PATH}
 RUN uv venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-# Copy storage directory for editable install
-COPY storage storage
+# Copy storage metadata for editable dependency resolution
+COPY storage/pyproject.toml storage/uv.lock storage/
 
 # ------------------ kserve deps ------------------
 COPY kserve/pyproject.toml kserve/uv.lock kserve/
@@ -50,8 +50,6 @@ FROM ${BASE_IMAGE} AS prod
 ARG VENV_PATH
 ENV VIRTUAL_ENV=${VENV_PATH}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-
-COPY third_party third_party
 
 RUN useradd kserve -m -u 1000 -d /home/kserve
 
--- a/python/.dockerignore
+++ b/python/.dockerignore
@@ -1,0 +1,14 @@
+# Exclude dev artifacts from Python image builds (context: python/)
+**/.venv/
+**/__pycache__/
+**/*.pyc
+**/*.pyo
+**/.pytest_cache/
+**/.ruff_cache/
+**/.mypy_cache/
+**/*.egg-info/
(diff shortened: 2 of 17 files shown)


python/artexplainer.Dockerfile
```diff
@@ -51,8 +51,6 @@ ARG VENV_PATH
 ENV VIRTUAL_ENV=${VENV_PATH}
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-COPY third_party third_party
-
 RUN useradd kserve -m -u 1000 -d /home/kserve
 
 COPY --from=builder --chown=kserve:kserve third_party third_party
```
bartoszmajsak, line 56: Before the change, the prod image had both:
  - `third_party/pip-licenses.py` (from build context)
  - `third_party/library/` (from builder stage)

See https://github.com/kserve/kserve/pull/5274/changes#diff-027458e2c18f8c7198876e420a407a733ac2a8db8d424c50d5f1bf3368f7a801R38-R43

  After the change, the prod image only has:
  - `third_party/library/`

The removal does change the final image contents, but `pip-licenses.py` seems like it is a build-time tool that has no purpose in the runtime image. 

If its presence before was intentional, not accidental I can restore this step.

spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5275, opened 2026-03-17
fix(ci): correct self-referencing path in llmisvc docker workflow

bartoszmajsak:

Fixes a typo in the llmisvc controller Docker publish workflow's path filter. The self-referencing path had a double dash and missing hyphens (`kserve--llmisvccontroller-docker-publish.yml` instead of `kserve-llmisvc-controller-docker-publish.yml`), which meant the workflow would never trigger on PRs that only modify the workflow file itself.

**Release note**:
```release-note
NONE
```


spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5276, opened 2026-03-17
fix(ci): enable coverage report comment on PRs

bartoszmajsak:

Fixes the PR coverage report comment that was never being posted. The step was gated on `pull_request_target` but the workflow only triggers on `pull_request`, making the entire comment block dead code. Also replaces the deprecated `::set-output` syntax with `$GITHUB_OUTPUT` in the same workflow.

The workflow already declares `pull-requests: write` at the workflow level (line 24), so the comment step has the permissions it needs once the condition is corrected.

I wonder however - since it was broken/ignored for a long time - do we really need it?

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -58,7 +58,7 @@ jobs:
                 chmod +x $GOPATH/bin/yq
                 make test
                 ./coverage.sh
-                echo ::set-output name=coverage::$(./coverage.sh | tr -s '\t' | cut -d$'\t' -f 3)
+                echo "coverage=$(./coverage.sh | tr -s '\t' | cut -d$'\t' -f 3)" >> $GITHUB_OUTPUT
             
             - name: Print coverage
               run: |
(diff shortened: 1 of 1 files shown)


spolti (approved):
/lgtm

bartoszmajsak:
```
Error: An error occurred trying to start process '/home/runner/actions-runner/cached/externals/node20/bin/node' with working directory '/home/runner/work/kserve/kserve'. Argument list too long
```

Do we need this job at all then? :)

Jooho:
The post coverage report step fails with Resource not accessible by integration because GitHub restricts GITHUB_TOKEN to read-only for fork PRs. This is a GitHub security policy, not a bug in our workflow.

A simple fix would be to integrate [Codecov](https://codecov.io/) (free for public repos), which posts PR comments from its own service, bypassing GitHub token limitations entirely. Many CNCF projects(istio,coreDNS,Rook and so on) already use it. @yuzisun WDYT?

Alternatively, we can just remove this comment step since coverage is already available in the build logs and as a build artifact.

PR #5277, opened 2026-03-17
chore(ci): align action versions across workflows

bartoszmajsak:

Eliminates action version drift across CI workflows. All Python Docker publish workflows were using `docker/build-push-action@v5` in their test jobs while the push jobs already used `@v6`. Similarly, one `actions/cache` step in `python-test.yml` was stuck on `v3` while the other 8 cache steps used `v4`. This mismatch means test and publish builds run on different action versions, which could mask compatibility issues.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/workflows/artexplainer-docker-publish.yml
+++ b/.github/workflows/artexplainer-docker-publish.yml
@@ -60,7 +60,7 @@ jobs:
           cache-binary: true
 
       - name: Run tests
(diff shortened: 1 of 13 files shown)


spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5283, opened 2026-03-18
chore: fix CVE-2026-32597 PyJWT crit header validation

spolti:
## Summary
- Pin PyJWT >= 2.12.0 to fix CVE-2026-32597: PyJWT accepts unknown `crit` header extensions violating RFC 7515 §4.1.11 MUST requirement
- PyJWT is a transitive dependency brought in via `azure-identity` -> `msal` -> `PyJWT`
- Pinning because it is a transitive dependency

## Test plan
- [x] `make precommit` passes
- [x] Storage initializer container builds successfully
- [x] Verified PyJWT 2.12.1 is installed in the built container image

### DIFF
--- a/python/aiffairness/uv.lock
+++ b/python/aiffairness/uv.lock
@@ -1124,6 +1124,7 @@ dependencies = [
     { name = "protobuf" },
     { name = "psutil" },
     { name = "pydantic" },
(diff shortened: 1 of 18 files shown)


sivanantha321 (approved):
/lgtm
/approve

PR #5287, opened 2026-03-20
ci: reduce Go image build times

bartoszmajsak:

Go Dockerfile builds are slow - each one sequentially compiles the binary, installs go-licenses, runs license checks, and generates license artifacts. 

Every image build redundantly re-validates licenses, and when a license violation occurs it surfaces as an opaque build failure with no clear signal.

This change cuts cold build times by 45-54% and gives license violations their own CI status. `go-licenses` check is extracted into a dedicated GitHub Actions workflow that runs once per PR covering both the main module and  qpext. License violations now surface as a separate CI check rather than being buried in Dockerfile build logs.

Each Dockerfile is split into three stages (deps, builder, license) with `--mount=type=cache` for Go module and compilation caches.

BuildKit runs builder and license concurrently; Buildah runs them sequentially but still benefits from persistent cache mounts across builds.

- [x] All 7 Dockerfiles verified with Docker BuildKit and Podman 5.7.1
- [x] Benchmarked across 5 scenarios on both engines

Cold build improvements (the CI-relevant scenario):
- Docker BuildKit: **-45%** for main Dockerfile (354s -> 193s)
- Podman/Buildah: **-54%** for main Dockerfile (435s -> 200s)

[Full benchmark results](https://gist.github.com/bartoszmajsak/f9209e3fc7a24f9a28de2ca6d7d724a1)

Cache mounts require Buildah 1.24+. ci-operator runs Buildah 1.42+ on RHEL 9 - verified via UBI9 image labels.

`go-licenses save` stays in each Dockerfile (produces `/third_party` in the runtime image). Only `check` moves to CI.

**Checklist**:

```release-note
NONE
```


qpext/LICENSE
```diff
@@ -0,0 +1 @@
+../LICENSE
```
bartoszmajsak, line 1: This is now needed as license check is done as CI job. Alternatively, instead of symlink, we could `--ignore github.com/kserve/kserve/qpext` as it's under our control

spolti:
can we use this same approach in the python container files?

bartoszmajsak:
> can we use this same approach in the python container files?

@spolti Sounds reasonable - I can do a follow-up PR to keep things scoped.

cjohannsen-cloudera (approved):
It looks like enforce-required-checks is failing:
`check run "test-predictor (kustomize)" is completed with conclusion "failure" (unsuccessful)`

Jooho (approved):
/lgtm
/approved

Jooho:
rebased

.github/workflows/go-license-check.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Install go-licenses
+        run: go install github.com/google/go-licenses@v1.6.0
+
+      - name: Check main module licenses
+        run: go-licenses check ./cmd/... ./pkg/... --disallowed_types="forbidden,unknown"
```
sivanantha321:
```suggestion
        run: go-licenses check ./cmd/... ./pkg/... ./qpext/... --disallowed_types="forbidden,unknown"
```

bartoszmajsak: Thanks, added!

bartoszmajsak: Good instinct, but qpext is a separate Go module with its own `go.mod` - `./qpext/...` can't be resolved from the root module's package paths and causes a build error. It's already covered by the "Check qpext licenses" step below which runs `go-licenses check ./...` under `working-directory: qpext`. Reverted.

bartoszmajsak:
@sivanantha321 PTAL

sivanantha321 (approved):
/lgtm
/approve

PR #5291, opened 2026-03-20
feat: add /v1/responses HTTPRoute for OpenAI Responses API support

RishabhSaini:
Route /v1/responses traffic through InferencePool in LLMInferenceService configs, enabling backends (vLLM, SGLang) that already implement the Responses API to receive requests. Added route rules to both kustomize and Helm chart configs, added RESPONSES_ENDPOINT constant to the Python proxy model, and regenerated quick-install scripts.

### DIFF
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -1530,6 +1530,25 @@ spec:
               backendRequest: 0s
               request: 0s
           - backendRefs:
+            - group: inference.networking.k8s.io
+              kind: InferencePool
+              name: '{{ ChildName .ObjectMeta.Name `-inference-pool` }}'
+              port: 8000
+              weight: 1
+            filters:
+            - type: URLRewrite
+              urlRewrite:
+                path:
+                  replacePrefixMatch: /v1/responses
+                  type: ReplacePrefixMatch
+            matches:
+            - path:
+                type: PathPrefix
+                value: /{{ .ObjectMeta.Namespace }}/{{ .ObjectMeta.Name }}/v1/responses
+            timeouts:
+              backendRequest: 0s
+              request: 0s
+          - backendRefs:
             - kind: Service
               name: '{{ ChildName .ObjectMeta.Name `-kserve-workload-svc` }}'
               port: 8000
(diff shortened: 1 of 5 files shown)


RishabhSaini:
@pierDipi tagging for review

python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
vivekk16, line 1: The changes to this file don't seem necessary here. 

Since traffic in the LLMInferenceService + InferencePool path goes directly from the gateway to the backend, `OpenAIProxyModel` is never involved. `RESPONSES_ENDPOINT` and `self._responses_endpoint` are set but never actually used. 

If the goal is just passthrough, the routing config changes are enough on their own and these lines can be dropped.

 If you did want `OpenAIProxyModel` to also support the `Responses API` (so subclasses can hook into it like they do for completions and chat), then there's quite a bit more needed like a class-level annotation, `create_responses()` and `generate_responses()` methods, a handler in `OpenAIEndpoints`, and a route in `register_openai_endpoints()`.

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
```diff
@@ -1529,6 +1529,25 @@ spec:
             timeouts:
               backendRequest: 0s
               request: 0s
+          - backendRefs:
+            - group: inference.networking.k8s.io
+              kind: InferencePool
+              name: '{{ ChildName .ObjectMeta.Name `-inference-pool` }}'
+              port: 8000
+              weight: 1
+            filters:
```
vivekk16, line 1538: Minor nit: the new rule has filters before matches, and replacePrefixMatch before type, which is the opposite order compared to the kustomize source in `config-llm-router-route.yaml`. No functional impact, but worth keeping the two consistent so it's clear one is derived from the other.

python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
vivekk16: If you plan to add the changes for `OpenAIProxyModel`, it would be good to have at least one test verifying the `HTTPRoute` template generates the correct rule for `/v1/responses`, similar to how the existing endpoints are covered in `test_openai_completion.py`.

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
RishabhSaini: the ordering difference is pre-existing across all rules in this file, not introduced by this PR

python/kserve/kserve/protocol/rest/openai/openai_proxy_model.py
RishabhSaini: dropped the proxy model changes. existing route tests are path agnostic

vivekk16:
lgtm

pierDipi (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

sivanantha321 (approved):
/lgtm
/approve

PR #5302, opened 2026-03-25
fix: enforce PSS restricted profile on LLMInferenceService default template

### LINKED ISSUE #5301: [Security] Enforce Pod Security Standards (Restricted) for LLMInferenceService templates
### Problem
Currently, in KServe v0.17, LLMInferenceService]templates—such as the core—are explicitly configured with insecure defaults:
* `runAsNonRoot: false`
* `readOnlyRootFilesystem: false`
* Missing explicit `runAsUser` and `runAsGroup` bounds
* Missing `allowPrivilegeEscalation` disablement and capability dropping.

These configurations default to insecure execution because standard LLM runtime images (like base CUDA images) historically run as `USER root` by default and attempt to write cache/temporary data directly to root-owned paths (like `/tmp` or `/home`). Falling back to `runAsNonRoot: false` violates Kubernetes Pod Security Standards (Restricted policy) and makes the default LLM deployments unusable on highly secured clusters (e.g., OpenShift `restricted-v2` SCC).

alokdangre:
This PR enforces compliance with the Kubernetes Pod Security Standards (Restricted) profile and resolves specific OpenShift Admission Controller failures for KServe's LLMInferenceServiceConfigtemplates.

To overcome this securely while supporting cross-platform edge cases, this PR:
1. Sets readOnlyRootFilesystem: true and runAsNonRoot: true across all config/llmisvcconfig/config-llm*.yaml templates, and successfully syncs these bounds natively to their identical Helm charts resources.yaml
2. Exposes mapping an emptyDir scratch volume to /tmp across all configurations to safely catch telemetry data, caches, and temp files that standard ML frameworks natively dump there, ensuring compatibility with immutable filesystems.
3. Adjusts all v1alpha2 inference config preset assertions to assert the new secure defaults.

Fixes #5301 

- [x] Verified local testing deployment: The `ghcr.io/llm-d/llm-d-cuda` base template applies identically with the hardened context. 
- [x] Ran `make precommit` to assert standard formatting and verification.

- Logs: N/A

**Checklist**:

**Release note**:
Enforce Pod Security Standards Restricted profile (runAsNonRoot: true, immutable root, explicitly defined non-root UIDs) on the default LLMInferenceService template [config-llm-template.yaml]


alokdangre:
@Jooho @pierDipi PTAL

vivekk16 (commented):
@alokdangre The unit test in `pkg/controller/v1aplha2/llimsvc/` has hardcoded expected values that still assert older values `readOnlyRootFilesystem: false` and `runAsNonRoot: false`. The tests have to be updated to reflect the new `securityContext`.

Also there are other templates under `config/llmisvcconfig` that have same problem with PSS compliance. So those should be considered for hardening too.

vivekk16 (commented):
Update the helm charts as well that are identical copies of these templates. Otherwise, users installing via helm won't get the hardened config.

config/llmisvcconfig/config-llm-template.yaml
```diff
@@ -155,8 +157,9 @@ spec:
             value: /models
         securityContext:
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: false
-          runAsNonRoot: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          runAsUser: 1000
```
vivekk16: Worth noting that `runAsUser: 1000` works for upstream Kubernetes PSS Restricted, but will cause admission failures on OpenShift. The restricted-v2 SCC uses `MustRunAsRange` and only allows UIDs from the namespace's allocated range (typically values like 1000620000+), so a hardcoded 1000 will be rejected. If OpenShift compatibility is a goal, consider either fixing the container image to declare a non-root USER (so `runAsNonRoot: true` alone is sufficient) or making the UID configurable via a template variable.

alokdangre: @vivekk16 Thanks for helping with this

alokdangre:
@juliusvonkohout ,To guarantee these fixes satisfy the cluster admission controller dynamically, I created a local kind cluster tonight and mapped the strict pod-security.kubernetes.io/enforce=restricted and enforce-version=latest labels to a new test namespace.

By pulling the new hardened securityContext out of our cleaned 

config-llm-template.yaml
 (which now correctly drops ALL capabilities and sets readOnlyRootFilesystem: true + runAsNonRoot: true) and dry-running it natively through the restricted namespace, the Admission Controller successfully accepted

config/llmisvcconfig/config-llm-decode-worker-data-parallel.yaml
```diff
@@ -245,13 +245,9 @@ spec:
...
           allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: false
-          runAsNonRoot: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
           capabilities:
-            add:
-              - "IPC_LOCK"
-              - "SYS_RAWIO"
-              - "NET_RAW"
```
pierDipi, line 254: multi node DP doesn't work without these capabilities (same for other DP configs)

alokdangre: thanks for the review, I think we need to compromise on its PSS restriction

alokdangre:
@juliusvonkohout As @pierDipi noted above, multi-node GPU inference is fundamentally incompatible with the strict "no capabilities" rule of PSS Restricted due to low-layer memory pooling requirements.

Therefore, data-parallel configs remain as hardened as physically possible (runAsNonRoot: true, readOnlyRootFilesystem: true), but retain their necessary capabilities

pierDipi:
To expand a bit, the expectation is that for multi node DP the service accounts for the workers on more restrictive platforms have a role binding to a Security context constraints (SCCs, or similar equivalent procedure) that allows the workload to run with these capabilities.

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
vivekk16, line 1: The `tmp-dir` emptyDir volume and `/tmp` volumeMount were added to all 6 config templates under `config/llmisvcconfig/` (`config-llm-template`, `config-llm-decode-template`, `config-llm-prefill-template`, and the three `*-worker-data-parallel` variants), but none of these additions are reflected in this file. Without this, Helm-deployed pods will have `readOnlyRootFilesystem: true` but no writable `/tmp`, causing runtime failures when ML frameworks attempt to write temp/cache files.

pkg/controller/v1alpha2/llmisvc/config_presets_test.go
vivekk16, line 1: The unit tests don't assert the newly added tmp-dir volume and /tmp volumeMount. Kindly add the assertions.

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
alokdangre: thanks for it added now

pkg/controller/v1alpha2/llmisvc/config_presets_test.go
alokdangre: done

Jooho (commented):
/lgtm

alokdangre:
@vivekk16 i ran `make precommit`, and generated files, it passess now
@pierDipi PTAL

vivekk16 (commented):
lgtm

Jooho (approved):
/lgtm
/approved

PR #5303, opened 2026-03-25
ci: optimize Python Dockerfile builds with cache mounts

bartoszmajsak:

Optimizes all 18 Python Dockerfiles for faster incremental builds. The main bottleneck today is that any kserve source change triggers full dependency reinstalls - including expensive ML packages like vllm, flashinfer, and lmcache on the HuggingFace images.

Three key changes drive the improvement:

- **BuildKit cache mounts** for uv/pip/apt/ccache across all Dockerfiles. The lightweight images used \`--no-cache\` to avoid polluting layers (correct without mounts), while the GPU HF Dockerfile already had cache mounts but still passed \`--no-cache\` to uv - making the mounts useless. This PR adds cache mounts everywhere and drops \`--no-cache\` so they actually work.
- **Layer reordering on HF images** - ML dependencies (vllm, lmcache, flashinfer) are installed before kserve source code, so source changes don't cascade into 300+ second reinstalls. Uses \`uv sync --inexact\` to prevent uv from removing ML packages that aren't in the project's lockfile but were installed in earlier layers.
- **Storage metadata-first pattern** - only \`pyproject.toml\` and \`uv.lock\` are copied for dependency resolution instead of the full storage directory, preventing storage source changes from busting the kserve dependency layer.

Also cleans up stale patterns: deterministic uv install via \`COPY --from\` instead of \`curl | sh\`, removal of dead code, \`sharing=locked\` on apt cache mounts, and a \`.dockerignore\` to shrink build contexts.

CPU HF Dockerfile gets a deeper restructure: base split into runtime/build stages so compilers don't leak into the production image, and ccache for incremental vllm C++ builds.

**Note on two reverted changes:** The original version of this PR also dropped \`google-perftools\` (tcmalloc) from the CPU runtime image and switched \`libibverbs-dev\` to \`libibverbs1\` in the GPU image. Both turned out to be regressions caught by CI:

- The LD_PRELOAD analysis was correct (ld.so resolves left-to-right, so tcmalloc was the effective allocator and jemalloc was shadowed) — but the conclusion was wrong. Jemalloc is not a performance-equivalent replacement for tcmalloc on vllm CPU workloads. Model loading timed out in E2E tests without it. \`google-perftools\` and the original \`LD_PRELOAD\` are restored.
- \`libibverbs1\` (runtime-only) caused vllm engine core init to crash. \`libibverbs-dev\` is restored.

Benchmark results ([full details with script](https://gist.github.com/bartoszmajsak/76f29289e95a1a8d3f01f02d75a87847)):

| Image | Scenario | Master | This PR | Speedup |
|---|---|---|---|---|
| custom_model | kserve source change | 99s | 66s | 33% faster |
| custom_model | storage source change | 92s | 1s | 92x faster |
| HF GPU | kserve source change | 561s | 220s | 61% faster |
| HF GPU | hfserver source change | 334s | 100s | 70% faster |

Cold builds are roughly the same (cache mounts start empty, reorder doesn't help first build). All gains are on incremental rebuilds - the common case during development and CI.

Functional equivalence verified by comparing installed packages between master and feature images. GPU image differs only in numpy patch version (2.2.0 vs 2.2.4). CPU image is identical.

- [x] All 18 Dockerfiles build successfully
- [x] GPU image package parity verified (213 packages, only numpy patch diff)
- [x] CPU image package parity verified (zero diff)
- [x] Incremental build benchmarks measured on cold-pruned docker cache

The HF Dockerfiles use \`uv sync --active --inexact\` after the ML dependency block. The \`--inexact\` flag prevents \`uv sync\` from removing packages (vllm, flashinfer, etc.) that aren't in the project's lockfile but were installed in earlier layers. Without it, the reorder silently strips ML dependencies from the final image.

**Checklist**:

\`\`\`release-note
NONE
\`\`\`

### DIFF
--- a/.github/workflows/huggingface-vllm-docker-publish-manual.yml
+++ b/.github/workflows/huggingface-vllm-docker-publish-manual.yml
@@ -19,11 +19,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: 
+        image:
           - version: ${{ inputs.version }}
             path: 'python/huggingface_server_cpu.Dockerfile'
+            cache-scope: hf-cpu
           - version: ${{ inputs.version }}-gpu
             path: 'python/huggingface_server.Dockerfile'
+            cache-scope: hf-gpu
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source
@@ -67,7 +69,9 @@ jobs:
           context: python
           file: ${{ matrix.image.path }}
           push: true
-          tags: ${{ env.IMAGE_ID }}:${{ env.VERSION }} 
+          tags: ${{ env.IMAGE_ID }}:${{ env.VERSION }}
           # https://github.com/docker/buildx/issues/1533
           provenance: false
           sbom: true
+          cache-from: type=gha,scope=${{ matrix.image.cache-scope }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.image.cache-scope }}
--- a/python/.dockerignore
+++ b/python/.dockerignore
@@ -1,14 +1,16 @@
-# Exclude dev artifacts from Python image builds (context: python/)
-**/.venv/
+# Exclude test/doc/build artifacts from all allowed directories
 **/__pycache__/
 **/*.pyc
 **/*.pyo
-**/.pytest_cache/
-**/.ruff_cache/
-**/.mypy_cache/
 **/*.egg-info/
+**/.pytest_cache/
+**/tests/
+**/test/
+**/docs/
+**/.openapi-generator/
+**/.venv/
+**/build/
 **/dist/
-kserve/test/
-kserve/docs/
-**/.idea/
-**/.vscode/
+**/.mypy_cache/
+**/.ruff_cache/
+**/third_party/library/
--- a/python/aiffairness.Dockerfile
+++ b/python/aiffairness.Dockerfile
@@ -5,12 +5,11 @@ ARG VENV_PATH=/prod_venv
 FROM ${BASE_IMAGE} AS builder
 
 # Required for building packages for arm64 arch
-RUN apt-get update && apt-get install -y --no-install-recommends curl python3-dev build-essential && apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+RUN --mount=type=cache,sharing=locked,target=/var/cache/apt --mount=type=cache,sharing=locked,target=/var/lib/apt/lists \
+    apt-get update && apt-get install -y --no-install-recommends python3-dev build-essential
 
 # Install uv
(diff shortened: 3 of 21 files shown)


Jooho:
rebased.

sivanantha321:
@bartoszmajsak the test is failing with following error

```
        bfloat16
      State:       Waiting
        Reason:    CrashLoopBackOff
      Last State:  Terminated
        Reason:    Error
        Message:    361, in start
      asyncio.run(self._serve())
    File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
      return loop.run_until_complete(main)
    File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
      return future.result()
    File "/prod_venv/lib/python3.10/site-packages/kserve/model_server.py", line 316, in _serve
      await asyncio.gather(*self.servers)
    File "/huggingfaceserver/huggingfaceserver/vllm/vllm_model.py", line 109, in start_engine
      async with build_async_engine_client_from_engine_args(
    File "/usr/lib/python3.10/contextlib.py", line 199, in __aenter__
      return await anext(self.gen)
    File "/huggingfaceserver/huggingfaceserver/vllm/utils.py", line 97, in build_async_engine_client_from_engine_args
      async_llm = AsyncLLM.from_vllm_config(
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/async_llm.py", line 228, in from_vllm_config
      return cls(
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/async_llm.py", line 155, in __init__
      self.engine_core = EngineCoreClient.make_async_mp_client(
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/core_client.py", line 122, in make_async_mp_client
      return AsyncMPClient(*client_args)
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/core_client.py", line 819, in __init__
      super().__init__(
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/core_client.py", line 479, in __init__
      with launch_core_engines(vllm_config, executor_class, log_stats) as (
    File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
      next(self.gen)
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/utils.py", line 933, in launch_core_engines
      wait_for_engine_startup(
    File "/prod_venv/lib/python3.10/site-packages/vllm/v1/engine/utils.py", line 992, in wait_for_engine_startup
      raise RuntimeError(
  RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}
  Error in sys.excepthook:
  ```

bartoszmajsak:
/hold

Needs further investigation

PR #5305, opened 2026-03-26
fix(llmisvc): migrate tokenProcessorConfig and blockSize

pierDipi:
In llm-d-inference-scheduler v0.6.0, the precise-prefix-cache-scorer plugin changed its config schema: tokenProcessorConfig moved from being nested inside indexerConfig to a top-level field under the plugin parameters. Without migration, the scorer silently falls back to defaults (blockSize: 16) and produces zero scores, breaking prefix cache routing entirely.

Add WithMigrateTokenProcessorConfig mutation that detects tokenProcessorConfig nested inside indexerConfig and moves it to the top level of the plugin parameters. The migration skips if a top-level tokenProcessorConfig already exists to avoid overwriting explicit user configuration. Update sample docs and e2e fixtures to use the new schema.

the prefix-cache-scorer plugin renamed the blockSize field (characters) to blockSizeTokens (tokens).
If only the deprecated blockSize is present the plugin refuses to start. Add WithMigrateBlockSizeToBlockSizeTokens mutation that copies blockSize to blockSizeTokens when the new field is not already set.

### DIFF
--- a/docs/samples/llmisvc/precise-prefix-kv-cache-routing/llm-inference-service-qwen2-7b-gpu-kv-cache-routing.yaml
+++ b/docs/samples/llmisvc/precise-prefix-kv-cache-routing/llm-inference-service-qwen2-7b-gpu-kv-cache-routing.yaml
@@ -19,12 +19,12 @@ spec:
             - type: single-profile-handler
             - type: precise-prefix-cache-scorer
               parameters:
+                tokenProcessorConfig:
+                  blockSize: 64       # Must match vLLM --block-size (default is 16)
+                  hashSeed: "42"      # Must match PYTHONHASHSEED in vLLM pods
                 kvEventsConfig:
                   zmqEndpoint: "tcp://*:5557"
                 indexerConfig:
-                  tokenProcessorConfig:
-                    blockSize: 64       # Must match vLLM --block-size (default is 16)
-                    hashSeed: "42"      # Must match PYTHONHASHSEED in vLLM pods
                   kvBlockIndexConfig:
                     enableMetrics: true                   # Enable Prometheus metrics for KV cache index
                     metricsLoggingInterval: 60000000000   # Log metrics every 60 seconds (in nanoseconds)
--- a/pkg/controller/v1alpha2/llmisvc/controller_int_scheduler_config_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/controller_int_scheduler_config_test.go
@@ -1141,6 +1141,9 @@ schedulingProfiles:
 				g.Expect(found).To(BeTrue(), "Expected to find --config-text in scheduler deployment")
 				g.Expect(configText).To(ContainSubstring("modelName: base"))
 				g.Expect(configText).To(ContainSubstring("socketFile: /tmp/tokenizer/tokenizer-uds.socket"))
+				// Verify tokenProcessorConfig was migrated from indexerConfig to top-level parameters
+				g.Expect(configText).To(ContainSubstring("tokenProcessorConfig"))
+				g.Expect(configText).To(ContainSubstring("blockSize: 16"))
 				return nil
 			}).WithContext(ctx).Should(Succeed())
 		})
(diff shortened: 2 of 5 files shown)


pkg/controller/v1alpha2/llmisvc/controller_int_scheduler_config_test.go
```diff
@@ -1267,6 +1273,134 @@ schedulingProfiles:
 		})
 	})
 
+	Context("tokenProcessorConfig migration from indexerConfig to top-level parameters", func() {
```
bartoszmajsak, line 1276: Maybe as follow-up we could:

- ensure we don't have overlapping test cases and slice it to a few files based on areas/contexts; this is getting big
- since mutating functions are exported maybe we could unit test them for more edge cases

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -821,6 +825,89 @@ func WithUdsTokenizerConfig(u *unstructured.Unstructured) error {
...
+
+	for _, plugin := range plugins {
+		pluginMap, ok := plugin.(map[string]interface{})
+		if !ok || pluginMap["type"] != precisePrefixCacheScorerPlugin {
+			continue
+		}
+
+		// Skip if tokenProcessorConfig already exists at the top level
+		if _, exists, _ := unstructured.NestedFieldNoCopy(pluginMap, "parameters", "tokenProcessorConfig"); exists {
+			continue
```
bartoszmajsak, line 852: Worth signaling this in the logs at least?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -821,6 +825,89 @@ func WithUdsTokenizerConfig(u *unstructured.Unstructured) error {
...
+		}
+
+		// Check if tokenProcessorConfig exists inside indexerConfig (old location)
+		tpc, found, err := unstructured.NestedFieldCopy(pluginMap, "parameters", "indexerConfig", "tokenProcessorConfig")
+		if err != nil || !found {
+			continue
```
bartoszmajsak, line 862: Log error?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -821,6 +825,89 @@ func WithUdsTokenizerConfig(u *unstructured.Unstructured) error {
 	return nil
 }
 
+// WithMigrateTokenProcessorConfig migrates tokenProcessorConfig from inside
+// indexerConfig to the top level of the plugin parameters for the
+// precise-prefix-cache-scorer plugin. This handles the schema change in
+// llm-d-inference-scheduler v0.6.0 where tokenProcessorConfig was promoted
+// from indexerConfig to a top-level plugin parameter.
+func WithMigrateTokenProcessorConfig(u *unstructured.Unstructured) error {
```
bartoszmajsak: Maybe move them all to `scheduler_migration.go`?

docs/samples/llmisvc/precise-prefix-kv-cache-routing/llm-inference-service-qwen2-7b-gpu-kv-cache-routing.yaml
```diff
@@ -19,12 +19,12 @@ spec:
             - type: single-profile-handler
             - type: precise-prefix-cache-scorer
               parameters:
+                tokenProcessorConfig:
+                  blockSize: 64       # Must match vLLM --block-size (default is 16)
```
bartoszmajsak, line 23: Maybe a silly question, but do we have a way to ensure it / report mismatch?

bartoszmajsak (commented):
I like the composable migrations!

Few minor comments for your consideration, but non-blocking really.

docs/samples/llmisvc/precise-prefix-kv-cache-routing/llm-inference-service-qwen2-7b-gpu-kv-cache-routing.yaml
pierDipi: you mean report mismatch with the vLLM config? Not yet

bartoszmajsak: Yeah. That's what I meant - is tracked somewhere?

bartoszmajsak (approved):
/lgtm
/approved

israel-hdez (approved):
/lgtm

PR #5308, opened 2026-03-27
fix(llmisvc): remove cert-hash restart annotation from scheduler

bartoszmajsak:

The scheduler now handles certificate reloading internally via its --enable-cert-reload flag, eliminating the need for the controller to trigger restarts through a cert-hash pod template annotation.

This also removes a source of unnecessary scheduler churn, as the hash covered the full certificate including workload pod IPs as SANs.

**Release note**:
```release-note
NONE
```


pierDipi (approved):
/lgtm

vivekk16 (commented):
Very clean code changes.
lgtm

sivanantha321 (approved):
/lgtm
/approve

PR #5310, opened 2026-03-27
fix(llmisvc): rename EPP port name to tcp-ext-proc for Istio compatibility

janghyukjin:

When Istio sidecar injection is enabled in the cluster, the EPP (Endpoint Picker) pod gets an Istio sidecar injected. Istio uses the container port name to determine traffic handling: port names starting with \`grpc\`, \`http\`, or \`http2\` trigger L7 HTTP/2 processing, while names starting with \`tcp\` are treated as raw TCP (L4 passthrough).

With the port named \`grpc\`, the Istio sidecar injected into the EPP pod intercepts incoming traffic and attempts to apply HTTP/2 L7 parsing to the ext-proc stream. The stream does not match the HTTP/2 format Istio expects, so it is rejected with \`http2.invalid.header.field\`, which causes the ext-proc connection to terminate and the gateway to return HTTP 500.

Renaming the port to \`tcp-ext-proc\` tells Istio to handle it as raw TCP, preventing HTTP/2 parsing of the ext-proc stream.

> **Note:** This issue is not specific to Istio Gateway. It occurs whenever Istio sidecar injection is enabled in the EPP pod's namespace, regardless of whether Istio Gateway or Envoy Gateway is used as the gateway component.

**Related upstream issues:**
- istio/istio#14909 — port named \`grpc\` triggers HTTP/2 L7 processing
- istio/istio#59577 — same issue observed with LLMInferenceService + Istio + ext-proc

N/A

Verified on a cluster with Istio 1.28.3 (\`STRICT\` mTLS PeerAuthentication), inference-gateway (Envoy-based) → EPP (port 9002) → vLLM.

- [x] Before fix: HTTP 500 with \`http2.invalid.header.field\` in EPP pod logs
- [x] After fix: HTTP 200 OK, E2E inference working correctly

The \`tcp-\` prefix is the standard Istio convention for raw TCP ports. This does not affect gRPC functionality — it only prevents Istio from applying HTTP/2 L7 processing to the ext-proc stream.

**Checklist**:

**Release note**:
\`\`\`release-note
Fix EPP ext-proc port name from \`grpc\` to \`tcp-ext-proc\` to prevent Istio from applying HTTP/2 L7 processing, which caused \`http2.invalid.header.field\` errors and HTTP 500 responses when Istio sidecar injection is enabled.
\`\`\`

### DIFF
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -27,6 +27,8 @@ spec:
           - name: main
             ports:
               - containerPort: 9002
+                # "tcp-" prefix prevents Istio from applying HTTP/2 L7 parsing to the ext-proc port.
+                # The port still uses gRPC; this only affects Istio protocol classification.
                 name: tcp-ext-proc
                 protocol: TCP
               - containerPort: 9003
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -27,7 +27,7 @@ spec:
           - name: main
             ports:
               - containerPort: 9002
-                name: grpc
+                name: tcp-ext-proc
                 protocol: TCP
               - containerPort: 9003
                 name: grpc-health
--- a/pkg/controller/v1alpha2/llmisvc/controller_int_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/controller_int_test.go
@@ -476,13 +476,13 @@ var _ = Describe("LLMInferenceService Controller", func() {
 					eppSvc := &corev1.Service{}
 					g.Expect(envTest.Client.Get(ctx, client.ObjectKey{Name: svcName + "-epp-service", Namespace: llmSvc.GetNamespace()}, eppSvc)).To(Succeed())
 
-					// Verify all expected ports are present (grpc, grpc-health, metrics, zmq)
+					// Verify all expected ports are present (tcp-ext-proc, grpc-health, metrics, zmq)
 					portNames := make(map[string]int32)
 					for _, port := range eppSvc.Spec.Ports {
 						portNames[port.Name] = port.Port
 					}
 
-					g.Expect(portNames).To(HaveKeyWithValue("grpc", int32(9002)))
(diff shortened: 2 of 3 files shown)


vivekk16 (commented):
lgtm

One small suggestion: add a short comment explaining the renaming (why the port uses `tcp-` rather than `grpc-`), so a future contributor doesn't revert it thinking it's a mistake.

janghyukjin:
@vivekk16 , Thanks for the suggestion!

Added a short comment in both config-llm-scheduler.yaml and scheduler.go 
to clarify why the "tcp-" prefix is used instead of "grpc".  :)

janghyukjin:
@vivekk16

After additional testing, I'm no longer confident that renaming the ext-proc port from `grpc-*` to `tcp-*` is a generally safe fix.

In our environment:
- `tcp-*` avoids sidecar L7 interpretation but can break gateway behavior (loss of HTTP/2 hint)
- `grpc-*` works better for the gateway (ensuring HTTP/2), but we also observed intermittent failures depending on the mesh configuration

We also observed that the llm-d inference scheduler binary defaults `--secureServing` to `true` when not explicitly set. This interacts with Istio mTLS and may contribute to connection failures (e.g., gRPC status 14), suggesting the issue involves multiple interacting factors beyond port naming alone.

Given this, I think this PR in its current form may not be appropriate as a general fix. I'll follow up with a minimal reproducible case and investigation instead.

Closing this for now based on the above findings.

janghyukjin:
@vivekk16 :)
After further investigation, I'm not fully confident that this rename alone is a safe general fix, so I'm keeping this PR closed for now.

However, I'd appreciate feedback from maintainers on the following:

- The grpc port name seems to trigger Istio L7 classification on the destination side, which can interfere with ext-proc communication.
- Renaming the port (e.g. to ext-proc) avoids that, but when port 9002 is exposed via a Service, some Istio environments may require additional outbound HTTP/2 configuration (e.g. EnvoyFilter) on the gateway side.

So it seems like there are two separate layers involved:
1. Port naming / protocol classification (KServe scope)
2. Gateway outbound protocol configuration (deployment scope)

Would it make sense for KServe to at least adopt a role-based port name like ext-proc, even if additional configuration may still be required in some environments?

Happy to follow up if this direction seems worth pursuing.

janghyukjin:
@vivekk16 Opening an issue for broader discussion: #5331

PR #5313, opened 2026-03-28
feat(ingress): add DisableHTTPRouteTimeout config flag for GKE Gateway

### LINKED ISSUE #5311: HTTPRoute timeouts field breaks GKE Gateway controller — need option to disable
**Title:** HTTPRoute `timeouts` field breaks GKE Gateway controller — need option to disable

/kind bug

When using KServe v0.17.0 in Raw Deployment mode with Gateway API on GKE, the HTTPRoute reconciler unconditionally sets `spec.rules[*].timeouts.request: "60s"` on every generated HTTPRoute (`httproute_reconciler.go`, `createHTTPRouteRule()`).

GKE's Gateway controller does not implement the `spec.rules.timeouts` field ([GKE GatewayClass capabilities](https://cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities)). It rejects the route with:

```
status:
  parents:
  - conditions:
    - type: Accepted
      status: "False"
      reason: UnsupportedValue
```

This causes `reconcileHTTPRouteStatus()` to set `IngressReady=False`, so external ingress never comes up.

The timeout is hardcoded as a package-level default:
```go
var DefaultTimeout = toGatewayAPIDuration(60)
```

And `createHTTPRouteRule()` always wraps it in a non-nil `Timeouts` struct — there is no code path that omits the field. Setting `TimeoutSeconds` on the InferenceService spec can change the value but cannot remove the field.

Manual workaround (`kubectl patch --type=json` to strip the timeouts field) is temporary — the reconciler re-adds it on the next loop.

HTTPRoutes generated by KServe should be accepted by the GKE Gateway controller. Ideally, the timeout field should be configurable/optional so it can be disabled for Gateway implementations that don't support it.

**Proposed fix:** Add a `disableHTTPRouteTimeout` flag to `inferenceservice-config` (ingress section). When set, the reconciler skips setting `Timeouts` on HTTPRouteRules. This is backwards-compatible — default behavior is unchanged for controllers that support it (Envoy Gateway, Istio).

```yaml
kind: InferenceService
metadata:
  name: distilbert-sentiment
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      storageUri: hf://distilbert/distilbert-base-uncased-finetuned-sst-2-english
      resources:
        requests:
          cpu: "1"
          memory: 2Gi
        limits:
          cpu: "2"
          memory: 4Gi
      args:
        - --task=sequence_classification
```

- GKE handles timeouts via `GCPBackendPolicy` (per-Service `timeoutSec`), not via HTTPRoute `timeouts` field.
- The Gateway API spec marks `timeouts` as "extended" conformance, meaning implementations are not required to support it ([GEP-1742](https://gateway-api.sigs.k8s.io/geps/gep-1742/)).
- Relevant code: `pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go` — `DefaultTimeout`, `createHTTPRouteRule()`, and all `createRaw*HTTPRoute()` functions.
- Related PR: [#3952](https://github.com/kserve/kserve/pull/3952) (introduced Gateway API support).

**Environment:**

- Istio Version: N/A (not using Istio, using GKE Gateway API)
- Knative Version: N/A (Raw Deployment mode)
- KServe Version: 0.17.0
- Kubeflow version: N/A
- Cloud Environment: GCP (GKE)
- Minikube/Kind version: N/A
- Kubernetes version: v1.31 (GKE)
- OS: GKE managed nodes (Container-Optimized OS)

sophieliu15:
## Summary

GKE's Gateway controller does not support `spec.rules.timeouts` on HTTPRoutes. KServe v0.17.0 unconditionally sets `timeouts: {request: 60s}` on every HTTPRoute via `httproute_reconciler.go`, causing GKE Gateway to reject the route with `Accepted: False` / `UnsupportedValue`. There is currently no annotation, config flag, or InferenceService field to disable this.

This PR adds a `disableHTTPRouteTimeout` field to the `inferenceservice-config` ConfigMap (ingress section). When set to `true`, the controller omits the `Timeouts` field entirely from HTTPRoutes, allowing GKE Gateway (and any other controller that doesn't implement this optional field to accept the route.

Default behavior is **unchanged** — existing users who rely on the implicit 60s timeout are unaffected.

Fixes: #5311

## Changes

  - pkg/apis/serving/v1beta1/configmap.go — Added `DisableHTTPRouteTimeout bool`  to `IngressConfig`
  - pkg/controller/.../httproute_reconciler.go — Added `resolveTimeout()` helper  that returns nil when the flag is set; modified `createHTTPRouteRule()` to only populate `rule.Timeouts` when timeout is non-nil; updated all 9 timeout
assignment blocks across `createRawPredictorHTTPRoute`,`createRawTransformerHTTPRoute`,`createRawExplainerHTTPRoute`, and
  `createRawTopLevelHTTPRoute`
  - pkg/controller/.../httproute_reconciler_test.go — Added 3 new test functions(6 cases): `TestResolveTimeout`, `TestCreateHTTPRouteRuleNilTimeout`, `TestCreateRawPredictorHTTPRouteDisableTimeout`

## Usage

```
 // inferenceservice-config ConfigMap, ingress section:

   {

     "enableGatewayApi": true,

     "kserveIngressGateway": "kserve/kserve-ingress-gateway",

     "disableHTTPRouteTimeout": true

   }
```

## Test plan

- [x] Unit tests: all new and existing tests passed
- [x] **Manual cluster test on GKE**: Built a custom controller image from this branch, deployed to a GKE cluster (`v1.34.4-gke.1130000`, Gateway class `gke-l7-global-external-managed`), and verified:
  - HTTPRoute `spec.rules[*].timeouts`  is **absent** when `disableHTTPRouteTimeout: true`
  - GKE Gateway accepts the HTTPRoute (`Accepted: True`)
  - End-to-end inference works through the external load balancer: `curl → GKE Gateway → InferenceService → {"predictions":[1,0]}` 
  - Without the fix (stock v0.17.0), the same HTTPRoute is rejected with by Gateway with  `UnsupportedValue`


vivekk16 (commented):
Code changes lgtm

Two suggestions

- `config/configmap/inferenceservice.yaml` and `config/overlays/test/configmap/inferenceservice.yaml` don't mention the new flag. It will be good to add a commented-out entry there like the other ingress fields, so the new flag can be documented and be actually discovered without reading source code.
- It might be worth wiring this through the Helm chart too — adding it to `charts/kserve-resources/values.yaml` and the ConfigMap template so Helm users can set it via `--set` or `values.yaml`. Without that, they will have to manually patch the ConfigMap after install, which gets overwritten on the next `helm upgrade`.

sophieliu15:
Thanks for the suggestions @vivekk16! I've updated the PR:
- Added `disableHTTPRouteTimeout` to `config/configmap/inferenceservice.yaml` and `config/overlays/test/configmap/inferenceservice.yaml` with inline documentation
- Wired the flag through both Helm charts (`kserve-resources` and `kserve-llmisvc-resources`) — values.yaml, ConfigMap templates, and READMEs

sivanantha321:
@sophieliu15 Can you help resolve the conflicts?

sivanantha321:
@sophieliu15 Can you also update the docs here https://github.com/kserve/website/blob/main/docs/admin-guide/configurations.md ?

sophieliu15:
@sivanantha321 Both changes are done:
- Conflicts with master have been resolved.
- Docs PR created at https://github.com/kserve/website/pull/635 to add `disableHTTPRouteTimeout` to the ingress configuration guide.

sivanantha321:
@sophieliu15 Can you run `make precommit` and commit the changes ?

sophieliu15:
@sivanantha321 Done — ran `make precommit` and committed the generated file changes.

sivanantha321 (approved):
/lgtm
/approve

PR #5314, opened 2026-03-29
fix: add storageContainerName field for explicit CSC selection

### LINKED ISSUE #5299: ClusterStorageContainer is completely ignored
/kind bug

The defined ClusterStorageContainer is ignored by InferenceService (or LLMInferenceService, which is a known bug). Neither the environment variables (such as HF_TOKEN), nor container limits are applied.

Expected steps in the [documentation](https://kserve.github.io/website/docs/getting-started/genai-first-isvc) to produce described result.

```yaml
kind: InferenceService
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"serving.kserve.io/v1beta1","kind":"InferenceService","metadata":{"annotations":{},"name":"unsloth-nemotron-30b-a3b-nvfp4","namespace":"default"},"spec":{"predictor":{"model":{"modelFormat":{"name":"huggingface"},"resources":{"limits":{"cpu":"6","memory":"16Gi","nvidia.com/gpu":"1"}},"storageUri":"hf://unsloth/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4"}}}}
    modelFormat: huggingface
    serving.kserve.io/deploymentMode: Standard
  creationTimestamp: "2026-03-24T15:15:12Z"
  finalizers:
  - inferenceservice.finalizers
  generation: 1
  name: unsloth-nemotron-30b-a3b-nvfp4
  namespace: default
  resourceVersion: "106144"
  uid: 8c1bf192-10f7-4370-acc5-a84dcdb1c413
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      name: ""
      resources:
        limits:
          cpu: "6"
          memory: 16Gi
          nvidia.com/gpu: "1"
      storageUri: hf://unsloth/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4
status:
  clusterServingRuntimeName: kserve-huggingfaceserver
  components:
    predictor: {}
  conditions:
  - lastTransitionTime: "2026-03-24T15:15:12Z"
    reason: Predictor ingress not created
    status: "False"
    type: IngressReady
  - lastTransitionTime: "2026-03-24T15:15:13Z"
    message: Deployment does not have minimum availability.
    reason: MinimumReplicasUnavailable
    status: "False"
    type: PredictorReady
  - lastTransitionTime: "2026-03-24T15:18:09Z"
    reason: Predictor ingress not created
    status: "False"
    type: Ready
  - lastTransitionTime: "2026-03-24T15:15:12Z"
    severity: Info
    status: "False"
    type: Stopped
  deploymentMode: Standard
  modelStatus:
    copies:
      failedCopies: 0
    lastFailureInfo:
      exitCode: 137
      message: |
        2026-03-24 15:17:57.950 1 storage.initializer INFO [initializer-entrypoint:<module>():16] Initializing, args: (src_uri, dest_path): [('hf://unsloth/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4', '/mnt/models')]
        2026-03-24 15:17:57.950 1 storage.initializer INFO [kserve_storage.py:download():161] Copying contents of hf://unsloth/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4 to local
      reason: ModelLoadFailed
    states:
      activeModelState: ""
      targetModelState: FailedToLoad
    transitionStatus: BlockedByFailedLoad
  observedGeneration: 1
```

The CSC:

```yaml
kind: ClusterStorageContainer
metadata:
  name: hf-hub
spec:
  container:
    name: storage-initializer
    image: kserve/storage-initializer:latest
    env:
    - name: HF_TOKEN
      valueFrom:
        secretKeyRef:
          name: hf-secret
          key: HF_TOKEN
          optional: false
    resources:
      requests:
        memory: 2Gi
        cpu: "1"
      limits:
        memory: 4Gi
        cpu: "1"
  supportedUriFormats:
    - prefix: hf://
```

**Environment:**

- KServe Version: 0.17.0
- Cloud Environment: upcloud
- Kubernetes version: (use `kubectl version`): v1.34.2

RunlinXiao:
This PR is to provide optional CSC specification when multiple CSC supports the same storage. Currently auto-match returns first alphabetical CSC when name not specified and user will have no flexibility on selecting the CSC to be used, this change will be backward compatibility 

fixes #5299

- Added unit tests

- We can add similar support for `llmisvc` in a separate PR once this is approved
- I have considered per-URI storageContainerName (on each StorageUri entry), but the current reconciler uses a single init container for all StorageUris — [it only looks at StorageUris[0] to determine the container spec](https://github.com/kserve/kserve/blob/v0.17.0/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L774). Per-URI CSC selection would require multiple init containers, which is a bigger change. I believe a single field at the PredictorExtensionSpec level is the correct scope for now. 

**Checklist**:

**Release note**:
```release-note
Users can now specify `ClusterStorageContainer` sources to be used for the InferenceServiceusing the new `storageContainerName` field. This enables use cases like multiple CSC supports the same format, and user would be able to specify which one to use instead of a auto-match behavior. The field is optional for backwards compatibility.

  Example:
  ```yaml
spec:
  predictor:
    storageContainerName: hf-hub
    model:
      modelFormat:
        name: huggingface
      name: ""
      resources:
        limits:
          cpu: "6"
          memory: 16Gi
          nvidia.com/gpu: "1"
      storageUri: hf://unsloth/NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4

  Note: referenced `ClusterStorageContainer` must exist
```


ybenmosh:
Hi @RunlinXiao, this PR is the solution for the exact problem we've hit (CSC basically being ignored). Is this expected to be merged any time soon? would be happy to assist if this can be useful in any way.
Thanks!

RunlinXiao:
> Hi @RunlinXiao, this PR is the solution for the exact problem we've hit (CSC basically being ignored). Is this expected to be merged any time soon? would be happy to assist if this can be useful in any way. Thanks!

Hi @ybenmosh , I will follow up with the team and address any feedback provided, hopefully get it merged soon!

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -59,6 +59,13 @@ type PredictorSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	StorageUris []StorageUri `json:"storageUris,omitempty"`
 
+	// StorageContainerName specifies the name of the ClusterStorageContainer to use
+	// for downloading model artifacts. When set, this ClusterStorageContainer is used
+	// instead of auto-matching based on the storage URI scheme.
+	// This is similar to storageClassName on PersistentVolumeClaims.
+	// +optional
+	StorageContainerName *string `json:"storageContainerName,omitempty"`
+
```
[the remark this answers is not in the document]
RunlinXiao: Updated the release note

cjohannsen-cloudera (approved):
lgtm

RunlinXiao:
Hi @Jooho could you help merge the PR if it looks good to you as well, thanks!

RunlinXiao:
Hi @cjohannsen-cloudera @Jooho  — Would you mind merging this when you get a chance? Thanks!

ybenmosh:
> Hi @cjohannsen-cloudera @Jooho — Would you mind merging this when you get a chance? Thanks!

Hi again, and thanks @RunlinXiao for the contribution! We have been waiting for this feature to be merged, and I can see @cjohannsen-cloudera approved it a while ago. Any hint on when this can be merged? Thanks!

RunlinXiao:
Hi @ybenmosh, I have followed up offline in Slack with the team to , the PR will be merged once the workflow is triggered and passed.

spolti:
@cjohannsen-cloudera @RunlinXiao can please any of you provide the docs updates for this feature?

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -649,7 +693,11 @@ func (mi *StorageInitializerInjector) InjectStorageInitializer(ctx context.Conte
 	storageURIs := []v1beta1.StorageUri{{Uri: srcURI, MountPath: constants.DefaultModelLocalMountPath}}
 
 	// Get storage container spec for the URI
-	storageContainerSpec, err := GetStorageContainerSpec(ctx, srcURI, mi.client)
+	var storageContainerName *string
+	if name, ok := pod.Annotations[constants.StorageContainerNameAnnotationKey]; ok {
```
spolti, line 697: Is there a need to add a new annotation for it? Wouldn't it be easier just query the CSC by name, if found, ok, if not default it and emit a event?

RunlinXiao:
> @cjohannsen-cloudera @RunlinXiao can please any of you provide the docs updates for this feature?

Hi @spolti , yes I will take care of the doc!

pkg/webhook/admission/pod/storage_initializer_injector.go
RunlinXiao: Thanks for reviewing! When you said query the CSC by name, I'd assume you mean from the `inferenceService` spec, but we are in the webhook path not the controller part, and all we got is a pod and context: https://github.com/kserve/kserve/pull/5314/changes#diff-19ab20f87e5f9bff37c2eada3e341c832cf91029b78a75b9f70acf6a22dacb8cL608.

Also for the fall back, this is something explicitly configured, I think a fast fail would be better than fall back to default, which aligns better with k8s convention IMHO but happy to discuss further if needed(e.g. when a secret or config map not found k8s fail it and not default to empty value)

spolti: Hi @RunlinXiao, thanks for your reply, adding annotation might not be necessary, we just want to make sure that the csc name in the isvc exists and is usable, if thats the case, ok use it, otherwise, we can either give a warning to not block isvc deployment, and, use the default csc or just don't use nothing.

RunlinXiao: Hi @spolti, thanks for the follow-up! 

> adding annotation might not be necessary, we just want to make sure that the csc name in the isvc exists and is usable

Digging in a bit more, it seems like the mutator fetched the isvc, so in principle the injector could read name from there rather than via the annotation and indeed it's not strictly inaccessible.

That said, I'd still lean toward keeping the annotation:
- The whole controller→webhook storage contract already flows through pod annotations: the [storageUri](https://github.com/kserve/kserve/blob/854688ac0a5cfaa8ec843eb0f1785be9e79faefe/pkg/webhook/admission/pod/storage_initializer_injector.go#L654) itself (`StorageInitializerSourceUriInternalAnnotationKey`), and the storage-spec. To me the `InjectStorageInitializer` is a pure function of the pod by design, and the CSC name as a sibling annotation keeps that boundary intact.
- Also the isvc lookup is best-effort/non-fatal. If CSC selection depended on it, a failed/slow GET would silently drop the user's choice while the annotation is part of the admission payload, so it's always presented.

>if thats the case, ok use it, otherwise, we can either give a warning to not block isvc deployment, and, use the default csc or just don't use nothing.

I'm fully with you on surfacing this more clearly. Today in the single-URI webhook path a bad name only fail at pod admission, which might not obvious to user. I'd like to add a controller-side validation (in [addStorageInitializerAnnotations](https://github.com/kserve/kserve/blob/854688ac0a5cfaa8ec843eb0f1785be9e79faefe/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L279-L280)) so any issue with the CSC surfaces as a clear `InternalError` event on the isvc, which is the same way the multi-URI path already does.

My one hesitation is silently falling back to the default/auto-matched CSC: naming a CSC pins a specific downloader image (and any env/resources it carries), and the whole premise of #5299 is that multiple CSCs match the same prefix and auto-match picks the "wrong" one, so a user who explicitly names one is doing it because the choice matters in their setup so silent-wrong feels worse than loud-failed here, and fail-fast here don't take down a running service. 

Either way, I think it's a judgment call, curious to hear your thoughts, and happy to adjust if you feel strongly about the fallback.

PR #5317, opened 2026-03-30
feat(llmisvc): add reconciliation logic for static lora adapters

andresllh:

This PR adds support for static LoRA (Low-Rank Adaptation) adapters in LLMInferenceService. LoRA adapters enable efficient fine-tuning of large language models by applying adapter weights on top of a base model without modifying the base model itself.

The implementation includes:
- New LoRA API spec that allows defining multiple adapters via `spec.model.lora.adapters`
- Reconciliation logic to configure vLLM with LoRA-specific flags (`--enable-lora`, `--max-lora-rank`, `--lora-modules`)
- Support for multiple storage backends:
  - `hf://` (Hugging Face) and `s3://` URIs via storage initializer
  - `pvc://` URIs via PVC volume mounts
- Validation to ensure adapter names are unique and don't conflict with the base model name
- Mount handling for LoRA adapters at `/mnt/lora/<adapter-name>`
- Unit tests for the reconciliation logic

N/A

- [x] Added unit tests for LoRA adapter reconciliation logic
- [x] Tested with hf://, s3://, and pvc:// URI schemes
- [x] Validated vLLM container args include correct LoRA flags
- [x] E2E tests pending

This is the initial implementation for static LoRA adapter support. Dynamic adapter loading/unloading at runtime is not part of this PR.

The LoRA adapters are mounted at `/mnt/lora` instead of under the base model path (`/mnt/models`) because the base model is mounted read-only and nested volume mounts would require mkdir permissions.

**Checklist**:

**Release note**:
```release-note
Add support for static LoRA adapters in LLMInferenceService with hf://, s3://, and pvc:// URI schemes
```


pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
```diff
@@ -334,6 +343,9 @@ func (r *LLMISVCReconciler) expectedPrefillMultiNodeLWS(ctx context.Context, llm
 			if err := r.attachModelArtifacts(ctx, serviceAccount, llmSvc, currLWS.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec, &expected.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec, config, "main", constants.DefaultModelLocalMountPath); err != nil {
 				return nil, fmt.Errorf("failed to attach model artifacts to prefill worker template: %w", err)
 			}
+			if err := r.attachLoRAAdapters(ctx, serviceAccount, llmSvc, currLWS.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec, &expected.Spec.LeaderWorkerTemplate.WorkerTemplate.Spec, config); err != nil {
+				return nil, fmt.Errorf("failed to attach LoRA adapters to prefill worker template: %w", err)
+			}
```
pierDipi: can we do this within attachModelArtifacts? by doing that we won't miss places we need this like scheduler (now it's needed there too and we're missing it)

andresllh: Done

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -168,6 +168,8 @@ type LLMModelSpec struct {
 // LoRASpec defines the configuration for LoRA adapters.
 type LoRASpec struct {
 	// Adapters is the static specification for one or more LoRA adapters.
+	// The LLMInferenceService controller reconciles these into vLLM (LoRA CLI flags on main container Args,
+	// storage-initializer for hf:// and s3://, PVC mounts for pvc://).
```
spolti: Can we clarify this comment? I am not sure what it means.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,237 @@
...
+		switch scheme {
+		case constants.HfURIPrefix, constants.S3URIPrefix:
+			if storageInitializerDisabled {
+				return nil, errors.New("LoRA adapter with hf:// or s3:// URI requires storage initializer (set storageInitializer.enabled to true)")
+			}
+		case constants.PvcURIPrefix:
+			if storageInitializerDisabled {
+				return nil, errors.New("LoRA adapter with pvc:// URI requires a mounted volume (do not set storageInitializer.enabled to false)")
+			}
+		case constants.OciURIPrefix:
```
spolti, line 98: can't it be moved to the default response as not supported?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,237 @@
...
+
+// resolvedLoRAAdapter is one adapter after URI validation (hf/s3 downloads are handled in attachModelArtifacts).
+type resolvedLoRAAdapter struct {
+	name      string
+	mountPath string
+	uri       string
+	scheme    string
+}
+
+// enumerateLoRAAdapters validates spec.model.lora.adapters and returns mount paths and schemes in order.
```
spolti: what is the order in which it is returned?

python/storage-initializer/scripts/initializer-entrypoint
```diff
@@ -15,6 +16,12 @@ dest_paths = sys.argv[2::2]
 
 logger.info("Initializing, args: (src_uri, dest_path): %s", list(zip(src_uris, dest_paths)))
 
+# Create all destination dirs before downloads (HF snapshot_download and legacy
+# os.mkdir in Storage.download expect parents to exist).
+for dest in dest_paths:
+    if dest:
+        os.makedirs(dest, exist_ok=True)
+
```
spolti, line 24: How is it different than the other dir creation?

python/storage/kserve_storage/kserve_storage.py
```diff
@@ -134,14 +133,20 @@ def download_files(
...
             ignore_patterns=ignore_patterns,
         )
-        with ThreadPoolExecutor() as executor:
-            model_dirs = list(executor.map(download_fn, source_uris, out_dirs))
-        return model_dirs
+        # Sequential: parallel snapshot_download of the same Hugging Face repo into
+        # different local_dir paths races on the process-wide HF cache (locks, temp files).
+        return [
+            download_fn(uri, out)
+            for uri, out in zip(source_uris, out_dirs, strict=True)
```
spolti, line 148: can we add a unit test in the storage module for this addition?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -311,6 +430,110 @@ func (r *LLMISVCReconciler) attachStorageInitializer(modelUri string, curr corev
 	return nil
 }
 
+// attachMultiStorageDownloads adds one storage-initializer init container with multiple
+// src_uri dest_path pairs (see storage-initializer entrypoint) and mounts the shared
+// emptyDir at the common parent of all destination paths.
+func (r *LLMISVCReconciler) attachMultiStorageDownloads(
```
spolti, line 390: can't we reuse the storageInitializer multi-model feature for this?
adding the missing bits there will also benefit isvc.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,237 @@
...
+	)
+	argv = append(argv, loraModules...)
+	main.Args = append(main.Args, argv...)
+}
+
+func sanitizeLoRAPathSegment(s string) string {
+	var b strings.Builder
+	for _, r := range s {
+		switch {
+		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9', r == '-', r == '_', r == '.':
```
spolti: isn't better to use regex?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,237 @@
...
+	if len(adapters) == 0 {
+		return nil
+	}
+
+	var loraModules []string
+	for i, a := range adapters {
+		switch a.scheme {
+		case constants.PvcURIPrefix:
+			volName := fmt.Sprintf("lora-pvc-%d", i)
+			if err := attachLoraPVCAdapter(a.uri, podSpec, containerName, a.mountPath, volName); err != nil {
```
spolti, line 149: instead mouting pvcs around, isn't it easier to rely on the storage module top copy the adapters into the same volume as the model?

e.g.:
```
/mnt/models/<model_files>
/mnt/models/loras/<lora1,lora2, etc>
```

it would save some complexity here and could be reusable by isvc as well.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,237 @@
...
+			return true
+		}
+	}
+	joined := strings.Join(c.Command, " ") + " " + strings.Join(c.Args, " ")
+	return strings.Contains(joined, "--lora-modules")
+}
+
+// appendLoRAVLLMWorkloadArgs appends vLLM LoRA flags to main.Args so the LLMInferenceServiceConfig
+// entrypoint can pass them to `vllm serve` (eval "... $@") after the trailing `--` argv separator.
+func appendLoRAVLLMWorkloadArgs(main *corev1.Container, n int, loraModules []string) {
```
spolti: we will need to have the docs updated to reflect it and how to use loras with llmisvc

python/storage-initializer/scripts/initializer-entrypoint
andresllh: the `os.makedirs` function creates all directories in the path (if needed) vs just the last dir in the path that `os.makedir` creates

pkg/controller/v1alpha2/llmisvc/workload_storage.go
andresllh: Good catch! You're right that there's duplicated logic between the webhook's multi-download implementation (`storage_initializer_injector.go:326-357`) and our controller's `attachMultiStorageDownloads` function.

Both implementations follow the same pattern:
1. Find common parent path (`utils.FindCommonParentPath`)
2. Build args as alternating URI/path pairs
3. Create init container (`utils.CreateInitContainerWithConfig`)
4. Add volume mounts (`utils.AddModelMount`)

We do reuse the core utility functions, but the orchestration logic is duplicated. We could extract this into a shared utility like `utils.SetupMultiDownloadInitContainer()` that both the webhook and controller could use.

To keep this PR focused and simpler, I've filed #5360 to track the refactoring work as a follow-up. The new utility would benefit both InferenceService and LLMInferenceService by centralizing the multi-download pattern.

Does that sound like a reasonable approach?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: Great suggestion! I like the idea of unifying everything under `/mnt/models/loras/` - it's cleaner and would simplify the main container's volume mounts. Let me explain the current blocker:

## The Problem: Symlinks vs Copies

The storage-initializer currently uses **symlinks** for PVC/local sources instead of copying files:

```python
# python/storage/kserve_storage/kserve_storage.py:1157-1159
logger.info("Linking: %s to %s", src, dest_path)
if not os.path.exists(dest_path):
    os.symlink(src, dest_path)
```

**Your proposed flow:**
1. Mount PVC to init container: `/mnt/pvc-sources/adapter1`
2. Storage-initializer: `/mnt/pvc-sources/adapter1 /mnt/models/loras/adapter1`
3. Main container: only mount `/mnt/models` (everything in one place ✅)

**Why it doesn't work:**
- Storage-initializer creates symlink: `/mnt/models/loras/adapter1 -> /mnt/pvc-sources/adapter1`
- Init container completes, PVC unmounted
- Symlink is now broken (points to non-existent path)
- Main container can't access adapter

## What Would Need to Change

To make this work, we'd need to modify storage-initializer to **copy** PVC files instead of symlinking:

```python
# Detect PVC source and copy instead of symlink
if src.startswith("/mnt/pvc"):  # or similar detection
    shutil.copytree(src, dest_path)  # actual copy
else:
    os.symlink(src, dest_path)  # existing symlink logic
```

**Impact:**
- Changes `python/storage/kserve_storage/kserve_storage.py` (shared by ISVC and LLMISvc)
- Copy overhead: time + disk space (data duplicated in PVC + emptyDir)

## Benefits of Your Approach

If we implement this:
- ✅ Cleaner architecture: everything under `/mnt/models/`
- ✅ Simpler pod: one volume mount to main container
- ✅ Uniform paths: `/mnt/models/loras/adapter1`, `/mnt/models/loras/adapter2`
- ✅ More maintainable

## Options

**Option A: Implement in this PR**
- Modify storage-initializer to copy PVC files
- Change adapter paths to `/mnt/models/loras/`
- Expands PR scope to include storage-initializer changes

**Option B: Defer to follow-up**
- Keep current direct-mount approach for this PR
- File issue to refactor PVC adapter handling with copy-based approach
- Allows this PR to stay focused on core LoRA functionality

I'm happy to do either - what's your preference? The copy-based approach is definitely cleaner if we're willing to accept the disk/time overhead of duplicating PVC data.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: I will work on the kserve website updates.

andresllh: https://github.com/kserve/website/pull/654

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
```
pierDipi: Should we embed this logic in the base llmisvc config? the benefit is that we get versioning for free through the prefix llmisvcconfig feature at the cost of some repeated logic but we don't risk "restarts on upgrade" issues.

We would need to call `enumerateLoRAAdapters` in config_merge and pass it to the Config passed to the template

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -76,26 +141,78 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
 		return nil
 	}
 
+	var loraPairs []storageDownloadPair
+	var err error
+	if containerName == "main" {
```
pierDipi: should this behaviour be explicitly enabled/disabled with a parameter instead of assuming container names?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -76,26 +141,78 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
...
+			}
+		}
+
+	default:
+		return fmt.Errorf("unsupported schema in model URI: %s", modelUri)
 	}
 
-	return fmt.Errorf("unsupported schema in model URI: %s", modelUri)
+	// Attach LoRA adapters (PVC mounts + vLLM flag injection) after model downloads
+	if containerName == "main" {
```
pierDipi: same here

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -222,7 +222,7 @@ jobs:
 
       - name: Run predictor E2E tests
         id: predictor-tests
-        timeout-minutes: 40
+        timeout-minutes: 60
```
bartoszmajsak, line 225: I guess w/o it, it's more likely to fail?

We could also rename this step. I am not sure it's `predictor-tests` :)

docs/samples/llmisvc/lora-adapters/README.md
```diff
@@ -0,0 +1,545 @@
...
+- **Multi-Tenancy**: Serve multiple adapted versions of a model from a single deployment
+- **Fast Switching**: Load and unload adapters dynamically without restarting the service
+- **Task Specialization**: Different adapters for translation, summarization, coding, etc.
+
+## Prerequisites
+
+- Kubernetes cluster with GPU nodes
+- Base model accessible via HuggingFace, S3, or PVC
+- LoRA adapter weights accessible via supported URI schemes
+- vLLM-compatible runtime (default for LLMInferenceService)
```
bartoszmajsak, line 23: Worth linking to vLLM docs that cover LoRA? Probably helpful for the users.

hack/setup/infra/manage.kserve-kustomize.sh
```diff
@@ -187,6 +187,12 @@ if [ -n "${KSERVE_OVERLAY_DIR}" ]; then
         TARGET_DEPLOYMENT_NAMES+=("kserve-controller-manager")
         TARGET_DEPLOYMENT_NAMES+=("kserve-localmodel-controller-manager")
     elif [ "${KSERVE_OVERLAY_DIR}" == "test-llmisvc" ]; then
+        # Update test-llmisvc overlay image tags if version is set
+        if [ -n "${SET_KSERVE_VERSION}" ]; then
+            log_info "Updating test-llmisvc overlay image tags to ${SET_KSERVE_VERSION}..."
+            sed -i -e "s/latest/${SET_KSERVE_VERSION}/g" config/overlays/test-llmisvc/llmisvc_image_patch.yaml
+            sed -i -e "s/latest/${SET_KSERVE_VERSION}/g" config/configmap/inferenceservice.yaml
+        fi
```
bartoszmajsak, line 195: Separate PR?

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -184,6 +184,8 @@ type LLMModelSpec struct {
 // LoRASpec defines the configuration for LoRA adapters.
 type LoRASpec struct {
 	// Adapters is the static specification for one or more LoRA adapters.
+	// The LLMInferenceService controller reconciles these into vLLM (LoRA CLI flags on main container Args,
+	// storage-initializer for hf:// and s3://, PVC mounts for pvc://).
```
bartoszmajsak: Do we  need to be that specific about implementation details in the API docs?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
+/*
+Copyright 2025 The KServe Authors.
```
bartoszmajsak: nit: if we require those headers, lets fix the year for net-new files.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+			if storageInitializerDisabled {
+				return nil, errors.New("LoRA adapter with pvc:// URI requires a mounted volume (do not set storageInitializer.enabled to false)")
+			}
+		default:
+			// Note: oci:// is intentionally not supported for LoRA adapters.
+			// OCI models run as sidecar containers ("modelcars") with shared process namespaces,
+			// but only one modelcar per pod is currently supported. Supporting multiple OCI
+			// adapters would require architectural changes (multiple sidecars, complex path
+			// management). Most LoRA adapters are distributed via HuggingFace (hf://), so the
+			// limitation has minimal real-world impact. Workaround: package in PVC and use pvc://.
```
bartoszmajsak: maybe for OCI that could be dedicated match with helpful error message instead? We could even link to docs for better UX

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+		mountPath := filepath.Join(loraAdaptersMountRoot, sanitizeLoRAPathSegment(adapterName))
+
+		switch scheme {
+		case constants.HfURIPrefix, constants.S3URIPrefix:
+			if storageInitializerDisabled {
+				return nil, errors.New("LoRA adapter with hf:// or s3:// URI requires storage initializer (set storageInitializer.enabled to true)")
+			}
+		case constants.PvcURIPrefix:
+			if storageInitializerDisabled {
+				return nil, errors.New("LoRA adapter with pvc:// URI requires a mounted volume (do not set storageInitializer.enabled to false)")
```
bartoszmajsak: Like for OCI - linking to docs could improve UX?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+
+	storageInitializerDisabled := llmSvc.Spec.StorageInitializer != nil &&
+		llmSvc.Spec.StorageInitializer.Enabled != nil &&
+		!*llmSvc.Spec.StorageInitializer.Enabled
+
+	baseModelName := ptr.Deref(llmSvc.Spec.Model.Name, llmSvc.Name)
+	adapters := llmSvc.Spec.Model.LoRA.Adapters
+	out := make([]resolvedLoRAAdapter, 0, len(adapters))
+
+	for i, adapter := range adapters {
```
bartoszmajsak: Should we handle duplicates?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+			return nil, fmt.Errorf("LoRA adapter name %q must differ from base model name %q", adapterName, baseModelName)
+		}
+
+		uri := adapter.URI.String()
+		schema, _, sepFound := strings.Cut(uri, "://")
+		if !sepFound {
+			return nil, fmt.Errorf("LoRA adapter %q: invalid URI %q", adapterName, uri)
+		}
+		scheme := schema + "://"
+		mountPath := filepath.Join(loraAdaptersMountRoot, sanitizeLoRAPathSegment(adapterName))
```
bartoszmajsak, line 87: Can we escape with adapter name `..`?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -311,6 +430,110 @@ func (r *LLMISVCReconciler) attachStorageInitializer(modelUri string, curr corev
...
+	}
+
+	initPtr := utils.GetInitContainerWithName(podSpec, constants.StorageInitializerContainerName)
+	if initPtr == nil {
+		return errors.New("storage-initializer init container not found after attachMultiStorageDownloads")
+	}
+
+	if serviceAccount == nil {
+		serviceAccount = &corev1.ServiceAccount{}
+		err := r.Get(ctx, types.NamespacedName{Name: "default", Namespace: llmSvc.Namespace}, serviceAccount)
```
bartoszmajsak: Shouldn't we use constant for SA name?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+	storageInitializerDisabled := llmSvc.Spec.StorageInitializer != nil &&
+		llmSvc.Spec.StorageInitializer.Enabled != nil &&
+		!*llmSvc.Spec.StorageInitializer.Enabled
+
+	baseModelName := ptr.Deref(llmSvc.Spec.Model.Name, llmSvc.Name)
+	adapters := llmSvc.Spec.Model.LoRA.Adapters
+	out := make([]resolvedLoRAAdapter, 0, len(adapters))
+
+	for i, adapter := range adapters {
+		if adapter.Name == nil || *adapter.Name == "" {
```
bartoszmajsak: Some of these rules are pure validation - can't we move it to the webhook?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+	if llmSvc.Spec.Model.LoRA == nil || len(llmSvc.Spec.Model.LoRA.Adapters) == 0 {
+		return nil, nil
+	}
+
+	storageInitializerDisabled := llmSvc.Spec.StorageInitializer != nil &&
+		llmSvc.Spec.StorageInitializer.Enabled != nil &&
+		!*llmSvc.Spec.StorageInitializer.Enabled
+
+	baseModelName := ptr.Deref(llmSvc.Spec.Model.Name, llmSvc.Name)
+	adapters := llmSvc.Spec.Model.LoRA.Adapters
```
bartoszmajsak: Do we want this stable? Especially that we rely on indexing for `pvc`s (`lora-pvc-%d`)

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -39,11 +39,76 @@ import (
...
+
 var tokenizerOnlyDownload = corev1.EnvVar{
 	Name:  "STORAGE_ALLOW_PATTERNS",
 	Value: `["tokenizer.json", "tokenizer_config.json", "special_tokens_map.json", "vocab.json", "merges.txt", "config.json", "generation_config.json"]`,
 }
 
+// stripPriorControllerStorageInitializer removes storage-initializer init containers that would
+// duplicate the one the controller is about to add: merged/user templates often already
+// define "storage-initializer", and older reconciles used per-LoRA "storage-initializer-lora-*"
+// plus "kserve-provision-location-lora-*" volumes.
```
bartoszmajsak: Do we have older per-LoRA  init-containers? Is that a documented approach? I wonder if we need this conversion/removal logic.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+// attachLoRAAdapters reconciles spec.model.lora.adapters into vLLM CLI flags appended to the main
+// container's Args. hf:// and s3:// adapters are downloaded in attachModelArtifacts via one
+// storage-initializer; pvc:// adapters are mounted here.
+func (r *LLMISVCReconciler) attachLoRAAdapters(
+	ctx context.Context,
+	_ *corev1.ServiceAccount,
+	llmSvc *v1alpha2.LLMInferenceService,
+	_ corev1.PodSpec,
+	podSpec *corev1.PodSpec,
+	_ *Config,
```
bartoszmajsak: Is there a contract we need to fulfill forcing us to have blank params?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+	return strings.Contains(joined, "--lora-modules")
+}
+
+// appendLoRAVLLMWorkloadArgs appends vLLM LoRA flags to main.Args so the LLMInferenceServiceConfig
+// entrypoint can pass them to `vllm serve` (eval "... $@") after the trailing `--` argv separator.
+func appendLoRAVLLMWorkloadArgs(main *corev1.Container, n int, loraModules []string) {
+	argv := make([]string, 0, 5+len(loraModules))
+	argv = append(argv,
+		"--enable-lora",
+		fmt.Sprintf("--max-lora-rank=%d", defaultMaxLoRARank),
```
bartoszmajsak: Why not adding it to the spec? WDYT @pierDipi?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	appendLoRAVLLMWorkloadArgs(main, len(loraModules), loraModules)
+
+	return nil
+}
+
+func userSuppliedLoRAConfig(c *corev1.Container) bool {
+	for _, e := range c.Env {
+		if e.Name == "VLLM_ADDITIONAL_ARGS" && strings.Contains(e.Value, "--lora-modules") {
```
bartoszmajsak, line 204: Just curious - what about `valueFrom`? Should we handle it or warn as non-supported? Probably speculating too much, but you never know what kind of interesting use cases users can come up with :)

pkg/controller/v1alpha2/llmisvc/workload_lora.go
pierDipi: I like that

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: If a user is providing them, then we can assume they put lora adapters in the spec so we just warn them.

andresllh: Solution I went with was provide a warning

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,282 @@
...
+
+// appendLoRAVLLMWorkloadArgs appends vLLM LoRA flags to main.Args so the LLMInferenceServiceConfig
+// entrypoint can pass them to `vllm serve` (eval "... $@") after the trailing `--` argv separator.
+func appendLoRAVLLMWorkloadArgs(main *corev1.Container, loraModules []string, maxRank, maxAdapters, maxCpuAdapters int) {
+	argv := make([]string, 0, 5+len(loraModules))
+	argv = append(argv,
+		"--enable-lora",
+		fmt.Sprintf("--max-lora-rank=%d", maxRank),
+		fmt.Sprintf("--max-loras=%d", maxAdapters),
+		fmt.Sprintf("--max-cpu-loras=%d", maxCpuAdapters),
```
pierDipi: should we only pass this when set explicitly? vLLM sets this to None by default

andresllh: Good catch, I will update all defaults to left empty with a comment of what vllm sets them to.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+		case constants.PvcURIPrefix:
+			if storageInitializerDisabled {
+				return nil, fmt.Errorf("LoRA adapter %q: pvc:// requires a mounted volume — do not set storageInitializer.enabled to false (see %s)", adapterName, loraAdapterDocsURL)
+			}
+		case constants.OciURIPrefix:
+			// oci:// is intentionally not supported for LoRA adapters. OCI models run as sidecar
+			// containers ("modelcars") with shared process namespaces, but only one modelcar per pod
+			// is currently supported. Workaround: package the adapter in a PVC and use pvc://.
+			return nil, fmt.Errorf("LoRA adapter %q: oci:// is not supported for LoRA adapters; use hf://, s3://, or pvc:// instead (see %s)", adapterName, loraAdapterDocsURL)
```
pierDipi, line 102: maybe this could be done in validation webhook as well?

pierDipi (approved):
/lgtm

bartoszmajsak (approved):
/lgtm
/approve

docs/samples/llmisvc/lora-adapters/README.md
```diff
@@ -0,0 +1,563 @@
+# LoRA Adapter Examples
```
spolti, line 1: are you planning to send it to the official docs as well?

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -403,6 +404,75 @@ func (l *LLMInferenceServiceValidator) validateSchedulerConfig(svc *LLMInference
...
+			continue
+		}
+
+		adapterName := *adapter.Name
+
+		if adapterName == "." || adapterName == ".." {
+			allErrs = append(allErrs, field.Invalid(
+				namePath,
+				adapterName,
+				"adapter name must not be \".\" or \"..\" (path traversal risk)",
```
spolti:
```suggestion
				"adapter name must not include \".\" or \"..\" (path traversal risk)",
```

docs/samples/llmisvc/lora-adapters/README.md
andresllh: https://github.com/kserve/website/pull/654

pkg/controller/v1alpha2/llmisvc/workload.go
```diff
@@ -43,6 +43,11 @@ const (
 	routingSidecarContainerName = "llm-d-routing-sidecar"
 
 	defaultServiceAccountName = "default"
+
```
spolti: shouldn't we move constants to a single file, constants.go?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
spolti: +1

pkg/controller/v1alpha2/llmisvc/workload_lora.go
spolti: cant this and `hasValueFromLoRAConfig` be unified?

pkg/controller/v1alpha2/llmisvc/workload_lora.go
```diff
@@ -0,0 +1,283 @@
...
+
+// appendLoRAVLLMWorkloadArgs appends vLLM LoRA flags to main.Args so the LLMInferenceServiceConfig
+// entrypoint can pass them to `vllm serve` (eval "... $@") after the trailing `--` argv separator.
+// maxRank, maxAdapters, and maxCpuAdapters are only injected when non-nil (explicitly set in the spec);
+// vLLM applies its own defaults otherwise.
+// See: https://github.com/vllm-project/vllm/blob/main/vllm/config/lora.py
+func appendLoRAVLLMWorkloadArgs(main *corev1.Container, loraModules []string, maxRank, maxAdapters, maxCpuAdapters *int32) {
+	argv := make([]string, 0, 5+len(loraModules))
+	argv = append(argv, "--enable-lora")
+	if maxRank != nil {
```
spolti: May we rename it to avoind confusion with https://github.com/kserve/kserve/pull/5317/changes#diff-9fd32e5d187473c2530707d7cec2ccdf3fe694d36ebd214ee006b8aa9e2994e2R417 ?

pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
```diff
@@ -206,7 +206,7 @@ func (r *LLMISVCReconciler) expectedMainMultiNodeLWS(ctx context.Context, llmSvc
 			currLeaderSpec = currLWS.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec
 		}
 
-		if err := r.attachModelArtifacts(ctx, serviceAccount, llmSvc, currLeaderSpec, &expected.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec, config, "main", constants.DefaultModelLocalMountPath); err != nil {
+		if err := r.attachModelArtifacts(ctx, serviceAccount, llmSvc, currLeaderSpec, &expected.Spec.LeaderWorkerTemplate.LeaderTemplate.Spec, config, "main", constants.DefaultModelLocalMountPath, withLoRAAdapters); err != nil {
```
spolti: Shouldn't this be dynamic based on the configurations provided by the user?
I mean, if user provides LoRa adapters, then this variable is dynamically set.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
spolti: okay, thanks.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Done — reworded the comment. The `Adapters` field now reads: "Each adapter is specified by name and URI (supports hf://, s3://, and pvc:// schemes). The controller automatically downloads adapters and configures the runtime to use them."

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: Kept as an explicit case rather than folding into `default` — @bartoszmajsak suggested providing a dedicated match with a helpful error message and docs link for better UX. The current message explains why oci:// is unsupported for LoRA and suggests the pvc:// workaround.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: Fixed — the comment now reads: "Returns adapters sorted by name for stable pod spec generation."

python/storage/kserve_storage/kserve_storage.py
andresllh: Added — tests for `Storage.download_files` are in `python/storage/test/test_multi_download.py`.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: Updated to use regex — sanitization now uses `regexp.MustCompile` for both path segments (`loraPathInvalidCharsRe`) and volume names (`loraVolumeNameInvalidCharsRe`).

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
andresllh: Fixed — changed to "adapter name must not include \".\" or \"..\"".

pkg/controller/v1alpha2/llmisvc/workload.go
andresllh: Done — created `pkg/controller/v1alpha2/llmisvc/constants.go` and moved `routingSidecarContainerName`, `defaultServiceAccountName`, and `withoutLoRAAdapters` there. Removed the `const` block from `workload.go`.

pkg/controller/v1alpha2/llmisvc/workload_lora.go
andresllh: Renamed the parameters in `addLoRAVLLMArgs` (previously `appendLoRAVLLMWorkloadArgs`) from `maxRank`, `maxAdapters`, `maxCpuAdapters` to `loraMaxRank`, `loraMaxAdapters`, `loraMaxCpuAdapters` to distinguish them from the spec field names. Let me know if you had a different rename in mind!

pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
andresllh: Made dynamic — replaced the constant `withLoRAAdapters` with `len(config.ResolvedLoRAAdapters) > 0` at all call sites in `workload_multi_node.go` and `workload_single_node.go`. LoRA injection now only runs when the user has actually configured adapters.

PR #5318, opened 2026-03-30
feat: add LocalModelCache support for LLMInferenceService

### LINKED ISSUE #5116: Allow LLMInferenceServices to use models cached using LocalModelCache
/kind feature

LLMInferenceService should support local model caching, the same way InferenceService does today via LocalModelCache and LocalModelNamespaceCache.

  Currently, every LLMInferenceService pod must download the full model from remote storage (e.g., Hugging Face, S3) at startup — even when the same model has already been cached locally
  on the node. For large language models that can be tens or hundreds of gigabytes, this results in significantly longer startup times, wasted network bandwidth, and increased costs.

  With this feature, if a LocalModelCache or LocalModelNamespaceCache exists for a model, an LLMInferenceService using that same model would automatically use the local copy — no
  configuration changes required. Users create their LLMInferenceService exactly as they do today:
```
  kind: LLMInferenceService
  metadata:
    name: my-llm
  spec:
    model:
      uri: hf://meta-llama/Llama-4-Scout-17B
    workloadSpec:
      replicas: 1
```

If a matching cache exists, the system handles everything transparently: URI matching, PVC routing, and status tracking. Users can also opt out per-service with the existing
  serving.kserve.io/disable-localmodel annotation, and select specific node groups with serving.kserve.io/nodegroup.

  This should work for both single-node (Deployment) and multi-node (LeaderWorkerSet) LLM workloads, including prefill/decode disaggregation setups.

VedantMahabaleshwarkar:

Adds LocalModelCache support for LLMInferenceService (LLMISVC), enabling model caching for LLM workloads with the same deduplication and multi-namespace support available for InferenceService.

Fixes : #5116

Key changes:
- **Webhook defaulter**: Mutating webhook that detects when an LLMInferenceService's model URI matches an existing LocalModelCache or LocalModelNamespaceCache and sets the appropriate labels/annotations
- **Cache status tracking**: Adds `LLMInferenceServices` field to `LocalModelCacheStatus` so reconcilers can track which LLMInferenceServices use a cached model
- **Reconciler integration**: LocalModelCache and LocalModelNamespaceCache reconcilers now watch LLMInferenceServices, create PVs/PVCs for their namespaces, and track them in cache status
- **Label propagation**: Workload controllers propagate localmodel labels/annotations to Deployment and LeaderWorkerSet pod templates
- **ValidateDelete protection**: Delete validation webhooks now check both InferenceServices and LLMInferenceServices before allowing cache deletion
- **URI rewriting**: `attachModelArtifacts` rewrites the model URI from the original (e.g. `hf://model`) to `pvc://<cached-pvc>/models/<storageKey>/` when a local model cache is active
- **RBAC**: Added permissions for LLMInferenceService controller to read localmodel caches and for the localmodel controller to watch LLMInferenceServices

- [x] Unit tests for LLMInferenceService webhook defaulter (cluster-scoped, namespace-scoped, no-match, node group matching, stale metadata removal)
- [x] Unit tests for ValidateDelete with active LLMInferenceService (both LocalModelCache and LocalModelNamespaceCache)
- [x] Unit tests for localmodel label propagation to Deployment pod templates
- [x] Integration test (Ginkgo/envtest) for URI rewriting from `hf://` to `pvc://` when local model cache labels are present
- [x] E2E test verifying webhook sets labels and cache status tracks LLMInferenceService
- [x] `make precommit` passes
- [x] `make test` passes (full suite)

LLMInferenceService uses a controller-driven approach (building pod specs directly via `attachModelArtifacts`) rather than the pod-level webhook used by InferenceService. This means local model cache URI rewriting is handled in `workload_storage.go` rather than in `storage_initializer_injector.go`.

**Checklist**:

**Release note**:
```release-note
Add LocalModelCache support for LLMInferenceService, enabling local model caching for LLM workloads with automatic URI rewriting and delete protection.
```


go.sum
pierDipi, line 1: we need to clean this with `make tidy`

config/webhook/llmisvc/manifests.yaml
```diff
@@ -1,4 +1,31 @@
...
+    failurePolicy: Fail
+    matchPolicy: Exact
+    name: llminferenceservice.kserve-webhook-server.defaulter
+    sideEffects: None
+    admissionReviewVersions: [ "v1", "v1beta1" ]
+    rules:
+      - apiGroups:
+          - serving.kserve.io
+        apiVersions:
+          - v1alpha2
```
pierDipi, line 22: don't we need v1alpha1 too?

pkg/webhook/admission/llminferenceservice/defaulter.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/utils"
+)
+
+var defaulterLogger = logf.Log.WithName("llminferenceservice-defaulter")
+
+// +kubebuilder:object:generate=false
+// +k8s:openapi-gen=false
+
+// +kubebuilder:webhook:path=/mutate-serving-kserve-io-v1alpha2-llminferenceservice,mutating=true,failurePolicy=fail,sideEffects=None,groups=serving.kserve.io,resources=llminferenceservices,verbs=create;update,versions=v1alpha2,name=llminferenceservice.kserve-webhook-server.defaulter,admissionReviewVersions=v1
```
bartoszmajsak: Do we need a defaulting webhook? I wonder if it can't be part of reconciliation and ` attachModelArtifacts()` path?

Also, do we have to deal with `v1alpha1.LocalModelCache` and `v1alpha1.LocalModelNamespaceCache` events, such as create/update/delete and reconcile LLMInferenceService accordingly?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -76,6 +77,25 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
...
+			return fmt.Errorf("annotation %s not found", constants.LocalModelPVCNameAnnotationKey)
+		}
+		subPath, _ := strings.CutPrefix(modelUri, sourceUri)
+		if !strings.HasPrefix(subPath, "/") {
+			subPath = "/" + subPath
+		}
+		storageKey := v1alpha1.GetStorageKey(sourceUri)
+		modelUri = "pvc://" + pvcName + "/models/" + storageKey + subPath
+		schema = "pvc"
+	}
```
pierDipi, line 97: how are we going handle LoRAs with just an annotation https://github.com/kserve/kserve/pull/5317? is the expectation that every model is in the cache? don't we need more fine grained control in this multi-model cases?

cc @andresllh

pkg/webhook/admission/llminferenceservice/defaulter.go
VedantMahabaleshwarkar: I used a webhook to try to keep the logic as similar as possible to what modelcache does for ISVCs. 

> Also, do we have to deal with v1alpha1.LocalModelCache and v1alpha1.LocalModelNamespaceCache events, such as create/update/delete and reconcile LLMInferenceService accordingly? 

no

bartoszmajsak: > > Also, do we have to deal with v1alpha1.LocalModelCache and v1alpha1.LocalModelNamespaceCache events, such as create/update/delete and reconcile LLMInferenceService accordingly?

> no

I wonder what will happen when we delete the Cache CR and try to scale the deployment.

bartoszmajsak: > I used a webhook to try to keep the logic as similar as possible to what modelcache does for ISVCs.

Is there a reason to do that when these are two separate control planes? 

I am also not sure if we want to fail-close where we have a few API calls within the current flow that can fail. 

If we stick with the webhook we should also consider using cached client instead of creating an instance for each and every defaulting call.

VedantMahabaleshwarkar: >I wonder what will happen when we delete the Cache CR and try to scale the deployment. 

we reject deletion of the cache CR via webhook if an isvc/llmisvc is using the cache

pkg/controller/v1alpha2/llmisvc/workload_storage.go
VedantMahabaleshwarkar: LoRA adapter local cache support will be a follow-up story in [RHOAIENG-58262](https://redhat.atlassian.net/browse/RHOAIENG-58262).

pkg/webhook/admission/llminferenceservice/defaulter.go
VedantMahabaleshwarkar: I considered moving this to the reconciler / `attachModelArtifacts()` path, but I think using the webhook is better because the labels must be persisted regardless — the LocalModel reconciler (in a separate binary) finds LLMISVCs via a field index on `LocalModelLabel` to provision PVs/PVCs and track cache status. If we moved cache matching to the reconciler, it would still need to `client.Update()` the LLMISVC to set those labels, introducing re-entrancy (update triggers another reconcile) and a race window where the LocalModel reconciler can't find the LLMISVC yet.

On the fail-close and cached client concerns: I've updated the defaulter to use the manager's cached client and an injected clientset instead of creating new ones on every admission call.

go.sum
VedantMahabaleshwarkar: Done.

config/webhook/llmisvc/manifests.yaml
VedantMahabaleshwarkar: Done — added v1alpha1 to the apiVersions and updated the kubebuilder marker.

pkg/webhook/admission/llminferenceservice/defaulter.go
```diff
@@ -0,0 +1,181 @@
...
+		defaulterLogger.Error(err, "unable to get configmap", "name", constants.InferenceServiceConfigMapName, "namespace", constants.KServeNamespace)
+		return err
+	}
+	localModelConfig, err := v1beta1.NewLocalModelConfig(configMap)
+	if err != nil {
+		return err
+	}
+
+	_, localModelDisabled := llmSvc.Annotations[constants.DisableLocalModelKey]
+	if !localModelDisabled && localModelConfig.Enabled {
```
andresllh, line 155: Stale labels are not cleaned up when the local model feature is disabled.

If `localModelDisabled` is `true` or `localModelConfig.Enabled` is `false`, `SetLocalModelLabel` is never called — so any cache labels/annotations previously stamped on the `LLMInferenceService` (e.g. `LocalModelLabel`, `LocalModelSourceUriAnnotationKey`, `LocalModelPVCNameAnnotationKey`) will persist. The controller will then continue rewriting the model URI to PVC even though the user has explicitly opted out.

Suggestion: call `DeleteLocalModelMetadata(llmSvc)` in the else branch:

```go
if !localModelDisabled && localModelConfig.Enabled {
    // ... existing path
    SetLocalModelLabel(llmSvc, models, nsModels)
} else {
    DeleteLocalModelMetadata(llmSvc)
}
```

pkg/webhook/admission/localmodelcache/localmodelcache_validator.go
```diff
@@ -112,6 +113,22 @@ func (v *LocalModelCacheValidator) ValidateDelete(ctx context.Context, obj runti
 			return admission.Warnings{}, fmt.Errorf("LocalModelCache %s is being used by InferenceService %s", localModelCache.Name, isvcMeta.Name)
 		}
 	}
+
+	// Check if current LocalModelCache is being used by LLMInferenceServices
+	for _, llmIsvcMeta := range localModelCache.Status.LLMInferenceServices {
+		llmIsvc := v1alpha2.LLMInferenceService{}
+		if err := v.Get(ctx, client.ObjectKey(llmIsvcMeta), &llmIsvc); err != nil {
+			localModelCacheValidatorLogger.Error(err, "Error getting LLMInferenceService", "name", llmIsvcMeta.Name)
+			return nil, err
```
andresllh, line 129: If an `LLMInferenceService` tracked in status has already been deleted (e.g. manual deletion before the reconciler has a chance to update the status), `v.Get` returns `IsNotFound` and this returns an error — blocking cache deletion indefinitely until the reconciler runs and removes the stale entry.

The same issue exists in the ISVC path, but worth fixing here too:

```go
if err := v.Get(ctx, client.ObjectKey(llmIsvcMeta), &llmIsvc); err != nil {
    if apierrors.IsNotFound(err) {
        continue
    }
    localModelCacheValidatorLogger.Error(err, "Error getting LLMInferenceService", "name", llmIsvcMeta.Name)
    return nil, err
}
```

pkg/webhook/admission/localmodelnamespacecache/local_model_namespace_cache_validation.go
```diff
@@ -108,6 +109,24 @@ func (v *LocalModelNamespaceCacheValidator) ValidateDelete(ctx context.Context,
 				localModelNamespaceCache.Namespace, localModelNamespaceCache.Name, isvcMeta.Namespace, isvcMeta.Name)
 		}
 	}
+
+	// Check if current LocalModelNamespaceCache is being used by LLMInferenceServices in the same namespace
+	for _, llmIsvcMeta := range localModelNamespaceCache.Status.LLMInferenceServices {
+		llmIsvc := v1alpha2.LLMInferenceService{}
+		if err := v.Get(ctx, client.ObjectKey(llmIsvcMeta), &llmIsvc); err != nil {
+			localModelNamespaceCacheValidatorLogger.Error(err, "Error getting LLMInferenceService", "name", llmIsvcMeta.Name, "namespace", llmIsvcMeta.Namespace)
+			return nil, err
```
andresllh, line 125: Same `IsNotFound` issue as in `localmodelcache_validator.go`: if an `LLMInferenceService` tracked in status has already been deleted, this blocks `LocalModelNamespaceCache` deletion indefinitely.

```go
if err := v.Get(ctx, client.ObjectKey(llmIsvcMeta), &llmIsvc); err != nil {
    if apierrors.IsNotFound(err) {
        continue
    }
    localModelNamespaceCacheValidatorLogger.Error(err, "Error getting LLMInferenceService", "name", llmIsvcMeta.Name, "namespace", llmIsvcMeta.Namespace)
    return nil, err
}
```

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
```diff
@@ -295,8 +344,25 @@ func (c *LocalModelNamespaceCacheReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&v1alpha1.LocalModelNamespaceCache{}).
 		Owns(&corev1.PersistentVolumeClaim{})
 
+	llmIsvcPredicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldNsLabel := e.ObjectOld.GetLabels()[constants.LocalModelNamespaceLabel]
+			newNsLabel := e.ObjectNew.GetLabels()[constants.LocalModelNamespaceLabel]
+			return oldNsLabel != newNsLabel
```
andresllh: The `UpdateFunc` predicate for `llmIsvcPredicates` only checks `LocalModelNamespaceLabel`, not `LocalModelLabel`. If the model name label changes while the namespace label stays the same, the reconciler will not fire.

This mirrors an existing gap in the ISVC predicate for namespace caches — not a regression from this PR, but worth fixing in both while touching this code:

```go
UpdateFunc: func(e event.UpdateEvent) bool {
    oldModel := e.ObjectOld.GetLabels()[constants.LocalModelLabel]
    newModel := e.ObjectNew.GetLabels()[constants.LocalModelLabel]
    oldNs := e.ObjectOld.GetLabels()[constants.LocalModelNamespaceLabel]
    newNs := e.ObjectNew.GetLabels()[constants.LocalModelNamespaceLabel]
    return oldModel != newModel || oldNs != newNs
},
```

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -76,6 +77,25 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
 		return nil
 	}
 
+	// Rewrite model URI to use cached PVC when local model cache is active
+	if _, ok := llmSvc.Labels[constants.LocalModelLabel]; ok {
+		sourceUri, ok := llmSvc.Annotations[constants.LocalModelSourceUriAnnotationKey]
+		if !ok {
+			return fmt.Errorf("annotation %s not found", constants.LocalModelSourceUriAnnotationKey)
```
andresllh: These error messages don't include the `LLMInferenceService` name/namespace, which makes them hard to correlate in controller logs. Suggestion:

```go
return fmt.Errorf("LLMInferenceService %s/%s: annotation %s not found", llmSvc.Namespace, llmSvc.Name, constants.LocalModelSourceUriAnnotationKey)
```

Same applies to the `LocalModelPVCNameAnnotationKey` error a few lines below.

pkg/webhook/admission/localmodelnamespacecache/local_model_namespace_cache_validation.go
VedantMahabaleshwarkar: fixed

pkg/webhook/admission/localmodelcache/localmodelcache_validator.go
VedantMahabaleshwarkar: fixed in llmisvc, will create a followup for isvc. I don't want to increase the scope of this PR.

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
VedantMahabaleshwarkar: fixed in llmisvc, will create a followup for isvc. I don't want to increase the scope of this PR.

vivekk16 (commented):
lgtm

danehans:
`Precommit Check / precommit-check (pull_request)`- Run `make precommit` and commit the generated diffs.

`ModelCache E2E Tests / test-modelcache (helm) (pull_request)`- Helm failed before installing Kksrve:
```sh
Loading image huggingfaceserver-8d8662...
unexpected EOF
ssh: Process exited with status 1
```
^ looks like a flaky image tar load.

PR #5327, opened 2026-04-01
fix(helm): template llmisvc CRD namespace from release namespace

### LINKED ISSUE #5325: [Bug] Hardcoded namespace in `kserve-llmisvc-crd` chart prevents flexible deployment
/kind bug

I was reviewing the Helm charts for the LLM Inference Service (`llmisvc`) and noticed that the namespace is hardcoded as `kserve` in the `kserve-llmisvc-crd` chart. 

Specifically, in `charts/kserve-llmisvc-crd/templates/serving.kserve.io_llminferenceserviceconfigs.yaml`, the `cert-manager` annotation and the `webhook.clientConfig.service.namespace` field are fixed to `kserve`. This prevents the chart from being deployed correctly in environments where a different namespace is used.

The namespace should be templated (e.g., using `{{ .Release.Namespace }}` or a configurable value) to allow flexible deployments, consistent with how it is handled in the `kserve-llmisvc-resources` chart.

**What's the CRD yaml:**
```yaml
apiVersion: apiextensions.k8s.io/v1
metadata:
  annotations:
    cert-manager.io/inject-ca-from: kserve/llmisvc-serving-cert # Hardcoded namespace
    controller-gen.kubebuilder.io/version: v0.19.0
  name: llminferenceserviceconfigs.serving.kserve.io
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          name: llmisvc-webhook-server-service
          namespace: kserve # Hardcoded namespace
          path: /convert
          port: 443
```

I would like to resolve this by applying a templating logic similar to the one used in the `kserve-llmisvc-resources` chart. 

I am happy to contribute a fix for this. Could you please let me know if I can go ahead and submit a Pull Request?

Yuuuuuu0319:
### What this PR does

This PR fixes hardcoded namespace usage in the `kserve-llmisvc-crd` Helm chart webhook configuration by templating it with `.Release.Namespace`.

### Changes

* Replaced hardcoded `cert-manager.io/inject-ca-from: kserve/llmisvc-serving-cert`
  with `{{ .Release.Namespace }}/llmisvc-serving-cert`
* Replaced hardcoded `webhook.clientConfig.service.namespace: kserve`
  with `{{ .Release.Namespace }}`
* Applied the change only in:

  * `charts/kserve-llmisvc-crd/templates/serving.kserve.io_llminferenceservices.yaml`
  * `charts/kserve-llmisvc-crd/templates/serving.kserve.io_llminferenceserviceconfigs.yaml`

### Why

The chart currently assumes the release namespace is always `kserve`, which breaks deployments in custom namespaces.

This fix keeps the Kustomize source under `config/` unchanged and applies the namespace templating only on the Helm chart side, which matches the intended installation flow for `kserve-llmisvc-crd`.

### Verification

Ran:

```bash
helm lint ./charts/kserve-llmisvc-crd
helm template test-llmisvc-crd ./charts/kserve-llmisvc-crd --namespace team-a
```

Confirmed that:

* `cert-manager.io/inject-ca-from` renders as `team-a/llmisvc-serving-cert`
* `webhook.clientConfig.service.namespace` renders as `team-a`

Fixes #5325

### DIFF
--- a/config/crd/full/llmisvc/llmisvc_conversion_webhook_patch.yaml
+++ b/config/crd/full/llmisvc/llmisvc_conversion_webhook_patch.yaml
@@ -5,7 +5,7 @@ kind: CustomResourceDefinition
 metadata:
   name: llminferenceservices.serving.kserve.io
   annotations:
-    cert-manager.io/inject-ca-from: kserve/llmisvc-serving-cert
+    cert-manager.io/inject-ca-from: "{{ .Release.Namespace }}/llmisvc-serving-cert"
 spec:
   conversion:
     strategy: Webhook
@@ -13,7 +13,7 @@ spec:
       clientConfig:
         service:
           name: llmisvc-webhook-server-service
-          namespace: kserve
+          namespace: "{{ .Release.Namespace }}"
           path: /convert
           port: 443
       conversionReviewVersions:
--- a/config/crd/full/llmisvc/llmisvc_conversion_webhook_patch.yaml
+++ b/config/crd/full/llmisvc/llmisvc_conversion_webhook_patch.yaml
@@ -5,7 +5,7 @@ kind: CustomResourceDefinition
 metadata:
   name: llminferenceservices.serving.kserve.io
   annotations:
-    cert-manager.io/inject-ca-from: "{{ .Release.Namespace }}/llmisvc-serving-cert"
+    cert-manager.io/inject-ca-from: kserve/llmisvc-serving-cert
 spec:
   conversion:
     strategy: Webhook
@@ -13,7 +13,7 @@ spec:
       clientConfig:
         service:
           name: llmisvc-webhook-server-service
-          namespace: "{{ .Release.Namespace }}"
+          namespace: kserve
           path: /convert
           port: 443
       conversionReviewVersions:
--- a/charts/kserve-llmisvc-crd/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
+++ b/charts/kserve-llmisvc-crd/templates/serving.kserve.io_llminferenceserviceconfigs.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/llmisvc-serving-cert
+    cert-manager.io/inject-ca-from: '{{ .Release.Namespace }}/llmisvc-serving-cert'
     controller-gen.kubebuilder.io/version: v0.19.0
   name: llminferenceserviceconfigs.serving.kserve.io
 spec:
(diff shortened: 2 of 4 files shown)


vivekk16 (commented):
lgtm

sivanantha321 (approved):

Yuuuuuu0319:
@sivanantha321 Hi, the precommit check is failing with:
"Found GitHub Actions not pinned to a full SHA. Run 'make pin-actions' to fix."
However, it seems that `make pin-actions` is not available in this repository locally.
Could you please advise how this should be handled, or if this check can be rerun/adjusted?

juhyeon-cha:
@sivanantha321 @Yuuuuuu0319 I noticed that only the `charts/` were updated, while the source in `config/crd/full/llmisvc` still has the hardcoded namespace.

Won't this cause a regression in the next release when the charts are synced with the `config/` directory?

Yuuuuuu0319:
@juhyeon-cha Thanks for catching this.That updating only the chart templates could lead to regression when they are regenerated from `config/crd/full/llmisvc`.

I've updated the source webhook patch files under `config/crd/full/llmisvc` and regenerated all downstream artifacts (Helm chart templates, quick-install manifests, and test CRDs) to ensure consistency across the generation pipeline.

This should prevent any regression when the charts are synced with the config directory in future releases.

config/crd/full/llmisvc/llmisvc_conversion_webhook_patch.yaml
```diff
@@ -5,15 +5,15 @@ kind: CustomResourceDefinition
 metadata:
   name: llminferenceservices.serving.kserve.io
   annotations:
-    cert-manager.io/inject-ca-from: kserve/llmisvc-serving-cert
+    cert-manager.io/inject-ca-from: "{{ .Release.Namespace }}/llmisvc-serving-cert"
```
pierDipi: this is used for the Kustomize installation, to enable this feature/fix, we need to have transformations on the chart side only (similar to how it's done with files and templates)

pierDipi:
> @sivanantha321 @Yuuuuuu0319 I noticed that only the `charts/` were updated, while the source in `config/crd/full/llmisvc` still has the hardcoded namespace.
> 
> Won't this cause a regression in the next release when the charts are synced with the `config/` directory?

`config/` it's used for kustomize-based installations, we can't have helm templates in that directory

Yuuuuuu0319:
@pierDipi Thanks for the clarification.
I had mistakenly propagated the Helm templating into the Kustomize source and generated artifacts. I've now reverted those changes from `config/crd/full/llmisvc`, quick-install manifests, and test CRDs, and kept the namespace templating only in the Helm chart templates.
After the latest push, the PR should only contain the chart-side fix for `kserve-llmisvc-crd`.

Yuuuuuu0319:
The PR has been updated based on the latest review feedback.

The namespace templating is now kept only on the Helm chart side, while the Kustomize source and generated manifests remain unchanged.

Yuuuuuu0319:
This PR is still relevant. I’m still working on it and will follow up as needed.

Yuuuuuu0319:
This PR is still relevant. I’m still working on it and will follow up as needed.

pierDipi:
@Yuuuuuu0319 note that charts resources are auto-generated and precommit is failing

PR #5328, opened 2026-04-01
fix: merge automation

pierDipi:
* Add issues and pull-requests write permissions at the workflow level
so the prow action can manage labels, assignees, and milestones.

* Reduce the schedule interval from hourly to every 10 minutes for
faster PR automerge on lgtm label.

* Add issue_comment trigger so users can re-run the enforce-all-checks
job by commenting /rerun on a PR. Also increase polling interval and
initial delay to reduce API usage.

### DIFF
--- a/.github/workflows/prow-github.yml
+++ b/.github/workflows/prow-github.yml
@@ -5,6 +5,11 @@ on:
   issue_comment:
     types: [created]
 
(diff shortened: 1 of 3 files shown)


Jooho (approved):
/lgtm
/approved

PR #5329, opened 2026-04-01
fix(ci): pins github actions in license check

bartoszmajsak:

License check workflows didn't have their actions pinned to explicit shas - causing `make precommit` failures.

This commit pins to exact versions by running `make pin-actions` target.

**Release note**:
```release-note
NONE
```


Jooho (approved):
/lgtm
/approved

sivanantha321 (approved):
/lgtm
/approve

PR #5332, opened 2026-04-02
fix: validate HPA metrics when autoscalerClass is unset

### LINKED ISSUE #4548: Controller panics if autoscaler class is not set to keda with external autoscaling config
/kind bug

I had these resources [1] and without the `serving.kserve.io/autoscalerClass: keda` the controller panics on this line https://github.com/kserve/kserve/blob/9e55d59c9823b2c4337d613da8c0a21681f5616b/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go#L64 because `Resource` is `nil`.

I would expect it to fail validation and give me an actionable error message.

[1]
```yaml
kind: ServingRuntime
metadata:
  name: vllm-runtime
  namespace: keda-demo
spec:
  annotations:
    prometheus.io/path: /metrics
    prometheus.io/port: "8080"
  containers:
    - args:
        - --port=8080
        - --model=/mnt/models
        - --served-model-name={{.Name}}
      command:
        - python
        - '-m'
        - vllm.entrypoints.openai.api_server
      env:
        - name: HF_HOME
          value: /tmp/hf_home
      image: quay.io/modh/vllm@sha256:8a3dd8ad6e15fe7b8e5e471037519719d4d8ad3db9d69389f2beded36a6f5b21
      name: kserve-container
      ports:
        - containerPort: 8080
          protocol: TCP
  multiModel: false
  supportedModelFormats:
  - autoSelect: true
    name: vLLM

---
kind: InferenceService
metadata:
  name: granite
  namespace: keda-demo
  annotations:
spec:
  predictor:
    minReplicas: 1
    maxReplicas: 2
    autoScaling:
      metrics:
        - type: External
          external:
            authenticationRef:
              name: inference-prometheus-auth
            metric:
              authModes: bearer
              backend: "prometheus"
              serverAddress: "https://thanos-querier.xyz.svc:9092"
              query: "sum(vllm:num_requests_running{namespace='keda-demo', job='granite-metrics'})"
              namespace: 'keda-demo'
            target:
              type: Value
              value: "2"
    model:
      modelFormat:
        name: vLLM
      args:
        - --model_name=granite-3-2-8b-instruct
        - --model_dir=/mnt/models
      storageUri: hf://ibm-granite/granite-3.2-8b-instruct
      env:
        - name: VLLM_CACHE_ROOT
          value: '/tmp'
      resources:
        limits:
          cpu: '6'
          memory: 24Gi
          nvidia.com/gpu: '1'
        requests:
          cpu: '1'
          memory: 8Gi
          nvidia.com/gpu: '1'
      runtime: vllm-runtime
    tolerations:
      - effect: NoSchedule
        key: nvidia.com/gpu
        operator: Exists

```

I would expect it to fail validation and give me an actionable error message.

See above.

- We need to have a test to validate that the controller doesn't panic in this case and gives the user an actionable and clear error message.

**Environment:**

- Istio Version: Any
- Knative Version: 1.16
- KServe Version: 0.15
- Cloud Environment: Any
- Minikube/Kind version: Any
- Kubernetes version: (use `kubectl version`): Any
- OS (e.g. from `/etc/os-release`): Any

JohnsonWang1015:
## What this PR does

  Fixes the controller panic (nil pointer dereference) reported in #4548 when an `InferenceService` configures `External` autoscaling metrics without
  setting the `serving.kserve.io/autoscalerClass: keda` annotation.

  ## Root Cause

  The validation webhook in `validateAutoScalingCompExtension` had two gaps:

  1. **Missing default case for `autoscalerClass`**: When `deploymentMode` is `Standard` and `autoscalerClass` is not set, the reconciler defaults to HPA
  (`DefaultAutoscalerClass = AutoscalerClassHPA`), but the validation switch did not match the empty string case — it silently returned `nil`, skipping all
  HPA metric validation. This allowed `External` metric types to pass through unchecked, and the HPA reconciler then panicked on `metric.Resource.Name`
  because `Resource` is `nil` for External metrics.

  2. **Legacy `RawDeployment` mode not handled**: The legacy `RawDeployment` deployment mode (deprecated alias for `Standard`) was not included in the
  `Standard` case of the switch, causing it to fall into the `default` branch and run KPA validation instead of HPA validation.

  ## Changes

  ### `pkg/apis/serving/v1beta1/inference_service_validation.go`
  - Added `""` (empty string) to the `autoscalerClass` switch so that when no autoscaler class annotation is set, HPA validation runs by default — matching
  the reconciler's default behavior.
  - Added `LegacyRawDeployment` (`"RawDeployment"`) to the `deploymentMode` switch alongside `Standard`.
  - Improved the error message to be more actionable: it now suggests setting `serving.kserve.io/autoscalerClass` to `keda` when `External` or `PodMetric`
  metric types are used.

  ### `pkg/apis/serving/v1beta1/inference_service_validation_test.go`
  - Updated the existing `"Invalid HPA CPU metrics with wrong metric source"` test to match the new error message.
  - Added `"Invalid External metrics without autoscalerClass annotation in Standard mode"` test — verifies that External metrics are rejected with a clear
  error when `autoscalerClass` is not set in Standard mode.
  - Added `"Invalid External metrics without autoscalerClass annotation in RawDeployment mode"` test — verifies the same validation applies to the legacy
  `RawDeployment` mode.

  ## How to reproduce the original panic

  Apply the YAML from #4548 (an `InferenceService` with `serving.kserve.io/deploymentMode: RawDeployment`, External autoscaling metrics, and **no**
  `serving.kserve.io/autoscalerClass` annotation). Before this fix, the controller crashes with:

  panic: runtime error: invalid memory address or nil pointer dereference
    github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice/reconcilers/hpa.getHPAMetrics

  After this fix, the webhook rejects the request with:

  invalid HPA metric source type with value [External], valid metric source types are Resource.
  For External or PodMetric types, set the annotation serving.kserve.io/autoscalerClass to keda

  ## Test plan
  - [x] `go test ./pkg/apis/serving/v1beta1/ -run TestAutoscalerClassHPA` — all 12 subtests pass
  - [x] `go test ./pkg/apis/serving/v1beta1/` — full package tests pass

  Fixes #4548

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -258,9 +258,9 @@ func validateAutoScalingCompExtension(annotations map[string]string, compExtSpec
 	autoscalerClass := annotations[constants.AutoscalerClass]
 
 	switch deploymentMode {
-	case string(constants.Standard):
+	case string(constants.Standard), string(constants.LegacyRawDeployment):
 		switch autoscalerClass {
-		case string(constants.AutoscalerClassHPA):
+		case string(constants.AutoscalerClassHPA), "":
 			return validateScalingHPACompExtension(compExtSpec)
 		case string(constants.AutoscalerClassKeda):
 			return validateScalingKedaCompExtension(compExtSpec)
@@ -364,8 +364,9 @@ func validateScalingHPACompExtension(compExtSpec *ComponentExtensionSpec) error 
 					return errors.New("metricSpec.Resource is not set for resource metric source type")
 				}
 			default:
-				return fmt.Errorf("invalid HPA metric source type with value [%s],"+
-					"valid metric source types are Resource", metricType)
+				return fmt.Errorf("invalid HPA metric source type with value [%s], "+
+					"valid metric source types are Resource. "+
+					"For External or PodMetric types, set the annotation serving.kserve.io/autoscalerClass to keda", metricType)
 			}
 		}
 	}
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -366,7 +366,7 @@ func validateScalingHPACompExtension(compExtSpec *ComponentExtensionSpec) error 
 			default:
 				return fmt.Errorf("invalid HPA metric source type with value [%s], "+
 					"valid metric source types are Resource. "+
-					"For External or PodMetric types, set the annotation serving.kserve.io/autoscalerClass to keda", metricType)
+					"For External or PodMetric types, set the annotation %s to keda", metricType, constants.AutoscalerClass)
 			}
 		}
 	}
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -253,12 +253,12 @@ func validateMultiNodeVariables(isvc *InferenceService) error {
 
 // Validate scaling options component extensions
 func validateAutoScalingCompExtension(annotations map[string]string, compExtSpec *ComponentExtensionSpec) error {
-	deploymentMode := annotations["serving.kserve.io/deploymentMode"]
+	deploymentMode := constants.ParseDeploymentMode(annotations[constants.DeploymentMode])
 	annotationClass := annotations[autoscaling.ClassAnnotationKey]
 	autoscalerClass := annotations[constants.AutoscalerClass]
 
 	switch deploymentMode {
-	case string(constants.Standard), string(constants.LegacyRawDeployment):
+	case constants.Standard:
 		switch autoscalerClass {
 		case string(constants.AutoscalerClassHPA), "":
 			return validateScalingHPACompExtension(compExtSpec)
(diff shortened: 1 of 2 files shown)


Jooho:
@JohnsonWang1015  The code changes look good overall, but it seems the pre-commit checks and raw CI are failing. Would you mind taking a look and addressing them?

JohnsonWang1015:
Hello @yuzisun ,
Thanks for the review! I've verified the code locally. The pre-commit and E2E failures appear to be unrelated to my changes. Could you help re-run the failed CI jobs?

Thank you very much!

spolti:
@jlost @Jooho It seems to be related with https://github.com/kserve/kserve/pull/5427

pkg/apis/serving/v1beta1/inference_service_validation.go
jlost: Running this through `constants.ParseDeploymentMode` should eliminate the need to add additional cases.

JohnsonWang1015: @jlost Thanks for the review! I've addressed the feedback:
- Switched to constants.ParseDeploymentMode per @jlost's suggestion — this also aligns webhook validation with the controller's deployment-mode resolution.
- Replaced hard-coded deploymentMode annotation key with constants.DeploymentMode.
- Updated TestCustomizeDeploymentStrategyUnsupportedForServerless to explicitly set the Knative annotation (previously relied on an implicit default that no longer matches controller behavior).

PR #5339, opened 2026-04-02
fix(llmisvc): always set status.URL with a discovered address

pierDipi:
Promote first address to top-level status.URL as some "cluster external" addresses are technically within "virtual private networks" and we cannot detect that from just IPs.

Even if it's a cluster-local URL, the status URL is just for discovery and easy access, and it's not a problem to have a private URL while we continue to prioritize external addresses as before.

The status.URL being unset causes users to be confused about how to send requests to it


bartoszmajsak:
@pierDipi probably worth clarifying the CRD docs too

https://github.com/kserve/kserve/blob/cd2ef02cbc32565b23336d48344c9eda6332f247/pkg/apis/serving/v1alpha2/llm_inference_service_types.go#L489

bartoszmajsak (approved):
/lgtm
/approve

sivanantha321 (approved):
/lgtm
/approve

PR #5342, opened 2026-04-03
fix:  CVE-2026-33186 gRPC authorization bypass

sivanantha321:

Fixes CVE-2026-33186


go.sum
pierDipi, line 1: There are many new GCP dependencies is this expected?

sivanantha321: I just ran`go mod tidy` and this is the result. The precommit check also passing.

cjohannsen-cloudera (approved):
LGTM

Jooho (approved):
/lgtm
/approved

PR #5347, opened 2026-04-04
feat(ingress): add pathMatchType config flag for GKE Gateway compatibility

### LINKED ISSUE #5319: HTTPRoute RegularExpression path match breaks GKE Gateway — need configurable path match type
/kind bug

When using KServe v0.17.0 in Raw Deployment mode with Gateway API on GKE, the HTTPRoute reconciler unconditionally uses `PathMatchRegularExpression` for all path matches (`httproute_reconciler.go`, `createHTTPRouteMatch()`).

GKE's Gateway controller does not support `RegularExpression` path match type. It rejects the route with:

```
GWCER104: Paths must start with a '/' character "^/.*$"
```

This causes `reconcileHTTPRouteStatus()` to set `IngressReady=False`, so external ingress never comes up.

The path match type is hardcoded in `createHTTPRouteMatch()`:
```go
func createHTTPRouteMatch(prefix string) gwapiv1.HTTPRouteMatch {
    return gwapiv1.HTTPRouteMatch{
        Path: &gwapiv1.HTTPPathMatch{
            Type:  ptr.To(gwapiv1.PathMatchRegularExpression), // hardcoded
            Value: ptr.To(prefix),
        },
    }
}
```

The function is called 9 times across the reconciler. The most common pattern is `FallbackPrefix()` = `^/.*$` (6 of 9 call sites), which is functionally equivalent to `PathPrefix: /`.

Manual workaround (`kubectl patch --type=json` to change the path match type to `PathPrefix`) is temporary — the reconciler re-adds `RegularExpression` on the next loop.

HTTPRoutes generated by KServe should be accepted by the GKE Gateway controller. The path match type should be configurable so operators can use `PathPrefix` for Gateway implementations that don't support `RegularExpression`.

**Proposed fix:** Add a `pathMatchType` field to the ingress section of `inferenceservice-config`. When set to `"PathPrefix"`, the reconciler uses `PathMatchPathPrefix` with equivalent prefix paths instead of regex patterns. Default behavior is unchanged (`RegularExpression`) for controllers that support it (Envoy Gateway, Istio).

This follows the same pattern as `disableHTTPRouteTimeout` (#5311 / #5313), which was added to address another GKE Gateway incompatibility with the same reconciler.

Regex-to-prefix conversion table:

| Regex Pattern | PathPrefix Equivalent |
|---|---|
| `^/.*$` (FallbackPrefix) | `/` |
| `^/v1/models/[\\w-]+:explain$` (ExplainPrefix) | `/v1/models/` |
| `path + (/v1/models/[\\w-]+:explain)$` (PathBasedExplainPrefix) | `path + /v1/models/` |
| `path + /` (path-based routing) | `path + /` |

```yaml
kind: InferenceService
metadata:
  name: distilbert-sst2
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      storageUri: hf://distilbert/distilbert-base-uncased-finetuned-sst-2-english
      resources:
        requests:
          cpu: "1"
          memory: 2Gi
        limits:
          cpu: "2"
          memory: 4Gi
      args:
        - --task=sequence_classification
```

- The Gateway API spec marks `RegularExpression` path match as "Extended" conformance, meaning implementations are not required to support it ([Gateway API HTTPRouteMatch](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPPathMatch)).
- GKE Gateway only supports `Exact` and `PathPrefix` path match types.
- The newer `llmisvc` controller already uses `PathMatchPathPrefix` in its routing logic (`pkg/controller/v1alpha2/llmisvc/fixture/gwapi_builders.go`), suggesting a precedent for prefix-based matching.
- Related: #5311 / #5313 (timeout incompatibility — same reconciler, same GKE Gateway limitation pattern).
- Relevant code: `pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go` — `createHTTPRouteMatch()` and all `createRaw*HTTPRoute()` functions.
- Relevant code: `pkg/constants/constants.go` — `FallbackPrefix()`, `ExplainPrefix()`, `PathBasedExplainPrefix()`.

**Environment:**

- Istio Version: N/A (not using Istio, using GKE Gateway API)
- Knative Version: N/A (Raw Deployment mode)
- KServe Version: 0.17.0
- Kubeflow version: N/A
- Cloud Environment: GCP (GKE)
- Minikube/Kind version: N/A
- Kubernetes version: v1.34 (GKE)
- OS: GKE managed nodes (Container-Optimized OS)

sophieliu15:
## Summary

GKE's Gateway controller does not support `RegularExpression` path matches on HTTPRoutes (Extended conformance). KServe hardcodes `PathMatchRegularExpression` with patterns like `^/.*$` in `httproute_reconciler.go`, causing GKE Gateway to reject the route with `GWCER104: Paths must start with '/' character`. There is currently no config flag to change the path match type.

This PR adds a `pathMatchType` field to the `inferenceservice-config` ConfigMap (ingress section). When set to `"PathPrefix"`, the controller uses `PathMatchPathPrefix` with equivalent prefix paths instead of regex patterns. Default behavior is **unchanged** — existing users who rely on regex path matching are unaffected.

Follows the same pattern as `disableHTTPRouteTimeout` (#5313).

Fixes: #5319
Docs: kserve/website#646

## Changes

**Core Go code (3 files):**
- `pkg/apis/serving/v1beta1/configmap.go` — Added `PathMatchType string` to `IngressConfig`
- `pkg/controller/.../httproute_reconciler.go` — Added `resolvePathMatch()` helper that returns a `PathPrefix` match when the flag is set; updated all 9 path match call sites across `createRawPredictorHTTPRoute`, `createRawTransformerHTTPRoute`, `createRawExplainerHTTPRoute`, and `createRawTopLevelHTTPRoute`
- `pkg/controller/.../httproute_reconciler_test.go` — Added 7 test cases (5 unit + 2 integration)

**Helm charts, ConfigMaps, quick-install scripts, OpenAPI, Python SDK** — updated following the same pattern as #5313 (20 additional files via manual edits + `make generate`).

## Usage

```json
// inferenceservice-config ConfigMap, ingress section:
{
    "enableGatewayApi": true,
    "kserveIngressGateway": "kserve/kserve-ingress-gateway",
    "disableHTTPRouteTimeout": true,
    "pathMatchType": "PathPrefix"
}
```

When `pathMatchType` is empty or unset, the default `RegularExpression` matching is used.

## Test plan

- [x] Unit tests: all new and existing tests pass
- [x] **Manual cluster test on GKE**: Built a custom controller image from this branch, deployed to a GKE cluster (Gateway class `gke-l7-global-external-managed`), and verified:
  - HTTPRoutes use `PathPrefix: /` instead of `RegularExpression: ^/.*$`
  - GKE Gateway accepts the HTTPRoutes (`Accepted: True`, `ResolvedRefs: True`, `Reconciled: True`)
  - InferenceService reaches `Ready: True`
  - End-to-end inference works through the external load balancer: `curl -> GKE Gateway -> InferenceService -> {"predictions":[1,0]}`
  - Without the fix (stock v0.17.0), the same HTTPRoute is rejected by GKE Gateway with `GWCER104`

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
@@ -117,6 +117,22 @@ func createHTTPRouteMatch(prefix string) gwapiv1.HTTPRouteMatch {
 	}
 }
 
+// resolvePathMatch returns an HTTPRouteMatch with the appropriate path match type.
+// When pathMatchType is "PathPrefix", it uses PathMatchPathPrefix with the provided
+// prefixEquivalent instead of the regex pattern. This is needed for Gateway
+// implementations (e.g. GKE) that do not support RegularExpression path matches.
+func resolvePathMatch(pathMatchType string, regexPattern string, prefixEquivalent string) gwapiv1.HTTPRouteMatch {
+	if pathMatchType == "PathPrefix" {
+		return gwapiv1.HTTPRouteMatch{
+			Path: &gwapiv1.HTTPPathMatch{
+				Type:  ptr.To(gwapiv1.PathMatchPathPrefix),
+				Value: ptr.To(prefixEquivalent),
+			},
+		}
+	}
+	return createHTTPRouteMatch(regexPattern)
+}
+
 func addIsvcHeaders(name string, namespace string) gwapiv1.HTTPRouteFilter {
 	return gwapiv1.HTTPRouteFilter{
 		Type: gwapiv1.HTTPRouteFilterRequestHeaderModifier,
@@ -191,7 +207,7 @@ func createRawPredictorHTTPRoute(isvc *v1beta1.InferenceService, ingressConfig *
 		return nil, fmt.Errorf("failed to generate predictor ingress host: %w", err)
 	}
 	allowedHosts = append(allowedHosts, gwapiv1.Hostname(predictorHost))
-	routeMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(constants.FallbackPrefix())}
+	routeMatch := []gwapiv1.HTTPRouteMatch{resolvePathMatch(ingressConfig.PathMatchType, constants.FallbackPrefix(), "/")}
 	timeout := resolveTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Predictor.TimeoutSeconds)
 	httpRouteRules = append(httpRouteRules, createHTTPRouteRule(routeMatch, filters, predictorName, isvc.Namespace, constants.CommonDefaultHttpPort, timeout))
 
@@ -251,7 +267,7 @@ func createRawTransformerHTTPRoute(isvc *v1beta1.InferenceService, ingressConfig
 		return nil, fmt.Errorf("failed to generate transformer ingress host: %w", err)
 	}
 	allowedHosts = append(allowedHosts, gwapiv1.Hostname(transformerHost))
-	routeMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(constants.FallbackPrefix())}
+	routeMatch := []gwapiv1.HTTPRouteMatch{resolvePathMatch(ingressConfig.PathMatchType, constants.FallbackPrefix(), "/")}
 	timeout := resolveTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Transformer.TimeoutSeconds)
 	httpRouteRules = append(httpRouteRules, createHTTPRouteRule(routeMatch, filters, transformerName, isvc.Namespace,
 		constants.CommonDefaultHttpPort, timeout))
@@ -314,7 +330,7 @@ func createRawExplainerHTTPRoute(isvc *v1beta1.InferenceService, ingressConfig *
 	allowedHosts = append(allowedHosts, gwapiv1.Hostname(explainerHost))
 
 	// Add explainer host rules
-	routeMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(constants.FallbackPrefix())}
+	routeMatch := []gwapiv1.HTTPRouteMatch{resolvePathMatch(ingressConfig.PathMatchType, constants.FallbackPrefix(), "/")}
 	timeout := resolveTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Explainer.TimeoutSeconds)
 	httpRouteRules = append(httpRouteRules, createHTTPRouteRule(routeMatch, filters, explainerName, isvc.Namespace,
 		constants.CommonDefaultHttpPort, timeout))
@@ -405,7 +421,7 @@ func createRawTopLevelHTTPRoute(isvc *v1beta1.InferenceService, ingressConfig *v
 		timeout := resolveTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Explainer.TimeoutSeconds)
 		// Add toplevel host :explain route
 		// :explain routes to the explainer when there is only explainer
-		explainRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(constants.ExplainPrefix())}
+		explainRouteMatch := []gwapiv1.HTTPRouteMatch{resolvePathMatch(ingressConfig.PathMatchType, constants.ExplainPrefix(), "/v1/models/")}
 		httpRouteRules = append(httpRouteRules, createHTTPRouteRule(explainRouteMatch, filters,
 			explainerName, isvc.Namespace, constants.CommonDefaultHttpPort, timeout))
 	}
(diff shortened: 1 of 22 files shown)


sophieliu15:
Friendly ping! Happy to improve PR based on any suggestions!

sophieliu15:
Holding off on updates here until #5394 lands — see discussion on #5319. Plan is to rebase onto the new `rawDeployment` config block and move `pathMatchType` under it, while keeping the broader scope (predictor/transformer/explainer + host-based catch-all routes). Converting to draft in the meantime.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -117,6 +117,22 @@ func createHTTPRouteMatch(prefix string) gwapiv1.HTTPRouteMatch {
...
+		return gwapiv1.HTTPRouteMatch{
+			Path: &gwapiv1.HTTPPathMatch{
+				Type:  ptr.To(gwapiv1.PathMatchPathPrefix),
+				Value: ptr.To(prefixEquivalent),
+			},
+		}
+	}
+	return createHTTPRouteMatch(regexPattern)
+}
+
```
vivekk16, line 135: `resolvePathMatch` silently falls through to regex for any value that isn't "PathPrefix". If a user sets "pathprefix", "PathPrefx", or "Exact" (unsupported), there's no indication the config was ignored, they'll just see GKE rejecting routes with `GWCER104` and have no idea why.

Could we log a warning in `NewIngressConfig()` in `configmap.go,` after the `PathTemplate` validation, so the user sees what's happening? Something like this,

```
if ingressConfig.PathMatchType != "" && ingressConfig.PathMatchType != "PathPrefix" {
    log.Info("WARNING: unrecognized pathMatchType value, falling back to RegularExpression",
        "pathMatchType", ingressConfig.PathMatchType,
        "supportedValues", []string{"PathPrefix"})
    ingressConfig.PathMatchType = ""
}
```

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler_test.go
```diff
@@ -3420,3 +3420,106 @@ func TestCreateRawPredictorHTTPRouteDisableTimeout(t *testing.T) {
...
+		g.Expect(*match.Path.Value).To(Equal("/v1/models/"))
+	})
+
+	t.Run("PathPrefix with path-based routing", func(t *testing.T) {
+		match := resolvePathMatch("PathPrefix", "/serving/default/test-isvc/", "/serving/default/test-isvc/")
+		g.Expect(*match.Path.Type).To(Equal(gwapiv1.PathMatchPathPrefix))
+		g.Expect(*match.Path.Value).To(Equal("/serving/default/test-isvc/"))
+	})
+
+	t.Run("RegularExpression explicitly set", func(t *testing.T) {
```
vivekk16, line 3451: nit: The test `RegularExpression explicitly set `isn't actually testing explicit recognition. It passes because `RegularExpression` isn't `PathPrefix`, so it falls through to the default. `kserve` would pass the same way. Maybe rename to something like `unrecognized value falls back to RegularExpression`?

vivekk16 (commented):
When `pathMatchType` is `PathPrefix` and an InferenceService has both a `predictor` and an `explainer`, the top-level `HTTPRoute` creates:

- Rule 1 (explain): PathPrefix /v1/models/ → explainer service
- Rule 2 (fallback): PathPrefix / → predictor/transformer service

With regex, `ExplainPrefix()` (`^/v1/models/[\w-]+:explain$`) only matches `:explain` requests. But with `PathPrefix`, `/v1/models/` matches all requests under that path, including `/v1/models/my-model:predict`. Per Gateway API spec, the longest matching prefix wins, so predict traffic would route to the `explainer` instead of the predictor.

I understand there may not be a better prefix available here. `PathPrefix` simply cannot express the suffix anchoring (`:explain$`) or pattern matching (`[\w-]+`) that the regex uses. `/v1/models/` is the closest approximation.

Did you consider this scenario? 

If yes, a few follow-up questions:

- Is this an acceptable limitation for GKE users deploying predictor+explainer together? 

- If it is a calculated trade-off, it would be worth documenting in the companion website PR (kserve/website#646), which currently doesn't mention this.

- Would it make sense to add test cases covering all scenarios with `PathPrefix` ?

vivekk16:
Just looked at comments in #5319. Will take a look at the related PR.
@sophieliu15

sophieliu15:
Thanks for the careful review — the explainer routing point is the big one and you're right that `PathPrefix /v1/models/` shadows `:predict` and misroutes it.

Planning to go with an **Exact** match for the explain path + `PathPrefix /` fallback — let me know if you see a hole in this before I rebase, otherwise I'll put up the changes once #5394 lands. Per Gateway API precedence, `Exact` wins over `PathPrefix` regardless of rule order, so `:explain` hits the explainer and everything else (including `:predict`) falls through correctly. Exact is Core conformance, so GKE supports it. This is also tighter than the current regex, which matches `:explain` against any model name.

Plan for the next revision (after rebasing onto #5394's `rawDeployment` config block):

1. Replace the `PathPrefix` approximation with `Exact /v1/models/<isvc-name>:explain` + `PathPrefix /` fallback. Same treatment for v2 (`/v2/models/<name>/explain`) and for path-based routing (`/serving/<ns>/<isvc>/...`).
2. Add the validation warning in `NewIngressConfig()` for unrecognized `pathMatchType` values, per your first comment.
3. Rename `"RegularExpression explicitly set"` → `"unrecognized value falls back to RegularExpression"`.
4. Expand tests to cover predictor-only, predictor+explainer (both `:predict` and `:explain`), and predictor+transformer+explainer under `PathPrefix` mode.
5. Update kserve/website#646 to document the Exact-match behavior.

Holding these changes until #5394 merges to avoid a double rebase.

sophieliu15:
Holding these changes until https://github.com/kserve/kserve/pull/5394 merges to avoid a double rebase.

PR #5351, opened 2026-04-07
fix(charts): remove workloadType from ClusterStorageContainer resources

elukey:

The CRD ClusterStorageContainer doesn't contain any workloadType attribue, and in environments where CI checks if resources matches their spec it may end up in validation failures.

Remove workloadType from ClusterStorageContainer definitions.

Deployed the chart in my kubernetes environment and made sure CI passed tests.

**Release note**:
```release-note
NONE
```


elukey:
@Jooho Hi! I am struggling to figure out what I did wrong to fail the two checks :(

elukey:
Abandoning since the field needs to be there.

PR #5352, opened 2026-04-07
feat(llmisvc): add Managed DRA support for LLMInferenceService

### LINKED ISSUE #5294: Proposal: Add DRA (Dynamic Resource Allocation) Support for GPU Management
/kind feature

## Summary

Proposal to add first-class support for Kubernetes [Dynamic Resource Allocation (DRA)](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/) in KServe, alongside the existing device plugin model. DRA is now GA in Kubernetes v1.34 and provides attribute-based GPU selection, dynamic partitioning, and better scheduling awareness compared to device plugins. This builds on the initial discussion in #4834.

### Current State: Device Plugin Model

KServe currently manages GPU resources exclusively via the device plugin model:
```yaml
resources:
  limits:
    nvidia.com/gpu: "2"   # Opaque integer count
```

**Limitations:**
- **No attribute selection**: Cannot request specific GPU models (e.g., H100 vs A100), memory sizes, or capabilities
- **No topology awareness**: Scheduler doesn't understand NVLink/NVSwitch connectivity between GPUs
- **Static partitioning**: MIG profiles must be pre-configured; no dynamic GPU sharing
- **Privileged DaemonSets**: Device plugins require full host access

### DRA Model

DRA uses `DeviceClass` + `ResourceClaim` (analogous to `StorageClass` + `PVC`):
```yaml
resourceClaims:
  - name: gpu-claim
    resourceClaimTemplateName: gpu-template
---
apiVersion: resource.k8s.io/v1beta1
kind: ResourceClaimTemplate
metadata:
  name: gpu-template
spec:
  spec:
    devices:
      requests:
        - name: gpu
          deviceClassName: nvidia-gpu
          count: 2
          selectors:
            - cel:
                expression: "device.attributes['gpu.nvidia.com/memory'].isGreaterThan(quantity('40Gi'))"
```

**Advantages:**
- CEL-based attribute selection (memory, architecture, vendor capabilities)
- Topology-aware scheduling via ResourceSlice
- Dynamic MIG creation and GPU time-slicing
- Better security model (no privileged DaemonSets)

### Ecosystem Readiness

- **Kubernetes**: DRA GA in v1.34
- **NVIDIA**: Production DRA driver ([k8s-dra-driver-gpu](https://github.com/NVIDIA/k8s-dra-driver-gpu))
- **Cloud providers**: AKS, GKE, EKS, Alibaba ACK all support DRA
- **NIM Operator 3.0**: Already bridges KServe + DRA externally
- **OpenShift 4.20**: Integrates KServe with DRA for GPU management

KServe should support DRA natively rather than relying on external operators to bridge the gap.

## Current DRA State in KServe

The `ResourceClaims` field already exists in KServe's PodSpec (`pkg/apis/serving/v1beta1/podspec.go:309-325`) as a **pass-through** to Kubernetes. However:
- No controller logic creates or manages ResourceClaims/ResourceClaimTemplates
- GPU count extraction (`computeRayNodeAndGPUs()`) only reads from container `resources.requests`
- Deployment reconciler (`addGPUResourceToDeployment()`) always injects device-plugin style resources
- Validation webhooks may reject DRA-based specs as "unknown GPU resource type"

### Phase 1: Pass-through DRA Support (Non-breaking)

Allow users to specify DRA ResourceClaims that KServe passes through correctly without interfering.

| Task | Files |
|------|-------|
| Update `IsGPUEnabled()` to detect GPU via ResourceClaims, not just resource requests | `pkg/utils/utils.go` |
| Update `computeRayNodeAndGPUs()` to handle DRA-based GPU count (from ResourceClaim count or annotation) | `pkg/controller/v1beta1/inferenceservice/components/predictor.go` |
| Update `addGPUResourceToDeployment()` to skip device-plugin injection when ResourceClaims are present | `pkg/controller/v1beta1/inferenceservice/reconcilers/deployment/deployment_reconciler.go` |
| Update validation to not reject DRA-based GPU specs | `pkg/apis/serving/v1beta1/inference_service_validation.go` |
| Ensure LLMInferenceService (v1alpha2) passes ResourceClaims to generated LeaderWorkerSets | `pkg/apis/serving/v1alpha2/`, controller |
| E2E tests with DRA-enabled cluster | `test/e2e/` |

### Phase 2: Managed DRA (Controller-generated ResourceClaimTemplates)

KServe controller auto-generates ResourceClaimTemplates based on higher-level GPU specifications.

**New field or annotation for DeviceClass selection:**
```yaml
kind: InferenceService
metadata:
  name: llm-with-dra
  annotations:
    serving.kserve.io/device-class: "nvidia-gpu-h100"
    serving.kserve.io/gpu-memory-min: "80Gi"
spec:
  predictor:
    model:
      modelFormat:
        name: huggingface
      runtime: kserve-vllm
      storageUri: hf://meta-llama/Llama-3-8B
    resourceClaims:
      - name: gpu-claim
        resourceClaimTemplateName: gpu-template
    containers:
      - name: kserve-container
        resources:
          claims:
            - name: gpu-claim
```

| Task | Description |
|------|-------------|
| Add `DeviceClass` field or annotation to InferenceService spec | Allow users to specify GPU class without writing full ResourceClaimTemplate |
| Controller generates ResourceClaimTemplate from spec | Simplifies user experience |
| GPU count extraction from generated claims | Automatic parallelism configuration |

## Backwards Compatibility

- **DRA is strictly opt-in**: Device plugin model remains the default
- **Detection logic**: If `ResourceClaims` are present in the spec, use DRA path; otherwise, use device-plugin path
- **No breaking changes**: All existing InferenceService specs continue to work unchanged
- **Minimum K8s version**: Document that DRA requires Kubernetes 1.31+ (beta) or 1.34+ (GA)

## Risks

| Risk | Severity | Mitigation |
|------|----------|------------|
| Backwards compatibility breakage | High | DRA is opt-in; device-plugin remains default path |
| GPU count extraction differs between models | Medium | Use ResourceClaim `count` field; fall back to annotation |
| K8s version requirement | Medium | Feature-gate the DRA code path; clear documentation |
| DRA driver availability (only NVIDIA mature) | Medium | Start with NVIDIA; design for extensibility |
| E2E testing needs DRA-enabled clusters | High | Use kind with DRA feature gate for CI |

## Related

- #4834 — Initial DRA question
- [NVIDIA DRA Driver](https://github.com/NVIDIA/k8s-dra-driver-gpu)
- [Kubernetes DRA documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
- [NIM Operator 3.0 with DRA + KServe](https://developer.nvidia.com/blog/deploy-scalable-ai-inference-with-nvidia-nim-operator-3-0-0)

anshulmav:
This PR adds Managed DRA (Dynamic Resource Allocation) support to the v1alpha2 `LLMInferenceService` controller. 

It allows users to request GPU resources dynamically by adding the `serving.kserve.io/dra-device-class` and `serving.kserve.io/dra-gpu-count` annotations to their `LLMInferenceService`. The controller will automatically generate the required `ResourceClaimTemplates` and inject the `ResourceClaims` into the PodSpecs for both single-node and multi-node (LeaderWorkerSet) deployments.

Partially Fixes #5294

- [x] Unit tests added for DRA helper functions (`managed_dra_test.go`)
- [x] Integration tests added for single-node DRA injection (`controller_int_test.go`)
- [x] Integration tests added for multi-node LeaderWorkerSet DRA injection (`controller_int_multi_node_test.go`)
- [x] Manually verified on a local `kind` cluster with DRA feature gates enabled.

Opened as a draft to verify Prow presubmit tests.

**Checklist**:

**Release note**:
```release-note
feat(llmisvc): add Managed DRA support for LLMInferenceService deployments

### DIFF
--- a/config/crd/full/serving.kserve.io_clusterservingruntimes.yaml
+++ b/config/crd/full/serving.kserve.io_clusterservingruntimes.yaml
@@ -14,476 +14,581 @@ spec:
     singular: clusterservingruntime
   scope: Cluster
   versions:
-    - additionalPrinterColumns:
-        - jsonPath: .spec.disabled
-          name: Disabled
-          type: boolean
-        - jsonPath: .spec.supportedModelFormats[*].name
-          name: ModelType
-          type: string
-        - jsonPath: .spec.containers[*].name
-          name: Containers
-          type: string
-        - jsonPath: .metadata.creationTimestamp
-          name: Age
-          type: date
-      name: v1alpha1
-      schema:
-        openAPIV3Schema:
-          properties:
-            apiVersion:
-              type: string
-            kind:
-              type: string
-            metadata:
-              type: object
-            spec:
-              properties:
-                affinity:
-                  properties:
-                    nodeAffinity:
-                      properties:
-                        preferredDuringSchedulingIgnoredDuringExecution:
-                          items:
-                            properties:
-                              preference:
-                                properties:
-                                  matchExpressions:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  matchFields:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                type: object
-                                x-kubernetes-map-type: atomic
-                              weight:
-                                format: int32
-                                type: integer
-                            required:
-                              - preference
-                              - weight
-                            type: object
-                          type: array
-                          x-kubernetes-list-type: atomic
-                        requiredDuringSchedulingIgnoredDuringExecution:
+  - additionalPrinterColumns:
+    - jsonPath: .spec.disabled
+      name: Disabled
+      type: boolean
+    - jsonPath: .spec.supportedModelFormats[*].name
+      name: ModelType
+      type: string
+    - jsonPath: .spec.containers[*].name
+      name: Containers
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              affinity:
+                properties:
+                  nodeAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
                           properties:
-                            nodeSelectorTerms:
+                            preference:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                matchFields:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                              type: object
+                              x-kubernetes-map-type: atomic
+                            weight:
+                              format: int32
+                              type: integer
+                          required:
+                          - preference
+                          - weight
+                          type: object
+                        type: array
+                        x-kubernetes-list-type: atomic
+                      requiredDuringSchedulingIgnoredDuringExecution:
+                        properties:
+                          nodeSelectorTerms:
+                            items:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                matchFields:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                              type: object
+                              x-kubernetes-map-type: atomic
+                            type: array
+                            x-kubernetes-list-type: atomic
+                        required:
+                        - nodeSelectorTerms
+                        type: object
+                        x-kubernetes-map-type: atomic
+                    type: object
+                  podAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            podAffinityTerm:
+                              properties:
+                                labelSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                            x-kubernetes-list-type: atomic
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                      x-kubernetes-list-type: atomic
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                  x-kubernetes-map-type: atomic
+                                matchLabelKeys:
+                                  items:
+                                    type: string
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                mismatchLabelKeys:
+                                  items:
+                                    type: string
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                namespaceSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
+                                              type: string
+                                            type: array
+                                            x-kubernetes-list-type: atomic
+                                        required:
+                                        - key
+                                        - operator
+                                        type: object
+                                      type: array
+                                      x-kubernetes-list-type: atomic
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                  x-kubernetes-map-type: atomic
+                                namespaces:
+                                  items:
+                                    type: string
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                topologyKey:
+                                  type: string
+                              required:
+                              - topologyKey
+                              type: object
+                            weight:
+                              format: int32
+                              type: integer
+                          required:
+                          - podAffinityTerm
+                          - weight
+                          type: object
+                        type: array
+                        x-kubernetes-list-type: atomic
+                      requiredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            labelSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                              x-kubernetes-map-type: atomic
+                            matchLabelKeys:
                               items:
-                                properties:
-                                  matchExpressions:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  matchFields:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                type: object
-                                x-kubernetes-map-type: atomic
+                                type: string
                               type: array
                               x-kubernetes-list-type: atomic
+                            mismatchLabelKeys:
+                              items:
+                                type: string
+                              type: array
+                              x-kubernetes-list-type: atomic
+                            namespaceSelector:
+                              properties:
+                                matchExpressions:
+                                  items:
+                                    properties:
+                                      key:
+                                        type: string
+                                      operator:
+                                        type: string
+                                      values:
+                                        items:
+                                          type: string
+                                        type: array
+                                        x-kubernetes-list-type: atomic
+                                    required:
+                                    - key
+                                    - operator
+                                    type: object
+                                  type: array
+                                  x-kubernetes-list-type: atomic
+                                matchLabels:
+                                  additionalProperties:
+                                    type: string
+                                  type: object
+                              type: object
+                              x-kubernetes-map-type: atomic
+                            namespaces:
+                              items:
+                                type: string
+                              type: array
+                              x-kubernetes-list-type: atomic
+                            topologyKey:
+                              type: string
                           required:
-                            - nodeSelectorTerms
+                          - topologyKey
                           type: object
-                          x-kubernetes-map-type: atomic
-                      type: object
-                    podAffinity:
-                      properties:
-                        preferredDuringSchedulingIgnoredDuringExecution:
-                          items:
-                            properties:
-                              podAffinityTerm:
-                                properties:
-                                  labelSelector:
-                                    properties:
-                                      matchExpressions:
-                                        items:
-                                          properties:
-                                            key:
+                        type: array
+                        x-kubernetes-list-type: atomic
+                    type: object
+                  podAntiAffinity:
+                    properties:
+                      preferredDuringSchedulingIgnoredDuringExecution:
+                        items:
+                          properties:
+                            podAffinityTerm:
+                              properties:
+                                labelSelector:
+                                  properties:
+                                    matchExpressions:
+                                      items:
+                                        properties:
+                                          key:
+                                            type: string
+                                          operator:
+                                            type: string
+                                          values:
+                                            items:
                                               type: string
-                                            operator:
-                                              type: string
-                                            values:
-                                              items:
-                                                type: string
-                                              type: array
-                                              x-kubernetes-list-type: atomic
-                                          required:
-                                            - key
-                                            - operator
-                                          type: object
-                                        type: array
-                                        x-kubernetes-list-type: atomic
-                                      matchLabels:
-                                        additionalProperties:
-                                          type: string
+                                            type: array
+                                            x-kubernetes-list-type: atomic
+                                        required:
+                                        - key
+                                        - operator
                                         type: object
-                                    type: object
-                                    x-kubernetes-map-type: atomic
-                                  matchLabelKeys:
-                                    items:
-                                      type: string
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  mismatchLabelKeys:
-                                    items:
-                                      type: string
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  namespaceSelector:
-                                    properties:
-                                      matchExpressions:
-                                        items:
-                                          properties:
-                                            key:
-                                              type: string
-                                            operator:
-                                              type: string
-                                            values:
-                                              items:
-                                                type: string
-                                              type: array
-                                              x-kubernetes-list-type: atomic
-                                          required:
-                                            - key
-                                            - operator
-                                          type: object
-                                        type: array
-                                        x-kubernetes-list-type: atomic
-                                      matchLabels:
-                                        additionalProperties:
-                                          type: string
-                                        type: object
-                                    type: object
-                                    x-kubernetes-map-type: atomic
-                                  namespaces:
-                                    items:
-                                      type: string
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  topologyKey:
+                                      type: array
+                                      x-kubernetes-list-type: atomic
+                                    matchLabels:
+                                      additionalProperties:
+                                        type: string
+                                      type: object
+                                  type: object
+                                  x-kubernetes-map-type: atomic
+                                matchLabelKeys:
+                                  items:
                                     type: string
-                                required:
-                                  - topologyKey
-                                type: object
-                              weight:
-                                format: int32
-                                type: integer
-                            required:
-                              - podAffinityTerm
-                              - weight
-                            type: object
-                          type: array
-                          x-kubernetes-list-type: atomic
-                        requiredDuringSchedulingIgnoredDuringExecution:
-                          items:
-                            properties:
-                              labelSelector:
-                                properties:
-                                  matchExpressions:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  matchLabels:
-                                    additionalProperties:
-                                      type: string
-                                    type: object
-                                type: object
-                                x-kubernetes-map-type: atomic
-                              matchLabelKeys:
-                                items:
-                                  type: string
-                                type: array
-                                x-kubernetes-list-type: atomic
-                              mismatchLabelKeys:
-                                items:
-                                  type: string
-                                type: array
-                                x-kubernetes-list-type: atomic
-                              namespaceSelector:
-                                properties:
-                                  matchExpressions:
-                                    items:
-                                      properties:
-                                        key:
-                                          type: string
-                                        operator:
-                                          type: string
-                                        values:
-                                          items:
-                                            type: string
-                                          type: array
-                                          x-kubernetes-list-type: atomic
-                                      required:
-                                        - key
-                                        - operator
-                                      type: object
-                                    type: array
-                                    x-kubernetes-list-type: atomic
-                                  matchLabels:
-                                    additionalProperties:
-                                      type: string
-                                    type: object
-                                type: object
-                                x-kubernetes-map-type: atomic
-                              namespaces:
-                                items:
-                                  type: string
-                                type: array
-                                x-kubernetes-list-type: atomic
-                              topologyKey:
-                                type: string
-                            required:
-                              - topologyKey
-                            type: object
-                          type: array
-                          x-kubernetes-list-type: atomic
-                      type: object
-                    podAntiAffinity:
-                      properties:
-                        preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: pkg/controller/v1alpha2/llmisvc/managed_dra.go [hunk 1]
```diff
@@ -0,0 +1,202 @@
...
+	return equality.Semantic.DeepEqual(expected.Spec, curr.Spec)
+}
+
+func semanticResourceClaimIsEqual(expected *resourcev1.ResourceClaim, curr *resourcev1.ResourceClaim) bool {
+	return equality.Semantic.DeepEqual(expected.Spec, curr.Spec)
+}
+
+// injectManagedDRA wires the managed DRA claim into the PodSpec:
+//   - Adds a pod-level resourceClaim entry pointing to the generated ResourceClaim or ResourceClaimTemplate
+//   - Adds container-level resources.claims entries so every container can access the GPU(s)
```
cjohannsen-cloudera: This injects into all containers, correct? If a pod has sidecar containers (e.g., routing proxy, storage init), they'll also get GPU claims.  Is this the desired outcome or do we need to filter?

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,202 @@
...
+		requests[i] = req
+	}
+	return requests
+}
+
+// reconcileManagedDRA creates or updates the ResourceClaim/ResourceClaimTemplate
+// that backs managed DRA for this LLMInferenceService.
+func (r *LLMISVCReconciler) reconcileManagedDRA(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+	if !hasManagedDRA(llmSvc) {
+		return nil
```
cjohannsen-cloudera: If llmSvc previous had managed DRA and now it does not (managed DRA has been disabled) this will leave orphaned ResourceClaim/ResourceClaimTemplates in the cluster.  These should be cleaned up to prevent accumulation or they will not get deleted until the llmisvc is deleted.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,202 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
+)
+
+const (
+	managedDRASuffix      = "-managed-dra"
+	managedDRAClaimName   = "managed-gpu"
```
cjohannsen-cloudera: Is GPU the only managed claim type?

config/crd/full/serving.kserve.io_clusterservingruntimes.yaml
Jooho, line 1: why the format of crd files change?  does not relate to this pr.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
Jooho, line 1: This PR covers LLMISVC only, but Issue #5294 scopes both Phase 1 (pass-through) and Phase 2 (managed) for ISVC as well. Is there a plan for ISVC DRA support, or is this LLMISVC-only for now?

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,202 @@
...
+// reconcileManagedDRA creates or updates the ResourceClaim/ResourceClaimTemplate
+// that backs managed DRA for this LLMInferenceService.
+func (r *LLMISVCReconciler) reconcileManagedDRA(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+	if !hasManagedDRA(llmSvc) {
+		return nil
+	}
+
+	deviceClass := llmSvc.Annotations[constants.ManagedDRADeviceClassAnnotationKey]
+	celSelector := llmSvc.Annotations[constants.ManagedDRACelSelectorAnnotationKey]
+	isShared := llmSvc.Annotations[constants.ManagedDRASharingAnnotationKey] == "true"
```
Jooho: Curious about the use case for dra-sharing: true in the context of LLM serving. vLLM typically takes exclusive ownership of GPU memory, so multiple replica pods sharing the same ResourceClaim would likely OOM. Is there a specific scenario where shared mode makes sense for LLMInferenceService?

Jooho (commented):
We discussed this earlier in the community meeting — you asked whether annotation or API field is better, and I suggested going with API fields. Still think that's the right direction.

Annotations work as a quick prototype, but for something that's going to be a core part of how users request GPUs:
- No admission validation — invalid dra-gpu-count like "abc" only fails at reconcile
- No type safety — everything is a string
- Not discoverable via kubectl explain                          
- Easy to typo annotation keys silently                                                            
                                                                  
Since v1alpha2 is the right place to experiment with new API surface, I'd lean toward adding DRA fields to the spec directly rather than keeping annotations long-term.

pierDipi (commented):
I left a comment on the doc as well, I think if we want first-class support for this automation, typed APIs are better than annotations. 

Since it's a UX improvement, it seems better to support typed APIs from the start (and decide the typed API shape now) rather than supporting annotations and being stuck with supporting annotations forever.

anshulmav:
@Jooho @pierDipi  Thanks for the feedback! I agree that typed APIs are the best long-term goal for stable features.

Before I move these fields into the v1alpha2 spec, I wanted to share a few thoughts we had about the scope of this feature, just to get your perspective. We initially leaned toward annotations because we see this "Managed DRA" feature as a quick, experimental way for users to try out DRA claims.

With only these three fields (gpu-count, device-class,cel-selectors), we are only covering the most basic use cases. If we add them as official API fields to the CRD, the only real advantage for the user is that it saves them one step of reconciliation (they don't have to write a ResourceClaimTemplate and attach it to the PodSpec). But for any topology that is even slightly more complex, they will have to manually write that ResourceClaimTemplate anyway.

Because of that, adding these fields feels like we are permanently bloating the CRD for a very narrow, experimental convenience feature. We also run into a slippery slope: where do we stop adding fields? To support complex topologies natively, we’d have to keep adding more and more fields until our CRD basically just reinvents the ResourceClaimTemplate spec.

And to address the very valid concerns about type safety and silent failures (like someone typing "abc" for gpu-count), we can add strict validation to the KServe admission webhook. This gives us the immediate failure and safety of an API, without the permanent CRD commitment.

Given that this is meant to be a lightweight, experimental effort rather than a comprehensive DRA API, I'd love to hear your thoughts. Do you feel the UX benefits of having these in the CRD outweigh the potential API bloat, or would incubating them as validated annotations be a safer middle ground for now?

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: I've updated the injectManagedDRA logic to filter and only inject the GPU claims into the main container.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: I've added a cleanupManagedDRA function that automatically deletes any orphaned ResourceClaimTemplate objects if a user disables Managed DRA on their service.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: yes

config/crd/full/serving.kserve.io_clusterservingruntimes.yaml
anshulmav: fixed.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: llmisvc only for now.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: dra-sharing was there initially to support a very specific use case where a user might want to share the GPU amongst 2 pods (with reduced GPU utilization configured to avoid OOM). However, since we are moving forward with broader and simpler use cases I have completely removed this annotation.

pierDipi:
@anshulmav a couple of thoughts:

- given the experimental nature, should that be explicit in the annotation keys?
- should we record this decision somewhere (comment on the constants or managed_dra.go level comment or anything that would record the intentionally limited scope) that we won't support the full breath of DRA in this feature and we won't accept extensions? or is it too restrictive as of now?

anshulmav:
@pierDipi 
> @anshulmav a couple of thoughts:
> 
> * given the experimental nature, should that be explicit in the annotation keys?
> * should we record this decision somewhere (comment on the constants or managed_dra.go level comment or anything that would record the intentionally limited scope) that we won't support the full breath of DRA in this feature and we won't accept extensions? or is it too restrictive as of now?

Yeah I agree with making the experimental nature explicit. I have updated the annotation keys to prefix them with `exp` (e.g., `serving.kserve.io/exp-dra-gpu-count`). And i think a hard "no extensions ever" rule might be a bit too restrictive right now. Instead, I have added a package-level comment in `managed_dra.go` and on the constants documenting that this is an intentionally limited-scope feature for basic DRA use cases, and that complex topologies should use native `ResourceClaimTemplates` directly. This leaves room for minor, practical additions if they become necessary.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,244 @@
...
+
+// hasManagedDRA checks if the LLMInferenceService has the required annotations to enable Managed DRA.
+func hasManagedDRA(llmSvc *v1alpha2.LLMInferenceService) bool {
+	_, ok := llmSvc.Annotations[constants.ManagedDRADeviceClassAnnotationKey]
+	return ok
+}
+
+// managedDRAResourceName generates the name of the ResourceClaimTemplate.
+func managedDRAResourceName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return llmSvc.GetName() + managedDRASuffix
```
pierDipi: to prevent longer than possible names, we need to use ChildName utility as used in other places

pkg/constants/constants.go
```diff
@@ -137,6 +137,13 @@ var (
 	DisableAutoUpdateAnnotationKey              = KServeAPIGroupName + "/disable-auto-update"
 	ModelFormatAnnotationKey                    = "modelFormat"
 	InferencePoolMigratedAnnotationKey          = KServeAPIGroupName + "/inferencepool-migrated"
+	// Managed DRA Experimental Annotations
+	// These annotations provide an intentionally limited-scope convenience feature for basic DRA use cases.
+	// Complex DRA topologies should use native Kubernetes ResourceClaimTemplate objects directly.
+	// These serve as a strong guideline to prevent feature bloat while leaving room for minor, practical additions.
+	ManagedDRADeviceClassAnnotationKey = KServeAPIGroupName + "/exp-dra-device-class"
+	ManagedDRACelSelectorAnnotationKey = KServeAPIGroupName + "/exp-dra-cel-selector"
+	ManagedDRAGpuCountAnnotationKey    = KServeAPIGroupName + "/exp-dra-gpu-count"
```
pierDipi: should we replace `gpu` and use `device` as it's not GPU specific?


```suggestion
	ManagedDRADeviceCountAnnotationKey    = KServeAPIGroupName + "/exp-dra-device-count"
```

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,244 @@
...
+// cleanupManagedDRA removes any previously generated ResourceClaimTemplate
+// owned by this LLMInferenceService.
+func (r *LLMISVCReconciler) cleanupManagedDRA(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+	ok, err := utils.IsCrdAvailable(r.Config, resourcev1.SchemeGroupVersion.String(), "ResourceClaimTemplate")
+	if err != nil {
+		return fmt.Errorf("failed to check if ResourceClaimTemplate CRD is available: %w", err)
+	}
+	if !ok {
+		return nil // Nothing to clean up if the API doesn't exist
+	}
```
pierDipi: Delete already ignores no Kind errors
```
		if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
			return nil
		}
```

```suggestion
```

pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
pierDipi, line 1: if we call `injectManagedDRA` in `config_merge.go` we wouldn't need to call it in every if branch and would handle both single and multi node cases

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,244 @@
...
+	managedDRAClaimName   = "managed-gpu"
+	defaultManagedDRAName = "gpu"
+	llmMainContainerName  = "main"
+)
+
+// hasManagedDRA checks if the LLMInferenceService has the required annotations to enable Managed DRA.
+func hasManagedDRA(llmSvc *v1alpha2.LLMInferenceService) bool {
+	_, ok := llmSvc.Annotations[constants.ManagedDRADeviceClassAnnotationKey]
+	return ok
+}
```
pierDipi: the other condition for the cleanup is the stop feature, see similar usage for `utils.GetForceStopRuntime(llmSvc)`

pkg/controller/v1alpha2/llmisvc/managed_dra.go
pierDipi: @anshulmav can you expand on why that limitation?

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/utils"
+)
+
+const (
+	managedDRASuffix      = "-managed-dra"
+	managedDRAClaimName   = "managed-gpu"
+	defaultManagedDRAName = "gpu"
+	llmMainContainerName  = "main"
```
saileshd1402: llmisvc does not define a default container name like isvc("kserve-container"), so adding ResourceClaim only to container "main" may be too strict. Current samples appear single-container, but to avoid future breakage we could:
(a) allow explicit container names via annotation
(b) deafult to first container of pod when annotation is absent.

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -588,3 +592,85 @@ func ValidateWorkloadScaling(basePath *field.Path, workload *WorkloadSpec) field
...
+	}
+
+	if hasDeviceClass {
+		if trimmed := strings.TrimSpace(deviceClass); trimmed == "" {
+			allErrs = append(allErrs, field.Invalid(
+				annotationsPath.Key(constants.ManagedDRADeviceClassAnnotationKey),
+				deviceClass,
+				"device class must not be empty",
+			))
+		} else if !dnsLabelLikePattern.MatchString(trimmed) {
```
saileshd1402: can we use `validation.IsDNS1123Subdomain(trimmed)` from `"k8s.io/apimachinery/pkg/util/validation"`? I think this is what DRA code uses as well. Please check once

pkg/controller/v1alpha2/llmisvc/workload_multi_node.go
anshulmav: done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: fixed.

pkg/constants/constants.go
anshulmav: yess , done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: @pierDipi  Initially I was more focused on the GPU side, but yeah I've made it generic to any accelerator now since DRA itself isn't GPU-specific. The constant is now `managedDRAClaimName = "managed-device"`.

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
anshulmav: yess , done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: Agreed, implemented both (a) and (b) exactly as you suggested. Added serving.kserve.io/exp-dra-container-name with a fallback to the first container when the annotation is absent.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,272 @@
...
+// that backs managed DRA for this LLMInferenceService.
+func (r *LLMISVCReconciler) reconcileManagedDRA(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+	if !hasManagedDRA(llmSvc) || utils.GetForceStopRuntime(llmSvc) {
+		return r.cleanupManagedDRA(ctx, llmSvc)
+	}
+
+	// Return an error if the cluster does not support the DRA ResourceClaimTemplate API.
+	if ok, err := utils.IsCrdAvailable(r.Config, resourcev1.SchemeGroupVersion.String(), "ResourceClaimTemplate"); err != nil || !ok {
+		return fmt.Errorf("managed DRA is requested but the ResourceClaimTemplate API (%s) is not available in this cluster", resourcev1.SchemeGroupVersion.String())
+	}
```
pierDipi: Can we remove this one? `IsCrdAvailable` uses a cache, if I install it later it would auto-resolve without requiring a controller restart while getting the same error naturally through `Reconcile`

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,272 @@
...
+
+	deviceClass := llmSvc.Annotations[constants.ManagedDRADeviceClassAnnotationKey]
+	celSelectors := parseManagedDRACelSelectors(llmSvc)
+
+	deviceCount, err := parseManagedDRADeviceCount(llmSvc)
+	if err != nil {
+		return err
+	}
+
+	deviceRequests := buildDeviceRequests(deviceClass, celSelectors, deviceCount)
```
pierDipi: nit: this could be encapsulated in `expectedManagedDRATemplate` and would follow other resources pattern

pierDipi: I would also put the main managed DRA method, reconcileManagedDRA at the very top of the file

pierDipi (commented):
Is it possible to add a sample in `docs/samples/llmisvc/managed-dra`?

config/rbac/llmisvc/role.yaml
pierDipi, line 1: given this change, you might need to run `make precommit` (in particular manifests, chart, etc regeneration)

pkg/controller/v1alpha2/llmisvc/managed_dra.go
```diff
@@ -0,0 +1,272 @@
...
+
+	parts := strings.Split(raw, "\n")
+	selectors := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if s := strings.TrimSpace(p); s != "" {
+			selectors = append(selectors, s)
+		}
+	}
+	return selectors
+}
```
pierDipi: if we encapsulate these in a function on the llmisvc type, we can reuse them in validation as now the parsing is repeated there too (eg in `pkg/apis/serving/v1alpha2/llm_inference_service_types_func.go`)

pierDipi: Same for `hasManagedDRA`

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: fixed

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: done.

anshulmav:
@pierDipi 
> Is it possible to add a sample in `docs/samples/llmisvc/managed-dra`?

added some example YAMLs and a README

config/rbac/llmisvc/role.yaml
anshulmav: done.

pkg/controller/v1alpha2/llmisvc/managed_dra.go
anshulmav: done.

docs/samples/llmisvc/managed-dra/llm-inference-service-managed-dra-multi-device.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: serving.kserve.io/v1alpha2
+kind: LLMInferenceService
+metadata:
+  name: qwen2-7b-instruct-dra-multi
+  # Multi-device claim with CEL selectors and explicit container targeting.
+  annotations:
+    serving.kserve.io/exp-dra-device-class: gpu.nvidia.com
+    serving.kserve.io/exp-dra-device-count: "2"
+    serving.kserve.io/exp-dra-container-name: vllm
```
pierDipi:
```suggestion
    serving.kserve.io/exp-dra-container-name: main
```

docs/samples/llmisvc/managed-dra/llm-inference-service-managed-dra-multi-device.yaml
```diff
@@ -0,0 +1,47 @@
...
+  template:
+    containers:
+      - name: sidecar
+        image: busybox:1.36
+        command: [ "sh", "-c", "while true; do sleep 3600; done" ]
+        resources:
+          requests:
+            cpu: 100m
+            memory: 64Mi
+      - name: vllm
```
pierDipi:
```suggestion
      - name: main
```

pierDipi: `main` is the one assumed by LLMInferenceServiceConfig and will always be there (merge semantic), if that's not the case, we have a bug somewhere

docs/samples/llmisvc/managed-dra/README.md
```diff
@@ -0,0 +1,189 @@
...
+kubectl apply -f llm-inference-service-managed-dra-multi-device.yaml
+```
+
+**YAML snippet:**
+```yaml
+metadata:
+  annotations:
+    serving.kserve.io/exp-dra-device-class: gpu.nvidia.com
+    serving.kserve.io/exp-dra-device-count: "2"
+    serving.kserve.io/exp-dra-container-name: vllm
```
pierDipi:
```suggestion
    serving.kserve.io/exp-dra-container-name: main
```

docs/samples/llmisvc/managed-dra/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+### 2. Full Surface - multi-device, CEL filters, container targeting ([llm-inference-service-managed-dra-multi-device.yaml](llm-inference-service-managed-dra-multi-device.yaml))
+
+Combines all four `exp-dra-*` annotations in one file to showcase the full surface. Each
+annotation is independent, use whichever subset matches your workload:
+
+- `exp-dra-device-count: "2"` requests multiple devices (default is `1`).
+- `exp-dra-cel-selector` filters eligible devices. Each non-empty line of the YAML `|`
+  block scalar becomes one selector, and all selectors must evaluate to true.
+- `exp-dra-container-name: vllm` routes the claim to the named container instead of the
```
pierDipi:
```suggestion
- `exp-dra-container-name: main` routes the claim to the named container instead of the
```

docs/samples/llmisvc/managed-dra/README.md
```diff
@@ -0,0 +1,189 @@
...
+  first one, only relevant when the pod template has multiple containers.
+
+**Configuration:**
+- Model: Qwen2.5-7B-Instruct, 1 replica
+- Devices: 2 of class `gpu.nvidia.com`, filtered by `type == 'A100'` AND `memory >= 40Gi`
+- Containers: `sidecar` (first) and `vllm`; claim targets `vllm`
```
pierDipi:
```suggestion
- Containers: `sidecar` (first) and `main`; claim targets `main`
```

pierDipi:
I guess I don't have enough power to trigger tests :sob:

spolti:
I tried to approve the ci, but it didn't work, @anshulmav do you mind sending a new commit, or an empty one, or one addressing any review suggestions?

docs/samples/llmisvc/managed-dra/llm-inference-service-managed-dra-multi-device.yaml
anshulmav: done

docs/samples/llmisvc/managed-dra/llm-inference-service-managed-dra-multi-device.yaml
anshulmav: done

docs/samples/llmisvc/managed-dra/README.md
anshulmav: done

docs/samples/llmisvc/managed-dra/README.md
anshulmav: done

docs/samples/llmisvc/managed-dra/README.md
anshulmav: done

anshulmav:
> I tried to approve the ci, but it didn't work, @anshulmav do you mind sending a new commit, or an empty one, or one addressing any review suggestions?

@spolti raised a new commit ...

pierDipi:
Thanks @anshulmav, this is great!

/lgtm

PR #5071, opened 2026-02-13
test: remove blanket sleeps from E2E test utilities

bartoszmajsak:
**What this PR does / why we need it**:

Remove hardcoded `asyncio.sleep(5)` and `time.sleep(10)` calls from `_predict`, `explain_response`, and `_openai_request`. The retry-based HTTP clients already handle transient failures properly.

**Release note**:
```release-note
NONE
```

Fixes #604 (hopefully)

PR #5357, opened 2026-04-08
fix(test): assert worker count via exec instead of log PIDs

jlost:
## Summary

- `test_sklearn_runtime_kserve` verified `--workers=2` by extracting PIDs from `kserve.trace` log lines after 25 concurrent requests. After #5071 removed blanket sleeps, httpx connection pooling can route all requests through a single TCP connection, so only one worker handles traffic and `assert len(process_ids) == 2` flakes -- particularly on Istio/OpenShift where Envoy manages upstream connections.
- Replace the traffic-dependent log assertion with a deterministic in-container process check via `kubectl exec`. A Python one-liner scans `/proc/*/cmdline` for multiprocessing `spawn_main` children. This is network-layer agnostic and works on slim Debian images that lack `pgrep`.
- Remove the now-unused `extract_process_ids_from_logs` utility and the post-predict `asyncio.sleep(5)` + log fetch.

## Test plan

- [ ] Verify `test_sklearn_runtime_kserve` passes in `test-predictor` (istio) CI job
- [ ] Verify `test_sklearn_runtime_kserve` passes in `test-kourier` CI job
- [ ] Confirm no regressions in other predictor E2E tests

### DIFF
--- a/test/e2e/common/utils.py
+++ b/test/e2e/common/utils.py
@@ -15,12 +15,14 @@
 import asyncio
 import json
 import os
+import time
 from concurrent import futures
 from typing import Union, List, Dict
 from urllib.parse import urlparse
 
 import portforward
 from kubernetes import client as k8s_client
+from kubernetes.stream import stream as k8s_stream
 from orjson import orjson
 
 from httpx import HTTPStatusError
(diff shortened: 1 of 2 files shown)


jlost:
This was observed in konflux but I haven't seen it reproduced since - let's close for now and reopen if it's seen again

PR #5363, opened 2026-04-08
fix: hardcode Status.Address scheme to http in raw deployment

jlost:

`createAddress` in `kube_ingress_reconciler.go` and `httproute_reconciler.go` used
`ingressConfig.UrlScheme` for the `Status.Address.URL.Scheme`. This conflated the
external URL scheme with the internal address scheme.

`Status.Address` is the cluster-internal endpoint (`*.svc.cluster.local`). The v1beta1
InferenceService raw deployment path has no pod-level TLS -- no cert injection, no TLS
sidecar, no `AppProtocol:"https"`. Pods always speak plain HTTP. `UrlScheme` controls
the external-facing `Status.URL` (gateway/ingress edge), not the in-cluster address.

Setting `urlScheme: https` (to reflect external TLS) would incorrectly report internal
addresses as HTTPS, even though the pod speaks plain HTTP.

This aligns the raw deployment reconcilers (Ingress and HTTPRoute) with:
- KServe's Knative-mode reconciler (`ingress_reconciler.go:116`), which hardcodes `http`
- Knative Serving's own route reconciler (`route.go:136`), which hardcodes `http`

If pod-level TLS is added to v1beta1 in the future, the address scheme should be derived
from a dedicated flag (analogous to LLMInferenceService's `EnableTLS`), not from
`UrlScheme`.

- [x] Unit tests for `createAddress` covering ClusterIP, headless, and transformer cases
- [x] Integration tests for `RawIngressReconciler.Reconcile` with `urlScheme: http` and `urlScheme: https`, verifying that `Status.Address` is always `http` while `Status.URL` respects the configured scheme
- [x] Cluster-local visibility test

**Checklist**:

```release-note
Fix Status.Address.URL.Scheme in raw deployment mode to always use "http" for the cluster-internal endpoint, instead of incorrectly inheriting the external URL scheme from ingressConfig.UrlScheme.
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
@@ -897,7 +897,7 @@ func (r *RawHTTPRouteReconciler) Reconcile(ctx context.Context, isvc *v1beta1.In
 	if isvc.Status.URL, err = createRawURL(isvc, r.ingressConfig); err != nil {
 		return ctrl.Result{}, err
 	}
-	isvc.Status.Address, err = createAddress(ctx, r.client, isvc, r.ingressConfig)
+	isvc.Status.Address, err = createAddress(ctx, r.client, isvc)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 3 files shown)


Jooho (approved):
/lgtm
/approved

PR #5366, opened 2026-04-09
feat(multinode): add groundwork for multinode support without ray

Jooho:
## Summary

- Add worker headless service (ClusterIP=None, PublishNotReadyAddresses=true) for multi-node deployments, enabling worker pod DNS discovery
- Add `multinode/executor-backend` annotation on ServingRuntime spec to switch between ray (default) and mp mode
- mp mode: PP determines node count, TP determines GPUs per node. Existing ray logic unchanged
- Inject PIPELINE_PARALLEL_SIZE, TENSOR_PARALLEL_SIZE, WORKER_SVC env vars into worker container
- Worker service name includes generation for rollout support (e.g. `foo-workers-1`)
- Add sample ClusterServingRuntime for vllm multi-node with `--distributed-executor-backend=mp`

Currently, the huggingfaceserver runtime supports multi-node deployments, but it relies on Ray as a required component for distributed execution. With recent vLLM updates, multi-node execution is now also possible without Ray by leveraging the mp (multiprocessing) distributed executor backend.

However, this approach requires passing specific vLLM CLI parameters (e.g., --master-addr, --master, --nnode) which are not supported by the existing huggingfaceserver runtime. For this reason, a dedicated ServingRuntime based directly on the vLLM image is required to fully enable this functionality.

Additionally, the node calculation logic differs between Ray and mp modes:

In Ray mode, node allocation is managed externally by Ray.
In mp mode, the number of nodes is derived from pipeline parallelism (PP), and GPUs per node are determined by tensor parallelism (TP).

To support this distinction, an annotation (multinode/executor-backend) has been introduced so that the controller can apply different logic depending on the execution backend.

A vLLM-based ServingRuntime is not included in this PR. However:

It has been tested locally and verified to work as expected.
The current huggingfaceserver runtime was also tested and confirmed not to work correctly with this approach.

A dedicated PR introducing the vLLM ServingRuntime will follow.

### Known Limitations
- Worker headless service name uses `isvcGeneration` as selector. If only the ServingRuntime 
  image changes (without ISVC spec change), generation won't increment and old/new worker pods 
  may share the same headless service during rolling updates. A follow-up PR will switch to 
  `pod-template-hash` based selector.

## Test plan

- [ ] Unit tests for `computeMpNodeAndGPUs` (PP/TP combinations)
- [ ] Unit tests for worker headless service creation in `createService`
- [ ] Existing ray-mode tests pass without changes
- [ ] E2E: deploy multi-node InferenceService with mp-mode ServingRuntime, verify worker service created and pods communicate

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -343,7 +343,10 @@ var (
 	IstioMeshGateway = "mesh"
 )
 
-const WorkerNodeSuffix = "worker"
+const (
+	WorkerNodeSuffix       = "worker"
+	WorkerNodeSuffixPlural = "workers"
+)
 
 // InferenceService Component enums
 const (
@@ -683,10 +686,18 @@ const (
 	DefaultPipelineParallelSize = 1
 )
 
+// MultiNode executor backend annotation
+// If not set, defaults to ray
+const (
+	MultiNodeExecutorBackendAnnotationKey = "multinode/executor-backend"
+	MultiNodeExecutorBackendMp            = "mp"
+)
+
 // Multi Node Labels
 var (
 	MultiNodeRoleLabelKey = "multinode/role"
 	MultiNodeHead         = "head"
+	MultiNodeWorker       = "worker"
 )
 
 // GetRawServiceLabel generate native service label
@@ -703,6 +714,12 @@ func GetRawWorkerServiceLabel(service string) string {
 func GetHeadServiceName(service string, isvcGeneration string) string {
 	isvcName := strings.TrimSuffix(service, "-predictor")
 	return isvcName + "-" + MultiNodeHead + "-" + isvcGeneration
+}
+
+// GetWorkerServiceName generate worker headless service name
+func GetWorkerServiceName(service string, isvcGeneration string) string {
+	isvcName := strings.TrimSuffix(service, "-predictor")
+	return isvcName + "-" + WorkerNodeSuffixPlural + "-" + isvcGeneration
 }
 
 func (e InferenceServiceComponent) String() string {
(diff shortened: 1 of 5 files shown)


brettmthompson (commented):
/lgtm

gavrissh:
>A dedicated PR introducing the vLLM ServingRuntime will follow.

There was an effort to do this, but it did not take off. Maybe we can consider this now - https://github.com/kserve/kserve/pull/4769

Jooho:
@gavrissh Hey good to see you :) 

I took a quick look at that PR, and it doesn’t seem to include the essential changes required to properly support multi-node / multi-GPU setups.

In contrast, my PR introduces controller-level changes in KServe to enable multi-node / multi-GPU support without relying on Ray.

For the vLLM runtime, I believe it should be handled separately from PR 4769. Similar to the HuggingFace server, a dedicated ServingRuntime specifically designed for multi-node setups is required, so it makes sense to proceed independently.

I’m not sure whether PR 4769 has been tested on an actual GPU cluster, but based on my understanding, using the default vLLM ServingRuntime alone would lead to several issues:

Pods (head/worker) would not restart properly upon failure
Node rank assignment is manual and not managed by KServe
There is no mechanism to handle these orchestration concerns

Because of these gaps, I expect that it may not even start correctly in a real multi-node environment.

Therefore, for multi-node / multi-GPU support, I believe this PR is the more appropriate approach.

That said, in the long term, we should align toward using LLMISvc for multi-node / multi-GPU workloads. While it requires LWS, it enables more advanced capabilities and better flexibility.

terrytangyuan (approved):

mwaykole:
this is nice

PR #5368, opened 2026-04-10
fix: recover multi download flag in default CSC

israel-hdez:

In #4702, a `supportsMultiModelDownload: true` flag was introduced and Helm charts where updated, although an update to Kustomize manifests is missing.

Later, in a manifest refactor, such default of `true` was lost. This recovers the right default for the provided storage-initializer.

- Run `kubectl apply -k config/overlays/standalone/kserve` as mentioned in [v0.17 install docs](https://kserve.github.io/website/docs/install/kserve-install#method-1-kustomize) and check that the right configuration of the ClusterStorageContainer resource is installed in the cluster.
- Create an ISVC with multiple storageURIs, and check that it deploys correctly (no message in logs about the storage-init not supporting multi downloads).

**Checklist**:

- [N/A] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/kserve-llmisvc-resources/files/common/storagecontainer.yaml
+++ b/charts/kserve-llmisvc-resources/files/common/storagecontainer.yaml
@@ -23,4 +23,5 @@ spec:
   - regex: https://(.+?).blob.core.windows.net/(.+)
   - regex: https://(.+?).file.core.windows.net/(.+)
   - regex: https?://(.+)/(.+)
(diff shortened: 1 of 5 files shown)


spolti (approved):
/lgtm

Jooho (approved):
@israel-hdez good catch! 
/lgtm
/approved

PR #5236, opened 2026-03-12
fix: fix nil pointer errors in LLMInferenceServiceConfig templates (#5231)

SeanCondon:
Changes:
- `config-llm-decode-worker-data-parallel.yaml`: Add nil checks for .Spec.Parallelism.{Expert,Tensor,Data,DataLocal,DataRPCPort}
- `config-llm-prefill-worker-data-parallel.yaml`: Add nil checks for .Spec.Prefill.Parallelism.{Expert,Tensor,Data,DataLocal,DataRPCPort}
- `config-llm-worker-data-parallel.yaml`: Add nil checks for .Spec.Parallelism.{Expert,Tensor,Data,DataLocal,DataRPCPort}
- `config_merge_test.go`: Enhanced unit tests to verify all nil-safety cases for both Parallelism and Prefill.Parallelism

**What this PR does / why we need it**:
Add comprehensive nil-safety checks to all template expressions accessing `Parallelism` and `Prefill.Parallelism` fields in well-known configs. This prevents nil pointer errors when `LLMInferenceService` specs don't define parallelism configurations.

**Which issue(s) this PR fixes**:
Fixes #5231 

**Feature/Issue validation/testing**:

- [ ] Added new Unit tests in [config_merge_test.go](pkg/controller/v1alpha2/llmisvc/config_merge_test.go)
- [ ] Ran e2e tests `uv run --project python/kserve pytest test/e2e/llmisvc -m "llminferenceservice and cluster_nvidia" -v`

- Logs

**Special notes for your reviewer**:
Thanks for reviewing

**Checklist**:

- [X] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [X] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note
NONE
```

PR #5371, opened 2026-04-12
fix: nil pointer panic in baseRefs merge when Parallelism set via baseRef

### LINKED ISSUE #4945: LLMInferenceService baseRefs merge fails with "nil pointer evaluating *v1alpha1.ParallelismSpec.Expert"
When using baseRefs to combine multiple LLMInferenceServiceConfig resources, the merge fails if any config contains a parallelism field.

  Steps to Reproduce:

  1. Create an LLMInferenceServiceConfig with parallelism:
  metadata:
    name: workload-dp2
    namespace: kserve-test
  spec:
    parallelism:
      tensor: 1
      data: 2
      dataLocal: 1
    # ... template and worker specs

  2. Create another LLMInferenceServiceConfig without parallelism:
  metadata:
    name: model-config
    namespace: kserve-test
  spec:
    model:
      uri: pvc://models/model-name
      name: model-name

  3. Create an LLMInferenceService using baseRefs:
  kind: LLMInferenceService
  metadata:
    name: test-service
    namespace: kserve-test
  spec:
    baseRefs:
      - name: model-config
      - name: workload-dp2
    replicas: 1

  Expected Behavior:
  The configs should merge successfully.

  Actual Behavior:
  Reconciliation fails with error:
  failed to merge config: template: config:1:8595: executing "config" at <.Spec.Parallelism.Expert>: nil pointer evaluating *v1alpha1.ParallelismSpec.Expert

  Root Cause:
  The merge template accesses .Spec.Parallelism.Expert without checking if Parallelism is nil first.

  Workaround:
  Define all configuration inline in LLMInferenceService instead of using baseRefs.

nickaggarwal:

Fixes a nil pointer panic in `LLMInferenceService` reconciliation when `Parallelism` fields (Expert, Tensor, Data, DataLocal, DataRPCPort) are configured via `baseRef` configs instead of directly on the service spec.

When `ReplaceVariables` renders config templates, it receives the **original** `LLMInferenceService` which may have `Parallelism: nil` if parallelism comes from a `baseRef`. Templates then panic trying to access nil pointers.

**Root cause**: `combineBaseRefsConfig` merges all specs and builds the merged config, but passes the original `llmSvc` to `ReplaceVariables` for template rendering. Templates access fields that are nil in the original service.

**Fix**: Pass the **merged spec** to `ReplaceVariables` instead of the original service. This ensures template data includes all merged configuration from baseRefs.

Fixes #4945

- [ ] Create an `LLMInferenceServiceConfig` with `parallelism: {data: 2, dataLocal: 1, expert: true}`
- [ ] Create an `LLMInferenceService` that references it via `baseRefs` but has **no** `parallelism` field in its own spec
- [ ] Verify reconciliation succeeds without nil pointer panic
- [ ] Verify the generated vLLM command includes `--enable-expert-parallel` and correct `--data-parallel-size` values

This is a minimal architectural fix (5 lines of code in one place). Instead of adding nil guards to multiple YAML templates, we fix the root cause by ensuring template data uses the merged spec. This keeps templates clean and maintainable.

The approach:
1. `spec` variable already contains the merged configuration from all baseRefs
2. Create a temporary service copy with this merged spec: `tmpSvc.Spec = spec`
3. Pass the temporary service to `ReplaceVariables` for template rendering
4. Templates now have access to complete Parallelism configuration

**Checklist**:

  - Existing tests for `ReplaceVariables` will verify the fix. The function signature is unchanged; we just provide it with better data.
  - Yes, added clear comments explaining why we create a temporary service with merged spec.
  - No documentation changes needed for this internal reconciler fix.

**Release note**:
```release-note
Fix nil pointer panic in LLMInferenceService reconciler when ParallelismSpec fields (Expert, Tensor, Data, DataLocal, DataRPCPort) are configured via baseRefs instead of directly on the service spec.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_merge.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_merge.go
@@ -250,7 +250,11 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
 		llmSvcCfg.Spec.Router.Scheduler.Template.ServiceAccountName = kmeta.ChildName(llmSvc.GetName(), "-epp-sa")
 	}
 
-	llmSvcCfg, err = ReplaceVariables(llmSvc, llmSvcCfg, reconcilerConfig)
+	// Create a temporary service with the merged spec for template rendering.
+	// This ensures Parallelism and other fields from baseRefs are available to templates.
+	tmpSvc := llmSvc.DeepCopy()
+	tmpSvc.Spec = spec
+	llmSvcCfg, err = ReplaceVariables(tmpSvc, llmSvcCfg, reconcilerConfig)
 	if err != nil {
 		return llmSvcCfg, err
 	}


spolti (commented):
cc @pierDipi @vivekk16

bartoszmajsak:
I wonder if this is not fixing the issue? https://github.com/kserve/kserve/pull/5236

PR #5374, opened 2026-04-13
feat(llmisvc): add heterogeneous GPUs load balancing sample

pierDipi:
Deploy the same model across NVIDIA and AMD GPUs using two LLMInferenceService instances that share a single scheduler/EPP via a custom InferencePool selector.

Create one scheduler/EPP, n1-replicas of vllm using NVIDIA GPUs and n2-replicas of vllm using AMD GPUs in oder to utilise all cluster resources.

**Release note**:

```release-note
feat(llmisvc): add heterogeneous GPUs load balancing sample
```

### DIFF
--- a/docs/samples/llmisvc/multi-gpu-vendor/README.md
+++ b/docs/samples/llmisvc/multi-gpu-vendor/README.md
@@ -1,0 +1,141 @@
+# Multi-GPU-Vendor Deployment Example
+
+Deploy the same model across NVIDIA and AMD GPUs using two LLMInferenceService instances that share a single
+scheduler/EPP via a custom InferencePool selector.
+
+## Problem
+
+A cluster has both NVIDIA and AMD GPU nodes. A single LLMInferenceService can only target one GPU type because all
+replicas share the same pod template. Without this pattern, one GPU vendor's capacity sits idle.
+
+## Solution
+
+1. **NVIDIA instance** (`qwen2-7b-instruct-nvidia`) — creates the scheduler/EPP, InferencePool, HTTPRoute, and Gateway.
+   The InferencePool uses a custom selector (`llm-pool: qwen2-7b`) instead of the default name-based selector.
+2. **AMD instance** (`qwen2-7b-instruct-amd`) — has **no router** (no scheduler, no route, no gateway). Its pods carry
+   the same `llm-pool: qwen2-7b` label, so the EPP from the NVIDIA instance discovers and routes traffic to them.
+
+## Architecture
+
+```
+                         ┌──────────────┐
+                         │   Gateway    │
+                         └──────┬───────┘
+                                │
+                         ┌──────┴───────┐
+                         │  HTTPRoute   │
+                         └──────┬───────┘
+                                │
+                    ┌───────────┴───────────┐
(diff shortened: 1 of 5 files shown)


bartoszmajsak (approved):
On the verge here :) Smart "label collision" trick that makes it work, but might be confusing for an average user (and also maybe when surfaced through UI). README maybe should hint it's something we can improve moving forward, but assuming it is a sample for "advanced use cases" this is good to merge.

I wonder if we could enhance API so we could rely on custom labels?

```
# NVIDIA - custom pool selector 
  spec:
    router:
      scheduler:
        pool:
          spec:
            selector:
              matchLabels:
                llm-pool: qwen2-7b

  # AMD - explicit opt-in
  spec:
    labels:
      llm-pool: qwen2-7b
```

If implicit label is not good enough maybe `spec.inferencePoolRef` or something like that can be a good candidate?

bartoszmajsak:
/lgtm
/approve

Just make builds happy :)

vpatz:
@pierDipi  sharing a version with minor modifications that worked for me on a cluster with different GPU nodes.
But had to manually label pods.

```
# For pod on NVIDIA and scheduler
kind: LLMInferenceService
metadata:
  name: qwen2-7b-instruct-nvidia
spec:
  model:
    name: Qwen/Qwen2.5-7B-Instruct
  replicas: 1
  labels:
    llm-pool: qwen2-7b
  router:
    scheduler:
      pool:
        spec:
          # v1alpha2: selector is a flat map (label key -> string value), not matchLabels.
          selector:
            llm-pool: qwen2-7b
            kserve.io/component: workload
          targetPortNumber: 8000
          extensionRef:
            name: qwen2-7b-instruct-nvidia-epp-service
            kind: Service
            portNumber: 9002
    route: { }
    gateway: { }
  template:
    containers:
      - name: main
        resources:
          limits:
            cpu: '4'
            memory: 32Gi
            nvidia.com/gpu: "1"
          requests:
            cpu: '2'
            memory: 16Gi
            nvidia.com/gpu: "1" 
```


```
# For pod on AMD
kind: LLMInferenceService
metadata:
  name: qwen2-7b-instruct-amd
spec:
  model:
    name: Qwen/Qwen2.5-7B-Instruct
  replicas: 1
  # Same shared label as the NVIDIA instance so that the InferencePool
  # selector (llm-pool: qwen2-7b) matches these pods too.
  labels:
    llm-pool: qwen2-7b
  # No router — no scheduler, no route, no gateway.
  # Traffic reaches these pods through the shared InferencePool
  # managed by the NVIDIA LLMInferenceService's scheduler/EPP.
  template:
    containers:
      - name: main
        image: ghcr.io/llm-d/llm-d-rocm:v0.6.0
        resources:
          limits:
            cpu: '4'
            memory: 32Gi
            amd.com/gpu: "1"
          requests:
            cpu: '2'
            memory: 16Gi
            amd.com/gpu: "1"

```
Set label to pods

```
oc label pod qwen2-7b-instruct-nvidia-kserve-<...> llm-pool=qwen2-7b
oc label pod qwen2-7b-instruct-amd-kserve-<...> llm-pool=qwen2-7b

```

PR #5378, opened 2026-04-13
ci: add structured E2E test results (JUnit XML + JSON)

jlost:
## Summary

- Parametrize PYTEST_ARGS in `run-e2e-tests.sh` 
- Add `--junitxml` and `--json-report` flags to the E2E pytest invocation so every CI run produces structured per-test results
- Add `pytest-json-report` to the test dependency group
- Add `upload-artifact` steps to all E2E jobs in GitHub Actions

## Motivation

The E2E test runner currently produces only console output with no structured test results. This means:
- No way to distinguish infrastructure failures from test failures at the per-test level
- No way to identify flaky tests or track trends over time

## What this produces

Two output files per E2E run, written to `$ARTIFACT_DIR` (or `/tmp` if unset):

| File | Format | Consumer |
|------|--------|----------|
| `junit_e2e.xml` | JUnit XML (built-in pytest) | Prow Spyglass JUnit lens, Sippy, TestGrid |
| `e2e_results.json` | JSON (pytest-json-report) | Custom analysis tooling (richer: per-test timings, setup/call/teardown stages, tracebacks) |

Both files are written even when tests fail, enabling partial failure analysis.

## Test plan

- [x] GitHub Actions: Verify `junit_e2e.xml` and `e2e_results.json` appear as uploaded artifacts

### DIFF
--- a/.github/workflows/e2e-test-llmisvc.yaml
+++ b/.github/workflows/e2e-test-llmisvc.yaml
@@ -237,6 +237,16 @@ jobs:
         run: |
           ./test/scripts/gh-actions/status-check.sh "llmisvc"
 
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: test-results-llmisvc-${{ matrix.install-method }}
+          path: |
+            /tmp/junit_e2e.xml
+            /tmp/e2e_results.json
+          if-no-files-found: ignore
+
   test-llmisvc-wva-autoscaling-hpa:
     runs-on: ubuntu-latest
     needs: [detect-changes, llmisvc-image-build]
@@ -329,6 +339,16 @@ jobs:
         if: always()
         run: |
           ./test/scripts/gh-actions/status-check.sh "llmisvc"
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: test-results-llmisvc-autoscaling-hpa-${{ matrix.install-method }}
+          path: |
+            /tmp/junit_e2e.xml
+            /tmp/e2e_results.json
+          if-no-files-found: ignore
 
   test-llmisvc-wva-autoscaling-keda:
     runs-on: ubuntu-latest
@@ -423,3 +443,13 @@ jobs:
         if: always()
         run: |
           ./test/scripts/gh-actions/status-check.sh "llmisvc"
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: test-results-llmisvc-autoscaling-keda-${{ matrix.install-method }}
+          path: |
+            /tmp/junit_e2e.xml
+            /tmp/e2e_results.json
+          if-no-files-found: ignore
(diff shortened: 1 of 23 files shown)


jlost:
Example output: 
https://github.com/kserve/kserve/actions/runs/25402625167/artifacts/6817926388
https://github.com/kserve/kserve/actions/runs/25402625167/artifacts/6818529507

spolti:
shouldn't we have a single file for the tests outputs? Like e2e test-predictor?

jlost:
@spolti 
Each e2e test run produces 2 files: 
- e2e_results.json
- junit_e2e.xml

They get zipped together and added as an artifact to the GHA Run (e.g. [E2E Tests / test-predictor (kustomize) (pull_request)](https://github.com/kserve/kserve/actions/runs/25402625167/job/74511528459?pr=5378)), which you can find through [the run details](https://github.com/kserve/kserve/actions/runs/25402625167/job/74511528459?pr=5378) under 'Upload test results'

The .json file is more detailed and captures runtime for different test phases. The .xml file is a standard junit format that is preferred by many observability tools.

So there is a zip containing both of those files for each: e2e-predictor, e2e-graph, e2e-raw, etc.

Not sure if that answers your question?

spolti:
it was, thanks.

jlost:
Added a docs section briefly explaining the artifacts and where to find them: https://github.com/kserve/website/pull/673

jlost:
@spolti @Jooho What do you think, is it ready to be merged?

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -236,6 +236,16 @@ jobs:
...
 
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: test-results-llmisvc-${{ matrix.install-method }}
+          path: |
+            /tmp/junit_e2e.xml
+            /tmp/e2e_results.json
+          if-no-files-found: ignore
```
bartoszmajsak: Can we make it a re-usable step with some parameters (`name` at least ;))? WDYT?

It's already established pattern, example being https://github.com/kserve/kserve/blob/b01b38f1cc11d917194e2b096bdb5984e5b2d41d/.github/actions/minikube-setup/action.yml

bartoszmajsak (commented):
One minor comment

.github/workflows/e2e-test-llmisvc.yaml
jlost: Good idea. Let me take a look.

jlost: Fixed

python/kserve/pyproject.toml
```diff
@@ -83,6 +83,7 @@ test = [
     "pytest<8.0.0,>=7.4.4",
     "pytest-cov<6.0.0,>=5.0.0",
     "pytest-xdist<4.0.0,>=3.0.2",
+    "pytest-json-report<2.0.0,>=1.5.0",
```
spolti, line 86: may we add it in the storage module as well?

jlost: Good idea, I added it for all of the tests in Python package / build (3.xx).

spolti: Did you add the dependencies in the storage module? Or is it not needed?
PS: the storage module is independent, it does not depend anymore on KServe's `pyproject.toml`.

jlost: In CI, python-test.yml activates the kserve/.venv to run pytest, so pytest-json-report is derived from there. In other words, the storage pyproject.toml doesn't have its own test dependencies section, and CI reuses the main one.

spolti (approved):

PR #5382, opened 2026-04-14
feat: add support for confidential model serving

### LINKED ISSUE #5080: Standards-Based Confidential Model Serving
/kind feature

This proposal introduces a hardware-rooted mechanism for protecting machine learning models at rest and in use. By extending the **KServe Storage Initializer** and leveraging **Modelcars**, we enable the decryption of models only within a verified **Trusted Execution Environment (TEE)**. The design relies on **OCI** `ocicrypt` and **JSON Web Encryption (JWE)** to remain agnostic of specific hardware vendors or Key Broker Services (KBS).

https://docs.google.com/document/d/1wrFE0flCEXUfW8YPwavNTjyc12DvnqBIVRQIbpF5e-U/edit?usp=sharing

font:
## Summary

Adds confidential model serving support to KServe, enabling encrypted ML model artifacts to be decrypted only inside Trusted Execution Environments (TEEs) using hardware attestation.

### What this PR does

**Go (API, controller, webhook):**
- Adds `ConfidentialSpec` (`enabled`, `resourceId`) to `PredictorExtensionSpec` (v1beta1) and `LLMModelSpec` (v1alpha2)
- Validation: requires `storageUri` when enabled, warns on OCI URIs, validates `kbs:///<repo>/<type>/<tag>` resource ID format
- Controller propagates confidential annotations to pods
- Webhook mutator injects `CONFIDENTIAL_ENABLED` + `CONFIDENTIAL_RESOURCE_ID` env vars into the storage initializer init container
- LLMISVC controller applies the same env var injection directly

**Python (confidential storage initializer):**
- `SecretResolver` abstract interface with `CDHSecretResolver` implementation (Confidential Data Hub, local to TEE guest)
- `JWEDecryptor` for detecting and decrypting JWE-encrypted model files (RFC 7516, A256KW + A256GCM)
- Confidential decryption logic integrated into the unified storage initializer image (no separate image required)
- RATS-protocol agnostic: works with any KBS backend (Trustee, Intel Trust Authority, etc.)

### How it works

1. User sets `confidential: enabled: true` on their InferenceService or LLMInferenceService
2. KServe injects confidential env vars into the storage initializer init container
3. The confidential storage initializer downloads model artifacts (from S3, HF, GCS, etc.)
4. It detects JWE-encrypted files (by `.jwe` extension or JWE header)
5. It contacts the Confidential Data Hub (CDH) running locally inside the TEE guest
6. CDH performs hardware attestation with a Key Broker Service (KBS) and retrieves the decryption key
7. The model is decrypted in-place inside the TEE, never exposed outside

### Example

```yaml
kind: InferenceService
metadata:
  name: my-confidential-model
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      storageUri: "s3://bucket/encrypted-model"
      confidential:
        enabled: true
        resourceId: "kbs:///default/keys/model-key"
```

## Test plan

- [x] Go unit tests for validation (confidential enabled/disabled, OCI URI warning, resource ID format)
- [x] Go unit tests for webhook (env var injection)
- [x] Go unit tests for LLMISVC controller (confidential config application)
- [x] Go integration tests pass (111 specs for InferenceService controller)
- [x] Python unit tests for `SecretResolver`, `CDHSecretResolver`, `JWEDecryptor` (16 tests)
- [x] [E2E tested on ARO cluster with CoCo peer-pods](https://github.com/font/kserve/tree/coco-e2e-tests/hack/confidential-test) (SEV-SNP TEE):
  - Unencrypted OCI modelcar on CoCo peer-pod ([Known issue with encrypted OCI images using CRI-O](https://github.com/font/kserve/tree/coco-e2e-tests/hack/confidential-test#encrypted-oci-images-not-supported-on-openshift))
  - JWE-encrypted sklearn model in MinIO (S3) → CDH → KBS → decrypted inside TEE → inference verified
  - LLMInferenceService confidential plumbing (env var injection)

## Related

- Design [KEP: Standards-Based Confidential Model Serving](https://docs.google.com/document/d/1wrFE0flCEXUfW8YPwavNTjyc12DvnqBIVRQIbpF5e-U/edit?usp=sharing)
- CoCo (Confidential Containers): https://confidentialcontainers.org

Fixes #5080


python/storage-initializer-confidential.Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+RUN uv venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# Install Python dependencies (including confidential extras)
+COPY storage/pyproject.toml storage/uv.lock storage/
+RUN cd storage && uv sync --active --no-cache
+
+COPY storage storage
+RUN cd storage && uv pip install ".[confidential]" --no-cache
+
```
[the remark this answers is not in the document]
font: Fixed — changed to `uv sync --active --extra confidential`.

hack/patch-confidential-image.sh
```diff
@@ -0,0 +1,23 @@
...
+
+kubectl get configmap inferenceservice-config -n "${NAMESPACE}" -o json | \
+  python3 -c "
+import json, sys
+cm = json.load(sys.stdin)
+si = json.loads(cm['data']['storageInitializer'])
+si['confidentialImage'] = '${IMAGE}'
+cm['data']['storageInitializer'] = json.dumps(si)
+json.dump(cm, sys.stdout)
+" | kubectl apply -f -
```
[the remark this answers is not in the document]
font: Fixed — now passes image via `sys.argv[1]`.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -534,11 +534,39 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
...
+		Name:  constants.ConfidentialEnabledEnvVar,
+		Value: "true",
+	})
+
+	if resourceId, ok := annotations[constants.ConfidentialResourceIdAnnotationKey]; ok && resourceId != "" {
+		initContainer.Env = append(initContainer.Env, corev1.EnvVar{
+			Name:  constants.ConfidentialResourceIdEnvVar,
+			Value: resourceId,
+		})
+	}
```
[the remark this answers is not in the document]
font: Fixed — switched to `utils.AddOrReplaceEnv()` for idempotent env var setting.

pkg/types/config.go
```diff
@@ -28,4 +28,6 @@ type StorageInitializerConfig struct {
 	MemoryModelcar          string `json:"memoryModelcar"`
 	EnableOciImageSource    bool   `json:"enableModelcar"`
 	UidModelcar             *int64 `json:"uidModelcar"`
+	ConfidentialImage              string `json:"confidentialImage,omitempty"`
+	EnableConfidentialAutoDiscovery bool   `json:"enableConfidentialAutoDiscovery,omitempty"`
```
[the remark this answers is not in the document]
font: Fixed — ran `gofmt`.

python/storage-initializer/scripts/initializer-entrypoint-confidential
```diff
@@ -0,0 +1,55 @@
...
+
+        total_decrypted = 0
+        for dest_path in dest_paths:
+            decrypted = decryptor.decrypt_directory(dest_path, resource_id=resource_id or None)
+            total_decrypted += len(decrypted)
+
+        logger.info("Confidential decryption complete: %d files decrypted", total_decrypted)
+    except Exception as e:
+        logger.error("Confidential decryption failed: %s", e)
+        sys.exit(1)
```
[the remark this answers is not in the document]
font: Fixed — `CONFIDENTIAL_RESOURCE_ID` is now required when confidential mode is enabled (fails early with error). Auto-discovery references have been removed from docs and the `enableConfidentialAutoDiscovery` config field has been removed.

python/storage/kserve_storage/confidential/jwe_decryptor.py
```diff
@@ -0,0 +1,152 @@
...
+        jwe_token.decrypt(symmetric_key)
+        plaintext = jwe_token.payload
+
+        # Determine output path (strip .jwe extension if present)
+        if path.suffix.lower() == _JWE_EXTENSION:
+            output_path = path.with_suffix("")
+        else:
+            output_path = path
+
+        output_path.write_bytes(plaintext)
```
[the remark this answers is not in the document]
font: Acknowledged — added a NOTE comment about this limitation. JWE Compact Serialization requires the full ciphertext in memory; streaming isn't possible with this format. For very large models, users should split files before encryption or increase container memory limits.

python/storage/kserve_storage/confidential/jwe_decryptor.py
```diff
@@ -0,0 +1,152 @@
...
+    def __init__(self, secret_resolver: SecretResolver, resource_id: str | None = None):
+        """
+        Args:
+            secret_resolver: Resolver used to obtain decryption keys.
+            resource_id: Default KBS resource ID for key resolution. If not
+                provided, must be discoverable per-file or set via environment.
+        """
+        self._resolver = secret_resolver
+        self._resource_id = resource_id
+
```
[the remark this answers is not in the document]
font: Fixed — updated docstring to match actual behavior (resource_id must be provided either at construction or per-call).

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -40,6 +41,9 @@ import (
 // variantCostPattern is compiled once at package init to avoid recompilation on every webhook call.
 var variantCostPattern = regexp.MustCompile(`^\d+(\.\d+)?$`)
 
+// kbsResourceIdRegexp validates KBS resource ID format: kbs:///<repo>/<type>/<tag>
+var kbsResourceIdRegexp = regexp.MustCompile(`^kbs:///[a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+/[a-zA-Z0-9_.-]+$`)
```
Jooho: The regex only validates the 3-segment path (kbs:///repo/type/tag) and would reject KBS URIs with query params, even though the spec appears to allow them. It may be worth allowing an optional query suffix, or documenting that params are intentionally unsupported.

Also, other protocols like hf:// or s3:// do not seem to have this level of webhook validation — is there a reason kbs:// needs stricter treatment, or is blocking KBS params intentional?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -40,6 +40,9 @@ import (
 	"github.com/kserve/kserve/pkg/validation"
 )
 
+// KBS resource ID format: kbs:///<repo>/<type>/<tag>
+var kbsResourceIdRegexp = regexp.MustCompile(`^kbs:///[a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+/[a-zA-Z0-9_.-]+$`)
```
Jooho: ConfidentialSpec type being defined separately in v1alpha2 and v1beta1 is fine — that follows K8s API versioning conventions.                                                                                  
                                                                                                                                                                                                                 
  But kbsResourceIdRegexp and the core validateConfidential logic (OCI URI warning, resourceId format check) are duplicated across both versions. These validate a protocol-level constraint (KBS resource ID
  format), not something version-specific.                                                                                                                                                                       
                                                                  
  Worth extracting the regex and shared validation into a common package (e.g. pkg/utils/ or pkg/constants/) so both webhooks import from one place. Otherwise any future change to KBS validation (e.g.         
  supporting query parameters) needs to be applied in two places independently.
                                                                                                                                                                                                                 
  Since this PR introduces both copies, cleaner to consolidate now rather than a follow-up refactor.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -408,3 +419,32 @@ func needCaBundleMount(caBundleConfigMapName string, initContainer *corev1.Conta
...
+	}
+
+	// Find the storage-initializer init container
+	for i := range podSpec.InitContainers {
+		if podSpec.InitContainers[i].Name != constants.StorageInitializerContainerName {
+			continue
+		}
+
+		// Swap image to confidential image if configured
+		if storageConfig.ConfidentialImage != "" {
```
Jooho: My understanding of which storage backends work with confidential model serving:

  Works — any protocol that goes through storage-initializer (s3://, hf://, gs://, Azure Blob, http(s)://). The init container image gets swapped to the confidential one, which handles JWE decryption.

  Doesn't work:
  - oci:// — uses modelcar, no storage-initializer to swap. Decryption relies on container runtime (kata-cc + image-rs). Known issue with CRI-O for encrypted OCI images.
  - pvc:// — direct volume mount, no storage-initializer. Setting confidential: enabled is silently ignored with no webhook warning.

  Does this match the intended scope? If so, worth documenting this in the website docs.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -408,3 +419,32 @@ func needCaBundleMount(caBundleConfigMapName string, initContainer *corev1.Conta
 	}
 	return result
 }
+
+// applyLLMConfidentialConfig swaps the storage-initializer init container image and injects
+// environment variables for confidential model serving when enabled in the LLMInferenceService spec.
+func applyLLMConfidentialConfig(llmSvc *v1alpha2.LLMInferenceService, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig) {
```
Jooho: Both ISVC and LLMISVC end up doing the same thing — find storage-initializer init container, swap image, inject env vars — but through different paths. ISVC goes through annotations (controller → webhook), 
  LLMISVC does it directly in the controller. That architectural difference makes sense, but the core logic is duplicated.                                                                                     
                                                                                                                                                                                                                 
  Would it make sense to extract the shared part into something like `pkg/utils/confidential.go`?

This is duplicated with here:  pkg/webhook/admission/pod/storage_initializer_injector.go `applyConfidentialConfig`

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -408,3 +419,32 @@ func needCaBundleMount(caBundleConfigMapName string, initContainer *corev1.Conta
...
+	confidential := llmSvc.Spec.Model.Confidential
```
Jooho: `applyLLMConfidentialConfig` has no unit tests. The webhook side (applyConfidentialConfig in storage_initializer_injector.go) has 4 test scenarios covering the same logic, but since these are two separate functions, both need coverage.

  This ties back to the earlier duplication point — if the core logic (image swap + env var injection) lived in a shared utility function, you'd only need one set of tests and both paths would be covered.  Right now the LLMISVC controller path is untested.

python/storage/kserve_storage/confidential/jwe_decryptor.py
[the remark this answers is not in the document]
Jooho: If the full ciphertext has to be in memory, the default memoryLimit: 1Gi for the storage initializer could be tight. JWE compact serialization + base64 overhead means roughly 2.3x the original file size in memory, so anything over ~400MB encrypted would risk OOM.
                                                                  
  Currently there's no way to set resources specifically for the confidential init container — StorageInitializerConfig only has confidentialImage, not separate memory/cpu fields. Bumping memoryLimit in the ConfigMap affects all storage initializers globally, including non-confidential ones.
   
  Might be worth adding dedicated resource fields (e.g. confidentialMemoryLimit) so applyConfidentialConfig can adjust resources when swapping to the confidential image. Or at minimum documenting that users should bump the global storageInitializer.memoryLimit when using confidential mode.

python/storage/kserve_storage/confidential/cdh_client.py
```diff
@@ -0,0 +1,85 @@
...
+import re
+
+import requests
+
+from .secret_resolver import SecretResolutionError, SecretResolver
+
+logger = logging.getLogger(__name__)
+
+# kbs:///<repo>/<type>/<tag>
+_KBS_RESOURCE_ID_RE = re.compile(
```
Jooho, line 26: Minor: the KBS resource ID regex differs between Go and Python.

  Go webhook uses `[a-zA-Z0-9_-]` for each segment, while cdh_client.py uses `[^/]+` which accepts any character except slash.

python/storage-initializer/scripts/initializer-entrypoint-confidential
```diff
@@ -0,0 +1,58 @@
...
+        # Select the appropriate SecretResolver based on available configuration.
+        # CDHSecretResolver talks to the Confidential Data Hub (CDH) running
+        # inside the TEE guest.  CDH handles attestation and key retrieval
+        # regardless of the underlying KBS/RATS implementation.
+        # Future alternatives for non-CoCo environments:
+        # - File-based resolver: read key from a mounted Kubernetes Secret
+        # - Environment variable resolver: key passed directly via env var
+        # - Cloud KMS resolver: retrieve key from a cloud provider KMS
+        resolver = CDHSecretResolver()
+        decryptor = JWEDecryptor(resolver, resource_id=resource_id or None)
```
Jooho: Nit: `resource_id or None` on lines 48 and 52 is dead code. Line 33 already exits if resource_id is empty, so by the time you reach lines 48/52, resource_id is always a non-empty string and or None never triggers.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
font: Good summary — that's exactly right. The confidential decryption path only applies to blob storage backends (S3, GCS, Azure Blob) where the storage initializer downloads files. OCI/modelcar images bypass the storage initializer entirely, so confidential decryption doesn't apply there (the webhook emits a warning if you try). I've added a comment to the `ValidateConfidentialSpec` function documenting this.

python/storage/kserve_storage/confidential/jwe_decryptor.py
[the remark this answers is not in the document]
font: Good point about memory. JWE requires the full ciphertext in memory for authenticated decryption (AEAD), so large model files can hit the default 1Gi memoryLimit. The overhead is roughly 2.3x the original file size (base64-encoded ciphertext + decoded bytes + decrypted plaintext).

Adding a `confidentialMemoryLimit` config field to let users tune this is a good idea. I'll track it as a follow-up since it's an optimization rather than a correctness issue — users can work around it today by setting the storage initializer memory limit via the existing `storageInitializer.memoryLimit` config. For very large models the typical pattern would be to encrypt individual files (e.g., per-shard) rather than a single monolithic archive.

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
font: Intentionally not supported. Query parameters are not part of the standard KBS resource addressing scheme defined by the Confidential Containers project — the three-segment path `kbs:///<repo>/<type>/<tag>` is the canonical format. Added a comment to the regex documenting this rationale.

Also extracted the regex and validation logic into a shared `pkg/validation/confidential.go` package so both v1beta1 and v1alpha2 webhooks use the same regex and validation.

pkg/apis/serving/v1beta1/inference_service_validation.go
font: Done. Extracted shared validation into `pkg/validation/confidential.go` with `ValidateConfidentialSpec()` and `KBSResourceIdRegexp`. Both v1beta1 and v1alpha2 webhooks now delegate to it.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
font: Done. Extracted the shared container configuration logic into `pkg/utils/confidential.go` with `ApplyConfidentialContainerConfig()`. Both the webhook's `applyConfidentialConfig` and the LLMISVC controller's `applyLLMConfidentialConfig` now delegate to it.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
font: Done. Added unit tests for `applyLLMConfidentialConfig` in `workload_storage_confidential_test.go` covering nil spec, disabled, enabled with/without resourceId, and empty confidentialImage cases. These also exercise the shared `ApplyConfidentialContainerConfig` utility transitively.

python/storage/kserve_storage/confidential/cdh_client.py
font: Done. Aligned the Python regex to match the Go regex: changed `[^/]+` to `[a-zA-Z0-9_.-]+` in `cdh_client.py`.

python/storage-initializer/scripts/initializer-entrypoint-confidential
font: Done. Removed the dead `or None` on both lines — just `resource_id` now.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -76,26 +76,37 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
...
+		return fmt.Errorf("unsupported schema in model URI: %s", modelUri)
+	}
+
+	if err != nil {
+		return err
 	}
 
-	return fmt.Errorf("unsupported schema in model URI: %s", modelUri)
+	// Apply confidential model serving configuration if enabled
+	applyLLMConfidentialConfig(llmSvc, podSpec, config.StorageConfig)
```
pierDipi: why is this logic here and not in attachStorageInitializer?

pierDipi: there is logic to avoid restarts on Kserve upgrades (ie `copied ...`) in that function that we need for this too

config/configmap/inferenceservice.yaml
```diff
@@ -649,7 +649,8 @@ data:
         "enableModelcar": true,
         "cpuModelcar": "10m",
         "memoryModelcar": "15Mi",
-        "uidModelcar": 1010
+        "uidModelcar": 1010,
+        "confidentialImage": ""
```
bartoszmajsak: nit: if we have `omitempty` in the struct do we really need this here?

pierDipi: should we set a value and have a GH action build + publish the image?

pkg/validation/confidential.go
```diff
@@ -0,0 +1,60 @@
...
+func ValidateConfidentialSpec(enabled bool, resourceId *string, uri string, basePath *field.Path) (admission.Warnings, field.ErrorList) {
+	var warnings admission.Warnings
+	var allErrs field.ErrorList
+
+	if !enabled {
+		return warnings, allErrs
+	}
+
+	// Warn if OCI URI is used with confidential
+	if len(uri) > 6 && uri[:6] == "oci://" {
```
bartoszmajsak: nit

```suggestion
	if strings.HasPrefix(uri, "oci://") {
```

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -721,3 +728,69 @@ func validateMultipleStorageURIs(isvc *InferenceService) error {
...
+		return warnings, errs.ToAggregate()
+	}
+
+	return warnings, nil
+}
+
+// GetConfidentialSpecFromPredictor extracts the ConfidentialSpec from any predictor implementation
+// that embeds PredictorExtensionSpec.
+func GetConfidentialSpecFromPredictor(predictor *PredictorSpec) *ConfidentialSpec {
+	switch {
```
bartoszmajsak, line 770: I wouldn't go as far as using reflection here for future-proofing, but maybe we can use `reflect` in the test to catch "forgotten case" (newly added predictor)? WDYT?

israel-hdez (commented):
I haven't reviewed all changes in detail, but I wonder why a different image is required?

Just from a quick scan, I see little reason to have two images. And given envvars trigger decryption, I'm failing to understand why we can't bundle everything into the existing image.

config/configmap/inferenceservice.yaml
font: Good catch — removed the empty `"confidentialImage": ""` from all configmap files. The `omitempty` on the struct field handles the zero value correctly.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
font: Good point. Moved the confidential config logic into `attachStorageInitializer` so it goes through the same restart-avoidance path (preserving the existing init container image from `curr`). The confidential image is now set on the copied config *before* the existing-image check, and env vars are injected before appending to the pod spec. Added a test case for the restart-avoidance behavior.

config/configmap/inferenceservice.yaml
font: Agree — adding a GH action to build and publish the confidential storage initializer image would be the right next step. I'll track this as a follow-up once the feature lands.

pkg/apis/serving/v1beta1/inference_service_validation.go
font: Good idea. The switch statement in `GetConfidentialSpecFromPredictor` would silently miss a new predictor type. A reflect-based test that discovers all `PredictorSpec` fields embedding `PredictorExtensionSpec` and verifies coverage would catch that. I'll address this as a follow-up since it's an improvement to pre-existing code outside the scope of this PR.

pkg/validation/confidential.go
font: Done.

font (commented):
@israel-hdez A separate image was settled on in the design doc and the community also preferred it. The goal is to minimize the Trusted Computing Base (TCB) — in confidential computing, everything inside the TEE is part of the attack surface. The confidential image adds `jwcrypto` (for JWE decryption) and a CDH client, which are only needed inside a TEE. Bundling these into the standard storage initializer would add crypto dependencies to every KServe deployment even when confidential serving isn't used.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -573,11 +573,28 @@ func CommonStorageInitialization(ctx context.Context, params *StorageInitializer
 		// This is done after merging to avoid conflicts with user-defined environment variables.
 		// See https://github.com/kserve/kserve/issues/4761
 		utils.AddDefaultHuggingFaceEnvVars(initContainer)
+
+		// Apply confidential model serving configuration if enabled via annotations
+		applyConfidentialConfig(initContainer, params.IsvcAnnotations, params.Config)
```
israel-hdez: We should make this compatible with `ClusterStorageContainer` CRD ([docs here](https://kserve.github.io/website/docs/model-serving/storage/storage-containers)).

As it is, it can be inconsistent, because this can replace a custom storage-init container with the built-in one, which would not have user customizations nor support any user protocols, resulting in a broken setup. OTOH, a user custom image may not have confidential capabilities which is a guard that we should have and surface in the Status.

Nowadays, `ClusterStorageContainer` is present in default setups. Thus, at least on my side, this is a blocker for an approval. Although, I'm OK to let it go if some other reviewer disagrees.

font: The right approach may be: when a `ClusterStorageContainer` matches, don't swap the image — only inject the confidential env vars (`CONFIDENTIAL_ENABLED`, `CONFIDENTIAL_RESOURCE_ID`). This way the custom image's protocol support and user customizations are preserved. If the custom image doesn't handle the confidential env vars, it will simply ignore them and download the model unencrypted (same behavior as today without this feature).

For the guard you mentioned — we could add a validation warning when confidential is enabled and a `ClusterStorageContainer` matches, letting the user know their custom image needs to support confidential decryption for it to take effect.

Does that approach work for you, or would you prefer we block the request entirely when there's a `ClusterStorageContainer` mismatch?

Do any others have an opinion?

israel-hdez: The problem that I see is that `ClusterStorageContainer` is not only for custom configs. We ship a default configuration for the built-in storage initializer. It is here: https://github.com/kserve/kserve/blob/master/config/storagecontainers/default.yaml

We can use what you suggest, but it would mean that most setups would not swap the image and would only inject the envvars. This makes me go back to asking why we don't prefer a unified image as opposed to separate ones? IMO, we should use a single pattern: either we use a unified image, or we use different ones and also extend `ClusterStorageContainer` to consistently swap the image in case of confidential serving.

font: Good point about `ClusterStorageContainer`. To clarify the issue: when `confidential.enabled: true`, `applyConfidentialConfig` replaces the init container image with the built-in confidential one. This happens after the `ClusterStorageContainer` merge, so it would overwrite any custom image set by the CRD. Non-confidential workloads are unaffected.

For the default `ClusterStorageContainer`, this is fine — the built-in confidential image is the intended replacement. The issue is when a platform admin has a custom `ClusterStorageContainer` with a custom storage initializer (e.g., supporting a proprietary protocol) and a user enables confidential serving — our swap would replace their custom image with the built-in confidential image, which doesn't support their protocol.

I see a few options:

**Option 1: Unified image** — Ship a single storage initializer image that handles both confidential and non-confidential. The env vars (`CONFIDENTIAL_ENABLED`, `CONFIDENTIAL_RESOURCE_ID`) control behavior. No image swap needed. The confidential entrypoint already falls through to the standard one when confidential is not enabled, so it's effectively a superset. Tradeoff: slightly larger image (CDH client bundled even when not used).

**Option 2: Skip image swap when custom `ClusterStorageContainer` matches** — Only inject the env vars and skip the image swap when a non-default `ClusterStorageContainer` is in use. The admin is responsible for ensuring their custom image supports confidential decryption. For the default `ClusterStorageContainer`, the swap still happens normally. The custom storage initializer would only get swapped when confidential is enabled.

**Option 3: Extend `ClusterStorageContainer` CRD** — Add a `confidentialContainer` field to the CRD spec, so each `ClusterStorageContainer` can define its own confidential variant. The swap happens within the CRD mechanism. Architecturally clean, but increases scope significantly (CRD schema change, migration, docs).

I think I lean toward option 1 for simplicity, but open to other opinions. Thoughts?

israel-hdez: I'm happy with either option 1 or option 3. Unfortunately, I don't have the context about the community discussion for initially choosing different images. So, I'm not sure I can correctly recommend one over the other. Although, I can say that option 2 is my less preferred one.

font:
Updated based on review feedback regarding `ClusterStorageContainer` compatibility.

Instead of swapping the storage initializer image when confidential is enabled, we now use a single unified image that includes the confidential dependencies (~4MB larger). The confidential entrypoint falls through to the standard path when confidential is not enabled, so `ClusterStorageContainer` customizations are fully preserved regardless.

**Summary of changes:**
- Merged `storage-initializer-confidential.Dockerfile` into the standard `storage-initializer.Dockerfile`
- Removed `confidentialImage` from `StorageInitializerConfig` and all image-swap logic
- Simplified `applyConfidentialConfig` to only inject env vars (no image override)
- Deleted `hack/patch-confidential-image.sh` (no longer needed)
- Rebased onto latest master and squashed into 3 clean commits (production, generated, tests)

**CI note:** A few e2e tests are failing but they are pre-existing flakes — the same tests (`test-huggingface-server-vllm`, `test-llm`, `test-predictor`, `test-raw`) also fail on master's E2E workflow. Failures are minikube resource exhaustion and reconciliation race conditions, unrelated to this PR.

@Jooho @israel-hdez @bartoszmajsak @pierDipi

pkg/webhook/admission/pod/storage_initializer_injector.go
font: Ok, this should be resolved now. PTAL.

pkg/validation/confidential.go
```diff
@@ -0,0 +1,61 @@
...
+// webhooks to avoid duplicating the validation logic.
+func ValidateConfidentialSpec(enabled bool, resourceId *string, uri string, basePath *field.Path) (admission.Warnings, field.ErrorList) {
+	var warnings admission.Warnings
+	var allErrs field.ErrorList
+
+	if !enabled {
+		return warnings, allErrs
+	}
+
+	// Warn if OCI URI is used with confidential
```
Jooho, line 44: Don't we need a validation check for pvc too?
pvc:// with confidential: enabled: true is silently ignored — no warning, no error.
Storage-initializer is not injected for pvc.

python/storage-initializer.Dockerfile
```diff
@@ -59,11 +59,11 @@ COPY --from=builder --chown=kserve:kserve $VIRTUAL_ENV $VIRTUAL_ENV
 COPY --from=builder storage storage
 COPY ./storage-initializer /storage-initializer
 
-RUN chmod +x /storage-initializer/scripts/initializer-entrypoint
+RUN chmod +x /storage-initializer/scripts/initializer-entrypoint-confidential
```
Jooho: we are using the same image so we don't need to change this. I think you missed to clean up this part.

python/storage-initializer/scripts/initializer-entrypoint-confidential
Jooho, line 1: we are using the same image so we don't need to change this. I think you missed to clean up this part.

pkg/validation/confidential.go
font: Good catch — addressed in the latest commit. The webhook now warns when pvc:// is used with confidential enabled, similar to the existing OCI warning.

python/storage-initializer.Dockerfile
font: Agreed — addressed in the latest commit. The Dockerfile now uses the standard initializer-entrypoint again. The confidential logic has been folded into it, gated by the CONFIDENTIAL_ENABLED env var.

python/storage-initializer/scripts/initializer-entrypoint-confidential
font: Done — deleted initializer-entrypoint-confidential and merged the confidential logic into the standard initializer-entrypoint. Single script, single image.

pkg/validation/confidential.go
```diff
@@ -0,0 +1,67 @@
...
+	if strings.HasPrefix(uri, "oci://") {
+		warnings = append(warnings,
+			"confidential has no effect with OCI URIs; OCI image decryption is handled by the container runtime via runtimeClassName")
+	}
+
+	// Warn if PVC URI is used with confidential
+	if strings.HasPrefix(uri, "pvc://") {
+		warnings = append(warnings,
+			"confidential has no effect with PVC URIs; the storage-initializer is not injected for PVC volumes")
```
israel-hdez, line 53: The PVC has a non-direct mount mode that would do copying with storage-init (unless already removed?). I want to confirm if we are not covering this case? From the changes in the injector, I think it is covered.

If so, the warning can be inaccurate.

pkg/apis/serving/v1beta1/inference_service_validation.go
israel-hdez: Alternatively, I think this whole func may also be replaced with something like:

```suggestion
func GetConfidentialSpecFromPredictor(predictor *PredictorSpec) *ConfidentialSpec {
	return predictor.GetImplementation().GetConfidential()
}
```

...but we "trade" the simpleness of a switch with a ¿less simple? approach of adding a `GetConfidential` func to `ComponentImplementation` interface and add it on all interface implementations.

(btw, I'm not sure if there is a risk of a nil pointer)

python/storage-initializer/scripts/initializer-entrypoint
```diff
@@ -30,3 +30,35 @@ except RuntimeError as e:
...
+        # regardless of the underlying KBS/RATS implementation.
+        # Future alternatives for non-CoCo environments:
+        # - File-based resolver: read key from a mounted Kubernetes Secret
+        # - Environment variable resolver: key passed directly via env var
+        # - Cloud KMS resolver: retrieve key from a cloud provider KMS
+        resolver = CDHSecretResolver()
+        decryptor = JWEDecryptor(resolver, resource_id=resource_id)
+
+        total_decrypted = 0
+        for dest_path in dest_paths:
```
israel-hdez, line 57: Mere feedback (non-actionable for now).

The python side seems to support multi-uri (see #4702), while the golang side doesn't.

python/storage/kserve_storage/confidential/jwe_decryptor.py
```diff
@@ -0,0 +1,155 @@
...
+                f"No resource_id provided for decryption of {path}. "
+                "Set resource_id on the decryptor or pass it to decrypt_file()."
+            )
+
+        logger.info("Decrypting %s with resource_id=%s", path, rid)
+
+        # NOTE: JWE Compact Serialization requires loading the entire file into
+        # memory.  For very large models this may be a concern; consider splitting
+        # large files before encryption or increasing container memory limits.
+        key_bytes = self._resolver.resolve_key(rid)
```
israel-hdez, line 106: Isn't any major impact for the repeated requests to the CDH?

pkg/apis/serving/v1alpha2/llm_inference_service_validation_test.go
```diff
@@ -1457,3 +1457,101 @@ func TestValidateManagedDRAAnnotations(t *testing.T) {
...
+			name:         "confidential enabled with valid resourceId",
+			confidential: &ConfidentialSpec{Enabled: true, ResourceId: ptr.To("kbs:///default/key/model-key")},
+			modelURI:     apis.URL{Scheme: "hf", Host: "meta-llama/Llama-2-7b"},
+			wantErrCount: 0,
+		},
+		{
+			name:         "confidential enabled without resourceId",
+			confidential: &ConfidentialSpec{Enabled: true},
+			modelURI:     apis.URL{Scheme: "hf", Host: "meta-llama/Llama-2-7b"},
+			wantErrCount: 0,
```
israel-hdez, line 1492: Without a resource id, wouldn't the python code complain?

spolti:
@font hi, thanks for this PR.
Can you also contribute to the docs?

pkg/apis/serving/v1alpha2/llm_inference_service_validation_test.go
font: Good question. [This test](https://github.com/kserve/kserve/blob/3a0a74b3f6ceb216c4063ac993469312ffb5962c/pkg/apis/serving/v1alpha2/llm_inference_service_validation_test.go#L1492-L1500) validates the **OCI-based confidential path**, where the container runtime (`kata-cc` + `image-rs`) handles decryption using key references embedded in the OCI image metadata — the storage-initializer and Python JWE decryptor are not involved.

  The Python code only runs for blob storage URIs (`s3://`, `hf://`, `gs://`), and the webhook requires `resourceId` for those paths. So this test case is valid: `resourceId` is only required when using blob storage with JWE encryption, not when using OCI encrypted images.

pkg/validation/confidential.go
font: The non-direct mount mode for PVC was removed — PVC URIs are now always direct-mounted as volumes (see [`storage_initializer_injector.go` line 407](https://github.com/kserve/kserve/blob/3a0a74b3f6ceb216c4063ac993469312ffb5962c/pkg/webhook/admission/pod/storage_initializer_injector.go#L407): _"Mount PVC storage URIs directly as volumes"_). The storage-initializer is never injected for `pvc://` URIs, so the warning is accurate.

python/storage-initializer/scripts/initializer-entrypoint
font: Correct — the Python storage initializer handles multi-URI ([parses URIs from argv](https://github.com/kserve/kserve/blob/2f9cbe816229e5888c3cdfef2013bd122fc99f0e/python/storage-initializer/scripts/initializer-entrypoint#L14-L15), [downloads all](https://github.com/kserve/kserve/blob/2f9cbe816229e5888c3cdfef2013bd122fc99f0e/python/storage-initializer/scripts/initializer-entrypoint#L26), then [applies JWE decryption to each dest_path](https://github.com/kserve/kserve/blob/2f9cbe816229e5888c3cdfef2013bd122fc99f0e/python/storage-initializer/scripts/initializer-entrypoint#L57-L59)). The Go webhook validation currently only validates the first URI's confidential config against the `InferenceService` spec. The actual download+decrypt pipeline works correctly for multi-URI; improving the Go-side validation to check all URIs would be a worthwhile enhancement in a follow-up PR.

python/storage/kserve_storage/confidential/jwe_decryptor.py
font: The CDH (Confidential Data Hub) runs as a localhost sidecar (`127.0.0.1:8006`), so latency per request is minimal. That said, [`resolve_key()`](https://github.com/kserve/kserve/blob/2f9cbe816229e5888c3cdfef2013bd122fc99f0e/python/storage/kserve_storage/confidential/jwe_decryptor.py#L106) returns the same key for a given `resource_id`, so caching it would avoid redundant HTTP round-trips when decrypting multiple files. I'll add key caching in a follow-up PR.

pkg/apis/serving/v1beta1/inference_service_validation.go
font: That's a cleaner API. The tradeoff is adding `GetConfidential()` to the [`ComponentImplementation` interface](https://github.com/kserve/kserve/blob/3a0a74b3f6ceb216c4063ac993469312ffb5962c/pkg/apis/serving/v1beta1/component.go#L64-L71) plus all implementations (`SKLearn`, `XGBoost`, `LightGBM`, `PMML`, `Paddle`, `Custom`, etc.), most of which just return `nil`. 

It's worth doing, but as a separate refactoring PR since it's a broader interface change that touches code outside the confidential serving feature. The [switch statement in the merged code](https://github.com/kserve/kserve/blob/2f9cbe816229e5888c3cdfef2013bd122fc99f0e/pkg/apis/serving/v1beta1/inference_service_validation.go#L770-L795) works correctly today. I can open a follow-up PR for this if you'd like.

PR #5389, opened 2026-04-14
feat(webhook): inject OVMS auto-versioning init container

bartoszmajsak:

OpenVINO Model Server (OVMS) requires models in a versioned directory structure (`/mnt/models/<version>/`), but the storage initializer downloads files flat into `/mnt/models`. This mismatch causes OVMS to fail at startup with:

```
No version found for model in path: /mnt/models
```

This is already handled in ModelMesh (it appends the versioned directory automatically), but on KServe raw deployment the user has no escape hatch - they'd need to manually restructure their storage layout to match what OVMS expects, making migrations from ModelMesh harder than they need to be.

When the `storage.kserve.io/ovms-auto-versioning` annotation is set to a positive integer on an InferenceService, an init container is injected after the storage initializer that reorganises the downloaded model files into the expected versioned layout. Hidden files (dotfiles) are handled correctly. The container image is digest-pinned to prevent non-determinism across deployments.

The annotation value becomes the version directory name, e.g. `storage.kserve.io/ovms-auto-versioning: "1"` places the model under `/mnt/models/1/`.

Ported from [opendatahub-io/kserve#968](https://github.com/opendatahub-io/kserve/pull/968) and [opendatahub-io/kserve#1330](https://github.com/opendatahub-io/kserve/pull/1330), following the upstream addition of the OpenVINO runtime in #4592.

- Deploy any InferenceService using the `kserve-openvino` runtime with a flat model storage URI (e.g. `https://github.com/spolti/aipg/raw/refs/heads/main/kserve/models/mnist/1/mnist.onnx`) and the annotation `storage.kserve.io/ovms-auto-versioning: "1"`.
- OVMS should start cleanly and the model should be reachable at `/mnt/models/1/`.
- Without the annotation, behaviour is unchanged.

`registry.access.redhat.com` is the public Red Hat registry - no authentication required. It follows the same approach as `python/predictiveserver.Dockerfile` already in this repo.

**Checklist**:

**Release note**:
```release-note
Adds optional OVMS auto-versioning init container injection for InferenceServices using OpenVINO Model Server. Set annotation storage.kserve.io/ovms-auto-versioning to a positive integer to automatically reorganise model files into the versioned directory structure OVMS requires.
```

### DIFF
--- a/charts/kserve-llmisvc-resources/files/common/configmap.yaml
+++ b/charts/kserve-llmisvc-resources/files/common/configmap.yaml
@@ -486,27 +486,6 @@ data:
            "cpuLimit": "1"
        }
 
-     # ====================================== OVMS VERSIONING CONFIGURATION ======================================
-     ovmsVersioning: |-
-       {
-           # image contains the init container image used to reorganise model files into the
-           # versioned directory structure that OpenVINO Model Server requires.
-           # registry.access.redhat.com is the public Red Hat registry (no authentication needed).
-           "image": "registry.access.redhat.com/ubi9/ubi-micro@sha256:2173487b3b72b1a7b11edc908e9bbf1726f9df46a4f78fd6d19a2bab0a701f38",
-
-           # memoryRequest is the requests.memory to set for the ovms versioning init container.
-           "memoryRequest": "64Mi",
-
-           # memoryLimit is the limits.memory to set for the ovms versioning init container.
-           "memoryLimit": "128Mi",
-
-           # cpuRequest is the requests.cpu to set for the ovms versioning init container.
-           "cpuRequest": "50m",
-
-           # cpuLimit is the limits.cpu to set for the ovms versioning init container.
-           "cpuLimit": "100m"
-       }
-
      # ====================================== ROUTER CONFIGURATION ======================================
      # Example
      router: |-
--- a/charts/kserve-llmisvc-resources/files/common/configmap.yaml
+++ b/charts/kserve-llmisvc-resources/files/common/configmap.yaml
@@ -486,6 +486,27 @@ data:
            "cpuLimit": "1"
        }
 
+     # ====================================== OVMS VERSIONING CONFIGURATION ======================================
+     ovmsVersioning: |-
+       {
+           # image contains the init container image used to reorganise model files into the
+           # versioned directory structure that OpenVINO Model Server requires.
+           # registry.access.redhat.com is the public Red Hat registry (no authentication needed).
+           "image": "registry.access.redhat.com/ubi9/ubi-micro@sha256:2173487b3b72b1a7b11edc908e9bbf1726f9df46a4f78fd6d19a2bab0a701f38",
+
+           # memoryRequest is the requests.memory to set for the ovms versioning init container.
+           "memoryRequest": "64Mi",
+
+           # memoryLimit is the limits.memory to set for the ovms versioning init container.
+           "memoryLimit": "128Mi",
+
+           # cpuRequest is the requests.cpu to set for the ovms versioning init container.
+           "cpuRequest": "50m",
+
+           # cpuLimit is the limits.cpu to set for the ovms versioning init container.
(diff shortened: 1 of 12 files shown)


Jooho:
The init container approach and the annotation-based injection pattern both look really solid.

I had one small thought around naming before this lands, mostly to keep things flexible for future use.

Right now, the annotation `storage.kserve.io/ovms-auto-versioning` being OVMS-specific totally makes sense. Different runtimes will naturally need different annotations.

But for the ConfigMap side, the ovmsVersioning key (and the image/resource config under it) feels a bit more like shared infrastructure — essentially a lightweight post-storage-init container. There’s nothing particularly OVMS-specific about things like the ubi-micro image or the default resource settings.

I’m wondering if it might be worth making that part runtime-agnostic — something like storagePostInit instead of ovmsVersioning.

That way, if other runtimes later need similar post-download steps (e.g. reorganizing model files like TF Serving version directories), they could reuse the same image/resource config without introducing additional runtime-specific blocks in inferenceservice-config.

```
annotation (runtime-specific)               configmap (shared infra)
─────────────────────────────               ─────────────────────────
storage.kserve.io/ovms-auto-versioning ──┐
storage.kserve.io/future-runtime-xxx   ──┼──→ storagePostInit { image, cpu, memory }
storage.kserve.io/another-post-init    ──┘
```

Also, more broadly, I’m not entirely sure if runtime-specific details belong in the `inferenceservice-config` ConfigMap in the first place. It feels like that ConfigMap is more intended for shared platform-level configuration, so keeping runtime-specific logic out of it might help avoid coupling and keep things cleaner long-term.

Just a suggestion — happy to hear your thoughts!

Jooho:
Or we can use ClusterStorageContainer

bartoszmajsak:
> Or we can use ClusterStorageContainer

@Jooho can you elaborate?

For the configmap piece - I will have a look, thanks for this feedback!

bartoszmajsak:
@Jooho I removed configmap piece, it's scope creep and wrong. Thanks for bringing it up

israel-hdez (commented):
In the past I discussed with @spolti about this.

This method is present in ODH, but I wonder if we can prefer to keep KServe agnostic?
IMO, the versioning is runtime concern. I think it is much better if we add `initContainers` to the `ServingRuntime` CRD, and we configure it in the [provided resource](https://github.com/kserve/kserve/blob/master/config/runtimes/kserve-openvino.yaml). We may require to use the downward API to pass the value of the annotation so that the init container knows if it needs to do something.

The downside is that the initContainer will always be present, but I see this as an acceptable trade-off to keep storage-initializer agnostic to the runtimes.

FWIW, InferenceService CRD already has `initContainers` field. So, I think adding the same field to SRs should be OK.

bartoszmajsak:
@israel-hdez this sounds reasonable. Let's discuss it with @spolti to align of the direction here, as I'm getting conflicting messaging https://github.com/opendatahub-io/kserve/pull/1330#issuecomment-4237714485 :)

spolti:
I am good with both approaches, the downside to migrate it to the SR would be that we might need to update downstream to align with upstream, or keep downstream as it is as the upstream changes will not affect it.

KillianGolds:
Hey, popping in here since this looks stale. Apologies if I've missed context elsewhere.

Going back through, looks like there are three approaches still in play:

1. Current annotation + webhook injection
2. @israel-hdez's `ServingRuntime.initContainers` route with downward API for the annotation value
3. @Jooho's `ClusterStorageContainer` idea, which I think still needs elaboration unless this has been discussed elsewhere, apologies if so.

@spolti's "I am good with both approaches" reply seems to clear the downstream/upstream concern from the linked ODH PR, so the main blocker looks like just picking a direction.

Without further detail from @Jooho on the CSC angle, option 2 feels like the strongest fit to me, keeping core runtime-agnostic and matching the pattern already on `InferenceService`.

bartoszmajsak:
/hold 

Unclear how we want to upstream it yet

PR #5392, opened 2026-04-14
fix: gate ClusterServingRuntime on CRD availability

bartoszmajsak:

Applies the existing `IsCrdAvailable` pattern (already used for Knative, KEDA, OTel, and Gateway API) to `ClusterServingRuntime`. In some distributions, the CSR CRD is not installed - making it an optional dependency. Without these guards, the controller panics or errors out in those environments.

Three callsites are covered:

- `cmd/manager/main.go`: skip CSR webhook registration when CRD is absent
- `pkg/controller/v1beta1/inferenceservice/controller.go`: skip CSR watch at startup
- `pkg/apis/serving/v1beta1/predictor_model.go`: treat `NoMatchError` as empty list when listing `ClusterServingRuntime`s
- `pkg/controller/v1beta1/inferenceservice/utils/utils.go`: treat `NoMatchError` as not-found when getting a named `ClusterServingRuntime`

The guards are a no-op in standard kserve deployments where the CSR CRD is always present.

Fixes #

- [x] `go build ./...` passes
- [x] `go vet ./...` clean
- [x] No behavior change in standard deployments - CSR CRD is always present there, guards never trigger

Uses the same `IsCrdAvailable` pattern already in use for Knative, KEDA, OTel, and Gateway API. Port of opendatahub-io/kserve#1327.

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -45,6 +45,7 @@ import (
 	"github.com/kserve/kserve/pkg/controller/v1alpha1/trainedmodel/reconcilers/modelconfig"
 	v1beta1controller "github.com/kserve/kserve/pkg/controller/v1beta1/inferenceservice"
 	kservescheme "github.com/kserve/kserve/pkg/scheme"
+	kserveutils "github.com/kserve/kserve/pkg/utils"
 	"github.com/kserve/kserve/pkg/webhook/admission/pod"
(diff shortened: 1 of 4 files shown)


pierDipi (approved):
/lgtm

bartoszmajsak:
@Jooho @sivanantha321 I would appreciate the review here.

spolti (approved):

PR #5393, opened 2026-04-14
fix: intermittent 403 Forbidden errors during S3 model download

agoston:
**What this PR does / why we need it**: Fixes intermittent 403 Forbidden and AccessDenied errors during S3 model downloads.

boto3 relies on botocore, which maintains a connection pool (urllib3) under the hood. When Python's multiprocessing forks a new worker process, the child process inherits the parent's memory space, including the default boto3 session and its open SSL sockets. If multiple worker processes try to read/write concurrently using the same shared SSL socket, the byte streams get interleaved. This corrupts the SSL state and leads to errors like BadStatusLine or SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC.

Tested via local stress-tester script.

### DIFF
--- a/python/storage/kserve_storage/kserve_storage.py
+++ b/python/storage/kserve_storage/kserve_storage.py
@@ -377,7 +377,8 @@ class Storage(object):
             import boto3
 
             kwargs = Storage._get_s3_client_kwargs()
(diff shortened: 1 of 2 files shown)


Jooho (approved):
good catch! 

/lgtm
/approved

PR #5394, opened 2026-04-14
feat(ingress): make RawDeployment's HTTPRoute configurable

### LINKED ISSUE #5384: RawDeployment HTTPRoute generation is not configurable enough for production Gateway API deployments
/kind feature

## Summary

When `enableGatewayApi: true` is set in `inferenceservice-config`, the `RawHTTPRouteReconciler` generates HTTPRoutes for RawDeployment `InferenceService` resources. 

The generated routes have several hardcoded behaviours that make them unsuitable for production Gateway API deployments using path-based routing. The issues force operators to set `disableIngressCreation: true` and maintain manual HTTPRoutes, defeating the purpose of the reconciler.

This issue is partially inspired by how HTTPRoute creation is handled for LLMIsvc.

## Configuration (inferenceservince-config CM, ingress block)

```json
{
  "enableGatewayApi": true,
  "kserveIngressGateway": "envoy-gateway-system/main-gateway",
  "ingressDomain": "example.domain.com",
  "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}",
  "domainTemplate": "{{ .Name }}-{{ .Namespace }}.{{ .IngressDomain }}",
  "urlScheme": "https",
  "disableIngressCreation": false
}
```

## Current vs. desired behaviour

**What is generated** (simplified):

```yaml
# HTTPRoute; ns raw-test, isvc sklearn-iris
spec:
  hostnames:
    - sklearn-iris-raw-test.example.domain.com
    - example.domain.com         # catches ALL traffic on the shared domain
  parentRefs:
    - name: main-gateway
      namespace: envoy-gateway-system
      # no sectionName — attaches to every listener
  rules:
    - matches:
        - path:
            type: RegularExpression
            value: "^/.*$"           # host catch-all, conflicts with other routes
      # no URLRewrite — backend receives /serving/ns/name/... and returns 404
      timeouts:
        request: 60s                 # too short for inference workloads
    - matches:
        - path:
            type: RegularExpression
            value: "/serving/raw-test/sklearn-iris/"
      # still no URLRewrite
```

**What is needed:**

```yaml
spec:
  parentRefs:
    - name: main-gateway
      namespace: envoy-gateway-system
      sectionName: https             # pinned to the correct listener
  rules:
    - matches:
        - path:
            type: PathPrefix         # portable, no regex engine required
            value: /serving/raw-test/sklearn-iris/
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: /  # backend receives /v1/models/... as expected
      timeouts:
        request: 300s
        backendRequest: 300s
```

## Problems

1. **Host catch-all hijacks the shared ingress domain.** When `pathTemplate` is set, `ingressDomain` is unconditionally added to `hostnames` and a `^/.*$` catch-all rule is generated. In a multi-tenant cluster every ISVC competes for the same hostname, intercepting traffic intended for other services. There is no config option to disable this.
2. **No URL rewrite on path-based rules.** The path prefix (e.g. `/serving/ns/name/`) is forwarded to the backend as-is. Most model servers expect requests at the root path and return 404. There is no config option to add a `URLRewrite` filter.
3. **Path match type is hardcoded to `RegularExpression`.** The generated values are plain prefixes; `PathPrefix` is more efficient, more portable, and the standard approach. There is no config option to change this.
4. **No `sectionName` on the `parentRef`.** The route attaches to all listeners on the gateway, including HTTP, potentially bypassing TLS or conflicting with hostname-specific listeners.
5. **No way to add custom labels to generated HTTPRoutes.** This prevents applying Gateway API `SecurityPolicy` resources via `targetSelectors` (label-based matching), which is the standard policy-attachment pattern.

## Proposed solution

We propose adding a `rawDeployment` sub-block to the existing `ingress` configmap key. This follows the same nested-struct pattern already used in `inferenceservice-config` for `credentials`, `deploy`, and `router`. New fields — covering `gatewayListenerName` (`sectionName`), `pathMatchType`, `pathRewriteTarget`, `disableHostBasedRouting`, `routeLabels`, `requestTimeout`, and `backendRequestTimeout` — would all default to zero values that reproduce the current v0.17.0 behaviour exactly, making the change fully backwards-compatible.

Two alternatives would be: (A) a full HTTPRouteSpec template inspired by `LLMInferenceServiceConfig` (flexible but heavy); and (B) updating the "ingress" block in the ConfigMap (messy though already contains some RawDeployment-specific values).

## Environment

- KServe: v0.17.0
- Gateway: Envoy Gateway v1.7.0
- Kubernetes: v1.32.13 (MicroK8s), GCP VMs

## Related

- The `pathTemplate` field was introduced in response to [#2257](https://github.com/kserve/kserve/issues/2257) but the implementation added only the path match rule without addressing URL rewriting, hostname conflicts, or path match type.

tmvfb:

In a nutshell, currently the automatically generated HTTPRoutes with `enableGatewayApi: true` in `inferenceservice-config` are unusable in production (#5384).

**Note**: this partially overlaps with the logic added in #5313, #5347 (found this out after creating the PR), see my comment: https://github.com/kserve/kserve/issues/5319#issuecomment-4253041437. My idea would be to consolidate options from those and current PRs in one `ingress.rawDeployment` block + extend the current PR to per-component routes (predictor/transformer/explainer). But I would love maintainer's feedback on the way to proceed.

---

Adds a `rawDeployment` sub-block to the `ingress` key of `inferenceservice-config` that makes HTTPRoute generation for RawDeployment `InferenceService` resources configurable via Gateway API. All new fields default to zero values that reproduce v0.17.0 behaviour exactly, so no existing deployment is affected.

From a design standpoint, this is a middle ground between Serverless ingress config (just the `ingress` block) and LLMIsvc ingress config (HTTPRoute configurable per LLMIsvc). This PR adds a single source of truth for all RawDeployments' HTTProutes while offering some essential configurability. 

Fixes #5384

- [x] Unit tests added for all new `RawDeploymentIngressConfig` fields and validation logic (`configmap_test.go`, `httproute_reconciler_test.go`)
- [x] Validated on a live cluster (MicroK8s, Kubernetes v1.32.13, Envoy Gateway v1.7.0) with `sklearn-iris` RawDeployment ISVC:
  - HTTPRoute status: `Accepted: True`, `ResolvedRefs: True`
  - `GET /serving/raw-test/sklearn-iris/v1/models/sklearn-iris` with valid `x-api-key`: **200 OK**
  - Same request without key: **401 Unauthorized** (gateway `SecurityPolicy` enforced via `routeLabels`)
  - Inference `POST /serving/.../sklearn-iris:predict` → `{"predictions":[1]}`

Recommended production config for path-based routing:

```json
{
  "enableGatewayApi": true,
  "pathTemplate": "/serving/{{ .Namespace }}/{{ .Name }}",
  "rawDeployment": {
    "gatewayListenerName": "https",
    "pathMatchType": "PathPrefix",
    "pathRewriteTarget": "/",
    "disableHostBasedRouting": true,
  }
}
```

`disableHostBasedRouting` only affects the **top-level HTTPRoute** for the ISVC (the per-component predictor/transformer/explainer subdomain routes are unaffected — suppressing/editing those is a follow-up, deliberately out of scope here).

Without the flag, the generated top-level HTTPRoute looks like this (from issue #5384):

```yaml
spec:
  hostnames:
  - sklearn-iris-raw-test.example.domain.com   # domainTemplate — won't match an exact-hostname listener
  - example.domain.com                          # ingressDomain — catch-all, hijacks all traffic on the shared domain
  rules:
  - matches:
    - path: { type: RegularExpression, value: "^/.*$" }   # host catch-all rule
  - matches:
    - path: { type: RegularExpression, value: "/serving/raw-test/sklearn-iris/" }
```

With `disableHostBasedRouting: true`:

```yaml
spec:
  hostnames: []   # empty — gateway listener's own hostname filter takes effect
  rules:
  - matches:
    - path: { type: PathPrefix, value: /serving/raw-test/sklearn-iris/ }
    filters:
    - type: URLRewrite
      urlRewrite:
        path: { type: ReplacePrefixMatch, replacePrefixMatch: / }
```

The empty `hostnames` is not a bug — per Gateway API spec a route with no hostnames matches whatever the listener's hostname allows. So if the listener is pinned to `example.domain.com`, that filter applies and the route is `Accepted: True`. Without the flag, the explicit hostnames in the route conflict with the listener's exact hostname and the route is `Accepted: False, NoMatchingListenerHostname`.

**Checklist**:

```release-note
Add `rawDeployment` HTTPRoute config block to `inferenceservice-config` ingress settings
```


tmvfb:
Rebased the branch, looks like failing CI is due to flaky tests, not due to changes

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -248,6 +262,68 @@ func createHTTPRouteRule(routeMatches []gwapiv1.HTTPRouteMatch, filters []gwapiv
...
+	if ingressConfig.RawDeployment == nil {
+		return
+	}
+	for k, v := range ingressConfig.RawDeployment.RouteLabels {
+		labels[k] = v
+	}
+}
+
+// resolveRawRequestTimeout returns the effective request timeout, preferring rawDeployment.requestTimeout when set.
+func resolveRawRequestTimeout(disableTimeout bool, componentTimeout *int64, rawCfg *v1beta1.RawDeploymentIngressConfig) *gwapiv1.Duration {
```
vivekk16, line 347: `resolveRawRequestTimeout` can silently override `disableHTTPRouteTimeout: true`.

When `disableHTTPRouteTimeout` is `true`, `resolveTimeout()` correctly returns `nil` (no timeout on the route). But if `rawDeployment.requestTimeout` is also set, this function returns a non-nil duration while adding the timeouts field back onto the `HTTPRoute`. On gateways that reject the timeouts field (e.g. GKE), this causes the route to be `Accepted: False` with no obvious indication that the two settings conflict.

vivekk16: Either validate the conflict in `validateRawDeploymentConfig`  at config parse time, or let `disableHTTPRouteTimeout` always win.

charts/kserve-resources/files/common/configmap-patch.yaml
```diff
@@ -87,7 +87,24 @@ data:
         "urlScheme": "{{ .Values.kserve.controller.gateway.urlScheme }}",
         "disableIstioVirtualHost": {{ .Values.kserve.controller.gateway.disableIstioVirtualHost }},
         "disableIngressCreation": {{ .Values.kserve.controller.gateway.disableIngressCreation }},
-        "disableHTTPRouteTimeout": {{ .Values.kserve.controller.gateway.disableHTTPRouteTimeout }}
+        "disableHTTPRouteTimeout": {{ .Values.kserve.controller.gateway.disableHTTPRouteTimeout }},
+        "rawDeployment": {
```
vivekk16, line 93: The `rawDeployment` block is always rendered in the Helm ConfigMap, even when all values are at their defaults (empty strings, `false`, `{}`). This means Helm deployed clusters will always parse `ingressConfig.RawDeployment` as a non-nil pointer to a zero value struct, while kustomize deployed clusters get `nil`.

This diverges from the "omit the block = nil = no behavior change" semantics tested in `TestNewIngressConfigWithoutRawDeployment`. Functionally equivalent today since the code checks individual fields, but it's a latent inconsistency that could surprise future contributors who check `if ingressConfig.RawDeployment != nil`.

I'd suggest conditionally rendering the block only when at least one field is non-default.

charts/kserve-resources/files/common/configmap-patch.yaml
```diff
@@ -87,7 +87,24 @@ data:
...
+            "gatewayListenerName": "{{ .Values.kserve.controller.gateway.rawDeployment.gatewayListenerName }}",
+            "pathMatchType": "{{ .Values.kserve.controller.gateway.rawDeployment.pathMatchType }}",
+            "pathRewriteTarget": "{{ .Values.kserve.controller.gateway.rawDeployment.pathRewriteTarget }}",
+            "disableHostBasedRouting": {{ .Values.kserve.controller.gateway.rawDeployment.disableHostBasedRouting }},
+            "routeLabels": {
```
vivekk16, line 98: When `routeLabels` is empty (the default), the Helm template renders `"routeLabels": {}`, which Go unmarshals as an empty non-nil map `(map[string]string{})`. The kustomize path (where rawDeployment is absent) gives a nil map. Both behave identically for range and len, but differ for != nil checks.

This is a subset of the "always emit rawDeployment" issue above. Fixing that (conditional rendering) fixes this too.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -352,6 +383,38 @@ func NewIngressConfig(isvcConfigMap *corev1.ConfigMap) (*IngressConfig, error) {
 	return ingressConfig, nil
 }
 
+// validateRawDeploymentConfig validates the RawDeploymentIngressConfig fields.
+// It is called from NewIngressConfig when the rawDeployment block is present.
+func validateRawDeploymentConfig(cfg *RawDeploymentIngressConfig) error {
```
vivekk16, line 410: `validateRawDeploymentConfig` validates `pathMatchType`, `pathRewriteTarget`, and timeout durations, but doesn't validate `routeLabels` keys/values against Kubernetes label rules. Invalid labels (spaces, too long, invalid characters) will pass config parsing fine but then cause the HTTPRoute to be rejected by the API server at runtime with a cryptic error.

Would be nice to add label validation using the standard Kubernetes validation utilities (`validation.IsQualifiedName` for keys, `validation.IsValidLabelValue` for values). That way misconfiguration gets caught at startup with a clear message, consistent with how the other fields are already validated.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -509,133 +591,194 @@ func createRawTopLevelHTTPRoute(ctx context.Context, client client.Client, isvc
...
+			httpRouteRules = append(httpRouteRules, createHTTPRouteRuleWithTimeouts(pathRouteMatch, applyRewriteFilter(filters, rewriteFilter), transformerName, isvc.Namespace,
+				constants.CommonDefaultHttpPort, timeout, rawBackendTimeout))
 		} else {
-			timeout := resolveTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Predictor.TimeoutSeconds)
+			timeout := resolveRawRequestTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Predictor.TimeoutSeconds, rawCfg)
 			// :predict routes to the predictor when there is only predictor
-			pathRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(path + "/")}
-			httpRouteRules = append(httpRouteRules, createHTTPRouteRule(pathRouteMatch, filters, predictorName, isvc.Namespace,
-				constants.CommonDefaultHttpPort, timeout))
+			pathRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatchWithType(path+"/", pathMatchType)}
```
vivekk16: Same as above.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -509,133 +591,194 @@ func createRawTopLevelHTTPRoute(ctx context.Context, client client.Client, isvc
...
+					Reason: "Transformer ingress not created",
+				})
+				return nil, nil
+			}
+			timeout := resolveRawRequestTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Transformer.TimeoutSeconds, rawCfg)
 			// :predict routes to the transformer when there are both predictor and transformer
-			pathRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatch(path + "/")}
-			httpRouteRules = append(httpRouteRules, createHTTPRouteRule(pathRouteMatch, filters, transformerName, isvc.Namespace,
-				constants.CommonDefaultHttpPort, timeout))
+			pathRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatchWithType(path+"/", pathMatchType)}
```
vivekk16: The path match value is constructed as path + "/" (e.g. `/serving/raw-test/sklearn-iris/`). With PathPrefix match type, a request to the bare path without trailing slash (`/serving/raw-test/sklearn-iris`) won't match (the request is shorter than the prefix).

While the full inference URLs (e.g. `.../v1/models/sklearn-iris:predict`) work fine, the bare path is commonly used for health checks or service discovery and will unexpectedly return 404.

I'd suggest to use the bare path as the prefix value.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -509,133 +591,194 @@ func createRawTopLevelHTTPRoute(ctx context.Context, client client.Client, isvc
...
+					Status: corev1.ConditionFalse,
+					Reason: "Explainer ingress not created",
+				})
+				return nil, nil
+			}
+			timeout := resolveRawRequestTimeout(ingressConfig.DisableHTTPRouteTimeout, isvc.Spec.Explainer.TimeoutSeconds, rawCfg)
+			// Add path based routing rule for :explain endpoint.
+			// PathBasedExplainPrefix() returns a regex fragment, so the match type must
+			// always be RegularExpression regardless of the configured pathMatchType.
+			explainerPathRouteMatch := []gwapiv1.HTTPRouteMatch{createHTTPRouteMatchWithType(path+constants.PathBasedExplainPrefix(), gwapiv1.PathMatchRegularExpression)}
```
vivekk16, line 793: The `:explain` rule uses RegularExpression match (since the pattern has `(/|:)explain`), but it still gets the `ReplacePrefixMatch` rewrite filter applied. That rewrite is only spec-defined for `PathPrefix` matches. With regex, the gateway doesn't know what the "prefix" is, so behavior varies by implementation (Envoy might do one thing, Istio another, GKE rejects regex entirely).

Since the comment above already calls out that this is always a regex rule, I think it makes sense to just skip the rewrite. If rewriting is actually needed for `:explain `in the future, `ReplaceFullPath` would be the right filter type since it works regardless of match type.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
tmvfb: Good catch, made `disableHTTPRouteTimeout` always win, updated tests accordingly.

charts/kserve-resources/files/common/configmap-patch.yaml
tmvfb: The `rawDeployment` block is now wrapped in `{{- if or ... }}` and only emitted when at least one field is non-default.

charts/kserve-resources/files/common/configmap-patch.yaml
tmvfb: Fixed in the same fashion as previous comment (conditional `rawDeployment` block).

pkg/apis/serving/v1beta1/configmap.go
tmvfb: Fixed. Added validation of `routeLabels` keys with `validation.IsQualifiedName` and values with `validation.IsValidLabelValue` (both already imported via `k8s.io/apimachinery/pkg/util/validation`).

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
tmvfb: That's a good finding, fixed. Extracted a `resolvePathValue()` helper that returns the bare path for `PathPrefix` (so `/serving/ns/isvc` matches both `/serving/ns/isvc` and `/serving/ns/isvc/...`) and path + "/" for `RegularExpression` (preserving existing behaviour). Applied to both the transformer and predictor-only predict rules.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
tmvfb: Fixed

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
tmvfb: Fixed, the `:explain` rule no longer gets the rewrite filter. Expanded the inline comment to explain why: `ReplacePrefixMatch` is only spec-defined for `PathPrefix` matches, so applying it to a regex rule produces implementation-defined behaviour. Noted `ReplaceFullPath` as the correct approach if rewriting is ever needed here.

tmvfb:
@vivekk16, thanks for the prompt and valuable feedback! I addressed your comments, please take a look.

Small question before proceeding:

`disableHTTPRouteTimeout` is currently a top-level ingress field (added in #5313), but it's exclusively relevant to the RawDeployment / Gateway API path; the same scope as everything inside `rawDeployment`. Should we maybe move `disableHTTPRouteTimeout` to that block in this PR as well? #5313 was a part of 0.18.0 release though, so we'd need to think a bit about backward compatibility.

vivekk16:
@tmvfb I would prefer to accept `disableHTTPRouteTimeout` in both locations (top level takes precedence), log a deprecation warning when the top level field is used, and remove the top level field in a future release. That way existing configs keep working and users get a clear migration path.

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
```diff
@@ -248,6 +262,81 @@ func createHTTPRouteRule(routeMatches []gwapiv1.HTTPRouteMatch, filters []gwapiv
...
+// disableHTTPRouteTimeout always takes precedence: when true, no timeout is emitted regardless of rawDeployment config.
+func resolveRawRequestTimeout(disableTimeout bool, componentTimeout *int64, rawCfg *v1beta1.RawDeploymentIngressConfig) *gwapiv1.Duration {
+	if disableTimeout {
+		return nil
+	}
+	if rawCfg != nil && rawCfg.RequestTimeout != "" {
+		d := gwapiv1.Duration(rawCfg.RequestTimeout)
+		return &d
+	}
+	return resolveTimeout(disableTimeout, componentTimeout)
```
vivekk16: Nit: when `resolveTimeout(disableTimeout, componentTimeout)` is reached, `disableTimeout` is guaranteed to be `false `

Just pass false directly to make the invariant obvious.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -352,6 +383,49 @@ func NewIngressConfig(isvcConfigMap *corev1.ConfigMap) (*IngressConfig, error) {
...
+			return fmt.Errorf("invalid ingress config - rawDeployment.routeLabels key %q: %s", k, strings.Join(errs, "; "))
+		}
+		if errs := validation.IsValidLabelValue(v); len(errs) > 0 {
+			return fmt.Errorf("invalid ingress config - rawDeployment.routeLabels value %q for key %q: %s", v, k, strings.Join(errs, "; "))
+		}
+	}
+
+	// Validate duration strings early so misconfigured values are caught before producing a rejected HTTPRoute.
+	if cfg.RequestTimeout != "" {
+		if _, err := time.ParseDuration(cfg.RequestTimeout); err != nil {
```
vivekk16, line 439: `time.ParseDuration` is more permissive than Gateway API's Duration type. It accepts values like "1.5s", "-5s", and "100us" that will pass this check but get rejected at runtime when the gateway controller validates the `HTTPRoute (Accepted: False).` 

Could we add a regex check matching the Gateway API pattern `^([0-9]{1,5}(h|m|s|ms)){1,4}$` after the `ParseDuration` call so users get a clear error at config load time instead of a confusing runtime failure?

charts/kserve-resources/files/common/configmap-patch.yaml
```diff
@@ -88,6 +88,26 @@ data:
         "disableIstioVirtualHost": {{ .Values.kserve.controller.gateway.disableIstioVirtualHost }},
         "disableIngressCreation": {{ .Values.kserve.controller.gateway.disableIngressCreation }},
         "disableHTTPRouteTimeout": {{ .Values.kserve.controller.gateway.disableHTTPRouteTimeout }}
+        {{- $rd := .Values.kserve.controller.gateway.rawDeployment -}}
+        {{- if or $rd.gatewayListenerName $rd.pathMatchType $rd.pathRewriteTarget $rd.disableHostBasedRouting (not (eq (len $rd.routeLabels) 0)) $rd.requestTimeout $rd.backendRequestTimeout }},
```
vivekk16: If a user sets `rawDeployment: null` in their values override (or omits it entirely), $rd will be nil and accessing `$rd.gatewayListenerName` will crash the Helm render. 

Could we guard with `{{- if and $rd (or $rd.gatewayListenerName ...) }}`? Same applies to the other two chart copies (kserve-llmisvc-resources and _common).

pkg/apis/serving/v1beta1/configmap_test.go
```diff
@@ -689,39 +770,141 @@ func TestNewIngressConfig_Validation(t *testing.T) {
 		g.Expect(cfg.KnativeLocalGatewayService).To(gomega.Equal("my-local-gateway-service"))
 	})
 
-	t.Run("returns error if pathTemplate is valid but ingressDomain is empty", func(t *testing.T) {
+	t.Run("returns config when all required fields are present", func(t *testing.T) {
```
vivekk16, line 787: The test for "pathTemplate set but ingressDomain empty" was removed here, but the validation it covers still exists. Could we keep it as a separate t.Run?

vivekk16 (commented):
@tmvfb Left a few minor comments on duration validation, a removed test, and a Helm nil guard. Please also add backward compatibility for disableHTTPRouteTimeout (accept it in both locations (top-level takes precedence + deprecation warning)).

tmvfb:
> @tmvfb I would prefer to accept `disableHTTPRouteTimeout` in both locations (top level takes precedence), log a deprecation warning when the top level field is used, and remove the top level field in a future release. That way existing configs keep working and users get a clear migration path.

Added this logic and added deprecation comments. The idea would be that the top-level value can be safely deleted after the next release. The controller also fires a log warning.

The only deviation: since both values have "false" by default, whoever *sets the value to "true"* will take precedence (OR logic).

---

I addressed the new comments, thanks for the review round!

pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
tmvfb: done

pkg/apis/serving/v1beta1/configmap.go
tmvfb: Initially omitted the validation deliberately, since I was not expecting users to test controller's patience :) Reasonable suggestion, added the guard

charts/kserve-resources/files/common/configmap-patch.yaml
tmvfb: addressed

pkg/apis/serving/v1beta1/configmap_test.go
tmvfb: Returned the removed test

charts/_common/common-patches/configmap-patch.yaml
vivekk16, line 1: Nit: My earlier nil guard comment on $rd was meant to cover the fields inside the block too (should have been more specific). 

Three spots could still break if individual keys end up nil,

`(len $rd.routeLabels)` panics on nil — guard with `(and $rd.routeLabels (not (eq (len $rd.routeLabels) 0)))`

`{{ $rd.disableHostBasedRouting }}` and `{{ $rd.disableHTTPRouteTimeout }}` render <no value> on nil — add `| default false`

Same for the other two chart copies.

vivekk16 (commented):
Kindly address the nit.
Rest of the changes LGTM!

cc @Jooho @spolti @sivanantha321

spolti:
@kserve/kserve-approvers @kserve/kserve-reviewers
Please take a look.

vivekk16:
@spolti can we have this reviewed please?

PR #5396, opened 2026-04-14
fix: de-hardcode torch wheels; upgrade torch/vision

jlost:
## Summary

- `custom_model` and `custom_transformer` declare `requires-python = "<3.13,>=3.10"` but pin torch/torchvision to cp311-only wheel URLs in `tool.uv.sources`
- uv 0.11+ validates dependency resolution across all declared Python versions, not just the active one
- This causes build failures when multiple Python interpreters are present (e.g. UBI9 base images ship both 3.9 and 3.11) or when `PYTHON_VERSION` is set to anything other than 3.11
- By using a uv index, we can de-hardcode the wheel URL to allow it to be constructed dynamically for different python versions, adding cross version support
- Updated torch version

### Reproduction Steps

```bash
cd python
podman build --build-arg PYTHON_VERSION=3.10 -f custom_transformer.Dockerfile .
```

Fails with:

```
error: Distribution `torch==2.9.0+cpu @ direct+https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl` can't be installed because the binary distribution is incompatible with the current platform

hint: You're using CPython 3.10 (`cp310`), but `torch` (v2.9.0+cpu) only has wheels with the following Python implementation tag: `cp311`
```

## Test plan

- [x] `podman build -f custom_transformer.Dockerfile .` passes (Python 3.11, default)
- [x] `podman build --build-arg PYTHON_VERSION=3.10 -f custom_transformer.Dockerfile .` now gives a clean "unsupported Python" error from uv instead of a confusing resolution failure
- [ ] CI passes

### DIFF
--- a/python/custom_model/pyproject.toml
+++ b/python/custom_model/pyproject.toml
@@ -6,8 +6,8 @@ license = {text = "Apache-2.0"}
 requires-python = "<3.13,>=3.10"
 dependencies = [
     "kserve[ray] @ file:///${PROJECT_ROOT}/../kserve",
-    "torchvision==0.24.0+cpu",
-    "torch==2.9.0+cpu",
+    "torchvision==0.26.0+cpu",
+    "torch==2.11.0+cpu",
     "pillow<11.0.0,>=10.3.0",
 ]
 name = "custom-model"
@@ -20,6 +20,10 @@ dev = [
 ]
 
 [tool.uv.sources]
-# For PyTorch CPU on Linux x86_64 with Python 3.11
-torch = { url = "https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl" }
-torchvision = { url = "https://download.pytorch.org/whl/cpu/torchvision-0.24.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl" }
+torch = [{ index = "pytorch-cpu" }]
+torchvision = [{ index = "pytorch-cpu" }]
+
+[[tool.uv.index]]
+name = "pytorch-cpu"
+url = "https://download.pytorch.org/whl/cpu"
+explicit = true
--- a/python/custom_model/uv.lock
+++ b/python/custom_model/uv.lock
@@ -421,8 +421,8 @@ dev = [
 requires-dist = [
     { name = "kserve", extras = ["ray"], directory = "../kserve" },
     { name = "pillow", specifier = ">=10.3.0,<11.0.0" },
-    { name = "torch", url = "https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl" },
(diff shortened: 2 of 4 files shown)


Jooho (approved):
/lgtm

python/custom_model/pyproject.toml
```diff
@@ -3,7 +3,7 @@ authors = [
     {name = "The KServe Authors"},
 ]
 license = {text = "Apache-2.0"}
-requires-python = "<3.13,>=3.10"
+requires-python = ">=3.11,<3.12"
```
spolti, line 6: Are we sure that we want to enforce only python 3.11?

jlost: Non-3.11 is broken per the repro steps, so the alternative would be to fix 3.10/3.12 support, something like: 
```
# PyTorch CPU wheels for Linux x86_64, per Python minor version
torch = [
  { url = "https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl", marker = "python_version == '3.10'" },
  { url = "https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl", marker = "python_version == '3.11'" },
  { url = "https://download.pytorch.org/whl/cpu/torch-2.9.0%2Bcpu-cp312-cp312-manylinux_2_28_x86_64.whl", marker = "python_version == '3.12'" },
]
torchvision = [
  { url = "https://download.pytorch.org/whl/cpu/torchvision-0.24.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl", marker = "python_version == '3.10'" },
  { url = "https://download.pytorch.org/whl/cpu/torchvision-0.24.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl", marker = "python_version == '3.11'" },
  { url = "https://download.pytorch.org/whl/cpu/torchvision-0.24.0%2Bcpu-cp312-cp312-manylinux_2_28_x86_64.whl", marker = "python_version == '3.12'" },
]
```

jlost: But since the other python versions didn't seem to be used anyway, it seemed simpler and lighter to remove the non-3.11 support

spolti: > But since the other python versions didn't seem to be used anyway,

can you please elaborate?

Wondering if we can't also upgrade the torch dependency. I don't think it is mandatory to stay on 2.9, could you please check?

jlost: >can you please elaborate?

I mean non-3.11 builds must not have been in use, since they wouldn't work anyway.

>Wondering if we can't also upgrade the torch dependency. I don't think it is mandatory to stay on 2.9, could you please check?

I pushed a different approach that allows the full range of Python versions to work without specifying specific wheel files for each one. I also updated the torch versions while I was at it. This solves the same issue without narrowing the python versions. Let me know which one you prefer.

spolti (approved):

PR #5381, opened 2026-04-14
fix(llmisvc): shorten webhook name to comply with 63-char label limit

spolti:
The LLMInferenceServiceConfig validating webhook name `llminferenceserviceconfig.kserve-webhook-server.v1alpha{1,2}.validator` was 66 characters, exceeding the Kubernetes 63-character label value limit. This caused OLM CSV installation to fail as it uses the webhook name as a generateName label value.

Shortened to `llmisvcconfig.kserve-webhook-server.v1alpha{1,2}.validator` (54 chars).

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Feature/Issue validation/testing**

- [ ] Test A
- [ ] Test B

**Checklist**:

- [ ] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note

```

PR #5398, opened 2026-04-15
revert(llmisvc): restore original webhook names, reverts #5381

bartoszmajsak:

Reverts #5381, which shortened the `ValidatingWebhookConfiguration` webhook names for `LLMInferenceServiceConfig` from `llminferenceserviceconfig.kserve-webhook-server.{v1alpha1,v1alpha2}.validator` to `llmisvcconfig.*`.

Renaming a webhook entry is a breaking change for any cluster with LLMInferenceService already deployed. When the webhook name changes, the old entry is not automatically removed during an upgrade. With OLM-managed operators in particular, the old `ValidatingWebhookConfiguration` entry can persist alongside the new one, causing every admission request for `LLMInferenceServiceConfig` to be evaluated twice - once by each webhook.

The 63-char label value limit that motivated #5381 is an OLM-side concern: OLM should sanitize (e.g. via hashing) the label values it derives from webhook names, not require upstream to rename the webhook itself. Fixing it by shortening the name just moves the problem while introducing a silent breaking upgrade.

- [ ] Verified webhook names are restored to their original values across all manifests and Go kubebuilder markers

The 63-char label limit is and the fix belongs on the OLM side (label sanitization/hashing), not here.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/kserve-llmisvc-resources/files/llmisvc/resources.yaml
+++ b/charts/kserve-llmisvc-resources/files/llmisvc/resources.yaml
@@ -1928,7 +1928,7 @@ webhooks:
       path: /validate-serving-kserve-io-v1alpha2-llminferenceserviceconfig
   failurePolicy: Fail
   matchPolicy: Exact
-  name: llmisvcconfig.kserve-webhook-server.v1alpha2.validator
+  name: llminferenceserviceconfig.kserve-webhook-server.v1alpha2.validator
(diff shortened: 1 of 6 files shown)


pierDipi (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5402, opened 2026-04-15
build: extend GOTAGS pass-through to the main controller Dockerfile

bartoszmajsak:

When the GOTAGS mechanism was introduced for the llmisvc controller in #5217, the main controller Dockerfile and its `docker-build` Makefile target were overlooked. Without `ARG GOTAGS` and `-tags` in the builder stage, any `//go:build` gated code is silently excluded from the main controller binary regardless of what `GOTAGS` is set to externally.

Extends the same pattern already in place for `llmisvc-controller.Dockerfile` so both controller builds handle build-tag customisation consistently.

Fixes #

- [x] `go build -tags distro ./cmd/manager` compiles correctly with the updated Dockerfile ARG wired through

Companion to #5217 - one-line oversight in `Dockerfile` and the `docker-build` target.

**Checklist**:

**Release note**:
```release-note
NONE
```


spolti (approved):
/lgtm

PR #5405, opened 2026-04-15
fix(llmisvc): extend storage migration retry window

bartoszmajsak:

Storage version migration could exhaust its retry budget during a rolling upgrade before the old pod was removed from Service Endpoints, leaving the controller in CrashLoopBackOff. Webhook calls hitting the old pod returned transient errors with no way out under the original fixed window.

Replaces the single exponential backoff (~150s worst case per resource group) with a two-phase strategy - fast backoff for quick initial detection, then steady-state polling for the remainder of a 1-hour configurable budget. Adds `--storage-migration-timeout` and `--storage-migration-poll-interval` flags for tuning.

Port of https://github.com/opendatahub-io/kserve/pull/1412

- [x] Unit tests for the two-phase retry state machine (phase fallthrough, fatal error short-circuit, timeout/cancellation error classification)
- [x] Existing e2e storage version migration test (default timeout updated to 900s to account for the new steady-state phase)

**Checklist**:

**Release note**:
```release-note
Extends storage version migration retry strategy to survive rolling upgrades.
The migration now retries for up to 1 hour (configurable) instead of a fixed
exponential backoff window (~150s), preventing CrashLoopBackOff when old pods
linger in Service Endpoints during upgrades.
```

### DIFF
--- a/cmd/llmisvc/main.go
+++ b/cmd/llmisvc/main.go
@@ -28,12 +28,10 @@ import (
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	apixclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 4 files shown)


cmd/llmisvc/main.go
```diff
@@ -275,37 +291,36 @@ func main() {
...
 	// that aren't serving yet.
-	// migrationBackoff allows enough time for Service endpoints to propagate
-	// after the webhook server starts.
-	migrationBackoff := wait.Backoff{
-		Duration: 2 * time.Second,
-		Factor:   1.5,
-		Jitter:   0.1,
-		Steps:    10,
-	}
+	//
```
spolti:
```suggestion
```

spolti (approved):
/lgtm

PR #5410, opened 2026-04-16
feat(llmisvc): add SectionName support to Gateway refs

bartoszmajsak:

Operators sharing a single Gateway across multiple services need a way to bind an `LLMInferenceService` to a specific listener rather than all listeners. Without this, every generated HTTPRoute attaches to every listener on the referenced Gateway, which can expose services on unintended ports or protocols (e.g. HTTP when only HTTPS was intended).

Introduces a dedicated `GatewayObjectReference` type that adds an optional `SectionName` field to gateway refs, targeting a specific named listener. `IngressSpec.Refs` stays on the generic `UntypedObjectReference` so the new field cannot be set where it has no meaning. When `SectionName` is omitted, existing behaviour is preserved - the route attaches to all compatible listeners.

The conflict-detection validator is updated to include `SectionName` in its equality key, so two refs pointing at the same gateway but different listeners are correctly treated as distinct. When the named listener does not exist, the Gateway API controller surfaces its standard `NoMatchingParent` condition, which propagates to the `LLMInferenceService` status without additional mapping.

- [x] Unit tests for `ToParentRefs` with and without `SectionName`
- [x] Validator tests covering matching/mismatching section names and asymmetric cases
- [x] Integration tests verifying `SectionName` propagates to the managed HTTPRoute `ParentReference`, and that omitting it preserves backward-compatible nil
- [x] E2e pytest tests for both the happy path and the backward-compat path

**Checklist**:

**Release note**:
```release-note
Adds optional `sectionName` field to `LLMInferenceService` gateway refs, enabling HTTPRoutes to target a specific named listener on a shared Gateway.
```


pierDipi (approved):
Great to see we can now have segmented gateways in a very easy to use manner!

/lgtm
/approve

PR #5411, opened 2026-04-16
fix(samples): add active-request-scorer and fix pluginRef typo

dagrayvid:
This PR updates the samples in docs/samples/llmisvc which use the llm-d prefix-aware-scorer routing configuration to add the active-requests scorer.
- Add active-request-scorer to prefix cache routing examples to improve load distribution.
- Fix pluginRef "prefix-cache-scorer" → "precise-prefix-cache-scorer" which would prevent scheduler startup.

This is needed because in some cases the default scorer/weight configuration with the prefix-cache-scorer can lead to all requests being routed to one Pod, and not balanced evenly. For example, a benchmark which ramps up traffic slowly where all requests share a common prefix / system prompt. The router scorer will dominate routing decisions and continue sending all requests to the Pod where the prefix is cached until load is high enough to cause queuing. This can be avoided by using the active-requests-scorer. 

[See this llm-d-inference-scheduler PR](https://github.com/llm-d/llm-d-inference-scheduler/pull/669) for more details on how this works and the benchmarks I ran to test this.

**Checklist**:
- [x] Have you linked the JIRA issue(s) to this PR?

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/samples/llmisvc/e2e-gpt-oss/README.md
+++ b/docs/samples/llmisvc/e2e-gpt-oss/README.md
@@ -200,7 +200,7 @@ kubectl apply -f llmisvc_config_prefix_cache.yaml -n kserve-lab
 This creates `llmisvc-prefix-caching`. It adds:
 
 - vLLM args: `--prefix-caching-hash-algo sha256_cbor`, `--block-size 64`, `--kv_transfer_config`, `--kv-events-config` (ZMQ to EPP)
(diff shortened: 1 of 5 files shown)


pierDipi (approved):
/lgtm
/approve

there is a PR title/description CI check failing

spolti:
Hi, can you please update the description to remove the issue tracker from there?

spolti (approved):

PR #5414, opened 2026-04-17
feat(llmisvc): report observed workload references in status

bartoszmajsak:

Operators had no way to discover which Deployment, LeaderWorkerSet, or Service backs an LLMInferenceService without knowing internal naming conventions or listing resources by label. The condition `MainWorkloadReady=True` gives no indication of which workload it refers to, whether the topology is single-node or multi-node, or what the workload Service name is.

Adds `status.workloads` with typed references (`TypedLocalObjectReference`) for the primary workload, prefill workload, workload service, and scheduler deployment. References are populated after a successful reconcile and cleared when the service is force-stopped, giving operators and tooling a stable way to navigate directly to the backing resources.

Example output:

```yaml
status:
  workloads:
    primary:
      apiGroup: apps
      kind: Deployment
      name: my-llm-kserve
    service:
      kind: Service
      name: my-llm-kserve-workload-svc
    scheduler:
      apiGroup: apps
      kind: Deployment
      name: my-llm-kserve-router-scheduler
```

- [x] Unit tests covering all topology variants (single-node, multi-node, P/D disaggregation, scheduler, force-stop, idempotency)
- [x] Integration tests (envtest) extending existing single-node, multi-node, stop, and scheduler test suites with `status.workloads` assertions
- [x] DeepEqual idempotency verified at both unit and integration level (no spurious status writes on no-op reconcile)

References are built from deterministic naming functions (the same ones the reconciler uses to create the resources) rather than `r.Get` calls. This avoids extra API calls and ensures `status.workloads` is populated on the same reconcile pass as conditions.

Also consolidates scattered inline name derivations into named functions and fixes test files that used raw string concatenation instead of `kmeta.ChildName`.

**Checklist**:

```release-note
Adds status.workloads to LLMInferenceService with typed references to backing Deployments, LeaderWorkerSets, and Services, enabling operators to discover workload resources directly from status.
```


pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -439,6 +447,67 @@ func (r *LLMISVCReconciler) enqueueOnHttpRouteChange(logger logr.Logger) handler
 	})
 }
 
+// enqueueOnInferencePoolChange creates an event handler that triggers reconciliation of
+// LLMInferenceServices that reference an external InferencePool via scheduler.pool.ref.
+// Managed pools are already covered by Owns(...) watches.
+func (r *LLMISVCReconciler) enqueueOnInferencePoolChange(logger logr.Logger) handler.EventHandler {
```
bartoszmajsak, line 531: Overlaps with #5417, as both are necessary. I will take care of potential merge conflicts whichever comes first.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -439,6 +447,67 @@ func (r *LLMISVCReconciler) enqueueOnHttpRouteChange(logger logr.Logger) handler
...
+		cfg, err := LoadConfig(ctx, r.Clientset)
+		if err != nil {
+			logger.Error(err, "Failed to load config")
+			return reqs
+		}
+
+		// When an InferencePool is modified, we need to find all LLMInferenceService instances
+		// that might depend on it and trigger their reconciliation.
+		// Use pagination to handle large numbers of services efficiently.
+		continueToken := ""
```
cjohannsen-cloudera, line 548: Given the potential for large numbers of services we need some telemetry metrics to track performance.

bartoszmajsak: Overall I think we don't need pagination here, as `r.List` goes through the `controller-runtime` cache. I can clean it up as a follow-up for each enqueue function (copy-pastie pattern I blindly followed).

As for the custom metrics - that's reasonable and I am happy to work on this. Is there anything more than we need besides what's already provided [1]? Should we open an issue so we can track this work?

[1] https://book.kubebuilder.io/reference/metrics-reference.html

cjohannsen-cloudera: Those defaults look fantastic, can you confirm that they are showing up with values populated when the code executes and add a screenshot to the PR description?  

Note: We should surface these metrics in the documentation (that can be another ticket not part of this).

pierDipi: Agree on making a pass to each `enqueue*` to remove pagination as a follow up

pierDipi (approved):

pkg/controller/v1alpha2/llmisvc/workload_status.go
```diff
@@ -0,0 +1,97 @@
...
+	if utils.GetForceStopRuntime(llmSvc) {
+		llmSvc.Status.Workloads = nil
+		return
+	}
+
+	ws := &v1alpha2.WorkloadStatus{}
+
+	if llmSvc.Spec.Worker != nil {
+		ws.Primary = &corev1.TypedLocalObjectReference{
+			APIGroup: ptr.To("leaderworkerset.x-k8s.io"),
```
spolti: should we move it to a constant?

spolti: also the other types below.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -494,6 +494,32 @@ type GatewayObjectReference struct {
...
+// WorkloadStatus records the workload resources observed during the last
+// successful reconciliation. Nil when no workload resources have been
+// created yet, or when the service is stopped.
+// +optional
+type WorkloadStatus struct {
+	// Primary is the main inference workload (Deployment or LeaderWorkerSet).
+	// When disaggregated serving is configured, this workload handles
+	// the decode phase; otherwise it handles both prefill and decode.
+	// +optional
+	Primary *corev1.TypedLocalObjectReference `json:"primary,omitempty"`
```
spolti, line 553: corev1.TypedLocalObjectReference only includes three fields, shouldn't we have more information about the status of each component?

spolti: As the objective is discovery, maybe renaming?

bartoszmajsak: What extra information would be helpful? 

FWIW we do have workload-related top-level conditions, these are intended as exact references for the involved objects.

pkg/controller/v1alpha2/llmisvc/workload_status.go
bartoszmajsak: Good idea!

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
spolti: my point is, the TypedLocalObjectReference, holds three fields only, mainly used for resource discovery, not to hold status.

spolti: non blocking, just my 2 cents, when Ive read it as status, I was expecting to see things like if the resource status is ready, if not, the reason.

spolti:
minor no blocking comments.

/lgtm

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
pierDipi: the status is already part of the conditions like "WorkloadsReady", etc

bartoszmajsak: > my point is, the TypedLocalObjectReference, holds three fields only, mainly used for resource discovery, not to hold status.

I understand this, however, as both me and @pierDipi pointed out, we already surface workload-related status through top-level conditions, for example we buble up deployment errors:

https://github.com/kserve/kserve/blob/5e056acb66da90884ed09bf2a33ea3acf11c1513/pkg/controller/v1alpha2/llmisvc/workload_single_node.go#L297

I wonder what other info you would find useful?

spolti: I get that, my point is, why name it `WorkloadStatus` since the purpose of it is not to propagate the resource status itself?

pierDipi: The idea is that WorkloadSpec is the spec type while WorkloadStatus is the status type, it's not indicating the status as in status conditions, small nuance but worth pointing out that status is not only conditions

pkg/controller/v1alpha2/llmisvc/controller.go
bartoszmajsak: FYI, here's the follow-up PR cleaning up pagination noise #5584

PR #5415, opened 2026-04-17
fix(llmisvc): fix long-name HTTPRoute BackendRefs

cyon13:

This PR fixes backendRef name normalization for LLMInferenceService router configuration when the LLMInferenceService name is long.

Previously, only default `InferencePool` backendRefs were normalized. However, workload `Service` backendRefs from templates or custom route configs could still use the plain concatenated name:

```text
<llmInferenceServiceName>-kserve-workload-svc
```

When the combined name exceeds the Kubernetes 63-character DNS label limit, the actual workload Service name is generated using `kmeta.ChildName()`, which applies truncation and hashing. In that case, the HTTPRoute backendRef could point to a non-existent Service.

This PR updates the reconciliation logic to:

- Normalize workload `Service` backendRefs using `kmeta.ChildName()`
- Match both hashed/truncated and plain concatenated workload Service names
- Match both default and plain concatenated default `InferencePool` names
- Apply workload `Service` backendRef normalization across scheduler-disabled, external scheduler pool ref, and managed scheduler pool configurations

This ensures that HTTPRoute backendRefs correctly reference the generated workload Service or configured scheduler pool for long LLMInferenceService names.

Fixes #

- [x] Added unit tests for `isDefaultBackendRef`
- [x] Verified that default `InferencePool` backendRefs match both `kmeta.ChildName()`-generated names and plain concatenated names
- [x] Verified that non-`InferencePool` backendRefs do not match `isDefaultBackendRef`
- [x] Added unit tests for `isDefaultWorkloadServiceBackendRef`
- [x] Verified that workload `Service` backendRefs match both `kmeta.ChildName()`-generated names and plain concatenated names
- [x] Verified that `Service` backendRefs with nil `Kind` are treated as default Kubernetes `Service` backendRefs
- [x] Verified that only core Kubernetes `Service` backendRefs with the expected workload Service name are matched
- [x] Added a long-name test to confirm that plain concatenation can exceed the 63-character Kubernetes name limit while `kmeta.ChildName()` produces a valid truncated/hash-based name

**Logs**:

```text
go test ./pkg/controller/v1alpha2/llminferenceservice/...
```

This change is intended to prevent `HTTPRoute` backendRefs from pointing to a non-existent workload `Service` when an `LLMInferenceService` has a long name.

For long names, the generated workload `Service` uses `kmeta.ChildName()` to stay within the Kubernetes 63-character name limit, while some `HTTPRoute` backendRefs may still use the plain concatenated name. This can cause a mismatch between the actual `Service` name and the `HTTPRoute` backendRef name.

This PR normalizes workload `Service` backendRefs to the same `kmeta.ChildName()`-based name used by the generated `Service`.

**Checklist**:

- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)? Not required for this bug fix

**Release note**:

```release-note
Fixed a mismatch between generated workload Service names and HTTPRoute backendRefs for LLMInferenceService resources with long names.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_merge.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_merge.go
@@ -272,10 +272,15 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
 		llmSvcCfg.Spec.Router.Scheduler == nil {
 		for i := range llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules {
 			for j := range llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs {
-				if isDefaultBackendRef(llmSvc, llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs[j].BackendRef) {
-					llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs[j].Group = ptr.To[gwapiv1.Group]("")
-					llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs[j].Kind = ptr.To[gwapiv1.Kind]("Service")
-					llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs[j].Name = gwapiv1.ObjectName(kmeta.ChildName(llmSvc.GetName(), "-kserve-workload-svc"))
+				ref := &llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules[i].BackendRefs[j]
+				if isDefaultBackendRef(llmSvc, ref.BackendRef) {
+					ref.Group = ptr.To[gwapiv1.Group]("")
+					ref.Kind = ptr.To[gwapiv1.Kind]("Service")
+					ref.Name = gwapiv1.ObjectName(kmeta.ChildName(llmSvc.GetName(), "-kserve-workload-svc"))
+				} else if isDefaultWorkloadServiceBackendRef(llmSvc, ref.BackendRef) {
+					// Ensure the workload service name uses ChildName for proper hash truncation
+					// when the combined name exceeds K8s 63-character limit.
+					ref.Name = gwapiv1.ObjectName(kmeta.ChildName(llmSvc.GetName(), "-kserve-workload-svc"))
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


cyon13:
Here is an example of the mismatch this PR is intended to fix.

For example, suppose the submitted `LLMInferenceService` includes the following router configuration:

```yaml
router:
  route:
    http:
      spec:
        hostnames:
        - "gw.model01.aa.com"
        parentRefs:
          - name: agentgateway-proxy
            namespace: agentgateway-system
        rules:
          - matches:
              - path:
                  type: PathPrefix
                  value: /ame/my-very-long-llm-inference-service-name-that-exceeds
            timeouts:
              backendRequest: 3600s
            filters:
              - type: URLRewrite
                urlRewrite:
                  path:
                    type: ReplacePrefixMatch
                    replacePrefixMatch: /
            backendRefs:
              - group: ""
                kind: Service
                name: my-very-long-llm-inference-service-name-that-exceeds-kserve-workload-svc
                port: 8000
```

For a long `LLMInferenceService` name, the generated workload `Service` name can be truncated/hashed, for example:

```text
my-very-lon43c2126c275981fbd479ad89f606c2c2-kserve-workload-svc
```

However, before this change, the `HTTPRoute` backendRef could still use the plain concatenated name:

```yaml
spec:
  hostnames:
  - gw.model01.aa.com
  parentRefs:
    kind: Gateway
    name: agentgateway-proxy
    namespace: agentgateway-system
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: my-very-long-llm-inference-service-name-that-exceeds-kserve-workload-svc
      port: 8000
      weight: 1
```

In this case, the generated `Service` name and the `HTTPRoute` backendRef name do not match, so the `HTTPRoute` can end up pointing to a `Service` that does not actually exist.

With this change, the workload `Service` backendRef is normalized to the generated `Service` name, so the `HTTPRoute` will instead become:

```yaml
spec:
  hostnames:
  - gw.model01.aa.com
  parentRefs:
    kind: Gateway
    name: agentgateway-proxy
    namespace: agentgateway-system
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: my-very-lon43c2126c275981fbd479ad89f606c2c2-kserve-workload-svc
      port: 8000
      weight: 1
```

pierDipi:
@cyon13 if the user is providing a configuration (for whatever reason) that cannot be true, meaning a name that is not valid, I think it should be a validation error and not a transformation.

maybe I'm missing something but in your example  `name: my-very-long-llm-inference-service-name-that-exceeds-kserve-workload-svc` this should be a validation error and not being magically transformed name

cyon13:
@pierDipi Thanks for the feedback. I completely agree with your point that if a user explicitly provides an invalid Service name in `backendRefs`, KServe should return a validation error rather than silently transforming it.

However, I’d like to clarify that this PR addresses a different issue: an internal naming mismatch between the actual workload Service generated for an `LLMInferenceService` and the generated `HTTPRoute`.

## The problem: internal inconsistency

When an `LLMInferenceService` is created, KServe automatically generates a workload Service and an `HTTPRoute` that points to it. To comply with the 63-character DNS label limit, KServe uses `kmeta.ChildName()` to truncate and hash the Service name when the generated workload Service name would be too long.

The issue is that while the actual Service is created using `kmeta.ChildName()`, the `HTTPRoute` `backendRef` is currently rendered using plain string concatenation. This can lead to a broken reference.

For example:

- LLMInferenceService name:
  `my-very-long-llm-inference-service-name-that-exceeds`

- Actual Service created:
  `my-very-long-llm-inference-service-na-abc123-kserve-workload-svc`
  via `kmeta.ChildName()`

- Current broken `backendRef`:
  `my-very-long-llm-inference-service-name-that-exceeds-kserve-workload-svc`
  via plain string concatenation

Because these two names do not match, the generated `HTTPRoute` points to a Service that KServe did not actually create.

## The fix

This PR makes the `HTTPRoute` controller use the same deterministic naming logic, `kmeta.ChildName()`, that KServe already uses when creating the workload Service.

So this is not about rewriting arbitrary user-provided names. It is about making the references between KServe-managed resources consistent. Without this fix, when the workload Service name is shortened by `kmeta.ChildName()`, the generated `HTTPRoute` can point to a Service that does not exist. As a result, inference requests routed through that `HTTPRoute` fail because there is no valid backend Service to receive the traffic.

PR #5417, opened 2026-04-17
feat(llmisvc): report observed routing topology in status

bartoszmajsak:

Operators verifying traffic shifts had to inspect child resources (Gateways, HTTPRoutes, InferencePools) individually because status only reported boolean conditions and discovered URLs. Event handlers re-derived the routing graph on every Gateway/HTTPRoute event by calling `combineBaseRefsConfig` per service - O(n services x m configs) in API server calls.

Adds `status.router` with structured references to the networking resources the controller observed during reconciliation. Gateways carry their matched listeners and bound HTTPRoutes, giving operators a single place to verify the routing topology. InferencePool and EPP Service refs are populated under `scheduler` when the scheduler is enabled.

Example output:

```yaml
status:
  router:
    gateways:
      kind: Gateway
      name: kserve-gateway
      namespace: kserve
      listeners:
      - https
      httpRoutes:
        kind: HTTPRoute
        name: my-llm-kserve-route
        namespace: my-namespace
    scheduler:
      inferencePool:
        group: inference.networking.k8s.io
        kind: InferencePool
        name: my-llm-inference-pool
      service:
        kind: Service
        name: my-llm-kserve-epp
```

Refactors `DiscoverURLs` to accept pre-fetched gateways, eliminating duplicate gateway fetches per reconcile. Event handlers now check `status.router` in-memory before falling back to the expensive config derivation path for services that predate this change.

- [x] 10 table-driven unit tests for `BuildObservedGateways` covering deduplication, sorting, nil-vs-empty slice semantics, multi-gateway aggregation
- [x] 4 envtest integration tests: managed route with gateway, force-stop clearing, no-router nil, DeepEqual stability
- [x] Full test suite passes (176 specs, 353s)

Uses `gwapiv1.ObjectReference` (from Gateway API) for all resource refs - not `corev1.ObjectReference` which carries noise fields (`UID`, `ResourceVersion`, `FieldPath`) that Kubernetes conventions discourage in new APIs.

`ObservedGateway` is the only new composite type - it embeds `gwapiv1.ObjectReference` inline and adds `Listeners` (gateway listener names) and `HTTPRoutes` (routes bound through that gateway). This nesting makes the gateway-to-route association explicit at a glance.

The event handler fast path includes a fallback for services that predate this change (`status.router == nil` falls back to `combineBaseRefsConfig`). A follow-on story can add field indexers to eliminate the full list operation entirely.

On transient errors (e.g. API server blip fetching a gateway), `status.router` from the last successful reconcile is preserved rather than zeroed - the enqueue handlers keep their fast-path lookup until the error clears.

**Checklist**:

**Release note**:
```release-note
Adds `status.router` to LLMInferenceService reporting the observed networking topology - gateways with matched listeners and bound HTTPRoutes, plus scheduler-observed InferencePool and EPP Service references. Eliminates expensive per-service config derivation in gateway/route event handlers.
```


pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -277,14 +280,17 @@ func (r *LLMISVCReconciler) expectedHTTPRoute(ctx context.Context, llmSvc *v1alp
 	return httpRoute
 }
 
-func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, routes ...*gwapiv1.HTTPRoute) error {
+func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, routes ...*gwapiv1.HTTPRoute) ([]ResolvedGateway, error) {
 	logger := log.FromContext(ctx)
 
+	// Zero out routing status on every entry to avoid stale state on early returns.
+	llmSvc.Status.Routing = nil
```
cjohannsen-cloudera, line 287: On early return the routing is always left nil, should the previous routing configuration be preserved and reused in this situation?

bartoszmajsak: Thanks for bringing it up. It was probably too broad and would allow any transient error to wipe the status which is probably not what we need. Moved it to two relevant paths to make it explicit.

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -661,3 +718,97 @@ func (r *LLMISVCReconciler) evaluateSingleInferencePoolCondition(ctx context.Con
...
+			}
+			a := acc[key]
+
+			if parentRef.SectionName != nil {
+				a.listeners = append(a.listeners, *parentRef.SectionName)
+			}
+
+			alreadyAdded := false
+			for _, existing := range a.routes {
+				if existing.Name == ref.Name {
```
cjohannsen-cloudera, line 764: Do we need to also consider namespace when deduplicating?

bartoszmajsak: These are routes attached to the given gateway (`a` is the value of `acc[gwNS + "/" + gwName`]), so I hope we are fine with only names here.

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
pierDipi, line 1: I'm wondering if we should match the spec structure and naming pattern (router vs routing) in the status too:
```yaml
status:
  router:
    gateways: ..
    scheduler:
      service: ...
```

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -500,6 +500,46 @@ type GatewayObjectReference struct {
...
+
+	// InferencePool is the InferencePool observed as active for this service.
+	// Nil when the scheduler is not configured.
+	// +optional
+	InferencePool *gwapiv1.ObjectReference `json:"inferencePool,omitempty"`
+
+	// SchedulerService is the EPP Service observed for this service.
+	// Nil when the scheduler is not configured.
+	// +optional
+	SchedulerService *gwapiv1.ObjectReference `json:"schedulerService,omitempty"`
```
pierDipi: I expect scheduler to grow over time so maybe a full struct will set this up for the future

pkg/apis/serving/v1alpha1/llm_inference_service_types.go
```diff
@@ -500,6 +500,46 @@ type GatewayObjectReference struct {
...
+	// Gateways lists the Gateway resources observed as attached to this service,
+	// each with the listeners and HTTPRoutes bound through them.
+	// +optional
+	// +listType=atomic
+	Gateways []ObservedGateway `json:"gateways,omitempty"`
+
+	// InferencePool is the InferencePool observed as active for this service.
+	// Nil when the scheduler is not configured.
+	// +optional
+	InferencePool *gwapiv1.ObjectReference `json:"inferencePool,omitempty"`
```
pierDipi: this could be within `scheduler` ?

bartoszmajsak:
Thanks for the suggestions @pierDipi - I aligned routing to mirror `.spec.router`, PTAL when you find some time.

PR #5418, opened 2026-04-17
feat(llmisvc): observed applied configs

bartoszmajsak:

When an `LLMInferenceService` reconciles, multiple `LLMInferenceServiceConfig`
resources get merged together - well-known defaults for the deployment pattern
plus any explicit baseRefs. Today, there's no way to see which configs actually
contributed to the result. Debugging precedence issues means mentally replaying
the merge logic, which is error-prone and slow.

This PR makes config precedence observable by adding `status.appliedConfigs` -
an ordered list of config refs tagged by source (`Preset` or `UserRef`),
each carrying the resolved name and namespace. After reconciliation, users and
tooling can inspect the status to see exactly which configs were applied, from
which namespace, and in what order.

### Changes

- **New types** - `AppliedConfigSource` enum and `AppliedConfigRef` struct using
  `gwapiv1.ObjectName` and `gwapiv1.Namespace` for inherited validation
  (maxLength, minLength, RFC 1123 pattern), matching how `UntypedObjectReference`
  and `GatewayObjectReference` already work in this codebase.
- **CombinedConfig wrapper** - `combineBaseRefsConfig` returns a `CombinedConfig`
  (merged config + applied refs) instead of a bare config pointer. Applied refs
  are populated with the resolved namespace from `getConfig`.
- **Status population** - `reconcileBaseRefs` writes `Status.AppliedConfigRefs`
  only on successful reconciliation. On force-stop with error, the field is
  cleared to reflect that no configs are actively applied.
- **Namespace-aware config lookup** - `IsUsingLLMInferenceServiceConfigInNamespace`
  checks `AppliedConfigRefs` first (authoritative when present), falling back to
  annotations/baseRefs only when the field is empty (new service or stopped).
- **Smarter config-change enqueue** - the watch handler uses name-only matching
  via `IsUsingLLMInferenceServiceConfig` since `getConfig` resolves namespace
  priority at reconcile time, not enqueue time. Cross-tenant isolation is handled
  by the list-namespace scoping.

- [x] Unit tests (`llm_inference_service_types_func_test.go`): name match,
  namespace match/mismatch, substring rejection, short-circuit when
  appliedConfigs is populated, cold-start fallback to annotations and baseRefs.
- [x] Reconcile test (`config_merge_reconcile_test.go`): verifies that a
  `combineBaseRefsConfig` error does not panic and does not populate
  `AppliedConfigRefs` on the error path.
- [x] Integration test (`controller_int_test.go`): creates a service with an
  explicit `baseRef`, verifies well-known configs appear before BaseRef entries
  in the correct precedence order, and checks namespace is populated on all entries.

- `AppliedConfigRef` uses `gwapiv1.ObjectName` and `gwapiv1.Namespace` types
  for inherited CRD validation, consistent with existing reference types in the
  codebase (`UntypedObjectReference`, `GatewayObjectReference`).
- `+listType=atomic` is intentional - the controller always writes the full list
  and ordering encodes merge precedence.
- The config-change watch handler uses name-only matching (`IsUsingLLMInferenceServiceConfig`)
  rather than namespace-aware matching. This is deliberate: `getConfig` resolves
  namespace priority at reconcile time (local ns first, then kserve ns), so a
  local config that shadows a global one must trigger re-reconciliation.

## Checklist

```release-note
Add status.appliedConfigs to LLMInferenceService to track which LLMInferenceServiceConfig resources were applied during reconciliation, tagged by source (Preset/UserRef) with resolved namespace.
```


pkg/controller/v1alpha2/llmisvc/controller_int_test.go
```diff
@@ -164,6 +164,72 @@ var _ = Describe("LLMInferenceService Controller", func() {
 			})).WithContext(ctx).Should(Succeed())
 		})
 
+		It("should populate status.appliedConfigs with well-known and explicit configs", func(ctx SpecContext) {
```
VedantMahabaleshwarkar: maybe a test case to cover llmisvc.status.appliedConfigRefs is cleared in case of error in reconciliation?

bartoszmajsak: Good idea, moved all net-new status tests to pkg/controller/v1alpha2/llmisvc/controller_int_status_test.go and added this case.

VedantMahabaleshwarkar:
lgtm

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -124,16 +133,18 @@ func WithSkipClearSchedulerConfigRef() CombineOption {
...
 
-			baseCfg = &v1alpha2.LLMInferenceServiceConfig{
+			return &v1alpha2.LLMInferenceServiceConfig{
 				Spec: *llmSvc.Spec.DeepCopy(),
-			}
-			return baseCfg, nil
+			}, nil
 		}
+
+		llmSvc.Status.AppliedConfigRefs = nil
```
pierDipi, line 146: don't we want to clear it also when stopped?

bartoszmajsak: I was wondering about it too - my thinking was that when we stop, it would still be good to know the last applied config? On the other hand it may become stale. On the verge here.

Though I noticed docs our outdated in the enqueue function - either way need to push some changes. 

WDYT?

pierDipi: let's keep it

bartoszmajsak:
@sivanantha321 I would appreciate if we could move this forward :) it finally passed all the tests (we have a lot of flakes :()

PR #5427, opened 2026-04-20
fix: normalize legacy deploymentMode in status

jlost:

After the deploymentMode rename (RawDeployment -> Standard, Serverless -> Knative in #5025), ISVCs created on older versions retain the legacy value in `status.DeploymentMode`. The defaulting webhook normalizes the annotation to the new name, but `validateDeploymentMode` compares the normalized annotation against the un-normalized status and rejects updates. This also causes `GetDeploymentMode` to return the legacy value to the controller, which writes it back to status -- so the status never self-corrects.

This PR uses `ParseDeploymentMode` to normalize legacy status values in:
- `GetDeploymentMode` -- so the controller returns and writes canonical values
- `validateDeploymentMode` -- so validation passes during the window before the controller reconciles

Fixes a regression from #5025 where upgraded ISVCs with legacy `status.DeploymentMode` values become un-updatable.

- [x] Unit tests: `TestGetDeploymentMode` with `LegacyRawDeploymentInStatus` and `LegacyServerlessInStatus` cases
- [x] Unit tests: `TestDeploymentModeUpdate` with legacy status accepted for same-mode, rejected for cross-mode

**Checklist**:

```release-note
Fix legacy status.DeploymentMode values (RawDeployment/Serverless) not being normalized after upgrade, causing validation to reject updates on ISVCs created before the deployment mode rename.
```

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -582,7 +582,7 @@ func validateCollocationStorageURI(predictorSpec PredictorSpec) error {
 
 // validates if the deploymentMode specified in the annotation is not different from the one recorded in the status
 func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService) error {
-	statusDeploymentMode := oldIsvc.Status.DeploymentMode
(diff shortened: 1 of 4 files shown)


spolti (approved):
/lgtm

Jooho (approved):
/lgtm
/approved

PR #5432, opened 2026-04-21
fix(llmisvc): add port in kv-cache routing sample topic

pierDipi:
the precise prefix cache scorer was updated to identify routing endpoints using a combination of the IP address and port in scheduler 0.7.

Previous versions relied solely on the IP address.


Jooho (approved):
not a blocker.

README file also have the same part: https://github.com/kserve/kserve/blob/fc000125fdd95a839e61e22ef86a46cae3eae22b/docs/samples/llmisvc/precise-prefix-kv-cache-routing/README.md?plain=1#L71

PR #5433, opened 2026-04-21
feat(llmisvc): add migration logic for llm-d v0.6 components upgrade

vivekk16:
Adds version-gated controller migration logic to ensure seamless upgrades to llm-d v0.6 components. Existing deployments with old configs are transparently migrated during reconciliation; new deployments start with correct defaults.

## llm-d Component Audit

| Component | Upgrade | Breaking Changes | Migration Strategy |
|---|---|---|---|
| `llm-d-inference-scheduler` (EPP) | v0.6 → v0.7.1 | Plugin renames, parameter restructuring, deprecated field removal, 5 CLI metric flags hard-rejected | Version-gated (≥ 0.7.0) single-pass YAML + CLI migration |
| `llm-d-routing-sidecar` | v0.6 → v0.7.1 | `--connector` renamed to `--kv-connector` | Template update only (binary accepts both via internal `Complete()` migration) |
| `llm-d-uds-tokenizer` | v0.6 → v0.7.1 | None | Image bump only |
| `llm-d-cuda` (vLLM runtime) | v0.6.0 | No changes in this PR | N/A |

## EPP Scheduler Migrations

All migrations are applied in a single YAML unmarshal/marshal pass inside `schedulerTransform`, gated by the `app.kubernetes.io/version` annotation on the pod template:

- **Plugin renames**: `prefill-header-handler` → `disagg-headers-handler`, `pd-profile-handler` → `disagg-profile-handler` (+ `pluginRef` updates in scheduling profiles)
- **Parameter restructure**: `deciderPluginName` (string) → `deciders` map (`prefill: <decider-name>`)
- **Field removal**: Strip deprecated `hashBlockSize` from plugin parameters
- **Metric flag migration**: Extract 5 deprecated CLI flags (`total-queued-requests-metric`, `total-running-requests-metric`, `kv-cache-usage-percentage-metric`, `lora-info-metric`, `cache-info-metric`) and inject values into `core-metrics-extractor` plugin `engineConfigs`. These flags are **hard-rejected** by GIE v1.4.0 `Validate()` — the binary will not start if they are explicitly set.

## Template & Default Updates (new deployments)

- `config-llm-scheduler.yaml`: Added `app.kubernetes.io/version: "0.7.0"` annotation so new deployments skip migrations
- `config-llm-decode-template.yaml` / `config-llm-decode-worker-data-parallel.yaml`: `--connector=nixlv2` → `--kv-connector=nixlv2`
- `schedulerConfigText()`: Updated P/D default with new plugin names and correct `deciders: {prefill: always-disagg-pd-decider}` format

## Version Gating Design

- **New deployments**: Start with `0.7.0` annotation from default config → all v0.7 migrations skipped
- **Existing deployments**: No annotation → defaults to `0.0.0` → `propagateSchedulerMetadata` sets `0.7.0` → migrations applied on first reconcile
- **v0.6 binaries**: Version < 0.7.0 → all migrations skipped (safe)

## Files Changed

| File | Change |
|---|---|
| `scheduler.go` | Migration functions, version gating, updated defaults |
| `scheduler_test.go` | Unit tests for each migration, version gating, full pipeline |
| `config-llm-scheduler.yaml` | Version annotation |
| `config-llm-decode-template.yaml` | `--kv-connector` |
| `config-llm-decode-worker-data-parallel.yaml` | `--kv-connector` |
| `config_presets_test.go` | Updated assertion for `--kv-connector` |
| `controller_int_scheduler_config_test.go` | Updated plugin name assertions |
| `go.mod` / `go.sum` | Added `github.com/coreos/go-semver` for version parsing |


pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -478,16 +510,17 @@ func schedulerConfigText(llmSvc *v1alpha2.LLMInferenceService) string {
 apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
 plugins:
-- type: prefill-header-handler
+- type: disagg-headers-handler
```
pierDipi, line 491: if we don't gate this update, it will restart components on Kserve upgrade

vivekk16: The `schedulerConfigText()` default is guarded. It's never injected into existing deployments on upgrade.
 
Here's the protection chain:

`preserveSchedulerConfig` has 4-level priority:

1. Explicit inline config -> wins
2. `--config-text` in the template args -> kept as-is
3. `--config-text` from the current Deployment on the cluster -> preserved across upgrades
4. Fresh default from `schedulerConfigText()` -> only if 1-3 don't apply

For an existing v0.6 deployment being upgraded: the merged template from config-llm-scheduler.yaml doesn't contain --config-text in its args (it only has --pool-name, --pool-namespace, etc.), so priority 2 is skipped. Priority 3 then picks up the existing --config-text from the live Deployment on the cluster, preserving the old v0.6 plugin names.

After preservation, the version-gated migration (via `schedulerTransform`) renames the plugins only if `app.kubernetes.io/version >= 0.7.0.` The annotation comes from the merged default config, so the migration fires and transforms the preserved config in-place.

The pod will restart on upgrade, but that's because the image changes from v0.6 to v0.7.1 (unavoidable for a binary upgrade), not because of an unguarded config default. The config and binary version stay in sync within the same reconcile.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -438,16 +443,43 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
...
-	}
 
-	r.propagateSchedulerMetadata(llmSvc, d)
+		// Pre-v0.7.0 schema migrations: promote fields that moved in v0.6.0.
+		// These are safe to run unconditionally (idempotent, understood by both v0.6 and v0.7 binaries).
+		if err := mutateSchedulerConfig(ctx, d,
+			WithMigrateTokenProcessorConfig,
+			WithMigrateBlockSizeToBlockSizeTokens,
+		); err != nil {
+			return d, fmt.Errorf("failed to migrate scheduler config: %w", err)
```
pierDipi: this was previously done within `isUsingTokenizerSidecar` which gates on v0.6

pierDipi (approved):

PR #5438, opened 2026-04-22
fix(agent): migrate AWS SDK from v1 to v2

### LINKED ISSUE #3709: Update to AWS Go SDK v2
Kserve is using AWS Go SDK v1.  The current version is v2.

https://github.com/kserve/kserve/blob/690e269ac1aa88d6e15872164c966305cd518169/go.mod#L7

The AWS Go SDK team [announced](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/) in January that maintenance mode for v1 starts on July 31, 2024, followed by end-of-life on July 31, 2025.

There's still time, but it would be good to [migrate to v2](https://aws.github.io/aws-sdk-go-v2/docs/migrating/) before approaching either of these dates, IMHO.

Thanks.

spolti:
Replace aws-sdk-go v1 with aws-sdk-go-v2 using the transfermanager package for S3 operations.
Add S3 integration tests with local S3-compatible server (SeaweedFS/MinIO).

Replaces the deprecated ask-sdk v1 with v2.
Added real-world tests by running local S3 buckets to test the feature.

Fixes #3709

- [x] Updated the tests to the v2 sdk
- [x] Added the `s3_integration_test.go` which targets running s3 buckets, can be run with build flags.

**Checklist**:

**Release note**:
```release-note
Upgrade  AWS SDK from v1 to v2
```


pkg/agent/storage/s3.go
```diff
@@ -17,95 +17,59 @@ limitations under the License.
...
-	}
-	return nil
-}
+	bucket := tokens[0]
 
-func (s *S3ObjectDownloader) GetAllObjects(s3Svc s3iface.S3API) ([]s3manager.BatchDownloadObject, error) {
-	resp, err := s3Svc.ListObjects(&s3.ListObjectsInput{
-		Bucket: aws.String(s.Bucket),
-		Prefix: aws.String(s.Prefix),
+	resp, err := m.Client.ListObjectsV2(ctx, &s3.ListObjectsV2Input{
```
mwaykole: This only fetches the first page (up to 1000 objects). If a model directory has more than 1000 files, the rest get silently dropped. The old code had the same limitation, but since you're rewriting this anyway, it'd be worth switching to `s3.NewListObjectsV2Paginator` to iterate through all pages. Models with lots of shards or safetensors files can easily exceed 1000 keys.

pkg/agent/storage/s3.go
```diff
@@ -17,95 +17,59 @@ limitations under the License.
...
+		Key:    aws.String(key),
+		Body:   strings.NewReader(string(object)),
 	})
+	return err
 }
 
 func (m *S3Provider) DownloadModel(modelDir string, modelName string, storageUri string) error {
-	log.Info("Download model ", "modelName", modelName, "storageUri", storageUri, "modelDir", modelDir)
+	log.Info("Download model", "modelName", modelName, "storageUri", storageUri, "modelDir", modelDir)
+	ctx := context.Background()
```
mwaykole, line 53: Both `DownloadModel` and `UploadObject` create `context.Background()` internally, so callers have no way to cancel or set a deadline on long-running S3 operations. Would be nice to accept a `context.Context` parameter — especially for large multi-GB model downloads that might need to be interrupted during shutdown.

pkg/agent/storage/s3.go
```diff
@@ -114,55 +78,49 @@ func (s *S3ObjectDownloader) GetAllObjects(s3Svc s3iface.S3API) ([]s3manager.Bat
...
+		}
+
+		if _, err := m.TransferClient.DownloadObject(ctx, &transfermanager.DownloadObjectInput{
+			Key:      object.Key,
+			Bucket:   aws.String(bucket),
+			WriterAt: file,
+		}); err != nil {
+			_ = file.Close()
+			return fmt.Errorf("failed to download %s: %w", *object.Key, err)
 		}
```
mwaykole: If `file.Close()` fails here, the error is logged but the download is treated as successful. On network-mounted filesystems (NFS, EFS) a close failure can mean data wasn't flushed — the model file would be silently incomplete. Might be worth returning the error instead of just logging it.

pkg/agent/storage/s3.go
spolti: hey @mwaykole, for this comment, I opened https://github.com/kserve/kserve/issues/5844
this would take more changes that is out of scope of this specific PR.

pkg/agent/storage/s3.go
spolti: addressed

pkg/agent/storage/s3.go
spolti: adressed

PR #5451, opened 2026-04-27
feat(isvc): add dual-protocol (REST/gRPC) routing for Standard mode

### LINKED ISSUE #5377: Support dual-protocol (gRPC + HTTP) routing in Gateway API HTTPRoute for InferenceServices
/kind feature

## Support dual-protocol (gRPC + HTTP) routing in Gateway API HTTPRoute for InferenceServices

When an InferenceService serves both gRPC and HTTP REST using the V2 Inference Protocol (e.g., gRPC on port 8081 and HTTP REST on port 8080), the current Gateway API implementation (#3952) generates a single HTTPRoute that routes all traffic to the Service's default port (port 80, mapped to the primary container port).

This means:
- If the primary protocol is gRPC (h2c), the auto-generated Service maps port 80 → containerPort 8081. The auto-generated HTTPRoute sends all traffic (including HTTP REST requests) to port 80, which lands on the gRPC server.
- If the primary protocol is HTTP, gRPC traffic routed to the HTTP server similarly fails.

### Current Workaround
Users must manually:

- Use the serving.kserve.io/extraPorts annotation to add the secondary port to the generated Service:
   ```yaml
   serving.kserve.io/extraPorts: '[{"name":"http-rest","containerPort":8080,"protocol":"TCP"}]'
   ```
- Create a second HTTPRoute on the inference cluster to match specific paths and route them to the secondary port:
```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: <isvc-name>-http
spec:
  hostnames:
    - <generated-hostname>
  parentRefs:
    - kind: Gateway
      name: kserve-ingress-gateway
      namespace: istio-system
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /v2/
      backendRefs:
        - kind: Service
          name: <isvc-name>-predictor
          port: 8080
```
This workaround is hard to maintain — the manually created HTTPRoute must track the auto-generated hostname and Service name, and must be applied on every inference cluster.

### Proposal
KServe should natively support dual-protocol routing when Gateway API is enabled. There are several possible approaches:
- Option 1: Generate a single HTTPRoute with protocol-aware rules
- Option 2: Generate both HTTPRoute and GRPCRoute
- Option 3:  Support multi-port routing in HTTPRoute generation

sivanantha321:

Adds **dual-protocol (REST + gRPC) routing support for InferenceService “Standard mode” when using Gateway API.** It does this by detecting REST and gRPC service ports from the Kubernetes Service and then generating two HTTPRoute rules: a more specific gRPC match first, followed by a REST/HTTP fallback match, enabling both protocols behind the same hostname.

- Fixes #5377

**What changed:**
**Routing logic:**

-     Updates the HTTPRoute reconciler to query the backing Service and detect:
       1. A REST port (default/fallback)
       2. A gRPC port (based on appProtocol = kubernetes.io/h2c and/or port name heuristics like containing grpc/h2c)
-     When both ports exist, it generates two HTTPRoute rules (capacity increased from 1 → 2 in multiple places):
        1. gRPC rule first: matches gRPC traffic using path + content-type header regex (application/grpc.*)
        2. HTTP rule second: matches remaining HTTP/REST traffic (fallback)
-     When dual ports are not detected (or Service not found early), it falls back to prior behavior (single rule to the default HTTP port).

**Tests:**
-     Updates unit tests to account for function signature changes (now passing ctx and a client).
-     Adds an e2e scenario validating a single SKLearn InferenceService can be reached via both gRPC and REST when ports are exposed.

**Alternatives considered:**
**1. Create both a GRPCRoute and an HTTPRoute for the same host (dual Route kinds)**

   This was considered as a clean separation of protocols using dedicated Gateway API route types. But it **does not work if the host name is same for both routes.**

**Why it doesn’t work for same hostname**: per Gateway API docs, if an HTTPRoute (A) and GRPCRoute (B) are attached to the same Listener and their hostnames intersect (non-empty intersection), the implementation MUST accept exactly one of the two routes. That makes simultaneously attaching both route kinds for the same host non-viable, so this PR instead keeps everything within a single HTTPRoute and uses match rules to discriminate gRPC vs REST.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1032,6 +1032,13 @@ jobs:
         timeout-minutes: 30
         run: |
           ./test/scripts/gh-actions/run-e2e-tests.sh "raw" "6" ${{ matrix.network-layer }}
+      
+      # Run dual protocol (gRPC/REST) test only for gatewayapi as it is the only one supporting it currently.
+      - name: Run E2E tests - Dual Protocol
+        if: matrix.network-layer == 'envoy-gatewayapi' || matrix.network-layer == 'istio-gatewayapi'
+        timeout-minutes: 10
+        run: |
+          ./test/scripts/gh-actions/run-e2e-tests.sh "dual_protocol" "6" ${{ matrix.network-layer }}
 
       - name: Patch inferenceservice config for path based routing
         if: matrix.network-layer == 'envoy-gatewayapi' || matrix.network-layer == 'istio-gatewayapi'
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/httproute_reconciler.go
@@ -135,6 +135,88 @@ func addIsvcHeaders(name string, namespace string) gwapiv1.HTTPRouteFilter {
 	}
 }
 
+// detectServiceProtocolPorts queries the Service to extract protocol port information.
+// It analyzes the service ports to identify REST and gRPC ports based on appProtocol annotations
+// and port names. Only the first port of each type is detected and returned.
+// Returns the REST port, gRPC port, and any error encountered.
+func detectServiceProtocolPorts(ctx context.Context, client client.Client, serviceName, namespace string) (restPort, grpcPort int32, err error) {
+	// Query the Service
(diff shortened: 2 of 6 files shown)


greenmoon55 (approved):
LGTM

PR #5465, opened 2026-04-29
fix: `hostPath` for `kserve-localmodelnode-agent` does not work for Talos

steffen-karlsson:
`/model` is not a writeable path in Talos, so in order to support LocalModelCache in KServe, we need to be able to override the `hostPath`. It defaults to /model so no changes in the default behaviours.

N/A

I ran a `helm template` test with and without the `kserve.localmodelnode.controller.hostPath` set and here are the diff result:
```yaml
--- Default (hostPath: /models)
+++ Custom (hostPath: /custom/models)
@@ -306,7 +306,7 @@
       tolerations: []
       volumes:
       - hostPath:
-          path: /models
+          path: /custom/models
           type: DirectoryOrCreate
         name: models
```
Do I need to bump the Chart Version?

**Checklist**:

**Release note**:
```release-note
`hostPath` for `kserve-localmodelnode-agent` is now configurable, defaults to `/model`.
```

### DIFF
--- a/charts/kserve-localmodel-resources/README.md
+++ b/charts/kserve-localmodel-resources/README.md
@@ -36,6 +36,7 @@ $ helm install kserve-localmodel-resources oci://ghcr.io/kserve/charts/kserve-lo
 | kserve.localmodel.controller.resources.requests.memory | string | `"200Mi"` |  |
 | kserve.localmodel.controller.tag | string | `""` |  |
 | kserve.localmodelnode.controller.affinity | object | `{}` |  |
+| kserve.localmodelnode.controller.hostPath | string | `"/models"` |  |
 | kserve.localmodelnode.controller.image | string | `"kserve/kserve-localmodelnode-agent"` |  |
 | kserve.localmodelnode.controller.imagePullPolicy | string | `"Always"` |  |
 | kserve.localmodelnode.controller.nodeSelector.kserve/localmodel | string | `"worker"` |  |
--- a/charts/kserve-localmodel-resources/files/daemonset-patch.yaml
+++ b/charts/kserve-localmodel-resources/files/daemonset-patch.yaml
@@ -20,3 +20,8 @@ spec:
         imagePullPolicy: {{ .Values.kserve.localmodelnode.controller.imagePullPolicy }}
         resources:
           {{- toYaml .Values.kserve.localmodelnode.controller.resources | nindent 10 }}
+      volumes:
+      - hostPath:
+          path: {{ .Values.kserve.localmodelnode.controller.hostPath }}
+          type: DirectoryOrCreate
+        name: models
--- a/charts/kserve-localmodel-resources/files/daemonset-patch.yaml
+++ b/charts/kserve-localmodel-resources/files/daemonset-patch.yaml
@@ -22,6 +22,6 @@ spec:
           {{- toYaml .Values.kserve.localmodelnode.controller.resources | nindent 10 }}
       volumes:
       - hostPath:
-          path: {{ .Values.kserve.localmodelnode.controller.hostPath }}
+          path: {{ .Values.kserve.localmodelnode.controller.hostPath | default "/models" }}
           type: DirectoryOrCreate
         name: models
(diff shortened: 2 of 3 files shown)


steffen-karlsson:
Hi @Jooho 
Any updates on reviewing this PR?
I has been waiting for 3 weeks and its blocking new version 0.17.0/0.18.0 to work on Talos Linux.

Thank you 🙏

steffen-karlsson:
@Jooho any updates?

Bafff:
Hi @steffen-karlsson, thanks for the fix. I hit the same non-templatable `hostPath` and reviewed your PR while investigating. The template mechanics are correct, but there is one issue that will quietly drop the fix later. Details below; everything was reproduced with `helm template` against your diff applied to the `v0.20.0-rc0` tree.

## What I verified works

The `deepMerge` helper in `_utils.tpl` merges arrays whose elements have a `name` field via `mergeArrayByName`, so the patched `volumes[name=models]` entry merges into the base entry and nested maps merge recursively: `hostPath.path` is overridden while `type: DirectoryOrCreate` is preserved.

- default render: `path: /models`, identical output to before the PR ✅
- `--set kserve.localmodelnode.controller.hostPath=/mnt/nvme/models`: `path: /mnt/nvme/models` ✅

## Issue: the new value is added to a generated file

`charts/kserve-localmodel-resources/values.yaml` is generated by `hack/setup/scripts/generate_chart_manifests.sh`:

```bash
# hack/setup/scripts/generate_chart_manifests.sh (lines 78-82)
yq eval-all '. as $item ireduce ({}; . * $item)' \
  ${REPO_ROOT}/charts/_common/common-sections.yaml \
  ${REPO_ROOT}/charts/_common/kserve-localmodel-resources-specific.yaml \
  > ${REPO_ROOT}/charts/kserve-localmodel-resources/values.yaml
```

(`charts/_common/README.md` also warns: *"Make sure you edited files in `charts/_common/`, not in `charts/*/values.yaml`"*.)

The PR adds `hostPath: /models` only to the generated `values.yaml`, not to `charts/_common/kserve-localmodel-resources-specific.yaml`. Because the file is generated, the next successful manifest-generation run will overwrite it and the value disappears. Since the template has no guard, the chart then renders an invalid DaemonSet with no error at template time:

```bash
$ ./hack/setup/scripts/generate_chart_manifests.sh
$ helm template t charts/kserve-localmodel-resources | grep -A4 'volumes:'
      volumes:
      - hostPath:
          path: null            # <- rejected by the apiserver: hostPath.path is required
          type: DirectoryOrCreate
        name: models
```

**Fix:** add the key to the `_common` source as well:

```yaml
# charts/_common/kserve-localmodel-resources-specific.yaml
kserve:
  localmodelnode:
    controller:
      ...
      hostPath: /models
```

I re-ran the full `generate_chart_manifests.sh` with this change applied: the value survives regeneration, the regenerated `values.yaml` matches, and `--set` overrides still work. Without it, the same script run drops the value and the chart renders `path: null`, which `helm template` and even `kubectl apply --dry-run=client` accept silently; only the apiserver rejects it.

## Two smaller suggestions

1. Guard the template so a missing value can never render `path: null`:

   ```yaml
   path: {{ .Values.kserve.localmodelnode.controller.hostPath | default "/models" }}
   ```

   Verified: with the value absent entirely, this still renders `path: /models`.

2. The PR description and release note say `/model`; the manifest default is `/models`.

Separate cleanup, not blocking this PR: across the current chart templates, the only reference to `kserve.localmodel.agent.*` is `reconcilationFrequencyInSecs` in the shared ConfigMap patch. The leftover `kserve.localmodel.agent.hostPath: /mnt/models` in the same values file is not consumed by anything. I'll file that separately.

steffen-karlsson:
Hi @Bafff
All good, PR has been updated.
Please assess again and merge if good

Bafff:
@steffen-karlsson I do not have such permissions :) Let's call code owners :)

Bafff:
@Jooho can you take a look please?

PR #5481, opened 2026-04-30
feat(llmisvc): distributed tracing API addition to LLMISVC

Gregory-Pereira:

Adds a tracing field to the LLMInferenceService API that gives users a clean, declarative way to enable OpenTelemetry distributed tracing on their inference deployments — no more manually wiring up env vars and args through pod template overrides.

**Design choices**

Per-component tracing, not top-level. We currently have two components that support tracing — the scheduler (EPP) and the inference server (vLLM). Rather than a single global tracing toggle, the spec has separate tracing blocks at spec.tracing (for the inference server) and spec.router.scheduler.tracing (for the scheduler). This lets users enable tracing on just one component, both, or neither. It also leaves room for future components to get their own tracing config without restructuring the API.

Follows the existing kserve "omit vs empty" pattern. If you omit tracing entirely, nothing changes — no tracing instrumentation is injected. If you set tracing: {}, you get tracing with all defaults filled in. And of course you can override individual fields.

Shared spec for decode and prefill, divergent service names. In a disaggregated (P/D) setup, the tracing config under spec.tracing is shared between decode and prefill workloads. The controller automatically appends -decode or -prefill to the service name so traces from each component are distinguishable in your collector. The -decode suffix is always appended to the decode workload, even in non-disaggregated setups — this is purely a semantic label for the OTel collector, so keeping it consistent makes traces easier to reason about regardless of topology.

Kubernetes resource attributes are always injected. When tracing is enabled, the controller automatically sets up OTEL_RESOURCE_ATTRIBUTES with the pod name, node name, and namespace via the downward API. These are always derived from k8s metadata and shouldn't need user configuration.

**Tracing API**:

A Minimal example following the pattern to opt into tracing with configs defaulted in:

```yaml
kind: LLMInferenceService
metadata:
  name: my-llm
spec:
  tracing: {}
  router:
    scheduler:
      tracing: {}
  model:
```

An example showcasing more custom overrides:

```yaml
kind: LLMInferenceService
metadata:
  name: my-llm
spec:
  tracing:
    serviceName: my-inference-worker
    otelExporterEndpoint: "http://my-collector:4317"
    tracesSamplerArg: "0.1"
  router:
    scheduler:
      tracing:
        serviceName: my-scheduler
        otelExporterEndpoint: "http://my-collector:4317"
        tracesSamplerArg: "0.1"
  model:
```

- [ ] Deploy an LLMISVC without DT to showcase nothing has been affected outside of the CRD change (not strictly required but helpful)
- [ ] Deploy an LLMISVC with simplified DT (no overrides) to showcase it works out of the box with reasonable production defaults
- [ ] Deploy an LLMISVC overriding some fields to showcase the correct behaviour that happens in the strategic merge

**Release note**:
```release-note
API changes to the llmisvc CRD creating first class support for distributed tracing for the `inference-scheduler` and `inference-server` components.
```

cc @vivekk16 @pierDipi


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -294,6 +302,12 @@ type SchedulerSpec struct {
 
 	// Replicas is the number of replicas for the scheduler.
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Tracing configuration for the scheduler component.
+	// When present (even as `{}`), distributed tracing is enabled with defaults.
+	// When omitted, no tracing instrumentation is injected.
+	// +optional
+	Tracing *TracingSpec `json:"tracing,omitempty"`
```
pierDipi: do we have a use case where tracing would be enabled at the scheduler level only and not everywhere? can we start with only the top-level tracing configuration that gets propagated to every managed component?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -444,6 +458,55 @@ type KEDAScalingSpec struct {
...
+	// Maps to the OTEL_TRACES_SAMPLER_ARG environment variable.
+	// Default: "0.05" (5% sampling rate)
+	// +optional
+	TracesSamplerArg *string `json:"tracesSamplerArg,omitempty"`
+
+	// TracesExporter specifies which exporter is used for traces.
+	// Maps to the OTEL_TRACES_EXPORTER environment variable.
+	// Default: "otlp"
+	// +optional
+	TracesExporter *string `json:"tracesExporter,omitempty"`
```
pierDipi:
```suggestion
	// Sampler specifies the sampler to use for traces.
	// Maps to the OTEL_TRACES_SAMPLER environment variable.
	// Default: "parentbased_traceidratio"
	// +optional
	Sampler *string `json:"sampler,omitempty"`

	// SamplerArg is an argument passed to the traces sampler (e.g. the sampling ratio).
	// Maps to the OTEL_TRACES_SAMPLER_ARG environment variable.
	// Default: "0.05" (5% sampling rate)
	// +optional
	SamplerArg *string `json:"samplerArg,omitempty"`

	// Exporter specifies which exporter is used for traces.
	// Maps to the OTEL_TRACES_EXPORTER environment variable.
	// Default: "otlp"
	// +optional
	Exporter *string `json:"exporter,omitempty"`
```

pierDipi: For consistency should we either have `Otel` prefix everywhere or none in any?

pkg/controller/v1alpha2/llmisvc/tracing.go
```diff
@@ -0,0 +1,166 @@
...
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					APIVersion: "v1",
+					FieldPath:  "metadata.name",
+				},
+			},
+		},
+		{
+			Name:  "OTEL_RESOURCE_ATTRIBUTES",
+			Value: "k8s.namespace.name=" + namespace + ",k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)",
```
pierDipi: should we also set llmisvc name?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
pierDipi: maybe prefix is not needed as OTEL is unlikely to be replaced any time soon in those components?

spolti: Are there plans to support ssl and auth?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
Gregory-Pereira: You can have different sampling rates for different components

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
Gregory-Pereira: Open to it but beyond the scope of initial PR support I think

pierDipi:
Can we add e2e tests?

- enable tracing (100% sampling)
- send requests
- verify tracing spans are in the backend (maybe using Jaeger All-in-One?)

pkg/controller/v1alpha2/llmisvc/tracing.go
```diff
@@ -0,0 +1,174 @@
+/*
+Copyright 2025 The KServe Authors.
```
bartoszmajsak: nit ;) we are in 2026 already.

pkg/controller/v1alpha2/llmisvc/tracing.go
```diff
@@ -0,0 +1,174 @@
...
+func injectSchedulerTracing(t *v1alpha2.TracingSpec, namespace, llmisvcName string, container *corev1.Container) {
+	if t == nil {
+		return
+	}
+
+	if !slices.Contains(container.Args, "--tracing=true") &&
+		!slices.Contains(container.Args, "-tracing=true") &&
+		!slices.Contains(container.Args, "--tracing") {
+		container.Args = append(container.Args, "--tracing=true")
+	}
```
bartoszmajsak: Maybe that's simpler and  respects any existing user settings?

```suggestion
	if !hasArgs(container.Args, "--tracing") &&
		!hasArgs(container.Args, "-tracing") {
		container.Args = append(container.Args, "--tracing=true")
	}
```

pkg/controller/v1alpha2/llmisvc/tracing.go
```diff
@@ -0,0 +1,174 @@
...
+// TracingSpec field values are expected to be populated by the well-known config
+// merge (kserve-config-llm-tracing) before this function is called.
+func injectServerTracing(t *v1alpha2.TracingSpec, namespace, llmisvcName, roleSuffix string, container *corev1.Container) {
+	if t == nil {
+		return
+	}
+
+	endpoint := ptr.Deref(t.ExporterEndpoint, "")
+	if endpoint == "" {
+		return
```
bartoszmajsak: The silent early return makes it easy to write tests that pass for the wrong reasons :)

WDYT about returning the mutated `*corev1.Container`?

test/e2e/llmisvc/test_tracing.py
```diff
@@ -0,0 +1,149 @@
...
+
+        logger.info("Inference request succeeded; verifying tracing spans in Jaeger...")
+
+        server_traces = assert_jaeger_traces_exist("inference-server-decode")
+        logger.info(f"Found {len(server_traces)} trace(s) for inference-server-decode")
+
+        try:
+            scheduler_traces = assert_jaeger_traces_exist(
+                "inference-scheduler", timeout=30
+            )
```
pierDipi: it would be good to have additional assertions like the vLLM spans assertion, ideally the assertions function would take a tracing tree as we need to assert that spans are linked to each other in a full request (tracing with no correlation is not very useful).
Example helpers https://github.com/knative/eventing/blob/main/test/conformance/helpers/tracing/traces.go#L147

In addition, if we add additional tests (like a simulated P/D) we're not grabbing the correct spans associated with the llmisvc under test, we might be getting spans associated with unrelated tests.

test/e2e/llmisvc/test_tracing.py
```diff
@@ -0,0 +1,149 @@
...
+            logger.info(
+                f"Found {len(scheduler_traces)} trace(s) for inference-scheduler"
+            )
+        except AssertionError:
+            logger.warning(
+                "inference-scheduler traces not found; EPP may not export "
+                "traces depending on its OTLP configuration"
+            )
```
pierDipi: do you think this test can be debugged with just these 2 logs? 
In addition, the message might be misleading if we assume that tracing is enabled with 100% spans generated (which is fair for these tests)

pierDipi: don't we also need to re-raise the assertion error to get the test to fail?

if the intention is to not fail the test (given the use of warning), what are actually getting out of this CI check? A log on a green check that nobody reads is not useful to catch regressions.

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -503,3 +504,105 @@ jobs:
...
+        run: |
+          minikube image load ./tmp/${{ env.LLMISVC_CONTROLLER_IMG }}-${{ env.TAG }}
+          echo "Images in minikube:"
+          minikube image ls | grep ${{ env.LLMISVC_CONTROLLER_IMG }} || echo "No llmisvc-controller images found"
+
+      - name: Load storage-initializer image into minikube
+        run: |
+          minikube image load ./tmp/${{ env.STORAGE_INIT_IMG }}-${{ env.TAG }}
+          echo "Images in minikube:"
+          minikube image ls | grep ${{ env.STORAGE_INIT_IMG }} || echo "No storage-initializer images found"
```
pierDipi: is this the same flow we use in the other test for llmisvc? why is this different? `./.github/actions/load-docker-images` has a reduce-flake logic that might be helpful

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -503,3 +504,105 @@ jobs:
...
+
+      - name: KServe dependency setup (tracing)
+        uses: ./.github/actions/kserve-dep-setup
+        with:
+          deploy-llmisvc: "true"
+          autoscaler: "tracing"
```
pierDipi: isn't this kinda strange? `autoscaler = tracing`?

test/e2e/llmisvc/test_tracing.py
pierDipi: CI is logging the error but is green

```
INFO     e2e.llmisvc.logging:test_llm_inference_service.py:1127 Waiting: No traces found for service 'inference-scheduler'
assert 0 > 0
 +  where 0 = len([])
ERROR    e2e.llmisvc.logging:test_llm_inference_service.py:1124 Timed out waiting: No traces found for service 'inference-scheduler'
assert 0 > 0
 +  where 0 = len([])
WARNING  e2e.llmisvc.logging:test_tracing.py:138 inference-scheduler traces not found; EPP may not export traces depending on its OTLP configuration
```

.github/workflows/e2e-test-llmisvc.yaml
```diff
@@ -280,7 +281,7 @@ jobs:
           # Run only CPU tests for now using pytest markers (cluster_)
           # Available GPU vendors: amd, nvidia, intel
           export PYTEST_MAXFAIL=3
-          ./test/scripts/gh-actions/run-e2e-tests.sh "llminferenceservice and cluster_cpu and not autoscaling" 0 "envoy-gateway"
+          ./test/scripts/gh-actions/run-e2e-tests.sh "llminferenceservice and cluster_cpu and not autoscaling and not tracing" 0 "envoy-gateway"
```
bartoszmajsak, line 284: Can we have consitent markers approach that is inclusion-based? Could be error-prone if we mix and it reads kinda backwards. Maybe we need categories like "base" or "core" and then add special ones with autoscaling and tracing?

vivekk16: Good suggestion. Will address this in a follow-up PR since it touches test files beyond the scope of this change.

.github/workflows/e2e-test-llmisvc.yaml
vivekk16: Updated this to have a separate block to install observability dependencies.

test/e2e/llmisvc/test_tracing.py
vivekk16: Added span link verification, cross-service trace ID correlation between scheduler and server, and scoped all Jaeger queries by k8s.namespace.name tag to isolate from unrelated tests.

test/e2e/llmisvc/test_tracing.py
vivekk16: Added trace detail logging that prints trace IDs, span counts, and operation names for each service. Should make CI failures much easier to debug.

vivekk16: Agreed, removed the try/except entirely. Scheduler trace assertion is now a hard failure. Also fixed the root cause. The fixture was using always_on sampler (unsupported by EPP) and a missing http:// scheme on the endpoint. Both components now reliably export traces.

.github/workflows/e2e-test-llmisvc.yaml
vivekk16: Updated.

test/e2e/llmisvc/fixtures.py
```diff
@@ -1368,14 +1376,19 @@ def _setup_test_case_service(kserve_client, tc, test_node_name, peer_index=None)
...
-            "apiVersion": "serving.kserve.io/v1alpha1",
+            "apiVersion": api_version,
             "kind": "LLMInferenceServiceConfig",
             "metadata": {
                 "name": unique_config_name,
                 "namespace": KSERVE_TEST_NAMESPACE,
             },
-            "spec": LLMINFERENCESERVICE_CONFIGS[base_ref],
+            "spec": original_spec,
         }
```
pierDipi: this signal we've lossy conversions between api versions, can we add tracing fields to v1alpha1 as well without this special casing?

vivekk16: Updated.

pierDipi (approved):

PR #5485, opened 2026-05-01
feat(llmisvc): add preStop hook and up terminationGracePeriod 

andresllh:
## Summary

- Adds a `/bin/sleep 15` preStop lifecycle hook to all vLLM container specs in the `LLMInferenceServiceConfig` preset templates
- Increases `terminationGracePeriodSeconds` from 30s to 60s across all vLLM templates
- Adds a behavioral e2e test verifying the preStop hook fires during pod termination

## How it works

When a pod is deleted, Kubernetes removes it from EndpointSlices immediately, but EPP needs a few seconds to propagate that change and stop routing new traffic to the pod. Without a preStop hook, SIGTERM fires immediately and EPP may still route new requests to a shutting-down vLLM, causing connection errors.

The 15s preStop sleep bridges this gap. After the sleep, SIGTERM fires and uvicorn drains in-flight requests (including streaming) before exiting — this is built-in uvicorn behavior, not version-gated.

With the previous `terminationGracePeriodSeconds: 30`, vLLM only had ~15s to drain after the preStop sleep. The new 60s gives uvicorn 45s to drain in-flight requests, which is more appropriate for long streaming LLM responses. Users can override `terminationGracePeriodSeconds` via `spec.workload.template.terminationGracePeriodSeconds` in their `LLMInferenceService` if needed.

## Test plan

- [x] Unit test (`TestPresetFiles`) updated with `Lifecycle` and `TerminationGracePeriodSeconds` assertions — all passing
- [x] Behavioral e2e test (`test/e2e/llmisvc/test_prestop_hook.py`) verifies pod termination takes `>= 12s`, confirming the preStop hook fired — verified passing on minikube


jlost (commented):
/lgtm

jlost:
I think it's worth looking into handling this from the routing side - for example, with EPP fallback endpoints or a gateway-based retry. That said, this looks good!

pierDipi:
@spolti @Jooho can you merge this PR?

PR #5486, opened 2026-05-01
e2e: reduce verbose payload logging in e2e tests

maskarb:

Disables verbose logging on the inference clients used in e2e tests by setting `verbose=False`. This eliminates multi-KB payload dumps (full gRPC tensor arrays, base64-encoded images, prediction outputs) from CI logs, making it easier to find actual test failures.

N/A

- [x] No behavioral changes — only disables verbose client logging
- [x] `ruff format` and `ruff check` pass

Three `verbose=True` → `verbose=False` changes:
- `test/e2e/conftest.py` — REST v1 and v2 client fixtures
- `test/e2e/common/utils.py` — gRPC client helper

No changes to the inference client library itself.

**Checklist**:

```release-note
Disable verbose payload logging on inference clients in e2e tests to reduce CI log noise.
```

### DIFF
--- a/test/e2e/common/utils.py
+++ b/test/e2e/common/utils.py
@@ -47,7 +47,7 @@ def grpc_client(host, cluster_ip):
     logger.info("gRPC target host: %s", host)
     return InferenceGRPCClient(
         cluster_ip,
-        verbose=True,
+        verbose=False,
         channel_args=[
             ("grpc.ssl_target_name_override", host),
         ],
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -69,7 +69,7 @@ async def rest_v1_client():
         config=RESTConfig(
             transport=transport,
             timeout=180,
-            verbose=True,
+            verbose=False,
             protocol=PredictorProtocol.REST_V1,
         )
     )
@@ -97,7 +97,7 @@ async def rest_v2_client():
         config=RESTConfig(
             transport=transport,
             timeout=180,
-            verbose=True,
+            verbose=False,
             protocol=PredictorProtocol.REST_V2,
         )
     )


python/kserve/kserve/inference_client.py
```diff
@@ -622,9 +619,8 @@ async def is_server_live(
             logger.info("url: %s, protocol_version: %s", url, self._config.protocol)
         response = await self._client.get(url, headers=headers, timeout=timeout)
         if self._config.verbose:
```
spolti, line 624: Is the verbose enabled by default?

python/kserve/kserve/inference_client.py
```diff
@@ -217,7 +217,7 @@ async def infer(
         else:
             raise InvalidInput("Invalid input format")
         if self._verbose:
-            logger.info(
+            logger.debug(
```
spolti: I guess the it would be better here if we figure out if the tests are enabling it or not.
imho, the puporse of keept it as INFO was to be easier to configure just by setting the verbose flag.

python/kserve/kserve/inference_client.py
maskarb: I believe yes, verbose is true by default in the e2e tests:
https://github.com/kserve/kserve/blob/master/test/e2e/conftest.py#L72
https://github.com/kserve/kserve/blob/master/test/e2e/conftest.py#L100

python/kserve/kserve/inference_client.py
maskarb: The tests do enable verbose. The issue is that full request/response payloads (which can be very large for model inference) flood the CI logs and make it hard to find actual test failures.

This PR takes a split approach rather than removing the logging entirely:
- **INFO**: metadata that's always useful for debugging (URLs, status codes, protocol) — stays visible with just `verbose=True`
- **DEBUG**: full request/response bodies — available when you need them by setting the log level

The verbose flag still gates all of it — nothing is logged without `verbose=True`. This just adds a layer so the common case (figuring out which endpoint was hit and whether it succeeded) doesn't require scrolling past megabytes of payload data.

spolti: maybe we could just move the verbose flag to false as default and leave the rest as it is?
@Jooho @israel-hdez wdyt?

israel-hdez (approved):
/lgtm

PR #5487, opened 2026-05-01
feat(llmisvc): default to FlowControl enabled with sample and docs

RishabhSaini:
- Enable `flowControl` feature gate in default `EndpointPickerConfig` for single-profile and prefill-decode paths
  - Add flow-control sample with three priority bands (1, 0, -1), SLO deadline ordering, and round-robin fairness
  - Add README with request headers, priority band configuration, and saturation tuning guidance


pierDipi (commented):
why is it not enabled by default in llm-d scheduler itself?

PR #5493, opened 2026-05-06
test(llmisvc): harden autoscaling e2e with pipeline health checks

vivekk16:
## Summary

Hardens the LLMISvc autoscaling E2E tests by adding pre-test pipeline validation and fixing several assertion gaps that could mask real failures.

### Changes

**New: `verify-autoscaling-health.sh` — fail-fast pipeline health check**
- Validates the full metrics pipeline _before_ tests run (replaces the old `|| true` checks in CI)
- Checks: Prometheus API reachable, WVA ServiceMonitor target scraped, External Metrics API healthy, operator pods ready
- Supports both `hpa` and `keda` modes
- Integrated into the CI workflow (`e2e-test-llmisvc.yaml`)

**Fixed: in-test assertions in `test_llm_autoscaling.py`**
- `send_load`: now tracks success/error counts with thread-safe counters; fails the test if all requests error (previously failures were silent)
- `assert_wva_metric_exists`: reads the correct field (`status.desiredOptimizedAlloc.numReplicas`) instead of the non-existent `status.desiredReplicas`
- `assert_hpa_active` / `assert_scaled_object_active`: poll with `wait_for` (60s timeout) instead of a single point-in-time check
- `assert_lws_replicas`: uses the correct controller-generated LWS resource name (`{name}-kserve-mn`)
- New `assert_metrics_pipeline_ready`: verifies APIService availability and operator pod health in-test

**Enhanced: `status-check.sh` diagnostics**
- Dumps ServiceMonitors, VariantAutoscalings, HPAs, ScaledObjects, WVA controller logs, and Prometheus Adapter logs on failure

### DIFF
--- a/test/e2e/llmisvc/test_llm_autoscaling.py
+++ b/test/e2e/llmisvc/test_llm_autoscaling.py
@@ -40,6 +40,7 @@ integration tests in pkg/controller/v1alpha2/llmisvc/scaling_int_test.go.
 import concurrent.futures
 import logging
 import os
+import threading
 import time
 
 import pytest
@@ -204,18 +205,36 @@ def send_load(service_url, model_name, concurrency=5, duration_seconds=30):
     headers = {"Content-Type": "application/json"}
 
     deadline = time.time() + duration_seconds
+    lock = threading.Lock()
+    counters = {"success": 0, "error": 0}
 
     def _worker():
         while time.time() < deadline:
             try:
-                requests.post(endpoint, json=payload, headers=headers, timeout=30)
+                resp = requests.post(
+                    endpoint, json=payload, headers=headers, timeout=30
+                )
+                with lock:
+                    if resp.status_code < 500:
+                        counters["success"] += 1
+                    else:
+                        counters["error"] += 1
             except Exception:
-                pass
+                with lock:
+                    counters["error"] += 1
             time.sleep(0.1)
 
     with concurrent.futures.ThreadPoolExecutor(max_workers=concurrency) as pool:
         futures = [pool.submit(_worker) for _ in range(concurrency)]
         concurrent.futures.wait(futures)
+
+    logger.info(
+        f"send_load complete: {counters['success']} ok, "
+        f"{counters['error']} errors to {endpoint}"
+    )
+    assert counters["success"] > 0, (
+        f"send_load: all {counters['error']} requests failed to {endpoint}"
+    )
 
 
 # --- Patching helpers ---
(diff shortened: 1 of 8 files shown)


test/e2e/llmisvc/test_llm_autoscaling.py
```diff
@@ -286,6 +305,110 @@ def assert_scaling_resources_deleted(
...
+    """Verify LeaderWorkerSet status reflects scale-up."""
+    api = client.CustomObjectsApi()
+    lws_name = _child_name(service_name, "-kserve-mn")
+    lws = api.get_namespaced_custom_object(
+        "leaderworkerset.x-k8s.io", "v1", namespace, "leaderworkersets", lws_name
+    )
+    actual = lws.get("status", {}).get("replicas", 0)
+    assert actual >= min_replicas, (
+        f"LWS {lws_name} replicas={actual}, expected >= {min_replicas}"
+    )
```
jlost: Is there a reason this doesn't follow the `wait_for` pattern like the others?

test/e2e/llmisvc/test_llm_autoscaling.py
```diff
@@ -286,6 +305,110 @@ def assert_scaling_resources_deleted(
...
+            label_selector="app.kubernetes.io/name=prometheus-adapter",
+        )
+        running = [p for p in pods.items if p.status.phase == "Running"]
+        assert len(running) > 0, "No running Prometheus Adapter pods found"
+    elif actuator == "keda":
+        v1 = client.CoreV1Api()
+        pods = v1.list_namespaced_pod(
+            namespace="keda",
+            label_selector="app.kubernetes.io/name=keda-operator",
+        )
```
jlost, line 341: The namespaces `monitoring` / `keda` are hard-coded here, but `verify-autoscaling-health.sh` suggests they can be overridden by env.

jlost (commented):
Looks good! I submitted a question and a nit.

test/e2e/llmisvc/test_llm_autoscaling.py
vivekk16: Good catch. No intentional reason, it was an oversight. Updated to use the wait_for polling pattern with a 60s timeout and 5s interval, consistent with assert_hpa_active and assert_scaled_object_active.

test/e2e/llmisvc/test_llm_autoscaling.py
vivekk16: Updated to read from KEDA_NAMESPACE and PROMETHEUS_NAMESPACE env vars (defaulting to "keda" and "monitoring"), matching what verify-autoscaling-health.sh already does.

jlost (commented):
/lgtm

test/e2e/llmisvc/test_llm_autoscaling.py
```diff
@@ -204,19 +208,37 @@ def send_load(service_url, model_name, concurrency=5, duration_seconds=30):
...
 
     def _worker():
         while time.time() < deadline:
             try:
-                requests.post(endpoint, json=payload, headers=headers, timeout=30)
+                resp = requests.post(
+                    endpoint, json=payload, headers=headers, timeout=30
+                )
+                with lock:
+                    if resp.status_code < 500:
```
bartoszmajsak: Do we consider 40x as a success?

vivekk16: No, 4xx definitely should not be treated as success, that was sloppy naming on my part. The goal of send_load is just to hammer the endpoint with traffic so autoscaling kicks in, not to check that responses are correct. During scale up we expect some 4xx (like 429s when the server is overloaded), so we don't want the load generator to fail on those.

I've split the counters into success (200), client_error (4xx), and server_error (5xx/exceptions) to be more clear. The assertion still only fails if everything came back as a server error.

PR #5496, opened 2026-05-06
feat(llmisvc): adding vLLM shutdown-timeout

andresllh:

Implements graceful drain for LLM inference pods to ensure active requests complete fully before a pod exits during rolling updates or scale-down events. Without this, SIGTERM caused vLLM to abort in-flight requests immediately.

Two complementary mechanisms are added:

1. **preStop hook** (`/bin/sleep 15`): gives EPP (Endpoint Picker) time to detect the pod's readiness probe failure and stop routing new requests to it before SIGTERM is sent.

2. **vLLM `--shutdown-timeout`**: set on all LLM config templates so vLLM drains in-flight requests gracefully after SIGTERM instead of aborting them immediately. The value is computed as `terminationGracePeriodSeconds - 15` (the preStop duration), defaulting to 45 s. If the user sets a custom `spec.template.terminationGracePeriodSeconds` on their `LLMInferenceService`, that value is used automatically via Go template helpers (`shutdownTimeout` + `tgps`).

`terminationGracePeriodSeconds` is also increased from the Kubernetes default (30 s) to 60 s across all templates to give enough room for both phases.

The drain sequence on pod termination is:
\`\`\`
preStop sleep (15 s) → EPP stops routing → SIGTERM → vLLM drains (≤45 s) → pod exits
\`\`\`

Fixes #

- Unit tests added to `TestReplaceVariables` covering:
  - `shutdownTimeout` default (45 s) when `spec.template` is nil
  - `shutdownTimeout` uses user-set `terminationGracePeriodSeconds` (e.g. 120 s → 105 s)
- E2E test added (`test/e2e/llmisvc/test_prestop_hook.py`) that sends streaming inference requests during a rolling update and asserts no requests are dropped.

The `tgps` FuncMap helper is needed because `.Spec.Template` is a `*corev1.PodSpec` (can be nil). Accessing `.Spec.Template.TerminationGracePeriodSeconds` directly in a Go template causes a nil-pointer error when the field is unset; the helper handles that safely in Go code.

**Checklist**:

**Release note**:
\`\`\`release-note
LLMInferenceService pods now drain in-flight requests gracefully on termination via a preStop hook and vLLM --shutdown-timeout, preventing request drops during rolling updates.
\`\`\`


config/llmisvcconfig/config-llm-prefill-template.yaml
```diff
@@ -140,6 +140,7 @@ spec:
                 --served-model-name "{{ .Spec.Model.Name }}" \
                 --port 8000 \
                 --disable-uvicorn-access-log \
+                --shutdown-timeout {{ shutdownTimeout (tgps .Spec.Template) 15 }} \
```
pierDipi: Don't we need to use `.Spec.Prefill.Template`? same for multi node using leader vs worker pod spec

andresllh: Done

pierDipi:
@andresllh please rebase

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -457,6 +457,38 @@ func ReplaceVariables(llmSvc *v1alpha2.LLMInferenceService, llmSvcCfg *v1alpha2.
...
+			},
+			// prefillWorkerTGPS safely extracts TerminationGracePeriodSeconds from the worker
+			// PodSpec of a potentially-nil WorkloadSpec (used for .Spec.Prefill.Worker).
+			"prefillWorkerTGPS": func(ws *v1alpha2.WorkloadSpec) *int64 {
+				if ws == nil || ws.Worker == nil {
+					return nil
+				}
+				return ws.Worker.TerminationGracePeriodSeconds
+			},
 		}).
```
pierDipi: can we use a single `shutdownTimeout` function? since those functions end up be APIs that need to be stable, the less we expose the better

config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -149,6 +149,7 @@ spec:
               --served-model-name "{{ .Spec.Model.Name }}" \
               --port 8001 \
               ${ACCESS_LOG_ARGS} \
+              --shutdown-timeout {{ shutdownTimeout (tgps .Spec.Template) 15 }} \
```
pierDipi: (same for others) we need to inject this arg only when the vLLM version supports it, similar to how it's done for `ACCESS_LOG_ARGS`

markmc:
Nice work!

Could you take a look at [this Gemini feedback](https://github.com/vllm-project/vllm/pull/43208#discussion_r2103399498) that when vLLM is launched via `/bin/sh -c` (or `/bin/bash -c`), the shell becomes PID 1 and doesn't forward SIGTERM to child processes by default. They suggest using `exec` so vLLM replaces the shell and receives SIGTERM directly.

The kserve template launches `eval "vllm serve ..."` which would have the same issue — bash would receive the SIGTERM but not forward it to the vLLM child process. Does `--shutdown-timeout` actually trigger in this setup, or is the drain relying entirely on the preStop hook delay + eventual SIGKILL?

config/llmisvcconfig/config-llm-decode-template.yaml
pierDipi:
```suggestion
            eval "exec vllm serve /mnt/models \
```
https://github.com/vllm-project/vllm/pull/43208/changes#r3273275082

(same for others)

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -457,6 +457,38 @@ func ReplaceVariables(llmSvc *v1alpha2.LLMInferenceService, llmSvcCfg *v1alpha2.
...
 		Funcs(map[string]any{
 			"ChildName": kmeta.ChildName,
+			// shutdownTimeout computes the vLLM --shutdown-timeout value:
+			// terminationGracePeriodSeconds - preStop, defaulting to 60 - preStop if unset.
+			"shutdownTimeout": func(tgps *int64, preStop int64) int64 {
+				const defaultTGPS = int64(60)
+				if tgps == nil {
+					return defaultTGPS - preStop
+				}
+				return *tgps - preStop
```
pierDipi:
```
				return Math.Max(0, *tgps - preStop - Math.Min(5, tgps))
```
This provides a buffer for signal propagation and final process cleanup (such as flushing logs or metrics) before Kubernetes sends a SIGKILL

pkg/controller/v1alpha2/llmisvc/config_merge.go
andresllh: Done — consolidated into a single `shutdownTimeout(spec any, preStop int64) int64` that accepts `*corev1.PodSpec` (or nil) directly, removing `tgps`, `prefillTGPS`, and `prefillWorkerTGPS` from the FuncMap.

config/llmisvcconfig/config-llm-decode-template.yaml
andresllh: Done — `--shutdown-timeout` is now gated behind a vLLM 0.18.0 version check (same pattern as `ACCESS_LOG_ARGS`), since that's the release where vllm-project/vllm#36666 landed.

config/llmisvcconfig/config-llm-decode-template.yaml
andresllh: Done — changed to `eval "exec vllm serve ..."` across all templates. This also directly addresses @markmc's comment: with `exec`, vLLM replaces bash and receives SIGTERM directly, so `--shutdown-timeout` actually triggers on pod termination.

pkg/controller/v1alpha2/llmisvc/config_merge.go
andresllh: Done — applied the buffer formula: `max(0, tgps - preStop - min(5, tgps))`. Default timeout is now 40 s (60 - 15 - 5) instead of 45 s.

andresllh:
@markmc great catch! You're right — the original eval 'vllm serve ...' left bash as the process receiving SIGTERM, so --shutdown-timeout would never fire.

Fixed: all templates now use eval 'exec vllm serve ...'. The exec replaces bash with the vLLM process, so SIGTERM hits vLLM directly and triggers the graceful drain via --shutdown-timeout.

pierDipi:
```
failed to merge config: template: config:1:25546: executing "config" at <.Spec.Prefill.Template>: nil pointer evaluating *v1alpha2.WorkloadSpec.Template
```

andresllh:
> ```
> failed to merge config: template: config:1:25546: executing "config" at <.Spec.Prefill.Template>: nil pointer evaluating *v1alpha2.WorkloadSpec.Template
> ```

already fixed, CI is rerunning

markmc:
(leaving this here for future reference)

> the original eval 'vllm serve ...' left bash as the process receiving SIGTERM, so --shutdown-timeout would never fire.

If in your previous testing, when you deleted the pod, it completed all in-flight requests, that would mean bash ignored the `SIGTERM` and kept running and was only killed by `SIGKILL` after the termination grace period

But ... why would bash ignore the signal? I think it's because of this: https://man7.org/linux/man-pages/man2/kill.2.html

> The only signals that can be sent to process ID 1, the init process, are those for which init has explicitly installed signal handlers.  This is done to assure the system is not brought down accidentally.

i.e. the shell hasn't registered a `SIGTERM` handler, so the kernel refuses to deliver the signal to PID 1 in the container

Issue #3750, opened 2024-06-18
Multi-Lora support


skonto:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

There are different directions:
- extend existing API for referencing multiple tuned models. For example can user utilize Triton runtime fetch multiple adapters:
https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/llama#run-llama-with-several-lora-checkpoints? 
Also see the comment here: https://github.com/kserve/kserve/issues/3413#issuecomment-2042948424 for downloading the fine-tuned models.
- support runtime servers that are optimized for multiple loras inference eg. [Lorax](https://github.com/predibase/lorax).
 
**Anything else you would like to add:**

Runtime support for multi-loras is a bit at early stage :

https://github.com/vllm-project/vllm/issues/2602
https://github.com/NVIDIA/TensorRT-LLM/issues/738
https://github.com/huggingface/text-generation-inference/issues/907

**Links to the design documents:**
[Optional, start with the short-form [RFC template](https://docs.google.com/document/d/1UcBeLfZ_JMGpVrPJmYtEIVH_9Y4U3AEKQdq_IKuOMrU) to outline your ideas and get early feedback.]
[Required, use the longer-form [design doc template](https://docs.google.com/document/d/1Mtoui_PP2a9N59NjYHnsvrdJ8t2iKFwIJAx1zRO_I1c) to specify and discuss your design in more detail]

cc @yuzisun

skonto:
@andyi2it hi, I saw you self-assigned this, pls take a look here as well https://github.com/kserve/kserve/issues/3413#issuecomment-2323990187 about current design. Would you be interested in collaborating on this one as I am also looking into it?

thedebugger:
Hi folks, Is there help needed on this feature? If yes, I might be able to help out.

torlar:
when will the kserve huggingface runtime/vllm image be updated to that of vllm where this is supported?

this works out of the box with vllm-openai image when not using kserve.

torlar:
or is LLMInferenceService the way to do this now?

kliukovkin:
Taking a swing at this. Sketching the design here before going deep on implementation — want to confirm the shape with maintainers first.

[...]

PR #5499, opened 2026-05-07
feat(llmisvc): declarative multi-LoRA adapter management

kliukovkin:
### What

Declarative multi-LoRA adapter management on `LLMInferenceService` (alpha, Preload-only).

Replaces the existing `Spec.Model.LoRA.Adapters` Schemaless field with a typed `[]LoRAAdapterSpec`. Each adapter declares Name, URI/Ref, LoadMode, Priority, Disabled. The controller now:

1. Renders `--enable-lora`, `--max-loras`, `--max-lora-rank`, and one `--lora-modules=<name>=/mnt/loras/<name>` per enabled adapter on the vLLM container.
2. Adds a per-adapter init container reusing the storage initializer to fetch weights into `/mnt/loras/<adapter-name>`. Multi-node (worker-data-parallel) inherits this automatically.
3. Emits one `gateway.networking.k8s.io.InferenceObjective` (apix/v1alpha2) per adapter with the user's Priority. The existing GIE EPP `lora-affinity-scorer` filter (fed by vLLM's `vllm:lora_requests_info` gauge per the model-server protocol) routes `model: <adapter-name>` requests to a replica that has it loaded — KServe contributes zero data-plane code.

Refs #3750. Implements the "LoRA adapters to InferenceService" roadmap entry.

cc @andyi2it — saw you're assigned to #3750. Please flag if any of this collides with WIP on your end.

### Commits

| # | What | Hash |
|---|---|---|
| 1 | typed `LoRAAdapterSpec` API + validation + v1alpha1↔v1alpha2 conversion + codegen | `4ce490b18` |
| 2 | render vLLM LoRA args from `spec.model.lora` across all 6 llmisvcconfig templates | `22bd0fb72` |
| 3 | emit `InferenceObjective` per adapter; conditional `Owns` guarded by `IsCrdAvailable` | `3e060cdc0` |
| 4 | per-adapter init-container fanout in storage initializer; ConfigMap/Secret Ref resolution | `bee79f937` |
| 5 | spec-level e2e smoke test + sample manifest + docs README | `777a00293` |

Each commit is independently green for `go test ./pkg/...` and `make go-lint`. The pre-existing envtest failure (`no such file: etcd`) is environmental and unrelated to this PR.

### Out of scope (followup PRs)

- `Dynamic` LoadMode + hot-swap reconciler (`POST /v1/load_lora_adapter` / `unload`). The validation webhook in this PR explicitly rejects `LoadMode: Dynamic` until the followup ships.
- Per-adapter `status.adapters[]` reporting with replica counts.
- `mlflow://` URI scheme (separate Custom Storage Container plugin contribution).
- Cluster-scoped adapter cache for cross-LLMISVC sharing.

### Design choices worth flagging

- **Embedded in `LLMInferenceService`, not a new top-level CRD.** Adapters are tightly coupled to a base model; embedding keeps lifecycle clean and avoids cross-namespace ref machinery.
- **`InferenceObjective` (apix/v1alpha2), not `InferenceModel`.** The latter is deprecated in GIE GA migration.
- **Two separate annotations for v1alpha1↔v1alpha2 round-trip.** Existing `internal.serving.kserve.io/lora-criticalities` (v1alpha1→v1alpha2 promotion) is kept untouched. New `internal.serving.kserve.io/lora-extras` carries the v1alpha2-only fields (URI, Ref, LoadMode, Priority, Disabled) on v1alpha2→v1alpha1 demotion. They travel in opposite directions, so coupling them would be misleading. Annotation size limit (~256KB) is documented in the conversion code; sane adapter counts fit comfortably.
- **Ref resolution: ConfigMap tried first, Secret as fallback** — `LocalObjectReference` only carries `Name`, no type discriminator. Both look up the URI under the `data.uri` key. Open to discriminated approach if reviewers prefer.
- **Storage initializer: per-adapter init container vs multi-URI fork of the existing initializer.** Picked per-adapter for clearer per-adapter status visibility in `kubectl describe pod` and to avoid touching the v1beta1 multi-URI machinery from PR #4702.
- **`r.reconcileChildren` doesn't exist in this package** — the InferenceObjective reconciliation uses a manual desired-vs-actual loop on top of the generic `Reconcile`/`Delete` primitives in `lifecycle_crud.go`, pattern-matching the existing `reconcileSchedulerInferencePool`.
- **Codegen note**: `hack/update-codegen.sh` (which uses `kube::codegen::gen_helpers` from `code-generator@v0.34.3`) skipped the new `LoRAAdapterSpec` due to a stale cache. Running `bin/controller-gen object paths="./pkg/apis/serving/v1alpha2/..."` directly produces the expected output. The committed `zz_generated.deepcopy.go` is from controller-gen direct, not the make target. Worth investigating the kube-codegen path separately.

### Tests

- Unit: 6 cases for vLLM arg rendering, 5 for `InferenceObjective` emission (including GC on adapter removal + custom Priority), 5 for storage initializer fanout (URI, ConfigMap Ref, missing Ref → error, Disabled skipped, no-LoRA no-op), validation cases for duplicate names + URI/Ref exclusivity + Dynamic-rejection + MaxAdapters bound.
- Integration: extended `controller_workload_storage_int_test.go` (envtest pre-existing-failure caveat noted above).
- E2e (smoke): deploys an LLMInferenceService with 2 Preload adapters using the llm-d-inference-simulator workload, asserts on rendered Deployment init containers, vLLM container args, InferenceObjective objects + owner refs. Conditionally skipped if the GIE CRD is not installed. Full request-routing e2e requires real GPU + adapter download time and is deferred to the Dynamic LoadMode followup PR where it carries more value.

### Sample

A runnable manifest + 45-line README live under `docs/samples/llmisvc/lora-adapters/`.

### A few things to flag

The adapters are embedded in `LLMInferenceService` rather than a separate top-level CRD. Adapters are coupled to a base model and embedding keeps lifecycle + cleanup simple. Not married to it but it's the choice that fell out naturally — flag if a separate CRD makes more sense for some reason I'm missing.

For Ref resolution the controller tries ConfigMap first then falls back to Secret with the same name. `LocalObjectReference` doesn't carry a type discriminator so the alternatives are either two separate fields (`RefConfigMap` / `RefSecret`) or a sum type. The implicit fallback felt cleanest but the explicit version is more honest about what the user wrote — fine either way, lmk preference.

Conversion does best-effort migration for existing v1alpha2 objects that used the old Schemaless `LoRASpec.Adapters` (`name` and `uri` map cleanly, anything else preserved in an annotation for forensics). Leaning soft-warn for one minor and hard-validate next, but if anyone's been depending on undocumented field shapes here please yell.


pierDipi:
@andresllh has a PR here https://github.com/kserve/kserve/pull/5317, can you both collaborate / sync?

kliukovkin:
> @andresllh has a PR here #5317, can you both collaborate / sync?

Took a look at #5317 — it's well past mine and clearly the right base. Closing #5499.

A few things from my draft that I think aren't covered in #5317, would make sense as separate followups once it lands. Flagging in case anyone wants to pick them up or wants me to send them as separate PRs:

- Typed `LoRAAdapterSpec` instead of the Schemaless `[]LLMModelSpec` recursion (proper OpenAPI validation, `listType=map` keyed by name)
- `Priority` field on adapters that emits one `InferenceObjective` per adapter for the EPP `lora-affinity-scorer` priority/shedding path
- `Ref` field for ConfigMap/Secret-backed URIs (credential rotation without touching spec)
- `Disabled` flag (declare without materializing)
- `LoadMode: Dynamic` + a hot-swap reconciler over `POST /v1/load_lora_adapter`
- Conversion-webhook round-trip of new fields via `internal.serving.kserve.io/lora-extras` annotation, parallel to the existing criticality one

@andresllh — definitely missed your PR in my dig before opening, sorry for the duplication.

PR #5489, opened 2026-05-04
fix(router): correct off-by-one in splitter pickupRoute random range

mwaykole:
**What this PR does / why we need it**:

Fixes an off-by-one error in the InferenceGraph Splitter router that can cause a nil pointer dereference panic, crashing the router process.

The `pickupRoute` function used `rand.Int(rand.Reader, big.NewInt(101))` which returns values in `[0, 100]`. When route weights sum to 100 (the expected case), a random value of `point=100` causes no route to match—the loop condition `point < end` evaluates to `100 < 100 = false`—so `pickupRoute` returns `nil`. The caller in `routeStep` then dereferences this nil pointer, panicking:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

The probability of this occurring is ~1/101 per Splitter request (~1%).

This PR:
1. Changes `big.NewInt(101)` → `big.NewInt(100)` so the random range becomes `[0, 99]`, guaranteeing a route is always selected when weights sum to 100.
2. Adds a defensive nil guard in `routeStep` for the Splitter case, matching the existing pattern already used by the Switch case, so that even if `pickupRoute` returns nil for any reason, the router returns a 500 error instead of panicking.

**Which issue(s) this PR fixes**:
Fixes #5408

**Feature/Issue validation/testing**:

- [x] `TestPickupRouteNeverReturnsNil` — runs `pickupRoute` 10,000× with weights 20/80, asserts non-nil every iteration
- [x] `TestPickupRouteAlwaysReturnsRouteForAllRandValues` — same with weights 30/70
- [x] `TestCryptoRandIntUpperBoundWithFix` — verifies `crand.Int(big.NewInt(100))` never returns ≥100 over 100,000 iterations
- [x] All existing router tests pass (14/14)
- [x] E2E on Kind cluster: deployed InferenceGraph with Splitter (sklearn-iris:20, xgboost-iris:80), sent 1,200 requests (200 sequential + 1,000 parallel) — **zero failures, zero pod restarts**

<details>
<summary>Unit test results</summary>

```
=== RUN   TestPickupRouteNeverReturnsNil
--- PASS: TestPickupRouteNeverReturnsNil (0.01s)
=== RUN   TestPickupRouteAlwaysReturnsRouteForAllRandValues
--- PASS: TestPickupRouteAlwaysReturnsRouteForAllRandValues (0.01s)
=== RUN   TestCryptoRandIntUpperBoundWithFix
[...]

Issue #5408, opened 2026-04-16
crand.NewInt(100) can return 100 and that can cause panic in routeStep.


ruseel:
/kind bug

**What steps did you take and what happened:**
surveying about kserve with LLM to know its feature, 
and GPT-5.4 said me about possibility of panic in splitter. 

**What did you expect to happen:**
even when crand.NewInt(100) returns 100.
should not panic. 

**What's the InferenceService yaml:**
N/A

**Anything else you would like to add:**
After seeing what GPT-5.4 high said, I tried to validated that claim very carefully. 
I checked that go reference document about crand.NewInt(..) is written 
  range is [0, max) 

and 

Made a that that reference document and actual behavior of crand.NewInt(..) is same. 

and 

Made a test that whether it is panic with seperate tests. 

verbose diff for test is here 
   https://github.com/ruseel/kserve/commit/61fc08c4115e3637e270ff9d425fb14eb7be1281

minimal patch is here 
   https://github.com/ruseel/kserve/commit/c57493128e484fd0a6f379aa980f7bff7333461b

**Environment:**

N/A

PR #5500, opened 2026-05-07
Fix off-by-one in pickupRoute causing ~1% nil pointer panic on Splitter requests

SebTardif:
## Problem

`pickupRoute` uses `rand.Int(rand.Reader, big.NewInt(101))` to generate a random value, which produces integers in **[0, 100] inclusive**. Route weights sum to 100, so when the random value is exactly 100, the loop condition `point < end` is never satisfied and the function returns `nil`. The nil pointer is then dereferenced in `handleSplitterORSwitchNode` (accessing `route.StepName`), panicking the router process.

This has a ~1/101 probability (~0.99%) of occurring on **every Splitter request**.

The existing comment even says `// generate num [0,100)`, confirming the intent was an exclusive upper bound.

## Fix

Change `big.NewInt(101)` to `big.NewInt(100)` so the generated range is [0, 100), matching the comment and ensuring a route is always selected when weights sum to 100.

## Tests added

Two new tests (`TestPickupRouteAlwaysReturnsRoute`, `TestPickupRouteSingleRoute`) run 1200+ iterations of `pickupRoute` and assert it never returns nil.

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -177,11 +177,11 @@ func callService(serviceUrl string, input []byte, headers http.Header) ([]byte, 
 }
 
 func pickupRoute(routes []v1alpha1.InferenceStep) *v1alpha1.InferenceStep {
-	randomNumber, err := rand.Int(rand.Reader, big.NewInt(101))
+	randomNumber, err := rand.Int(rand.Reader, big.NewInt(100))
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


SebTardif:
Closing as duplicate of #5489 which fixes the same off-by-one in pickupRoute with more comprehensive testing and is linked to #5408. Apologies for the duplicate.

PR #5501, opened 2026-05-07
fix: close file handles in agent storage and syncer

SebTardif:

Closes file handles that were opened but never closed in three locations within the model agent, preventing file descriptor leaks during model downloads and syncing.

1. **`pkg/agent/syncer.go` (`SyncModelDir`)**: `os.Open` is called for each model success file but the returned `*os.File` is never closed. Every success file discovered during sync leaks a file descriptor.

2. **`pkg/agent/storage/https.go` (`Download`, default case)**: `createNewFile` returns an `*os.File` that is written to via `io.Copy` but never closed, leaking a descriptor per single-file model download.

3. **`pkg/agent/storage/https.go` (`extractTarFiles`)**: `createNewFile` returns an `*os.File` for each extracted tar entry but never closes it. For large model archives with many files, this can exhaust the process file descriptor limit (`ulimit -n`, typically 1024), causing subsequent downloads to fail with "too many open files." The sibling `extractZipFiles` already closes files correctly (lines 226-228); this change matches that pattern.

- [x] All existing agent tests pass (17/17 in `pkg/agent`, 4/4 in `pkg/agent/storage`)
- [x] `go vet ./pkg/agent/...` passes with no warnings
- [x] `go build ./pkg/agent/...` compiles cleanly

The `extractTarFiles` fix follows the same close pattern already used by `extractZipFiles` in the same file: close on error, then close on success after `io.Copy`. The `syncer.go` fix uses `defer` which is appropriate since the Walk callback opens at most one file per invocation.

**Checklist**:

- [x] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)? N/A - no user-facing behavior change

```release-note
Fix file descriptor leaks in model agent: close file handles in SyncModelDir, HTTPS single-file download, and tar archive extraction.
```

### DIFF
--- a/pkg/agent/storage/https.go
+++ b/pkg/agent/storage/https.go
@@ -126,6 +126,7 @@ func (h *HTTPSDownloader) Download(client http.Client) error {
 		if err != nil {
 			return err
 		}
+		defer file.Close()
 		if _, err = io.Copy(file, resp.Body); err != nil {
 			return fmt.Errorf("unable to copy file content: %w", err)
 		}
@@ -277,7 +278,11 @@ func extractTarFiles(reader io.Reader, dest string) error {
 
 		limitReader := io.LimitReader(tr, DEFAULT_MAX_DECOMPRESSION_SIZE)
 		if _, err := io.Copy(newFile, limitReader); err != nil {
+			newFile.Close()
 			return fmt.Errorf("unable to copy contents to %s: %w", header.Name, err)
+		}
+		if closeErr := newFile.Close(); closeErr != nil {
+			return closeErr
 		}
 	}
 	return nil
--- a/pkg/agent/storage/https.go
+++ b/pkg/agent/storage/https.go
@@ -126,9 +126,12 @@ func (h *HTTPSDownloader) Download(client http.Client) error {
 		if err != nil {
 			return err
 		}
-		defer file.Close()
 		if _, err = io.Copy(file, resp.Body); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/agent/storage/https.go
```diff
@@ -126,6 +126,7 @@ func (h *HTTPSDownloader) Download(client http.Client) error {
 		if err != nil {
 			return err
 		}
+		defer file.Close()
```
vivekk16: since file is a write handle, consider using explicit close with error checking instead of `defer file.Close()`, to match the pattern used in extractTarFiles and extractZipFiles. `defer` discards the error from `Close()`, which on a write handle could silently hide a flush failure. Not a blocker, but would keep the approach consistent across the file.

vivekk16 (commented):
Thanks for the fix! The syncer.go and extractTarFiles changes look good.
One minor suggestion on the Download default case in https.go

pkg/agent/storage/https.go
SebTardif: Good catch, updated in bcf9ce289. Download now uses explicit close with error check, matching extractTarFiles.

vivekk16:
lgtm
@Jooho @spolti

SebTardif:
Still relevant. The file handle leaks in agent storage and syncer are real resource leaks on error paths. @Jooho @spolti any chance you could take a look when you have a moment?

PR #5502, opened 2026-05-07
fix(localmodel): use NodeSelector for jobs to fix PVC access

bartoszmajsak:

Download jobs for LocalModelCache use `NodeName` to pin pods to specific nodes, which bypasses the Kubernetes scheduler entirely. The [Node authorizer](https://kubernetes.io/docs/reference/access-authn-authz/node/) verifies that kubelets can only access PVCs related to pods bound to their node through a graph-based relationship model. When a pod is placed via `NodeName`, the scheduler never establishes this relationship, and the kubelet is denied access to the download PVC:

```
no relationship found between node 'minikube-m03' and this object
```

This has become more visible with Kubernetes 1.34, where the Node authorizer enforces stricter per-node scoping ([KEP-4601](https://github.com/kubernetes/enhancements/issues/4601), now stable). The `test-modelcache` E2E test was failing almost every run, blocking 19+ open PRs.

Switches to `NodeSelector` with `kubernetes.io/hostname`, which routes through the scheduler. The scheduler establishes the PVC-to-node binding that the Node authorizer requires, while still targeting the exact same node.

No migration concerns - download jobs are ephemeral (1h TTL), PVs/PVCs and on-disk models are unchanged, and no CRD changes are involved.

- [x] All 12 `localmodelnode` controller tests pass
- [x] All 17 `localmodel` reconciler tests pass
- [x] Added test assertion verifying `NodeSelector` is set and `NodeName` is empty on download jobs
- [x] `test-modelcache (kustomize)` E2E passes in CI

The switch from `NodeName` to `NodeSelector` changes scheduling behavior: the scheduler now evaluates taints and resource pressure on the target node. Previously `NodeName` silently bypassed these checks. This is strictly better - if a node is in trouble, you want the download job to stay `Pending` rather than being force-placed. The download job still targets the exact same node via the `kubernetes.io/hostname` label.

**Checklist**:

```release-note
Fix LocalModelCache download jobs failing on K8s 1.34+ by switching from NodeName to NodeSelector, allowing the scheduler to establish proper PVC-to-node binding for the Node authorizer.
```

### DIFF
--- a/pkg/controller/v1alpha1/localmodelnode/controller.go
+++ b/pkg/controller/v1alpha1/localmodelnode/controller.go
@@ -198,7 +198,7 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, localModelNode
 			TTLSecondsAfterFinished: &jobTTLSecondsAfterFinished,
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
-					NodeName:      nodeName,
+					NodeSelector:  map[string]string{"kubernetes.io/hostname": nodeName},
 					Containers:    []corev1.Container{*container},
 					RestartPolicy: corev1.RestartPolicyNever,
 					Volumes:       volumes,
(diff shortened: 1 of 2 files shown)


VedantMahabaleshwarkar (approved):
LGTM

sivanantha321 (approved):

PR #5507, opened 2026-05-07
feat(llmisvc): use targeted access-log flag with vLLM version fallback

KillianGolds:

Builds on #5154. vLLM 0.16 introduced [`--disable-access-log-for-endpoints`](https://docs.vllm.ai/en/latest/cli/serve/#-disable-access-log-for-endpoints) ([vllm-project/vllm#30011](https://github.com/vllm-project/vllm/pull/30011)), a targeted alternative to the blanket `--disable-uvicorn-access-log` that lets us silence access logs only for specific high-frequency endpoints. Switching the default LLM-D templates to it restores access log visibility for legitimate inference traffic (`/v1/completions`, `/v1/chat/completions`, etc.) while still preventing the log volume problem from EPP's 200ms `/metrics` scrape that #5154 fixed.

Each vLLM-invoking template now does a runtime `vllm --version` check at entrypoint and selects the new flag when vLLM ≥ 0.16, falling back to the existing blanket flag for older images. This keeps users on pre-0.16 vLLM (e.g. older pinned `llm-d-cuda` builds, or distributions that pin a specific vLLM version) working without changes. The detection block also emits two `[access-log-detect]` diagnostic lines so users can read pod logs to see exactly which version was parsed and which flag was selected.

The detection is hardened against noisy `vllm --version` failures: it pipes through `tail -1` to ignore prelude output, validates the extracted token with a `^[0-9]+\.[0-9]+` regex before doing the version compare, and falls back safely on any parse failure. This was added after a probe pod surfaced that some images dump multi-line `INFO` messages to stdout when `vllm --version` fails (e.g. running on a node without a usable GPU), which would otherwise leak garbage into the version comparison.

Endpoint list `/health,/metrics,/ping` matches the [vLLM docs example](https://docs.vllm.ai/en/latest/cli/serve/#-disable-access-log-for-endpoints). `install/*` release snapshots intentionally untouched. Charts and quick-install scripts regenerated via `make generate-chart-manifests` and `make generate-quick-install-scripts`.

Fixes #

Validated end-to-end on an OpenShift cluster with NVIDIA GPU Operator + KServe full install.

- [x] **Reproduced regression baseline.** Hit `/health`, `/metrics`, `/ping`, and `/v1/completions` against an unmodified-template pod. 0 access log entries for any of them, including the legitimate inference call. Confirms the loss-of-visibility issue this PR fixes.
- [x] **Test 1: default upstream image, vLLM 0.17.1+precompiled.** Detection picks the new flag. Diagnostic log line: `[access-log-detect] vllm version='0.17.1+precompiled'` → `selected ACCESS_LOG_ARGS='--disable-access-log-for-endpoints /health,/metrics,/ping'`. After hitting all four endpoints, `/v1/completions` logged exactly once, `/health`, `/metrics`, `/ping` stayed silent. The `+precompiled` suffix is a real-world example of a PEP 440 local-version identifier; the regex matches it because it only requires a leading `MAJOR.MINOR`.
- [x] **Test 2: forced-old vLLM (`vllm/vllm-openai:v0.15.0`, `v0.15.1`).** Detection takes the fallback branch. vLLM 0.15.x parses and accepts `--disable-uvicorn-access-log` (`'disable_uvicorn_access_log': True` in the engine's non-default args). Pods then crash on a CUDA driver mismatch unrelated to this change — proven via a control pod with no kserve scaffolding hitting the identical `Error 803` in `cudaGetDeviceCount`.
- [x] **Test 2 hardening probe.** Ran the hardened bash against `vllm/vllm-openai:v0.15.0` on a CPU-only node, where `vllm --version` exits non-zero and prints `INFO ... Triton not installed or not compatible; certain GPU-related functions will not be available.` to stdout. The regex correctly rejects the resulting `available.` token, so `ACCESS_LOG_ARGS` falls back to `--disable-uvicorn-access-log`. Confirms the noisy-failure fallback path.
- [x] **Test 3 hardening probe.** Same shape against a `0.11.0+rhai5` build on a CPU-only node where `vllm --version` exits non-zero with `INFO ... No platform detected, vLLM is running on UnspecifiedPlatform`. The regex correctly rejects `UnspecifiedPlatform`, fallback selected. The `+rhai5` suffix also confirms the regex tolerates PEP 440 local-version identifiers more generally.
- [x] **Forced-fallback runtime validation.** Pinned `ACCESS_LOG_ARGS="--disable-uvicorn-access-log"` on a working vLLM 0.17.1 image and re-ran the curl battery. 0 access log entries for any endpoint. Confirms the fallback flag's runtime silencing behaviour end-to-end on a vLLM that actually starts.

Happy to share specific log snippets and pod manifests on request.

- The runtime version-check is an intentional scope addition on top of the mechanical flag swap. The new flag is unrecognised by older vLLM CLI parsers, so passing it unconditionally would startup-error any deployment pinned to a pre-0.16 image. Detecting via `vllm --version` and a `sort -V` floor compare is more stable than introspecting vLLM's argparse internals, which can change shape across releases.
- The `[access-log-detect]` diagnostic echoes match the existing `[Infer RoCE] ...` pattern that already exists in these templates. Two lines per pod startup, no runtime impact, useful for operators debugging which path the detection took on a given image.
- Both flags coexist in vLLM 0.18.0 source ([`cli_args.py` lines 229 and 231](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/cli_args.py)), so the fallback path is forward-safe through at least the next vLLM minor release.
- No new unit/e2e tests added. The change is contained in the bash entrypoint of each template, which isn't currently covered by the Go test suite, and the cluster validation above provides the behavioural proof. Mirrors the testing approach of #5154 which made a similar mechanical template change. Happy to add a regression test asserting the rendered templates contain the expected bash markers if reviewers prefer.

**Checklist**:

**Release note**:

```release-note
Switch the default LLMInferenceServiceConfig templates from `--disable-uvicorn-access-log` to vLLM 0.16's targeted `--disable-access-log-for-endpoints /health,/metrics,/ping`, restoring access log visibility for legitimate inference traffic while continuing to prevent the log-volume issue from EPP's 200ms metrics scrape that #5154 addressed. A runtime `vllm --version` check falls back to the original blanket flag for vLLM < 0.16 images, so existing pre-0.16 deployments continue to work unchanged.
```

### DIFF
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -130,10 +130,20 @@ spec:
           fi
         fi
 
+        # --disable-access-log-for-endpoints landed in vLLM 0.16.0 (vllm-project/vllm#30011).
+        # Older versions still need the blanket --disable-uvicorn-access-log.
+        ACCESS_LOG_ARGS="--disable-uvicorn-access-log"
+        VLLM_VERSION=$(vllm --version 2>/dev/null | tail -1 | awk '{print $NF}')
+        echo "[access-log-detect] vllm version='${VLLM_VERSION}'"
+        if [[ "$VLLM_VERSION" =~ ^[0-9]+\.[0-9]+ ]] && [ "$(printf '%s\n%s\n' "0.16.0" "${VLLM_VERSION}" | sort -V | head -1)" = "0.16.0" ]; then
+          ACCESS_LOG_ARGS="--disable-access-log-for-endpoints /health,/metrics,/ping"
+        fi
+        echo "[access-log-detect] selected ACCESS_LOG_ARGS='${ACCESS_LOG_ARGS}'"
+
         eval "vllm serve /mnt/models \
           --served-model-name "{{ .Spec.Model.Name }}" \
           --port 8001 \
-          --disable-uvicorn-access-log \
+          ${ACCESS_LOG_ARGS} \
           {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }} \
           {{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt{{- end }} \
           {{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key{{- end }} \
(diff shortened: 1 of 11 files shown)


vivekk16 (commented):
lgtm

pierDipi (approved):

spolti (approved):

Issue #5480, opened 2026-04-30
Enable customization of storage-initializer configuration for LLMInferenceServices


brettmthompson:
/kind feature

Currently, there is no straightforward way to customize the storage-initializer init container configuration for LLMInferenceServices (v1alpha2). The controller always programmatically [creates and appends](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha2/llmisvc/workload_storage.go#L298-L299) a storage-initializer with default configuration, making it impossible to override specific settings like environment variables, resources, or image without fully disabling the automation.

When a user attempts to customize the storage-initializer via spec.template.initContainers or through LLMInferenceServiceConfig baseRefs, the controller appends a second storage-initializer instead of merging with the existing one, resulting in two init containers with the same name. This causes kuberenetes to reject pod creation with a validation error due to duplicate container names. 

As a workaround, users must explicitly disable the storage-initializer automation via `storageInitializer.enabled: false` and define a custom storage-initializer container spec in its entirety, manually replicating all default behavior including volume setup, credential injection, and CA bundle configuration.

**Describe the solution you'd like**
_Option 1: Enable storage-initializer merge support for llmisvc specs_

Modify the controller to check if storage-initializer already exists in `spec.template.initContainers` (from user spec or baseRefs merge) before appending. If found, strategically merge the user's configuration with defaults instead of appending a duplicate. Merge order would be defaults -> baseRefs -> user spec. The semantics of the `storageInitializer.enabled` field in the llmisvc spec would now be:
- true (default): Controller provides defaults and merges with user customizations
- false: Controller skips default injection, user handles everything 

_Option 2: ClusterStorageContainer support for llmisvcs_

Add ClusterStorageContainer lookup in the llmisvc controller, similar to how it's implemented for v1beta1 InferenceServices. The ClusterStorageContainer spec would then be merged with the default storage-initializer container for llmisvcs as well.

_Option 3: StorageInitializerConfig extension in the inferenceservice-config configmap_

Extend the StorageInitializerConfig struct to include additional fields for customization that get applied to all storage-initializers cluster-wide (or go a step further and allow customization based on uri prefix).

**Additional Context**
- This issue only affects LLMInferenceServices (v1alpha2), InferenceServices (v1beta1) use a webhook-based approach with ClusterStorageContainer support
[...]

PR #5510, opened 2026-05-08
fix(ci): stabilize HuggingFace model downloads in e2e tests

andresllh:

E2E tests that download models from HuggingFace were intermittently hanging or failing due to two compounding issues:

1. **XET client bugs** — \`hf-xet\` was locked at v1.1.3 (transitively via \`huggingface-hub\`) which has upstream bugs causing intermittent \`IncompleteMessage\` errors and infinite retries against the XET/xorbs CDN.

2. **Concurrent HF downloads** — all three llmisvc test jobs started simultaneously and each independently ran the storage-initializer to download \`facebook/opt-125m\` from HuggingFace, causing multiple concurrent XET sessions to compete under the 1-CPU limit.

3. **test-llm / test-huggingface-server-vllm failures** — the \`kserve-huggingfaceserver\` ClusterServingRuntime downloads models inline at pod startup via \`huggingface_hub\`, hitting the same XET CDN connectivity issues in the GHA runner network environment.

## Changes

**HuggingFace download stability:**
- **Pin \`hf-xet>=1.5.0\`** in \`python/storage/pyproject.toml\` and regenerate all lockfiles to pick up upstream XET bug fixes
- **Add \`TOKIO_WORKER_THREADS=1\`** to \`seaweedfs-init-job.yaml\` to limit tokio async worker thread concurrency under the 1-CPU init container limit

**llmisvc test pre-seeding (eliminates concurrent HF downloads):**
- **Add \`seed-model-cache\` job** — a discrete pre-download step that runs the storage-initializer once via \`docker run\`, packages the downloaded model as an OCI image, and uploads it as a GHA artifact
- **Load artifact in test jobs** — each test job downloads the artifact and loads it into minikube with \`minikube image load\`, then \`setup-kserve.sh\` uses it via \`OPT_125M_CACHE_IMAGE\` env var instead of re-downloading from HuggingFace
- **Add \`seaweedfs-init-job-from-cache.yaml\`** — alternative init job that copies model files from the pre-loaded OCI image into SeaweedFS, replacing the HF download init container

**HuggingFace server test overlay (test-only, does not affect production runtime):**
- **Add kustomize patch** at \`config/overlays/test/clusterresources/huggingfaceserver-tokio-patch.yaml\` that sets \`TOKIO_WORKER_THREADS=1\` and \`HF_HUB_DISABLE_XET=1\` on the \`kserve-huggingfaceserver\` ClusterServingRuntime — scoped to the test overlay only, production runtime is unchanged

Related to #5480

## Testing

- [x] All lockfiles regenerated with \`make precommit\`
- [x] \`seed-model-cache\` job passing in CI (49s); all three llmisvc test jobs passing
- [x] Fallback preserved: \`setup-kserve.sh\` still uses the original HF download path when \`OPT_125M_CACHE_IMAGE\` is unset (local dev unaffected)
- [x] Kustomize patch verified: \`kustomize build config/overlays/test/clusterresources\` confirms \`TOKIO_WORKER_THREADS\` and \`HF_HUB_DISABLE_XET\` are present alongside existing \`LMCACHE_USE_EXPERIMENTAL\` env var

\`\`\`release-note
NONE
\`\`\`

### DIFF
--- a/.github/workflows/e2e-test-llmisvc.yaml
+++ b/.github/workflows/e2e-test-llmisvc.yaml
@@ -121,9 +121,58 @@ jobs:
           compression-level: 0
           if-no-files-found: error
 
+  seed-model-cache:
+    runs-on: ubuntu-latest
+    needs: [llmisvc-image-build]
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Load KServe environment variables
+        run: ./kserve-images.sh --ci
+
+      - name: Free-up disk space
+        uses: ./.github/actions/free-up-disk-space
+
+      - name: Download storage-initializer image
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: ${{ env.STORAGE_INIT_IMG }}-${{ env.TAG }}
+          path: ./tmp
+
+      - name: Load storage-initializer image
+        run: docker load < ./tmp/${{ env.STORAGE_INIT_IMG }}-${{ env.TAG }}
+
+      - name: Download opt-125m from HuggingFace
+        timeout-minutes: 30
+        run: |
+          mkdir -p /tmp/model-cache && chmod a+rwx /tmp/model-cache
+          docker run --rm \
+            --network host \
+            -v /tmp/model-cache:/mnt/models \
+            -e HF_HUB_DISABLE_PROGRESS_BARS=1 \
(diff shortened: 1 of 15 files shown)


bartoszmajsak:
/lgtm

I hope this will work :)

jlost:
Looks good to me, we just need tests to pass. There are unresolved CI flakiness issues on master

jlost:
I think this is failing consistently for a reason. I want to try tweaking storage-initializer ENVs to make it work in CI. I'm going to try to fold it into #5614 due to the chicken-and-egg problem.

mwaykole (commented):
/lgtm

Jooho (approved):

PR #5512, opened 2026-05-08
fix(ingress): eliminate ExternalName service no-op update churn

bartoszmajsak:

`reconcileExternalService()` compares a minimal desired Service object against the fully-populated existing object using `equality.Semantic.DeepEqual`. This always fails because the desired object lacks server-populated fields (ClusterIP, UID, resourceVersion, managedFields).

Every reconcile cycle produces an API write that overwrites Labels and Annotations with nil - even when the Spec is identical. This strips metadata set by other controllers and generates unnecessary cross-controller reconciliation churn.

The fix replaces `DeepEqual` with `DeepDerivative`, which treats zero/nil fields in the desired object as "don't care". Server-populated metadata no longer causes false diffs. The unconditional Labels/Annotations overwrite is also removed - the reconciler should not be wiping metadata that other controllers set.

- [x] Unit test that creates an ExternalName service with labels and annotations from an external controller, reconciles with unchanged spec, and asserts metadata is preserved
- [x] Full ingress reconciler test suite passes

**Checklist**:

**Release notes**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/ingress_reconciler.go
@@ -337,7 +337,9 @@ func (ir *IngressReconciler) reconcileExternalService(ctx context.Context, isvc 
 			}
 
 			// Return if no differences to reconcile.
-			if equality.Semantic.DeepEqual(desired, existing) {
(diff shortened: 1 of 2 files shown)


spolti (commented):
/lgtm

spolti (approved):

PR #5521, opened 2026-05-12
feat(llmisvc): add model name based routing

pierDipi:
Add header-based model routing to LLMInferenceService HTTPRoute rules, enabling shared-gateway deployments where a single gateway dispatches to multiple models by the `X-Gateway-Model-Name` header (header name is configurable).

This header can be injected with components like llm-d BBR (aka IPP) or Gateway-specific native features.

Expand LoRA adapter matches dynamically in Go code after template processing so that adapter requests route through the same backend as the base model.

Key changes:
- Add model-based routing rules to the HTTPRoute template alongside existing path-based rules
- Add expandLoRAAdapterMatches to generate per-adapter header matches
- Add preferPathBasedURL for backward-compatible status.url selection
- Add address type names with -model-routing suffix for discovery
- Add e2e tests for model-based routing across /v1/completions, /v1/chat/completions, and /v1/models endpoints
- Fix extra_headers bug in wait_for_model_response where merged headers were discarded before the HTTP call

Follow ups:
- routing status update, after https://github.com/kserve/kserve/pull/5417 and https://github.com/kserve/kserve/pull/5414

```release-note
Add model-based routing support for LLMInferenceService, enabling shared-gateway deployments where a single gateway dispatches requests to multiple models using the `X-Gateway-Model-Name` header. LoRA adapter requests are automatically routed through the same backend as the base model by dynamically expanding header matches for each adapter defined in `.spec.model.lora.adapters`. 
```


pierDipi:
```
      Last State:  Terminated
        Reason:    OOMKilled
        Message:   2026-05-13 16:06:46.033 1 storage.initializer INFO [initializer-entrypoint:<module>():16] Initializing, args: (src_uri, dest_path): [('hf://Qwen/Qwen2.5-0.5B-Instruct', '/mnt/models')]
  2026-05-13 16:06:46.033 1 storage.initializer INFO [kserve_storage.py:download():161] Copying contents of hf://Qwen/Qwen2.5-0.5B-Instruct to local
```

docs/samples/llmisvc/opt-125m-cpu/llm-inference-service-facebook-opt-125m-cpu-no-scheduler.yaml
```diff
@@ -13,6 +13,8 @@ spec:
     containers:
       - name: main
         image: public.ecr.aws/q9t5s3a7/vllm-cpu-release-repo:v0.19.0
+        securityContext:
+          runAsNonRoot: false
```
bartoszmajsak, line 17: Why is this needed?

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -283,6 +283,26 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
...
+			if llmSvc.Spec.Model.LoRA != nil {
+				expandLoRAAdapterMatches(
+					llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules,
+					llmSvc.Namespace,
+					llmSvc.Spec.Model.LoRA.Adapters,
+					reconcilerConfig.ModelBasedRoutingHeaderName,
+				)
+			}
+		} else {
+			llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules = stripModelBasedRoutingRules(
```
bartoszmajsak, line 299: `X-Gateway-Model-Name` is not really a unique header name. Aren't we concerned we could strip too much in existing models where users provided their own `router.route.spec`?

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -283,6 +283,26 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
...
+		llmSvcCfg.Spec.Router.Route != nil &&
+		llmSvcCfg.Spec.Router.Route.HTTP.HasSpec() {
+		if r.isModelBasedRoutingEnabled(ctx, llmSvc, reconcilerConfig) {
+			if llmSvc.Spec.Model.LoRA != nil {
+				expandLoRAAdapterMatches(
+					llmSvcCfg.Spec.Router.Route.HTTP.Spec.Rules,
+					llmSvc.Namespace,
+					llmSvc.Spec.Model.LoRA.Adapters,
+					reconcilerConfig.ModelBasedRoutingHeaderName,
+				)
```
bartoszmajsak, line 296: Do we want to add LORAs also when users define their own `router.route.spec`?

docs/samples/llmisvc/opt-125m-cpu/llm-inference-service-facebook-opt-125m-cpu-no-scheduler.yaml
pierDipi: It doesn't work otherwise, image runs as root

bartoszmajsak: I mean in the scope if this PR :)

pkg/controller/v1alpha2/llmisvc/config_merge.go
pierDipi: I think so, they would be injected only when the custom injected rules include model routing matches.

managed provides automation, for full control, ref is our answer?

pkg/controller/v1alpha2/llmisvc/config_merge.go
pierDipi: maybe we can match by rule name suffix? any match kinda feels limited in their own way, any better alternative?

docs/samples/llmisvc/opt-125m-cpu/llm-inference-service-facebook-opt-125m-cpu-no-scheduler.yaml
pierDipi: yeah, I get annoyed and just fix things, it's existing forgotten when we switched back to runAsNonRoot = false being the default in well known configs

pkg/controller/v1alpha2/llmisvc/config_merge.go
bartoszmajsak: It might still collide on the exact header match, and as you mentioned in the other comment:

> managed provides automation, for full control, ref is our answer?

I think this simplifies things and protects us from weird edge cases, so I like this principle.

The only thing I wonder is if we should somehow signal what's potentially going to be altered - colliding rules or removal of some, as it all happens behind the scenes. I wonder if that would be better UX.

On the other hand it's in the spec of managed route, so maybe that's good enough?

pierDipi: > The only thing I wonder is if we should somehow signal what's potentially going to be altered - colliding rules or removal of some, as it all happens behind the scenes. I wonder if that would be better UX.

like an event with the stripped out rules + a descriptive message?

bartoszmajsak: Yes, that would probably help. But we can do it as a follow-up to avoid scope creep.

config/llmisvcconfig/config-llm-router-route.yaml
```diff
@@ -98,3 +196,18 @@ spec:
               timeouts:
                 backendRequest: 0s
                 request: 0s
+            - name: v1-catch-all-model-routing
```
bartoszmajsak, line 202: Assuming that reflects the intent - would it make sense to add short explanation for posterity?


```suggestion
            # Catch-all rules route to the workload Service directly (bypassing
            # InferencePool) for endpoints not handled by the scheduler, such as
            # /v1/models, /health, or custom paths exposed by the runtime.
            - name: v1-catch-all-model-routing
```

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -396,6 +416,87 @@ func isUsingTokenizerSidecar(spec v1alpha2.LLMInferenceServiceSpec) bool {
...
+		return false
+	}
+	switch cfg.ModelBasedRoutingMode {
+	case ModelBasedRoutingDisabled:
+		return false
+	case ModelBasedRoutingForced:
+		return true
+	default:
+		gateways, err := r.CollectReferencedGateways(ctx, llmSvc)
+		if err != nil {
```
bartoszmajsak, line 434: I'm fine with fail-open here, but maybe it's worth logging the error at least?

bartoszmajsak (approved):
non-blocking comments. Let me know your thoughts and we can do them as follow-ups to avoid gambling with flaky tests :)

config/llmisvcconfig/config-llm-router-route.yaml
pierDipi: Added to the first catch-all

PR #5523, opened 2026-05-12
fix(ci): stabilize version resolution in precommit

bartoszmajsak:

The precommit check was intermittently failing in CI because `generate-versions-from-gomod.py` re-resolved dependency versions from scratch on every run. For pseudo-versions (e.g. `gateway-api v1.4.2-0.xxx`), the script queries the GitHub releases API and HEAD-checks each candidate's download URL. A transient HEAD-check failure caused it to skip the correct version (`v1.4.1`) and fall back to an older one (`v1.4.0`), producing a spurious diff against the committed `kserve-deps.env`.

Two changes:

- **Prefer existing `kserve-deps.env` value for pseudo-versions** - when the go.mod version is a pseudo-version and the base version has no release artifact, the script now checks whether the existing value is in the same minor series and its URL still verifies. If so, it's reused without hitting the releases API. For real release versions, the full resolution path is preserved to catch actual drift.

- **Log URL check failures** - `check_url_exists` was silently swallowing all exceptions, making it impossible to diagnose why CI picked a wrong version. Failures are now printed so the root cause is visible.

- [x] Verified script resolves `GATEWAY_API_VERSION=v1.4.1` correctly with `--no-cache`
- [x] Verified URL check failure logging works (404 for non-existent v1.4.2 release is now printed)

The pseudo-version detection uses the same regex as `strip_pseudo_version` - if the stripped version differs from the raw version, it's a pseudo-version. Only pseudo-versions get the existing-value shortcut; real releases always go through full resolution to preserve drift detection. We could use `go list` to get full json of deps instead and know exactly what we are dealing with, but it's a bigger change.

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/setup/scripts/generate-versions-from-gomod.py
+++ b/hack/setup/scripts/generate-versions-from-gomod.py
@@ -106,7 +106,8 @@ def check_url_exists(url, timeout=5):
         req = Request(url, method="HEAD")
         with urllib.request.urlopen(req, timeout=timeout) as response:
             return 200 <= response.status < 400
-    except Exception:
+    except Exception as e:
+        print(f"    HEAD {url} failed: {e}")
         return False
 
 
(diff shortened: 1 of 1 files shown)


Jooho (approved):
/lgtm

PR #5524, opened 2026-05-12
feat(llmisvc): track gateway origin on each discovered address

bartoszmajsak:

When multiple Gateways serve an LLMInferenceService, there's no way to tell which URL came from which Gateway. `DiscoverURLs` flattens all gateway URLs into a single list, losing gateway identity. Consumers that need to select the endpoint associated with a specific Gateway have nothing to filter on.

This PR enriches each address in the status with an optional `origin` field (`gwapiv1.ObjectReference`) that records which networking resource produced it. A new `SourcedAddress` type extends `duckv1.Addressable` with origin tracking while keeping the JSON wire format backward-compatible - existing CRs deserialize cleanly with `origin` as nil.

Also fixes a pre-existing duplicate address bug: when a ClusterIP-backed gateway reports its service hostname in `Gateway.Status.Addresses`, both the external and internal discovery paths produce identical entries. The internal path now skips hosts already discovered through gateway status addresses.

Adds a `distro`-tagged extension point (`discoverAdditionalURLs`) for distribution-specific URL discovery.

**Sample status output (multi-gateway)**:
```yaml
status:
  url: "https://my-model.example.com/ns/my-model"
  addresses:
    - name: gateway-external
      url: "https://my-model.example.com/ns/my-model"
      origin:
        kind: Gateway
        name: inference-gateway
        namespace: istio-system
    - name: gateway-internal
      url: "https://inference-gateway.istio-system.svc.cluster.local/ns/my-model"
      origin:
        kind: Gateway
        name: inference-gateway
        namespace: istio-system
```

- [x] Unit tests for `DiscoverURLs` updated - all 58 subtests pass, including new Origin tracking assertions
- [x] Regression test for ClusterIP duplicate suppression (gateway status address matches backing service hostname)
- [x] Filter function tests updated for `DiscoveredURL` type
- [x] v1alpha1 <-> v1alpha2 conversion roundtrip tests in both directions (Origin loss is acceptable - controller repopulates on next reconcile)
- [x] e2e test assertions for address origin validation (exact gateway match for custom gateway tests)
- [x] `go build ./...` compiles cleanly
- [x] `make manifests` - CRD schema includes `origin` under `addresses` with `x-kubernetes-list-type: atomic`

The `duckv1.AddressStatus` embed is replaced with explicit `Address` and `Addresses` fields. This is acceptable because (a) the API is alpha, (b) kserve controls the only conversion webhook, (c) no Knative duck-typing is used on LLMInferenceService. The `Address` (singular) field is retained for CRD schema stability but marked deprecated - it was never populated.

Origin uses `gwapiv1.ObjectReference` (Gateway API's standard reference type) rather than `corev1.TypedObjectReference` - cleaner field names (`group` not `apiGroup`), value types instead of pointers, no PVC-related doc baggage, and already imported throughout the codebase.

**Checklist**:

**Release note**:
```release-note
Add origin tracking to LLMInferenceService addresses, enabling consumers to identify which Gateway produced each URL. Fix duplicate address entries for ClusterIP-backed gateways.
```


pierDipi (approved):
/lgtm
/approve

pierDipi:
there are conflicts

PR #5525, opened 2026-05-12
fix(llmisvc): wire enableLLMInferenceServiceTLS in reconcile pipeline

madhugoutham:

`enableLLMInferenceServiceTLS` was added in #5249 but the status URL path in `router.go` was hardcoded to `https://` regardless of the flag value. This PR wires the status URL to respect the TLS toggle and adds test coverage for the TLS on/off paths.

When `enableLLMInferenceServiceTLS: false`:
- Self-signed cert is still generated (unused, keeps volume mounts stable)
- Workload service port switches from `https` to `http` (correct protocol detection)
- Scheduler and P/D sidecar template flags render correctly based on TLS state
- Status URL reports `http://` instead of `https://`

When `enableLLMInferenceServiceTLS` is unset or `true`: behavior is unchanged.

- [x] Template rendering tests: scheduler TLS flags (`--secure-serving`, `--enable-cert-reload`, `--model-server-metrics-scheme`, `--cert-path`) with EnableTLS=true and EnableTLS=false
- [x] Template rendering tests: P/D sidecar flags (`--decoder-use-tls`, `--prefiller-use-tls`) with both TLS states
- [x] Envtest integration test: full controller reconcile with `enableLLMInferenceServiceTLS: false`, verifies cert secret presence, service port name `http`, status URL `http://` scheme
- [x] 214 existing envtest tests pass (no regressions)
- [x] `go vet` clean, `go build` clean
- [x] `make generate-chart-manifests` produces no additional diff

- **Cert kept unconditional (Option 2):** Per review discussion, cert generation stays unconditional. When TLS is off the cert is unused but present. This avoids `optional` field complexity and deployment rollout churn on TLS toggle.
- The envtest fixture ConfigMap now explicitly sets `enableLLMInferenceServiceTLS: true` to match the TLS-on test path. This ensures envtest exercises TLS-on behavior.
- **Default ConfigMap note:** The Helm chart's default `inferenceservice-config` does not include `enableLLMInferenceServiceTLS`, so it defaults to `false` (Go zero value). This matches the upstream design from #5249 (BackendTLSPolicy not yet implemented). Admins who need TLS can set the flag explicitly in their ConfigMap.
- `SSL_CERT_DIR` in scheduler config is unconditionally set, tracked separately in #5537.

**Checklist**:

- [x] Have you added unit/integration tests that prove your fix is effective or that this feature works?

**Release note**:
```release-note
Wire enableLLMInferenceServiceTLS through status URL path and add TLS toggle test coverage.
```

### DIFF
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -297,4 +297,4 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
-          optional: '{{ if .GlobalConfig.EnableTLS }}false{{ else }}true{{ end }}'
+          optional: true
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -268,9 +268,9 @@ spec:
           - "--enable-ssrf-protection=true"
           - "--pool-group=inference.networking.x-k8s.io"
           - '{{ if .GlobalConfig.EnableTLS }}--secure-proxy=true{{else}}--secure-proxy=false{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- else }}--cert-path={{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- else }}--decoder-use-tls=false{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- else }}--prefiller-use-tls=false{{- end }}'
         volumeMounts:
           - mountPath: /var/run/kserve/tls
             name: tls-certs
@@ -297,3 +297,4 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
+          optional: true
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -297,4 +297,4 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
-          optional: {{ if .GlobalConfig.EnableTLS }}false{{ else }}true{{ end }}
+          optional: '{{ if .GlobalConfig.EnableTLS }}false{{ else }}true{{ end }}'
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -297,4 +297,4 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
-          optional: true
+          optional: {{ if .GlobalConfig.EnableTLS }}false{{ else }}true{{ end }}
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -268,9 +268,9 @@ spec:
           - "--enable-ssrf-protection=true"
           - "--pool-group=inference.networking.x-k8s.io"
           - '{{ if .GlobalConfig.EnableTLS }}--secure-proxy=true{{else}}--secure-proxy=false{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- else }}--cert-path={{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- else }}--decoder-use-tls=false{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- else }}--prefiller-use-tls=false{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- end }}'
         volumeMounts:
           - mountPath: /var/run/kserve/tls
             name: tls-certs
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -297,4 +297,3 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
-          optional: true
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -163,7 +163,7 @@ spec:
           - name: tls-certs
             secret:
               secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
-              optional: true
+              optional: {{ if .GlobalConfig.EnableTLS }}false{{ else }}true{{ end }}
           - name: tokenizer-uds
             emptyDir: {}
           # Writable /tmp for the tokenizer container (see volumeMounts comment above).
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -72,10 +72,10 @@ spec:
               - "9002"
               - --grpc-health-port
               - "9003"
-              - '{{ if .GlobalConfig.EnableTLS }}--enable-cert-reload=true{{- else }}--enable-cert-reload=false{{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--enable-cert-reload=true{{- end }}'
               - '{{ if .GlobalConfig.EnableTLS }}--secure-serving=true{{- else }}--secure-serving=false{{- end }}'
-              - '{{ if .GlobalConfig.EnableTLS }}--model-server-metrics-scheme=https{{- else }}--model-server-metrics-scheme=http{{- end }}'
-              - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- else }}--cert-path={{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--model-server-metrics-scheme=https{{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- end }}'
             env:
               - name: SSL_CERT_DIR
                 value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -163,7 +163,7 @@ spec:
           - name: tls-certs
             secret:
               secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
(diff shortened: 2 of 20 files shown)


madhugoutham:
@pierDipi could you review this when you get a chance?

pierDipi:
@madhugoutham can you describe why the global flag is not enough for your use and we need the per service API? 

if your admins decide to force service mesh it's usually that you want the full cluster to use service mesh and per service config is just not helpful as individual services can disable or enable TLS which could break one way or another.

madhugoutham:
> @madhugoutham can you describe why the global flag is not enough for your use and we need the per service API?
> 
> if your admins decide to force service mesh it's usually that you want the full cluster to use service mesh and per service config is just not helpful as individual services can disable or enable TLS which could break one way or another.

I went with per-CR thinking it gives operators a safer rollout path disable on one service, validate with the mesh, then go cluster-wide. But you're right that TLS scope should be an admin-level decision. I'll rework to global-only the existing enableLLMInferenceServiceTLS isn't fully wired on master. Cert gen, port name, and status URL all ignore it. I'll fix those paths and drop spec.tls.

madhugoutham:
@pierDipi Addressed feedback reworked to global-only. PTAL when you get a chance.

config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -259,9 +259,9 @@ spec:
           - "--enable-ssrf-protection=true"
           - "--pool-group=inference.networking.x-k8s.io"
           - '{{ if .GlobalConfig.EnableTLS }}--secure-proxy=true{{else}}--secure-proxy=false{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- end }}'
-          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--cert-path=/var/run/kserve/tls{{- else }}--cert-path={{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--decoder-use-tls=true{{- else }}--decoder-use-tls=false{{- end }}'
+          - '{{ if .GlobalConfig.EnableTLS }}--prefiller-use-tls=true{{- else }}--prefiller-use-tls=false{{- end }}'
```
pierDipi: do we need the else?

madhugoutham: Without the else, the template renders an empty string arg ('') in the container args the YAML list item stays, just empty. The else keeps it explicit and consistent with the existing --secure-proxy pattern on the line above, which already uses true/false.

config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -288,3 +288,4 @@ spec:
       - name: tls-certs
         secret:
           secretName: "{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}"
+          optional: true
```
pierDipi: shouldn't we gate the optional true/false based on the flag?

madhugoutham: The volume block is always rendered by the template, but when TLS is off the cert secret gets deleted so without optional the pod crashes referencing a non-existent secret. That said, gating it makes sense so TLS-on still fails loudly: optional: `{{ if .GlobalConfig.EnableTLS}}false{{ else }}true{{ end }}`.

madhugoutham: Gated it across all templates

config/llmisvcconfig/config-llm-decode-template.yaml
pierDipi: secure proxy is different because the default is/used to be `true` while these ones can rely on defaults and so empty is the non-TLS case

madhugoutham: Makes sense those flags default to the non-TLS value when omitted, so the if-only form is sufficient. Reverted those, kept the else only for --secure-serving and --secure-proxy since both default to `true`.

madhugoutham:
Pushed a fix for the CI failures the optional template value needed quotes for kustomize compatibility

pierDipi:
@madhugoutham I don't think that will work as optional is boolean and not a string

madhugoutham:
@pierDipi The only option is to keep `optional: true` unconditionally can't gate it in the template since it's a boolean field. Reverting the gating.

config/llmisvcconfig/config-llm-decode-template.yaml
madhugoutham: optional is a boolean so it can't be gated in the template keeping `optional: true` since the controller always creates the cert before the deployment.

pierDipi:
optional true when it's not really optional carries the risk of having volume not bound when it's actually needed

There are other options:
1) remove `optional: true` in configs and remove the volume when `enable TLS == false` in the controller as part of the `expected*` for workloads, etc
2) remove `optional: true` in configs, always create the secret and only reconfigure arguments (equivalent to what it's currently done on master)
3) set optional = true only when `enable TLS == false` in the controller as part of the `expected*` for workloads, etc

madhugoutham:
Going with option 2 keep the cert always present and only reconfigure the args. Removes the cert deletion and the optional field entirely. Closest to master behavior and avoids the template boolean limitation.

Option 1 adds controller code to manipulate volumes which couples template and controller maintenance. Option 3 has the same coupling plus a nil vs &false diff that triggers unnecessary deployment rollouts on upgrade.

madhugoutham:
@pierDipi Pushed a precommit fix, quick-install scripts were out of sync. Verified on Kind + Istio with TLS disabled, all checks pass.

madhugoutham:
@pierDipi could you approve the pending workflow on the Actions tab?

madhugoutham:
@pierDipi @Jooho let me know if anything else needs attention.

mwaykole (commented):
Test review comment

mwaykole (commented):
## PR Review: `fix(llmisvc): wire enableLLMInferenceServiceTLS in reconcile pipeline`

**Risk Level: MEDIUM** — Behavioral change to a core reconcile path, safe for existing TLS-on installs but carries a backward-compatibility risk for new installs and a double-API-server-hit per reconcile.

### What this PR does well ✅
- The fix is **correct and complete** for the three paths it targets (status URL scheme, service port name, scheduler args).
- **Cert kept unconditional** (Option 2) is the right call — avoids pod-restart churn on TLS toggle and optional-field complexity.
- Both unit tests (`TestReplaceVariables_TLSConditional`, `TestReplaceVariables_PDSidecarTLSConditional`) and the integration test (`controller_int_tls_test.go`) are well-structured.
- The hack/setup scripts and Helm chart file are consistently updated.
- The fixture ConfigMap correctly adds `enableLLMInferenceServiceTLS: true` to avoid masking the old bug.

### Required changes 🔴

1. **Backward-compatibility breakage for new installs** — The Helm chart default ConfigMap does NOT set `enableLLMInferenceServiceTLS`. Go zero-value for `bool` is `false`, so every new install will default to TLS **off**. The PR description mentions this as a follow-up, but shipping this fix without the ConfigMap update means anyone doing a fresh install will silently lose TLS. This must be addressed before merge.

2. **`updateRoutingStatus` should accept `cfg *Config` as a parameter** — The newly added `LoadConfig` call (line 317) and the pre-existing one (~line 339) both perform API server hits per reconcile. `reconcileRouter` already receives `cfg *Config` — thread it through to `updateRoutingStatus` and remove both internal `LoadConfig` calls.

3. **Integration test `It` description is contradictory** — Named `"should not create cert secret"` but asserts the cert secret IS created. Rename to accurately reflect behavior.

### Non-blocking observations 🟡

4. **`--enable-cert-reload` has no `else` branch** — When TLS is off, the template renders an empty string `""` for this arg while `--secure-serving` gets an explicit `false`. Inconsistent. Add `{{- else }}--enable-cert-reload=false{{- end }}` for parity, or document why the asymmetry is intentional.

5. **`Eventually` blocks lack explicit timeouts** — Add `.WithTimeout(30 * time.Second).WithPolling(500 * time.Millisecond)` to match other controller integration tests and avoid flakes on slow envtest reconciliation.

6. **Template YAML line wrapping** — The Helm/hack script files wrap `--secure-serving=false{{-\n            end }}` awkwardly. Use the single-line form from `config/llmisvcconfig/config-llm-scheduler.yaml` for consistency.

7. **`SSL_CERT_DIR` unconditional** — Acknowledged in the PR description, tracked in #5537. Fine to defer.

---

**Verdict: REQUEST_CHANGES** — Items 1-3 are required before merge. Item 1 is the most critical: this fix makes the missing Helm ConfigMap default load-bearing for production installs.

madhugoutham:
Thanks for the detailed review @mwaykole.

  **1. ConfigMap default:** `enableLLMInferenceServiceTLS` was introduced upstream in #5249 without a ConfigMap default. Before this PR the flag was not fully wired, so the missing default had no impact. This PR completes the wiring, which makes the missing default load-bearing. Will add `"enableLLMInferenceServiceTLS": true` to the default ConfigMap in the next push.

  **2. Double LoadConfig:** The two calls are in mutually exclusive branches. The first (line 317) is inside an early-return block that exits at line 332 via `return nil, nil`, so it never reaches the second call at line 338. Only one executes per reconcile. The existing call at line 338 is upstream code with a TODO for caching. Changing the function signature would diverge from upstream master and expand the scope of this PR.

  **3. Test description:** Will update to `"should keep cert secret and should use HTTP port name"` in the next push.

  **4. --enable-cert-reload:** The asymmetry is intentional. This flag defaults to `false` in the scheduler binary, so
  omitting it produces the correct behavior. `--secure-serving` defaults to `true`, which is why it needs the explicit else branch. This was covered in an earlier review thread.

  **5. Eventually timeouts:** Following existing test conventions in this package. No other envtest in `llmisvc/` sets explicit timeouts.

  **6. Template line wrapping:** The multi-line form in `resources.yaml` and hack scripts is auto-generated by `make precommit`. The source template in `config/llmisvcconfig/` is single-line.

  **7. SSL_CERT_DIR:** Tracked in #5547.

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -73,7 +73,7 @@ spec:
               - --grpc-health-port
               - "9003"
               - '{{ if .GlobalConfig.EnableTLS }}--enable-cert-reload=true{{- end }}'
-              - '{{ if .GlobalConfig.EnableTLS }}--secure-serving=true{{- end }}'
+              - '{{ if .GlobalConfig.EnableTLS }}--secure-serving=true{{- else }}--secure-serving=false{{- end }}'
```
pierDipi: it seems CI is failing with [1], the major change is this one, it might be related and maybe explains why the else wasn't set before

[1] ```FAILED llmisvc/test_llm_inference_service.py::test_llm_inference_service[cluster_cpu-cluster_single_node-router-managed-workload-single-cpu-model-fb-opt-125m] - AssertionError: Service returned 500:```

madhugoutham: The --secure-serving else clause only takes effect when enableLLMInferenceServiceTLS is explicitly set to false in the ConfigMap. The default path (field absent or true) renders the same --secure-serving=true as before. CI uses the default ConfigMap, so this change is not exercised in these test runs. The failures are repo-wide, PRs #5596 and #5594 have the same test-llmisvc errors.

madhugoutham: > it seems CI is failing with [1], the major change is this one, it might be related and maybe explains why the else wasn't set before
> 
> [1] `FAILED llmisvc/test_llm_inference_service.py::test_llm_inference_service[cluster_cpu-cluster_single_node-router-managed-workload-single-cpu-model-fb-opt-125m] - AssertionError: Service returned 500:`

Checked the logs. The 500 comes from FailedMount: secret "...-kserve-self-signed-certs" not found, which prevents the pod from starting. The same FailedMount error appears on PR #5596 : [job logs](https://github.com/kserve/kserve/actions/runs/26574709300/job/78291861153). This is a pre-existing cert timing issue in CI, not related to the --secure-serving else clause.

madhugoutham:
Addressed review feedback:
- Added `"enableLLMInferenceServiceTLS": true` default to ConfigMap, chart, and test overlay
- Fixed stale test description

pierDipi:
> * Added `"enableLLMInferenceServiceTLS": true` default to ConfigMap, chart, and test overlay

I don't think this will work

madhugoutham:
@pierDipi I traced this further through the standalone chart path.

I'm going to remove `enableLLMInferenceServiceTLS: true` from the production ConfigMaps and revert `--secure-serving` back to if-only. Without a BackendTLSPolicy/transport_socket, enabling TLS by default would make the backend serve HTTPS without a trusted TLS path from Envoy. The `--secure-serving` else clause also overrides the EPP binary default (SecureServing=true) when the field is absent from the ConfigMap, breaking the ext_proc connection.

The rest of the toggle wiring stays: cert gating, port name, status URL, and template flags. Test overlay and envtest fixture keep `enableLLMInferenceServiceTLS: true` for test coverage. Test description fix included.

CI: builds, precommit, and unit tests pass. E2E llmisvc failures are pre-existing FailedMount cert timing, same on #5596 and #5594. CI does not exercise the gateway to backend TLS path.

Does this approach work for you?

madhugoutham:
@spolti @Jooho PTAL for /approve.

PR #5181, opened 2026-03-05
feat: add id2label support

avishay-akamai:
**What this PR does / why we need it**:

- Adds optional support in the HuggingFace encoder server to return human-readable labels (from `model_config.id2label`) instead of numeric class indices for classification tasks.
- Introduces a new CLI bool flag `--use_id2label` to enable this behavior.
- Improves output readability for:
   - `return_probabilities=True` (probability dict keys become labels).
   - `return_raw_logits=True` (logit dict keys become labels).
   - default sequence classification output (predicted class index becomes label).
- Includes a safe helper (`_get_label_or_index`) that:
   - falls back to indices if id2label is missing/empty.
   - handles id2label keys that may be int or str.

**Which issue(s) this PR fixes**:

This PR completes and replaces the unmerged work from https://github.com/kserve/kserve/pull/4444, which introduced the requested id2label support. Users asked to have this finalized.

**Feature/Issue validation/testing**:

- [x] Unit tests: added `python/huggingfaceserver/tests/test_encoder_label_mapping.py`
- [x] Manual test: run huggingfaceserver with `--use_id2label` and a sequence-classification model that defines `id2label`
- Logs: N/A

**Feature/Issue validation/testing**:

- The change is gated behind `--use_id2label` and defaults to the existing behavior (numeric indices).
- Label mapping is only applied when `model_config.id2label` is present; otherwise it safely falls back to indices.

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:
```release-note
HuggingFace encoder server: add optional `--use_id2label` support to return model `id2label` strings instead of numeric class indices for classification outputs (predictions/probabilities/raw logits).
```

PR #5526, opened 2026-05-13
feat: add id2label support

oleksii-donets:

Adds an opt-in `--use_id2label` CLI flag to the HuggingFace encoder server so that sequence (text) classification outputs can be returned with the model's human-readable `id2label` strings instead of numeric class indices.

When enabled, label mapping is applied to:
- the default top-1 prediction (argmax),
- `--return_probabilities` dict keys,
- `--return_raw_logits` dict keys.

The helper falls back to numeric indices when the model's `config.id2label` is missing or empty, so behavior is unchanged for models without a labels mapping. The flag itself defaults to off, so existing deployments are unaffected.

> Note: label strings come verbatim from the model's `config.id2label` and therefore vary per model — e.g. `POSITIVE` / `NEGATIVE` for `distilbert-sst2`, `LABEL_0` / `LABEL_1` for models that never set real labels (such as `textattack/bert-base-uncased-yelp-polarity` used in the E2E test), `5 stars` for `nlptown/bert-base-multilingual-uncased-sentiment`. Clients wanting consistent labels across models must normalize on their side.

Supersedes #5181 — continuation of @avishay-akamai's work after the original PR went stale. Authorship of the original feature commit is preserved in the git history. This PR rebuilds it on top of current `master` and addresses the open review feedback:

- **Fixes the failing `build (3.10|3.11|3.12)` jobs from #5181.** Two unit tests (`test_get_label_or_index_with_valid_labels`, `test_multiple_samples_label_mapping`) call the default (non-probability) prediction path and expect mapped label strings; the implementation now applies `_get_label_or_index` there too (this also matches Copilot's suppressed low-confidence review comment on the original PR).
- **Tightens the `--use_id2label` CLI help text** to match what's actually implemented (sequence/text classification), call out the int → label fallback, and warn that label strings come from the model.
- **Updates `_process_sequence_classification`'s return-type annotation** from `List[Union[int, Dict]]` to `List[Union[int, str, Dict]]` and refreshes the docstring.
- **Drops misleading test comments** that referred to `_get_label_or_index` as a "nested function created in postprocess" — it is a dedicated instance method.

The E2E test added in #5181 is kept as-is: it uses `textattack/bert-base-uncased-yelp-polarity` (which has the HF-default `LABEL_0` / `LABEL_1` placeholders) and asserts exactly those strings, so it tests the mechanism without coupling to a specific label vocabulary.

- [x] Unit tests: `python/huggingfaceserver/tests/test_encoder_label_mapping.py` — all 7 tests pass locally.
- [x] Regression check: `tests/test_token_classification_offsets.py` still passes (other encoder paths unaffected).
- [x] Lint / format: `ruff check` and `ruff format --check` clean on touched files.
- [x] E2E test: `test_huggingface_v2_sequence_classification_with_id2label` (already added in #5181) exercises `--use_id2label` + `--return_probabilities` against a real model.

## Checklist

## Release note

\`\`\`release-note
HuggingFace encoder server: add optional `--use_id2label` flag to return model `id2label` strings instead of numeric class indices for sequence/text classification outputs (top-1 prediction, return_probabilities, return_raw_logits). Falls back to numeric indices when the model config has no id2label mapping.
\`\`\`


gavrissh (approved):
Thanks, I think this is the 3rd PR with similar change

PR #5528, opened 2026-05-13
fix(ci): collect pod logs on fail, tune maxfail

jlost:

Improves failure diagnostics in the LLMInferenceService e2e test suite by collecting pod and init-container logs (including storage-initializer) inline with the failing test's output, and tunes `--maxfail` so that failing runs abort before the job timeout kills the process — preserving those diagnostics.

### Diagnostic log collection

- **`collect_pod_logs()`** in `diagnostic.py`: emits logs for every pod matching the test's labels -- init containers and regular containers, current and previous restart (`--previous` for CrashLoopBackOff). Called from `_collect_diagnostics` on failure.
- **`logger` instead of `print`**: routes diagnostic output through Python's logging module so it is correctly captured under pytest-xdist (`-n N --dist worksteal`). Bare `print()` from xdist workers is often lost or misattributed; `--log-cli-level=INFO` captures logger output reliably.
- **Resilient `_collect_diagnostics`**: wraps `get_llmisvc()` in try/except so diagnostics still run even when the LLMISVC object was never successfully created.

### Maxfail tuning

Makes `--maxfail` configurable via `PYTEST_MAXFAIL` (default 5, down from 10) so failing runs abort sooner and save CI minutes. For serial or low-parallelism jobs where 5 timeout-failures would exceed the job's `timeout-minutes` — causing GitHub Actions to kill the process and lose all diagnostic output — tighter overrides are set in the workflow YAML:

| Job | Parallelism | Per-test timeout | MAXFAIL | Worst-case fail time | Job timeout |
|---|---|---|---|---|---|
| vllm | 1 (serial) | 10 min | 3 | 30 min | 45 min |
| llmisvc | 0 (serial) | 15 min | 3 | 45 min | 60 min |
| llmisvc autoscaling HPA | 0 (serial) | 21 min | 2 | 42 min | 60 min |
| llmisvc autoscaling KEDA | 0 (serial) | 21 min | 2 | 42 min | 60 min |

All other jobs have sufficient parallelism (typically 6) that 5 failures overlap in a single batch and finish well within their timeout.

**Which issue(s) this PR fixes**: N/A

- Builds on the `_collect_diagnostics` / `print_all_events_table` pattern already in the llmisvc suite
- `collect_pod_logs` retrieves `--previous` logs when `restart_count > 0`, useful for init-container CrashLoopBackOff
- The maxfail values were derived by computing `ceil(maxfail / parallelism) * per_test_wait_timeout` and ensuring it fits within `timeout-minutes`

**Checklist**:

```release-note
NONE
```


jlost:
@bartoszmajsak @pierDipi What do you think?

pierDipi (approved):
/lgtm

PR #5533, opened 2026-05-14
fix(llmisvc): set workload service appProtocol based on TLS config

### LINKED ISSUE #5531: LLMInferenceService workload Service uses appProtocol: https for non-TLS vLLM backend, causing Traefik Gateway API upstream TLS failure
/kind bug

## Title

LLMInferenceService workload Service uses `appProtocol: https` for non-TLS vLLM backend, causing Traefik Gateway API upstream TLS failure

## What steps did you take and what happened?

I deployed a `LLMInferenceService` using the KServe LLMInferenceService controller with Traefik Gateway API.

This issue occurs when deploying a vLLM workload without TLS. The vLLM backend is intentionally configured to serve plain HTTP, but the LLMInferenceService controller still generates the workload Service with `name: https` and `appProtocol: https`.

The generated `HTTPRoute` correctly points to the workload Service on port `8000`:

```yaml
backendRefs:
- group: ""
  kind: Service
  name: facebook-opt-125m-single-kserve-workload-svc
  port: 8000
```

The generated workload Service selects the correct vLLM pod and the EndpointSlice is ready:

```yaml
endpoints:
- addresses:
  - 10.233.126.108
  conditions:
    ready: true
    serving: true
    terminating: false
ports:
- appProtocol: https
  name: https
  port: 8000
  protocol: TCP
```

However, the generated workload Service declares the backend protocol as HTTPS:

```yaml
ports:
- appProtocol: https
  name: https
  port: 8000
  protocol: TCP
  targetPort: 8000
```

The actual vLLM server behind this Service is serving plain HTTP on port `8000`.

From inside the cluster:

```bash
curl -v http://facebook-opt-125m-single-kserve-workload-svc:8000/v1/models
```

returns `200 OK`:

```text
< HTTP/1.1 200 OK
< server: uvicorn
```

But HTTPS fails:

```bash
curl -vk https://facebook-opt-125m-single-kserve-workload-svc:8000/v1/models
```

with:

```text
TLS connect error: error:0A00010B:SSL routines::wrong version number
curl: (35) TLS connect error: error:0A00010B:SSL routines::wrong version number
```

With Traefik Gateway API, the Service `appProtocol: https` appears to make Traefik treat the workload Service as an HTTPS upstream and attempt a TLS connection to the backend. Since the vLLM workload is not configured with TLS, this causes requests through the Gateway to fail due to backend protocol mismatch.

To confirm the cause, I temporarily patched the generated workload Service from:

```yaml
name: https
appProtocol: https
```

to:

```yaml
name: http
appProtocol: http
```

After that, the request through the Gateway succeeded:

```bash
curl -v http://172.16.200.184/kserve-test/facebook-opt-125m-single/v1/models
```

Result:

```text
< HTTP/1.1 200 OK
< Content-Type: application/json
< Server: uvicorn
```

The response contains the expected vLLM model list.

The patched Service showed the HTTP port name:

```bash
kubectl describe svc -n kserve-test facebook-opt-125m-single-kserve-workload-svc
```

```text
Name:                     facebook-opt-125m-single-kserve-workload-svc
Namespace:                kserve-test
Selector:                 app.kubernetes.io/name=facebook-opt-125m-single,app.kubernetes.io/part-of=llminferenceservice,kserve.io/component=workload
Type:                     ClusterIP
IP:                       10.233.58.177
Port:                     http  8000/TCP
TargetPort:               8000/TCP
Endpoints:                10.233.126.108:8000
```

When the LLMInferenceService controller reconciles the resource again, it restores the Service back to `name: https` and `appProtocol: https`, so this workaround is temporary.

After the LLMInferenceService controller was restarted and running again:

```bash
kubectl get po -n kserve
```

```text
NAME                                          READY   STATUS    RESTARTS   AGE
llmisvc-controller-manager-5cfbd6444f-2zr9z   1/1     Running   0          36s
```

the Gateway request started failing again:

```bash
curl -v http://172.16.200.184/kserve-test/facebook-opt-125m-single/v1/models
```

```text
*   Trying 172.16.200.184:80...
* Connected to 172.16.200.184 (172.16.200.184) port 80
> GET /kserve-test/facebook-opt-125m-single/v1/models HTTP/1.1
> Host: 172.16.200.184
> User-Agent: curl/8.7.1
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
< Date: Thu, 14 May 2026 07:18:27 GMT
< Content-Length: 21
```

This indicates that the manual Service patch is only a temporary workaround. Once the LLMInferenceService controller reconciles the workload Service again, the Service protocol metadata is restored to `name: https` / `appProtocol: https`, and the Gateway path fails again.

The issue seems to come from the LLMInferenceService controller creating the workload Service with `Name: "https"` and `AppProtocol: ptr.To("https")` even though this vLLM backend is deployed without TLS and serves HTTP on port `8000`.

Relevant generated Service:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: facebook-opt-125m-single-kserve-workload-svc
  namespace: kserve-test
spec:
  ports:
  - appProtocol: https
    name: https
    port: 8000
    protocol: TCP
    targetPort: 8000
  selector:
    app.kubernetes.io/name: facebook-opt-125m-single
    app.kubernetes.io/part-of: llminferenceservice
    kserve.io/component: workload
  type: ClusterIP
```

## What did you expect to happen?

When deploying a vLLM workload without TLS, the generated workload Service should advertise HTTP, not HTTPS:

```yaml
ports:
- appProtocol: http
  name: http
  port: 8000
  protocol: TCP
  targetPort: 8000
```

The controller should only set `appProtocol: https` when the generated workload is actually configured to serve TLS.

With a non-TLS vLLM backend, Gateway API implementations such as Traefik should be able to route HTTP traffic to the workload Service without attempting HTTPS to the backend.

## What's the InferenceService yaml?

This is using `LLMInferenceService`, not the classic `InferenceService`.

```bash
kubectl get llmisvc facebook-opt-125m-single -n kserve-test -oyaml
```

```yaml
kind: LLMInferenceService
metadata:
  namespace: kserve-test
spec:
  model:
    name: facebook/opt-125m
    uri: hf://facebook/opt-125m
  replicas: 1
  router:
    route: {}
  template:
    containers:
    - env:
      - name: VLLM_LOGGING_LEVEL
        value: DEBUG
      - name: VLLM_CPU_KVCACHE_SPACE
        value: "1"
      image: public.ecr.aws/q9t5s3a7/vllm-cpu-release-repo:v0.19.0
      name: main
      resources:
        limits:
          cpu: "1"
          memory: 10Gi
        requests:
          cpu: 100m
          memory: 8Gi
```

## Anything else you would like to add?

The rendered vLLM workload is intentionally deployed without TLS.

vLLM supports HTTPS when TLS-related options are configured, for example:

```text
--ssl-keyfile
--ssl-certfile
--ssl-ca-certs
--enable-ssl-refresh
--ssl-cert-reqs
--ssl-ciphers
```

However, this LLMInferenceService workload does not configure those TLS options, so it serves plain HTTP. This matches the observed runtime behavior: HTTP requests to the workload Service succeed, while HTTPS requests fail with `wrong version number`.

Therefore, the generated workload Service should not advertise `appProtocol: https` unless the workload is actually configured to serve TLS.

The generated `HTTPRoute` is accepted and references are resolved:

```yaml
status:
  parents:
  - conditions:
    - reason: Accepted
      status: "True"
      type: Accepted
    - reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: traefik.io/gateway-controller
```

The Gateway is also programmed:

```yaml
status:
  conditions:
  - reason: Accepted
    status: "True"
    type: Accepted
  - reason: Programmed
    status: "True"
    type: Programmed
```

So the routing objects themselves appear to be valid. The failure appears to be caused by the generated workload Service protocol metadata.

The relevant controller code appears to be in the LLMInferenceService workload Service generation logic, where the Service port is created with:

```go
Name:        "https",
AppProtocol: ptr.To("https"),
Port:        8000,
TargetPort: intstr.IntOrString{
    Type:   intstr.Int,
    IntVal: 8000,
},
```

A possible fix is to make the workload Service port name and `appProtocol` reflect the actual backend protocol instead of hardcoding `https`.

For a non-TLS vLLM backend, the generated Service should use:

```yaml
name: http
appProtocol: http
```

If workload TLS is explicitly enabled, then the Service can use:

```yaml
name: https
appProtocol: https
```

If this is expected behavior or a configuration issue on my side, please feel free to close this issue. Any guidance on the recommended setup for a non-TLS vLLM backend with Gateway API would be appreciated.

## Environment

- Istio Version: N/A
- Knative Version: N/A
- KServe Version: `v0.18.0`
- Kubeflow version: N/A
- Cloud Environment: bare metal Kubernetes / Traefik Gateway API
- Minikube/Kind version: N/A
- Kubernetes version: TODO, output of `kubectl version`
- OS: TODO, output of `/etc/os-release`
- Resource type: `LLMInferenceService`
- Gateway API implementation: Traefik Gateway API
- Backend runtime: vLLM
- Backend port: `8000`

walnutkim:

This PR updates the LLMInferenceService workload Service port metadata to respect `config.EnableTLS`.

Previously, the workload Service was always generated with:

```go
Name:        "https",
AppProtocol: ptr.To("https"),
```

even when the vLLM workload was deployed without TLS and served plain HTTP on port 8000.

With Gateway API implementations such as Traefik, appProtocol: https can make the gateway treat the workload Service as an HTTPS upstream. This causes backend TLS failures for non-TLS vLLM workloads.

This PR sets the workload Service port name and appProtocol to:

- http when config.EnableTLS is false
- https when config.EnableTLS is true

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:

Fixes #5531

Manually verified with a non-TLS vLLM backend:

- HTTP request to the workload Service succeeds:
  - `curl http://facebook-opt-125m-single-kserve-workload-svc:8000/v1/models`

  - returns `200 OK`

- HTTPS request to the same non-TLS workload Service fails:
  - `curl -vk https://facebook-opt-125m-single-kserve-workload-svc:8000/v1/models`

  - fails with `wrong version number`

- After changing the generated Service metadata from `https` to `http`, the Traefik Gateway API path succeeds:
  - `curl http://172.16.200.184/kserve-test/facebook-opt-125m-single/v1/models`

  - returns `200 OK`

Local test:

```bash

go test ./pkg/controller/v1alpha2/llmisvc/...
ok      github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc 250.805s
?       github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc/fixture [no test files]
```

- [ ] Test A
- [ ] Test B

- Logs

This change preserves the existing `https` Service metadata when LLMInferenceService TLS is enabled, while avoiding advertising HTTPS for non-TLS vLLM workloads.

**Checklist**:

**Release note**:
```release-note
LLMInferenceService workload Service port metadata now respects the TLS configuration, using HTTP metadata for non-TLS workloads and HTTPS metadata when TLS is enabled.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/workload.go
+++ b/pkg/controller/v1alpha2/llmisvc/workload.go
@@ -96,7 +96,7 @@ func (r *LLMISVCReconciler) reconcileWorkload(ctx context.Context, llmSvc *v1alp
 	}
 
 	// Create Service to expose workload pods
-	if err := r.reconcileWorkloadService(ctx, llmSvc); err != nil {
+	if err := r.reconcileWorkloadService(ctx, llmSvc, config); err != nil {
 		llmSvc.MarkMainWorkloadNotReady("ReconcileWorkloadServiceError", err.Error())
 		return fmt.Errorf("failed to reconcile workload service: %w", err)
 	}
@@ -117,7 +117,11 @@ func (r *LLMISVCReconciler) reconcileWorkload(ctx context.Context, llmSvc *v1alp
 	return nil
 }
 
-func (r *LLMISVCReconciler) reconcileWorkloadService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+func (r *LLMISVCReconciler) reconcileWorkloadService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, config *Config) error {
+	workloadServiceProtocol := "http"
+	if config != nil && config.EnableTLS {
+		workloadServiceProtocol = "https"
+	}
 	expected := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      kmeta.ChildName(llmSvc.GetName(), "-kserve-workload-svc"),
@@ -142,9 +146,9 @@ func (r *LLMISVCReconciler) reconcileWorkloadService(ctx context.Context, llmSvc
 			// "main receiver" port.
 			Ports: []corev1.ServicePort{
 				{
-					Name:        "https",
+					Name:        workloadServiceProtocol,
 					Protocol:    corev1.ProtocolTCP,
-					AppProtocol: ptr.To("https"),
+					AppProtocol: ptr.To(workloadServiceProtocol),
 					Port:        8000,
 					TargetPort: intstr.IntOrString{
 						Type:   intstr.Int,


madhugoutham:
This is covered by #5525 which adds a full per-CR TLS toggle port name, appProtocol, cert generation, and status URL all switch based on spec.tls.enabled.

pierDipi:
@madhugoutham I think this PR can go on its own the other PR needs more discussions as it's introducing a new API field.

pierDipi:
@walnutkim can you fix the DCO issue https://github.com/kserve/kserve/pull/5533/checks?check_run_id=76025039908?

walnutkim:
> @walnutkim can you fix the DCO issue https://github.com/kserve/kserve/pull/5533/checks?check_run_id=76025039908?

Done! I've signed off the commit and force-pushed. Please check again.

bartoszmajsak (approved):
/lgtm

Thanks for this improvement!

PR #5540, opened 2026-05-14
feat(llmisvc): bubble up HPA/KEDA scaling status to service conditions

### LINKED ISSUE #5530: feat(llmisvc): bubble up ScaledObject/HPA status to LLMInferenceService conditions
## Problem

Currently, `LLMInferenceService` does not surface any status from its managed autoscaling resources (KEDA `ScaledObject` or `HorizontalPodAutoscaler`). The only scaling-related feedback on the CR today is:

- **Reconcile errors** (`ScalingCRDNotFound`, `ReconcileScalingError`) set when the scaling resource fails to create/update
- **Workload readiness** from Deployment/LWS status, which is indirect and only reflects whether pods are available — not whether autoscaling itself is healthy

This means failures like a broken metric pipeline (HPA `AbleToScale=False`, `ScalingActive=False`), KEDA trigger errors, or a ScaledObject stuck in an error state are invisible to users looking at the `LLMInferenceService` status. The only way to discover these issues today is to manually inspect the child HPA or ScaledObject, which is a poor user experience.

## Proposal

Introduce a new condition (e.g. `ScalingReady`) on `LLMInferenceService` that reflects the health of the managed autoscaling resources. This condition should:

1. **Read status from the active actuator:**
   - When using **HPA**: check `AbleToScale` and `ScalingActive` conditions from `HorizontalPodAutoscaler.status.conditions`
   - When using **KEDA ScaledObject**: check `ScaledObject.status.conditions` (e.g. `Ready`, `Active`, `Fallback`)

2. **Roll up into the existing condition hierarchy:**
   - Add `ScalingReady` as a sub-condition under `WorkloadsReady` (alongside `MainWorkloadReady`, `WorkerWorkloadReady`, etc.), or as a peer — to be discussed
   - If prefill scaling is separately configured, consider `PrefillScalingReady` as well

3. **Handle edge cases:**
   - When no scaling is configured, the condition should be absent (not False)
   - When scaling is configured but the actuator CRD is missing, the existing `ScalingCRDNotFound` reason on `MainWorkloadNotReady` is sufficient
   - When KEDA manages its own HPA underneath a ScaledObject, surface ScaledObject status (not the KEDA-generated HPA)

## Current status conditions for reference

| Condition | Role |
|---|---|
| `Ready` | Top-level, derived from `WorkloadsReady` + `RouterReady` |
| `WorkloadsReady` | Rollup of `MainWorkloadReady`, `WorkerWorkloadReady`, `PrefillWorkloadReady`, `PrefillWorkerWorkloadReady` |
| `RouterReady` | Rollup of `GatewaysReady`, `HTTPRoutesReady`, `InferencePoolReady`, `SchedulerWorkloadReady` |
| `PresetsCombined` | Preset merge status (not in the Ready rollup) |

## Relevant code

- Scaling reconciliation: `pkg/controller/v1alpha2/llmisvc/scaling.go`
- Lifecycle/conditions: `pkg/apis/serving/v1alpha2/llm_inference_service_lifecycle.go`
- Controller watches (already owns HPA and ScaledObject): `pkg/controller/v1alpha2/llmisvc/controller.go`
- Workload status propagation pattern to follow: `pkg/controller/v1alpha2/llmisvc/workload_single_node.go` (`propagateDeploymentStatus`)

## Acceptance criteria

- [ ] New `ScalingReady` condition type on `LLMInferenceService`
- [ ] HPA `AbleToScale` / `ScalingActive` status propagated when HPA is the actuator
- [ ] KEDA `ScaledObject` status conditions propagated when KEDA is the actuator
- [ ] Condition is absent when no scaling is configured
- [ ] Condition participates in the `WorkloadsReady` (or top-level) rollup so `Ready` reflects scaling health
- [ ] Unit tests covering all actuator states (healthy, degraded, missing)

vivekk16:
## What

Surfaces the health of autoscaling resources (HPA or KEDA ScaledObject) directly in the LLMInferenceService status via two new conditions: `ScalingReady` and `PrefillScalingReady`.

Closes #5530

## Why

Users currently have to inspect child HPA/ScaledObject resources separately to diagnose scaling issues. This change propagates autoscaler health into the parent CR so `kubectl get`/`describe` shows the full picture.

## Changes

* **New conditions**: `ScalingReady`, `PrefillScalingReady` with marker methods and rollup into `WorkloadsReady` via `DetermineWorkloadReadiness()`
* **Status propagation**: `propagateHPAStatus()` maps AbleToScale/ScalingActive; `propagateScaledObjectStatus()` maps Ready
* **Refactor**: Deduplicated main/prefill scaling into shared `reconcileWorkloadScaling()` via `workloadScalingParams`

## Follow-ups

* Informational states like ScalingLimited (HPA), Fallback (KEDA), and Paused (KEDA) are intentionally not surfaced in this PR. These are better suited as non-terminal status conditions rather than events and will be addressed separately.

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/scaling.go
+++ b/pkg/controller/v1alpha2/llmisvc/scaling.go
@@ -26,6 +26,7 @@ import (
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
@@ -267,6 +268,10 @@ func (r *LLMISVCReconciler) propagateScalingStatus(ctx context.Context, llmSvc *
 func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *autoscalingv2.HorizontalPodAutoscaler, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
 	curr := &autoscalingv2.HorizontalPodAutoscaler{}
 	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
+		if apierrors.IsNotFound(err) {
+			notReady("HPAProgressing", "HPA not yet visible in cache")
+			return nil
+		}
 		return fmt.Errorf("failed to get current HPA %s/%s: %w", expected.GetNamespace(), expected.GetName(), err)
 	}
 
@@ -313,6 +318,10 @@ func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *au
 func (r *LLMISVCReconciler) propagateScaledObjectStatus(ctx context.Context, expected *kedav1alpha1.ScaledObject, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
 	curr := &kedav1alpha1.ScaledObject{}
 	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
+		if apierrors.IsNotFound(err) {
+			notReady("ScaledObjectProgressing", "ScaledObject not yet visible in cache")
+			return nil
+		}
 		return fmt.Errorf("failed to get current ScaledObject %s/%s: %w", expected.GetNamespace(), expected.GetName(), err)
 	}
 
--- a/pkg/controller/v1alpha2/llmisvc/scaling.go
+++ b/pkg/controller/v1alpha2/llmisvc/scaling.go
@@ -264,7 +264,7 @@ func (r *LLMISVCReconciler) propagateScalingStatus(ctx context.Context, llmSvc *
 
 // propagateHPAStatus reads the live HPA status and maps its conditions to a ScalingReady
 // condition on the LLMInferenceService. AbleToScale=False or ScalingActive=False means the
-// metrics pipeline is broken; ScalingLimited is informational (scaling works but is capped).
+// metrics pipeline is broken and sets ScalingReady=False.
 func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *autoscalingv2.HorizontalPodAutoscaler, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
 	curr := &autoscalingv2.HorizontalPodAutoscaler{}
 	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
@@ -279,25 +279,11 @@ func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *au
 
 	for _, cond := range curr.Status.Conditions {
 		switch cond.Type {
-		case autoscalingv2.AbleToScale:
+		case autoscalingv2.AbleToScale, autoscalingv2.ScalingActive:
 			foundAny = true
 			if cond.Status == corev1.ConditionFalse {
 				notReady(cond.Reason, cond.Message)
 				return nil
-			}
-		case autoscalingv2.ScalingActive:
-			foundAny = true
-			if cond.Status == corev1.ConditionFalse {
-				notReady(cond.Reason, cond.Message)
-				return nil
-			}
-		case autoscalingv2.ScalingLimited:
-			// ScalingLimited=True is informational: the HPA is functioning but capped at
-			// min or max replicas. This is not a failure — scaling is still healthy — so we
-			// emit an event rather than marking ScalingReady=False. The event is visible via
-			// `kubectl describe llminferenceservice` in the Events section.
-			if cond.Status == corev1.ConditionTrue {
-				r.Eventf(expected, corev1.EventTypeNormal, "ScalingLimited", "HPA is at scaling limit: %s", cond.Message)
 			}
 		}
 	}
@@ -313,8 +299,7 @@ func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *au
 
 // propagateScaledObjectStatus reads the live KEDA ScaledObject status and maps its conditions
 // to a ScalingReady condition on the LLMInferenceService. Ready=False means a trigger/config
-// issue; Paused=True means scaling is paused; Fallback=True is a soft warning (scaling degraded
-// but still operating).
+// issue and sets ScalingReady=False.
 func (r *LLMISVCReconciler) propagateScaledObjectStatus(ctx context.Context, expected *kedav1alpha1.ScaledObject, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
 	curr := &kedav1alpha1.ScaledObject{}
 	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
@@ -331,12 +316,6 @@ func (r *LLMISVCReconciler) propagateScaledObjectStatus(ctx context.Context, exp
 		return nil
 	}
 
-	pausedCond := conditions.GetPausedCondition()
-	if pausedCond.Status == metav1.ConditionTrue {
-		notReady(pausedCond.Reason, pausedCond.Message)
-		return nil
-	}
-
 	readyCond := conditions.GetReadyCondition()
 	if readyCond.Status == metav1.ConditionFalse {
 		notReady(readyCond.Reason, readyCond.Message)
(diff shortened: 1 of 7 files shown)


test/e2e/llmisvc/test_llm_autoscaling.py
pierDipi, line 1: why not augmenting the existing test and we need another test + marker for this?

vivekk16: The separate test+marker was added initially because `ScalingReady` is a new condition and I wanted to test it in isolation first, but you are right, augmenting the existing tests gives better coverage with less duplication. I've included this in existing test without separate marker.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -202,6 +227,128 @@ func (r *LLMISVCReconciler) reconcileActuator(ctx context.Context, llmSvc *v1alp
...
+}
+
+// propagateHPAStatus reads the live HPA status and maps its conditions to a ScalingReady
+// condition on the LLMInferenceService. AbleToScale=False or ScalingActive=False means the
+// metrics pipeline is broken; ScalingLimited is informational (scaling works but is capped).
+func (r *LLMISVCReconciler) propagateHPAStatus(ctx context.Context, expected *autoscalingv2.HorizontalPodAutoscaler, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
+	curr := &autoscalingv2.HorizontalPodAutoscaler{}
+	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
+		return fmt.Errorf("failed to get current HPA %s/%s: %w", expected.GetNamespace(), expected.GetName(), err)
+	}
```
KillianGolds, line 276: Here

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -202,6 +227,128 @@ func (r *LLMISVCReconciler) reconcileActuator(ctx context.Context, llmSvc *v1alp
...
+
+// propagateScaledObjectStatus reads the live KEDA ScaledObject status and maps its conditions
+// to a ScalingReady condition on the LLMInferenceService. Ready=False means a trigger/config
+// issue; Paused=True means scaling is paused; Fallback=True is a soft warning (scaling degraded
+// but still operating).
+func (r *LLMISVCReconciler) propagateScaledObjectStatus(ctx context.Context, expected *kedav1alpha1.ScaledObject, ready func(), notReady func(reason, messageFormat string, messageA ...interface{})) error {
+	curr := &kedav1alpha1.ScaledObject{}
+	if err := r.Get(ctx, client.ObjectKeyFromObject(expected), curr); err != nil {
+		return fmt.Errorf("failed to get current ScaledObject %s/%s: %w", expected.GetNamespace(), expected.GetName(), err)
+	}
```
KillianGolds, line 311: and here.

KillianGolds (commented):
One thing I wanted to ask about: in `propagateHPAStatus` and `propagateScaledObjectStatus` ([scaling.go#L269-L271](pkg/controller/v1alpha2/llmisvc/scaling.go#L269-L271), [scaling.go#L315-L317](pkg/controller/v1alpha2/llmisvc/scaling.go#L315-L317)), the `Get` error is wrapped uniformly. I might be missing something about the timing here, but I was wondering whether a fresh-create `NotFound` could briefly bubble up from the cache-backed reader and cause `reconcileWorkloadScaling` to return an error before the propagation callbacks run. If that does happen in practice, would it be worth special-casing `apierrors.IsNotFound` as a `Progressing` signal? The integration tests look like they exercise the steady-state happy path, so I'm not sure whether they'd cover that narrow window. I could be reading way to much into the ordering so feel free to push back!

vivekk16:
@KillianGolds The controller self-heals on the next requeue so it's not a functional issue, but it does cause a transient error flash in the status. I've updated both `propagateHPAStatus` and `propagateScaledObjectStatus` to treat `NotFound` as a Progressing signal, which is semantically the same as "exists but no conditions yet".

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -202,6 +228,136 @@ func (r *LLMISVCReconciler) reconcileActuator(ctx context.Context, llmSvc *v1alp
...
+				return nil
+			}
+		case autoscalingv2.ScalingLimited:
+			// ScalingLimited=True is informational: the HPA is functioning but capped at
+			// min or max replicas. This is not a failure — scaling is still healthy — so we
+			// emit an event rather than marking ScalingReady=False. The event is visible via
+			// `kubectl describe llminferenceservice` in the Events section.
+			if cond.Status == corev1.ConditionTrue {
+				r.Eventf(expected, corev1.EventTypeNormal, "ScalingLimited", "HPA is at scaling limit: %s", cond.Message)
+			}
```
pierDipi: Because these are continuous states rather than point-in-time occurrences, using events could cause an event flood. This type of telemetry is better suited for a status condition. Keep in mind that not all conditions need to be 'terminal' or affect overall readiness; we can use non-terminal conditions purely for observability. If we want to avoid making changes in this PR, I suggest we remove the event-emitting logic for now and follow up later with a non-terminal condition (as there might be side effects unrelated to this PR).

Also, regarding the current code: for an event to actually be visible on the LLMInferenceService, we must pass the llmisvc resource to Eventf instead of expected. The event recorder requires an object with a hydrated Kind and UID to successfully link the event, which expected is not.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -202,6 +228,136 @@ func (r *LLMISVCReconciler) reconcileActuator(ctx context.Context, llmSvc *v1alp
...
+			if cond.Status == corev1.ConditionFalse {
+				notReady(cond.Reason, cond.Message)
+				return nil
+			}
+		case autoscalingv2.ScalingActive:
+			foundAny = true
+			if cond.Status == corev1.ConditionFalse {
+				notReady(cond.Reason, cond.Message)
+				return nil
+			}
```
pierDipi:
```suggestion
		case autoscalingv2.AbleToScale, autoscalingv2.ScalingActive:
			foundAny = true
			if cond.Status == corev1.ConditionFalse {
				notReady(cond.Reason, cond.Message)
				return nil
			}
```

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -202,6 +228,136 @@ func (r *LLMISVCReconciler) reconcileActuator(ctx context.Context, llmSvc *v1alp
...
+	if conditions == nil || !conditions.AreInitialized() {
+		notReady("ScaledObjectProgressing", "ScaledObject conditions not yet available")
+		return nil
+	}
+
+	pausedCond := conditions.GetPausedCondition()
+	if pausedCond.Status == metav1.ConditionTrue {
+		notReady(pausedCond.Reason, pausedCond.Message)
+		return nil
+	}
```
pierDipi: I'm not sure if paused should affect the overall readiness of llmisvc? maybe ready scaling with a reason of "paused" or something like that?

are we exposing pause in llmisvc?

vivekk16: No, we are not exposing pause in llmisvc. I've removed the Paused handling for now. When/if we add a scaling pause feature to the LLMISVC API, we can design the condition semantics properly at that point.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Combined into a single case autoscalingv2.AbleToScale, autoscalingv2.ScalingActive.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Thanks for flagging this. I removed the event-emitting logic entirely. For now, ScalingLimited is simply ignored (HPA is healthy, ScalingReady=True is set). I'll follow up with a non-terminal condition for observability in a separate PR.

KillianGolds (commented):
LGTM, this looks solid.

Separate from the observability states above, one non-blocking thought that came from this [discussion](https://github.com/kserve/kserve/pull/5540#discussion_r3281102417) on `ScalingReady` itself: since it rolls into `WorkloadsReady`, a freshly created HPA in its metric-warmup window (`ScalingActive=False` / `FailedGetResourceMetric` for the first bit) will briefly show the service as `NotReady` until metrics flow. I think that's reasonable as-is right? im just wondering whether that transient warmup is something we'd want terminal, or worth treating as `Progressing` down the line. cc: @pierDipi 

TL;DR: do we want top-level Ready to mean "serving," or "serving and autoscaling correctly"? Today it's the latter right?

andresllh:
@vivekk16 there seems to be some insufficient CPU issue in many PRs, not just this one. Working on a fix for this.

Issue #5543, opened 2026-05-16
feat(llmisvc): make creation of GIE CRDs optional


neelabhsinha:
/kind feature

**Describe the solution you'd like**

Make creation of the Gateway API Inference Extension (GIE) CRDs in the `kserve-llmisvc-resources` chart **opt-out** via a new boolean value.

Today `charts/kserve-llmisvc-resources/files/llmisvc/resources.yaml` bundles five CRDs that the LLMInferenceService controller depends on:

- `inferencemodelrewrites.inference.networking.x-k8s.io`
- `inferenceobjectives.inference.networking.x-k8s.io`
- `inferencepoolimports.inference.networking.x-k8s.io`
- `inferencepools.inference.networking.k8s.io`
- `inferencepools.inference.networking.x-k8s.io`

`charts/kserve-llmisvc-resources/templates/llmisvc/resources.yaml` renders the file unconditionally, so the chart cannot be installed alongside any other component that already owns those CRDs (a separate `gateway-api-inference-extension` release, an operator bundle that ships GIE, Istio in some configurations, etc.). Helm fails the install with the standard ownership error:

```
Error: Unable to continue with install: CustomResourceDefinition
"inferenceobjectives.inference.networking.x-k8s.io" in namespace "" exists
and cannot be imported into the current release: invalid ownership metadata;
annotation validation error: key "meta.helm.sh/release-name" must equal
"kserve-llmisvc-resources": current value is "<other-release>"
```

There is a way to adopt existing GIE CRDs in Kserve as per https://github.com/kserve/kserve/blob/143e9b835e4c90418ee84dc49cac3afaac5a8ce1/hack/setup/infra/manage.kserve-helm.sh#L332

But vice-versa is not possible.

Adding a single boolean flag mirrors the established pattern for transitive CRD dependencies in the rest of the Helm ecosystem (cert-manager's `installCRDs`) and removes the things currently blocking those integrations.

henrichter-sap:
I'd like to challenge the current default here. A few things feel off:

  1. The docs already list Gateway API CRDs and Gateway API Inference Extension CRDs as prerequisites [1]. So when a user follows the prerequisites correctly, the default `kserve-llmisvc-resources` install fails with a Helm ownership conflict. The default is set up to fail precisely for users who did the right thing.
[...]

PR #5544, opened 2026-05-16
feat(charts): make llmisvc GIE CRD creation optional

neelabhsinha:

Adds a new Helm value `kserve.llmisvc.createGIECRDs` (default: `true`) to `kserve-llmisvc-resources` so operators can skip creating bundled Gateway API Inference Extension (GIE) CRDs when those CRDs are already managed elsewhere.

When the value is set to `false`, the chart renders all `llmisvc` resources except `CustomResourceDefinition` objects, avoiding Helm ownership conflicts in shared-cluster setups. This keeps default behavior unchanged while adding a safe opt-out path for pre-provisioned CRDs.

Also updates chart schema/docs for the new value and includes generated OpenAPI formatting-only updates.

- https://github.com/kserve/kserve/issues/5543

- Verified Helm values and schema include `kserve.llmisvc.createGIECRDs` with default `true`
- Verified chart template logic filters out CRDs only when `createGIECRDs=false`
- Verified default behavior still renders full resource set (including CRDs)

**Notes for the reviewer**:

- Default install behavior remains backward compatible (`createGIECRDs=true`)
- `createGIECRDs=false` is intended for environments where GIE CRDs are installed/owned by another release or platform component
- LLMISVC controller still requires these CRDs to exist at runtime; this flag only controls chart ownership/creation

**Checklist**:

**Release note**:

```release-note
Add `kserve.llmisvc.createGIECRDs` Helm value to optionally skip creating bundled GIE CRDs in the kserve-llmisvc-resources chart.

### DIFF
--- a/charts/kserve-llmisvc-resources/templates/llmisvc/resources.yaml
+++ b/charts/kserve-llmisvc-resources/templates/llmisvc/resources.yaml
@@ -1,5 +1,19 @@
-{{- include "kserve-common.renderMultiResourceWithPatches" (dict
+{{- $rendered := include "kserve-common.renderMultiResourceWithPatches" (dict
   "baseFile" "files/llmisvc/resources.yaml"
   "patchGlob" "files/llmisvc/*-patch.yaml"
   "certName" "llmisvc-serving-cert"
-  "context" .) -}}
+  "context" .) }}
+{{- if .Values.kserve.llmisvc.createGIECRDs -}}
+{{ $rendered }}
+{{- else -}}
+{{- range $doc := splitList "\n---\n" $rendered -}}
+{{- $trimmed := trim $doc -}}
+{{- if $trimmed -}}
+{{- $parsed := fromYaml $trimmed -}}
+{{- if and $parsed (ne $parsed.kind "CustomResourceDefinition") }}
+---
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: charts/kserve-llmisvc-resources/templates/llmisvc/resources.yaml [hunk 1]
```diff
@@ -1,5 +1,19 @@
...
-  "context" .) -}}
+  "context" .) }}
+{{- if .Values.kserve.llmisvc.createGIECRDs -}}
+{{ $rendered }}
+{{- else -}}
+{{- range $doc := splitList "\n---\n" $rendered -}}
+{{- $trimmed := trim $doc -}}
+{{- if $trimmed -}}
+{{- $parsed := fromYaml $trimmed -}}
+{{- if and $parsed (ne $parsed.kind "CustomResourceDefinition") }}
```
pierDipi: should we add a condition on the GIE CRD group/name?

neelabhsinha: Yes, thanks @pierDipi. Added a filter now. Is that fine?

gavrissh (approved):

PR #5546, opened 2026-05-17
feat(llmisvc): make creation of GIE CRDs optional

### LINKED ISSUE #5543: feat(llmisvc): make creation of GIE CRDs optional

skbhudolia12:

Adds a `kserve.llmisvc.installGIECRDs` boolean value (defaults to `true`) to the `kserve-llmisvc-resources` Helm chart, making the five Gateway API Inference Extension (GIE) CRDs opt-out.

Today `templates/llmisvc/resources.yaml` renders the GIE CRDs unconditionally. When another component (standalone `gateway-api-inference-extension` release, an operator bundle, etc.) already owns those CRDs, `helm install` fails with an ownership annotation error.

Setting `kserve.llmisvc.installGIECRDs: false` skips the five CRDs while still rendering all controller resources.

Fixes #5543

Yes. New Helm value `kserve.llmisvc.installGIECRDs` (default `true`).

## How Has This Been Tested?

Verified with `helm template` across three scenarios:

- Default (no flag): 5 GIE CRDs rendered, all controller resources present
- installGIECRDs=true: 5 GIE CRDs rendered, all controller resources present
- installGIECRDs=false: 0 GIE CRDs rendered, all controller resources still present

### DIFF
--- a/charts/kserve-llmisvc-resources/README.md
+++ b/charts/kserve-llmisvc-resources/README.md
@@ -58,6 +58,7 @@ $ helm install kserve-llmisvc-resources oci://ghcr.io/kserve/charts/kserve-llmis
 | kserve.inferenceservice.resources.limits.memory | string | `"2Gi"` |  |
 | kserve.inferenceservice.resources.requests.cpu | string | `"1"` |  |
 | kserve.inferenceservice.resources.requests.memory | string | `"2Gi"` |  |
+| kserve.llmisvc.installGIECRDs | bool | `true` | Whether to install Gateway API Inference Extension (GIE) CRDs. Set to false if already managed by another release. |
 | kserve.llmisvc.controller.affinity | object | `{}` |  |
 | kserve.llmisvc.controller.annotations | object | `{}` |  |
 | kserve.llmisvc.controller.containerSecurityContext.allowPrivilegeEscalation | bool | `false` |  |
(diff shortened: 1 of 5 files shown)


neelabhsinha:
Thanks @skbhudolia12, but we are already managing this change via https://github.com/kserve/kserve/pull/5544.

skbhudolia12:
Thanks for letting me know @neelabhsinha! I wasn't aware of #5544 — happy to close this one in favor of your implementation. Let me know if there's anything from my approach that would be useful to incorporate.

I have a few other PRs open as well (#5547, #5548, #5550) — would appreciate any feedback on those when you get a chance!

PR #5547, opened 2026-05-17
fix(llmisvc): gate SSL_CERT_DIR /var/run/kserve/tls prefix on EnableTLS

### LINKED ISSUE #5537: fix(llmisvc): SSL_CERT_DIR in scheduler config should be conditional on TLS
**What happened:**

`SSL_CERT_DIR` in `config/llmisvcconfig/config-llm-scheduler.yaml` (line 80) is set unconditionally:

```yaml
env:
  - name: SSL_CERT_DIR
    value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
```

The surrounding args on lines 75-78 are all gated by `{{ if .GlobalConfig.EnableTLS }}`, but this env var is not. When TLS is disabled, it references `/var/run/kserve/tls` which won't exist.

**What you expected to happen:**

`SSL_CERT_DIR` should be wrapped in the same `{{ if .GlobalConfig.EnableTLS }}` conditional to stay consistent with `--cert-path`, `--secure-serving`, and `--enable-cert-reload`.

**Impact:**

Low. Go ignores missing cert directories so nothing breaks, but it's inconsistent with the rest of the TLS conditionals and could cause confusing debug output.

skbhudolia12:

`SSL_CERT_DIR` was set unconditionally in the scheduler and routing-sidecar
container specs of three LLMInferenceServiceConfig templates, always
referencing `/var/run/kserve/tls` even when TLS is disabled via
`GlobalConfig.EnableTLS`. This is inconsistent with the surrounding TLS
args (`--secure-serving`, `--cert-path`, `--decoder-use-tls`, etc.) which
are all properly gated behind `{{ if .GlobalConfig.EnableTLS }}`.

Make the `/var/run/kserve/tls` prefix conditional on `EnableTLS` while
keeping the service-account CA and system cert directories **always**
available — those are needed regardless of user-facing TLS so the
container can still verify in-cluster connections (e.g. to the K8s API
server):

```yaml
- name: SSL_CERT_DIR
  value: '{{ if .GlobalConfig.EnableTLS }}/var/run/kserve/tls:{{- end }}/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs'
```

This keeps the change minimal, valid YAML (template stays inside the
quoted string like the existing `--cert-path` pattern), and addresses
the original concern: no path to a non-existent `/var/run/kserve/tls`
when TLS is off.

## Files changed

- `config/llmisvcconfig/config-llm-scheduler.yaml` — scheduler main container
- `config/llmisvcconfig/config-llm-decode-template.yaml` — routing sidecar
- `config/llmisvcconfig/config-llm-decode-worker-data-parallel.yaml` — routing sidecar
- `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml` — regenerated from the three sources via `bash hack/setup/scripts/generate_chart_manifests.sh` so the Helm-bundled copies stay in sync
- `pkg/controller/v1alpha2/llmisvc/config_presets_test.go` — updated the `SSL_CERT_DIR` expected value in `TestPresetFiles` to match the new conditional output (the test's `kserveSystemConfig` leaves `EnableTLS` at its zero value `false`)

Fixes #5537

No behavioral change when TLS is enabled — the rendered value is
unchanged. When TLS is disabled, `SSL_CERT_DIR` no longer references the
non-existent `/var/run/kserve/tls` directory, only the service-account
CA and system certs.

- All 8 `TestPresetFiles` subtests pass locally.
- `go vet`, `go build`, and `golangci-lint run` on the package are clean.
- Verified the regenerated chart file is byte-identical to running
  `bash hack/setup/scripts/generate_chart_manifests.sh` on top of the
  source changes.

### DIFF
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -243,7 +243,7 @@ spec:
           fieldRef:
             fieldPath: metadata.namespace
       - name: SSL_CERT_DIR
-        value: /var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs
+        value: '{{ if .GlobalConfig.EnableTLS }}/var/run/kserve/tls:{{- end }}/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs'
       image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.7.1
       imagePullPolicy: IfNotPresent
       livenessProbe:
@@ -579,7 +579,7 @@ spec:
           fieldRef:
             fieldPath: metadata.namespace
       - name: SSL_CERT_DIR
-        value: /var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs
+        value: '{{ if .GlobalConfig.EnableTLS }}/var/run/kserve/tls:{{- end }}/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs'
       image: ghcr.io/llm-d/llm-d-routing-sidecar:v0.7.1
       imagePullPolicy: IfNotPresent
       livenessProbe:
@@ -1888,7 +1888,7 @@ spec:
             }}'
           env:
           - name: SSL_CERT_DIR
-            value: /var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs
+            value: '{{ if .GlobalConfig.EnableTLS }}/var/run/kserve/tls:{{- end }}/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs'
           image: ghcr.io/llm-d/llm-d-inference-scheduler:v0.7.1
           imagePullPolicy: IfNotPresent
           livenessProbe:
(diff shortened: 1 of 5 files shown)


skbhudolia12:
Rebased onto master in 87bcadd50 to clear the merge conflict — branch was sitting on `412193ca` while master moved to `950ba64df`. Single conflict on `config/llmisvcconfig/config-llm-decode-template.yaml` from #5485 (preStop hook PR): master bumped `terminationGracePeriodSeconds: 30 → 60` on the line directly below the `SSL_CERT_DIR` env var this PR touches. Resolved by taking both — my conditional `SSL_CERT_DIR` value + master's `terminationGracePeriodSeconds: 60`. The other three source YAMLs (`config-llm-decode-worker-data-parallel.yaml`, `config-llm-scheduler.yaml`) and `config_presets_test.go` auto-merged cleanly.

Re-ran `bash hack/setup/scripts/generate_chart_manifests.sh` so `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml` is back in sync with the source kustomize inputs; the diff there is exactly the new conditional template embedded in two places (decode containers in `decode-template.yaml` and `decode-worker-data-parallel.yaml`).

`go build ./...`, `go vet`, `TestPresetFiles` (all 8 subtests), and `golangci-lint run ./pkg/controller/v1alpha2/llmisvc/...` (0 issues) all green on the rebased tip. Net diff against master unchanged: +9 / −7 across the same 5 files.

Friendly ping to `@pierDipi` and `@vivekk16` — this one's been waiting for a first look since 2026-05-18; happy to address any feedback quickly.

madhugoutham:
LGTM, clean fix. Good call keeping the service-account and system cert dirs always available regardless of TLS setting.

PR #5548, opened 2026-05-17
fix(llmisvc): remove duplicate yaml.Unmarshal in loadConfig test helper

skbhudolia12:

`loadConfig()` in `pkg/controller/v1alpha2/llmisvc/config_presets_test.go`
called `yaml.Unmarshal` twice on the same data into the same struct —
a copy-paste bug. The second call is redundant and is removed here.

> The PR was originally scoped to also gate `SSL_CERT_DIR` in
> `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml`,
> but that file is generated from `config/llmisvcconfig/*.yaml` via
> `kustomize build` (see `hack/setup/scripts/generate_chart_manifests.sh`),
> so editing it directly is incorrect — thanks @pierDipi for catching this.
> The SSL_CERT_DIR fix is being addressed at the source in #5547; once
> that lands, `make generate-chart-manifests` will regenerate the chart
> file with the conditional automatically.

No — test-only cleanup with no behavior change.

- Pure deletion of 4 lines; the first `yaml.Unmarshal` call still populates
  `config`, so all subsequent assertions in `loadConfig` are unaffected.
- Package compiles cleanly (`go vet`, `go build`, `golangci-lint run`).

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_presets_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_presets_test.go
@@ -613,10 +613,6 @@ func loadConfig(t *testing.T, data []byte, filePath string) *v1alpha2.LLMInferen
 		t.Errorf("Failed to unmarshal YAML from %s: %v", filePath, err)
 		return nil
 	}
-	if err := yaml.Unmarshal(data, config); err != nil {
-		t.Errorf("Failed to unmarshal YAML from %s: %v", filePath, err)
-		return nil
-	}
 
 	expectedGroupVersion := v1alpha2.LLMInferenceServiceConfigGVK.GroupVersion().String()
 	if config.APIVersion != expectedGroupVersion {


charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
pierDipi, line 1: This is a generated file, we cannot change it as it's done here

skbhudolia12: Thanks for catching this @pierDipi! You're right — `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml` is generated by `kustomize build config/llmisvcconfig` in `hack/setup/scripts/generate_chart_manifests.sh`, so editing it directly is wrong.

I've force-pushed to drop that change — this PR is now scoped only to the unrelated duplicate-`yaml.Unmarshal` removal in `config_presets_test.go` (the title/body have been updated to match). The SSL_CERT_DIR fix is being handled at the source in #5547; once that lands, running `make generate-chart-manifests` will regenerate the chart file with the conditional. PTAL when you have a chance.

bartoszmajsak (approved):
/lgtm

skbhudolia12:
Thanks for the lgtm and rerun @bartoszmajsak — sorry for the slow turnaround, was buried in other things for a few days.

Quick note for whoever picks this up: the only red check on the rerun (`test-llmisvc (kustomize)`) is again the 60-min e2e step timeout, not a real failure on this change — different last-running test this time (`prestop-hook-test` instead of `stop-feature-test`), same `##[error]The action 'Run LLMInferenceService E2E tests' has timed out after 60 minutes.` This branch is a one-line dedup of a duplicate `yaml.Unmarshal` in `pkg/controller/v1alpha2/llmisvc/config_presets_test.go`'s `loadConfig` test helper — strictly test-only, can't possibly affect e2e cluster behavior. Same flake hit #5550 too — looks like the e2e suite is brushing up against the action's wall-clock budget right now.

Happy to rebase / kick another rerun if helpful.

spolti (approved):

skbhudolia12:
Thanks @spolti for the review and approval — appreciate it. And thanks @bartoszmajsak for the earlier lgtm and the rerun. Happy to rebase if anything goes stale before merge.

PR #5554, opened 2026-05-18
feat: support storage initializer resource annotations

### LINKED ISSUE #5326: Huggingface models are unusable
/kind bug

This is related to #5299 . Due to that issue, it is impossible to modify the initContainer. Consequently, because of https://github.com/huggingface/huggingface_hub/issues/3300 attempting to download any model larger than 8Gb will cause an OOM error. This effectively makes HF models unusable in KServe 0.17 because init containers cannot be changed and the default initContainer is incompatible with the behavior of the huggingface_hub library.

The model downloads properly and runs.

```yaml
kind: LLMInferenceService
metadata:
  name: qwen3.5-35b-a3b-single
spec:
  model:
    uri: hf://Qwen/Qwen3.5-35B-A3B
    name: Qwen/Qwen3.5-35B-A3B
  template:
    containers:
      - name: main
        resources:
          limits:
            cpu: '4'
            memory: 32Gi
            nvidia.com/gpu: "1"
          requests:
            cpu: '2'
            memory: 16Gi
            nvidia.com/gpu: "1"
        livenessProbe:
          httpGet:
            path: /health
            port: 8000
            scheme: HTTPS
          initialDelaySeconds: 120
          periodSeconds: 30
          timeoutSeconds: 30
          failureThreshold: 5
```

The PR opened to fix the linked issue would enable a workaround, but the default initContainer limits should still be adjusted.

**Environment:**

- KServe Version: 0.17.0
- Cloud Environment: upcloud
- Kubernetes version: (use `kubectl version`): v1.34.2

Zakharden:

Adds per-service annotations for overriding storage initializer CPU and memory requests/limits:

- `storage.kserve.io/initializer-cpu-request`
- `storage.kserve.io/initializer-cpu-limit`
- `storage.kserve.io/initializer-memory-request`
- `storage.kserve.io/initializer-memory-limit`

This lets large model deployments, especially HuggingFace and LLM workloads, increase storage initializer resources without changing the global `inferenceservice-config` defaults for every workload.

Fixes #5326

- Added unit coverage for successful storage initializer resource overrides.
- Added unit coverage for invalid resource quantity validation.
- Ran `go test -c ./pkg/webhook/admission/pod`.
- `go test ./pkg/webhook/admission/pod` could not run locally because the environment is missing `/usr/local/kubebuilder/bin/etcd`.

This keeps the existing global defaults unchanged and only applies overrides when users explicitly set annotations.

**Checklist**:

**Release note**:
```release-note
Add storage initializer resource override annotations for per-service CPU and memory requests/limits.
```


docs/samples/llmisvc/single-node-gpu/README.md
```diff
@@ -14,6 +14,20 @@ testing, or production deployments of smaller models (e.g., 7B parameter models)
...
+Large HuggingFace models may need more CPU or memory while the storage initializer downloads and prepares model files.
+Use these annotations on an `LLMInferenceService` to override the storage initializer resources for a single service:
+
+```yaml
+metadata:
+  annotations:
+    storage.kserve.io/initializer-cpu-request: "500m"
+    storage.kserve.io/initializer-cpu-limit: "2"
+    storage.kserve.io/initializer-memory-request: "2Gi"
+    storage.kserve.io/initializer-memory-limit: "4Gi"
```
hexfusion, line 28: question: what this PR is proposing makes sense but it feels more like a spec declaration than it does an annotation.  there is already corev1.ResourceRequirements so it would align nicely. i am new to this code base so if an arch conversation has already happened disregard. but the cost of maintaining annotations overtime can be high.

api changes are more work but do we need something more like
```
  spec:
    model:
      uri: hf://Qwen/Qwen3-35B-A3B
    storageInitializer:
      resources:
        requests:
          cpu: 500m
          memory: 2Gi
        limits:
          cpu: "2"
          memory: 4Gi
```

PR #5555, opened 2026-05-18
fix(llmisvc): surface SchedulerPortMismatch on missing scheduler ports

### LINKED ISSUE #5441: llmisvc: scheduler port mismatch creates broken Service without surfacing status condition
/kind bug

When a custom scheduler `PodTemplateSpec` omits one or more required container ports (`grpc`, `grpc-health`, `metrics`, `zmq`), the controller detects the mismatch but only logs a warning and continues. 

The resulting `Service` is created with an incomplete port set, rendering the Endpoint Picker (EPP) unreachable to the Gateway. The operator has no observable signal in the `LLMInferenceService` status, only a log line.

**Affected code**

`pkg/controller/v1alpha2/llmisvc/scheduler.go` — `expectedSchedulerService`:

```go
desiredPorts := sets.New("grpc", "grpc-health", "metrics", "zmq")
actualPorts := make(map[string]*corev1.ContainerPort)
// ... populate actualPorts from scheduler container ports ...

if len(desiredPorts) != len(actualPorts) {
    // TODO should this be raised as failing condition?
    logger.Info("some ports are not matching", "desired", desiredPorts, "actual", maps.Keys(actualPorts))
}
// Service created regardless
```

**Proposed fix outline**

When `desiredPorts` and `actualPorts` differ, the controller could:

In `expectedSchedulerService`, return an error to pause reconciliation when ports are missing:

```go
if diff := desiredPorts.Difference(sets.KeySet(actualPorts)); len(diff) > 0 {
    return nil, fmt.Errorf("scheduler container missing required ports: %v", diff.UnsortedList())
}
```

In `reconcileSchedulerService`, call `llmSvc.MarkSchedulerWorkloadNotReady()` with reason `SchedulerPortMismatch` and a message listing missing ports:

```go
expected, err := r.expectedSchedulerService(ctx, llmSvc)
if err != nil {
    llmSvc.MarkSchedulerWorkloadNotReady("SchedulerPortMismatch", err.Error())
    return err
}
```

skbhudolia12:

When a custom scheduler `PodTemplateSpec` omits one or more of the required
container ports (`grpc`, `grpc-health`, `metrics`, `zmq`), the controller
used to detect the mismatch but only log a warning and continue, producing
an incomplete `Service` that left the Endpoint Picker unreachable to the
Gateway. The operator had no observable signal in the `LLMInferenceService`
status — only a log line.

This PR addresses the `TODO` at `scheduler.go:260` and turns the mismatch
into a proper status condition:

- `expectedSchedulerService` now returns `(*Service, error)`. When any
  required port is missing it returns an error listing the missing port
  names instead of building an incomplete `Service`.
- `reconcileSchedulerService` surfaces that error as a
  `SchedulerWorkloadReady=False` condition with reason
  `SchedulerPortMismatch` and a message naming the missing ports, then
  returns the error so reconciliation pauses until the user fixes the
  template.
- The deletion / force-stop / external-pool path skips the validation —
  a user tearing down a service with an invalid template can still do so
  without hitting the new error.

The `TODO`'s second clause ("check if grpc port matches what's defined in
the inferencepool") is left as a follow-up TODO so this PR stays scoped to
the status-condition fix the issue asks for.

Fixes #5441

Yes (status surface only — no API change). An `LLMInferenceService` whose
scheduler template is missing required ports now reports
`SchedulerWorkloadReady=False` with reason `SchedulerPortMismatch` and a
message naming the missing ports, e.g.:

```
status:
  conditions:
    - type: SchedulerWorkloadReady
      status: "False"
      reason: SchedulerPortMismatch
      message: 'scheduler container template is missing required ports: [grpc-health zmq]'
```

Previously this state produced only a controller-log warning, which is now
also gone (the error message subsumes it).

Added `TestExpectedSchedulerService_PortValidation` (table-driven) and
`TestReconcileSchedulerService_PortMismatchSetsStatusCondition` to
`scheduler_test.go`. Cases covered:

- All four required ports present (happy path).
- One required port missing (error mentions that port).
- Multiple required ports missing (error lists them all).
- All required ports missing.
- Extra non-required ports alongside the required set (ignored).
- Required ports split across multiple containers (no error).
- No scheduler template (deletion path) — no error, `Service` shell only.
- Reconciler-level: `SchedulerWorkloadReady` is set to `False` with reason
  `SchedulerPortMismatch` and the missing port names appear in the message.

Local checks:

- `go vet ./pkg/controller/v1alpha2/llmisvc/...` — clean.
- `go build ./pkg/controller/v1alpha2/llmisvc/...` — clean.
- `golangci-lint run ./pkg/controller/v1alpha2/llmisvc/...` — 0 issues.
- New tests + `TestPresetFiles` + `TestPreserveSchedulerConfig` all pass.

cc @KaveeshKhattar (issue author) @pierDipi @vivekk16


pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -213,10 +213,24 @@ func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context,
...
 	if utils.GetForceStopRuntime(llmSvc) || llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Scheduler == nil || llmSvc.Spec.Router.Scheduler.Template == nil || llmSvc.Spec.Router.Scheduler.Pool.HasRef() {
-		return Delete(ctx, r, llmSvc, expected)
+		toDelete, _ := r.expectedSchedulerService(ctx, llmSvc)
+		return Delete(ctx, r, llmSvc, toDelete)
+	}
+
+	expected, err := r.expectedSchedulerService(ctx, llmSvc)
+	if err != nil {
+		llmSvc.MarkSchedulerWorkloadNotReady(SchedulerPortMismatchReason, "%s", err.Error())
+		return err
```
bartoszmajsak: Wouldn't returning an error cause infinite exponential backoff retries?

I don't think missing ports will fix themselves and since the controller watches spec changes, returning nil after setting the condition is sufficient. Same pattern as `ErrPreconditionNotMet` in `reconcileRouter` for example.

Also it's overwritten further up, so it will get lost in translation: https://github.com/kserve/kserve/blob/777a66cfc9926326fcec31d64f5d0ebd8b718a7d/pkg/controller/v1alpha2/llmisvc/router.go#L90

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -213,10 +213,24 @@ func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context,
...
+
 func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
-	expected := r.expectedSchedulerService(ctx, llmSvc)
+	// Deletion path: we don't validate scheduler ports because the user is
+	// either stopping the runtime or referencing an external InferencePool.
+	// expectedSchedulerService still returns a Service shell with the correct
+	// name/namespace which is all Delete needs.
 	if utils.GetForceStopRuntime(llmSvc) || llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Scheduler == nil || llmSvc.Spec.Router.Scheduler.Template == nil || llmSvc.Spec.Router.Scheduler.Pool.HasRef() {
-		return Delete(ctx, r, llmSvc, expected)
+		toDelete, _ := r.expectedSchedulerService(ctx, llmSvc)
```
bartoszmajsak: I think we can call it once and handle `err != nil` afterwards.

But I think there's also an edge case here that we should take into consideration: if we have missing ports we are returning `nil`  and passing it to `toDelete` - wouldn't it cause panic? https://github.com/kserve/kserve/blob/777a66cfc9926326fcec31d64f5d0ebd8b718a7d/pkg/controller/v1alpha2/llmisvc/lifecycle_crud.go#L63

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -256,9 +269,13 @@ func (r *LLMISVCReconciler) expectedSchedulerService(ctx context.Context, llmSvc
...
 
-		if len(desiredPorts) != len(actualPorts) {
-			// TODO should this be raised as failing condition? + check if grpc port matches what's defined in the inferencepool
-			logger.Info("some ports are not matching", "desired", desiredPorts, "actual", maps.Keys(actualPorts))
+		if missing := desiredPorts.Difference(sets.KeySet(actualPorts)); missing.Len() > 0 {
+			// Returning an error here lets reconcileSchedulerService surface a
+			// SchedulerPortMismatch status condition instead of silently building
+			// an incomplete Service that leaves the Endpoint Picker unreachable
+			// to the Gateway.
+			// TODO also cross-check the grpc port against the InferencePool spec.
```
bartoszmajsak: Is this planned as follow-up?

bartoszmajsak (commented):
Thanks for hunting this! 

Couple of things to consider:

- I wonder if we shouldn't moving port validation earlier, as currently this might result in partial state. `reconcileScheduler` runs `reconcileSchedulerDeployment` before `reconcileSchedulerService`. With missing ports, the deployment gets created but the service doesn't. Are we good with such a state of things?
- shouldn't we update [`sample.go`](https://github.com/kserve/kserve/blob/777a66cfc9926326fcec31d64f5d0ebd8b718a7d/pkg/controller/v1alpha2/llmisvc/sample.go#L193-L208) with this new validation?
- "grpc", "grpc-health", "metrics", "zmq" are string literals in multiple places. Since the PR adds validation against them, extracting to constants might help keeping it consistent.

skbhudolia12:
Thanks for the careful read @bartoszmajsak — every one of those was a real issue. All four addressed in 39a7edf3d:

**1. Partial state: validate before the Deployment, not after.**
Port validation now lives at the top of `reconcileScheduler`, right after `reconcileSchedulerServiceAccount` and before `reconcileSchedulerDeployment`. With missing ports we short-circuit (set the condition, `return nil`) so no Deployment is ever created — the previous \"Deployment exists but no Service ever could\" state is gone.

**2. Return `err` causing infinite backoff + getting overwritten upstream.**
You're right on both counts. The path now mirrors the `ErrPreconditionNotMet` pattern in `reconcileRouter`: mark `SchedulerWorkloadReady=False` with reason `SchedulerPortMismatch` and **return `nil`** — no requeue storms for a non-transient condition, and `router.go:90`'s `MarkSchedulerWorkloadNotReady(\"SchedulerReconcileError\", ...)` no longer overwrites the more specific reason. The watch on `LLMInferenceService` re-triggers reconcile as soon as the user edits the template.

**3. Call once, and the `nil` → `Delete()` panic.**
`expectedSchedulerService` now returns just `*corev1.Service` (no error). It's called once at the top of `reconcileSchedulerService`, and the branch is now \"delete or reconcile\" based on a single `schedulerShouldBeDeleted(llmSvc)` helper. The panic vector you flagged (`existing := expected.DeepCopyObject().(T)` on a typed-nil at `lifecycle_crud.go:63`) is removed by construction — `expectedSchedulerService` always returns a non-nil shell now.

**4. String literals → constants.**
Extracted `schedulerPortGRPC` / `schedulerPortGRPCHealth` / `schedulerPortMetrics` / `schedulerPortZMQ` plus a `requiredSchedulerPorts` set, used by both the validation and the populate-from-template logic. While I was at it I also brought `pkg/controller/v1alpha2/llmisvc/sample.go` (which was missing `metrics`) and two scheduler-config integration test templates (which were missing `zmq`) into compliance — those would otherwise have been silently broken by the new validation.

**5. \"Is this planned as follow-up?\" on the InferencePool grpc-port cross-check.**
Yes — kept intentionally out of this PR to keep it scoped. Moved the `TODO` next to the early-validation block so it's adjacent to the natural place to add it, and left it as a follow-up.

New `reconcileScheduler` shape, for orientation:

```go
func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
    log.FromContext(ctx).Info(\"Reconciling Scheduler\")

    if err := r.reconcileSchedulerServiceAccount(ctx, llmSvc); err != nil {
        return err
    }

    // TODO also cross-check the grpc port against the InferencePool spec.
    if !schedulerShouldBeDeleted(llmSvc) {
        if missing := missingSchedulerPorts(llmSvc.Spec.Router.Scheduler.Template); len(missing) > 0 {
            llmSvc.MarkSchedulerWorkloadNotReady(
                SchedulerPortMismatchReason,
                \"scheduler container template is missing required ports: %v\", missing,
            )
            return nil
        }
    }
    // ... deployment, service, inference pool ...
}
```

Local checks on the rewritten code: `go vet`, `go build`, `golangci-lint run` (0 issues), and the unit tests in the package all pass. Unit tests reshaped around the new shape:

* `TestMissingSchedulerPorts` — pure-function table test (all present, missing one/some/all, extras ignored, ports split across containers, nil template).
* `TestExpectedSchedulerService_PopulatesPresentPorts` — confirms the function is non-validating and just surfaces present ports.
* `TestSchedulerShouldBeDeleted` — covers each branch of the delete predicate including `Pool.HasRef()`.

PTAL when you have a moment — happy to iterate further.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -213,21 +288,26 @@ func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context,
...
 
+// SchedulerPortMismatchReason is the SchedulerWorkloadReady condition reason
+// set by reconcileScheduler when the scheduler PodTemplateSpec is missing
+// one or more required container ports.
+const SchedulerPortMismatchReason = "SchedulerPortMismatch"
+
 func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
 	expected := r.expectedSchedulerService(ctx, llmSvc)
-	if utils.GetForceStopRuntime(llmSvc) || llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Scheduler == nil || llmSvc.Spec.Router.Scheduler.Template == nil || llmSvc.Spec.Router.Scheduler.Pool.HasRef() {
+	if schedulerShouldBeDeleted(llmSvc) {
```
bartoszmajsak: Maybe a simpler approach would be to always construct a minimal service shell for the deletion path as Delete only needs name/namespace?

skbhudolia12:
Good call @bartoszmajsak — done in be553853c. Extracted a small `schedulerServiceShell(llmSvc)` that returns just the `ObjectMeta` (name + namespace), and the delete branch in `reconcileSchedulerService` now uses that instead of running the full `expectedSchedulerService` build:

```go
func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
    if schedulerShouldBeDeleted(llmSvc) {
        return Delete(ctx, r, llmSvc, schedulerServiceShell(llmSvc))
    }
    return Reconcile(ctx, r, llmSvc, &corev1.Service{}, r.expectedSchedulerService(ctx, llmSvc), semanticServiceIsEqual)
}

func schedulerServiceShell(llmSvc *v1alpha2.LLMInferenceService) *corev1.Service {
    return &corev1.Service{
            Name:      llmSvc.Spec.Router.EPPServiceName(llmSvc),
            Namespace: llmSvc.GetNamespace(),
        },
    }
}
```

Verified against `lifecycle_crud.go:59-100` — `Delete()` uses `client.ObjectKeyFromObject(expected)` to look up the live object, and the ownership / `DeletionTimestamp` checks all run against the *fetched* copy, so the input shell really only needs name + namespace. No more wasted PodSpec `DeepCopy`, port collection or sorting on the delete path.

Also tightened the godoc on `expectedSchedulerService` to make this split explicit (it's now \"used by the reconcile path only; see `schedulerServiceShell` for delete\").

Added `TestSchedulerServiceShell` that asserts the shell carries name + namespace and nothing else (no `Spec`, no labels, no owner refs) and that it still works when `llmSvc.Spec.Router == nil` — `EPPServiceName` falls back to the llmSvc-derived default so a stray Service left over from a previous reconcile can still be located and torn down.

`go vet`, `go build`, `golangci-lint run` and the package unit tests all clean. PTAL.

bartoszmajsak:
Why not just returning svc instead of `nil` like it was in the original proposal?

skbhudolia12:
Fair point @bartoszmajsak — you're right, I over-engineered the previous round. Reverted in 88f07606f: `schedulerServiceShell` is gone, `reconcileSchedulerService` is back to the single-call shape:

```go
func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
    expected := r.expectedSchedulerService(ctx, llmSvc)
    if schedulerShouldBeDeleted(llmSvc) {
        return Delete(ctx, r, llmSvc, expected)
    }
    return Reconcile(ctx, r, llmSvc, &corev1.Service{}, expected, semanticServiceIsEqual)
}
```

This matches every other reconciler in the file (`reconcileSchedulerDeployment`, `reconcileV1InferencePool`, `reconcileV1Alpha2InferencePool`) — call `expected...` once, share the result across the delete and reconcile branches. `expectedSchedulerService` is already non-nil-always after the earlier refactor (the panic vector from the original `(nil, err)` return is gone), so passing it to `Delete` is safe — `Delete` only uses `client.ObjectKeyFromObject` from it and ignores the rest. The supposed \"wasted work\" on the delete path is negligible (DeepCopy + port collection only run when `Template != nil`, and the cost is dominated by the actual Delete API call anyway).

`TestSchedulerServiceShell` removed along with the helper. Remaining unit coverage (`TestMissingSchedulerPorts`, `TestExpectedSchedulerService_PopulatesPresentPorts`, `TestSchedulerShouldBeDeleted`) still covers the actual behaviour. `go vet`, `go build`, `golangci-lint run` (0 issues) and the non-envtest tests all clean.

Net diff vs head before this commit: −57 / +9. Apologies for the round-trip — should have stuck with the round-2 shape and trusted the pattern that was already in the file.

bartoszmajsak:
Do we really need `schedulerShouldBeDeleted`?

skbhudolia12:
Good call @bartoszmajsak — addressed in 4d26d8b9e. The helper was an outlier: 6 other delete-or-reconcile pairs in `scheduler.go` (`reconcileSchedulerAuthDelegatorBinding`, `reconcileSchedulerRole`, `reconcileSchedulerRoleBinding`, `reconcileSchedulerServiceAccount`, `reconcileSchedulerDeployment`, `reconcileSchedulerInferencePool`) inline the same `forceStop || Router==nil || Scheduler==nil || Template==nil || Pool.HasRef()` disjunction, so extracting it for only 2 of 8 callsites was inconsistent.

New shape mirrors the `reconcileSchedulerInferencePool → reconcileV1InferencePool(ctx, llmSvc, shouldDelete bool)` pattern already in the file — compute `shouldDelete` once at the entry-point, thread it down:

```go
func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
    ...
    if err := r.reconcileSchedulerServiceAccount(ctx, llmSvc); err != nil {
        return err
    }

    shouldDelete := utils.GetForceStopRuntime(llmSvc) ||
        llmSvc.Spec.Router == nil ||
        llmSvc.Spec.Router.Scheduler == nil ||
        llmSvc.Spec.Router.Scheduler.Template == nil ||
        llmSvc.Spec.Router.Scheduler.Pool.HasRef()

    // TODO also cross-check the grpc port against the InferencePool spec.
    if !shouldDelete {
        if missing := missingSchedulerPorts(llmSvc.Spec.Router.Scheduler.Template); len(missing) > 0 {
            llmSvc.MarkSchedulerWorkloadNotReady(SchedulerPortMismatchReason, ...)
            return nil
        }
    }

    if err := r.reconcileSchedulerDeployment(ctx, llmSvc); err != nil { return err }
    if err := r.reconcileSchedulerService(ctx, llmSvc, shouldDelete); err != nil { return err }
    if err := r.reconcileSchedulerInferencePool(ctx, llmSvc); err != nil { return err }
    ...
}

func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, shouldDelete bool) error {
    expected := r.expectedSchedulerService(ctx, llmSvc)
    if shouldDelete {
        return Delete(ctx, r, llmSvc, expected)
    }
    return Reconcile(ctx, r, llmSvc, &corev1.Service{}, expected, semanticServiceIsEqual)
}
```

Bonus: the port-validation gate and the delete-service decision now share the same `shouldDelete` local, so we re-evaluate the disjunction once per reconcile instead of twice.

`TestSchedulerShouldBeDeleted` removed along with the helper. Behaviour is exercised end-to-end through `reconcileScheduler`'s existing integration tests; the remaining unit coverage (`TestMissingSchedulerPorts`, `TestExpectedSchedulerService_PopulatesPresentPorts`) covers the parts that genuinely benefit from isolated tests. `go vet`, `go build`, `golangci-lint run` (0 issues) and the non-envtest tests all clean. Net diff: −98 / +15.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -213,21 +281,31 @@ func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context,
...
+// SchedulerPortMismatchReason is the SchedulerWorkloadReady condition reason
+// set by reconcileScheduler when the scheduler PodTemplateSpec is missing
+// one or more required container ports.
+const SchedulerPortMismatchReason = "SchedulerPortMismatch"
+
+// reconcileSchedulerService reconciles or deletes the Endpoint Picker
+// Service. shouldDelete carries the caller's delete-path decision (see
+// reconcileScheduler) so the same disjunction isn't re-evaluated here,
+// mirroring the reconcileSchedulerInferencePool → reconcileV1InferencePool
+// pattern already used in this file.
```
bartoszmajsak: Is this really helpful and adds extra context to the code below? Feels somehow redundant.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -68,10 +88,36 @@ func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1al
...
+			// fixes the template.
+			return nil
+		}
+	}
+
 	if err := r.reconcileSchedulerDeployment(ctx, llmSvc); err != nil {
 		return err
 	}
-	if err := r.reconcileSchedulerService(ctx, llmSvc); err != nil {
+	if err := r.reconcileSchedulerService(ctx, llmSvc, shouldDelete); err != nil {
```
bartoszmajsak: `shouldDelete` is cheap; I don't feel like we need it as an additional parameter here. 

If we really want to avoid duplication, maybe we can have a helper func `shouldDeleteScheduler` for `LLMInferenceService` instead to encapsulate this check?

bartoszmajsak:
@skbhudolia12 PTAL at my last round of review :)

skbhudolia12:
Sorry for the delay @bartoszmajsak — got pulled into a few other things this week. Both addressed in 9de490444.

**`shouldDelete` parameter / godoc.** Agreed, threading it was over-engineering for something this cheap. Reverted the parameter and inlined the disjunction at the callsite, matching the existing inline style used by `reconcileSchedulerAuthDelegatorBinding`, `reconcileSchedulerRole`, `reconcileSchedulerRoleBinding`, `reconcileSchedulerServiceAccount` and `reconcileSchedulerDeployment` in the same file:

```go
func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
    expected := r.expectedSchedulerService(ctx, llmSvc)
    if utils.GetForceStopRuntime(llmSvc) || llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Scheduler == nil || llmSvc.Spec.Router.Scheduler.Template == nil || llmSvc.Spec.Router.Scheduler.Pool.HasRef() {
        return Delete(ctx, r, llmSvc, expected)
    }
    return Reconcile(ctx, r, llmSvc, &corev1.Service{}, expected, semanticServiceIsEqual)
}
```

With the parameter gone, the godoc explaining it had nothing left to add — dropped too. Net diff: −8 / +3.

`reconcileScheduler` keeps the local `shouldDelete` since it's used exactly once there (to guard port validation), so no duplication concern. Skipped the optional `(*LLMInferenceService).ShouldDeleteScheduler` extraction — happy to do it as a separate scoped PR that converts all 8 callsites in one shot if you'd prefer, since adding the helper here would just re-introduce the "extracted for some, inlined for others" inconsistency we walked back from earlier.

`go vet`, `go build`, `golangci-lint run` (0 issues) and the non-envtest tests all clean.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -213,21 +281,26 @@ func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context,
 	return Reconcile(ctx, r, llmSvc, &igwapiv1alpha2.InferencePool{}, expected, semanticInferencePoolV1Alpha2IsEqual)
 }
 
+// SchedulerPortMismatchReason is the SchedulerWorkloadReady condition reason
+// set by reconcileScheduler when the scheduler PodTemplateSpec is missing
+// one or more required container ports.
+const SchedulerPortMismatchReason = "SchedulerPortMismatch"
```
bartoszmajsak: nit: maybe worth placing it next to relevant port constants and narrowing the scope (i.e. we probably don't need to export it?)

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -68,6 +88,32 @@ func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1al
...
+	if !shouldDelete {
+		if missing := missingSchedulerPorts(llmSvc.Spec.Router.Scheduler.Template); len(missing) > 0 {
+			llmSvc.MarkSchedulerWorkloadNotReady(
+				SchedulerPortMismatchReason,
+				"scheduler container template is missing required ports: %v", missing,
+			)
+			// Returning nil avoids exponential backoff retries; the watch
+			// on LLMInferenceService re-triggers reconcile when the user
+			// fixes the template.
+			return nil
```
bartoszmajsak, line 116: That leaves `reconcileSchedulerServiceAccount` with orphaned RBAC - maybe we can move it as first check?

bartoszmajsak (commented):
Two minor things, otherwise LGTM

skbhudolia12:
Both addressed in 9ee66c9c2 @bartoszmajsak — thanks for the second pass.

**Const placement + visibility.** Moved `SchedulerPortMismatchReason` next to `requiredSchedulerPorts` (right under `schedulerPortGRPC` / `schedulerPortGRPCHealth` / `schedulerPortMetrics` / `schedulerPortZMQ`) and unexported it to `schedulerPortMismatchReason`. The only callsite is the `MarkSchedulerWorkloadNotReady` call inside `reconcileScheduler` in the same file, so there's no reason it needs to live in the public API surface — confirmed `grep -r SchedulerPortMismatchReason` returns zero hits outside `scheduler.go`. All four related declarations (port names, required-set, and now the condition reason) read top-to-bottom in one block.

**Orphaned RBAC on missing-ports.** Good catch — hoisted the port-validation block above `reconcileSchedulerServiceAccount` so the order is now:

1. compute `shouldDelete`
2. if `!shouldDelete` and ports missing → `MarkSchedulerWorkloadNotReady(schedulerPortMismatchReason, …)` + `return nil`
3. `reconcileSchedulerServiceAccount` (which transitively handles AuthDelegatorBinding, Role, RoleBinding)
4. Deployment / Service / InferencePool

Previously a fresh create with a malformed template would have left a ServiceAccount, Role, RoleBinding and AuthDelegator ClusterRoleBinding behind with no Deployment to use them. Now the controller returns before touching any cluster state on the bad-template path — the user just sees `SchedulerWorkloadReady=False / reason=SchedulerPortMismatch` with the missing-ports list in the Message and nothing got created. Validation stays gated on `!shouldDelete` so the force-stop / Pool.HasRef() / no-template cleanup paths still flow through the SA reconciler and reach `reconcileSchedulerDeployment`'s `Stopped` reason for force-stop, as before.

Updated the inline comment on the validation block to call out *why* it has to run first (the orphaned-RBAC hazard), so the next reader doesn't accidentally walk it back below the SA reconcile.

`go build ./...`, `go vet ./pkg/controller/v1alpha2/llmisvc/...`, the non-envtest unit suite, and `golangci-lint run ./pkg/controller/v1alpha2/llmisvc/...` (0 issues) all clean. Net diff: +17 / −15 in `scheduler.go`, no test changes needed — the existing `TestMissingSchedulerPorts` / `TestExpectedSchedulerService_PopulatesPresentPorts` coverage doesn't depend on reconcile-step ordering.

bartoszmajsak (approved):

bartoszmajsak:
@skbhudolia12 can you take care of merge conflicts?

bartoszmajsak:
@skbhudolia12 friendly ping about merge conflicts

PR #5557, opened 2026-05-18
docs: add AGENTS.md with project conventions

bartoszmajsak:

Adds an `AGENTS.md` file that gives AI coding agents (Claude Code, Copilot, Cursor, etc.) the project context they need to produce accurate changes on the first pass.

AI agents work better when they understand the project's constraints, testing conventions, and controller-runtime patterns upfront - rather than guessing or re-discovering them each session. This means fewer review cycles and less back-and-forth fixing agent-generated code that doesn't match project conventions.

Covers:
- Hard constraints (autogenerated files, distro build tags, `make precommit`)
- Project structure and build commands
- Testing conventions (Ginkgo/Gomega, envtest, `pkg/testing/` helpers, fixture packages)
- Controller-runtime patterns (reconcile loop, status/conditions with Knative lifecycle helpers, watches/caching)
- PR template requirements

Documentation-only change, no functional impact.

This file is only consumed by AI coding agents - it has no effect on builds, tests, or CI. The content reflects the project's existing conventions (gathered from the Makefile, test files, and controller code) rather than introducing new ones. A `CLAUDE.md` symlink is included for Claude Code compatibility.

**Checklist**:

```release-note
NONE
```


bartoszmajsak:
As a follow-up we can create and link a doc with e2e troubleshooting or any other helpful tips /cc @jlost

AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+
+```bash
+make test          # full Go test suite (downloads envtest assets automatically)
+make test-qpext    # qpext tests
+```
+
+Run tests for a specific package:
+
+```bash
+KUBEBUILDER_ASSETS="$(setup-envtest use -p path)" go test ./pkg/controller/v1beta1/...
```
pierDipi:
```suggestion
make setup-envtest
KUBEBUILDER_ASSETS="$(./bin/setup-envtest use $(go list -m -f '{{ .Version }}' k8s.io/api | awk -F'[v.]' '{printf "1.%d", $3}') -p path)" go test ./pkg/controller/v1beta1/...
```

bartoszmajsak: That's definitely more robust! Thanks!

AGENTS.md
```diff
@@ -0,0 +1,127 @@
+# KServe Project Overview
+
+KServe is a Kubernetes-native platform for serving machine learning models. The Go codebase implements custom controllers using controller-runtime to manage inference services, serving runtimes, and related resources on Kubernetes.
+
+The project lives under `github.com/kserve/kserve`. Core controller logic is in `pkg/controller/`, API types in `pkg/apis/`, and webhook handling in `pkg/webhook/`.
+
+## Building and Testing
```
israel-hdez: Lately, I prefer we implement _incremental disclosure_ to keep used context low. Not mandatory, but I would prefer we split some of these sub-headings in dedicated files and we refer them here.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
+# KServe Project Overview
+
+KServe is a Kubernetes-native platform for serving machine learning models. The Go codebase implements custom controllers using controller-runtime to manage inference services, serving runtimes, and related resources on Kubernetes.
+
+The project lives under `github.com/kserve/kserve`. Core controller logic is in `pkg/controller/`, API types in `pkg/apis/`, and webhook handling in `pkg/webhook/`.
```
israel-hdez: In my experience, this is redundant. The previous paragraph already points that this project uses `controller-runtime` and I've seen that quite enough for AI assistants to correctly use/understand this structure.

I'd recommend to remove, unless really needed.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+`make precommit` runs all of these in the right order.
+
+## Development Conventions
+
+- API types live in `pkg/apis/serving/` with versioned subdirectories (`v1alpha1`, `v1alpha2`, `v1beta1`).
+- Controllers live in `pkg/controller/` with versioned subdirectories.
+- Tests live next to the code they test. Pure logic/unit tests often use fakes, while controller integration suites typically use envtest.
+- CRDs are generated from Go types using controller-gen markers.
+- Run `make precommit` before submitting changes to catch formatting, linting, and generation drift.
```
israel-hdez: This one is mentioned more than once in this file.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+Several artifacts are autogenerated - do not edit them by hand:
+- Helm chart `README.md` files are generated by `helm-docs` from `values.yaml` annotations
+- Helm chart templates under `charts/*/templates/` are generated from kustomize overlays via `generate-chart-manifests`
+- Common Helm helpers (`_utils.tpl`, `_common.tpl`, `_resources.tpl`) are synced from `charts/_common/` to individual charts
+- CRD YAML files are copied from `config/crd/full/` into `charts/kserve-crd/templates/`
+
+`make precommit` runs all of these in the right order.
+
+## Development Conventions
```
israel-hdez: Similarly to my comment above, the mention of using `controller-runtime` has been enough for me for AI assistants to automatically follow all what is mentioned here (with the only exception of the `make precommit` which is anyway mentioned in other parts).

I'd recommend to remove this whole heading.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+- API types live in `pkg/apis/serving/` with versioned subdirectories (`v1alpha1`, `v1alpha2`, `v1beta1`).
+- Controllers live in `pkg/controller/` with versioned subdirectories.
+- Tests live next to the code they test. Pure logic/unit tests often use fakes, while controller integration suites typically use envtest.
+- CRDs are generated from Go types using controller-gen markers.
+- Run `make precommit` before submitting changes to catch formatting, linting, and generation drift.
+
+## Development Workflow
```
israel-hdez: This doesn't align to my personal workflow. I do implementation first, test second.

Do we really want everybody using this workflow?

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Pull Requests
+
+Use the template in `.github/PULL_REQUEST_TEMPLATE.md` when opening PRs. Every PR must include:
+
+- **What this PR does / why we need it** - clear description of the change and motivation
+- **Which issue(s) this PR fixes** - link issues with `Fixes #<number>` format
+- **Feature/Issue validation/testing** - describe how you tested the change, include relevant logs
+- **Checklist** - confirm unit/e2e tests, comments for complex code, and docs updates
+- **Release note** - write a release note or `NONE` if not user-facing
```
israel-hdez: Maybe, we should add these instructions to the template, rather than specifying here?

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+Run `make precommit` before opening a PR.
```
israel-hdez: This is already mentioned in the template.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Controller-Runtime Patterns
+
+### Reconcile Loop
+
+- Reconcile must be **idempotent** - same input run N times produces the same result
+- Use `context.Context` everywhere; avoid `context.Background()` inside reconcile
```
israel-hdez: Maybe this?

```suggestion
- Avoid using `context.Background()`, prefer propagating existing `context.Context` via function arguments.
```

israel-hdez (commented):
Most of my comments are for minimizing context usage. Not blockers, you choose if you want to change the PR.

AGENTS.md
```diff
@@ -0,0 +1,127 @@
...
+make generate                # codegen, OpenAPI/Python client generation, Helm chart READMEs
+make manifests               # CRDs, RBAC, webhook configs
+make generate-chart-manifests  # Helm chart manifests from kustomize overlays
+```
+
+Several artifacts are autogenerated - do not edit them by hand:
+- Helm chart `README.md` files are generated by `helm-docs` from `values.yaml` annotations
+- Helm chart templates under `charts/*/templates/` are generated from kustomize overlays via `generate-chart-manifests`
+- Common Helm helpers (`_utils.tpl`, `_common.tpl`, `_resources.tpl`) are synced from `charts/_common/` to individual charts
+- CRD YAML files are copied from `config/crd/full/` into `charts/kserve-crd/templates/`
```
spolti: and the install scripts.

AGENTS.md
bartoszmajsak: I think it is relatively short right now - only 128 lines (yay!), but I agree some more focused topics (like e2e troubleshooting) deserve their own docs (also for the ... human contributors).

More docs right from the start means more tool calling, maybe we can see how this works with the simple README as start and we can split later?

AGENTS.md
bartoszmajsak: From my experience, having good tests sketched first is a good signal if the agent actually understands the task at hand.

But maybe that should be the human part before the agent takes over?

AGENTS.md
bartoszmajsak: Maybe it's enough to just refer to the template really. I think in my /gh-pr skill it's all what's needed + a bit of steering on what should be the focus of the PR description. Let me rewrite that.

AGENTS.md
bartoszmajsak:
```suggestion
```

AGENTS.md
bartoszmajsak:
```suggestion
```

AGENTS.md
israel-hdez: I'm fine with this as a start, as long as this doesn't grow in several thousands of lines :-)

AGENTS.md
israel-hdez: Yeah, this really looks more like a personal workflow thing and about what works for each individual.

Personally, I've adopted spec-driven development with quite good results: I let the agent to write a markdown file for the implementation that I review. This is where I ensure that the agent has the right idea and I do edits where I see it needed. Then, I let it implement.

Later, I ask the agent to write yet another spec focused on the tests. The resulting `md` file is what I review/filter so that only the tests that really make sense are coded.

spolti:
/lgtm

Let's wait for others to jump in :)

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+For faster iteration on a specific package, run `make setup-envtest` once, then:
+
+```
+KUBEBUILDER_ASSETS="$(./bin/setup-envtest use $(go list -m -f '{{ .Version }}' k8s.io/api | awk -F'[v.]' '{printf "1.%d", $3}') -p path)" \
+  go test ./pkg/controller/v1beta1/... -run TestSpecificName -v
+```
+
+## Testing
+
+- Tests live next to the code they test
```
spolti, line 41: Should we add instructions to run the Python e2e tests locally?

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+**Isolation and cleanup:**
+- Each test creates its own namespace
+- Clean up resources with `defer` immediately after creation
+- Use `retry.RetryOnConflict` when updating resources the controller is also reconciling
+
+**Assertions:**
+- Use `Eventually`/`Consistently`, never `time.Sleep`
+- Simulate external controller behavior (status updates) with helper functions
+
+## Development Workflow
```
israel-hdez, line 65: I'm still not very happy with outlining a workflow. It will simply break my own workflow and some personal skills I have.

bartoszmajsak: Open to suggestions :)

AGENTS.md
bartoszmajsak: Is a timely feedback loop worth making an agent aware of? One thing I was thinking of is to run relevant e2e tests as the final verification. WDYT?

AGENTS.md
pierDipi: maybe let's remove the specific section for now?

bartoszmajsak:
It seems I got couple of LGTMs here, can we get this merged @kserve/kserve-approvers?

Issue #4083, opened 2024-11-26
Harmonizing OCI Image model support


rhuss:
There are three ways how to run models that are stored in OCI images:

* The [Modelcar](https://kserve.github.io/website/latest/modelserving/storage/oci/) approach that leverages a sidecar for direct access to the model data in an OCI image. It leverages the `oci://` schema in the `storageUri`
* A custom storage initializer [oci-storage-initializer](https://github.com/lampajr/oci-storage-initializer/tree/main) that does pulling of the image programmatically and copies the model locally (essentially in the same way as other storage options). This custom initializer uses `oci-artifact://` as schema currently.
* This [draft PR](https://github.com/kserve/kserve/pull/3539) covers Kubernetes-native OCI image volumes, which IMO should be the future for handling models stored in OCI images. Since cri-o 1.31 now supports image volumes (see this [demo instructions](https://gist.github.com/rhuss/7ba76324dcf26497d84a5277a8267d69) for running image volumes on Minikube)

I want to cover all of these with a consistent user experience. Only modelcars are currently part of core KServe, so I propose to extend the modelcar configuration and specification as follows:

* Change the global KServe configuration from 

```
{ 
   ...
   "enableModelcar": true,
   ...
}
```

to 

```
{ 
   ...
   # Feature flag could be removed later. 
   # Also support "enableModelcar" for backward compatibility
   "enableOciModelSupport": "true",
   
   # Which mode to use by default:
   # - "modelcar" Modelcar as known before
   # - "native" Kubernetes native OCI image volumes
   # - "fetch" Download OCI image internally and store it on a volume
   "ociModelMode": "modelcar"
}
```

Then for the `storageUri`, the proposal would be to extend the `oci://` schema with some _mode_ added with `+` (like in `webdav+ssl` or `github+ssh`):

* `oci+modelcar://....` - Use modelcar to fetch the model
* `oci+native://...` - Leverage native Kubernetes support for mounting the OCI image
* `oci+fetch://....` - Download the image programmatically and store it in an emptyDir volume locally

A difficult question is how to deal with `oci://...` which is the current schema for modelcars. Possible implementations could be:

* Use the default as configured globally (see above). Only then does a global default make sense. However, if you depend on the configured default, an InferenceService with a `storageUri` that uses `oci://` behaves differently on different installations, which might or might not be an issue.
[...]

PR #5558, opened 2026-05-19
feat(storage): add oci+native:// ImageVolume mounting (Step 2 of #4083)

kliukovkin:
### What

This is Step 2 of the OCI storage harmonization roadmap from issue #4083. Step 1
(modelcar, already merged) dispatches bare `oci://` URIs to a sidecar container that
shares `/mnt/models`. This PR adds the `oci+native://` scheme, which materializes a
Kubernetes [ImageVolume](https://kubernetes.io/docs/concepts/storage/volumes/#image)
(KEP-4639) directly on the pod spec — no sidecar, no storage initializer download.
The model container image is pulled and mounted by the container runtime at
`/mnt/models` as a read-only volume.

@rhuss, @lizzzcai — this is the answer to the open question from #4083 about how
`oci://` dispatches between modelcar and native. Short version: `oci://` stays modelcar
by default (backcompat unchanged), `oci+native://` is an explicit per-service override,
and `ociModelMode: "native"` in `inferenceservice-config` flips the cluster-wide default.
The dispatch logic lives in `ResolveOciModelMode` in `pkg/types/config.go`. Details below.

The change covers `InferenceService` (v1beta1), `LLMInferenceService` (v1alpha2), and
the webhook injector path (`InjectModelcar` + `CommonStorageInitialization`). `oci+fetch://`
(Step 3) and arbitrary OCI artifact support (Step 4) are not in this PR.

### Commits

| Hash | Description |
|------|-------------|
| `79027fcc9` | feat(config): add `enableOciModelSupport` and `ociModelMode` ConfigMap fields |
| `85b23f569` | feat(storage): parse `oci+native://` scheme and route `oci://` by configured mode |
| `3a867425e` | feat(webhook): materialize ImageVolume for `oci+native://` |
| `0111493fd` | feat(controller): warn on incompatible K8s version when `oci+native://` resolved |
| `9a603f60d` | feat(e2e): smoke test, sample, and docs for `oci+native://` |

Each commit is independently buildable. `go test ./pkg/...` and `make go-lint` are clean
at every hash.

### Design decisions

**`enableOciModelSupport` is opt-in and defaults to `false`.** The new field sits alongside
the existing `enableModelcar` in `StorageInitializerConfig`. Setting it to `true` without
`ociModelMode` defaults to `"modelcar"` — same as `enableModelcar: true` — so it functions
as an alias for clusters migrating away from the older field. `ociModelMode: "native"` is
the explicit switch to ImageVolume mounting. I kept the default `false` because ImageVolume
requires K8s ≥ 1.31 and containerd ≥ 2.0 or CRI-O ≥ 1.31; most existing clusters won't
meet that bar, and an accidental opt-in would silently fail on incompatible nodes. If the
project prefers to default to `"native"` on K8s ≥ 1.33 (autodetect at controller startup),
`ResolveOciModelMode` is a three-line change — happy to revisit.

**Two-path asymmetry in the injector.** `InjectModelcar` runs as a webhook mutator, gated by
`config.EnableOciImageSource`. It reads the `internal.serving.kserve.io/storage-initializer-sourceuri`
annotation on the pod (set by the controller from `spec.predictor.model.storageUri`) and
processes OCI URIs only. When `ResolveOciModelMode` returns empty — meaning neither
`enableOciModelSupport` nor `ociModelMode` is configured — `InjectModelcar` falls back to
`OciModelModeModelcar`. This is intentional: the function predates config-based dispatch and
serves clusters that have `EnableOciImageSource` set but no explicit mode configured.
`CommonStorageInitialization`, by contrast, is called directly by the Raw and Knative
reconcilers (`raw_kube_reconciler.go`, `ksvc_reconciler.go`) from the ISVC spec's
`storageUri` — not through the pod annotation path. When `IsLegacyURI=false` and
`ResolveOciModelMode` returns empty for an OCI URI, it skips the URI rather than falling
back to modelcar. The asymmetry is intentional: `InjectModelcar` has a legacy user base
that expects a modelcar default; the reconciler path is newer and can afford the stricter
no-fallback contract.

**The materializer is idempotent.** `ConfigureOciNativeToContainer` in `pkg/utils/storage.go`
is a no-op if the named volume already exists on the pod spec. Volume name is derived from
mount path via the existing `GetVolumeNameFromPath` helper (e.g., `/mnt/models` →
`mnt-models`), matching the naming convention used by `ConfigureModelcarToContainer`. If a
different volume is already mounted at the target path, the function returns an error rather
than silently clobbering the mount. This makes re-admission on pod spec updates safe and
keeps the collision detection deterministic.

**The K8s version condition is advisory, not a hard gate.** `warnIfImageVolumeUnsupported`
in `pkg/controller/v1beta1/inferenceservice/oci_version_warn.go` (and its mirror in the
LLMInferenceService controller) sets an `OciImageVolumeCompatible` condition with severity
`Info` when the cluster minor version is below 31, or between 31 and 32. The condition does
not influence `ConditionReady`. I chose advisory over hard-reject at the webhook for three
reasons: the controller doesn't own admission for this code path; users targeting alpha
clusters have already opted in knowingly; and a warning that survives an upgrade is more
debuggable than an error that disappears. The condition clears automatically once the cluster
reaches 1.33 (beta, gate on by default). If the project prefers a hard webhook rejection
for K8s < 1.31, that's a straightforward addition to `ConfigureOciNativeToContainer` using
a server version check — it just needs a client in the injector, which it doesn't have today.

**subPath defaults to `"models"` for modelcar compatibility.** The container runtime mounts
the OCI image under `/mnt/models` with a `subPath: "models"` VolumeMount, so existing
modelcar images (which uniformly pack model files under `/models/*` per the [KServe OCI
provider docs](https://kserve.github.io/website/docs/model-serving/storage/providers/oci))
work unchanged under `oci+native://`. Requires K8s ≥ 1.33: subPath on `ImageVolume` was
forbidden in alpha (1.31/1.32). For images with non-`/models/` layouts, a URI-fragment
override (e.g., `oci+native://image:tag#path`) is a contained followup — `ParseOciScheme`
already recognizes the fragment slot.

**`enableModelcar` semantics are unchanged.** The existing `enableModelcar` bool is preserved
as a backcompat alias; its presence in `ResolveOciModelMode` is unchanged. No existing
behavior changes when this PR is applied to a cluster that doesn't set `enableOciModelSupport`
or `ociModelMode`.

### Out of scope

`oci+fetch://` (Step 3 of #4083) — storage initializer pulls image layers via the existing
Python fetcher. The scheme parser in this PR recognizes `oci+fetch` as a mode token and
would route it through a not-yet-implemented handler. Nothing in this PR produces or depends
on that handler.

The Python storage initializer (`kserve/python/kserve/storage/`) has no changes. The
`oci+native://` path routes through the webhook materializer only; the storage initializer
is not invoked and needs no changes for native mounting.

Arbitrary OCI artifact support (Step 4 of #4083) — non-model OCI artifacts (datasets,
adapter weights packed as OCI) would build on this infrastructure. Not in this PR.

`LLMInferenceService` gets the same K8s version warning as `InferenceService` but the LoRA
adapter code path is untouched. LoRA adapters continue to use the modelcar pattern and are
managed by a separate feature branch.

### How to verify locally

Apply the sample (replace the image ref with one you can pull):

```bash
kubectl apply -f docs/samples/storage/oci-image-volume/inference_service.yaml
kubectl get inferenceservice sklearn-oci-native -n kserve-test
kubectl describe pod -n kserve-test \
    -l serving.kserve.io/inferenceservice=sklearn-oci-native
```

The pod spec should show an `image` volume and a read-only mount on `kserve-container`:

```
Volumes:
  mnt-models:
    Type:       Image (an OCI container image)
    Reference:  ghcr.io/my-org/my-sklearn-model:v1
    Pull Policy: IfNotPresent

...
    Mounts:
      /mnt/models from mnt-models (ro)
```

To exercise the K8s version warning, apply to a cluster running K8s 1.30 with
`ociModelMode: "native"` set in `inferenceservice-config`. The ISVC should reach
`Ready=True` but `kubectl get inferenceservice sklearn-oci-native -o jsonpath='{.status.conditions}'`
should include an `OciImageVolumeCompatible=False` entry with reason `ImageVolumeUnsupported`.

To exercise global default mode without the explicit scheme, set `ociModelMode: "native"` in
the ConfigMap and use a bare `oci://` URI. The webhook should materialise the same ImageVolume
structure.

Unit tests cover all materializer logic paths:

```bash
go test ./pkg/types/... ./pkg/utils/... \
    ./pkg/webhook/admission/pod/... \
    ./pkg/controller/v1beta1/inferenceservice/... \
    ./pkg/controller/v1alpha2/llmisvc/...
```

Lint:

```bash
make go-lint
```

### Open questions / flags for review

**`enableOciModelSupport` default.** Currently `false` (opt-in). If the project wants
opt-out for K8s ≥ 1.33 clusters (e.g., check server version at controller startup and
default to `"native"` automatically), I'm open to it. Fine either way — lmk preference.

**`OciImageVolumeCompatible` condition exposure.** The condition is set on both
`InferenceService.status` and `LLMInferenceService.status`. It sits outside the main
`conditionSet` so it doesn't roll up into `ConditionReady`. If the project prefers it stay
internal until ImageVolume graduates (i.e., log-only, no status field), I can remove it —
the version check would become a controller log line only. Current form is more debuggable
for operators. Lmk preference.

**K8s version thresholds.** I used minor < 31 → `ImageVolumeUnsupported`, 31–32 →
`ImageVolumeAlpha` (gate exists but is on by default only from 1.33), ≥ 33 → condition
cleared. These match the KEP-4639 graduation milestones. If there's reason to lower the
"clean" threshold to 1.31 (treating alpha clusters as fully supported), the message strings
in `oci_version_warn.go` are the only change. Lmk.

**Collision error vs warning.** `ConfigureOciNativeToContainer` returns an error when a
different volume is already bound to `/mnt/models`. This surfaces as a webhook admission
error. An alternative is to log a warning and skip injection (let the pod admit with the
existing mount). The strict behavior seemed safer — a collision is almost certainly a
misconfiguration — but if there's a real use case for pre-existing mounts at the model
path, I can soften it.

### Verification status

Go tests: 40 unit tests across all modified packages — all pass.

```
pkg/types:                          6 cases (ResolveOciModelMode, config parsing)
pkg/utils:                          5 cases (ParseOciScheme) + 5 cases (ConfigureOciNativeToContainer)
pkg/webhook/admission/pod:          6 cases (InjectModelcar + CommonStorageInitialization OCI paths)
pkg/controller/v1beta1/inferenceservice: 8 cases (warnIfImageVolumeUnsupported, all version branches)
pkg/controller/v1alpha2/llmisvc:    5 cases (LLMInferenceService version warn)
pkg/apis/serving/v1beta1:           5 cases (GetStorageInitializerConfigs parsing + validation)
```

Lint: `make go-lint` — 0 issues.

`envtest`: not available in local environment. Controller reconciler paths that require it
are not covered by the test suite in this PR. The materializer, scheme parser, and version
warn logic are all unit-testable without a running API server and are covered directly.

E2E: `test/e2e/storage/test_oci_native.py` runs under the `storage` pytest marker. It skips
automatically on K8s < 1.31 and on 1.31–1.32 (alpha gate state is undetectable from outside
the cluster). Needs K8s ≥ 1.33 to run fully. The test image placeholder
(`ghcr.io/kserve/oci-native-test-fixture:v1`) will need to be published or replaced with an
accessible OCI-format image before CI can complete the end-to-end assertion. Happy to
coordinate or use an existing fixture if the project has one.

### References

- Issue #4083: OCI storage harmonization roadmap
- [KEP-4639](https://github.com/kubernetes/enhancements/issues/4639): OCI Volume Source
  (alpha K8s 1.31, beta K8s 1.33)
- `docs/samples/storage/oci-image-volume/` — runnable manifest + README


test/e2e/storage/test_oci_native.py
pierDipi, line 1: can we add a similar e2e test for llmisvc?

kliukovkin: Added in 9c0871b90. Mirrors the InferenceService variant — same skip logic and assertions, just creates LLMInferenceService with an inline pod template so the controller materializes the ImageVolume via workload_storage.go. Thanks for the catch.

kliukovkin:
Rebased onto master to pick up #5470 (multiple OCI sources). Resolved conflicts in storage_initializer_injector.go, storage_test.go, and storage_initializer_injector_test.go — took both #5470's ociIndex tracking and our native/modelcar dispatch, so both features coexist cleanly.

kliukovkin:
Pushed two fixes:

- Fix the actual panic surfaced by the Test job: `oci+native://` now goes through `ParseOciScheme` in the `CommonStorageInitialization` skip filter (was missing the `oci+native://` case). Regression test added.
- Run `make precommit` to regenerate the configmap/overlay artifacts I missed locally.

The four cancelled kustomize e2e jobs (test-predictor / test-graph / test-kourier / test-huggingface-server-vllm) were all 'runner has received a shutdown signal' — no assertion failures in logs. /retest those when convenient.

kliukovkin:
Triage of the five failing checks — all upstream, not introduced on this branch:

**Three kustomize jobs** (test-autoscaling, test-predictor, test-raw istio-gatewayapi) all fail in the **download-hf-model init container** with a HuggingFace Hub network error. The init container was introduced by #5566 (cache opt-125m in SeaweedFS, merged 2026-06-03) and master has been red on the same jobs since — last green master e2e run was 2026-04-07. The helm twins of these tests passed on different runners where the Hub was reachable, confirming the failure is runner-side network flakiness, not in this PR.

**Two helm jobs** (test-huggingface-server-vllm, test-llm) hit InferenceService startup timeout and 45-minute workflow timeout respectively. The new `warnIfImageVolumeUnsupported` controller hook only fires a discovery call when the resolved mode is `native`, and the shipped default is `modelcar`, so it doesn't run on these test paths. The make-precommit regen in 0c67615f1 only touches the two new fields (`enableOciModelSupport: false`, `ociModelMode: "modelcar"`) — no incidental changes to HF/S3/AWS/MLflow keys.

Local gates on this HEAD are green: `go test`, `go-lint`, `kubectl kustomize config/default` and `config/overlays/test`. precommit-check and the Test job both passed in CI on this same SHA.

Could one of you /retest just the failing jobs when convenient? Thanks for the channel ping @pierDipi.

pkg/controller/v1alpha2/llmisvc/oci_version_warn.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if minor < 31 {
+		mgr.MarkFalse(ociImageVolumeCompatible, "ImageVolumeUnsupported",
+			"Cluster K8s %s.%s does not support ImageVolume (introduced in 1.31). Falling back to modelcar may be required.",
+			v.Major, v.Minor)
+		return
+	}
+
+	if minor < 33 {
```
brettmthompson: @kliukovkin  it looks like the feature gate defaults to false in k8s 1.33 and 1.34 as well, it was set to true by default in 1.35.

You can view the `ImageVolume` feature gate status per release through this link:
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

pkg/controller/v1beta1/inferenceservice/oci_version_warn.go
```diff
@@ -0,0 +1,101 @@
+/*
```
brettmthompson: Given the oci version check + warning logic here is identical to `pkg/controller/v1alpha2/llmisvc/oci_version_warn.go`, would this logic be better as a utility in `pkg/utilities` that can be reused across both controllers?

pkg/controller/v1beta1/inferenceservice/controller_oci_warn_test.go
```diff
@@ -0,0 +1,138 @@
+/*
```
brettmthompson, line 1: Same here - tests can be consolidated for a single utility.

pkg/utils/storage.go
```diff
@@ -582,3 +582,75 @@ func ConfigureModelcarToContainer(modelUri string, podSpec *corev1.PodSpec, targ
...
+
+	// Reject if modelPath is already claimed by a different mount.
+	for _, m := range targetContainer.VolumeMounts {
+		if m.MountPath == modelPath && m.Name != volName {
+			return fmt.Errorf("mountPath %q already used by volume %q", modelPath, m.Name)
+		}
+	}
+
+	// Idempotent: skip if the volume already exists (e.g. duplicate adapter path).
+	for _, v := range podSpec.Volumes {
```
brettmthompson, line 680: Is this required? `AddVolumeMountIfNotPresent` already skips if a matching volume mount name is found.
https://github.com/kserve/kserve/blob/master/pkg/utils/utils.go#L304

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -162,33 +162,51 @@ func (mi *StorageInitializerInjector) InjectModelcar(pod *corev1.Pod) error {
...
 		return nil
 	}
 
+	// Determine effective mode: explicit suffix wins; bare oci:// falls back to config or modelcar default.
+	effectiveMode := parsedMode
+	if effectiveMode == "" {
+		if m := types.ResolveOciModelMode(mi.config); m != "" {
+			effectiveMode = m
+		} else {
+			effectiveMode = types.OciModelModeModelcar
```
brettmthompson: Is this else block necessary? `ResolveOciModelMode` will never return an empty string here because `enableOCIImageSource` must be true in order for this webhook to execute:
https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/mutator.go#L141

test/e2e/storage/__init__.py
brettmthompson, line 1: nit: other test package init files are empty:
https://github.com/kserve/kserve/blob/master/test/e2e/predictor/__init__.py

brettmthompson:
@kliukovkin if you are planning for the `enableOciModelSupport` setting to eventually take over and deprecate the `enableModelcar` setting, have you investigated everywhere in the code that `enableOciModelSupport` will need to be used in addition to `enableModelcar` during the transition (with the intention for long term replacement)? Are you planning on implementing these changes as a follow up?

kliukovkin:
Thanks @cjohannsen-cloudera for the Slack diagnosis and the runner enable, and @brettmthompson for the careful review. Round 3 pushed at `d2fa0c3` (rebased onto current master to keep the branch clean — auto-merge handled the test-fixture renames upstream made while this was in review).

### cjohannsen's two diagnoses

- **`torch_dtype` → `dtype` rename** in `13bec96`. Applied across all three callsites of the same pattern (`encoder_model.py:177`, `generative_model.py:237`, `time_series_model.py:103`) so we don't surface this again on the next CI iteration. `__main__.py` references to `model_config.torch_dtype` are left alone — those read the AutoConfig attribute, separate thing, not deprecated.
- **vLLM CPU v0 fallback** in `ad33907`. Added `VLLM_USE_V1=0` to all four Qwen vLLM CPU test fixtures in `test_huggingface_vllm_cpu.py`. Cory listed three by name; I added it to the fourth (chat-stream) for consistency since the env block is shared.

### brettmthompson's review

Inline replies on each thread; quick summary here:

- **K8s feature gate threshold**: you were right — gate is default-off in 1.33 and 1.34 too, default-on only from 1.35. Verified against the K8s feature-gates doc. Threshold corrected to `<31=Unsupported`, `31..34=NeedsFeatureGate`, `≥35=clear`. Message strings and unit tests updated to match.
- **Refactor**: consolidated both `oci_version_warn.go` files into a single `pkg/utils/imagevolume.go` exposing `CheckImageVolumeCompatibility`. Both controllers are thin call sites now (a few lines each); tests live under the helper. See `85f3451`.
- **Volume idempotency loop**: kept it — `AddVolumeMountIfNotPresent` dedupes per-container `VolumeMounts` by name, but the loop you flagged is on pod-level `Volumes`, which is a separate API resource with no pre-existing dedup helper. Added an explanatory comment in `c3e481b` so future readers don't trip on the same question.
- **Else block in `InjectModelcar`**: you were right again — verified via `ResolveOciModelMode` at `config.go:52`. The disjunction `EnableOciImageSource || EnableOciModelSupport` guarantees the resolver returns non-empty inside `InjectModelcar` (which is only registered when `EnableOciImageSource=true`). Dropped in `b36c758`.
- **`test/e2e/storage/__init__.py`**: emptied to match the predictor package style. See `d2fa0c3`.

### Deprecation plan for `enableModelcar`

Yes, the long-term intent is that `enableOciModelSupport` replaces `enableModelcar`. Audit of the current active surface (excluding frozen install snapshots and chart values templates that just mirror the configmap):

- `pkg/types/config.go` `ResolveOciModelMode` is the only resolution site for both fields today. Current check order is `OciModelMode → enableOciModelSupport → enableModelcar`, which keeps clusters that set only `enableModelcar=true` behaving exactly as before. That's intentional backcompat.
- Three chart READMEs (`charts/kserve-{,llmisvc-,localmodel-}resources/README.md`), `config/configmap/inferenceservice.yaml`, and `charts/_common/common-sections.yaml` mention `enableModelcar`. None need code changes to keep working — just doc updates at deprecation time.
- Tests in `configmap_test.go`, `storage_initializer_injector_test.go`, the LLMISVC controller test suite, and the fixture `required_resources.go` cover the existing field. New cases needed only when we add the deprecation warning.

Concrete followup plan I'm signing up for:

1. **Tracking issue + audit doc** filed today if you want to bookmark it before this PR merges; otherwise immediately after merge. Pin in #4083 thread so the OCI work has it.
2. **PR next minor release**: controller log warning when `enableModelcar` is set without `enableOciModelSupport`. Update chart READMEs and the configmap inline comment to point at the new field. No behavior change.
3. **PR after that** (gated on `oci+native://` having soaked on the K8s 1.36-stable baseline): treat `enableModelcar=true` internally as `enableOciModelSupport=true; ociModelMode=modelcar`. Old field stays in schema, behavior unified.
4. **PR one more minor release later**: remove `enableModelcar` from `StorageInitializerConfig`, update docs and tests, bump the configmap schema version.

Happy to file the tracking issue now or after merge — your call.

### Verification

After rebase, the five local gates are clean: `go test ./pkg/...`, `make go-lint`, `ruff check` on the three Python files touched, `kubectl kustomize config/default` and `config/overlays/test`. Branch shows no conflicts with base. CI on `d2fa0c3` should turn the previously-failing helm twins green (the two issues Cory diagnosed are the only known causes of those failures).

pkg/controller/v1alpha2/llmisvc/oci_version_warn.go
kliukovkin: Updated — threshold is now <31=Unsupported, 31..34=NeedsFeatureGate, ≥35=clear, matching the gate timeline in the K8s docs. See 85f3451.

pkg/controller/v1beta1/inferenceservice/oci_version_warn.go
kliukovkin: Consolidated both helpers into pkg/utils/imagevolume.go with a single CheckImageVolumeCompatibility function. Both controllers are thin call sites now. See 85f3451.

pkg/controller/v1beta1/inferenceservice/controller_oci_warn_test.go
kliukovkin: Tests moved into pkg/utils/imagevolume_test.go alongside the helper. See 85f3451.

pkg/utils/storage.go
kliukovkin: Kept the loop — AddVolumeMountIfNotPresent dedupes per-container VolumeMounts by name, but this loop is on pod-level Volumes, a separate API resource with no pre-existing dedup helper. Added an explanatory comment in c3e481b so future readers see the layer distinction.

pkg/webhook/admission/pod/storage_initializer_injector.go
kliukovkin: You were right — ResolveOciModelMode at config.go:52 uses EnableOciImageSource || EnableOciModelSupport, and InjectModelcar is only registered when EnableOciImageSource=true, so the resolver always returns non-empty here. Dropped in b36c758.

test/e2e/storage/__init__.py
kliukovkin: Done — emptied to match the predictor package style. See d2fa0c3.

test/e2e/predictor/test_huggingface_vllm_cpu.py
```diff
@@ -78,6 +78,10 @@ def test_huggingface_vllm_cpu_openai_chat_completions():
                     name="VLLM_ENABLE_V1_MULTIPROCESSING",
                     value="0",
                 ),
+                client.V1EnvVar(
```
yashaswipiplani: This env var is removed from vLLM since `v0.15.0`. When we upgrade the image, we would need to remove it and find the proper fix. Do you think the alternate suggestion is better?

> Alternatively, fixing just the multiprocessing method (less invasive):
> YAML
> 
> env:
>   - name: VLLM_WORKER_MULTIPROC_METHOD
>     value: "spawn"

cc: @cjohannsen-cloudera

test/e2e/predictor/test_huggingface_vllm_cpu.py
yashaswipiplani, line 77: If we do go ahead with using `VLLM_USE_V1: 0` for now, I think this env is redundant as we would not be using V1 engine at all

brettmthompson:
@kliukovkin should `ResolveOciModelMode` also be used [here](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/mutator.go#L140)? Without it `enableModelCar` is still the sole gate for this functionality for InferenceServices. If `enableModelCar` is false the `InjectModelCar` mutating webhook will never be registered regardless of whether `OciModelMode` or `EnableOciModelSupport` are configured.

kliukovkin:
@brettmthompson — you're right, and good catch. Pushed in a4a90a5: changed `mutator.go:140` from gating on `EnableOciImageSource` alone to `types.ResolveOciModelMode(storageInitializer.config) != ""`, so a cluster opting in only via `enableOciModelSupport` (with `enableModelcar` still false) now correctly gets `InjectModelcar` registered.

Added a test case in `mutator_test.go` covering the `EnableOciModelSupport=true, EnableOciImageSource=false` opt-in pattern to lock the behavior.

The round-3 simplification inside `InjectModelcar` (dropping the `OciModelModeModelcar` fallback in the else branch) stays correct under the new gate — `ResolveOciModelMode` is guaranteed non-empty at the call site precisely because the gate now checks for that. Updated the explanatory comment in `storage_initializer_injector.go` to reflect the new precondition.

Three files touched, +46/-3 lines. Local gates all green.

test/e2e/predictor/test_huggingface_vllm_cpu.py
kliukovkin: Thanks @yashaswipiplani — both points are valid. The `VLLM_USE_V1=0` choice here was per @cjohannsen-cloudera's primary diagnosis (Slack thread in #kserve-contributors), with `VLLM_WORKER_MULTIPROC_METHOD=spawn` listed as the less-invasive alternative.

Cory — quick question before we commit to one: do you prefer (a) keeping `VLLM_USE_V1=0` and dropping the now-redundant `VLLM_ENABLE_V1_MULTIPROCESSING`, or (b) switching to `VLLM_WORKER_MULTIPROC_METHOD=spawn` so we stay on the v1 engine path? Happy to apply either — just want to honor your original call.

test/e2e/predictor/test_huggingface_vllm_cpu.py
kliukovkin: Agreed — under `VLLM_USE_V1=0` this env is effectively a no-op. Will fold the cleanup into whichever path Cory picks for the broader env decision above.

pkg/controller/v1alpha2/llmisvc/oci_version_warn.go
```diff
@@ -0,0 +1,67 @@
+/*
```
brettmthompson: nit: are `pkg/controller/v1alpha2/llmisvc/oci_version_warn.go` and `pkg/controller/v1beta1/inferenceservice/oci_version_warn.go` still needed? Can we not combine everything from these files into `pkg/utils/imagevolume.go` and invoke the methods from `pkg/utils/imagevolume.go` within the controllers?

pkg/webhook/admission/pod/storage_initializer_injector.go
brettmthompson: One thing to keep in mind, I think the validation in the `ValidateOCIMountPaths` method can be less restrictive when `ImageVolumes` are used. When multiple storage uris are specified, this method currently rejects any that share the same parent directory. This is because with model cars an empty dir is mounted to the parent directory for each uri. On the other hand, `ImageVolumes`s can be mounted directly at the path.

pkg/controller/v1alpha2/llmisvc/oci_version_warn.go
kliukovkin: Good call — collapsed both per-controller `oci_version_warn.go` files into private helpers inside the respective `controller.go`. The version-check logic stays in `pkg/utils.CheckImageVolumeCompatibility` as the single source of truth; each controller handles its own condition translation inline (<30 lines).

Didn't move the translation logic into `pkg/utils` itself because that would require utils to import `pkg/apis/serving/{v1beta1,v1alpha2}` types — uncommon dependency direction in this codebase. Happy to revisit if you prefer the full move; just wanted to flag the layering trade-off.

See `10b0c60`. Tests moved into `controller_oci_warn_test.go` next to each `controller.go`.

pkg/webhook/admission/pod/storage_initializer_injector.go
kliukovkin: Good catch — added a `mode` parameter to `ValidateOCIMountPaths` in `7e49ee0`. Native mode validates exact-path uniqueness only (siblings under the same parent directory are fine, only exact-path duplicates rejected); modelcar (or any other / empty mode, defensively) keeps the existing parent-dir collision check.

Updated the call site in `storage_initializer_injector.go` to pass `effectiveMode`. Existing test cases pass `OciModelModeModelcar` explicitly and still behave the same; new sub-tests cover native-mode's relaxed semantics and exact-path rejection.

pkg/controller/v1alpha2/llmisvc/oci_version_warn.go
brettmthompson: This separation makes sense. The differing APIs used to interact with llmisvc and isvc status makes it impossible to re-use a single helper function across both resource types.

pkg/utils/storage.go
```diff
@@ -582,3 +603,79 @@ func ConfigureModelcarToContainer(modelUri string, podSpec *corev1.PodSpec, targ
...
+// mount shadowing.
+func ConfigureOciNativeToContainer(modelUri string, podSpec *corev1.PodSpec, targetContainerName, modelPath string, _ *types.StorageInitializerConfig) error {
+	targetContainer := GetContainerWithName(podSpec, targetContainerName)
+	if targetContainer == nil {
+		return fmt.Errorf("no container found with name %s", targetContainerName)
+	}
+
+	imageRef := strings.TrimPrefix(modelUri, constants.OciURIPrefix)
+
+	volName := GetVolumeNameFromPath(modelPath)
```
brettmthompson, line 661: nit: I was looking into the naming convention for model storage volumes and it seems like we follow the following:
1. For all legacy URIs a single volume is created named `kserve-provision-location`
2. For all non-legacy URIs that are not OCI, a single volume is mounted at the common parent path of all storage URIs.
3. For all non-legacy URIs that are OCI, each URI has a volume created called `kserve-provision-location-n` with n being the # of the storage URI index.

brettmthompson:
@kliukovkin thank you for addressing all my feedback so quickly. The current implementation LGTM.

I am curious to hear your thoughts on the design choice though. I'm wondering if there is a reason for combining the OCI storage conditions behind a single webhook vs exposing multiple webhooks per OCI backend and then [calling](https://github.com/kserve/kserve/blob/master/pkg/webhook/admission/pod/mutator.go#L141) the webhooks conditionally based on the isvc spec?

kliukovkin:
Thanks @brettmthompson — the four-round review genuinely tightened this PR.

On the design question: single webhook with internal dispatch was the conservative call, for two structural reasons.

1. **Consistency with the existing storage entry point.** `InjectStorageInitializer` delegates to a single `CommonStorageInitialization` that handles every storage scheme — `s3://`, `gs://`, `pvc://`, `hf://`, `http(s)://`, and OCI — by internal switch. Splitting OCI into per-backend webhooks would diverge from that pattern without an established reason in the codebase.

2. **Single source of truth for mode resolution.** `ResolveOciModelMode` lives in `pkg/types/config.go:48` and is consulted from exactly two places: the registration gate in `mutator.go` and the dispatch site in `InjectModelcar`. With multi-webhook, that resolution would either be duplicated across N registration gates or extracted into a helper called from each — same code, more glue, easier to forget to update one site.

Fair trade-offs the multi-webhook alternative would buy:

- Cleaner per-backend failure isolation — a bug in `oci+fetch://` materialization wouldn't risk affecting `oci+native://`.
- Independent per-webhook configuration (`failurePolicy`, `sideEffects`, `matchPolicy`).
- Smaller blast radius for per-backend integration tests.

None of those felt heavy enough to justify the split now, but if Step 4 of #4083 (arbitrary OCI artifacts) materially expands the backend surface, refactoring is contained: `ResolveOciModelMode` already gives us the discriminator, the dispatch sites in injector and reconciler would split into per-backend mutators, and the gate becomes a mode → mutator-set translation. Worth revisiting at that point.

Re: deprecation — filed the tracking issue: #5679.

pkg/utils/storage.go
kliukovkin: Good observation — the naming differs intentionally between modelcar and native.

`ConfigureModelcarToContainer` uses indexed `kserve-provision-location-N` (via `ModelcarNames(ociIndex)`) because each OCI URI needs its own emptyDir volume: the modelcar sidecar creates a `/proc/<PID>/root` symlink specific to its container image, and a shared emptyDir would cause symlink conflicts. The index keeps those distinct.

ImageVolume has no such sharing concern — the container runtime mounts the OCI image directly at `modelPath`, so the natural identifier is the path itself. `GetVolumeNameFromPath` (e.g., `/mnt/models` → `mnt-models`) keeps the volume name readable from the pod's volume listing and avoids coupling it to the URI's position in the spec, which keeps re-admission idempotent (same path → same name across reorderings).

Happy to switch to the `kserve-provision-location-N` convention if you'd rather keep naming uniform across OCI backends — it's a small change.

brettmthompson: Your understanding here is correct. I especially like your point about how the use of `GetVolumeNameFromPath` makes the operation idempotent while using `kserve-provision-location-N` does not allow for idempotency.

With your current implementation all scenarios have distinct naming conventions:
1. Legacy = `kserve-provision-location`
2. Non-legacy non-OCI = single volume named after shared parent dir
3. Non-legacy OCI with model car = `kserve-provision-location-N` for each uri
4. Non-legacy OCI with image volume = volume per uri, named after uri path

> Happy to switch to the kserve-provision-location-N convention if you'd rather keep naming uniform across OCI backends

Curious to see if other's have any opinion on this. For speed I am ok with either atm.

pkg/utils/storage.go
```diff
@@ -582,3 +603,79 @@ func ConfigureModelcarToContainer(modelUri string, podSpec *corev1.PodSpec, targ
...
+	}
+
+	// Idempotent: skip if the pod-level Volume already exists.
+	// This is distinct from AddVolumeMountIfNotPresent, which deduplicates per-container
+	// VolumeMounts. Both layers need independent duplicate detection: duplicate pod Volumes
+	// are rejected by the API server, while duplicate container VolumeMounts are rejected
+	// by the kubelet. No shared helper covers pod-level Volumes, so this loop is necessary.
+	for _, v := range podSpec.Volumes {
+		if v.Name == volName {
+			return nil
```
brettmthompson: I was looking this part over a little more - is returning correct here, meaning no volume or volume mount is created, or should we simply skip volume creation if the volume already exists in the pod?

I believe model car storage currently has a single shared volume that is mounted to the pod for legacy URIs (mount to the shared volume per component). With this logic a volume per component is required. I think the same should be done for legacy URIs with image volumes.

pkg/utils/storage.go
```diff
@@ -582,3 +603,79 @@ func ConfigureModelcarToContainer(modelUri string, podSpec *corev1.PodSpec, targ
...
+		Name: volName,
+		VolumeSource: corev1.VolumeSource{
+			Image: &corev1.ImageVolumeSource{
+				Reference:  imageRef,
+				PullPolicy: corev1.PullIfNotPresent,
+			},
+		},
+	})
+
+	AddVolumeMountIfNotPresent(targetContainer, volName, modelPath, true)
```
brettmthompson: I think the way we are adding volume mounts here make OCI images used in the past incompatible with image volumes (unless a storage path is explicitly specified). Model car expects models to be under the `/models/*` path in the container and will seamlessly link `/proc/<pid>/root/models` of the model car container to `mnt/models` in the inference container. With image volumes the same configuration will not successfully load by default.

A possible solution is to specify the `models` subpath in the volumemount. This will better align with the current expectations for model car.

pkg/utils/storage.go
kliukovkin: You're right — restructured the loop to mirror `ConfigureModelcarToContainer`'s pattern. The pod-level Volume check still guards against duplicate Volume entries (no upstream helper covers pod Volumes), but `AddVolumeMountIfNotPresent` is now always called so each target container correctly receives its mount.

Added a regression test for the multi-container case (same modelPath, two target containers → one Volume, two VolumeMounts).

See `3ce81e6`.

pkg/utils/storage.go
kliukovkin: Took your suggestion — default `subPath: "models"` so existing modelcar OCI images mount their `/models/*` at `/mnt/models/*` unchanged, matching the runtime contract vLLM / sklearn-server / TorchServe expect. Surveyed the published `quay.io/redhat-ai-services/modelcar-catalog` and KServe's official OCI provider docs to confirm `/models/` is the dominant convention.

Caveat: subPath on `ImageVolume` VolumeMounts was forbidden in K8s 1.31/1.32 alpha (kubernetes/kubernetes#125660 validation) and only added in 1.33 beta (kubernetes/kubernetes#130135). Bumped the `OciImageVolumeCompatible` condition with a new `ImageVolumeSubPathUnsupported` state for 1.31/1.32, so operators on those alpha clusters get an actionable message ("Upgrade to K8s 1.33+ for full oci+native:// support") rather than a confusing admission failure.

URI-fragment override for custom image layouts (`oci+native://image:tag#path`) stays as the deferred followup originally noted in the PR description — small change, can land right after merge if there's demand.

See `65882ae`.

test/e2e/predictor/test_huggingface_vllm_cpu.py
kliukovkin: Switched to your suggestion in 24df525: replaced `VLLM_USE_V1=0` with `VLLM_WORKER_MULTIPROC_METHOD=spawn` (Cory's listed alternative on Slack, future-proof since `VLLM_USE_V1` is what gets removed in vLLM 0.15+).

Thanks for the catch and for clarifying which env was actually deprecated — the original ping was on a different track. Applied across all five vLLM CPU test fixtures (four Qwen variants + one bge-reranker) for consistency.

test/e2e/predictor/test_huggingface_vllm_cpu.py
kliukovkin: Resolved in the same commit (24df525) — `VLLM_ENABLE_V1_MULTIPROCESSING` dropped from all five fixtures alongside the switch to `VLLM_WORKER_MULTIPROC_METHOD=spawn`. Env block is down to `VLLM_CPU_KVCACHE_SPACE` + the multiproc method now.

kliukovkin:
CI surface update on 848a9a6:

- `test-vllm-runtime (helm)` was failing with the same vLLM v1 EngineCore CPU init pattern Cory diagnosed for `test-huggingface-server-vllm`. The fix lived one level up — added `VLLM_WORKER_MULTIPROC_METHOD=spawn` to `config/runtimes/kserve-vllmserver.yaml`, which cascades to all 8 test functions in `test_vllm.py` and benefits production users running vLLM on CPU nodes. Same pattern as Round 7, just at the runtime layer rather than the test fixtures. Diff is +4 lines across the source runtime + its regenerated chart copy.

- `test-quick-install (kustomize)` is failing in the MetalLB ConfigMap setup step (`get_metallb_config | kubectl apply -f -`) — the CI runner uses the old MetalLB v0.9.x ConfigMap API that was removed in v0.13+. The job hits this before any KServe code or configmap is applied; our diff vs master in that path is +24 lines / 0 deletions, purely additive config fields. Out of scope for this PR.

Local gates clean. Once CI confirms the vLLM runtime fix, ready for the final review pass.

tarilabs:
EDIT: sorted.

hi, I'm a User of KServe, I want to thank you for this PR 🙏 

> subPath is deferred. Some model images pack files under a subdirectory rather than
serving directly from /mnt/models. The materializer currently mounts the image root at
/mnt/models

It's not "some", but practically _all_ ModelCar ("model images") following the [KServe documentaton](https://kserve.github.io/website/docs/model-serving/storage/providers/oci?_highlight=modelcar#prepare-an-oci-image-with-model-data)

> ```
> COPY data/ /models/
> ```

(from [KServe docs](https://kserve.github.io/website/docs/model-serving/storage/providers/oci?_highlight=modelcar#prepare-an-oci-image-with-model-data))

meaning effectively "model images" following the instruction of the KServe documentation for the ModelCar, would contain model weights ~like:

```
/models/:
drwxr-xr-x. 1 user user         26 Jun 18 08:28 .
dr-xr-xr-x. 1 user user         28 Jun 18 09:02 ..
-rw-rw-r--. 1 user user       3196 Jun 10 09:13 README.md
-rw-rw-r--. 1 user user        608 Jun 10 09:13 config.json
-rw-rw-r--. 1 user user        566 Jun 10 09:13 eval_results.json
-rw-rw-r--. 1 user user        124 Jun 10 09:13 generation_config.json
-rw-rw-r--. 1 user user 1234567890 Jun 10 09:13 model.safetensors
-rw-rw-r--. 1 user user       3197 Jun 15 08:39 modelcard.md
-rw-rw-r--. 1 user user        551 Jun 10 09:13 special_tokens_map.json
-rw-rw-r--. 1 user user    1842767 Jun 10 09:13 tokenizer.json
-rw-rw-r--. 1 user user     499723 Jun 10 09:13 tokenizer.model
-rw-rw-r--. 1 user user       1289 Jun 10 09:13 tokenizer_config.json
```

inside `/models/`,

not root (`/`),

so mounting the root of the "model images" or the ModelCar would result in the model weights being only accessible in a subdirectory of the intended mountmount, and by my very limited experiences ServingRuntimes do not support that.

Hence, I very humbly believe `subPath` is not deferrable so to have forward-compatibility of existing ModelCar ("model images") to be used in this way, _without requiring to re-package, re-publish Models_

I thought was worth of sharing in the hope it's helpful angle to consider

brettmthompson:
@tarilabs I believe your concern is already addressed :smile: See this [thread](https://github.com/kserve/kserve/pull/5558#discussion_r3424689477)

tarilabs:
> @tarilabs I believe your concern is already addressed 😄 See this [thread](https://github.com/kserve/kserve/pull/5558#discussion_r3424689477)

didn't notice as it was folded 😅 thank you for making me notice I can see in 65882ae

kliukovkin:
Thanks @tarilabs for the careful read — you're absolutely right that "some" understated it; the `/models/` layout is the de-facto convention for the whole modelcar catalog, which is why Round 6 made `subPath: "models"` the default (and surfaced the K8s 1.31/1.32 alpha caveat for those clusters). The "subPath is deferred" wording in the PR description is now stale — fixing that in the next step. Appreciate the second eye.

kliukovkin:
Pushed regen fix in 29a60fe. The precommit-check failure on 848a9a6 was my miss — Round 8 added `VLLM_WORKER_MULTIPROC_METHOD=spawn` to `config/runtimes/kserve-vllmserver.yaml` and the Helm chart copy but the 3 `hack/setup/quick-install/*-with-manifests.sh` scripts that embed the runtime YAML inline weren’t regenerated in the same pass. CI’s precommit-check caught the drift; full regen via `make generate-quick-install-scripts` brings them in sync. Diff is +6 lines across the 3 scripts, nothing else.

Thanks for the /rerun on the prior HEAD — the new commit supersedes that run. Workflow approval gate will need one more nudge once CI starts the fresh batch.

kliukovkin:
Rebased onto current master in 9f66bde to resolve the conflict introduced by #5382 (TestApplyConfidentialConfig added at the end of `storage_initializer_injector_test.go`). Union-merge — kept both upstream's new test and our `TestInjectModelcarOciNative` / `TestCommonStorageInitializationWithOciNativeURI`. Diff vs the prior HEAD is just the +70 lines of upstream's test function; `go build`, `go vet`, `make go-lint` all clean.

Workflow approval gate reset on the push — last cycle, all review threads addressed across rounds 3-9.

kliukovkin:
Ran the full local CI suite on 7d90e98 — everything we can run without a GPU cluster is green:

- `go build` / `go vet` ✓
- `make go-lint` ✓ (0 issues)
- `go test ./pkg/...` with envtest ✓ (including `pkg/webhook/admission/pod/...` where the Round 10 compile error surfaced)
- `make precommit` ✓ (PR-surface clean; 85 codegen drifts are upstream `a6a11b34c` dynamic-year template churn, unrelated)
- `kubectl kustomize` on `config/default` + `config/overlays/test` ✓

Plus installed a pre-push hook locally that re-validates HEAD compiles before any future push, so the brace-loss class of failure can't repeat. Workflow approval gate just needs the one nudge to kick off the fresh CI batch on this HEAD.

kliukovkin:
Final CI rollup on 7d90e98 — 88 successful, 1 real failure (`test-vllm-runtime (kustomize)`), plus the `enforce-required-checks` aggregator.

The kustomize failure is the same class of vLLM CPU image instability that already affects `test-huggingface-server-vllm (kustomize)` on master (failing since 2026-05-04). Specifically: `VllmWorker-0` crashes during KV-cache profiling in `shm_broadcast.py:701` with a shared-memory queue teardown RuntimeError — a different failure path than the fork-init hang Round 8 addressed. Verified our `VLLM_WORKER_MULTIPROC_METHOD=spawn` env var IS correctly deployed (controller logs confirm) — the spawn switch fixes the multiprocessing init phase but doesn't reach this deeper KV-cache crash, which appears to be a CPU image bug. The helm variant of the same test passed on this HEAD.

Brett's LGTM still applies; all review threads addressed across rounds 3-11; tracking issue #5679 filed for the deprecation arc. Ready for the final `/approve` `/lgtm` whenever a Member has a moment. @pierDipi @cjohannsen-cloudera

kliukovkin:
Correction on the prior triage — I was wrong on two specific points, fixed in 7d27d59.

**What I got wrong:**

The earlier comment implied the `test-vllm-runtime (kustomize)` failure was the same class as a master-baseline `test-huggingface-server-vllm (kustomize)` issue. Closer investigation shows that's a misframe: `test-vllm-runtime` was introduced by #4769 (@yashaswipiplani, merged 2026-06-12) and has never run on master CI (last master push CI was 2026-05-28, before #4769 merged). This PR is the first run to exercise the job. The Round 8 spawn env var is correctly delivered to the pod (verified via the env-merge code path in `MergeRuntimeContainers` at `pkg/controller/v1beta1/inferenceservice/utils/utils.go:252`, not just the kustomize build output).

**Actual root cause:**

`VllmWorker-0` triggers AOT torch compilation during EngineCore init, which takes ~130s on the CI CPU runner. `shm_broadcast`'s startup poll window is 2×60s = 120s. EngineCore gets cancelled at 120s, worker finishes compilation at 130s, hits `RuntimeError: cancelled`. Tests 2 and 3 are cascade failures (test 1's pod consumes node resources, subsequent ISVCs can't schedule).

**Fix in 7d27d59:**

Test-overlay-only kustomize patch (`config/overlays/test/clusterresources/kserve-vllmserver-test-patch.yaml`) adds `--enforce-eager` to the vllmserver args. Skips AOT compilation; startup drops from ~130s to <1s, well under the timeout window. Production runtime config (`config/runtimes/kserve-vllmserver.yaml`) is unchanged — patch lives only in the test overlay so CI runners get the fast path while production stays on the default compilation behaviour.

@yashaswipiplani — flagging since the test is yours from #4769. The `--enforce-eager` patch is intentionally test-overlay-only and reversible; happy to revert and let you address upstream-side if you'd prefer that path. Otherwise the patch unblocks the kustomize job.

Apart from this one job: 88 jobs successful on the prior HEAD, all other E2E paths green. Brett's LGTM still applies; all review threads addressed across rounds 3-12; tracking issue #5679 filed for the deprecation arc.

@pierDipi @cjohannsen-cloudera — once CI confirms the test-overlay patch resolves the timeout, ready for the final `/approve` `/lgtm`.

yashaswipiplani:
> Correction on the prior triage — I was wrong on two specific points, fixed in [7d27d59](https://github.com/kserve/kserve/commit/7d27d59516ebdf1742aefa1d85337f4d1408cbf1).
> 
> **What I got wrong:**
> 
> The earlier comment implied the `test-vllm-runtime (kustomize)` failure was the same class as a master-baseline `test-huggingface-server-vllm (kustomize)` issue. Closer investigation shows that's a misframe: `test-vllm-runtime` was introduced by #4769 (@yashaswipiplani, merged 2026-06-12) and has never run on master CI (last master push CI was 2026-05-28, before #4769 merged). This PR is the first run to exercise the job. The Round 8 spawn env var is correctly delivered to the pod (verified via the env-merge code path in `MergeRuntimeContainers` at `pkg/controller/v1beta1/inferenceservice/utils/utils.go:252`, not just the kustomize build output).
> 
> **Actual root cause:**
> 
> `VllmWorker-0` triggers AOT torch compilation during EngineCore init, which takes ~130s on the CI CPU runner. `shm_broadcast`'s startup poll window is 2×60s = 120s. EngineCore gets cancelled at 120s, worker finishes compilation at 130s, hits `RuntimeError: cancelled`. Tests 2 and 3 are cascade failures (test 1's pod consumes node resources, subsequent ISVCs can't schedule).
> 
> **Fix in [7d27d59](https://github.com/kserve/kserve/commit/7d27d59516ebdf1742aefa1d85337f4d1408cbf1):**
> 
> Test-overlay-only kustomize patch (`config/overlays/test/clusterresources/kserve-vllmserver-test-patch.yaml`) adds `--enforce-eager` to the vllmserver args. Skips AOT compilation; startup drops from ~130s to <1s, well under the timeout window. Production runtime config (`config/runtimes/kserve-vllmserver.yaml`) is unchanged — patch lives only in the test overlay so CI runners get the fast path while production stays on the default compilation behaviour.
> 
> @yashaswipiplani — flagging since the test is yours from #4769. The `--enforce-eager` patch is intentionally test-overlay-only and reversible; happy to revert and let you address upstream-side if you'd prefer that path. Otherwise the patch unblocks the kustomize job.
> 
> Apart from this one job: 88 jobs successful on the prior HEAD, all other E2E paths green. Brett's LGTM still applies; all review threads addressed across rounds 3-12; tracking issue #5679 filed for the deprecation arc.
> 
> @pierDipi @cjohannsen-cloudera — once CI confirms the test-overlay patch resolves the timeout, ready for the final `/approve` `/lgtm`.

Thank you so much for the fix. I think we can get it in as part of this PR only so that you are not blocked. Although I am curious why did the vLLM e2e tests never ran on Master. They did ran on the PR and passed(I even checked logs) before merge. Regardless, thanks for the fix

yashaswipiplani:
Also, I see @Jooho had a comment as to how to add a similar patch on some other CI failure fix: https://github.com/kserve/kserve/pull/5657#discussion_r3396308403. Should we take it here as well?

kliukovkin:
Thanks @yashaswipiplani. Looked at #5657 — Jooho's pattern is for production-appropriate envs (auto-propagate via base config). `--enforce-eager` is the opposite: it disables `torch.compile()`, which would be a GPU performance regression in production. Scoping to the test overlay keeps production on AOT while CI's CPU runners get the fast path. Happy to skip vLLM CPU tests entirely instead if maintainers prefer.

Re: never ran on master — timing only. Last master push CI was 2026-05-28; #4769 merged 2026-06-12, no master push since. Our PR is just first to trigger it.

Ready for /approve /lgtm @pierDipi @cjohannsen-cloudera once CI's green on 7d27d59.

yashaswipiplani:
> Thanks @yashaswipiplani. Looked at #5657 — Jooho's pattern is for production-appropriate envs (auto-propagate via base config). `--enforce-eager` is the opposite: it disables `torch.compile()`, which would be a GPU performance regression in production. Scoping to the test overlay keeps production on AOT while CI's CPU runners get the fast path. Happy to skip vLLM CPU tests entirely instead if maintainers prefer.
> 
> Re: never ran on master — timing only. Last master push CI was 2026-05-28; #4769 merged 2026-06-12, no master push since. Our PR is just first to trigger it.
> 
> Ready for /approve /lgtm @pierDipi @cjohannsen-cloudera once CI's green on [7d27d59](https://github.com/kserve/kserve/commit/7d27d59516ebdf1742aefa1d85337f4d1408cbf1).

Understood. Thank you so much for the detailed explanation

kliukovkin:
test-vllm-runtime helm flipped to failing on the prior HEAD (same AOT vs shm_broadcast race, just won the race in kustomize and lost in helm this time). Patching each path individually is whack-a-mole, so going with the alternative I offered earlier: skipping the three vLLM CPU tests with explicit reason in b4488cc. Round 12 `--enforce-eager` patch stays in place — it takes effect automatically when the tests are re-enabled.

Re-enable when: GPU CI runner available, OR upstream vLLM resolves the race, OR fixtures move to a smaller model that compiles under 120s.

@yashaswipiplani — flagging since it's your test from #4769. Decorator is `pytest.mark.skip`; trivial to remove. Happy to revisit if you'd rather a different approach.

@pierDipi @cjohannsen-cloudera — should be the actual last cycle. Ready for /approve /lgtm once CI's green on b4488cc.

kliukovkin:
test-huggingface-server-vllm (helm + kustomize) are failing on HuggingFace CDN 403 errors during model weight download — HF recently migrated to XET/Xethub CDN and the CI runner is hitting `403 Forbidden` rate-limit/auth errors on `us.aws.cdn.hf.co/xet-bridge-us/...`. Not fixable from our PR; needs either HF_TOKEN injected into CI secrets or retry once HF CDN recovers.

test-vllm-runtime had a 4th test (test_vllm_openai_text_completion_streaming) that wasn't in Round 13's skip list — fixing in Round 14. Round 12 --enforce-eager and the 3 prior skips remain in place.

kliukovkin:
Three fixes on 81a5a81:

- **Round 14**: skipped the 4th vLLM CPU test missed in Round 13.
- **Round 15**: kubeconfig load fix in `test_oci_native*.py` — `_get_k8s_minor()` was silently false-skipping with `-1`.
- **Round 16**: added `oci+native://` to `SupportedStorageURIPrefixList` in `pkg/controller/v1beta1/inferenceservice/utils/utils.go`. Caught running our full local CI replica against a built controller image — `ValidateStorageURI` was rejecting the scheme at the controller level before `ParseOciScheme` could normalize. Storage tests had been false-skipping past this. End-to-end verified on local kind: `test_oci_native_image_volume_spec` PASSES.

`test-huggingface-server-vllm` is failing on HF CDN 403s (HF migrated to XET/Xethub; rate-limit on `us.aws.cdn.hf.co/xet-bridge-us/...`). Not PR-side; needs `HF_TOKEN` in CI secrets to unblock.

@cjohannsen-cloudera — can you please "Approve and run workflows"?

config/configmap/inferenceservice.yaml
```diff
@@ -187,6 +189,14 @@ data:
            # using a source URL with an "oci://" schema.
            "enableModelcar": false,
 
+           # enableOciModelSupport enables any OCI-backed model storage path (modelcar, native ImageVolume, or fetch).
+           # This is the newer master switch; enableModelcar is kept as a backcompat alias for the "modelcar" mode.
+           "enableOciModelSupport": false,
+
+           # ociModelMode selects the materialization strategy when a storageUri uses oci:// without an explicit
```
lizzzcai, line 196: thanks for the PR. when ociModelMode: "native", `oci://` will be the same as `oci+native://`, am i correct?

lizzzcai:
Hi @kliukovkin ,thanks for the PR. I look into the oci spec, raising a design question before this merges, to make sure we're aligned on the scope of `oci+native://` vs the deferred `oci+fetch://` (Step 3 of #4083).

**Background: there are two distinct OCI artifact formats in use today.**

The first is what this PR targets — an OCI container image (standard `application/vnd.oci.image.layer.v1.tar+gzip` layers), where the model files live under `/models/` in the image filesystem. The modelcar catalog and `rhuss/kserving-example-sklearn:1.0` are examples of this.

The second is the **ModelPack OCI artifact format**, now a CNCF sandbox specification at `github.com/modelpack/model-spec`. This is the format that Docker's AI model catalog (`hub.docker.com/u/ai`) uses — the `docker model push` / `docker model pull` workflow produces artifacts with `artifactType: application/vnd.cncf.model.manifest.v1+json` and weight layers typed as `application/vnd.cncf.model.weight.v1.raw`. Docker's blog post ([OCI Artifacts for AI Model Packaging](https://www.docker.com/blog/oci-artifacts-for-ai-model-packaging/)) describes this as the direction for standardized model distribution. The [ai/gemma4:E2B](https://hub.docker.com/r/ai/gemma4) manifest, for example, looks like:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.cncf.model.manifest.v1+json",
  "layers": [
    {
      "mediaType": "application/vnd.cncf.model.weight.v1.raw",
      "size": 3184494720,
      "annotations": {
        "org.cncf.model.filepath": "gemma-4-E2B-it-UD-Q4_K_XL.gguf"
      }
    }
  ]
}
```

**The question: should `oci+native://` handle both formats?**

The ModelPack [getting-started docs](https://github.com/modelpack/model-spec/blob/main/docs/getting-started.md) describe mounting these artifacts in Kubernetes via ImageVolume (KEP-4639) as a first-class use case. CRI-O explicitly supports this via `oci_artifact_mount_support: true` (default on). For CRI-O clusters, `spec.volumes[].image` with a ModelPack reference may already work at the runtime layer.

For containerd clusters (including the one I checked: containerd 2.2.4 on K8s 1.35.5), containerd's ImageVolume implementation goes through the overlay snapshotter, which expects filesystem tarballs. It would not know how to unpack `vnd.cncf.model.weight.v1.raw` blobs into a directory tree. So containerd clusters would need KServe to handle the extraction explicitly — reading `org.cncf.model.filepath` annotations, writing files to `/mnt/models/`, before the pod starts.

**If `oci+native://` is extended to detect the `artifactType` and handle ModelPack extraction itself** (rather than delegating to the container runtime), it could cover both formats under one scheme and make `oci+fetch://` redundant. The file path would come from the `org.cncf.model.filepath` annotation on each layer rather than from a `/models/` subpath in an image filesystem.

The questions I'd like to get clarity on before this PR merges:

1. Is the current scope of `oci+native://` intentionally limited to OCI container images (filesystem layers), with ModelPack artifacts explicitly deferred to `oci+fetch://`?
2. Or is the expectation that `oci+native://` should eventually handle the ModelPack artifact format too (or only support the ModelPack artifact format which seems to be a standard now) — in which case the `subPath: "models"` assumption baked in here (commit `65882ae`) would not apply to ModelPack artifacts?
3. If `oci+native://` is extended to handle ModelPack, is `oci+fetch://` still needed, or does it become redundant?

Happy to follow up in #4083 if this is out of scope for the current PR — just want to make sure the naming and design contract are intentional before the scheme lands in a release.

config/configmap/inferenceservice.yaml
kliukovkin: Correct. `ParseOciScheme` returns `mode=""` for bare `oci://`; the webhook then fills the effective mode from the configured `ociModelMode`. So `oci://...` ≡ `oci+native://...` when the mode is `"native"`. Same pattern for `"modelcar"` and (once Step 3 lands) `"fetch"`. Covered in `pkg/utils/storage_test.go` / `TestParseOciScheme`.

kliukovkin:
@lizzzcai —

Q1: Yes, `oci+native://` is scoped to OCI container images by design. "Native" = the container runtime mounts via ImageVolume (KEP-4639); containerd's snapshotter only unpacks filesystem layers, so the runtime-mounted path can only carry filesystem images. The `subPath: "models"` convention (commit 65882ae) applies in that world.

Q2: native stays scoped. ModelPack artifacts need artifact-aware extraction that containerd's ImageVolume path can't do — overlay snapshotter expects filesystem tarballs. CRI-O's `oci_artifact_mount_support` can, but tying the user-facing scheme to runtime-specific behavior would give silent cluster-to-cluster divergence. Extraction belongs on the KServe side.

Q3: `oci+fetch://` isn't redundant — it's Step 3 of @rhuss's original 4-step plan above (KServe-side download, runtime-agnostic). ModelPack is Step 4 on top of fetch. I have started working on Step 3 and will raise the PR once #5558 merges.

Happy to add you as a reviewer on both — your runtime-layer read was exactly the right framing.

kliukovkin:
@cjohannsen-cloudera @pierDipi — current red checks on 81a5a8121 are external (CI infra), not PR-side. Quick triage:

- `test-llm` (helm) + `test-kourier` (kustomize): artifact-pipeline race — `huggingfaceserver-${TAG}` wasn't available when tests tried to download it. Both were green on Round 13 (b4488cc8c).
- `test-llmisvc-wva-autoscaling-hpa` (helm): InferencePool CRD (`inference.networking.x-k8s.io/v1alpha2`) not installed in helm cluster; llmisvc webhook unreachable. Green on Round 13.
- `test-huggingface-server-vllm` (helm + kustomize): pre-existing HF CDN/XET timeout — pytest hangs ~69 min then killed. Needs `HF_TOKEN` in CI secrets to unblock permanently.

Asymmetry rules out code: each newly-failing surface fails on one install method but passes on the other (helm vs kustomize).

A "Re-run failed jobs" should clear the first three. HF tests stay red until the token's in CI secrets.

cjohannsen-cloudera (approved):
LGTM

PR #5545, opened 2026-05-16
fix(llmisvc): use `model-server-protocol-metrics` 

pierDipi:
The scheduler migration was using the wrong plugin name for the
core metrics extractor, causing a mismatch with the actual GIE
plugin type. Also adds an e2e test with a custom scheduler template
to validate the migration path for deprecated metric flags.

transformation logic assumes core-metrics-extractor but scheduler 0.7 (GIE 1.4) expects `model-server-protocol-metrics`
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.4/pkg/epp/framework/plugins/datalayer/extractor/metrics/factories.go#L28

vs GIE 1.5 using `core-metrics-extractor` (future version)
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.5/pkg/epp/framework/plugins/datalayer/extractor/metrics/factories.go#L31

Now we get the correct args for the matching llm-d scheduler version (0.7 currently)
```yaml
  - args:
    - --cert-path
    - /var/run/kserve/tls
    - --pool-group
    - inference.networking.x-k8s.io
    - --pool-name
    - scheduler-custom-template-test-inference-pool
    - --pool-namespace
    - kserve-ci-e2e-test
    - --zap-encoder
    - json
    - --grpc-port
    - "9002"
    - --grpc-health-port
    - "9003"
    - --secure-serving
    - --model-server-metrics-scheme
    - https
    - --config-text
    - |
      apiVersion: inference.networking.x-k8s.io/v1alpha1
      kind: EndpointPickerConfig
      plugins:
      - type: single-profile-handler
      - type: queue-scorer
      - type: active-request-scorer
      - type: prefix-cache-scorer
      - name: model-server-protocol-metrics
        parameters:
          defaultEngine: vllm
          engineConfigs:
          - kvUsageSpec: vllm:kv_cache_usage_perc
            name: vllm
          engineLabelKey: inference.networking.k8s.io/engine-type
        type: model-server-protocol-metrics
      schedulingProfiles:
      - name: default
        plugins:
        - pluginRef: queue-scorer
          weight: 2
        - pluginRef: active-request-scorer
          weight: 2
        - pluginRef: prefix-cache-scorer
          weight: 3
    command:
    - /app/epp
    - --pool-name
    - scheduler-custom-template-test-inference-pool
    - --pool-namespace
    - kserve-ci-e2e-test
    - --zap-encoder
    - json
    - --grpc-port
    - "9002"
[...]

PR #5560, opened 2026-05-20
feat(llmisvc): migrate non-zero threshold to prefix-based-pd-decider

### LINKED ISSUE #5561: feat(llmisvc): migrate non-zero threshold to prefix-based-pd-decider during v0.7 scheduler upgrade
## Problem

When upgrading llm-d-inference-scheduler from v0.6 to v0.7, the controller's scheduler migration logic (#5433) skips configs that have a non-zero `threshold` value on the `pd-profile-handler` / `disagg-profile-handler` plugin. These deployments remain on the deprecated v0.6 schema with no plugin rename and no parameter restructuring.

The v0.7 binary still accepts the old format as a deprecated alias, but this leaves users on a dead-end config that will eventually be removed.

## Background

In v0.6, the `pd-profile-handler` plugin used a `threshold` parameter representing the number of **non-cached characters** in user input to decide whether to disaggregate:

```go
if (1.0-hitPercentagePrefix)*float64(len(userInput)) < float64(h.pdThreshold) {
    // skip prefill
}
```

In v0.7, this was replaced by dedicated decider plugins:
- `always-disagg-pd-decider` — always disaggregates (equivalent to `threshold: 0`)
- `prefix-based-pd-decider` — uses `nonCachedTokens` (in **tokens**, not characters)

The migration for `threshold: 0` was implemented in #5433. The non-zero case was deferred because there was no confirmed conversion formula at the time.

## Proposed Solution

Migrate non-zero threshold values using:

```
nonCachedTokens = ceil(threshold / 4)
```

The `/4` is based on the empirical ~4:1 character-to-token ratio for English text
(confirmed by upstream maintainers).

### Migration mapping

| Old config (v0.6)    | New config (v0.7)                                                  |
|----------------------|--------------------------------------------------------------------|
| `threshold: 0`       | `always-disagg-pd-decider` (no params) — already implemented      |
| `threshold: 100`     | `prefix-based-pd-decider` with `nonCachedTokens: 25`              |
| `threshold: 5`       | `prefix-based-pd-decider` with `nonCachedTokens: 2`               |

### Important caveats to document

- The original v0.6 threshold was never properly tuned (per upstream feedback)
- The 4:1 ratio is an approximation — actual ratio varies by language and tokenizer
- A runtime warning should be logged advising users to review the converted value

## Acceptance Criteria

- [x] Non-zero threshold configs are migrated instead of skipped
- [x] `prefix-based-pd-decider` plugin entry is added to the top-level plugins list
- [x] Plugin rename (`pd-profile-handler` → `disagg-profile-handler`) applies to all configs
- [x] Runtime warning log emitted with original threshold and converted nonCachedTokens
- [x] Code comments document the unit shift (chars → tokens) and conversion rationale
- [x] Unit tests cover threshold values: 100, 5, 1, 0.5, and idempotency
- [x] Existing migration tests updated (no regressions)

vivekk16:
## Summary

Closes #5561

Completes the v0.6 → v0.7 scheduler migration for non-zero `threshold` values in the disagg-profile-handler plugin, which were previously skipped entirely by #5433.

- `threshold: 0` → `always-disagg-pd-decider` (already implemented in #5433)
- `threshold > 0` → `prefix-based-pd-decider` with `nonCachedTokens = ceil(threshold / 4)`

The `/4` conversion is based on the ~4:1 character-to-token ratio for English text. The original v0.6 threshold (non-cached characters) was never properly tuned, so a runtime warning is logged advising users to review the converted value for their workload.

## Changes

- **Removed** `hasNonZeroThreshold` guard that skipped migration for non-zero thresholds
- **Added** Path C in `WithMigrateDisaggProfileParams` for `threshold > 0` without `deciderPluginName`
- **Added** `thresholdToNonCachedTokens` helper (handles int64/float64 from YAML, minimum 1 token)
- **Added** runtime warning log with original threshold and converted nonCachedTokens values
- **Added** code comments documenting the unit shift (chars → tokens) and conversion rationale

## Test plan

- [x] New unit tests for Path C: threshold 100 → 25, threshold 5 → 2, threshold 1 → 1
- [x] New unit test for idempotency (prefix-based-pd-decider already exists)
- [x] New `TestThresholdToNonCachedTokens` for int64/float64 edge cases
- [x] New `TestFullMigrationPipelineNonZeroThreshold` end-to-end test
- [x] Updated existing "skips all" tests to expect migration proceeds
- [x] All existing tests still pass (no regressions)

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/scheduler.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"math"
 	"path"
 	"slices"
 	"sort"
@@ -996,8 +997,25 @@ func WithRenamePlugin(oldType, newType string) mutateSchedulerConfigFunc {
 }
 
 // WithMigrateDisaggProfileParams migrates the disagg-profile-handler (formerly
-// pd-profile-handler) from the old flat deciderPluginName parameter to the new
-// deciders map structure introduced in llm-d-inference-scheduler v0.7.0.
+// pd-profile-handler) from the old flat deciderPluginName/threshold parameters
+// to the new deciders map structure introduced in llm-d-inference-scheduler v0.7.0.
+//
+// Migration paths:
+//
+//	Path A: deciderPluginName present → use that name in the deciders map, strip threshold.
+//	Path B: threshold == 0, no deciderPluginName → always-disagg-pd-decider.
+//	Path C: threshold > 0, no deciderPluginName → prefix-based-pd-decider with nonCachedTokens.
+//
+// Path C conversion rationale (threshold → nonCachedTokens):
+// In v0.6, `threshold` was the number of **non-cached characters** in the user input
+// that determined whether to disaggregate:
+//
+//	if (1.0 - hitPercentagePrefix) * len(userInput) < threshold { /* skip prefill */ }
+//
+// In v0.7, the prefix-based-pd-decider uses `nonCachedTokens` (in **tokens**, not chars).
+// The conversion uses ceil(threshold / 4) based on the empirical ~4:1 character-to-token
+// ratio for English text. Note: the original v0.6 threshold was never properly tuned
+// (per upstream feedback), so the converted value should be reviewed for each workload.
 func WithMigrateDisaggProfileParams(ctx context.Context, u *unstructured.Unstructured) error {
 	val, found, err := unstructured.NestedFieldNoCopy(u.Object, "plugins")
 	if err != nil || !found {
@@ -1008,7 +1026,11 @@ func WithMigrateDisaggProfileParams(ctx context.Context, u *unstructured.Unstruc
 		return nil
 	}
 
-	needDeciderPluginEntry := false
+	type deciderPluginEntry struct {
+		pluginType string
+		params     map[string]interface{}
+	}
+	var deciderPluginsToInject []deciderPluginEntry
 
 	for _, plugin := range plugins {
 		pluginMap, ok := plugin.(map[string]interface{})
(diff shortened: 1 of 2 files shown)


KillianGolds (commented):
Nice work on this. I pulled the branch locally and went through it. Build, `go vet`, and `gofmt` are all clean, and the full scheduler unit suite passes (the threshold paths, the helper edge cases, idempotency, and the full-pipeline test).

I checked the new plugin and parameter names against the llm-d-inference-scheduler v0.7.0 source and they all match: `disagg-profile-handler`, the `deciders.prefill` map, `prefix-based-pd-decider`, `nonCachedTokens`, and `always-disagg-pd-decider`. I also liked the `÷4` choice. It looks like the decider does the same thing internally (the README says it estimates tokens by "dividing raw character length by 4"), so the conversion feels well grounded. Sending `threshold:0` to `always-disagg-pd-decider` instead of `nonCachedTokens:0` also looks right, since `0` seems to turn the decider off. The min-1 floor helps for the same reason.

One thing I wanted to raise: I think the plugin order in the output might cause a problem at startup. The migration adds the new `prefix-based-pd-decider` at the end of the list, after the `disagg-profile-handler` that points to it. From reading the loader (GIE v1.4.0 and the v0.7.0 handler factory), it looks like plugins load in list order, and each one is only registered after it is built. The handler is built first, and it errors with `deciders.prefill plugin not found` if the decider is not registered yet. The README also says deciders should be "declared earlier in the same configuration." If that is right, the config may fail to load. I might be missing something here, so it is worth a second look.

If it does turn out to be an issue, I think the fix would be to declare the decider before the handler that references it rather than at the end of the list. I also noticed Path B (`always-disagg-pd-decider`) adds it the same way, so it may be worth checking both, and a small test that the decider comes before the handler would help lock it down.

Everything else looks good to me.

andresllh (commented):
/lgtm
Looks good on my end.

vivekk16:
@KillianGolds I've fixed both Path B and Path C to insert the decider before the handler instead of appending at the end. Also added a `validateDeciderOrder` test helper that runs on every migration test case to enforce the decider must appear before handler invariant, so this can't regress silently.

vivekk16:
@pierDipi I'll add e2e tests to cover the complete migration logic in a follow up PR.

pierDipi:
> @pierDipi I'll add e2e tests to cover the complete migration logic in a follow up PR.

why? the previous PR didn't have tests and we found the bug [too late](https://github.com/kserve/kserve/pull/5545)

vivekk16:
@pierDipi Added the e2e tests - https://github.com/kserve/kserve/pull/5564

spolti:
Hi @vivekk16 @pierDipi is this ready to go?

spolti (approved):

PR #5566, opened 2026-05-21
feat(ci): cache opt-125m in SeaweedFS

jlost:

Pre-cache facebook/opt-125m in SeaweedFS during E2E cluster setup to
eliminate HuggingFace downloads that cause ~15 min timeouts and
intermittent rate-limit failures in LLMISVC CI.

Changes:
- Add storage-initializer initContainer to the SeaweedFS init job that
  downloads opt-125m from HuggingFace and syncs it to S3
- Add `activeDeadlineSeconds: 600` to the init job to prevent orphaned
  jobs when network operations hang
- Deploy SeaweedFS in the LLMISVC-only CI path (previously only deployed
  in the KServe+LLMISVC path)
- Parametrize `hf://facebook/opt-125m` in test fixtures via
  `OPT_125M_MODEL_URI` env var (defaults to `hf://`, set to `s3://` in
  the LLMISVC workflow)
- Add S3 credentials secret for storage-initializer authentication

- [ ] LLMISVC E2E tests pass using `s3://example-models/facebook/opt-125m`
- [ ] KServe E2E tests pass (init job completes with additional opt-125m download)
- [ ] SeaweedFS deploys and init job completes within 300s timeout
- [ ] `activeDeadlineSeconds` is set on the init job

- The `kserve/storage-initializer:latest` image ref in the init job YAML
  is substituted to the CI-built tag via `sed` in `setup-kserve.sh`
- LoRA adapter URIs (`hf://edbeeching/opt-125m-lora`) are intentionally
  left as `hf://` since they're small (~2MB)
- `OPT_125M_MODEL_URI` defaults to `hf://facebook/opt-125m` so local
  dev/test without SeaweedFS continues to work

**Checklist**:

```release-note
NONE
```

### DIFF
--- a/config/overlays/test/s3-local-backend/seaweedfs-s3-creds-secret.yaml
+++ b/config/overlays/test/s3-local-backend/seaweedfs-s3-creds-secret.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: seaweedfs-s3-creds
+  annotations:
+    serving.kserve.io/s3-endpoint: "s3-service.kserve:8333"
+    serving.kserve.io/s3-usehttps: "0"
+    serving.kserve.io/s3-verifyssl: "0"
+type: Opaque
+stringData:
+  AWS_ACCESS_KEY_ID: s3admin
+  AWS_SECRET_ACCESS_KEY: s3admin123
--- a/.github/workflows/e2e-test-llmisvc.yaml
+++ b/.github/workflows/e2e-test-llmisvc.yaml
@@ -226,6 +226,8 @@ jobs:
       - name: Run LLMInferenceService E2E tests
         id: llmisvc-e2e-tests
         timeout-minutes: 90
+        env:
+          OPT_125M_MODEL_URI: "s3://example-models/facebook/opt-125m"
         run: |
           # Run only CPU tests for now using pytest markers (cluster_)
           # Available GPU vendors: amd, nvidia, intel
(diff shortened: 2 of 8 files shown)


pierDipi (commented):
One concern I have is that now we wouldn't be testing "using HF models" anymore

config/overlays/test/s3-local-backend/seaweedfs-s3-creds-secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Secret
+metadata:
+  name: seaweedfs-s3-creds
+  annotations:
+    serving.kserve.io/s3-endpoint: "s3-service.kserve:8333"
+    serving.kserve.io/s3-usehttps: "0"
+    serving.kserve.io/s3-verifyssl: "0"
+type: Opaque
+stringData:
+  AWS_ACCESS_KEY_ID: s3admin
```
gavrissh, line 11: Will there be issues in rotating these in regular intervals?

jlost: This is a CI-only, cluster-internal "secret" for the Seaweed S3. It doesn't need to be rotated.

jlost:
> One concern I have is that now we wouldn't be testing "using HF models" anymore

I think we should weigh the positives and negatives...

Pro: Fewer flakes, saves 10-15 minutes run time on the test suites that use those models

Con: Less e2e HF coverage -- But this is mitigated by the fact that the tests added in #5317 will still use hf:// to get the smaller LoRAs. So it will still have coverage, but in a less bandwidth-intensive context.

IMO, this is still a clear win @pierDipi

andresllh (commented):
/lgtm

PR #5573, opened 2026-05-23
feat(llmisvc): cache inferenceservice-config and watch for changes

KaveeshKhattar:

`LoadConfig` was calling the API server directly on every reconcile loop and in every event handler (`enqueueOnGatewayChange`, `enqueueOnHttpRouteChange`, `enqueueOnInferencePoolChange`, `enqueueOnConfigMapChange`, `updateRoutingStatus`, `expectedSchedulerDeployment`). This adds unnecessary API server pressure under high reconciliation load.

This PR switches all call sites to a cached controller-runtime client read via `r.loadConfig(ctx)`, backed by a field-selector-scoped informer cache for the `inferenceservice-config` ConfigMap.

It also adds a dedicated `Watches` entry so that changes to `inferenceservice-config` trigger reconciliation of all `LLMInferenceService` objects across namespaces.

The existing ConfigMap watch is updated with a predicate to exclude `inferenceservice-config` so the two paths do not overlap.

- [x] `TestLoadConfigFromClient` — verifies the cached client read path returns correct config
- [x] `TestRequestsForInferenceServiceConfigMapChangeEnqueuesAllServices`: verifies that a change to `inferenceservice-config` enqueues all `LLMInferenceService` objects across namespaces
- [x] Existing `TestNewSchedulerConfig` and `TestRequestsForInferencePoolChange` pass unchanged
- Full integration tests require `envtest`; local etcd binary unavailable, covered by CI

Closes two TODOs in the codebase:
- `router.go`: `// TODO: LoadConfig fetches the configmap from the API server on every reconciliation...`
- `controller.go`: `// TODO(ctrl): add watch on CfgMap with predicate and cache tuning to trigger reconcile when it changes`

The `enqueueOnConfigMapChange` handler previously fanned out to all namespaces when a ConfigMap in `KServeNamespace` changed. That behavior is preserved for `LLMInferenceServiceConfig` objects via the existing `enqueueOnLLMInferenceServiceConfigChange` handler. 
The `inferenceservice-config` ConfigMap now has its own dedicated handler for this fan-out.

### DIFF
--- a/cmd/llmisvc/main.go
+++ b/cmd/llmisvc/main.go
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apixclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -51,6 +52,7 @@ import (
 
 	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
 	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
 	"github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc"
 	kservescheme "github.com/kserve/kserve/pkg/scheme"
 	llmisvcwebhook "github.com/kserve/kserve/pkg/webhook/admission/llminferenceservice"
@@ -197,7 +199,16 @@ func main() {
 					Label: llmSvcCacheSelector,
 				},
 				&corev1.ConfigMap{}: {
-					Label: llmSvcCacheSelector,
+					Namespaces: map[string]cache.Config{
+						cache.AllNamespaces: {
+							LabelSelector: llmSvcCacheSelector,
+						},
+						constants.KServeNamespace: {
+							// Namespace-specific cache configs do not merge with AllNamespaces.
+							// Keep the system namespace scope limited to the global config read by LLMISVC.
+							FieldSelector: fields.OneTermEqualSelector("metadata.name", constants.InferenceServiceConfigMapName),
+						},
+					},
 				},
 				&appsv1.Deployment{}: {
 					Label: llmSvcCacheSelector,
@@ -234,7 +245,7 @@ func main() {
 		os.Exit(1)
 	}
 	v1alpha1ConfigValidator := &v1alpha1.LLMInferenceServiceConfigValidator{
-		ConfigValidationFunc:   createV1Alpha1ConfigValidationFunc(clientSet),
+		ConfigValidationFunc:   createV1Alpha1ConfigValidationFunc(mgr.GetAPIReader()),
 		WellKnownConfigChecker: wellKnownConfigChecker,
 	}
 	if err = v1alpha1ConfigValidator.SetupWithManager(mgr); err != nil {
@@ -242,7 +253,7 @@ func main() {
 		os.Exit(1)
 	}
 	v1alpha2ConfigValidator := &v1alpha2.LLMInferenceServiceConfigValidator{
-		ConfigValidationFunc:   createV1Alpha2ConfigValidationFunc(clientSet),
+		ConfigValidationFunc:   createV1Alpha2ConfigValidationFunc(mgr.GetAPIReader()),
 		WellKnownConfigChecker: wellKnownConfigChecker,
 	}
 	if err = v1alpha2ConfigValidator.SetupWithManager(mgr); err != nil {
@@ -370,8 +381,8 @@ func wellKnownConfigChecker(name string) bool {
 
 // validateLLMISVCConfig validates a v1alpha2 LLMInferenceServiceConfig by loading the controller
 // config and validating the template variables.
-func validateLLMISVCConfig(ctx context.Context, clientSet kubernetes.Interface, config *v1alpha2.LLMInferenceServiceConfig) error {
-	cfg, err := llmisvc.LoadConfig(ctx, clientSet)
+func validateLLMISVCConfig(ctx context.Context, reader client.Reader, config *v1alpha2.LLMInferenceServiceConfig) error {
+	cfg, err := llmisvc.LoadConfig(ctx, reader)
 	if err != nil {
 		return err
 	}
@@ -381,19 +392,19 @@ func validateLLMISVCConfig(ctx context.Context, clientSet kubernetes.Interface, 
 
 // createV1Alpha1ConfigValidationFunc creates a validation function for v1alpha1 LLMInferenceServiceConfig.
 // It converts the config to v1alpha2 and validates using the v1alpha2 llmisvc package.
-func createV1Alpha1ConfigValidationFunc(clientSet kubernetes.Interface) func(ctx context.Context, config *v1alpha1.LLMInferenceServiceConfig) error {
+func createV1Alpha1ConfigValidationFunc(reader client.Reader) func(ctx context.Context, config *v1alpha1.LLMInferenceServiceConfig) error {
 	return func(ctx context.Context, config *v1alpha1.LLMInferenceServiceConfig) error {
 		v2Config := &v1alpha2.LLMInferenceServiceConfig{}
 		if err := config.ConvertTo(v2Config); err != nil {
 			return err
 		}
-		return validateLLMISVCConfig(ctx, clientSet, v2Config)
+		return validateLLMISVCConfig(ctx, reader, v2Config)
 	}
 }
 
 // createV1Alpha2ConfigValidationFunc creates a validation function for v1alpha2 LLMInferenceServiceConfig.
-func createV1Alpha2ConfigValidationFunc(clientSet kubernetes.Interface) func(ctx context.Context, config *v1alpha2.LLMInferenceServiceConfig) error {
+func createV1Alpha2ConfigValidationFunc(reader client.Reader) func(ctx context.Context, config *v1alpha2.LLMInferenceServiceConfig) error {
 	return func(ctx context.Context, config *v1alpha2.LLMInferenceServiceConfig) error {
-		return validateLLMISVCConfig(ctx, clientSet, config)
+		return validateLLMISVCConfig(ctx, reader, config)
 	}
 }
--- a/pkg/controller/v1alpha2/llmisvc/config_loader.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader.go
@@ -23,7 +23,8 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/kubernetes"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
 	"github.com/kserve/kserve/pkg/constants"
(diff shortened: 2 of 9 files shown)


pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -685,19 +695,18 @@ func (r *LLMISVCReconciler) enqueueOnConfigMapChange(logger logr.Logger) handler
 		sub := object.(*corev1.ConfigMap)
 		reqs := make([]reconcile.Request, 0)
 
+		if isInferenceServiceConfigMap(sub) {
```
vivekk16: This is a redundant guard. `nonInferenceServiceConfigMapPredicate` already prevents `inferenceservice-config` from reaching this handler. Any specific reason for having it additional check once again?

cmd/llmisvc/main.go
```diff
@@ -197,7 +199,14 @@ func main() {
 					Label: llmSvcCacheSelector,
 				},
 				&corev1.ConfigMap{}: {
-					Label: llmSvcCacheSelector,
+					Namespaces: map[string]cache.Config{
```
vivekk16, line 202: Quick note on the `Namespaces` map here. Controller-runtime won't merge the `AllNamespaces` config into namespace specific entries, so in `KServeNamespace` only `inferenceservice-config` ends up being cached and watched. Labeled ConfigMaps there won't generate events anymore.

Not a problem today since no labeled ConfigMaps exist in that namespace and reads still fall back to the API server, but it's easy to miss. Would be nice to have a comment on the `KServeNamespace` entry so the next person doesn't have to figure this out the hard way.

pkg/controller/v1alpha2/llmisvc/config_loader_test.go
```diff
@@ -89,3 +93,20 @@ func TestNewSchedulerConfig(t *testing.T) {
...
+		WithScheme(clientgoscheme.Scheme).
+		WithObjects(cachedConfigMap).
+		Build()
+
+	got, err := llmisvc.LoadConfigFromClient(t.Context(), c)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if got.UrlScheme != "https" {
```
vivekk16, line 109: nit: This only checks `UrlScheme`. Might be worth asserting a couple more fields (like `IngressGatewayName` or `StorageConfig != nil`) since `LoadConfigFromClient` is the new primary read path and the fixture ConfigMap populates all those keys anyway.

vivekk16:
@KaveeshKhattar Nice improvements. The InferenceService controller has the same pattern where `GetInferenceServiceConfigMap` hits the API server directly on every reconcile (and actually multiple times per reconcile across the component reconcilers). It also has no watch on `inferenceservice-config`, so config changes don't trigger reconciliation of existing InferenceServices. Would be great to see a follow up PR applying the same caching and watch approach there.

bartoszmajsak:
Duplicate of #5094?

KaveeshKhattar:
> @KaveeshKhattar Nice improvements. The InferenceService controller has the same pattern where `GetInferenceServiceConfigMap` hits the API server directly on every reconcile (and actually multiple times per reconcile across the component reconcilers). It also has no watch on `inferenceservice-config`, so config changes don't trigger reconciliation of existing InferenceServices. Would be great to see a follow up PR applying the same caching and watch approach there.

@vivekk16 thanks for the review! I have addressed your feedback.

KaveeshKhattar:
There's overlap in intent, but the scope differs, the goal here was to address the existing TODOs in llmisvc with a controller-scoped change and minimal architectural surface area. 
Could close in favor of #5094 if it's preferred to consolidate.

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -186,6 +188,41 @@ func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, e
 		return nil, fmt.Errorf("failed to load InferenceServiceConfigMap: %w", errCfgMap)
 	}
 
+	return configFromConfigMap(isvcConfigMap)
+}
+
+// LoadConfigFromClient loads configuration from the cached controller-runtime client.
+func LoadConfigFromClient(ctx context.Context, c client.Client) (*Config, error) {
```
bartoszmajsak: Do we still need `LoadConfig`? Maybe we can drop it in favor of this new func and rename it?

pkg/controller/v1alpha2/llmisvc/controller_handlers_test.go
```diff
@@ -309,6 +311,51 @@ func TestRequestsForInferencePoolChangeIncludesRefsEvenWhenPoolIsOwned(t *testin
 	}))
 }
 
+func TestRequestsForInferenceServiceConfigMapChangeEnqueuesAllServices(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	scheme := runtime.NewScheme()
+	g.Expect(v1alpha2.AddToScheme(scheme)).To(Succeed())
+	g.Expect(corev1.AddToScheme(scheme)).To(Succeed())
```
bartoszmajsak: As we are re-using the scheme in couple of tests maybe we can extract it for consistency?

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -734,6 +739,39 @@ func (r *LLMISVCReconciler) enqueueOnConfigMapChange(logger logr.Logger) handler
...
 
+func (r *LLMISVCReconciler) enqueueOnInferenceServiceConfigMapChange(logger logr.Logger) handler.EventHandler {
+	logger = logger.WithName("enqueueOnInferenceServiceConfigMapChange")
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, object client.Object) []reconcile.Request {
+		reqs := make([]reconcile.Request, 0)
+		if !isInferenceServiceConfigMap(object) {
+			return reqs
+		}
+
+		continueToken := ""
```
bartoszmajsak: `continueToken` is not necessary as List is against the cache - there's no pagination involved. I just opened PR with a clean-up - #5584 - so let's align it.

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -186,6 +188,41 @@ func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, e
...
+		&corev1.ConfigMap{},
+		WithGetFallbackAPIServerConfigMap(r.Clientset),
+	)
+	if fallbackErr != nil {
+		return nil, fmt.Errorf("failed to load InferenceServiceConfigMap: %w", fallbackErr)
+	}
+	return configFromConfigMap(configMap)
+}
+
+func configFromConfigMap(isvcConfigMap *corev1.ConfigMap) (*Config, error) {
```
bartoszmajsak: maybe `toConfig`?

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -186,6 +188,41 @@ func LoadConfig(ctx context.Context, clientset kubernetes.Interface) (*Config, e
...
+		Name:      constants.InferenceServiceConfigMapName,
+	}, isvcConfigMap); err != nil {
+		return nil, fmt.Errorf("failed to load InferenceServiceConfigMap: %w", err)
+	}
+
+	return configFromConfigMap(isvcConfigMap)
+}
+
+func (r *LLMISVCReconciler) loadConfig(ctx context.Context) (*Config, error) {
+	if r.Clientset == nil {
```
bartoszmajsak: Is this check necessary? We would fail while combining configs in `config_merge.go` if that's ever `nil`.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -329,7 +338,8 @@ func (r *LLMISVCReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Secret{}, builder.WithPredicates(childResourcesPredicate)).
 		Owns(&corev1.Service{}, builder.WithPredicates(childResourcesPredicate)).
 		Owns(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(childResourcesPredicate)).
-		Watches(&corev1.ConfigMap{}, r.enqueueOnConfigMapChange(logger)).
+		Watches(&corev1.ConfigMap{}, r.enqueueOnConfigMapChange(logger), builder.WithPredicates(nonInferenceServiceConfigMapPredicate)).
+		Watches(&corev1.ConfigMap{}, r.enqueueOnInferenceServiceConfigMapChange(logger), builder.WithPredicates(inferenceServiceConfigMapPredicate)).
```
bartoszmajsak, line 375: I wonder if we could save unnecessary reconciliation by scoping diff to `llmisvc.Config` struct. Unrelated changes in `.data` for things outside of the boundaries of `LLMInferenceService` would get ignored. It also will make it more focused and ignore e.g. metadata updates which we shouldn't be concerned about.

pkg/controller/v1alpha2/llmisvc/controller_handlers_test.go
```diff
@@ -309,6 +311,51 @@ func TestRequestsForInferencePoolChangeIncludesRefsEvenWhenPoolIsOwned(t *testin
 	}))
 }
 
+func TestRequestsForInferenceServiceConfigMapChangeEnqueuesAllServices(t *testing.T) {
```
bartoszmajsak, line 310: I wonder if it wouldn't be more helpful to cover this test case through `envtest` tests to validate the real reconcilition flow and ensure the changes from config are reflected?

vivekk16:
cc @bartoszmajsak @spolti @VedantMahabaleshwarkar

bartoszmajsak (commented):
Thanks for this contribution. I shared some suggestions for minor improvements. 

Overall I like that it's scoped, so lets ship it first and then #5094 can build on top.

FWIW I think you need to run `make precommit` to tidy up the code.

pkg/controller/v1alpha2/llmisvc/config_loader.go
KaveeshKhattar: `LoadConfig` is still needed for the webhook validation path, both in:

1. `cmd/llmisvc/main.go` (`validateLLMISVCConfig`)
2. `fixture/envtest.go`

where it's called with a raw `kubernetes.Interface` before the controller-runtime cache is ready. 

`LoadConfigFromClient` requires a cached `client.Client` so it can't substitute there. 
Will update `LoadConfig`'s godoc to make this distinction explicit.

pkg/controller/v1alpha2/llmisvc/controller_handlers_test.go
KaveeshKhattar: The unit test verifies that the watch handler correctly enqueues all services, that's the core behavior this PR introduces. A full envtest covering the config-to-reconcile flow would be valuable but feels broader than this focused change. Happy to open a follow-up issue to track it if you'd like.

pkg/controller/v1alpha2/llmisvc/config_loader.go
bartoszmajsak: FWIW `mgr.GetAPIReader()` is available at webhook setup time, so maybe we can unify both functions around `client.Reader`:

```golang
  func LoadConfig(ctx context.Context, r client.Reader) (*Config, error) {
      isvcConfigMap := &corev1.ConfigMap{}
      if err := r.Get(ctx, types.NamespacedName{
          Namespace: constants.KServeNamespace,
          Name:      constants.InferenceServiceConfigMapName,
      }, isvcConfigMap); err != nil {
          return nil, fmt.Errorf("failed to load InferenceServiceConfigMap: %w", err)
      }
      return configFromConfigMap(isvcConfigMap)
  }
```

pkg/controller/v1alpha2/llmisvc/controller_handlers_test.go
bartoszmajsak: I'm on the fence about this being broader :) It's the primary thing this PR introduces - caching + watch - so having near-e2e proof that the full loop works is valuable. The envtest scaffolding already exists, so maybe a single test case that:

1. Creates an LLMInferenceService
2. Updates inferenceservice-config (e.g. changes urlScheme)
3. Asserts the reconciled state reflects the new config

would be a good start? A handler unit test with a fake client doesn't exercise the informer cache, the predicate filtering, or the actual reconcile path - this would.

KaveeshKhattar: Sure, I've added a test case `"should reconcile existing services when inferenceservice-config changes"` in `controller_model_based_routing_int_test.go`

pkg/controller/v1alpha2/llmisvc/config_loader.go
KaveeshKhattar: Unified around `client.Reader`. `mgr.GetAPIReader()` is passed in `envtest.go`, and `main.go` passes it through the webhook closure where `mgr` is in scope. `LoadConfigFromClient` is dropped.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -76,6 +76,47 @@ var ChildResourcesLabelSelector = metav1.LabelSelector{
...
+	newConfigMap, newOK := newObj.(*corev1.ConfigMap)
+	if !oldOK || !newOK {
+		return true
+	}
+
+	oldConfig, oldErr := toConfig(oldConfigMap)
+	newConfig, newErr := toConfig(newConfigMap)
+	if oldErr != nil || newErr != nil {
+		return !equality.Semantic.DeepEqual(oldConfigMap.Data, newConfigMap.Data)
+	}
```
bartoszmajsak, line 117: That's reasonable take, however I think we should also surface the error(s) that we have malformed configmap. Maybe sth like this is good enough?


```suggestion
	if oldErr != nil || newErr != nil {
		ctrl.Log.WithName("LLMInferenceService").Error(
				errors.Join(oldErr, newErr),
				"Failed to parse inferenceservice-config, falling back to raw comparison",
		)
		return !equality.Semantic.DeepEqual(oldConfigMap.Data, newConfigMap.Data)
	}
```

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -76,6 +76,47 @@ var ChildResourcesLabelSelector = metav1.LabelSelector{
...
+	return obj.GetNamespace() == constants.KServeNamespace &&
+		obj.GetName() == constants.InferenceServiceConfigMapName
+}
+
+func inferenceServiceConfigChanged(oldObj, newObj client.Object) bool {
+	oldConfigMap, oldOK := oldObj.(*corev1.ConfigMap)
+	newConfigMap, newOK := newObj.(*corev1.ConfigMap)
+	if !oldOK || !newOK {
+		return true
+	}
```
bartoszmajsak: We probably can just ignore the ok check - that's consistent with the rest of the predicates (and it's specifically set to be always for the Configmap watch). 

```suggestion
	oldConfigMap := oldObj.(*corev1.ConfigMap)
	newConfigMap := newObj.(*corev1.ConfigMap)
```

Or we can use `predicate.TypedFuncs[*corev1.ConfigMap]` instead.

bartoszmajsak (commented):
Two minor things for the simplified solution. Thanks for taking my feedback into consideration.

pkg/controller/v1alpha2/llmisvc/controller.go
KaveeshKhattar: Makes sense, removed the ok check, consistent with the rest of the predicates.

pkg/controller/v1alpha2/llmisvc/controller.go
KaveeshKhattar: Sure, added the error log before falling back to raw comparison.

bartoszmajsak (approved):

KaveeshKhattar:
@Jooho @spolti could you please take a look?

PR #72, opened 2019-05-07
Consolidate to make fwks pluggable

rakelkar:
Consolidate all fwk related code under the framework package so that frameworks are easy to add and maintain. Comments on approach solicited!

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kfserving/72)

PR #5574, opened 2026-05-23
fix(ci): harden E2E test infrastructure to reduce flakiness

vivekk16:
## Summary
- Fix race condition in unit test where async K8s event recorder caused flaky baseline count
- Make image verification in LLMISvc workflow blocking (exit 1 on missing images)
- Fix bash iteration pattern and /tmp-images path typo in ModelCache workflow
- Load runtime images on all 3 minikube nodes for ModelCache tests
- Add kubectl wait pod readiness gate after KServe install with field-selector to exclude Completed pods
- Wait for old controller pod to terminate after KEDA rollout restart before readiness gate
- Wait for cert-manager webhook and API service availability before proceeding with KServe install
- Add deployment status, pod describe, and event diagnostics to wait_for_deployment on failure
- Increase LLMISvc E2E test step timeout from 60 to 90 minutes
- Add --field-selector=status.phase=Running to AWS kubectl wait calls

### DIFF
--- a/.github/workflows/e2e-test-llmisvc.yaml
+++ b/.github/workflows/e2e-test-llmisvc.yaml
@@ -173,21 +173,15 @@ jobs:
 
       - name: Load LLMISvc controller image into minikube
         run: |
-          # Load the tagged image into minikube
           minikube image load ./tmp/${{ env.LLMISVC_CONTROLLER_IMG }}-${{ env.TAG }}
-
-          # Verify image is loaded
-          echo "Images in minikube:"
-          minikube image ls | grep ${{ env.LLMISVC_CONTROLLER_IMG }} || echo "No llmisvc-controller images found"
+          echo "Verifying llmisvc-controller image is loaded:"
+          minikube image ls | grep ${{ env.LLMISVC_CONTROLLER_IMG }} || { echo "FATAL: llmisvc-controller image not found in minikube"; exit 1; }
 
       - name: Load storage-initializer image into minikube
         run: |
-          # Load the tagged image into minikube
           minikube image load ./tmp/${{ env.STORAGE_INIT_IMG }}-${{ env.TAG }}
-
-          # Verify image is loaded
-          echo "Images in minikube:"
-          minikube image ls | grep ${{ env.STORAGE_INIT_IMG }} || echo "No storage-initializer images found"
+          echo "Verifying storage-initializer image is loaded:"
+          minikube image ls | grep ${{ env.STORAGE_INIT_IMG }} || { echo "FATAL: storage-initializer image not found in minikube"; exit 1; }
 
       - name: KServe dependency setup
         uses: ./.github/actions/kserve-dep-setup
@@ -297,14 +291,14 @@ jobs:
       - name: Load LLMISvc controller image into minikube
         run: |
           minikube image load ./tmp/${{ env.LLMISVC_CONTROLLER_IMG }}-${{ env.TAG }}
-          echo "Images in minikube:"
-          minikube image ls | grep ${{ env.LLMISVC_CONTROLLER_IMG }} || echo "No llmisvc-controller images found"
+          echo "Verifying llmisvc-controller image is loaded:"
+          minikube image ls | grep ${{ env.LLMISVC_CONTROLLER_IMG }} || { echo "FATAL: llmisvc-controller image not found in minikube"; exit 1; }
 
       - name: Load storage-initializer image into minikube
         run: |
           minikube image load ./tmp/${{ env.STORAGE_INIT_IMG }}-${{ env.TAG }}
-          echo "Images in minikube:"
-          minikube image ls | grep ${{ env.STORAGE_INIT_IMG }} || echo "No storage-initializer images found"
+          echo "Verifying storage-initializer image is loaded:"
+          minikube image ls | grep ${{ env.STORAGE_INIT_IMG }} || { echo "FATAL: storage-initializer image not found in minikube"; exit 1; }
 
       - name: KServe dependency setup (HPA autoscaling)
         uses: ./.github/actions/kserve-dep-setup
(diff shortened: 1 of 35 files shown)


.github/workflows/e2e-test-modelcache.yaml
```diff
@@ -345,13 +344,16 @@ jobs:
...
+            echo "Loading image $(basename "$file")"
+            minikube ssh -n minikube -- docker image load -i "$file"
+            minikube ssh -n minikube-m02 -- docker image load -i "$file"
+            minikube ssh -n minikube-m03 -- docker image load -i "$file"
           done
-          sudo rm -rf tmp-images/*
+          sudo rm -rf /tmp-images/*
+          minikube ssh -n minikube -- docker image ls
           minikube ssh -n minikube-m02 -- docker image ls
+          minikube ssh -n minikube-m03 -- docker image ls
```
spolti: don't we need to add some filter here? like, if the images are not there, exit.

vivekk16: Sure! Added a pre-load check that fails early if no image files are found in /tmp-images, and a post-load verification that runs docker image inspect for each expected image on all three minikube nodes.

spolti:
@Jooho @sivanantha321 may you guys take a look as well?

hack/setup/infra/manage.cert-manager-helm.sh
```diff
@@ -71,6 +71,13 @@ install() {
 
     wait_for_pods "cert-manager" "app in (cert-manager,webhook,cainjector)" "180s"
 
+    log_info "Waiting for cert-manager webhook to be ready..."
+    kubectl wait --for=condition=Available deployment/cert-manager-webhook -n cert-manager --timeout=60s || {
+        log_error "cert-manager webhook deployment not available"
+        return 1
+    }
```
Jooho: It is duplicated code with line #72

hack/setup/infra/manage.cert-manager-helm.sh
```diff
@@ -71,6 +71,13 @@ install() {
...
+    kubectl wait --for=condition=Available apiservice v1.cert-manager.io --timeout=60s 2>/dev/null || true
```
Jooho: Isn’t this a bit too defensive? If we add this for cert-manager, shouldn’t we do the same for the others too?

Also, do we have any flaky test or CI failure links related to this?

.github/workflows/e2e-test-llmisvc.yaml
Jooho, line 1: First of all, I like the fast-fail way :) 

I have two thoughts on this:

1. Why doesn’t LLMISVC use the `docker load files` like the main/modelcache e2e tests do? If we used the same approach, we could probably enhance/reuse the existing action(or new action) across all cases.
3. Why don’t we add retries when loading images before exiting?

.github/workflows/e2e-test-modelcache.yaml
```diff
@@ -319,17 +319,36 @@ jobs:
...
-            minikube ssh -n minikube-m02 -- docker image load -i ${file}
-            minikube ssh -n minikube-m03 -- docker image load -i ${file}
+          file_count=$(find /tmp-images -maxdepth 1 -type f | wc -l)
+          if [[ "$file_count" -eq 0 ]]; then
+            echo "::error::No image files found in /tmp-images"
+            exit 1
+          fi
+          for file in $(find /tmp-images -maxdepth 1 -type f); do
+            echo "Loading image $(basename "$file")"
+            minikube ssh -n minikube -- docker image load -i "$file"
```
Jooho, line 329: Fyi,this is also using `docker image load`  so only llmisvc is using minikube load.

yea it is not in this PR scope, but I think making it standard would be great later then all test cases will use the same ci env.

Jooho:
reviewing

Jooho (commented):
There are no major blockers.

There is some duplication in cert-manager script, but I don’t think it’s necessary.

It would also be good if we could standardize the e2e test CI environment. LLMISVC and other e2e tests already follow the same pattern for building and loading images, so there’s no real need to use different approaches across tests.

hack/setup/infra/manage.cert-manager-helm.sh
vivekk16: Good catch! Removed it.

.github/workflows/e2e-test-llmisvc.yaml
vivekk16: Agreed, there was no good reason for the divergence (pattern was there since the LLMISvc workflow was first added). LLMISvc uses minikube with driver none (same as the main e2e tests), so docker image load via the shared load-docker-images action works the same way. Switched all 3 LLMISvc jobs to use the action now.

On retries: docker image load will fail the step with a non-zero exit code if the tar file is missing or corrupt, which is a clear signal. Adding retries is a good idea for transient Docker daemon errors though, I can add that as a follow-up enhancement to the load-docker-images action so all workflows benefit from it.

.github/workflows/e2e-test-modelcache.yaml
vivekk16: ModelCache is a different case, it uses minikube with driver docker and 3 nodes, so it needs `minikube ssh -- docker image load` per node with a mounted path. The load-docker-images action won't work there since it loads into the host Docker daemon which is separate from minikube's when using the docker driver.

hack/setup/infra/manage.cert-manager-helm.sh
vivekk16: We don't have a specific CI failure link for this. It's non-blocking (|| true) and checks API server registration rather than pod readiness, so I kept it as a low-risk safety net. 

Removed it.

andresllh (commented):
https://github.com/kserve/kserve/pull/5568 Just merged. Can you rebase master onto this branch and retest?

Jooho (approved):
/approve

PR #5575, opened 2026-05-23
fix(sdk): add client_configuration parameter in KServeClient constructor

### LINKED ISSUE #5464: sdk: `KServeClient` ignores the `client_configuration` parameter
/kind bug

When constructing `KServeClient` from outside the cluster and passing a `kubernetes.client.Configuration` object via the `client_configuration` constructor parameter, the supplied configuration has no effect. The client ignores the provided host and credentials and falls back to `~/.kube/config` instead.

`KServeClient` should use the host and credentials from the supplied configuration object for all management operations against the Kubernetes API server.

N/A — the `InferenceService` is never reached; the client fails to authenticate before any resource is created.

The `config_dict` constructor parameter works correctly as a workaround: wrapping the credentials in a minimal kubeconfig dict and passing it via `config_dict` results in the expected behaviour. The `client_configuration` parameter appears to be broken in the current implementation.

**Environment:**

- Istio Version: 1.24
- Knative Version: N/A, using Standard mode
- KServe Version: v0.16.0
- Kubeflow version: 1.11
- Cloud Environment: istio_dex
- Minikube/Kind version: KinD v0.27.0
- Kubernetes version: v1.32
- OS: Ubuntu 24.04

vishwagandhi1610:
`KServeClient` accepts a `client_configuration` parameter intended to let callers pass a pre-built `kubernetes.client.Configuration` object (with a custom host, token, or SSL settings) instead of loading a kubeconfig file.

However, when `client_configuration` was passed alone (without `config_file` or `config_dict`), it had no effect. The constructor fell through to `config.load_kube_config()`, which overwrote the provided configuration object with values from `~/.kube/config`. The resulting API clients (`core_api`, `app_api`, etc.) were then created with no explicit configuration, picking up the global default — not the caller's config.

This fix adds a dedicated branch: when `client_configuration` is provided without `config_file` or `config_dict`, the configuration is wired directly into a shared `ApiClient` which is passed to each API client. This matches the pattern recommended by the Kubernetes Python SDK: https://github.com/kubernetes-client/python/blob/master/examples/remote_cluster.py

Fixes #5464

Added `python/kserve/test/test_client_configuration.py` following the same pattern as `test_kubeconfig_dict.py`:

- [x] `test_client_configuration_is_wired_to_all_api_clients` — verifies the provided `host` is reflected in all four API clients (`core_api`, `app_api`, `api_instance`, `hpa_v2_api`) without reading any kubeconfig
- [x] `test_inferenceservice_client_create/get/patch/replace/delete` — verifies the full CRUD surface works correctly when initialized via `client_configuration`

Run tests:
```bash
cd python /path/to/.venv/bin/pytest -W ignore kserve/test/test_client_configuration.py -v
```

**Checklist**:

**Release note**:
```release-note
Fix KServeClient ignoring the client_configuration parameter when config_file and config_dict are not provided.
```

### DIFF
--- a/python/kserve/kserve/api/kserve_client.py
+++ b/python/kserve/kserve/api/kserve_client.py
@@ -48,30 +48,35 @@ class KServeClient(object):
         :param client_configuration: kubernetes configuration object
         :param persist_config:
         """
-        if client_configuration and not config_file and not config_dict:
+        if config_dict:
+            config.load_kube_config_from_dict(
+                config_dict=config_dict,
+                context=context,
+                client_configuration=None,
+                persist_config=persist_config,
+            )
+        elif config_file:
+            config.load_kube_config(
+                config_file=config_file,
+                context=context,
+                client_configuration=client_configuration,
+                persist_config=persist_config,
+            )
+        elif client_configuration:
             api_client = client.ApiClient(configuration=client_configuration)
             self.core_api = client.CoreV1Api(api_client=api_client)
             self.app_api = client.AppsV1Api(api_client=api_client)
             self.api_instance = client.CustomObjectsApi(api_client=api_client)
             self.hpa_v2_api = client.AutoscalingV2Api(api_client=api_client)
             return
-        if config_file or config_dict or not utils.is_running_in_k8s():
-            if config_dict:
-                config.load_kube_config_from_dict(
-                    config_dict=config_dict,
-                    context=context,
-                    client_configuration=None,
-                    persist_config=persist_config,
-                )
-            else:
-                config.load_kube_config(
-                    config_file=config_file,
-                    context=context,
-                    client_configuration=client_configuration,
-                    persist_config=persist_config,
-                )
-        else:
+        elif utils.is_running_in_k8s():
             config.load_incluster_config()
+        else:
+            config.load_kube_config(
+                context=context,
+                client_configuration=client_configuration,
+                persist_config=persist_config,
+            )
         self.core_api = client.CoreV1Api()
         self.app_api = client.AppsV1Api()
         self.api_instance = client.CustomObjectsApi()
--- a/python/kserve/kserve/api/kserve_client.py
+++ b/python/kserve/kserve/api/kserve_client.py
@@ -48,6 +48,13 @@ class KServeClient(object):
         :param client_configuration: kubernetes configuration object
         :param persist_config:
         """
+        if client_configuration and not config_file and not config_dict:
+            api_client = client.ApiClient(configuration=client_configuration)
+            self.core_api = client.CoreV1Api(api_client=api_client)
+            self.app_api = client.AppsV1Api(api_client=api_client)
+            self.api_instance = client.CustomObjectsApi(api_client=api_client)
+            self.hpa_v2_api = client.AutoscalingV2Api(api_client=api_client)
+            return
         if config_file or config_dict or not utils.is_running_in_k8s():
             if config_dict:
                 config.load_kube_config_from_dict(
--- a/python/kserve/test/test_client_configuration.py
+++ b/python/kserve/test/test_client_configuration.py
@@ -1,4 +1,4 @@
-# Copyright 2021 The KServe Authors.
+# Copyright 2026 The KServe Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


python/kserve/test/test_client_configuration.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def test_inferenceservice_client_delete():
+    """Unit test for kserve delete api with client_configuration"""
+    with patch(
+        "kserve.api.kserve_client.KServeClient.delete", return_value=mocked_unit_result
+    ):
+        assert mocked_unit_result == kserve_client.delete(
+            "flower-sample", namespace="kubeflow"
+        )
```
Jooho: The CRUD tests are mocking KServeClient.create/get/... at the class level, so the real implementation never runs. Since these tests don’t exercise the actual logic at all, I’m not sure they add much value. If this pattern was copied from existing tests for consistency, that makes sense, but from a coverage/regression perspective it feels a bit closer to noise than meaningful validation.

python/kserve/test/test_client_configuration.py
```diff
@@ -0,0 +1,123 @@
...
+    return isvc
+
+
+def test_client_configuration_is_wired_to_all_api_clients():
+    """client_configuration must be passed directly to every underlying API
+    client without reading ~/.kube/config or setting a global default."""
+    assert kserve_client.core_api.api_client.configuration.host == cfg.host
+    assert kserve_client.app_api.api_client.configuration.host == cfg.host
+    assert kserve_client.api_instance.api_client.configuration.host == cfg.host
+    assert kserve_client.hpa_v2_api.api_client.configuration.host == cfg.host
```
Jooho, line 32: Don't we need to check api_key?

python/kserve/test/test_client_configuration.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2021 The KServe Authors.
```
Jooho: 2026?

python/kserve/kserve/api/kserve_client.py
```diff
@@ -48,6 +48,13 @@ def __init__(
         :param client_configuration: kubernetes configuration object
         :param persist_config:
         """
+        if client_configuration and not config_file and not config_dict:
```
Jooho: The fix looks correct for the standalone client_configuration case and is consistent with the typical Kubernetes Python client pattern of creating an ApiClient with an explicit configuration.

One subtle point: this branch only applies when neither config_file nor config_dict is set, so client_configuration is not treated as a strict precedence option. For example, if both config_dict and client_configuration are provided, the config_dict path will take priority and the client_configuration path will be ignored.

That may be fine depending on the intended behavior, but the current structure makes the precedence rules a bit hard to infer at a glance. A clearer if/elif structure or an explicit note in the constructor docs would help make the intended priority more obvious.

python/kserve/test/test_client_configuration.py
vishwagandhi1610: resolved

python/kserve/test/test_client_configuration.py
vishwagandhi1610: I have added checks for api_key

python/kserve/kserve/api/kserve_client.py
vishwagandhi1610: Thanks for the feedback! I've refactored the constructor to use an explicit if/elif chain that makes the precedence order self-documenting:

`config_dict > config_file > client_configuration > in-cluster > default kubeconfig`

Regarding the config_dict + client_configuration case — worth noting that client_configuration is silently ignored when config_dict is provided, as it is explicitly passed as None to `load_kube_config_from_dict`. This matches the original behavior, but I wanted to flag it: is that intentional, or is it also a bug worth addressing? 

I've added `test_config_dict_takes_priority_over_client_configuration` to explicitly assert this behavior so it's no longer implicit. Similarly added `test_config_file_takes_priority_over_client_configuration` for the config_file case. Also removed the CRUD tests — they were initially kept for consistency with the other test files but you're right, they add noise without testing anything meaningful.

python/kserve/test/test_client_configuration.py
vishwagandhi1610: Removed the CRUD tests — they were initially kept for consistency with the other test files but you're right, they add noise without testing anything meaningful.

python/kserve/kserve/api/kserve_client.py
```diff
@@ -48,23 +48,35 @@ def __init__(
...
+            self.app_api = client.AppsV1Api(api_client=api_client)
+            self.api_instance = client.CustomObjectsApi(api_client=api_client)
+            self.hpa_v2_api = client.AutoscalingV2Api(api_client=api_client)
+            return
+        elif utils.is_running_in_k8s():
             config.load_incluster_config()
+        else:
+            config.load_kube_config(
+                context=context,
+                client_configuration=client_configuration,
```
spolti, line 77: Shouldn't we check if client_configuration is empty

vishwagandhi1610: It will be empty that is the reason "else" loop would be executed at the first place. Otherwise it would get into "elif client_configuration" loop before. Do we want to explicitly pass "None" to the client_configuration than?

PR #5578, opened 2026-05-25
feat: added ppc64le support to ARTExplainer.

Sunidhi-Gaonkar1:
**What this PR does / why we need it**: We have updated the existing Dockerfile and added a new workflow for Artexplainer to add Power support.

The generated image was validated on a local Power VM. A container was created using the built ART-explainer image and validated using basic curl command.

- Logs
```
docker run -it  localhost/art:test bash
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/prod_venv/lib/python3.11/site-packages/art/estimators/certification/__init__.py:30: UserWarning: PyTorch not found. Not importing DeepZ or Interval Bound Propagation functionality
  warnings.warn("PyTorch not found. Not importing DeepZ or Interval Bound Propagation functionality")
2026-05-25 09:06:26.954 1 kserve INFO [model_server.py:register_model():439] Registering model: model
2026-05-25 09:06:26.954 1 kserve INFO [model_server.py:setup_event_loop():315] Setting max asyncio worker threads as 20
2026-05-25 09:06:27.081 1 kserve INFO [server.py:_register_endpoints():124] OpenAI endpoints not registered
2026-05-25 09:06:27.082 1 kserve INFO [server.py:_register_endpoints():132] Time series endpoints not registered
2026-05-25 09:06:27.082 1 kserve INFO [server.py:start():183] Starting uvicorn with 1 workers
2026-05-25 09:06:27.157 1 uvicorn.error INFO:     Started server process [1]
2026-05-25 09:06:27.158 1 uvicorn.error INFO:     Waiting for application startup.
2026-05-25 09:06:27.163 1 kserve INFO [server.py:start():70] Starting gRPC server with 4 workers
2026-05-25 09:06:27.163 1 kserve INFO [server.py:start():71] Starting gRPC server on [::]:8081
2026-05-25 09:06:27.165 1 uvicorn.error INFO:     Application startup complete.
2026-05-25 09:06:27.166 1 uvicorn.error INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)

```

```
curl http://localhost:8080
{"status":"alive"}
```

- This PR only introduces Power (ppc64le) image build support for ARTExplainer.
- For the packages that do not have power support, we are fetching wheels from `https://wheels.developerfirst.ibm.com/ppc64le/linux`. We have added this index to `project.toml` file of kserve and artexplainer.
- The build time taken for Power is around 9 mins, which is comparable to linux/arm64/v8 when built separately.

**Release note**:
```release-note
Added ppc64le support to ART-explainer.
```


python/kserve/pyproject.toml
```diff
@@ -101,15 +103,53 @@ dev = [
...
+
 [build-system]
 requires = ["setuptools>=61.0"]
 build-backend = "setuptools.build_meta"
 
 # UV workspace source for local dependency resolution
 [tool.uv.sources]
 kserve-storage = { path = "../storage", editable = true }
+grpcio = [
+    { index = "pypi", marker = "platform_machine != 'ppc64le'" },
```
spolti: Why not keep these into the art explainer pyproject.toml only?

spolti (commented):
Is there a reason to have the ppc64 in the main pyproject.yaml?

python/kserve/pyproject.toml
Sunidhi-Gaonkar1: `grpcio` gets installed as a part of kserve dependency [here,](https://github.com/kserve/kserve/blob/596c65315de98a586b4724c7b17e61d88233eab4/python/artexplainer.Dockerfile#L26) but PyPI doesn’t provide wheels for ppc64le. As a result, it builds from source, which takes around an hour. To avoid this, we’ve added `grpcio` in `python/kserve/pyproject.toml` using the IBM developerfirst index, which provides prebuilt wheels.

Sunidhi-Gaonkar1:
Hi @spolti ,
while running uv commands for ARTExplainer few packages are getting built from source as their wheels are not available on pypi for ppc64le, we have updated python/kserve/pyproject.toml to use the IBM Power wheel index (https://wheels.developerfirst.ibm.com/ppc64le/linux) and generated a single uv.lock file.

The current approach of modifying the main pyproject.toml to use IBM DevPI is breaking non-PPC builds. Should we instead maintain separate files per architecture?
- The first pyproject.toml will fetch wheels from pypi for amd64 and arm64. 
- second pyproject-ppc64le.toml will fetch wheels from IBM power wheel index.
- Dockerfiles would conditionally copy the right files based on TARGETARCH. This keeps standard builds unchanged while enabling PPC-optimized wheels.


Additional questions:

1. What's the recommended process for updating pyproject.toml and regenerating uv.lock files in this project?
2. Should contributors manually run uv lock after dependency changes, or is this handled automatically?
3. For the PPC-specific files, should we commit both pyproject.toml.ppc64le and uv.lock.ppc64le to the repository?
4. Please let us know if this approach acceptable? and Should we apply it to all Python components?

python/kserve/pyproject.toml
spolti: by adding this directly into the artexplainer one would have the same output, I believe, no?

I would like to not have dozens of lines added to the main module to build just one container.
Ideally, it would be best to live only under the ppc64 builds.

spolti:
@yuzisun @terrytangyuan  thoughts?

terrytangyuan:
I am not sure if it makes sense to provide OS specific additional files at the project root level.

sandeepgupta12:
Hi @spolti  @Jooho, 
We updated this PR based on the suggestion to handle the pyproject.toml changes from python/artexplainer.Dockerfile by adding the IBM package index there for dependency resolution. The branch has now been cleaned up and the PR contains only the intended changes. Could you please approve the pending workflows so they can run, and review the PR when you have a chance?

sandeepgupta12:
I'll take a look at the failing jobs.

sandeepgupta12:
@Jooho @spolti @cjohannsen-cloudera , please approve the workflows to run.
Also please review this PR.

python/artexplainer.Dockerfile
```diff
@@ -23,16 +25,98 @@ COPY storage/pyproject.toml storage/uv.lock storage/
 
 # ------------------ kserve deps ------------------
 COPY kserve/pyproject.toml kserve/uv.lock kserve/
+
+# On ppc64le: patch pyproject.toml to add the ppc64le package index and sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
+            -e '/^index-strategy\s*=.*/a \\' \
+            -e '/^index-strategy\s*=.*/a [[tool.uv.index]]' \
```
cjohannsen-cloudera, line 34: **Correctness (high):** This sed insertion lands `[[tool.uv.index]]` right after `index-strategy = ...`, which closes the `[tool.uv]` table context in TOML. The `override-dependencies = [...]` block that follows (CVE version overrides for urllib3, aiohttp, h11, cryptography, pyjwt, pyasn1, pillow, python-multipart) now parses as `tool.uv.index[0].override-dependencies` instead of `tool.uv.override-dependencies`. `uv` silently ignores the unknown field on the index struct — `uv lock` succeeds with no error, but the ppc64le image resolves those CVE-impacted packages at uncontrolled versions (e.g. urllib3 < 2.7.0, CVE-2026-21441).

python/artexplainer.Dockerfile
```diff
@@ -23,16 +25,98 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            -e '/^kserve-storage\s*=.*/a grpcio-tools = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a numpy = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pandas = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a psutil = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pyyaml = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a httptools = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a uvloop = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pillow = { index = "ppc64le-wheels" }' \
+            kserve/pyproject.toml && \
+        cd kserve && uv lock && \
```
cjohannsen-cloudera, line 51: **Reliability:** `uv lock` runs live here instead of using a committed lockfile, so ppc64le builds are non-reproducible — two builds a week apart against the same commit could resolve different dependency versions if anything changes on PyPI or `wheels.developerfirst.ibm.com`. x86/arm64 builds don't have this problem since they run `uv sync --no-cache` against the checked-in `uv.lock`. Also means if the IBM wheel index is unreachable, the build blocks indefinitely with no fallback. Same pattern repeats below for `artexplainer` (line ~103).

python/artexplainer.Dockerfile
```diff
@@ -23,16 +25,98 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+# On ppc64le: patch pyproject.toml to add the ppc64le package index and sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
+            -e '/^index-strategy\s*=.*/a \\' \
+            -e '/^index-strategy\s*=.*/a [[tool.uv.index]]' \
+            -e '/^index-strategy\s*=.*/a name = "ppc64le-wheels"' \
+            -e '/^index-strategy\s*=.*/a url = "https://wheels.developerfirst.ibm.com/ppc64le/linux"' \
+            -e '/^index-strategy\s*=.*/a explicit = true' \
+            -e '/^\s*"pyasn1>=[^,]*"$/s/"$/",/' \
```
cjohannsen-cloudera, line 38: **Correctness (latent):** This sed address `/^\s*"pyasn1>=[^,]*"$/` never matches the current file, since the actual line ends with `", # CVE-2026-30922` rather than a bare quote — so this trailing-comma guard is permanently a no-op. Harmless today, but if `pyasn1` ever becomes the last element of `override-dependencies` with no trailing comma/comment, the two `a` commands right below will inject `httptools`/`uvloop` after a string with no separating comma — invalid TOML, breaking `uv lock` only on ppc64le.

python/artexplainer.Dockerfile
```diff
@@ -23,16 +25,98 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
 RUN cd kserve && uv sync --active --no-cache
 
 # ------------------ artexplainer deps ------------------
 COPY artexplainer/pyproject.toml artexplainer/uv.lock artexplainer/
+
+# On ppc64le: patch pyproject.toml to add extra direct dependencies and the ppc64le package index/sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
+            -e '/^    "h5py/a\    "scikit-learn==1.6.1",' \
```
cjohannsen-cloudera, line 77: **Maintainability:** `scikit-learn==1.6.1`, `scipy==1.15.2`, and `ml-dtypes==0.5.1` are hardcoded here instead of being tracked in `kserve-deps.env` (the project's canonical version registry). If `adversarial-robustness-toolbox` raises its minimum scikit-learn requirement, `uv lock` on ppc64le will fail here while x86/arm64 CI (using the committed lockfile) keeps passing — a silent, hard-to-trace breakage down the line.

python/artexplainer.Dockerfile
```diff
@@ -23,16 +25,98 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            -e '/^    "h5py/a\    "scipy==1.15.2",' \
+            -e '/^    "h5py/a\    "ml-dtypes==0.5.1",' \
+            artexplainer/pyproject.toml && \
+        printf '%s\n' \
+            '' \
+            '[[tool.uv.index]]' \
```
cjohannsen-cloudera, line 83: **Correctness (latent):** This `printf >> artexplainer/pyproject.toml` appends `[[tool.uv.index]]`/`[tool.uv.sources]` with nothing guarding against duplicate section headers. If a follow-up moves these sections into `artexplainer/pyproject.toml` directly (matching the pattern other components use) without also removing this printf, the next ppc64le build gets duplicate TOML headers — invalid syntax, breaks `uv lock` only on ppc64le.

python/artexplainer.Dockerfile
```diff
@@ -5,7 +5,9 @@ ARG VENV_PATH=/prod_venv
 FROM ${BASE_IMAGE} AS builder
 
 # Required for building packages for arm64 arch
-RUN apt-get update && apt-get install -y --no-install-recommends curl python3-dev build-essential && apt-get clean && \
+RUN apt-get update && apt-get install -y --no-install-recommends curl python3-dev build-essential && \
+    if [ "$(uname -m)" = "ppc64le" ]; then apt-get install pkg-config libssl-dev gcc gfortran cmake pkg-config libssl-dev libopenblas-dev libjpeg-dev libhdf5-dev wget -y; fi && \
```
cjohannsen-cloudera, line 9: **Efficiency / nit:** Two small things here: (1) missing `--no-install-recommends`, unlike the main install two lines up — pulls in a lot of unneeded transitive packages on the already-slow QEMU build path; (2) `pkg-config` and `libssl-dev` are each listed twice, likely from copy-paste — `apt-get` dedupes silently so it's harmless, but worth cleaning up.

cjohannsen-cloudera (commented):
Automated review pass on the ppc64le/ART Explainer build changes. One finding (production stage missing runtime shared libs) isn't tied to a line this PR touches, so it's called out in this summary rather than inline: the production stage never installs runtime libs like `libopenblas.so.0`/`libgfortran.so.5` that the ppc64le-compiled numpy/scipy/pillow/h5py extensions link against at runtime (they're compiled from source in the builder stage using `libopenblas-dev`/`gfortran`/`libhdf5-dev`/`libjpeg-dev`, none of which carry over). Worth confirming `artserver` actually starts on a ppc64le image before merging.

.github/workflows/artexplainer-ppc64le-docker-publish.yml
```diff
@@ -0,0 +1,138 @@
...
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+    paths:
+      - "python/artexplainer/**"
```
cjohannsen-cloudera, line 16: **Correctness:** This `pull_request.paths` filter doesn't include `python/kserve/**`, but the Dockerfile explicitly `COPY`s and installs `kserve`. A PR that only touches kserve source (e.g. `python/kserve/kserve/protocol/...`) won't trigger this ppc64le workflow, even though the amd64/arm64 `artexplainer` workflow (which does include `python/kserve/**`) will fire. A broken ppc64le image could ship silently on the next tag.

.github/workflows/artexplainer-ppc64le-docker-publish.yml
```diff
@@ -0,0 +1,138 @@
...
+          git fetch --unshallow origin
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+          git config user.email "ci@kserve.io"
+          git config user.name "CI Bot"
+          git merge --no-edit origin/${{ github.event.pull_request.base.ref }}
+
+      - name: Free-up disk space
+        uses: ./.github/actions/free-up-disk-space
+
+      - name: Setup QEMU
```
cjohannsen-cloudera, line 57: **Efficiency:** QEMU setup isn't scoped to `platforms: linux/ppc64le` here (or in the `push` job further down), so every push to master registers binfmt handlers for all emulated architectures unnecessarily — adds ~30-60s per run for no benefit. The existing `agent-docker-publish.yml` workflow in this repo already scopes this.

PR #5579, opened 2026-05-25
feat(llmisvc): model-based routing gates and models in status

pierDipi:
Gate model-based routing on the serving.kserve.io/model-based-routing-enabled
  annotation, which is now set in all default config templates. This prevents
  older presets (without the alternative served model name) from silently
  enabling model-based routing. Move routing rule expansion/stripping from
  config merge to HTTPRoute construction for better separation of concerns.

  Add ModelSourcedAddressStatus to surface served model names in
  status.addresses, including LoRA adapter names. Refactor label/annotation
  propagation to use PropagateMap with skipKeys to prevent the routing gate
  annotation from leaking into pod templates.

Example status.addresses

```yaml
    addresses:
    - models:
      - name: publishers/default/models/facebook/opt-125m
      name: internal-model-routing
      origin:
        kind: Gateway
        name: kserve-ingress-gateway
        namespace: kserve
      url: http://172.18.0.3/
    - models:
      - name: publishers/default/models/facebook/opt-125m
      - name: facebook/opt-125m
      name: internal
      origin:
        kind: Gateway
        name: kserve-ingress-gateway
        namespace: kserve
      url: http://172.18.0.3/default/facebook-opt-125m-single
```


pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -522,7 +531,7 @@ func (r *LLMISVCReconciler) CollectReferencedGateways(ctx context.Context, llmSv
 
 // EvaluateHTTPRouteConditions evaluates the readiness of all HTTPRoutes referenced by the LLMInferenceService
 // and updates the HTTPRoutesReady condition accordingly. Also detects Gateway rejection of v1alpha2 backendRefs.
-func (r *LLMISVCReconciler) EvaluateHTTPRouteConditions(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+func (r *LLMISVCReconciler) EvaluateHTTPRouteConditions(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, cfg *Config) error {
```
bartoszmajsak, line 534: This threading of cfg is a bit noisy, but once #5573 lands, we should be good. Just marking it here so we can address it in the other PR.

pkg/controller/v1alpha2/llmisvc/router_discovery_filter.go
```diff
@@ -77,23 +82,74 @@ func IsModelRoutingURL(url *apis.URL) bool {
...
 		typeName = "internal"
 	}
 
-	if IsModelRoutingURL(url) {
-		return typeName + "-model-routing"
+	models := make([]v1alpha2.ModelSourcedAddressStatus, 0, 2)
+	const modelRoutingFmt = "publishers/%s/models/%s"
+
+	// Ensure llmSvc.Spec.Model.Name is set.
+	llmSvc.SetDefaults(ctx)
```
bartoszmajsak: There's a little bit more going on behind the scenes when we default top-level, maybe we could simply call

```golang
modelName := ptr.Deref(llmSvc.Spec.Model.Name, llmSvc.GetName())
```

if it's the only thing that matters here?

pkg/utils/utils.go
```diff
@@ -116,14 +117,17 @@ func PropagatePrefixedMap(source map[string]string, dest *map[string]string, pre
 
 // PropagateMap copies all key-value pairs from source to dest.
 // Initializes dest if nil. No-op if source is empty.
-func PropagateMap(source map[string]string, dest *map[string]string) {
+func PropagateMap(source map[string]string, dest *map[string]string, skipKeys ...string) {
```
bartoszmajsak, line 120: Should we also call add `skipKeys` here:

https://github.com/kserve/kserve/blob/e5e9eb9882ccb842c133eeb798b28df0602ba82a/pkg/controller/v1alpha2/llmisvc/workload.go#L157

pierDipi: Done

pkg/controller/v1alpha2/llmisvc/router_discovery_filter.go
pierDipi: Done

PR #5584, opened 2026-05-26
refactor(llmisvc): no pagination in enqueue handlers

bartoszmajsak:

The `enqueueOn*Change` handlers in the LLMISvc controller used continue-token pagination loops when listing `LLMInferenceService` resources. Since these lists go through the controller-runtime informer cache - which always returns the full result set in a single call - the pagination was dead code adding unnecessary complexity.

Removes the pagination loops from all four handlers (Gateway, HTTPRoute, InferencePool, LLMInferenceServiceConfig), making them consistent with `enqueueOnConfigMapChange` which was already written without pagination.

- [x] Code compiles cleanly
- [x] Existing unit tests pass - no behavioral change, purely structural simplification

The informer cache's `List()` ignores `Continue` tokens entirely - they're only meaningful for direct API server calls. This is a no-op change in terms of behavior.

**Checklist**:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/controller.go
+++ b/pkg/controller/v1alpha2/llmisvc/controller.go
@@ -387,69 +387,59 @@ func (r *LLMISVCReconciler) enqueueOnGatewayChange(logger logr.Logger) handler.E
 			return reqs
 		}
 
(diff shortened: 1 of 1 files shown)


vivekk16 (commented):
LGTM. Clean removal of dead code.

PR #5589, opened 2026-05-26
fix: add WorkloadRef to PodSpec and bump k8s.io/* to v0.35.1

rajatvig:
`k8s.io/api` v0.35 added `WorkloadRef` to `corev1.PodSpec`, breaking the direct struct cast in `NewCustomExplainer`, `NewCustomPredictor`, and `NewCustomTransformer`. Adds the missing field to kserve's `PodSpec` copy to restore structural identity, and bumps all `k8s.io/*` deps to v0.35.1 (required since `storagemigration/v1alpha1` was promoted to `v1beta1` in v0.35).

Fixes #

- [x] `go build ./...` passes
- [x] `go test ./pkg/apis/serving/v1beta1/...` passes

**Checklist**:

**Release note**:
```release-note
Fix build failure when using k8s.io/api v0.35+: add WorkloadRef field to kserve PodSpec
```


pkg/apis/serving/v1beta1/podspec.go
```diff
@@ -351,4 +351,11 @@ type PodSpec struct {
 	// +featureGate=HostnameOverride
 	// +optional
 	HostnameOverride *string `json:"hostnameOverride,omitempty" protobuf:"bytes,41,opt,name=hostnameOverride"`
+	// WorkloadRef provides a reference to the Workload object that this Pod belongs to.
+	// This field is used by the scheduler to identify the PodGroup and apply the
+	// correct group scheduling policies.
+	//
+	// +featureGate=GenericWorkload
+	// +optional
+	WorkloadRef *corev1.WorkloadReference `json:"workloadRef,omitempty" protobuf:"bytes,42,opt,name=workloadRef"`
```
vivekk16, line 364: nit: Since this file is an explicit mirror of corev1.PodSpec, could we keep the comment in sync with upstream? The [upstream comment](https://github.com/kubernetes/api/blob/v0.35.1/core/v1/types.go) includes immutability and scheduling-policy caveats that are useful context:

```
	// WorkloadRef provides a reference to the Workload object that this Pod belongs to.
	// This field is used by the scheduler to identify the PodGroup and apply the
	// correct group scheduling policies. The Workload object referenced
	// by this field may not exist at the time the Pod is created.
	// This field is immutable, but a Workload object with the same name
	// may be recreated with different policies. Doing this during pod scheduling
	// may result in the placement not conforming to the expected policies.
	//
	// +featureGate=GenericWorkload
	// +optional
	WorkloadRef *corev1.WorkloadReference `json:"workloadRef,omitempty" protobuf:"bytes,42,opt,name=workloadRef"`
```

go.mod
vivekk16, line 1: The go.mod has a replace directive pinning controller-runtime to v0.19.7 (designed for k8s.io v0.31.x) while k8s.io deps are now at v0.35.1. This was pre-existing but the gap is widening. Any plans to bump controller-runtime or remove the replace?

vivekk16 (commented):
nit: Consider adding a roundtrip test that sets WorkloadRef on kserve's PodSpec and verifies it's preserved after the corev1.PodSpec() cast. The compile check catches missing fields but not type mismatches that could silently drop data.

pkg/apis/serving/v1beta1/podspec.go
rajatvig: Updated

go.mod
rajatvig: Dropped the pin. It did imply fixing some deprecation warnings

rajatvig:
@vivekk16 Do the changes look good?

PR #5599, opened 2026-05-28
feat(llmisvc): add kvCacheOffloading spec for CPU KV cache tiering

andresllh:
## Summary

- Adds `kvCacheOffloading` as a structured field on `WorkloadSpec` (available as `spec.kvCacheOffloading` and `spec.prefill.kvCacheOffloading` for P/D disaggregated deployments) with two sub-fields:
  - `cpu` — CPU memory to reserve for KV cache offloading, expressed as a Kubernetes quantity (e.g. `10Gi`)
  - `evictionPolicy` — optional eviction policy string passed through to vLLM
- Adds a `kvTransferConfig` Go template function in `config_merge.go` that renders the `--kv-transfer-config` JSON flag from the spec, converting the `resource.Quantity` to bytes via `.Value()`.
- All six vLLM entrypoint templates gain a `KV_TRANSFER_ARGS` bash block that injects `--kv-transfer-config` when `kvCacheOffloading` is set, and skips injection when `--kv-transfer-config` (or `--kv_transfer_config`) is already present in `VLLM_ADDITIONAL_ARGS` or positional args to avoid duplicate flags.
- The `kserve-config-llm-cpu-offload` preset has been removed — users set `kvCacheOffloading` directly in their `LLMInferenceService` spec instead.
- CRD, chart, and install-script manifests regenerated via `make precommit`.

## How it works

Set `kvCacheOffloading` directly on the workload spec:

```yaml
spec:
  kvCacheOffloading:
    cpu: 10Gi
```

For P/D disaggregated deployments, set it on the prefill spec:

```yaml
spec:
  prefill:
    kvCacheOffloading:
      cpu: 10Gi
```

The controller renders this into the vLLM entrypoint as:
```
--kv-transfer-config '{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"spec_name":"TieringOffloadingSpec","cpu_bytes_to_use":10737418240}}'
```

Users who need secondary tier support or custom connector configs can still pass `--kv-transfer-config` directly via `VLLM_ADDITIONAL_ARGS`; the entrypoint detects this and skips automatic injection.

## Test plan

- [ ] `go test ./pkg/controller/v1alpha2/llmisvc/...` passes (all config_merge tests including nil, basic, and eviction-policy kvCacheOffloading scenarios)
- [ ] Deploy with `kvCacheOffloading.cpu: 10Gi` on a cluster with GPU nodes and verify `--kv-transfer-config` appears in the vLLM serve command
- [ ] Verify KV cache offloading is active by observing reduced GPU memory pressure under load
- [ ] Verify that setting `--kv-transfer-config` in `VLLM_ADDITIONAL_ARGS` suppresses automatic injection


zdtsw:
is this plan to intergrate with llm-d 0.8.0 and only for cpu offloading?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -144,6 +144,33 @@ type WorkloadSpec struct {
...
+	// +kubebuilder:default=lru
+	EvictionPolicy string `json:"evictionPolicy,omitempty"`
+
+	// SecondaryTiers lists optional additional storage tiers for KV cache cascading.
+	// Each entry is an opaque object with at minimum a "type" field identifying the tier
+	// implementation (e.g. {"type":"fs_python","path":"/mnt/kv-cache"}).
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	SecondaryTiers []runtime.RawExtension `json:"secondaryTiers,omitempty"`
```
pierDipi: maybe for now should we hold off at doing this since I expect we need to "manage" the connection details to remote storage for other tiers and a more structured approach might be better?

I think that's where the API adds more value since doing it through args is still possible today without locking ourself into an unstructured API?

andresllh: I've removed this unstructured API.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -144,6 +144,25 @@ type WorkloadSpec struct {
...
+	// +optional
+	KVCacheOffloading *KVCacheOffloadingSpec `json:"kvCacheOffloading,omitempty"`
+}
+
+// KVCacheOffloadingSpec configures KV cache offloading via vLLM's OffloadingConnector.
+type KVCacheOffloadingSpec struct {
+	// CPUBytesToUse is the number of CPU RAM bytes to allocate as the primary KV cache tier
+	// (maps to vLLM kv_connector_extra_config.cpu_bytes_to_use).
+	// +kubebuilder:validation:Minimum=1
+	CPUBytesToUse int64 `json:"cpuBytesToUse"`
```
pierDipi: I'm wondering if we should use a higher level [k8s quantity](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity) like "10Gi" that gets converted to bytes when passed down

pierDipi: consequentially field name would become `cpu`

```
kvCacheOffloading:
  cpu: 50Gi
```

reads and it's easier to use than converting GiB to bytes beforehand

andresllh: Done

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -604,6 +581,37 @@ func ReplaceVariables(llmSvc *v1alpha2.LLMInferenceService, llmSvcCfg *v1alpha2.
 			// (or nil): max(0, tgps - preStop - min(5, tgps)), defaulting tgps to 60 when unset.
 			// The 5-second buffer reserves time for signal propagation and final process cleanup
 			// before Kubernetes sends SIGKILL.
```
pierDipi: this comment is out of place now, should be down

pierDipi (approved):

PR #5600, opened 2026-05-29
test(e2e): added http/https download test cases for multi-model serving

### LINKED ISSUE #1576: Multi-model serving e2e test for http/https
/kind feature

Multi-model serving e2e test only test downloading from gcs. We should update this to **also** download from http/https. https://github.com/kubeflow/kfserving/blob/master/test/e2e/predictor/test_multi_model_serving.py

Madhumasa84:
    This PR resolves and closes issue #1576 by updating the multi-model serving E2E tests (`test_multi_model_serving.py`) to test downloading model artifacts from HTTP/HTTPS storage URIs 
  in addition to GCS.                                                                                                                                  
                                                                                                                                                       
    We added HTTPS counterparts for:                                                                                                                   
    - Sklearn `v1` and `v2` protocols under `test_mms_sklearn_kserve`.                                                                                 
    - XGBoost `v1` and `v2` protocols under `test_mms_xgboost_kserve`.                                                                                 
                                                                                                                                                       
    Fixes #1576                                                                                                                                        
                                                                                                                                                       
    - Verified all public HTTP/HTTPS model file endpoints are reachable and return `200 OK`.                                                           
    - Checked and compiled the test script file: `python3 -m py_compile test/e2e/predictor/test_multi_model_serving.py`.
    - Ran the core storage initializer HTTP/HTTPS unit tests to verify the downloader resolution behaves correctly:
      ```bash
      PYTHONPATH=python/storage pytest python/storage/test/test_storage.py -k "test_http_uri_path or test_https_uri_path"
      # Output: 23 passed, 33 deselected
  
  Special notes for your reviewer:
  Note that the HTTP/HTTPS storage URIs point directly to individual model files ( model.joblib  and  model.bst ) instead of directory paths. Since    
  HTTP/HTTPS protocol does not support native directory listing, the storage-initializer requires a specific file endpoint to download the model       
  successfully and avoid raising  No filename contained in URI  errors.
  
  Checklist:
  
  [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
  [ ] Has code been commented, particularly in hard-to-understand areas?
  [ ] Have you made corresponding changes to the documentation https://github.com/kserve/website?
  
  Release note:
    test(e2e): Add HTTP/HTTPS download test cases for multi-model serving

### DIFF
--- a/test/e2e/predictor/test_multi_model_serving.py
+++ b/test/e2e/predictor/test_multi_model_serving.py
@@ -44,6 +44,14 @@ from ..common.utils import KSERVE_TEST_NAMESPACE
             "v2",
             "gs://seldon-models/sklearn/mms/lr_model",
         ),
+        (
+            "v1",
+            "https://storage.googleapis.com/kfserving-examples/models/sklearn/1.0/model/model.joblib",
+        ),
+        (
+            "v2",
+            "https://storage.googleapis.com/seldon-models/sklearn/mms/lr_model/model.joblib",
+        ),
     ],
 )
 @pytest.mark.mms
@@ -164,6 +172,14 @@ async def test_mms_sklearn_kserve(
         (
             "v2",
             "gs://seldon-models/xgboost/mms/iris",
+        ),
+        (
+            "v1",
+            "https://storage.googleapis.com/kfserving-examples/models/xgboost/1.5/model/model.bst",
+        ),
+        (
+            "v2",
+            "https://storage.googleapis.com/seldon-models/xgboost/mms/iris/model.bst",
         ),
     ],
 )
--- a/test/e2e/predictor/test_multi_model_serving.py
+++ b/test/e2e/predictor/test_multi_model_serving.py
@@ -34,37 +34,50 @@ from ..common.utils import KSERVE_TEST_NAMESPACE
 
 
 @pytest.mark.parametrize(
-    "protocol_version,storage_uri",
+    "protocol_version,storage_uri,storage_type",
     [
         (
             "v1",
+            # GCS directory URI: storage-initializer downloads all files under the prefix recursively
             "gs://kfserving-examples/models/sklearn/1.0/model",
+            "gcs",
         ),
         (
             "v2",
+            # GCS directory URI: storage-initializer downloads all files under the prefix recursively
             "gs://seldon-models/sklearn/mms/lr_model",
+            "gcs",
         ),
         (
             "v1",
+            # HTTPS single-file URI: storage-initializer fetches one file via HTTP GET.
+            # Both GCS and HTTPS cases result in the same model layout inside the container.
             "https://storage.googleapis.com/kfserving-examples/models/sklearn/1.0/model/model.joblib",
+            "https",
         ),
         (
             "v2",
+            # HTTPS single-file URI: storage-initializer fetches one file via HTTP GET.
+            # Both GCS and HTTPS cases result in the same model layout inside the container.
             "https://storage.googleapis.com/seldon-models/sklearn/mms/lr_model/model.joblib",
+            "https",
         ),
     ],
-    ids=["v1-gcs", "v2-gcs", "v1-http", "v2-http"],
+    ids=["v1-gcs", "v2-gcs", "v1-https", "v2-https"],
 )
 @pytest.mark.mms
 @pytest.mark.asyncio(scope="session")
 async def test_mms_sklearn_kserve(
     protocol_version: str,
     storage_uri: str,
+    storage_type: str,
     rest_v1_client,
     rest_v2_client,
     network_layer,
 ):
-    service_name = f"isvc-sklearn-mms-{protocol_version}"
+    # Include storage_type in the service name to avoid resource name collisions
+    # when multiple parametrized cases share the same protocol_version.
+    service_name = f"isvc-sklearn-mms-{protocol_version}-{storage_type}"
     # Define an inference service
     predictor = V1beta1PredictorSpec(
         min_replicas=1,
@@ -164,37 +177,50 @@ async def test_mms_sklearn_kserve(
 
 
 @pytest.mark.parametrize(
-    "protocol_version,storage_uri",
+    "protocol_version,storage_uri,storage_type",
     [
         (
             "v1",
+            # GCS directory URI: storage-initializer downloads all files under the prefix recursively
             "gs://kfserving-examples/models/xgboost/1.5/model",
+            "gcs",
         ),
         (
             "v2",
+            # GCS directory URI: storage-initializer downloads all files under the prefix recursively
             "gs://seldon-models/xgboost/mms/iris",
+            "gcs",
         ),
         (
             "v1",
+            # HTTPS single-file URI: storage-initializer fetches one file via HTTP GET.
+            # Both GCS and HTTPS cases result in the same model layout inside the container.
             "https://storage.googleapis.com/kfserving-examples/models/xgboost/1.5/model/model.bst",
+            "https",
         ),
         (
             "v2",
+            # HTTPS single-file URI: storage-initializer fetches one file via HTTP GET.
+            # Both GCS and HTTPS cases result in the same model layout inside the container.
             "https://storage.googleapis.com/seldon-models/xgboost/mms/iris/model.bst",
+            "https",
         ),
     ],
-    ids=["v1-gcs", "v2-gcs", "v1-http", "v2-http"],
+    ids=["v1-gcs", "v2-gcs", "v1-https", "v2-https"],
 )
 @pytest.mark.mms
 @pytest.mark.asyncio(scope="session")
 async def test_mms_xgboost_kserve(
     protocol_version: str,
     storage_uri: str,
+    storage_type: str,
     rest_v1_client,
     rest_v2_client,
     network_layer,
 ):
-    service_name = f"isvc-xgboost-mms-{protocol_version}"
+    # Include storage_type in the service name to avoid resource name collisions
+    # when multiple parametrized cases share the same protocol_version.
+    service_name = f"isvc-xgboost-mms-{protocol_version}-{storage_type}"
     # Define an inference service
     predictor = V1beta1PredictorSpec(
         min_replicas=1,
--- a/test/e2e/predictor/test_multi_model_serving.py
+++ b/test/e2e/predictor/test_multi_model_serving.py
@@ -53,6 +53,7 @@ from ..common.utils import KSERVE_TEST_NAMESPACE
             "https://storage.googleapis.com/seldon-models/sklearn/mms/lr_model/model.joblib",
         ),
     ],
+    ids=["v1-gcs", "v2-gcs", "v1-http", "v2-http"],
 )
 @pytest.mark.mms
 @pytest.mark.asyncio(scope="session")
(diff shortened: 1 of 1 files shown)


test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -44,6 +44,14 @@
...
             "gs://seldon-models/sklearn/mms/lr_model",
         ),
+        (
+            "v1",
+            "https://storage.googleapis.com/kfserving-examples/models/sklearn/1.0/model/model.joblib",
+        ),
+        (
+            "v2",
+            "https://storage.googleapis.com/seldon-models/sklearn/mms/lr_model/model.joblib",
+        ),
```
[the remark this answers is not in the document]
Madhumasa84: The `https://storage.googleapis.com/...` URLs point to the standard public HTTPS gateways of the **exact same GCS buckets** (`gs://kfserving-      
  examples` and `gs://seldon-models`) that the existing GCS test cases already depend on. Because they query the same underlying Google Cloud Storage  
  infrastructure, this does not introduce any new third-party host dependency or increase flakiness beyond what is already there.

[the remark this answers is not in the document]
Madhumasa84: Thanks! I have updated the branch and added explicit pytest parameter IDs (`ids=["v1-gcs", "v2-gcs", "v1-http", "v2-http"]`) for both sklearn and xgboost test decorators. This ensures the test IDs in logs are clean and uniquely identifiable (e.g.,`test_mms_sklearn_kserve[v1-gcs]` vs`test_mms_sklearn_kserve[v1-http]`).

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -165,6 +173,14 @@ async def test_mms_sklearn_kserve(
...
             "gs://seldon-models/xgboost/mms/iris",
         ),
+        (
+            "v1",
+            "https://storage.googleapis.com/kfserving-examples/models/xgboost/1.5/model/model.bst",
+        ),
+        (
+            "v2",
+            "https://storage.googleapis.com/seldon-models/xgboost/mms/iris/model.bst",
+        ),
```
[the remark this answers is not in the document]
Madhumasa84: For single-file frameworks like `sklearn` and `xgboost`, both GCS prefix downloads and HTTPS file downloads result in the exact same model         
  directory structure:
    - The GCS downloader fetches the contents under the prefix (which contains only `model.joblib` or `model.bst`) and saves it to the target directory.
    - The HTTPS downloader fetches the file directly (`model.joblib` or `model.bst`) and saves it to the target directory.
  
    In both cases, the final layout inside the container contains the exact same file in the root path, ensuring the downstream serving code path is   
  identical

Madhumasa84:
ignore-stale

spolti:
@Madhumasa84, hey, do you mind sending an empty commit to rettriger CI?

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -44,7 +44,16 @@
             "v2",
             "gs://seldon-models/sklearn/mms/lr_model",
         ),
+        (
```
mwaykole, line 51: `service_name` is derived from `protocol_version` only (`f"isvc-sklearn-mms-{protocol_version}"`), so the new HTTP parametrize cases with the same protocol version will collide with the GCS cases on the resource name. If tests run sequentially and cleanup succeeds, this works, but any failure mid-test leaves the ISVC behind and the next same-version case hits a 409. Consider encoding the storage type in the service name (e.g., `isvc-sklearn-mms-{protocol_version}-http`) to make them independent.

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -44,7 +44,16 @@
...
+            "v1",
+            "https://storage.googleapis.com/kfserving-examples/models/sklearn/1.0/model/model.joblib",
+        ),
```
mwaykole, line 57: The GCS URIs point to directories (storage-initializer downloads all files recursively), while these HTTPS URIs point to single files. These exercise fundamentally different download paths in the storage-initializer (GCS client `download_prefix` vs HTTP `requests.get` to a single file). Might be worth a brief comment noting this distinction, since a reader might assume they're equivalent URIs just with different schemes.

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -44,7 +44,16 @@
...
+        (
+            "v2",
+            "https://storage.googleapis.com/seldon-models/sklearn/mms/lr_model/model.joblib",
+        ),
     ],
+    ids=["v1-gcs", "v2-gcs", "v1-http", "v2-http"],
```
mwaykole: Nit: these are all `https://` URLs, not `http://`. The PR title says "http/https" but there's no plain-HTTP test case. Is plain `http://` download actually supported and worth testing separately, or should the title/ids just say "https"?

mwaykole (commented):
The test additions look reasonable but the service_name collision between GCS and HTTP cases sharing the same protocol_version is a real fragility risk in CI.

test/e2e/predictor/test_multi_model_serving.py
Madhumasa84: I've added `storage_type` (`"gcs"` or `"https"`) as a third parameter to the `@pytest.mark.parametrize` decorator and embedded it in the service name:
```python
service_name = f"isvc-sklearn-mms-{protocol_version}-{storage_type}"
```
This gives each parametrized case a unique resource name (e.g. `isvc-sklearn-mms-v1-gcs` vs `isvc-sklearn-mms-v1-https`), so a mid-test failure leaving an ISVC behind won't affect the next case.

test/e2e/predictor/test_multi_model_serving.py
Madhumasa84: Added inline comments on each URI tuple to make the distinction clear:

- **GCS entries:** `# GCS directory URI: storage-initializer downloads all files under the prefix recursively`
- **HTTPS entries:** `# HTTPS single-file URI: storage-initializer fetches one file via HTTP GET. Both GCS and HTTPS cases result in the same model layout inside the container.`

test/e2e/predictor/test_multi_model_serving.py
Madhumasa84: The test IDs have been updated to `"v1-https"` and `"v2-https"` to accurately reflect the scheme used.

Regarding plain `http://` the storage-initializer does support it (the HTTP handler resolves both `http://` and `https://` schemes), but all publicly available model artifacts for this test (from GCS and Seldon buckets) are served over HTTPS. Adding a plain `http://` test case would require a dedicated unauthenticated HTTP server hosting the model files, which is outside the scope of this PR.

The PR title has been phrased as "http/https" to reflect the general protocol support being exercised, but the test coverage here is specifically HTTPS. Happy to update the PR title to `test(e2e): add HTTPS download test cases for multi-model serving` if that's clearer.

Madhumasa84:
Would appreciate a review whenever someone has a chance. 
Thank you

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -34,28 +34,50 @@
...
     storage_uri: str,
+    storage_type: str,
     rest_v1_client,
     rest_v2_client,
     network_layer,
 ):
-    service_name = f"isvc-sklearn-mms-{protocol_version}"
+    # Include storage_type in the service name to avoid resource name collisions
+    # when multiple parametrized cases share the same protocol_version.
+    service_name = f"isvc-sklearn-mms-{protocol_version}-{storage_type}"
```
cjohannsen-cloudera, line 80: The same `storage_type` suffix is missing from `model_names` a few lines below (lines 111–113). `v1-gcs` and `v1-https` both resolve to `model1-sklearn-v1` / `model2-sklearn-v1`. If `v1-gcs` fails before cleanup, `v1-https` hits a Kubernetes 409 Conflict on `create_trained_model` — a failure completely unrelated to HTTP fetching.

```python
# Currently (lines 111–113):
model_names = [
    f"model1-sklearn-{protocol_version}",
    f"model2-sklearn-{protocol_version}",
]

# Should be:
model_names = [
    f"model1-sklearn-{protocol_version}-{storage_type}",
    f"model2-sklearn-{protocol_version}-{storage_type}",
]
```

cjohannsen-cloudera (changes requested):
This PR correctly adds `storage_type` to `service_name` to prevent `InferenceService` name collisions, but the same fix is incomplete — `model_names` in both test functions still template on `protocol_version` only. Additionally, cleanup is not guarded by `finally`, which turns the name-collision bug from a theoretical race into a near-certain session failure.

**Issue 3 — cleanup not in `finally`**

Lines 173–176 (`test_mms_sklearn_kserve`) and lines 317–320 (`test_mms_xgboost_kserve`) clean up resources only on the happy path. If any assertion fails, `TrainedModel` and `InferenceService` objects remain in the namespace. Combined with the shared `model_names` bug, a single flaky `v1-gcs` case permanently blocks `v1-https` for the rest of the session with a 409 Conflict that looks unrelated to HTTP fetching.

```python
# Wrap assertions so cleanup always runs
try:
    # ... responses / assert ...
finally:
    for model_name in model_names:
        kserve_client.delete_trained_model(model_name, KSERVE_TEST_NAMESPACE)
    kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
```

test/e2e/predictor/test_multi_model_serving.py
```diff
@@ -155,28 +177,50 @@ async def test_mms_sklearn_kserve(
...
     storage_uri: str,
+    storage_type: str,
     rest_v1_client,
     rest_v2_client,
     network_layer,
 ):
-    service_name = f"isvc-xgboost-mms-{protocol_version}"
+    # Include storage_type in the service name to avoid resource name collisions
+    # when multiple parametrized cases share the same protocol_version.
+    service_name = f"isvc-xgboost-mms-{protocol_version}-{storage_type}"
```
cjohannsen-cloudera, line 224: Same issue here — `model_names` at lines 254–256 still templates on `protocol_version` only. `v1-gcs` and `v1-https` will both create `model1-xgboost-v1` / `model2-xgboost-v1`, causing the same 409 Conflict on collision.

```python
# Currently (lines 254–256):
model_names = [
    f"model1-xgboost-{protocol_version}",
    f"model2-xgboost-{protocol_version}",
]

# Should be:
model_names = [
    f"model1-xgboost-{protocol_version}-{storage_type}",
    f"model2-xgboost-{protocol_version}-{storage_type}",
]
```

Madhumasa84:
Thankyou @cjohannsen-cloudera for the review, will resolve them.

Madhumasa84:
I have updated the PR to resolve all the requested changes:

1. **Include `storage_type` in `model_names`**:
       - Updated `model_names` in `test_mms_sklearn_kserve` to `model1-sklearn-{protocol_version}-{storage_type}` and `model2-sklearn-{protocol_version}-{storage_type}`.
       - Updated `model_names` in `test_mms_xgboost_kserve` to `model1-xgboost-{protocol_version}-{storage_type}` and `model2-xgboost-{protocol_version}-{storage_type}`.
       - This ensures each parametrized test case (`v1-gcs`, `v1-https`, etc.) creates uniquely named `TrainedModel` resources and avoids Kubernetes 409 Conflict errors.

2. **Guarded Cleanup with `try...finally`**:
       - Wrapped test creation and predictions inside a `try` block in both test functions.
       - Placed the resource deletion calls (`delete_trained_model` and `delete`) inside the `finally` block to ensure cleanup always executes, even if assertions or readiness checks fail mid-test.

    Verified all test collections and unit tests locally.

PR #5590, opened 2026-05-27
feat(kernelcache): add kernelcache/mcv initial code donation

spolti:
- Add kernelcache/mcv Go module to fmt, vet, go-lint, and tidy Makefile targets
- Add docker-build-mcv and docker-push-mcv targets
- Update hack/boilerplate.sh and boilerplate.go.txt to cover mcv files with 2026 copyright headers
- Clean up mcv.Dockerfile: remove entrypoint script, trim unused runtime packages, use debian:bookworm-slim base
- Exclude kernelcache/mcv/example from ruff linting

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Part of https://github.com/kserve/kserve/issues/5476


**Feature/Issue validation/testing**:
- `cmd/main_test.go` — CLI flag combination validation (TestValidateFlagCombinations)                                                                                                                                                                                  
- `pkg/cache/vllm_test.go` — vLLM cache detection: no-cache, Mega AOT single/multi rank, rank skip, metadata marshaling                                                                                                                                                
- `pkg/client/client_test.go` — Client API: extract with GPU enabled/disabled, skip precheck, cache dir, invalid image names, inspect, preflight check, stub mode, timeout                                                                                             
- `pkg/config/config_test.go` — Config initialization defaults, environment overrides, setters/getters                                                                                                                                                                 
- `pkg/fetcher/fetcher_test.go` — Docker/Podman fetcher failures, remote fetcher, fallback to remote, tarball loading                                                                                                                                                  
- `pkg/imgbuild/builder_test.go` — Builder selection: Buildah available, Docker fallback, unsupported, auto-detect                                                                                                                                                     
- `pkg/imgbuild/utils_test.go` — Dockerfile generation, directory cleanup, cleanup with timeout                                                                                                                                                                        
- `pkg/logformat/logformat_test.go` — Log level configuration                                                                                                                                                                                                          
- `pkg/utils/utils_test.go` — File path existence, app detection, group JSON sanitization, MCV dir cleanup                                                                                                                                                             
                
GHA action also added that covers:
- Builds MCV container, runs go test, builds and validates cache images from all test fixtures
- Added GHA to publish the MCV container.                                                                                       

- Logs

**Special notes for your reviewer**:
This is part of the KernelCache feature delivery; this is the CLI tool and API for cache extraction and preflight checks.
[...]

PR #5601, opened 2026-05-29
feat(kernelcache): add GPU kernel caching support for LLM workloads

### LINKED ISSUE #5476: Proposal: Add GPU Kernel Caching Support to KServe
/kind feature

## Description
When serving Large Language Models (LLMs), frameworks like vLLM, PyTorch and Triton translate GPU kernels implemented through higher-level programming languages into CUDA PTX or ROCm HASCO assembly before being executed by the GPU driver (for example, via torch.compile). This just-in-time (JIT) compilation occurs each time a model is loaded and can significantly delay model startup by 30-120 seconds each time it is deployed in Kubernetes Pods, longer for larger models. This proposal adds support to KServe to manage GPU kernel caches for LLM workloads running in KServe Clusters, allowing LLM workloads to startup with a warm Kernel Cache, skipping the JIT compilation and reducing startup time dramatically.

## Design Documents
[Add GPU Kernel Caching to KServe: RFC](https://docs.google.com/document/d/1pfVUCUzhGWGsFruDA8YlgJNu0SP8Kjo4Qa-B74kPdMA/edit?usp=sharing)
Add GPU Kernel Caching to KServe: Design Document - **In progress**

Billy99:

This PR implements GPU kernel caching for KServe to dramatically reduce LLM inference startup times by eliminating JIT compilation overhead.

**Problem:** GPU frameworks (vLLM, PyTorch, Triton) perform just-in-time kernel compilation on every pod startup, causing 30-120 second delays for model deployment. Each time a model loads, kernels are recompiled to CUDA PTX or ROCm HASCO assembly.

**Solution:** This PR adds KernelCache and KernelCacheNode CRDs to manage pre-compiled GPU kernels packaged as OCI images. Kernels are extracted once per cluster/node and mounted into inference pods, enabling warm cache startup with near-instant model loading.

**Architecture:**
- Controller (operator): Creates Download PVCs, manages KernelCacheNodes per worker node, aggregates status
- Agent (DaemonSet): Launches extraction Jobs using gkm-extract container, tracks per-node cache status, populates GPU hardware info via MCV
- Webhooks: Image digest resolution with cosign verification, HMAC signature protection, deletion protection when caches in use 

**Key Features:**
- MCV GPU Detection: Auto-detects GPU hardware (AMD/NVIDIA) on each node, supports heterogeneous configurations
- Cosign Integration: Verifies OCI image signatures for supply chain security
- Runtime Configuration: ConfigMap-driven settings (no restart needed for image updates, timeouts, testing flags)
- Production-Ready Deletion Protection: Webhooks prevent cache deletion when pods are using them

**Integration Points:**
- Follows LocalModelCache patterns for PVC management and Job launching
- Integrates with existing inferenceservice-config ConfigMap
- Uses KServe cosign verification infrastructure

Fixes #5476

Test Coverage:
- Unit tests: Controller reconcilers, webhooks, GPU detection, status aggregation
- E2E tests: Multi-node extraction, PVC lifecycle, status tracking (test/e2e/kernelcache/)
  
All tests passing locally.

**Deployment and Usage:**

NOTE: `inferenceservice-config` ConfigMap was updated with Kernel Cache fields, so KServe must be built and deployed from this PR. Because Kernel Cache is an optional feature, it must be enabled in ConfigMap before being used.

Build and deploy to Cluster with GPU:
```
# 1. Create KIND cluster with local registry, or skip if on hardware
cd ~/src/kserve/
./hack/setup/dev/manage.kind-with-registry.sh

# 2. Set environment variables (add to ~/.bashrc for persistence)
export KO_DOCKER_REPO=localhost:5001
export KO_DEFAULTPLATFORMS=linux/amd64

# 3. Deploy KServe (includes LocalModel & KernelCache controllers)

# Deploy KServe
make deploy-dev

# Deploy LocalModel and KernelCache to KServe
## A. Hardware
make deploy-dev-localmodel

## B. KIND Cluster
make deploy-dev-localmodel-kind

# 4. Enable KernelCache in ConfigMap:

## A. Hardware
kubectl patch configmap inferenceservice-config -n kserve --type merge -p '{"data": {"kernelcache":{\"enabled\": true}"}}'

## B. KIND Cluster
kubectl patch configmap -n kserve inferenceservice-config --type=merge -p '{"data":{"kernelcache":"{\"enabled\": true, \"noGPU\": true, \"enablePermissionInitContainer\": true}"}}'

KernelCache ConfigMap settings:
- enabled: false|true Enable/disable KernelCache feature
- jobNamespace: Namespace for extraction jobs (default: kserve)
- extractImage: OCI image extraction tool (adjust for your architecture)
   - ARM64: quay.io/billy99/gkm-extract:aarch64
   - AMD64: quay.io/billy99/gkm-extract:amd64
- captureImage: MCV in Container for Capture Side-car (adjust for your architecture)
   - ARM64: quay.io/billy99/kserve-mcv:latest-aarch64
   - AMD64: quay.io/billy99/kserve-mcv:latest
- fsGroup: File system group for cache volume permissions
- jobTTLSecondsAfterFinished: How long to keep completed extraction jobs
- reconcileIntervalSeconds: How often to check cache status
- noGPU: false|true Allow extraction without GPU (for testing), needs to be true in KIND
- enablePermissionInitContainer: Fix PVC permissions before extraction, needs to be true in KIND
- allowUnsigned: false|true Need to set to true for KernelCacheCapture until MCV-PR#169 merges.

# 5. Verify deployment
kubectl get pods -n kserve | grep -E 'localmodel|kernelcache'

Expected output:
kserve-kernelcachenode-agent-xxxxx           1/1     Running
kserve-localmodel-controller-manager-xxxxx   1/1     Running
```

Iterating After Code Changes
```
## A. Hardware
make deploy-dev-localmodel

## B. KIND Cluster
make deploy-dev-localmodel-kind
```
  
Create a KernelCache using PVC and Extraction Job:
```
cat <<EOF | kubectl apply -f -
kind: KernelCache
metadata:
  name: test-cache
  namespace: default
spec:
  image: quay.io/gkm/cache-examples:vector-add-cache-rocm-v2
EOF
```

OR Create a KernelCache using Kubernetes Image Volume feature (no PVCs or Extraction Jobs):
```
cat <<EOF | kubectl apply -f -
kind: KernelCache
metadata:
  name: test-cache
  namespace: default
spec:
  image: quay.io/gkm/cache-examples:vector-add-cache-rocm-v2
  mountType: imageVolume
EOF
```

Verify:
```
# Check KernelCache created
kubectl get kernelcache -A

# Check KernelCacheNodes created
kubectl get kernelcachenodes

# Check cache status
kubectl get kernelcache test-cache -o yaml

# Check extraction Jobs (If using)
kubectl get jobs -n kserve -l app=kernel-cache-extract
```

To capture a Kernel Cache from an ISVC, create a KernelCacheCapture CR:

```
cat <<EOF | kubectl apply -f -
kind: KernelCacheCapture
metadata:
  name: test-cache-capture
  namespace: default
spec:
  # Target OCI image where cache will be pushed
  targetImage: kind-registry:5000/test-captured-cache:v1

  # Use vllm preset (resolves to /home/kserve/.cache/vllm)
  cachePreset: vllm

  # Auto-create KernelCache when capture completes
  createKernelCache:
    enabled: true  # default
    mountType: imageVolume  # default
```

Sample Inference Service:
```
kind: InferenceService
metadata:
  name: sklearn-iris-with-kc
  namespace: default
  labels:
    # This label triggers sidecar injection to enable Cache to be Captured from ISVC
    internal.serving.kserve.io/kernelcachecapture: "test-cache-capture"
    # KernelCache label - tells webhook to mount KC PVC
    internal.serving.kserve.io/kernelcache: "test-cache"
  annotations:
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
```

Then to trigger the Capture (once the ISVC comes up)"

```
kubectl patch kcc test-cache-capture --type=merge -p '{"spec":{"trigger":true}}'
```

OR

Sample Pod mounting PVC:
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: cache-test-pod
  namespace: default
  labels:
    pod: kernel-cache-test
spec:
  containers:
  - command:
    - sleep
    - 365d
    image: quay.io/fedora/fedora-minimal
    imagePullPolicy: IfNotPresent
    name: test
    securityContext:
      runAsNonRoot: true
      runAsUser: 1000
    volumeMounts:
    - mountPath: /kernel-caches
      name: kernel-volume
      readOnly: true
  securityContext:
    fsGroup: 1000
  tolerations:
  - effect: NoSchedule
    key: gpu
    operator: Equal
    value: "true"
  volumes:
  - name: kernel-volume
    persistentVolumeClaim:
      claimName: test-cache
EOF
```

For KIND or MiniKube, include Init-Container if using PVCs:
```
  initContainers:
  - command:
    - sh
    - -c
    - |
      chown -R 1000:1000 /cache
      chmod -R 775 /cache
    image: quay.io/fedora/fedora-minimal
    name: fix-permissions
    securityContext:
      runAsUser: 0
    volumeMounts:
    - mountPath: /cache
      name: kernel-volume
```

Phase 1 Scope (this PR):
- Namespace-scoped CRDs only (ClusterKernelCache deferred to Phase 2)
- RWX storage requirement (ReadWriteMany PVCs for multi-node sharing)
- Stub GPU mode for testing without real GPUs (noGPU: true)
- Manual KernelCache creation (automatic ISVC integration deferred to Phase 2)
  
MCV Integration (#5590):
- Imports MCV library from GKM (github.com/redhat-et/GKM/mcv) until #5590 is merged.
- Uses GetSystemGPUInfo() for hardware detection (stub mode in tests)
- gkm-extract container (quay.io/gkm/gkm-extract:latest) for OCI extraction
  
Future Work (Phase 2):
- NodeGroups support for RWO storage (following LocalModelCache pattern)
- ISVC auto-discovery webhooks (framework + GPU matching)
- ClusterKernelCache cluster-scoped variant 
- Sidecar auto-creation from compiled kernels
  
Testing Note: E2E CI workflow not included in this PR to keep initial review focused. Will add .github/workflows/e2e-test-kernelcache.yaml in follow-up commits once core functionality reviewed.

**Checklist**:

**Release note**:

```release-note
feat(kernelcache): Add GPU kernel caching support to accelerate LLM inference startup. Includes KernelCache/KernelCacheNode CRDs, controller/agent, webhooks with cosign verification, and MCV GPU detection integration. Reduces model deployment startup from 30-120s to near-instant by eliminating JIT compilation overhead.
```


Billy99:
Remaining items to do for this PR:

- [x] Filesystem cleanup after deletion
  - Agent doesn't delete /mnt/kernel-cache/ directories when cache removed
  - Add file_utils.go with removeModel() function (copy LocalModelNode pattern)
- [x] Configurable node selector
  - Currently hardcoded kserve/kernelcache: worker in DaemonSet
  - Make ConfigMap-configurable or use Kustomize overlays
  - Part of E2E testing
- [x] Unified Cache Controller (merge LocalModel + KernelCache, 25% resource savings)

config/configmap/inferenceservice.yaml
```diff
@@ -629,6 +629,27 @@ data:
          "disableVolumeManagement": false
        }
 
+     # ====================================== KERNELCACHE CONFIGURATION ======================================
+     # Example
+     kernelcache: |-
```
spolti, line 644: Shouldn't we use the same configuration section as modelCache? Some configurations can be reused by both, wdyt?

config/configmap/inferenceservice.yaml
```diff
@@ -629,6 +629,27 @@ data:
...
+       {
+         "enabled": false,
+         # jobNamespace specifies the namespace where the kernel cache extraction jobs and download PV/PVC will be created.
+         "jobNamespace": "kserve",
```
spolti, line 648: not sure if I understand it, the jobs are expected to extract the cache on the nodes where the model will be running, right? 

For community, kserve namespace is where the the kserve and other controllers are deployed, is it ok to keep it in the same ns?

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -306,7 +306,7 @@ func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1a
 
 	if utils.GetForceStopRuntime(llmSvc) {
 		llmSvc.Status.Router = nil
-		llmSvc.Status.Address = nil //nolint:staticcheck // retained for schema compatibility
+		llmSvc.Status.Address = nil
```
spolti: Any reason to remove it?

pkg/controller/v1beta1/inferenceservice/components/explainer.go
```diff
@@ -127,7 +128,11 @@ func (e *Explainer) Reconcile(ctx context.Context, isvc *v1beta1.InferenceServic
 		isvc.Spec.Explainer.Containers[0] = *container
 	}
 
-	podSpec := corev1.PodSpec(isvc.Spec.Explainer.PodSpec)
+	// Convert PodSpec to corev1.PodSpec via JSON marshaling
+	// This handles any field type differences between k8s.io/api versions
+	data, _ := json.Marshal(isvc.Spec.Explainer.PodSpec)
+	var podSpec corev1.PodSpec
+	_ = json.Unmarshal(data, &podSpec)
```
spolti: May you elaborate on why it is needed?

pkg/cosign/verify.go
```diff
@@ -0,0 +1,173 @@
+/*
+Copyright 2025 The KServe Authors.
```
spolti, line 2: nit: 2026

.github/workflows/e2e-test-kernelcache.yaml
```diff
@@ -0,0 +1,236 @@
+name: KernelCache E2E Tests
+
+on:
+  pull_request:
+    branches: [master, release*]
+    paths:
+      - "pkg/controller/v1alpha1/kernelcache*/**"
+      - "pkg/apis/serving/v1alpha1/kernel*"
+      - "pkg/webhook/**"
```
spolti: Shouldn't it be `pkg/webhook/kernelcache/**`

pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go
```diff
@@ -87,7 +87,7 @@ func (dst *LLMInferenceService) ConvertFrom(srcRaw conversion.Hub) error {
 		URL:    src.Status.URL,
 		Status: src.Status.Status,
 	}
-	dst.Status.Address = src.Status.Address //nolint:staticcheck // retained for schema compatibility
```
spolti, line 91: also here, why removing?

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,485 @@
...
+	"github.com/google/go-containerregistry/pkg/authn"
+	gcrname "github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	gcrremote "github.com/google/go-containerregistry/pkg/v1/remote"
+
+	"github.com/kserve/kserve/pkg/cosign"
+)
+
+// Webhook annotation keys (KServe namespaced)
+const (
```
spolti, line 49: shouldn't this webhook be in the webhooks package?

Also, may we move the contants to the `constants.go` file?

Makefile
```diff
@@ -492,6 +540,18 @@ docker-build-localmodelnode-agent:
 docker-push-localmodelnode-agent: docker-build-localmodelnode-agent
 	${ENGINE} buildx build ${ARCH} --push -t ${KO_DOCKER_REPO}/${LOCALMODEL_AGENT_IMG} -f localmodel-agent.Dockerfile .
 
+docker-build-kernelcache:
+	${ENGINE} buildx build ${ARCH} --build-context mcv=/home/bmcfall/src/GKM/mcv -t ${KO_DOCKER_REPO}/${KERNELCACHE_CONTROLLER_IMG} -f kernelcache.Dockerfile .
```
spolti: shouldn't we use the same controller than modelCache?

spolti: it is still pointing to your home directory.

Makefile
```diff
@@ -629,3 +689,21 @@ manifests-distro:
...
 # Optional local/downstream overrides (ignored if absent)
 -include Makefile.overrides.mk
+
+# KernelCache webhook secret management
+.PHONY: kernelcache-webhook-secret-file
+kernelcache-webhook-secret-file:
+	@mkdir -p config/secret
+	@[ -s config/secret/mutation.env ] || \
+	  (echo 'Generating config/secret/mutation.env'; \
+	   printf 'MUTATION_SIGNING_KEY=%s\n' "$$(head -c 32 /dev/urandom | base64 | tr -d '\n')" > config/secret/mutation.env)
```
spolti, line 742: for what it is used for?

Makefile
```diff
@@ -463,7 +511,7 @@ bump-version:
 
 # Build the docker image
 docker-build:
-	${ENGINE} buildx build ${ARCH} --build-arg GOTAGS=${GOTAGS} . -t ${KO_DOCKER_REPO}/${CONTROLLER_IMG}
+	${ENGINE} buildx build ${ARCH} --build-context mcv=/home/bmcfall/src/GKM/mcv --build-arg GOTAGS=${GOTAGS} . -t ${KO_DOCKER_REPO}/${CONTROLLER_IMG}
```
spolti: what it is needed for?

pkg/apis/serving/v1alpha1/llm_inference_service_conversion_test.go
```diff
@@ -922,8 +922,8 @@ func TestLLMInferenceServiceConversion_StatusRoundtrip_V1Alpha1ToV1Alpha2(t *tes
 	require.NoError(t, src.ConvertTo(hub))
 
 	assert.Equal(t, src.Status.URL.String(), hub.Status.URL.String())
-	require.NotNil(t, hub.Status.Address, "Address (singular) should be preserved in ConvertTo") //nolint:staticcheck // testing deprecated field
-	assert.Equal(t, externalURL.String(), hub.Status.Address.URL.String())                       //nolint:staticcheck // testing deprecated field
```
spolti, line 927: Any reason to remove this linter?

pkg/apis/serving/v1alpha1/kernel_cache_status.go
```diff
@@ -0,0 +1,130 @@
...
+	TotalPodsReady       int `json:"totalPodsReady"`       // Total ready pods
+	TotalPodsTerminating int `json:"totalPodsTerminating"` // Pods being deleted
+
+	// Per-namespace breakdown (for debugging)
+	// +optional
+	NamespaceCounts map[string]NamespaceServingCounts `json:"namespaceCounts,omitempty"`
+}
+
+// NamespacedName is already defined in local_model_cache_status.go
+// Reusing that type to avoid duplication
```
spolti: is this comment in the right place?

pkg/apis/serving/v1alpha1/kernel_cache_status.go
```diff
@@ -0,0 +1,130 @@
...
+	// Total incompatible GPU count
+	TotalIncompatibleGPUs int `json:"totalIncompatibleGPUs"`
+}
+
+// ServingStatus tracks serving PVC usage (Phase 2)
+type ServingStatus struct {
+	// Aggregate counts across all nodes/namespaces (Phase 2)
+	TotalNamespaces      int `json:"totalNamespaces"`      // Namespaces with serving PVCs
+	TotalPods            int `json:"totalPods"`            // Total pods using cache
+	TotalPodsReady       int `json:"totalPodsReady"`       // Total ready pods
```
spolti, line 122: What is the total pods-ready count here?

charts/kserve-crd/templates/serving.kserve.io_clusterservingruntimes.yaml
```diff
@@ -1982,6 +1982,10 @@ spec:
                                       type: integer
                                     signerName:
                                       type: string
+                                    userAnnotations:
```
spolti: I didn't understand where it is coming from. Can you please elaborate?

charts/kserve-kernelcache-crd-minimal/templates/serving.kserve.io_kernelcachenodes.yaml
```diff
@@ -0,0 +1,53 @@
...
+  names:
+    kind: KernelCacheNode
+    listKind: KernelCacheNodeList
+    plural: kernelcachenodes
+    singular: kernelcachenode
+  scope: Cluster
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .status.counts.cachesInUse
+      name: Caches-In-Use
```
spolti, line 18: I think the correct convention for these status columns is all uppercase.

charts/kserve-kernelcache-crd-minimal/templates/serving.kserve.io_kernelcaches.yaml
```diff
@@ -0,0 +1,65 @@
...
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .status.state
+      name: Status
+      type: string
+    - jsonPath: .status.counts.nodeCnt
+      name: Nodes
+      type: integer
+    - jsonPath: .status.counts.nodeInUseCnt
+      name: Node-In-Use
```
spolti, line 26: same

charts/kserve-kernelcache-crd-minimal/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: kserve-kernelcache-crd-minimal
+version: v0.18.0
```
spolti: we already released 0.19.0.cr1

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: kserve/kserve-controller:latest
```
spolti, line 11: We shouldn't change it.

config/kernelcachenodes/nodeselectors/gpu-amd.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
```
spolti, line 1: how are these node selectors used at runtime?

config/kernelcachenodes/nodeselectors/README.md
```diff
@@ -0,0 +1,445 @@
...
+Pre-made nodeSelector examples for KernelCache agent DaemonSet.
+
+This directory contains example patches for common node selection scenarios.
+Copy the patch content into your own Kustomize overlay to use.
+
+## Available Examples
+
+- **gpu-nvidia.yaml** - Run agent only on nodes with NVIDIA GPUs (`nvidia.com/gpu: "true"`)
+- **gpu-amd.yaml** - Run agent only on nodes with AMD GPUs (`amd.com/gpu: "true"`)
+- **kernelcache.yaml** - Run agent only on nodes labeled `kserve/kernelcache: worker`
```
spolti, line 12: shouldn't we use the default workers label?

config/overlays/test/configmap/inferenceservice.yaml
```diff
@@ -752,6 +771,17 @@ data:
       "disableVolumeManagement": false
     }
 
+  kernelcache: |-
+    {
+      "enabled": false,
```
spolti, line 788: First question: I think we discussed at some point that it would share the same controller as modelCache, as well as the same configuration block.

Second, where is this enable flag being used? I might have overlooked it; I haven't found it so far.
As the controllers are shared, shouldn't we use it inside the modelcache controller to enable or not the kernel cache controller?

go.mod
```diff
@@ -27,27 +28,31 @@ require (
...
 	gopkg.in/go-playground/validator.v9 v9.31.0
 	istio.io/api v1.27.1
 	istio.io/client-go v1.27.1
-	k8s.io/api v0.34.5
-	k8s.io/apiextensions-apiserver v0.34.3
-	k8s.io/apimachinery v0.34.5
-	k8s.io/client-go v0.34.5
-	k8s.io/code-generator v0.34.3
-	k8s.io/component-helpers v0.34.3
+	k8s.io/api v0.35.0
```
spolti: does kernel cache require it? If not, let's keep the previous version.

pkg/cosign/verify.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create verifier: %w", err)
+	}
+
+	// Verify each bundle until we find a valid one
+	var lastErr error
+	for i, bundle := range bundles {
+		log.V(1).Info("Verifying bundle", "index", i, "totalBundles", len(bundles))
+
+		policy := verify.NewPolicy(artifactDigestPolicyOption, verify.WithoutIdentitiesUnsafe())
```
spolti, line 130: This means any valid Sigstore signature from any signer passes verification -- a malicious image signed by any Fulcio-issued certificate would be accepted.

We need to find a more secure way to verify it, we could do it later, however, we can't let it as it is very easy to bypass.

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,485 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	gcrname "github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	gcrremote "github.com/google/go-containerregistry/pkg/v1/remote"
```
spolti, line 43: Imported twice.

config/kernelcachenodes/manager.yaml
```diff
@@ -0,0 +1,101 @@
...
+kind: DaemonSet
+metadata:
+  name: kserve-kernelcachenode-agent
+  namespace: kserve
+  labels:
+    app.kubernetes.io/name: kserve-kernelcachenode-agent
+    app: kserve-kernelcachenode-agent
+    control-plane: kserve-kernelcachenode-agent
+    controller-tools.k8s.io: "1.0"
+spec:
```
spolti, line 11: there are no probes, is this expected?

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1009 @@
...
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.KernelCache{}).
+		Owns(&corev1.PersistentVolume{}).      // Watch PV changes
```
spolti: does it work for clustered resources?

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1009 @@
...
+		Owns(&corev1.PersistentVolumeClaim{}). // Watch PVC changes
```
spolti, line 1136: don't we need to filter the PVCs here?

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,485 @@
...
+		// Invalid value, default to disabled and log a warning
+		kernelcacheLog.Info("Invalid value for KYVERNO_VERIFICATION_ENABLED, defaulting to disabled", "value", envValue)
+		return false
+	}
+}
+
+// resolveImageDigest resolves an image reference to its digest without verifying signatures.
+// This is used when Kyverno verification is disabled (development/testing mode).
+// It returns the image digest string (sha256:...) if successful.
+func resolveImageDigest(ctx context.Context, imageRef string) (string, error) {
```
spolti: is this function being called?

config/rbac/kernelcache/role.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - nodes/status
+  verbs:
+  - get
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  - persistentvolumes
+  verbs:
```
spolti, line 34: aren't these too permissive?
May we inherit the same roles than model cache?

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1009 @@
...
+	// Step 1: Handle deletion
+	if !kc.DeletionTimestamp.IsZero() {
+		return r.handleDeletion(ctx, kc, kernelCacheConfig)
+	}
+
+	// Add finalizer if not present
+	if !controllerutil.ContainsFinalizer(kc, KernelCacheFinalizerName) {
+		controllerutil.AddFinalizer(kc, KernelCacheFinalizerName)
+		if err := r.Update(ctx, kc); err != nil {
+			if strings.Contains(err.Error(), "object has been modified") {
```
spolti: For all occurrences of it, please update to `apierrors.IsConflict()`

config/kernelcachenodes/manager.yaml
```diff
@@ -0,0 +1,101 @@
...
+        securityContext:
+          runAsUser: 0
+          runAsNonRoot: false
+        volumeMounts:
+          - mountPath: /mnt/kernel-cache
+            name: kernel-caches
+      containers:
+      - command:
+        - /manager
+        image: ko://github.com/kserve/kserve/cmd/kernelcachenode
```
spolti, line 63: is this expected?

config/kernelcachenodes/manager.yaml
```diff
@@ -0,0 +1,101 @@
...
+      serviceAccountName: kserve-kernelcachenode-agent
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+        fsGroup: 1000
+        seccompProfile:
+          type: RuntimeDefault
+      initContainers:
+      - name: fix-permissions
+        image: quay.io/fedora/fedora-minimal:latest
```
spolti: lets use ubi micro here.
Or the same than modelCache.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -169,6 +170,18 @@ type LocalModelConfig struct {
 	DisableVolumeManagement      bool   `json:"disableVolumeManagement,omitempty"`
 }
 
+// +kubebuilder:object:generate=false
+type KernelCacheConfig struct {
+	Enabled                       bool   `json:"enabled"`
+	JobNamespace                  string `json:"jobNamespace"`
+	ExtractImage                  string `json:"extractImage,omitempty"`
+	FSGroup                       *int64 `json:"fsGroup,omitempty"`
```
spolti, line 178: this seems to be missing in the Python SDK

test/e2e/kernelcache/test_kernelcache_basic.py
```diff
@@ -0,0 +1,215 @@
...
+
+            except ApiException as e:
+                if e.status == 404:
+                    pytest.fail(
+                        f"KernelCacheNode {kcnode_name} was not created for node {node_name}"
+                    )
+                raise
+
+        # Wait for extraction Job to be created
+        await asyncio.sleep(15)
```
spolti: shouldn't we use pooling instead of sleed?

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -404,6 +417,17 @@ func NewLocalModelConfig(isvcConfigMap *corev1.ConfigMap) (*LocalModelConfig, er
 	return localModelConfig, nil
 }
 
+func NewKernelCacheConfig(isvcConfigMap *corev1.ConfigMap) (*KernelCacheConfig, error) {
+	kernelCacheConfig := &KernelCacheConfig{}
+	if kernelCache, ok := isvcConfigMap.Data[KernelCacheConfigName]; ok {
+		err := json.Unmarshal([]byte(kernelCache), &kernelCacheConfig)
+		if err != nil {
+			return nil, fmt.Errorf("unable to unmarshall kernelcache: %w", err)
```
spolti:
```suggestion
			return nil, fmt.Errorf("unable to unmarshal kernelcache: %w", err)
```

pkg/controller/v1alpha1/kernelcachenode/controller.go
```diff
@@ -0,0 +1,900 @@
...
+func (r *KernelCacheNodeReconciler) jobCompleted(job *batchv1.Job) bool {
+	for _, cond := range job.Status.Conditions {
+		if cond.Type == batchv1.JobComplete && cond.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *KernelCacheNodeReconciler) cleanupJobs(
```
spolti: is there a reason to keep it since it is no-op and called on reconcile cycles?

pkg/controller/v1alpha1/kernelcachenode/file_utils.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return false, err
+}
+
+func (f *FileSystemHelper) ensureCacheRootFolderExists() error {
+	// If the folder already exists, this will do nothing
+	if err := os.MkdirAll(f.cachesRootFolder, os.ModePerm); err != nil { //nolint:gosec // G301: kernel cache must be readable by model server running as a different UID
```
spolti: ModePerm is `0777`, I don't think it is a good idea.
Would that lead on users being able to execute scripts and execute it?

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,485 @@
...
+	for imageRef, status := range verifications {
+		if status != "pass" {
+			return fmt.Errorf("Kyverno verification status for %s is not 'pass': %s", imageRef, status)
+		}
+	}
+
+	return nil
+}
+
+// isKyvernoVerificationEnabled checks if Kyverno verification is enabled.
```
spolti, line 631: wondering if we shouldn't move it to the configMap configuraiton

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,485 @@
...
+// It reads from the KYVERNO_VERIFICATION_ENABLED environment variable.
+// Defaults to false (disabled) if not set or invalid.
+func isKyvernoVerificationEnabled() bool {
+	envValue := os.Getenv(EnvKyvernoEnabled)
+	if envValue == "" {
+		// Default to disabled
+		return false
+	}
+
+	// Parse the value - accept "true", "1", "yes" as enabled
```
spolti: do we need to accept all of it?
Or can we just use parseBool here and expect onlyu  `true` ?

pkg/apis/serving/v1alpha1/kernel_cache_node_types.go
```diff
@@ -0,0 +1,131 @@
...
+	Message string `json:"message,omitempty"`
+
+	// LastUpdate timestamp
+	// +optional
+	LastUpdate metav1.Time `json:"lastUpdate,omitempty"`
+}
+
+// CacheNodeExtractionStatus is deprecated - use CacheNodeCacheInfo
+// Kept for backward compatibility during migration
+type CacheNodeExtractionStatus = CacheNodeCacheInfo
```
spolti: Do we need to keep it?
I mean it is a new API; it is weird it already starting with deprecated fields.

pkg/controller/v1alpha1/kernelcachenode/controller.go
```diff
@@ -0,0 +1,900 @@
...
+	nodeName         = os.Getenv("NODE_NAME")
+	cachesRootFolder = filepath.Join(kernelcachecommon.MountPath, "kernel-cache")
+	fsHelper         FileSystemInterface
+)
+
+type KernelCacheNodeReconciler struct {
+	client.Client
+	Clientset *kubernetes.Clientset
+	Log       logr.Logger
+	Scheme    *runtime.Scheme
```
spolti, line 73: May we move the fsHelper here so it is not lazily initialized?

pkg/controller/v1alpha1/kernelcachenode/controller.go
```diff
@@ -0,0 +1,900 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
+	"github.com/kserve/kserve/pkg/apis/serving/v1beta1"
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/controller/v1alpha1/kernelcachecommon"
+)
+
+var (
+	nodeName         = os.Getenv("NODE_NAME")
```
spolti, line 59: what happens if this env is empty?

config/configmap/inferenceservice.yaml
Billy99: So LocalModelCache defaults to "kserve-localmodel-jobs", but that namespace is not created by default, the user must create it. I agree that the Jobs pollute the "kserve" namespace, but I think it's weird to have a default that doesn't work out of the box. I have followed LMC on most other things, so I'm fine changing it to their pattern and document that the namespace must be created manually.

.github/workflows/e2e-test-kernelcache.yaml
Billy99: Done - Removing this line.
LMC has the same line in their workflow yaml. It is to cover both mutating and admission webhooks. KC does not have admission webhook at the moment so removing this line. If/When that is added, this line should be "pkg/webhook/admission/kernelcache/**".

charts/kserve-crd/templates/serving.kserve.io_clusterservingruntimes.yaml
Billy99: I did try to keep these isolated to it's own commit. After updating k8s.io dependencies, k8s.io/component-helpers moved v0.34.3 to
v0.35.0. Why k8s.io/component-helpers moved from v0.34.3 to v0.35.0?

Root Cause Chain:
1. KernelCache webhooks needed cosign v3 for OCI image signature verification
2. Cosign v3 dependencies required k8s.io/api v0.35.0 (commit 5aa70e7e9)
3. This upgrade broke controller-runtime v0.19.7 (bookmark timeout issues on k8s 1.35)
4. Had to remove the v0.19.7 replace directive and use v0.22.5 instead
5. controller-runtime v0.22.5 pulled in k8s.io/{api,apimachinery,client-go} v0.35.0
6. k8s.io/code-generator also upgraded to v0.35.0 to match
7. k8s.io/component-helpers was left at v0.34.3 initially (oversight)
8. Commit e3de5466b fixed the inconsistency by upgrading component-helpers to v0.35.0

Why it needed to be upgraded:
Having mixed k8s.io versions (some v0.34.3, most v0.35.0) causes:
- Potential dependency conflicts
- Inconsistent API behavior
- Build warnings from go mod
- Best practice is to keep all k8s.io/* dependencies at the same minor version
  
In summary: The upgrade from v0.34.3 to v0.35.0 was necessary to maintain consistency with the other k8s.io dependencies that were upgraded to v0.35.0 due to cosign v3 requirements for the KernelCache webhook functionality.

charts/kserve-kernelcache-crd-minimal/templates/serving.kserve.io_kernelcachenodes.yaml
Billy99: Works as designed. Ignoring.

The convention is to use CamelCase and hyphens in code and Kubelet adjusts. See current output and other CRD definitions.

$ kubectl get kernelcachenode -o wide
NAME           CACHES-IN-USE   CACHES-NOT-IN-USE   CACHES-ERROR   POD-RUNNING   POD-DELETING   AGE
...

charts/kserve-kernelcache-crd-minimal/templates/serving.kserve.io_kernelcaches.yaml
Billy99: Works as designed. Ignoring.

charts/kserve-kernelcache-crd-minimal/Chart.yaml
Billy99: Done.
Updated to v0.19.0-rc0 (not rc1) to match the rest of the code base.

config/configmap/inferenceservice.yaml
Billy99: I could go either way on this. Looking at the list, 4-5 items are over laps, 4-5 are feature specific.

- **enabled**: Need separate fields
- **jobNamespace**: Probable separate fields, but guess we could reuse the same namespace for both
- **defaultJobImage/extractImage:** Need separate fields
- fsGroup: Same
- jobTTLSecondsAfterFinished: Same
- reconcilationFrequencyInSecs: Same
- disableVolumeManagement: Same (not used by KC now, but could if KC adopts NodeGroups)
- **noGPU**: KC Only
- **enablePermissionInitContainer**: KC Only, but if LMC wants to run in KIND they could adopt it.

config/default/manager_image_patch.yaml
Billy99: Agree. Fixed.

config/kernelcachenodes/nodeselectors/gpu-amd.yaml
Billy99: These are just examples. In the first pass, there was a hardcoded Node Selector on the Kernel Manager Agent Daemonset:`kserve/kernelcache: worker`
This followed the LocalModelCache implementation. This was then updated to stray from LMC in an attempt to make it more configurable. By default, there is no Node Selector, so Agent
will run on every Node. A new `nodeselectors/` directory was added to`config/kernelcachenodes/`, which contains example Kustomize patches for both Node Selectors and Tolerations, along with a README.md explaining how to use them and why.

config/kernelcachenodes/nodeselectors/README.md
Billy99: I can go back to doing this if that is what the team wants, just an extra step for the user to add the label. I think the realistic deployment is the Agent needs to run on Nodes with GPUs. Other than NFD (Node Feature Discovery), I don't think there is a standard for labeling nodes with GPUs is there?

config/kernelcachenodes/manager.yaml
Billy99: This is intentional and matches the LocalModelNode agent pattern. Probes are only on the controller Deployment, not the agent DaemonSet. If you think the agents should have probes, they should probably be added to both LocalModelNode and KernelCacheNode agents. Just let me know.

config/kernelcachenodes/manager.yaml
Billy99: All KServe controllers use this pattern:

- Main controller: ko://github.com/kserve/kserve/cmd/manager
- LocalModel: ko://github.com/kserve/kserve/cmd/localmodel
- LocalModelNode: ko://github.com/kserve/kserve/cmd/localmodelnode
- KernelCache: ko://github.com/kserve/kserve/cmd/kernelcache
- KernelCacheNode: ko://github.com/kserve/kserve/cmd/kernelcachenode

`ko://` is a standard image reference format used by the ko build tool, and it's used consistently across all KServe controllers. The image gets replaced by the Kustomize patch during deployment.

config/kernelcachenodes/manager.yaml
Billy99: Done.
Changed to `registry.access.redhat.com/ubi9/ubi-micro:latest`. I think I originally busybox and was getting docker rate pull errors. Made this change globally across this feature.

config/overlays/test/configmap/inferenceservice.yaml
Billy99: Yes, the plan is to share the same controller. I plan to tackle that next. As far as the flag, it is not used. The LMC enable flag is not used either. Once the same controller is implemented, they will both be used.

config/rbac/kernelcache/role.yaml
Billy99: Local Model has the same list except for `delete`. PV/PVC delete is required for manual cleanup because PVC are cross-namespace and PV are cluster-scoped resources and can't use owner references to delete on these objects. LocalModel has DeletePV() and DeletePVC() functions but lacks the RBAC permission, so there may be a bug there.

With this investigation we found that we can remove `delete` from pods though. Pods associated with Jobs are automatically deleted after `jobTTLSecondsAfterFinished`, but if the KC is deleted before that expires the code was manually deleting.  When you delete a Job with `propagationPolicy: Background` (the default), Kubernetes does delete the Pods. So the manual Pod deletion is redundant. So `delete` from Pods was removed and the logic to delete a Pod associated with a Job was removed.

pkg/apis/serving/v1alpha1/kernel_cache_node_types.go
Billy99: Done.
No, nothing in this feature should be deprecated. Removed this and removed the usage of it in the test file.

pkg/apis/serving/v1alpha1/kernel_cache_status.go
Billy99: This is the number of pods that are both using the cache (have volumeMount) and that are Ready (Kubernetes Ready condition = True). However, upon further investigation, it looks like:
- CacheCounts.PodRunningCnt == ServingStatus.TotalPods
- CacheCounts.PodDeletingCnt == ServingStatus.TotalPodsTerminating
So these Pod counts in CacheCounts were deleted. So it now looks like:

```
$ kubectl get kc -n team-a test-cache -o yaml
:
status:
  counts:
    nodeCnt: 3
    nodeErrorCnt: 0
    nodeInUseCnt: 0
    nodeNotInUseCnt: 3
  gpuCompatibility:
    totalCompatibleGPUs: 0
    totalIncompatibleGPUs: 0
  resolvedDigest: sha256:660955164c52231b202ca95b1e454515834e612584a68ada99824eaaebc18ba3
  servingStatus:
    totalNamespaces: 0
    totalPods: 0
    totalPodsReady: 0
    totalPodsTerminating: 0
  state: Extracted
 ```

File: pkg/apis/serving/v1alpha1/kernel_cache_status.go (hunk 13 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_status.go]: No, deleted.

File: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go (hunk 25 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_webhook.go:43]: Done.
Deleted the first instance.

File: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go (hunk 8 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_webhook.go:49]: There are two Different Webhook Patterns in KServe:

**Pattern 1:** Type-based webhooks (KernelCache, TrainedModel, InferenceService)
  - Methods implemented directly on the CRD type
  - func (kc *KernelCache) ValidateCreate(...)
  - Registration: (&v1alpha1.KernelCache{}).SetupWebhookWithManager(mgr)
  - Location: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go

**Pattern 2:** Separate validator struct (LocalModelCache, LocalModelNamespaceCache, ServingRuntime)
  - Methods on separate *Validator struct
  - func (v *LocalModelCacheValidator) ValidateCreate(...)
  - Registration: ctrl.NewWebhookManagedBy(mgr).For(&v1alpha1.LocalModelCache{}).WithValidator(&localmodelwebhook.LocalModelCacheValidator{...})
  - Location: pkg/webhook/admission/localmodelcache/localmodelcache_validator.go

Why the Difference?

LocalModelCache validator needs Client - it queries other resources during validation. The separate validator struct holds the client:
  type LocalModelCacheValidator struct {
      client.Client  // Injected during registration
  }
  
KernelCache validation is self-contained - only validates the resource itself, no external lookups needed.

So KernelCache webhook location IS consistent - it follows the same pattern as TrainedModel and InferenceService (type-based webhooks in pkg/apis/). However, if you want consistency with LocalModelCache pattern, we could move it to pkg/webhook/admission/kernelcache/ with a separate validator struct. This would be:
  - More consistent with LocalModelCache
  - Better if we later need Client for cross-resource validation (but don't think that will be needed)
  - Follows the newer pattern (LocalModelCache added more recently)

---

Yes, I would like all constants in `constants.go` but moving these would create a circular dependency. APIs shouldn't reference controller files. I did search to make sure there were no other constants floating around and found `KernelCacheFinalizerName` defined in `pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go`, so moved it to `constants.go`

File: pkg/apis/serving/v1alpha1/llm_inference_service_conversion_test.go (hunk 12 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/llm_inference_service_conversion_test.go:927]: Done
No, added back in

File: pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go (hunk 7 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/llm_inference_service_conversion.go:91]: Done
No, added back in

File: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go (hunk 29 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_webhook.go]: Done - Removed
No, it was used at one point, then when Cosign support as added, the functionality moved to a cosign related function and this was not cleaned up. Removed now.

File: pkg/apis/serving/v1beta1/configmap.go (hunk 36 above)
Comment (Billy99) [reply pkg/apis/serving/v1beta1/configmap.go]: Done.

File: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go (hunk 39 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_webhook.go:631]: No, Kyverno enabled should NOT be a ConfigMap field. The current approach (environment variable) is correct because it has infrastructure dependency. Kyverno requires installing CRDs, webhooks, and ImageVerificationPolicy resources. So it is not runtime-configurable and you can't toggle Kyverno on/off dynamically; it must match what's actually installed in the cluster. So environment variable is appropriate - Set at pod deployment time to reflect the cluster's actual state.

File: pkg/apis/serving/v1alpha1/kernel_cache_webhook.go (hunk 40 above)
Comment (Billy99) [reply pkg/apis/serving/v1alpha1/kernel_cache_webhook.go]: Done.
Not needed. Changed to support only true and false.

File: pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go (hunk 31 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go]: Done
Yes, I didn't realize that API existed, I like that much better.

File: pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go (hunk 27 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go]: Done.
No it doesn't. And there was no logic reacting to a PV change. Removed the PV watch line.

File: pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go (hunk 28 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go:1136]: No, we don't need to filter PVCs. The Owns() mechanism already filters - it only triggers for PVCs that have this specific KernelCache as an owner reference. Since only the Serving PVC has an owner ref, only that PVC will trigger reconciliation. The Download PVC won't trigger because it has no owner ref (cross-namespace).

File: pkg/controller/v1alpha1/kernelcachenode/controller.go (hunk 43 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcachenode/controller.go:59]: Done
There was a call from main that made sure it existed, so the signalHandler would never be started, but added a check early in main() to fail fast if not set..

File: pkg/controller/v1alpha1/kernelcachenode/controller.go (hunk 42 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcachenode/controller.go:73]: Done.

File: pkg/controller/v1alpha1/kernelcachenode/controller.go (hunk 37 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcachenode/controller.go]: Done
No, removed it. When I was reworking code I thought I might still need it, but then forgot to deleted it once I moved on.

File: pkg/controller/v1alpha1/kernelcachenode/file_utils.go (hunk 38 above)
Comment (Billy99) [reply pkg/controller/v1alpha1/kernelcachenode/file_utils.go]: Done
Agree and changed to 0o755. It was a Cut and Paste from LMC, but looks like they need it for a different reason.

File: pkg/controller/v1alpha2/llmisvc/router.go (hunk 3 above)
Comment (Billy99) [reply pkg/controller/v1alpha2/llmisvc/router.go]: Done.
No, added it back.

File: pkg/controller/v1beta1/inferenceservice/components/explainer.go (hunk 4 above)
Comment (Billy99) [reply pkg/controller/v1beta1/inferenceservice/components/explainer.go]: I isolated these changes to one commit and have a pretty thorough commit description. See description associated with commit `Fix PodSpec conversion for k8s.io/api v0.35.0`. If you already saw that and still need more, just let me know.

File: pkg/cosign/verify.go (hunk 5 above)
Comment (Billy99) [reply pkg/cosign/verify.go:2]: Done

File: pkg/cosign/verify.go (hunk 24 above)
Comment (Billy99) [reply pkg/cosign/verify.go:130]: Agree. I have it marked down as a ToDo in the list of `Production Hardening` work items:
- Cosign Allow Policy - Configurable flag to allow or reject unsigned OCI Images
- Cosign Sign Policy - Need a way to configure what Cosign signatures are allow. Can't allow just any signature.

File: test/e2e/kernelcache/test_kernelcache_basic.py (hunk 35 above)
Comment (Billy99) [reply test/e2e/kernelcache/test_kernelcache_basic.py]: Done
Change all the sleeps to polling.

File: go.mod (hunk 23 above)
Comment (Billy99) [reply go.mod]: The bump was required by Cosign v3. If this bump causes heartburn, we could down grade to only support Cosign v2, but I would prefer to keep Cosign V3 support.

File: Makefile (hunk 11 above)
Comment (Billy99) [reply Makefile]: Done
Nothing, it never should have been there. Removed reference to my local sandbox from multiple places.

File: Makefile (hunk 9 above)
Comment (Billy99) [reply Makefile]: Removed all occurrences to my local directory.

Same controller as LMC is the next feature I will work on.

File: Makefile (hunk 10 above)
Comment (Billy99) [reply Makefile:742]: TL;DR: The secret prevents users from tampering with the resolved digest annotation. Only the mutating webhook can create valid digest annotations because only it can sign them with the secret key.

The KernelCache webhook has two phases:
1. Mutating webhook - Resolves spec.image (e.g., "repo/cache:tag") to its digest (e.g., "sha256:abc123...") and stores it in annotations
2. Validating webhook - Verifies the digest annotation was set correctly

Without the signature:
- A user could manually add the internal.serving.kserve.io/resolved-digest annotation with an arbitrary/malicious digest
- The validating webhook wouldn't know if the digest came from the trusted mutating webhook or from a user
- This could allow digest injection attacks (user claims image X but points to malicious digest Y)

How the HMAC signature prevents this:
1. Mutating webhook (line 174-186):
  - Resolves the image to get the real digest
  - Generates HMAC signature: HMAC-SHA256(secret, image + "|" + digest)
  - Stores both in annotations:
    - resolved-digest: sha256:abc123...
    - mutation-sig: \<base64-encoded-HMAC\>
2. Validating webhook (line 214-224):
  - Reads the digest and signature from annotations
  - Recomputes HMAC using the same secret
  - Verifies signature matches (constant-time comparison to prevent timing attacks)
  - Rejects if signature is missing or invalid

Why it's secure:
- Only the webhook has access to the MUTATION_SIGNING_KEY secret (injected via env var from k8s Secret)
- Users cannot generate valid signatures because they don't have the secret
- Even if a user adds resolved-digest annotation manually, it won't have a valid signature
- Uses HMAC-SHA256 (cryptographically secure) with constant-time comparison

File: pkg/apis/serving/v1beta1/configmap.go (hunk 34 above)
Comment (Billy99) [reply pkg/apis/serving/v1beta1/configmap.go:178]: Done.
Yes, it was missing. So were all the KC CRDs as well. Added now.

**Review by Billy99 — COMMENTED** [review]
Thank you for the thorough review. All comments have been addressed or commented on.

File: pkg/controller/v1beta1/inferenceservice/components/explainer.go (hunk 4 above)
Comment (Billy99) [reply pkg/controller/v1beta1/inferenceservice/components/explainer.go]: I ended up removing this. Upstream moved to `k8s.io/api v0.35.0` with a `replace k8s.io/api => k8s.io/api v0.34.5`, which satisfied the Cosign requirement. So the commit containing these PodSec conversion was deleted.

File: charts/kserve-crd/templates/serving.kserve.io_clusterservingruntimes.yaml (hunk 15 above)
Comment (Billy99) [reply charts/kserve-crd/templates/serving.kserve.io_clusterservingruntimes.yaml]: I ended up removing this. Upstream moved to k8s.io/api v0.35.0 with a replace k8s.io/api => k8s.io/api v0.34.5, which satisfied the Cosign requirement. So the commit containing these `userAnnotations` fields was deleted.

File: go.mod (hunk 23 above)
Comment (Billy99) [reply go.mod]: Upstream moved to k8s.io/api v0.35.0 with a replace k8s.io/api => k8s.io/api v0.34.5, which satisfied the Cosign requirement. So this matches upstream now.

**Comment by vishwagandhi1610** [comment]
I was able to test the changes in this PR and it works as described. the extraction job and pvc is created in kserve namespace.

**Comment by Jooho** [comment]
@Billy99 Is this pr ready for review?

File: pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go [hunk 44]
```diff
@@ -0,0 +1,1259 @@
...
+	}
+
+	// Extract unique node names from running agent pods
+	agentNodes := make(map[string]bool)
+	for _, pod := range agentPods.Items {
+		if pod.Spec.NodeName != "" {
+			agentNodes[pod.Spec.NodeName] = true
+		}
+	}
+
```
Jooho, line 367: reconcilePVC lists agent pods and builds agentNodes  but neither is used afterward. Status aggregation reads KernelCacheNodes directly in updateAggregateStatus. This block does an unnecessary Pod List RPC every PVC reconcile.

I think this is dead code that was for ImageVolume.  Do we still need this?

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,802 @@
...
+	got, err := base64.StdEncoding.DecodeString(sigB64)
+	if err != nil {
+		return false
+	}
+	return hmac.Equal(want, got)
+}
+
+// isExtractionStarted checks if extraction has started on any node
+// Extraction is considered started if any node is in InProgress, Extracted, or Running state
+func isExtractionStarted(kc *KernelCache) bool {
```
Jooho, line 756: `isExtractionStarted` misses Downloading state (NodeCnt>0 but both InUse/NotInUse counts are 0), so storage fields stay mutable while the extraction Job runs.  `spec` can diverge from the already-created PVC. Gate on NodeCnt>0 (or State != Pending) instead.

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1259 @@
...
+	// Default access mode: ReadWriteMany for Phase 1 (multi-node sharing)
+	accessModes := kc.Spec.AccessModes
+	if len(accessModes) == 0 {
+		accessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteMany}
+	}
+
+	// Step 1: Create PV (for KIND - no dynamic provisioner)
+	// Use labels for tracking instead of owner references (PV is cluster-scoped, KernelCache is namespace-scoped)
+	pv := corev1.PersistentVolume{
```
Jooho, line 480: Sibling controller LocalModel also creates PVs but takes the PV spec from the user (NodeGroup.Spec.PersistentVolumeSpec) instead of hardcoding it. KernelCache hardcodes hostPath, baking the cluster's storage type into the controller. Creating a PV is fine (StorageClass isn't required for static binding) but the spec must come from the user/CRD like LocalModel does, so NFS/CSI/hostPath is the admin's choice, not the controller's.

pkg/apis/serving/v1alpha1/kernel_cache_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// +kubebuilder:validation:Required
+	Image string `json:"image"`
+
+	// MountType specifies how to mount the cache (pvc or imageVolume)
+	// +kubebuilder:validation:Enum=pvc;imageVolume
+	// +kubebuilder:default=pvc
+	// +optional
+	MountType KernelCacheMountType `json:"mountType,omitempty"`
+
+	// StorageClassName for PV/PVC (only used when mountType=pvc)
```
Jooho, line 60: These flat storage fields (StorageClassName/StorageSize/AccessModes) can't express the PV source, forcing the controller to hardcode hostPath. Sibling LocalModel exposes the full corev1.PersistentVolumeSpec (NodeGroup.Spec. PersistentVolumeSpec) so the admin picks hostPath/NFS/CSI. KernelCache should follow the same pattern instead of flat fields + hardcoded hostPath.

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1259 @@
...
+			"kernelcache.kserve.io/namespace": kc.Namespace,
+		}
+
+		// Delete Jobs (Pods auto-delete via propagation policy)
+		jobList := &batchv1.JobList{}
+		if err := r.List(ctx, jobList,
+			client.InNamespace(jobNamespace),
+			client.MatchingLabels(jobLabels),
+		); err == nil {
+			propagationPolicy := metav1.DeletePropagationBackground
```
Jooho, line 201: The Job is deleted with Background propagation, and then the Download PVC is deleted immediately (:213), without waiting for the Job's pod to terminate. If the pod is still using the PVC, the PVC can remain stuck in Terminating. We should use Foreground propagation and requeue until the Job and pod are gone before deleting the PVC. Also, deleting the hostPath PV doesn't remove the data under /kernel-caches on the node. We should confirm that the agent cleans it up.

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1259 @@
...
+			if err := r.Delete(ctx, servingPV); err != nil && !errors.IsNotFound(err) {
+				r.Log.Error(err, "Failed to delete Serving PV", "pv", servingPVName)
+				return ctrl.Result{}, err
+			}
+			r.Log.Info("Deleted Serving PV", "pv", servingPVName)
+		}
+	}
+
+	// Remove finalizer (both PVC and ImageVolume modes)
+	controllerutil.RemoveFinalizer(kc, kernelcachecommon.KernelCacheFinalizerName)
```
Jooho, line 264: The operator removes the finalizer after deleting the PV/PVC objects, but the actual /kernel-caches/<storageKey> directories are local to each node and can only be removed by the agent through deleteCaches/orphan cleanup.

There’s currently no coordination between these two. Once the finalizer is removed, the KernelCache is gone, so if the agent hasn’t removed the cache from CacheStatus yet—or if the agent is down or missing on a node the data can be left behind with nothing to trigger the cleanup.

Should we make sure the data has been removed before the finalizer is removed?

pkg/controller/v1alpha1/kernelcache/reconcilers/kernelcache_reconciler.go
```diff
@@ -0,0 +1,1259 @@
...
+			return nil
+		} else if !errors.IsNotFound(err) {
+			return err
+		}
+
+		return r.createExtractionJob(ctx, kc, pvcName, config)
+	} else if r.jobFailed(job) {
+		// Only recreate if failed job is old enough
+		age := time.Since(job.CreationTimestamp.Time)
+		if age > 5*time.Minute {
```
Jooho, line 607: Retrying a failed Job calls createExtractionJob, which creates the Job with the deterministic name {ns}-{name}-extract. Since the failed Job still exists (the default TTL is 1h), the create call returns AlreadyExists, causing every reconcile to fail. As a result, extraction can't retry for ~55 minutes despite the 5-minute retry gate.

Delete the old failed Job first using Foreground propagation, then recreate it (or requeue so it goes through the job == nil path).

Jooho (commented):
I've only reviewed the PV/PVC part so far, and I already have ~7 findings. A few of them are more about the overall design than individual lines—for example, the controller hardcodes a hostPath PV source, which means the storage type is effectively baked into the implementation rather than being configurable by the admin, as it is with LocalModel via PersistentVolumeSpec. There are also some concerns around RWX + hostPath not really supporting sharing across nodes, and around coordinating cleanup between the operator finalizer and the agent's on-disk cleanup.

There are still ~6 larger areas to look through (webhook/digest verification, KernelCacheNode agent, GPU detection, capture pipeline + sidecar injector, pod webhook mounting, and validation).

Given that, I wanted to check in on how you'd prefer to approach the review. A full review will probably result in quite a few comments—potentially 50+—and some of the issues seem to come from a handful of broader "Phase 1" design decisions that may still be evolving.

Would you prefer that I:

Continue with a full deep review of the remaining areas? I'm happy to do that, though it'll be fairly extensive and take some time.
Consider splitting the PR into smaller pieces that can be reviewed and merged independently? For example: API types, cosign, KernelCache controller behind the feature flag, agent, capture, and finally the shared-path webhook/validation changes.

My inclination is toward option 2, mainly because I worry that doing a detailed line-by-line review of the whole PR right now may result in a lot of feedback on code that's likely to change as some of these design questions get settled. Breaking it up could make the review a bit easier for everyone and also give us smaller, safer pieces to merge and roll back if needed.

That said, I don't want to slow things down unnecessarily. If the goal is to keep this as a single PR, I'm happy to continue the review in that direction as well.

Billy99:
@Jooho I will do whatever you prefer on getting the PR in. If you want the PR broken up, that might take some time, but we can go that route.

I do have the ImageVolume working. Do we plan to support both PVC and ImageVolume or just move to the ImageVolume? Design decision that may effect how you review the PR.

pkg/apis/serving/v1alpha1/kernel_cache_webhook.go
```diff
@@ -0,0 +1,803 @@
...
+// SetupWebhookWithManager registers the webhook with the controller-runtime manager
+func (kc *KernelCache) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(kc).
+		WithDefaulter(kc).
+		WithValidator(kc).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/mutate-serving-kserve-io-v1alpha1-kernelcache,mutating=true,failurePolicy=fail,sideEffects=None,groups=serving.kserve.io,resources=kernelcaches,verbs=create;update,versions=v1alpha1,name=kernelcache.kserve-webhook-server.defaulter,admissionReviewVersions=v1,reinvocationPolicy=Never
```
vishwagandhi1610, line 115: Shouldn't reinvocationPolicy be IfNeeded?

config/webhook/kernelcache/manifests.yaml
```diff
@@ -0,0 +1,55 @@
...
+webhooks:
+  - clientConfig:
+      service:
+        name: $(webhookServiceName)
+        namespace: $(kserveNamespace)
+        path: /mutate-serving-kserve-io-v1alpha1-kernelcache
+    failurePolicy: Fail
+    name: kernelcache.kserve-webhook-server.defaulter
+    sideEffects: None
+    reinvocationPolicy: Never
```
vishwagandhi1610, line 16: Shouldn't reinvocationPolicy be IfNeeded?

PR #5602, opened 2026-05-29
fix(llmisvc): fix probe thresholds, EPP grace period...

andresllh:
## Summary
... and rolling upgrade coordination.

Three inference-aware lifecycle gaps identified during review:

- **Probe thresholds were inverted** across all vLLM config templates: `readinessProbe` had `failureThreshold: 60` (10-minute detection window) while `livenessProbe` had `failureThreshold: 3` (restart after 30s, before traffic even stops). Fixed to `readiness: periodSeconds=1 / failureThreshold=1` and `liveness: failureThreshold=10`. `startupProbe` is unchanged.
- **EPP `terminationGracePeriodSeconds` was 30s**, less than the workload pod TGPS of 60s. The EPP is on the critical path for every response chunk (ext_proc `FULL_DUPLEX_STREAMED`); killing it before in-flight streams complete truncates responses. Bumped to 60s with a comment documenting the invariant.
- **Workload and EPP Deployments updated concurrently** by the reconciler. Added `isWorkloadRolling()` which checks the `DeploymentAvailable` condition (same definition used by `propagateDeploymentStatus`) and defers the EPP spec update until all workload Deployments are fully available.

## Test plan

- [ ] Unit tests added for `isWorkloadRolling` covering: fully available, `Available=False`, no condition yet, prefill mid-rollout, both ready, deployment not found
- [ ] `make precommit` passes
- [ ] E2e test `test_rolling_upgrade_coordination` triggers a rolling update via pod-template annotation patch and verifies the service recovers to a healthy, inference-serving state

### DIFF
--- a/config/llmisvcconfig/config-llm-template.yaml
+++ b/config/llmisvcconfig/config-llm-template.yaml
@@ -195,16 +195,16 @@ spec:
             port: 8000
             scheme: '{{ if .GlobalConfig.EnableTLS }}HTTPS{{else}}HTTP{{- end }}'
           periodSeconds: 10
-          timeoutSeconds: 10
-          failureThreshold: 3
+          timeoutSeconds: 1
+          failureThreshold: 10
         readinessProbe:
           httpGet:
             path: /health
             port: 8000
             scheme: '{{ if .GlobalConfig.EnableTLS }}HTTPS{{else}}HTTP{{- end }}'
-          periodSeconds: 10
-          timeoutSeconds: 5
-          failureThreshold: 60
+          periodSeconds: 1
+          timeoutSeconds: 1
+          failureThreshold: 1
         startupProbe:
           httpGet:
             path: /health
--- a/pkg/controller/v1alpha2/llmisvc/scheduler.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler.go
@@ -192,11 +192,12 @@ func (r *LLMISVCReconciler) reconcileSchedulerDeployment(ctx context.Context, ll
 	return r.propagateSchedulerDeploymentStatus(ctx, scheduler, llmSvc.MarkSchedulerWorkloadReady, llmSvc.MarkSchedulerWorkloadNotReady)
 }
 
-// isWorkloadRolling returns true if any workload Deployment for the LLMInferenceService is not
-// yet fully available. It uses the same readiness definition as propagateDeploymentStatus: the
-// workload is ready only when the DeploymentAvailable condition is True. This is used to defer
-// EPP spec updates until all workload Deployments are fully available, preventing an EPP restart
-// from truncating in-flight streams while vLLM pods are still coming up.
+// isWorkloadRolling returns true if any workload Deployment for the LLMInferenceService is
+// explicitly unavailable (DeploymentAvailable condition present and not True). Deployments with
+// no Available condition (brand-new or not yet observed by the controller, e.g. in envtest) are
+// treated as not rolling so that initial EPP creation is not blocked. This is used to defer EPP
+// spec updates during active workload rollouts, preventing an EPP restart from truncating
+// in-flight streams while vLLM pods are still coming up.
 func (r *LLMISVCReconciler) isWorkloadRolling(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) (bool, error) {
 	names := []string{mainDeploymentName(llmSvc)}
 	if llmSvc.Spec.Prefill != nil {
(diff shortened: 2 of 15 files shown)


andresllh:
cc @pierDipi

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -176,12 +177,52 @@ func (r *LLMISVCReconciler) reconcileSchedulerDeployment(ctx context.Context, ll
...
+	// Defer EPP spec updates until all workload pods have finished rolling out.
+	// The EPP is on the critical path for every response chunk; restarting it mid-stream
+	// truncates in-flight streaming responses on the client side.
+	if rolling, err := r.isWorkloadRolling(ctx, llmSvc); err != nil {
+		return err
+	} else if rolling {
+		logger.Info("Workload rollout in progress, deferring EPP deployment update")
+		return r.propagateSchedulerDeploymentStatus(ctx, scheduler, llmSvc.MarkSchedulerWorkloadReady, llmSvc.MarkSchedulerWorkloadNotReady)
+	}
 	if err := Reconcile(ctx, r, llmSvc, &appsv1.Deployment{}, scheduler, semanticDeploymentIsEqual, PreserveDeploymentReplicas()); err != nil {
```
KillianGolds, line 189: Minor, non-blocking thought on the `else if rolling` defer branch.

When `isWorkloadRolling` is true we return `propagateSchedulerDeploymentStatus(...)`, which assumes the EPP `Deployment` already exists. In the normal first-create flow it does (reconcile ordering means the workload has no conditions yet, so we don't defer here), so this is fine in practice.

The only edge I could think of is if the EPP is ever missing *while* a workload is mid-rollout (manual delete, or GC during a topology transition). The defer branch would then return `NotFound`, and since it runs before the Service / InferencePool steps, those get skipped until the workload becomes `Available`.

Might be worth only deferring when the EPP already exists, so an absent EPP still gets created:

```go
if rolling, err := r.isWorkloadRolling(ctx, llmSvc); err != nil {
    return err
} else if rolling && eppExists {
    // defer update
}
```

Could easily be a non-issue, just flagging in case it's a cheap guard to add.

config/llmisvcconfig/config-llm-template.yaml
```diff
@@ -195,16 +195,16 @@ spec:
...
           httpGet:
             path: /health
             port: 8000
             scheme: '{{ if .GlobalConfig.EnableTLS }}HTTPS{{else}}HTTP{{- end }}'
-          periodSeconds: 10
-          timeoutSeconds: 5
-          failureThreshold: 60
+          periodSeconds: 1
+          timeoutSeconds: 1
+          failureThreshold: 1
```
KillianGolds, line 207: With `periodSeconds: 1` / `timeoutSeconds: 1` / `failureThreshold: 1`, a single transient `/health` slower than 1s would pull the pod from the `InferencePool`, which I could imagine happening under heavy decode load when the event loop is briefly saturated.

`failureThreshold: 2` would still land within the "1-2 second" detection target from the AC while tolerating one slow probe:

```yaml
readinessProbe:
  periodSeconds: 1
  timeoutSeconds: 1
  failureThreshold: 2   # ~2s detection, survives a single slow /health
```

Same applies to the other five templates (`config-llm-decode-template.yaml:207`, `config-llm-prefill-template.yaml:208`, `config-llm-worker-data-parallel.yaml:243`, `config-llm-decode-worker-data-parallel.yaml:302`, `config-llm-prefill-worker-data-parallel.yaml:244`). Not sure it matters in practice for vLLM's `/health`, so purely a suggestion.

KillianGolds (commented):
Overall this looks solid, readiness now detects an unhealthy engine in ~1s instead of the old 10-minute window, EPP `terminationGracePeriodSeconds` is bumped to 60 so it outlives in-flight streams, and `isWorkloadRolling` defers EPP updates until the single-node workload finishes rolling out. The unit coverage on `isWorkloadRolling` and the e2e rollout test are nice additions.

I left two small, non-blocking notes (defer path when the EPP is absent, and an optional failureThreshold tweak for a bit more tolerance). Both are take-it-or-leave-it and shouldn't hold up the merge.

Nice work @andresllh!

/lgtm

pkg/controller/v1alpha2/llmisvc/scheduler.go
cjohannsen-cloudera: @andresllh please create a follow-up issue for this condition and submit the pr for the fix separately

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -176,12 +177,52 @@ func (r *LLMISVCReconciler) reconcileSchedulerDeployment(ctx context.Context, ll
 		}
 		return Delete(ctx, r, llmSvc, scheduler)
 	}
+	// Defer EPP spec updates until all workload pods have finished rolling out.
+	// The EPP is on the critical path for every response chunk; restarting it mid-stream
+	// truncates in-flight streaming responses on the client side.
+	if rolling, err := r.isWorkloadRolling(ctx, llmSvc); err != nil {
+		return err
```
pierDipi, line 184: if the scheduler is misconfigured and it's crash looping, this prevents the update and it will never recover.

it seems like we're trying to fix a problem that should be fixed somewhere else (in the scheduler to graceful shutdown) and increasing termination grace period further.

PR #5606, opened 2026-05-30
feat(apis): validate pathTemplate format in IngressConfig

lbcristaldo:
Implements the TODO in pkg/apis/serving/v1beta1/configmap.go:308 to validate that the pathTemplate ingress configuration generates a path that starts with / and uses both .Name and .Namespace to avoid route collisions. Returns an error when the template does not meet these criteria, consistent with the existing validation pattern in the function. Path without / already fails at runtime in GenerateUrlPath(), so this does not break any working configuration.

Which issue(s) this PR fixes: N/A
Feature/Issue validation/testing:

- [x] go test ./pkg/apis/serving/v1beta1/... all existing tests pass
- [x]  go test -race ./pkg/apis/serving/v1beta1/... no race conditions detected
- [x]  go vet ./pkg/apis/serving/v1beta1/... clean
- [x]  make generate && make manifests: no changes generated

Notes: 
- make precommit was attempted but failed during the Python venv setup step (ruff installation). The Go-specific checks (go vet, go test -race, make generate, make manifests) were run manually and all passed. No Python files were modified in this PR.

Checklist:
- [x]  Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [x]  Has code been commented, particularly in hard-to-understand areas?
- [ ]  Have you made corresponding changes to the documentation?

```
release note:
Validate pathTemplate in IngressConfig to return an error when the generated path does not start with '/' or does not use both .Name and .Namespace to avoid collisions.
```

### DIFF
--- a/pkg/apis/serving/v1beta1/configmap.go
+++ b/pkg/apis/serving/v1beta1/configmap.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 	"text/template"
 
@@ -311,16 +312,30 @@ func NewIngressConfig(isvcConfigMap *corev1.ConfigMap) (*IngressConfig, error) {
 			return nil, errors.New("invalid ingress config - ingressGateway is required")
 		}
 		if ingressConfig.PathTemplate != "" {
-			// TODO: ensure that the generated path is valid, that is:
-			// * both Name and Namespace are used to avoid collisions
-			// * starts with a /
-			// For now simply check that this is a valid template.
-			_, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
+			tmpl, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
 			if err != nil {
 				return nil, fmt.Errorf("invalid ingress config, unable to parse pathTemplate: %w", err)
 			}
 			if ingressConfig.IngressDomain == "" {
 				return nil, errors.New("invalid ingress config - ingressDomain is required if pathTemplate is given")
+			}
+			// Blocking validation: check template generates valid path format
+			// Validate that the generated path starts with '/' and uses both .Name and .Namespace
+			// to avoid route collisions. Returns an error if the path is invalid.
+			var buf strings.Builder
+			testParams := struct{ Name, Namespace string }{
+				Name: "test-service", Namespace: "test-namespace",
+			}
+			if err := tmpl.Execute(&buf, testParams); err != nil {
+				return nil, fmt.Errorf("invalid pathTemplate - execution failed: %w", err)
+			}
+			// Check if path starts with /
+			if _, err := url.ParseRequestURI(buf.String()); err != nil {
+				return nil, fmt.Errorf("invalid pathTemplate - generated path %q is not a valid URI path: %w", buf.String(), err)
+			}
+			// Check if template uses Name/Namespace to avoid collisions
+			if !strings.Contains(ingressConfig.PathTemplate, "{{ .Name }}") || !strings.Contains(ingressConfig.PathTemplate, "{{ .Namespace }}") {
+				return nil, errors.New("invalid pathTemplate - must include both {{ .Name }} and {{ .Namespace }} to avoid route collisions")
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/apis/serving/v1beta1/configmap.go
vivekk16: Please remove the TODO block or update it to describe the current behavior.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -315,13 +316,39 @@ func NewIngressConfig(isvcConfigMap *corev1.ConfigMap) (*IngressConfig, error) {
...
+				validated = false
+			}
+
+			// Check if template uses Name/Namespace to avoid collisions
+			if !strings.Contains(ingressConfig.PathTemplate, ".Name") || !strings.Contains(ingressConfig.PathTemplate, ".Namespace") {
+				validated = false
+			}
+
+			if !validated {
+				klog.Infof("PathTemplate %q may have collisions or wrong format (generated: %q), consider using .Name and .Namespace and starting with /",
```
vivekk16: Every other validation in this function returns an `error`, not a log warning. Using `klog.Infof` breaks that pattern and makes the issue invisible to callers. A path without `/` already fails at runtime in `GenerateUrlPath()`, so returning an error here wouldn't break any working config.

pkg/apis/serving/v1beta1/configmap_test.go
```diff
@@ -724,4 +724,68 @@ func TestNewIngressConfig_Validation(t *testing.T) {
...
+					"ingressGateway": "knative-serving/knative-ingress-gateway",
+					"ingressDomain": "mydomain.com",
+					"pathTemplate": "/{{ .Name }}/{{ .Namespace }}/inference"
+				}`,
+			},
+		}
+		cfg, err := NewIngressConfig(cm)
+		g.Expect(err).ShouldNot(gomega.HaveOccurred())
+		g.Expect(cfg).ShouldNot(gomega.BeNil())
+	})
```
vivekk16, line 793: These tests only assert err == nil, they'd still pass if the validation logic were deleted. Please add assertions that verify the expected behavior for the invalid cases vs the valid case.

pkg/apis/serving/v1beta1/configmap.go
```diff
@@ -315,13 +316,39 @@ func NewIngressConfig(isvcConfigMap *corev1.ConfigMap) (*IngressConfig, error) {
...
 			// For now simply check that this is a valid template.
-			_, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
+			tmpl, err := template.New("path-template").Parse(ingressConfig.PathTemplate)
 			if err != nil {
 				return nil, fmt.Errorf("invalid ingress config, unable to parse pathTemplate: %w", err)
 			}
 			if ingressConfig.IngressDomain == "" {
 				return nil, errors.New("invalid ingress config - ingressDomain is required if pathTemplate is given")
 			}
+			// Non-blocking validation: check template generates valid path format
```
vivekk16: `GenerateUrlPath()` in `pkg/controller/v1beta1/inferenceservice/reconcilers/ingress/path.go` already validates all of this, correct struct (`PathTemplateValues` with only `Name`/`Namespace`), template parse + execute, and `url.ParseRequestURI` (rejects paths without `/`). It can't be called directly from here due to circular imports, but consider replicating the same approach:

```go
testParams := struct{ Name, Namespace string }{
    Name: "test-name", Namespace: "test-namespace",
}
if err := tmpl.Execute(&buf, testParams); err != nil {
    return nil, fmt.Errorf("invalid pathTemplate - execution failed: %w", err)
}
if _, err := url.ParseRequestURI(buf.String()); err != nil {
    return nil, fmt.Errorf("invalid pathTemplate - generated path %q is not a valid URI path: %w", buf.String(), err)
}
```

This also fixes two bugs in the current approach:
1. The test struct here has `IngressDomain` which the runtime struct doesn't, templates using `{{ .IngressDomain }}` pass validation but fail at runtime.
2. `strings.Contains(template, ".Name")` is always true when `.Namespace` is present since `"Namespace"` starts with `"Name"`, making the collision check dead code.

If you still want the `.Name`/`.Namespace` collision check, match the full action syntax (`"{{ .Name }}"`) instead of bare `".Name"`.

lbcristaldo: Done: removed the TODO comment, changed the warning to an error consistent with the rest of the function, used url.ParseRequestURI for path validation, fixed the test struct to remove IngressDomain, matched the {{ .Name }} / {{ .Namespace }} syntax exactly, and updated the tests to assert on the specific error messages. Thank you!

vivekk16 (commented):
lgtm
cc @spolti @Jooho

PR #5541, opened 2026-05-15
fix: prevent Value/ValueFrom conflict when merging ClusterStorageContainer

richiefrost:
**What this PR does / why we need it**:

The problem:
When using a ClusterStorageContainer with environment variables, these can conflict when there are envs already injected into the existing `storage-initializer` initContainer.  

1. **Unmarshalling JSON of a modified object into an existing struct** - previously, the approach was to capture the container's state, perform a strategic JSON merge, and unmarshal that JSON back into the original state. However, unmarshalling into the same struct's env slices after merging unintentionally writes ClusterStorageContainer envs to the same slots that were in the original state. For example, if an env with name `AWS` was already in slot 0 set with `value`, and the CSC override added an env named 'CUSTOM' to the container set with `valueFrom`, it'd get written to slot 0 as well when unmarshalling. But since the `value` part of the overwritten entry didn't get cleared from the previous entry in JSON unmarshal, the final result would look like `'CUSTOM': {'value': 'foo', 'valueFrom': 'bar'}` when it should be `'CUSTOM': {'valueFrom': 'bar'}` to pass admission.

2. **Same-name field-wise merge.** Strategic merge patch will populate both `value` and `valueFrom` entries since they're keyed by field name. So if the ClusterStorageContainer manually overrides known envs like `AWS_ACCESS_KEY_ID` and provides them via `value` while the webhook may inject these envs via `valueFrom`, you'll get rejected by admission for that pod's spec.

This solution:
StrategicMergePatch with the CSC overrides as usual. Then, unmarshal into a fresh container struct to ensure no stale fields are left behind. Finally, reconcile environment variables to ensure there are no conflicts, favoring the values defined in the ClusterStorageContainer if there are duplicate definitions.

**Which issue(s) this PR fixes**:
Fixes #5516

**Feature/Issue validation/testing**:

Added `TestMergeContainerSpecs_EnvHandling` — a table-driven test with three subtests covering both bugs:

- [x] `valueFrom_overrides_value_on_same_name`: credential env injected as literal `Value` is correctly replaced by a CSC `valueFrom` reference on the same name; merged entry has only `valueFrom`.
- [x] `value_overrides_valueFrom_on_same_name`: symmetric case; merged entry has only `Value`.
[...]

PR #5614, opened 2026-06-02
feat(llmisvc): merge user storage-initializer customizations

### LINKED ISSUE #5480: Enable customization of storage-initializer configuration for LLMInferenceServices

jlost:

When users define a `storage-initializer` init container in `spec.template.initContainers`
or via LLMInferenceServiceConfig baseRefs, the controller strips it and appends a fresh
default — discarding user customizations (env vars, resources, image, volume mounts).
This causes duplicate container names that Kubernetes rejects.

This PR replaces the strip+append pattern with extract+merge: the controller extracts
the user's storage-initializer before stripping, creates the default, then merges user
customizations on top using `strategicpatch.StrategicMergePatch`. Controller-owned fields
(`Name`, `Args`) are always preserved.

**Usage example — passing env vars to the storage-initializer**:

To tune the storage-initializer runtime (e.g. disable HuggingFace xet high-performance
mode or limit Tokio worker threads), add a `storage-initializer` init container with
the desired env vars in `spec.template.initContainers`:

```yaml
kind: LLMInferenceService
metadata:
  name: my-llm
spec:
  model:
    name: my-model
    uri: hf://org/model-name
  template:
    initContainers:
      - name: storage-initializer
        env:
          - name: HF_XET_HIGH_PERFORMANCE
            value: "0"
          - name: TOKIO_WORKER_THREADS
            value: "1"
```

The controller merges these env vars into the default storage-initializer — the user's
values are preserved alongside controller-managed fields (args, name, volume mounts).

Fixes #5480

- [x] Unit tests for `extractAndStripStorageInitializer` (present, absent, nil cases)
- [x] Unit tests for `mergeStorageInitializerContainer` (resource override, image override, env var add/override, volume mount add, controller name/args preservation)
- [x] Unit test with real-world env vars (`HF_XET_HIGH_PERFORMANCE=0`, `TOKIO_WORKER_THREADS=1`) validating merge preserves them alongside controller defaults
- [x] Integration test (envtest) creating an LLMInferenceService with custom env vars on the storage-initializer and asserting they appear in the resulting Deployment after full reconciliation

```
go test ./pkg/controller/v1alpha2/llmisvc/... -count=1 -v
```

- The merge uses the same `strategicpatch.StrategicMergePatch` approach already used in `config_merge.go:mergeSpecs()`
- Image precedence is preserved: system default → current deployment image (upgrade stability) → user override
- `AddModelMount` is already idempotent (checks by VolumeName), so post-merge volume mounts don't duplicate

**Checklist**:

```release-note
Enable customization of storage-initializer init container for LLMInferenceServices. User-provided env vars, resources, image, and volume mounts from spec.template.initContainers or baseRefs are now merged with controller defaults instead of being discarded.
```


pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -51,21 +53,54 @@ var tokenizerOnlyDownload = corev1.EnvVar{
...
+		keptInit = append(keptInit, podSpec.InitContainers[i])
 	}
 	podSpec.InitContainers = keptInit
+	return extracted
+}
+
+// mergeStorageInitializerContainer merges user customizations onto the controller-generated
+// default storage-initializer container using strategic merge patch. The controller's Name
+// and Args are always preserved regardless of user overrides.
+func mergeStorageInitializerContainer(controllerDefault *corev1.Container, userOverride corev1.Container) (*corev1.Container, error) {
```
bartoszmajsak: We probably want `Value/ValueFrom` merged too - like we did in #5541

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -51,21 +53,54 @@ var tokenizerOnlyDownload = corev1.EnvVar{
...
+		return nil, fmt.Errorf("could not apply strategic merge patch: %w", err)
+	}
+
+	var merged corev1.Container
+	if err := json.Unmarshal(mergedJSON, &merged); err != nil {
+		return nil, fmt.Errorf("could not unmarshal merged container: %w", err)
+	}
+
+	merged.Name = controllerDefault.Name
+	merged.Args = controllerDefault.Args
```
bartoszmajsak: Do we want to warn user that for some reason set their own `Args`?

I also wonder if we want to protect `Command` too?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
bartoszmajsak: It also seems we have the logic in three different places,

- https://github.com/kserve/kserve/blob/d249eeff250ab25aacad5f8a324146dbd85f49f3/pkg/webhook/admission/pod/storage_initializer_injector.go#L775-L836
- https://github.com/kserve/kserve/blob/637ddfc999f0b4c0fe31e7ea770afd12b558299b/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L252-L286

maybe this PR is an opportunity to consolidate?

bartoszmajsak (commented):
A few minor suggestions.

I also wonder if we can add some helpful `godoc` around the `Template` spec, so the users will know what is happening under the hood:

```go
// Template for the main pod spec.
// ...existing lines...
//
// For the storage-initializer init container (named "storage-initializer"), users may
// customize fields such as env, resources, volumeMounts, and image. The container's
// name and args are controller-managed and any user-provided values for these fields
// will be overridden. To disable the storage-initializer entirely, use
// spec.storageInitializer.enabled=false.
//
// +optional
```

jlost:
Good catches! Addressed:
* Extracted the common container patch logic to a common helper that guards against the duplicate Value+ValueFrom case. Consolidated its use across 3 sites.
* Protected both Args and Command, adding an event if the user had attempted to specify an override.
* Added API comment.

@bartoszmajsak

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -259,7 +266,7 @@ func (r *LLMISVCReconciler) attachPVCModelArtifact(modelUri string, podSpec *cor
 //
 //	An error if the configuration fails, otherwise nil.
 func (r *LLMISVCReconciler) attachS3ModelArtifact(ctx context.Context, serviceAccount *corev1.ServiceAccount, llmSvc *v1alpha2.LLMInferenceService, modelUri string, curr corev1.PodSpec, podSpec *corev1.PodSpec, storageConfig *kserveTypes.StorageInitializerConfig, credentialConfig *credentials.CredentialConfig, containerName string, modelPath string) error {
-	if err := r.attachStorageInitializer(modelUri, curr, podSpec, storageConfig, llmSvc.Spec.Model.Confidential, containerName, modelPath); err != nil {
+	if err := r.attachStorageInitializer(llmSvc, modelUri, curr, podSpec, storageConfig, llmSvc.Spec.Model.Confidential, containerName, modelPath); err != nil {
```
bartoszmajsak: That's quite a few params and I think some of them are derived from `llmSvc` - maybe opportunity to make it simpler instead of adding new param?

jlost: Pruned the 'confidential' param

Jooho (approved):

PR #5615, opened 2026-06-02
feat: added power support for lgbserver component

PrernaKumbhar20:
**What this PR does / why we need it:**  We have updated the existing Dockerfile and added a new workflow for **lgbserver** to add Power support.

**Feature/Issue validation/testing:**

- Logs

`docker run -it --rm \
  -p 8080:8080 \
  -v /tmp/lgb-test-model:/models \
  localhost/lgb02:latest \
  --model_name lgb-iris \
  --model_dir /models
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/prod_venv/lib/python3.11/site-packages/lightgbm/basic.py:3714: UserWarning: Ignoring params argument, using parameters from model file.
  _log_warning("Ignoring params argument, using parameters from model file.")
2026-06-02 09:02:07.019 1 kserve INFO [model_server.py:register_model():439] Registering model: lgb-iris
2026-06-02 09:02:07.019 1 kserve INFO [model_server.py:setup_event_loop():315] Setting max asyncio worker threads as 20
2026-06-02 09:02:07.201 1 kserve INFO [server.py:_register_endpoints():124] OpenAI endpoints not registered
2026-06-02 09:02:07.201 1 kserve INFO [server.py:_register_endpoints():132] Time series endpoints not registered
2026-06-02 09:02:07.201 1 kserve INFO [server.py:start():183] Starting uvicorn with 1 workers
2026-06-02 09:02:07.256 1 uvicorn.error INFO:     Started server process [1]
2026-06-02 09:02:07.256 1 uvicorn.error INFO:     Waiting for application startup.
2026-06-02 09:02:07.260 1 kserve INFO [server.py:start():70] Starting gRPC server with 4 workers
2026-06-02 09:02:07.260 1 kserve INFO [server.py:start():71] Starting gRPC server on [::]:8081
2026-06-02 09:02:07.260 1 uvicorn.error INFO:     Application startup complete.
2026-06-02 09:02:07.261 1 uvicorn.error INFO:     Uvicorn running on http://0.0.0.0:8080 `

`curl http://localhost:8080
{"status":"alive"}`

**Special notes for your reviewer:**

- This PR only introduces Power (ppc64le) image build support for LightGBMServer.
- For the packages that do not have power support, we are fetching wheels from https://wheels.developerfirst.ibm.com/ppc64le/linux. We have added this index to `project.toml` file of kserve , lgbserver and storage.
- The build time taken for Power is below 5 mins, which is comparable to linux/arm64/v8 when built separately.

**Release note:**

`Added ppc64le support to LightGBMServer`


sandeepgupta12:
We are waiting for review on this PR.

sandeepgupta12:
I'll take a look at the failing jobs.

PrernaKumbhar20:
@Jooho  @spolti  @cjohannsen-cloudera , please approve the workflows to run.
Also please review this PR.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
 
 # Install dependencies for kserve using uv
 COPY kserve/pyproject.toml kserve/uv.lock kserve/
+
+# On ppc64le: patch pyproject.toml to add the ppc64le package index and sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
+            -e '/^index-strategy\s*=.*/a \\' \
+            -e '/^index-strategy\s*=.*/a [[tool.uv.index]]' \
```
cjohannsen-cloudera, line 34: **Security (high):** This sed insertion lands `[[tool.uv.index]]` right after `index-strategy = ...`, which closes the `[tool.uv]` table context in TOML. The `override-dependencies = [...]` block that follows (CVE version floors for urllib3, aiohttp, h11, cryptography, pyjwt, pyasn1, pillow, python-multipart) now parses as `tool.uv.index[0].override-dependencies` instead of `tool.uv.override-dependencies`. `uv` silently ignores the unknown field on the index struct — `uv lock` succeeds with no error, but the ppc64le image resolves those CVE-impacted packages with no minimum-version floor at all. Confirmed via `tomllib`: parsing the post-sed structure shows `override-dependencies=None` under `tool.uv`, with the full list stuck inside `tool.uv.index[0]`.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+
 RUN cd kserve && uv sync --active --no-cache
 
 # Install kserve-storage
 COPY storage storage
-RUN cd storage && uv pip install . --no-cache
+
+# On ppc64le: append ppc64le index + sources to storage/pyproject.toml,
+# regenerate uv.lock, then sync (same pattern as kserve/lgbserver).
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
```
cjohannsen-cloudera, line 76: **Security (high):** This storage patch appends `[tool.uv]` with `index-strategy`/`package = true` but no `override-dependencies` at all. storage/pyproject.toml declares `cryptography>=48.0.1`, so on ppc64le `uv lock` can resolve cryptography down to 48.x and `uv sync` installs it into the shared venv — silently downgrading below kserve's CVE-required 50.0.0 floor and reintroducing GHSA-537c-gmf6-5ccf / PYSEC-2026-36 in every ppc64le lgbserver image, while x86/arm64 stay patched via the committed lockfile.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            -e '/^    "pyasn1>=/a\    "pyyaml==6.0.2",' \
+            storage/pyproject.toml && \
+        printf '%s\n' \
+            '' \
+            '[tool.uv]' \
+            'index-strategy = "unsafe-best-match"' \
+            'package = true' \
+            '' \
+            '[build-system]' \
+            'requires = ["setuptools>=61.0"]' \
```
cjohannsen-cloudera, line 89: **Security (high):** This injected `[build-system]` block sets `requires = ["setuptools>=61.0"]`, but kserve/pyproject.toml deliberately pins `setuptools>=78.1.1` to address CVE-2025-47273. This is the only `[build-system]` declaration storage/pyproject.toml has, so on ppc64le `uv` can resolve and use a setuptools in the 61-77 range during the storage build step — reintroducing the CVE in the ppc64le builder image specifically.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            'pyyaml = { index = "ppc64le-wheels" }' \
+            'uvloop = { index = "ppc64le-wheels" }' \
+            'httptools = { index = "ppc64le-wheels" }' \
+            'scikit-learn = { index = "ppc64le-wheels" }' \
+            'lightgbm = { index = "ppc64le-wheels" }' \
+            'google-crc32c = { index = "ppc64le-wheels" }' \
+            'hf-xet = { index = "ppc64le-wheels" }' \
+            'pillow = { index = "ppc64le-wheels" }' \
+            >> lgbserver/pyproject.toml && \
+        cd lgbserver && uv lock && \
```
cjohannsen-cloudera, line 146: **Correctness (build-breaking risk):** `cd lgbserver && uv lock` regenerates the lock from lgbserver/pyproject.toml, which contains a `file:///${PROJECT_ROOT}/../kserve` path dependency. The non-ppc64le path never calls `uv lock` (it just consumes the committed uv.lock), so this variable substitution is never actually exercised there. If `uv` doesn't expand `${PROJECT_ROOT}` inside this Docker RUN environment (where it's unset), this fresh `uv lock` call could fail with a path-resolution error and break every ppc64le lgbserver build. Worth confirming this resolves correctly before merging — same pattern exists for kserve (line 52) and storage (line 102) but those may be less risky if PROJECT_ROOT happens to be set at that point in the build.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            -e '/^kserve-storage\s*=.*/a numpy = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pandas = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a psutil = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pyyaml = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a httptools = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a uvloop = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a scikit-learn = { index = "ppc64le-wheels" }' \
+            -e '/^kserve-storage\s*=.*/a pillow = { index = "ppc64le-wheels" }' \
+            kserve/pyproject.toml && \
+        cd kserve && uv lock && \
```
cjohannsen-cloudera, line 52: **Reliability:** `uv lock` runs live here instead of using a committed lockfile, so ppc64le builds are non-reproducible — two builds of the same commit on different days could resolve different dependency versions from PyPI and the IBM wheel index (e.g. a CVE-fixed point release lands between builds and silently changes which image is patched). x86/arm64 don't have this problem since they run against the checked-in `uv.lock`. Same pattern repeats for storage (line 102) and lgbserver (line 146) below.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
 
 # Install dependencies for kserve using uv
 COPY kserve/pyproject.toml kserve/uv.lock kserve/
+
+# On ppc64le: patch pyproject.toml to add the ppc64le package index and sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
```
cjohannsen-cloudera, line 32: **Reliability:** None of these sed invocations have a post-patch verification guard — `sed -i` exits 0 whether or not any pattern matched. If kserve/pyproject.toml is ever reformatted (renamed `index-strategy`, changed indentation, edited pyasn1 comment), these `-e` patches silently match zero lines, and the resulting failure surfaces later as a confusing `uv lock` resolver error about a missing ppc64le wheel rather than pointing at the sed that did nothing. A guard like `grep -q 'ppc64le-wheels' kserve/pyproject.toml || { echo 'FATAL: sed patch failed' >&2; exit 1; }` after the sed would catch this immediately.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+            -e '/^index-strategy\s*=.*/a \\' \
+            -e '/^index-strategy\s*=.*/a [[tool.uv.index]]' \
+            -e '/^index-strategy\s*=.*/a name = "ppc64le-wheels"' \
+            -e '/^index-strategy\s*=.*/a url = "https://wheels.developerfirst.ibm.com/ppc64le/linux"' \
+            -e '/^index-strategy\s*=.*/a explicit = true' \
+            -e '/^\s*"pyasn1>=[^,]*"$/s/"$/",/' \
+            -e '/^\s*"pyasn1>=/a\    "httptools==0.6.4",' \
```
cjohannsen-cloudera, line 39: **Correctness (latent):** This anchor `/^\s*"pyasn1>=/` matches pyasn1 at two places in kserve/pyproject.toml — once in `[project].dependencies` and once in `[tool.uv].override-dependencies` — so `httptools==0.6.4` and `uvloop==0.21.0` get inserted twice. Right now the override-dependencies copy lands inside the dead index-table context from the bug above and gets silently discarded, but if that's fixed independently, these duplicate inserts would force-pin httptools/uvloop workspace-wide inside override-dependencies and could conflict with uvicorn[standard]'s own version requirements.

python/lgb.Dockerfile
```diff
@@ -23,20 +25,141 @@ COPY storage/pyproject.toml storage/uv.lock storage/
...
+# On ppc64le: patch pyproject.toml to add the ppc64le package index and sources,
+# then regenerate uv.lock before syncing.
+RUN if [ "$(uname -m)" = "ppc64le" ]; then \
+        sed -i \
+            -e '/^index-strategy\s*=.*/a \\' \
+            -e '/^index-strategy\s*=.*/a [[tool.uv.index]]' \
+            -e '/^index-strategy\s*=.*/a name = "ppc64le-wheels"' \
+            -e '/^index-strategy\s*=.*/a url = "https://wheels.developerfirst.ibm.com/ppc64le/linux"' \
+            -e '/^index-strategy\s*=.*/a explicit = true' \
+            -e '/^\s*"pyasn1>=[^,]*"$/s/"$/",/' \
```
cjohannsen-cloudera, line 38: **Correctness (latent):** This comma-normalizing substitution `/^\s*"pyasn1>=[^,]*"$/s/"$/",/` never actually matches, since both pyasn1 lines in kserve/pyproject.toml end with trailing CVE comments (`"pyasn1>=0.6.4", # CVE-2026-30922, ...`), not a bare closing quote. Harmless today since the append commands below use a looser anchor, but this guard is dead code and would only bite if the surrounding line format changes.

python/lgb.Dockerfile
```diff
@@ -5,7 +5,9 @@ ARG VENV_PATH=/prod_venv
 FROM ${BASE_IMAGE} AS builder
 
 # Install system dependencies
-RUN apt-get update && apt-get install -y --no-install-recommends python3-dev curl build-essential && apt-get clean && \
+RUN apt-get update && apt-get install -y --no-install-recommends curl python3-dev build-essential && \
+    if [ "$(uname -m)" = "ppc64le" ]; then apt-get install pkg-config libssl-dev gcc gfortran cmake pkg-config libssl-dev libopenblas-dev libjpeg-dev libhdf5-dev wget -y; fi && \
```
cjohannsen-cloudera, line 9: **Efficiency / nit:** Two things on this line: (1) missing `--no-install-recommends`, unlike the main install two lines up — pulls in a lot of unneeded transitive packages (docs, locales, alternate lib builds) on the already-slow QEMU build path, and could even affect which python3-dev variant gets linked; (2) `pkg-config` and `libssl-dev` are each listed twice, likely from copy-paste — `apt-get` dedupes silently so it's harmless today, but worth cleaning up.

cjohannsen-cloudera (changes requested):
Automated review pass on the ppc64le/LightGBM server build changes. Requesting changes primarily because of the sed patch structurally corrupting kserve/pyproject.toml's `[tool.uv]` table (comment #1) — that one bug cascades into two concrete CVE-floor bypasses (cryptography and setuptools, comments #2 and #3) on ppc64le images specifically. There's also a plausible build-breaker around `${PROJECT_ROOT}` expansion in a fresh `uv lock` call (comment #4). Details inline; the rest are lower-severity reliability/latent-bug/efficiency notes.

PR #5616, opened 2026-06-02
feat(localmodel): align enhanceDownloadJob hook signature

bartoszmajsak:

The default (no-op) `enhanceDownloadJob` hook in `platform_default.go` accepts only the Job and storage key (2 params), while distribution-specific implementations need `context.Context` and `*LocalModelNodeReconciler` to access the API client and propagate context for platform-specific logic (e.g. resolving security contexts, MCS levels).

This expands the default stub signature to match the contract that platform hook implementations already depend on. The `ensureModelRootFolderExistsAndIsWritable` hook already follows this pattern (accepts context + reconciler).

N/A - signature alignment for distribution hook support.

- [x] `go build ./pkg/controller/v1alpha1/localmodelnode/...` compiles
- [x] `go test ./pkg/controller/v1alpha1/localmodelnode/...` - 4 tests pass

This is a no-op change - the default implementation ignores the extra parameters. The expanded signature enables platform-specific builds to provide implementations that need API client access without modifying the upstream controller code.

**Checklist**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1alpha1/localmodelnode/platform_default.go
+++ b/pkg/controller/v1alpha1/localmodelnode/platform_default.go
@@ -29,10 +29,12 @@ import (
 
 const MountPath = "/mnt/models"
 
-func enhanceDownloadJob(_ *batchv1.Job, _ string) error { return nil }
+func (c *LocalModelNodeReconciler) enhanceDownloadJob(_ context.Context, _ *batchv1.Job, _ string) error {
+	return nil
+}
 
 // TODO we need a way to ensure that the local path on persistent volume is the same as the local path of the node agent DaemonSet.
-func ensureModelRootFolderExistsAndIsWritable(_ context.Context, _ *LocalModelNodeReconciler,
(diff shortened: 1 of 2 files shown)


pkg/controller/v1alpha1/localmodelnode/platform_default.go
```diff
@@ -29,7 +29,9 @@ import (
 
 const MountPath = "/mnt/models"
 
-func enhanceDownloadJob(_ *batchv1.Job, _ string) error { return nil }
+func enhanceDownloadJob(_ context.Context, _ *LocalModelNodeReconciler, _ *batchv1.Job, _ string) error {
+	return nil
+}
```
pierDipi: should this be a method on `enhanceDownloadJob`? kinda weird to be passing `LocalModelNodeReconciler`

bartoszmajsak: Good catch, I adjusted it both here and midstream.

Jooho (approved):

PR #5617, opened 2026-06-02
feat(isvc): add platform hooks for service customization

bartoszmajsak:

The service reconciler creates Kubernetes Services with no extension point for platform-specific
modifications. Distributions that need to inject TLS annotations, override ports, or configure
auth proxy settings must edit the upstream file directly, creating merge conflicts on every sync.

This adds two hook call points in the service creation path:

- `customizeService(svc, componentMeta)` - called at the end of `createDefaultSvc`, allows
  platform builds to modify annotations, ports, or other service spec fields
- `customizeHeadSvc(svc, predictorSvcName)` - called at the end of `createHeadlessSvc`, allows
  platform builds to customize the head node service in multi-node deployments

Default implementations are no-ops behind `//go:build !distro`. Platform builds provide their
own implementations via companion `_odh.go` / `_ocp.go` files.

This follows the same hook pattern used elsewhere in the codebase (e.g. `extendControllerSetup`
in the llmisvc controller, `annotation_filter_ocp.go` in the ingress reconciler).

N/A - extension point for distribution-specific service customization.

- [x] `go build ./...` compiles
- [x] `go test ./pkg/controller/v1beta1/inferenceservice/reconcilers/service/...` - 11 tests pass

This is a no-op change for upstream. The hooks are called but the default implementations do
nothing. The value is enabling downstream distributions to customize services without editing
upstream code.

**Checklist**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler_default.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler_default.go
@@ -1,0 +1,30 @@
+//go:build !distro
+
+/*
+Copyright 2025 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


pierDipi (approved):

pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler_default.go
```diff
@@ -0,0 +1,30 @@
+//go:build !distro
+
+/*
+Copyright 2025 The KServe Authors.
```
spolti, line 4: nit: year

spolti (approved):
headers will all be fixed on a upcoming PR.

PR #5620, opened 2026-06-02
ci(e2e): split test jobs into per-file matrix for targeted reruns

andresllh:
## Summary

- Each e2e test job now runs one test file per matrix entry instead of an entire marker category
- When a single test fails and \`/rerun-all\` is triggered, only that file's matrix entry re-runs (not all tests in the category)
- Jobs converted: \`test-predictor\` (20 files), \`test-transformer-explainer-mms\` (7 files), \`test-path-based-routing\` (7 files), \`test-autoscaling\` (1 file), \`test-kourier\` (2 files), \`test-llm\` (10 files)
- \`test-raw\` excluded — sequential multi-step structure with config patches between run steps makes per-file splitting unsafe without a larger restructure

## How it works

\`PYTEST_ARGS="<file>"\` is passed to \`run-e2e-tests.sh\`, which appends it as a positional path to pytest: \`pytest -m <marker> <file.py>\`. This scopes each job to exactly one test file while still respecting marker filtering.

## Test plan

- [ ] Verify matrix job names appear as \`test-predictor (kustomize, sklearn)\` etc. in the Actions UI
- [ ] Confirm a single failing test triggers rerun of only its matrix entry
- [ ] Check that no test files have zero tests collected (would indicate a marker mismatch)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -359,6 +359,27 @@ jobs:
       fail-fast: false
       matrix:
         install-method: ${{ fromJSON(needs.detect-changes.outputs.install_methods) }}
+        test:
+          - { file: "batcher/test_batcher.py", name: "batcher" }
+          - { file: "batcher/test_batcher_custom_port.py", name: "batcher-custom-port" }
+          - { file: "custom/test_custom_model_grpc.py", name: "custom-model-grpc" }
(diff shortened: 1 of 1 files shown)


andresllh:
Closing in favor of @bartoszmajsak  solution: https://github.com/kserve/kserve/pull/5621

PR #5623, opened 2026-06-03
feat(llmisvc): add PVC storage e2e tests for LLMInferenceService

pierDipi:
Add three new parametrized test cases to validate PVC-based model storage across different deployment topologies:

1. Single-node: workload-single-cpu with model loaded from PVC
2. PD disaggregation: workload-pd-cpu with shared PVC mount
3. LWS multi-worker: workload-simulated-dp-ep-cpu with PVC

The tests use real vLLM CPU serving with facebook/opt-125m downloaded into a PVC via a Kubernetes Job running the kserve storage-initializer.

The PVC setup is idempotent and cluster-agnostic (no hardcoded StorageClass, no explicit security context) so it works on KinD, Minikube, and OpenShift.

### DIFF
--- a/test/e2e/llmisvc/conftest.py
+++ b/test/e2e/llmisvc/conftest.py
@@ -55,6 +55,4 @@ def pytest_configure(config):
         "markers", "model_routing: mark test as a model-based routing test"
     )
     config.addinivalue_line("markers", "lora: mark test as a LoRA adapter test")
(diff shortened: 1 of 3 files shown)


bartoszmajsak (approved):

spolti (approved):
/approve

PR #5636, opened 2026-06-04
fix(storage): add timeout and retry to HuggingFace snapshot_download

### LINKED ISSUE #5611: storage-initializer hangs indefinitely when HuggingFace download stalls mid-transfer
## Description

The `_download_hf` function in `kserve_storage.py` calls `snapshot_download` without any timeout or retry parameters. When a HuggingFace download stalls mid-transfer (e.g., due to CDN throttling, network issues, or platform-specific behavior like AWS SOCI), the storage-initializer init container hangs indefinitely, blocking the entire pod lifecycle.

## Location

`python/storage/kserve_storage/kserve_storage.py` — `_download_hf()` method, around the `snapshot_download(**kwargs)` call.

The current code:
```python
kwargs = dict(repo_id=repo_id, revision=revision, local_dir=temp_dir)
if allow_patterns:
    kwargs["allow_patterns"] = allow_patterns
if ignore_patterns:
    kwargs["ignore_patterns"] = ignore_patterns
snapshot_download(**kwargs)
```

No `etag_timeout`, no retry logic, no `HF_HUB_DOWNLOAD_TIMEOUT` env var support.

## Steps to Reproduce

1. Deploy an InferenceService with `storageUri: hf://Qwen/Qwen3-0.6B` on a cluster where network conditions can cause stalls (e.g., EKS with SOCI enabled)
2. Wait for the storage-initializer init container to start downloading
3. Observe the pod stays in `Init:0/1` indefinitely when the download stalls

## Observed Behavior

- Pod stuck in `Init:0/1` forever
- storage-initializer logs show download starts but produces no further output
- Partial download on disk (e.g., 1.4 GB of 1.6 GB), no progress
- HuggingFace is reachable from the container (not a DNS/network block issue)

## Expected Behavior

- `snapshot_download` should be called with a configurable timeout (e.g., `etag_timeout` parameter)
- When a download stalls, it should timeout and retry with exponential backoff
- Users should be able to configure timeout and retries via environment variables (e.g., `HF_HUB_DOWNLOAD_TIMEOUT`, `HF_HUB_DOWNLOAD_RETRIES`)

## Suggested Fix

```python
import os

download_timeout = int(os.environ.get("HF_HUB_DOWNLOAD_TIMEOUT", "300"))
max_retries = int(os.environ.get("HF_HUB_DOWNLOAD_RETRIES", "3"))

kwargs = dict(
    repo_id=repo_id,
    revision=revision,
    local_dir=temp_dir,
    etag_timeout=download_timeout,
)
# ... patterns ...

for attempt in range(1, max_retries + 1):
    try:
        snapshot_download(**kwargs)
        break
    except Exception as e:
        if attempt == max_retries:
            raise
        logger.warning(f"Download attempt {attempt}/{max_retries} failed: {e}. Retrying...")
        time.sleep(min(2 ** attempt, 60))
```

## Environment

- Platform: AWS EKS with SOCI enabled
- Model: `hf://Qwen/Qwen3-0.6B` (~1.6 GB)
- KServe version: latest master (also affects older releases)

Jwrede:
## What

Add configurable timeout and retry logic to the HuggingFace `snapshot_download` call in the storage-initializer.

Fixes #5611

## Why

`_download_hf` calls `snapshot_download` without `etag_timeout` or retry logic. When a download stalls mid-transfer (CDN throttling, network issues, AWS SOCI behavior), the storage-initializer init container hangs indefinitely, blocking the entire pod lifecycle. Operators have no way to configure timeout behavior.

## Changes

- Pass `etag_timeout` to `snapshot_download` (configurable via `HF_HUB_ETAG_TIMEOUT` env var, default 30s)
- Wrap `snapshot_download` in a retry loop with exponential backoff (configurable via `HF_HUB_DOWNLOAD_RETRIES` env var, default 3 attempts)
- Permanent errors (`RepositoryNotFoundError`, `RevisionNotFoundError`, `GatedRepoError`) are re-raised immediately without retry
- Transient failures (network errors, timeouts, stalls) are retried with backoff capped at 60s

## Testing

- Updated all existing `test_hf_storage.py` assertions to include `etag_timeout`
- New tests: custom timeout from env var, retry on transient failure, retry exhaustion, custom retry count from env var, no retry on `RepositoryNotFoundError`
- All 16 tests pass

### DIFF
--- a/python/storage/test/test_hf_storage.py
+++ b/python/storage/test/test_hf_storage.py
@@ -32,6 +32,7 @@ def test_download_model(mock_snapshot_download):
         repo_id=f"{repo}/{model}",
         revision=revision,
         local_dir=mock.ANY,
+        etag_timeout=30,
     )
 
 
@@ -45,6 +46,7 @@ def test_download_model_with_allow_patterns(mock_snapshot_download):
         repo_id="example.com/model",
         revision=None,
         local_dir="/tmp/out",
+        etag_timeout=30,
         allow_patterns=["*.safetensors", "*.json"],
     )
 
@@ -59,6 +61,7 @@ def test_download_model_with_ignore_patterns(mock_snapshot_download):
         repo_id="example.com/model",
         revision=None,
         local_dir="/tmp/out",
+        etag_timeout=30,
         ignore_patterns=["*.bin", "*.gguf"],
     )
 
@@ -78,6 +81,7 @@ def test_download_model_with_both_patterns(mock_snapshot_download):
         repo_id="example.com/model",
         revision=None,
         local_dir="/tmp/out",
+        etag_timeout=30,
         allow_patterns=["*.json"],
         ignore_patterns=["config.json"],
     )
@@ -93,6 +97,7 @@ def test_download_model_no_patterns_omits_kwargs(mock_snapshot_download):
         repo_id="example.com/model",
         revision=None,
         local_dir="/tmp/out",
+        etag_timeout=30,
     )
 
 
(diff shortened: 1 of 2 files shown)


python/storage/test/test_hf_storage.py
```diff
@@ -32,6 +32,7 @@ def test_download_model(mock_snapshot_download):
         repo_id=f"{repo}/{model}",
         revision=revision,
         local_dir=mock.ANY,
+        etag_timeout=30,
```
aneeshkp, line 35: My  download stalled at 1.4GB/1.6GB  hat's during the actual data transfer, not the metadata check. The TCP socket hung during read()  while downloading the file body.

pierDipi: @aneeshkp are you suggesting a different fix?

Jwrede: Good point. The etag_timeout parameter only covers the metadata HEAD request, not the file body transfer. For stalled body downloads, huggingface_hub uses the HF_HUB_DOWNLOAD_TIMEOUT environment variable (default 10s) as a socket-level read timeout. If the TCP connection hangs during a chunk read, that timeout fires and raises an exception, which our retry loop then catches and retries with exponential backoff.

So the retry loop in this PR does cover your scenario: a stalled download at 1.4GB/1.6GB would eventually hit the socket timeout, raise, and trigger a retry. If the default 10s is too aggressive or too lenient, it can be tuned via HF_HUB_DOWNLOAD_TIMEOUT.

Would it be useful to mention HF_HUB_DOWNLOAD_TIMEOUT in the code comment alongside HF_HUB_ETAG_TIMEOUT?

aneeshkp: The retry logic looks good.

Jwrede:
Thanks for the reviews. Is there anything else needed here before this can be approved and merged, or is it just waiting on a maintainer approval?

spolti (approved):

pierDipi (approved):
(I'm not an approver, but lgtm)

PR #5637, opened 2026-06-05
fix: set limits and increase requests in llmisvc

rsevilla87:
## Summary
Updates the resource requests and limits configuration for the LLM scheduler, to ensure full compatibility with Kubernetes `LimitRanges`.

### Why is this change necessary?
Previously, the scheduler omitted explicit resource limits/requests or used configurations that conflicted with our cluster's `LimitRanges`. In namespaces where strict limit ranges are enforced, pods would fail to deploy with validation errors (e.g., missing mandatory limits or exceeding maximum constraints). 

### Testing

These numbers are derived from the results after running some load tests using llm-d-scheduler 0.6 with 8 backends:

#### "default" scheduling, (empty value in the scheduler field of the LLMIinferenceService object)
image

#### "approximate" scheduling
<details>
<summary>**View config**</summary>

```yaml
kind: LLMInferenceService
metadata:
  name: qwen3-0-6b-approximate
  annotations:
    security.opendatahub.io/enable-auth: "false"
spec:
  model:
    uri: "hf://dummy-org/dummy-model"
    name: Qwen/Qwen3-0.6B
  replicas: 8
  router:
    scheduler:
      config:
        inline:
          kind: EndpointPickerConfig
          plugins:
          - type: queue-scorer
          - type: kv-cache-utilization-scorer
          - type: prefix-cache-scorer
          - type: no-hit-lru-scorer
          schedulingProfiles:
          - name: default
            plugins:
            - pluginRef: queue-scorer
              weight: 2
            - pluginRef: kv-cache-utilization-scorer
              weight: 2
            - pluginRef: prefix-cache-scorer
              weight: 3
            - pluginRef: no-hit-lru-scorer
              weight: 2
    route: {}
    gateway: {}
  template:
    containers:
      - name: main
        image: ghcr.io/llm-d/llm-d-inference-sim:v0.8.2
        command: ["/app/llm-d-inference-sim"]
        args:
          - --port
          - "8000"
          - --model
          - "{{ .Spec.Model.Name }}"
          - --mode
          - random
          - --hash-seed
          - "42"
          - --event-batch-size
          - "1"
          - --ssl-certfile
          - /var/run/kserve/tls/tls.crt
          - --ssl-keyfile
          - /var/run/kserve/tls/tls.key
          - --max-model-len
          - "131072"
        env:
        - name: POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: PYTHONHASHSEED
          value: "42"
        livenessProbe:
          httpGet:
            path: /health
            port: 8000
            scheme: HTTPS
          periodSeconds: 5
          timeoutSeconds: 30
          failureThreshold: 5
        readinessProbe:
          httpGet:
            path: /health
            port: 8000
            scheme: HTTPS
          periodSeconds: 5
          timeoutSeconds: 30
          failureThreshold: 5
        startupProbe:
          httpGet:
            path: /health
            port: 8000
            scheme: HTTPS
          periodSeconds: 10
          timeoutSeconds: 10
          failureThreshold: 10
```
</details>

image

---

I also conducted more experiments using a higher number of backends, using the approximate scheduling configuration again

#### **CPU & Memory with more replicas**

| RPS | 8 Replicas | 16 Replicas | CPU Increase (%) | 24 Replicas | CPU Increase (%) | 32 Replicas | CPU Increase (%) | 40 Replicas | CPU Increase (%) |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| 20 | 0.456 | 0.684 | \+50.0% | 0.938 | \+105.7% | 1.189 | \+160.7% | 1.409 | \+209.0% |
| 40 | 0.655 | 0.895 | \+36.6% | 1.162 | \+77.4% | 1.393 | \+112.7% | 1.633 | \+149.3% |
| 60 | 0.861 | 1.081 | \+25.6% | 1.360 | \+58.0% | 1.609 | \+86.9% | 1.865 | \+116.6% |
| 80 | 1.045 | 1.285 | \+23.0% | 1.364 | \+30.5% | 1.779 | \+70.2% | 1.984 | \+89.9% |
| 100 | 1.263 | 1.506 | \+19.2% | 1.749 | \+38.5% | 2.032 | \+60.9% | 2.243 | \+77.6% |
</div>

Although the limits may be insufficient on higher load cases, it helps to establish a reasonable baseline.

/cc @pierDipi

### DIFF
--- a/config/llmisvcconfig/config-llm-scheduler.yaml
+++ b/config/llmisvcconfig/config-llm-scheduler.yaml
@@ -86,9 +86,12 @@ spec:
               - name: SSL_CERT_DIR
                 value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
             resources:
+              limits:
+                cpu: 6
+                memory: 16Gi
               requests:
-                cpu: 256m
-                memory: 500Mi
+                cpu: 1
+                memory: 2Gi
             terminationMessagePath: /dev/termination-log
             terminationMessagePolicy: FallbackToLogsOnError
             securityContext:
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -1991,9 +1991,12 @@ spec:
             successThreshold: 1
             timeoutSeconds: 1
           resources:
+            limits:
+              cpu: 6
+              memory: 16Gi
             requests:
-              cpu: 256m
-              memory: 500Mi
+              cpu: 1
+              memory: 2Gi
           securityContext:
             allowPrivilegeEscalation: false
             capabilities:
--- a/hack/setup/quick-install/kserve-knative-mode-full-install-with-manifests.sh
+++ b/hack/setup/quick-install/kserve-knative-mode-full-install-with-manifests.sh
@@ -4594,9 +4594,12 @@ spec:
             successThreshold: 1
             timeoutSeconds: 1
           resources:
+            limits:
+              cpu: 6
+              memory: 16Gi
             requests:
-              cpu: 256m
-              memory: 500Mi
+              cpu: 1
+              memory: 2Gi
           securityContext:
             allowPrivilegeEscalation: false
             capabilities:
--- a/hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh
+++ b/hack/setup/quick-install/kserve-standard-mode-full-install-with-manifests.sh
@@ -4180,9 +4180,12 @@ spec:
             successThreshold: 1
             timeoutSeconds: 1
           resources:
+            limits:
+              cpu: 6
+              memory: 16Gi
             requests:
-              cpu: 256m
-              memory: 500Mi
+              cpu: 1
+              memory: 2Gi
           securityContext:
             allowPrivilegeEscalation: false
             capabilities:
--- a/test/e2e/llmisvc/fixtures.py
+++ b/test/e2e/llmisvc/fixtures.py
@@ -379,7 +379,25 @@ LLMINFERENCESERVICE_CONFIGS = {
         },
     },
     "router-managed": {
-        "router": {"scheduler": {}, "route": {}, "gateway": {}},
+        "router": {
+            "scheduler": {
+                "template": {
+                    "containers": [
+                        {
+                            "name": "main",
+                            "resources": {
+                                "requests": {
+                                    "cpu": "256m",
+                                    "memory": "500Mi",
+                                },
+                            },
+                        }
+                    ],
+                },
+            },
+            "route": {},
+            "gateway": {},
+        },
     },
     "router-no-scheduler": {
         "router": {"route": {}},
@@ -711,7 +729,21 @@ LLMINFERENCESERVICE_CONFIGS = {
     },
     "scheduler-managed": {
         "router": {
-            "scheduler": {},
+            "scheduler": {
+                "template": {
+                    "containers": [
+                        {
+                            "name": "main",
+                            "resources": {
+                                "requests": {
+                                    "cpu": "256m",
+                                    "memory": "500Mi",
+                                },
+                            },
+                        }
+                    ],
+                },
+            },
         },
     },
     "scheduler-with-inline-config": {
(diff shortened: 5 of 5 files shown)


config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -81,8 +81,11 @@ spec:
                 value: "/var/run/kserve/tls:/var/run/secrets/kubernetes.io/serviceaccount:/etc/pki/tls/certs"
             resources:
               requests:
-                cpu: 256m
-                memory: 500Mi
+                cpu: 1
+                memory: 2Gi
+              limits:
+                cpu: 6
+                memory: 16Gi
```
pierDipi: Since I know you've the reasoning for these numbers, can we leave a trail of that reasoning summary as a comment here?

pierDipi:
Thanks!

config/llmisvcconfig/config-llm-scheduler.yaml
rsevilla87: makes sense, I've updated the PR description, PTAL

pierDipi:
@spolti @Jooho can you please trigger tests?

yashaswipiplani:
@rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks

rsevilla87:
> @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks

done, let's see if there's 

> @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks

done, hope the ci doesn't complain again because of diff between llmisvc templates

yashaswipiplani:
> > @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks
> 
> done, let's see if there's
> 
> > @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks
> 
> done, hope the ci doesn't complain again because of diff between llmisvc templates

I think it probably would but regardless this doesn't seem the right way, I think. Actually, I meant passing the earlier set resources for e2e only here: [test/e2e/llmisvc/fixtures.py](https://github.com/kserve/kserve/blob/master/test/e2e/llmisvc/fixtures.py)

rsevilla87:
> > > @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks
> > 
> > 
> > done, let's see if there's
> > > @rsevilla87 the llmisvc e2e tests are failing. I believe it is because we use the presets from config-llm-scheduler.yaml for resources requests and limits in e2e tests as well which might be too much for the CI VMs. I think we should pass the same resources in e2e tests as earlier. Thanks
> > 
> > 
> > done, hope the ci doesn't complain again because of diff between llmisvc templates
> 
> I think it probably would but regardless this doesn't seem the right way, I think. Actually, I meant passing the earlier set resources for e2e only here: [test/e2e/llmisvc/fixtures.py](https://github.com/kserve/kserve/blob/master/test/e2e/llmisvc/fixtures.py)

done, let's try again

rsevilla87:
> /rerun-all

I think you need to approve the CI run

yashaswipiplani:
> > /rerun-all
> 
> I think you need to approve the CI run

Can you run make precommit please. llmisvc e2e tests are passing now it seems

rsevilla87:
> > > /rerun-all
> > 
> > 
> > I think you need to approve the CI run
> 
> Can you run make precommit please. llmisvc e2e tests are passing now it seems

done, precommit passed now

yashaswipiplani (commented):
LGTM!

rsevilla87:
> LGTM!

tests need approve to run

yuzisun (approved):

PR #5639, opened 2026-06-05
feat(logger): prefer X-Request-Id over Ce-Id for request ID

### LINKED ISSUE #5591: feat(logger): use dedicated request ID header instead of Ce-Id
The TODO in `pkg/logger/worker.go:43` acknowledges that Ce-Id is used as a request ID as a workaround. The current implementation in getOrCreateID (`pkg/logger/handler.go:210-216`) reads Ce-Id as the request identifier, but this has real implications for users:
- Ce-Id is the CloudEvents event identifier: its semantics belong to the messaging layer, not to inference request tracking
- If a user sends Ce-Id with their own value for CloudEvents purposes, that value appears in inference logs as a request ID even though it isn't one
- If a user wants to track inference requests by ID and sends X-Request-Id, the logger ignores it entirely

Additionally, x-request-id is already used in the Python inference server (`kserve/python/model.py`) for request tracking. The Go logger does not follow this convention, reading Ce-Id instead.

**How we could handle this:** add support for a dedicated header (e.g. X-Request-Id) with fallback to Ce-Id for backward compatibility, and eventually deprecate the use of Ce-Id as a request ID.

ItsVigneshMurugan:

The logger resolves the inference request ID from the `Ce-Id` header (`getOrCreateID` in `pkg/logger/handler.go`), which is the CloudEvents *event* identifier and belongs to the messaging layer, not inference request tracking. The code acknowledged this with a TODO in `pkg/logger/worker.go`:

```go
// TODO: ideally request id would have its own header but make do with ce-id for now
```

Consequences of the current behaviour:
- A client-supplied `X-Request-Id` (the conventional request tracking header, injected automatically by Istio/Envoy and already used for request tracking by the Python model server in `python/kserve/kserve/model.py`) is ignored entirely — payload log events get a random UUID, so they cannot be correlated with access logs, traces, or the model server's own logs.
- An incoming `Ce-Id` (e.g. when the InferenceService is invoked from a CloudEvents pipeline) is silently adopted as the inference request ID and reused as the emitted events' own IDs.

This PR resolves the request ID with the precedence suggested in #5591: **`X-Request-Id` → `Ce-Id` (backward compatibility) → generated UUID**, and removes the stale TODO.

Fixes #5591

- [x] Unit test: new table-driven `TestGetOrCreateID` in `pkg/logger/handler_test.go` covering all four scenarios (prefers `X-Request-Id` over `Ce-Id`, uses `X-Request-Id` alone, falls back to `Ce-Id`, generates a UUID when neither is present). `go test ./pkg/logger/...` and `./pkg/agent/...` pass; `golangci-lint run ./pkg/logger/...` reports 0 issues.
- [x] Verified end-to-end on a kind cluster: deployed a logger-enabled sklearn InferenceService with an event-display sink and sent three predictions (plain, with `X-Request-Id: my-trace-123`, with `Ce-Id: someone-elses-event-id`).

Before (stock agent):

| Request | CloudEvent `id` received by sink |
|---|---|
| plain | random UUID |
| `X-Request-Id: my-trace-123` | random UUID — header ignored |
| `Ce-Id: someone-elses-event-id` | `someone-elses-event-id` |

After (patched agent):

```
  type: org.kubeflow.serving.inference.request
  id: 066ce481-dbe5-4e2b-92bc-5890c8fb347d      <- plain: generated UUID (unchanged)
  type: org.kubeflow.serving.inference.request
  id: my-trace-123                              <- X-Request-Id now honoured
  type: org.kubeflow.serving.inference.request
  id: someone-elses-event-id                    <- Ce-Id fallback preserved
```

(each request event's matching response event carries the same id)

**Special notes for reviewer**:

The issue also suggests eventually deprecating `Ce-Id` as a request ID source; this PR intentionally keeps it as a fallback so existing CloudEvents-based callers are unaffected. Happy to adjust the precedence or follow up on deprecation if preferred.

**Checklist**:

**Release note**:
```release-note
The inference logger now uses the X-Request-Id header as the request ID for logged request/response events, falling back to Ce-Id for backward compatibility, and generating a UUID when neither is present.
```

### DIFF
--- a/pkg/logger/handler.go
+++ b/pkg/logger/handler.go
@@ -207,8 +207,15 @@ func (eh *LoggerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// getOrCreateID resolves the request ID used to correlate the logged request
+// and response events. It prefers the dedicated X-Request-Id header, falling
+// back to Ce-Id for backward compatibility, and generates a UUID when neither
+// is present.
 func getOrCreateID(r *http.Request) string {
-	id := r.Header.Get(CloudEventsIdHeader)
+	id := r.Header.Get(RequestIdHeader)
+	if id == "" {
+		id = r.Header.Get(CloudEventsIdHeader)
+	}
 	if id == "" {
 		id = guuid.New().String()
 	}
--- a/pkg/logger/handler_test.go
+++ b/pkg/logger/handler_test.go
@@ -27,6 +27,7 @@ import (
 	"testing"
 	"time"
 
+	guuid "github.com/google/uuid"
 	"github.com/onsi/gomega"
 	pkglogging "knative.dev/pkg/logging"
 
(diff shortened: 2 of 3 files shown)


ItsVigneshMurugan:
Friendly ping — this is my first contribution to KServe and the test workflows are currently awaiting maintainer approval. Could someone take a look when you have a moment? Happy to address any feedback.

/cc @sivanantha321 @spolti

PR #5650, opened 2026-06-09
fix(ci): retry corrupted docker image artifact downloads

pierDipi:
Add tar integrity verification before loading docker images. When a tarball is corrupted or truncated (e.g. unexpected EOF), re-download the artifact using gh CLI with up to 3 retries and backoff.


bartoszmajsak (approved):
Good stuff! Thanks!

PR #5652, opened 2026-06-09
fix: preserve storageUri during webhook defaulting

maskarb:
## Summary
- Preserve `model.storageUri` when `assign*Runtime()` overwrites the Model spec during legacy-to-Model conversion in the mutating webhook
- Add unit tests (5 scenarios) and e2e tests (4 scenarios) verifying storageUri preservation

## Root Cause

The `assign*Runtime()` methods (e.g. `assignSKLearnRuntime()`) create a **new** `ModelSpec` from the legacy spec's `PredictorExtensionSpec`, discarding any existing `model.storageUri`. This affects any workflow where both a legacy spec and a Model spec with `storageUri` coexist during webhook defaulting.

## Fix

Save `storageUri` before the legacy-to-Model conversion switch and restore it afterward if it was cleared. When the legacy spec carries its own `storageUri`, it takes precedence (the conversion copies it through).

## Test plan
- [x] Unit tests: 5 scenarios — model-only, sklearn+model, tensorflow+model, xgboost+model, legacy-wins
- [x] E2e tests: 4 scenarios exercising the real webhook pipeline
  - Legacy spec + model storageUri preserved on CREATE
  - storageUri preserved on PUT/replace (dashboard workflow)
  - No spurious storageUri injected when none set
  - Legacy storageUri takes precedence over model storageUri

### DIFF
--- a/test/e2e/predictor/test_webhook_defaults.py
+++ b/test/e2e/predictor/test_webhook_defaults.py
@@ -1,0 +1,234 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import time
+
+import pytest
+from kubernetes import client
+from kubernetes.client import V1ResourceRequirements
+
+from kserve import (
+    KServeClient,
+    V1beta1InferenceService,
+    V1beta1InferenceServiceSpec,
+    V1beta1ModelFormat,
+    V1beta1ModelSpec,
+    V1beta1PredictorSpec,
+    V1beta1SKLearnSpec,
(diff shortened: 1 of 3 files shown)


test/e2e/predictor/test_webhook_defaults.py
```diff
@@ -0,0 +1,234 @@
...
+        assert _get_storage_uri(service_name) == "oci://quay.io/example/model:latest"
+    finally:
+        kserve_client.delete(service_name, KSERVE_TEST_NAMESPACE)
+
+
+@pytest.mark.predictor
+def test_storage_uri_preserved_on_replace():
+    """storageUri must survive a PUT/replace that changes imagePullSecrets.
+
+    This mirrors the dashboard workflow (RHOAIENG-61105): delete old OCI
```
spolti:
```suggestion
```

test/e2e/predictor/test_webhook_defaults.py
```diff
@@ -0,0 +1,234 @@
...
+    secret, create new one, PUT the IS referencing the new secret.
```
spolti:
```suggestion
```

israel-hdez (commented):
Hi @maskarb 

TBH, this is something very unlikely to happen since the website no longer documents the legacy spec. I see this as a very, very edge case to the point I'm not sure we need this.

Is there really something to fix here? Mixing legacy spec and new spec sounds to me like something that should not happen, maybe we should have a validating webhook instead?

PR #5653, opened 2026-06-10
feat(localmodel): support tolerations for LocalModelCache download jobs

### LINKED ISSUE #5348: Support tolerations for LocalModelCache download Jobs
## Summary

When a `LocalModelCache` object is created, KServe spawns a Kubernetes `Job` (via `LocalModelNodeReconciler.launchJob`) to download the model onto the node. Currently, there is no way to specify `tolerations` for this Job's Pod, which makes it impossible to run download Jobs on tainted nodes.

## Motivation

It is common to run model-serving workloads on dedicated node pools with taints (e.g. GPU nodes, high-memory nodes, or nodes reserved for inference workloads). These nodes typically carry taints like:

```
nvidia.com/gpu=present:NoSchedule
dedicated=inference:NoSchedule
```

Since `LocalModelCache` is designed to pre-cache models on specific node groups, it is natural that those nodes may be tainted. Without toleration support, the download Job cannot be scheduled on those nodes even though the model is intended to be cached there.

## Current Behavior

`launchJob()` in `pkg/controller/v1alpha1/localmodelnode/controller.go` creates the Job with a hardcoded `PodSpec` that does not include `tolerations`:

```go
Spec: corev1.PodSpec{
    NodeName:      nodeName,
    Containers:    []corev1.Container{*container},
    RestartPolicy: corev1.RestartPolicyNever,
    Volumes:       volumes,
    SecurityContext: &corev1.PodSecurityContext{
        FSGroup: FSGroup,
    },
    // no Tolerations field
},
```

## Expected Behavior

It should be possible to configure tolerations for the download Job, either via `LocalModelNodeGroup` (as a node-group-level setting) or via the KServe config (as a global default), so that the Job's Pod can be scheduled on tainted nodes.

ammarasyad:
This PR adds optional job tolerations to `LocalModelNodeGroup`. This allows download jobs created by agents to have their tolerations specified, allowing jobs to be assigned to tainted nodes commonly used by node pools dedicated for inference or GPU-assigned nodes.

**Note: Existing jobs do not have their tolerations propagated after creation. Jobs will have to be recreated if tolerations are edited after the fact.**

Fixes #5348 

- [x] Build `localmodel-agent` and deploy to cluster
- [x] Create a `LocalModelCache` and observe created jobs assigned to tainted nodes

**Checklist**:

**Release note**:
```release-note
feat(localmodel): support for adding tolerations to LocalModelCache download jobs
```


aslafy-z:
This is a desired feature

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
```diff
@@ -31,6 +31,10 @@ type LocalModelNodeGroupSpec struct {
 	PersistentVolumeSpec corev1.PersistentVolumeSpec `json:"persistentVolumeSpec"`
 	// Used to create PersistentVolumeClaims for download and in inference service namespaces
 	PersistentVolumeClaimSpec corev1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec"`
+	// Used to specify tolerations for download jobs
+	// +optional
+	// +listType=atomic
+	JobTolerations []corev1.Toleration `json:"jobTolerations,omitempty"`
```
spolti: Can we keep it named just tolerations? Or make the toleration an inline field?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -114,7 +114,17 @@ func (c *LocalModelNodeReconciler) getNodeGroupFromNode(ctx context.Context, nod
 	return nil, fmt.Errorf("did not find matching nodegroup for node: %s", nodeName)
 }
 
+func (c *LocalModelNodeReconciler) getJobTolerationsFromNodeGroup(ctx context.Context, nodeGroupName string) ([]corev1.Toleration, error) {
+	nodeGroup := &v1alpha1.LocalModelNodeGroup{}
+	if err := c.Get(ctx, types.NamespacedName{Name: nodeGroupName}, nodeGroup); err != nil {
+		return nil, err
+	}
+
+	return nodeGroup.Spec.JobTolerations, nil
```
spolti: shouldn't we test if the field is populated before returning it?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -126,6 +136,15 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, localModelNode
 			return nil, err
 		}
 		nodeGroupName = nodeGroup.Name
+		jobTolerations = nodeGroup.Spec.JobTolerations
```
spolti: what happens if the field is empty?

pkg/controller/v1alpha1/localmodelnode/controller.go
```diff
@@ -126,6 +136,15 @@ func (c *LocalModelNodeReconciler) launchJob(ctx context.Context, localModelNode
...
+	} else {
+		tolerations, err := c.getJobTolerationsFromNodeGroup(ctx, nodeGroupName)
```
spolti, line 141: same here, as no check is done to make sure the field is populated, we can introduce faulty behavior.

pkg/apis/serving/v1alpha1/local_model_node_group_types.go
ammarasyad: I can rename this to tolerations.

pkg/controller/v1alpha1/localmodelnode/controller.go
ammarasyad: An empty tolerations list and a nil slice is functionally the same when serialized with omitempty json tag. What should this check before returning?

pkg/controller/v1alpha1/localmodelnode/controller.go
ammarasyad: The pod will have empty tolerations as expected. It will be omitted in the serialized json

pkg/controller/v1alpha1/localmodelnode/controller.go
ammarasyad: There's no need to check, if tolerations is an empty list or nil it's functionally the same (field will get omitted in the resulting json, the same behavior as upstream before this PR).

VedantMahabaleshwarkar:
along with the download jobs, don't we also need to provide a way to pass these tolerations to the localmodelnode agent daemonset yaml so that the agents are running on the correct nodes w.r.t tolerations?

ammarasyad:
> along with the download jobs, don't we also need to provide a way to pass these tolerations to the localmodelnode agent daemonset yaml so that the agents are running on the correct nodes w.r.t tolerations?

LocalModelNode agents can be set independently (from `Values.kserve.localmodelnode.controller.tolerations`). I'm not sure how we should approach matching the tolerations list.

When I tested this in our cluster with tainted GPU nodes, I simply specified the same tolerations for the LocalModelNode agents. I think that's a valid use flow.

VedantMahabaleshwarkar (approved):

ammarasyad:
@spolti @VedantMahabaleshwarkar any feedbacks? the tests seem to fail not from my changes, but if there are changes needed from my side let me know.

mwaykole:
Tolerations on the download Job look correct for both the inferred and explicit NodeGroup paths. Two things worth checking: (1) Jobs are only created (not patched), so changing NodeGroup tolerations later won't update an already-created download Job — worth a short note in the docs/PR description. (2) Confirm the localmodelnode agent DaemonSet doesn't also need these tolerations; otherwise downloads can land on tainted nodes while the agent still can't schedule there (or the reverse).

ammarasyad:
> Tolerations on the download Job look correct for both the inferred and explicit NodeGroup paths. Two things worth checking: (1) Jobs are only created (not patched), so changing NodeGroup tolerations later won't update an already-created download Job — worth a short note in the docs/PR description. (2) Confirm the localmodelnode agent DaemonSet doesn't also need these tolerations; otherwise downloads can land on tainted nodes while the agent still can't schedule there (or the reverse).

Sorry for my late response. Regarding (1) I agree, I'll add it to the PR description. However for (2) I am currently not in the position to test LocalModelCache again. During my testing I noticed it still worked without tolerations for the agent. Logically it shouldn't, so I might just be seeing them getting downloaded to untainted nodes. I'll add those in.

VedantMahabaleshwarkar:
@ammarasyad can you follow up on this PR? Please address any open comments and fix the failing E2E tests.

ammarasyad:
> @ammarasyad can you follow up on this PR? Please address any open comments and fix the failing E2E tests.

I forwarded the tolerations to the DaemonSet and rebased with upstream. With the rebase I noticed there's a new LocalModelNamespaceCache CRD, which I have to make explicit that this PR **does not** cover that path.

PR #5655, opened 2026-06-10
feat(llmisvc): auto-enable lora-affinity-scorer for LoRA adapters

andresllh:
## Summary

- When an `LLMInferenceService` has LoRA adapters specified, the controller now automatically injects the `lora-affinity-scorer` plugin into the default llm-d-router scheduling profile — no manual scheduler config required from the platform operator
- When no LoRA adapters are configured, behavior is unchanged
- The injection is idempotent and only targets the `default` schedulingProfile

## Background

The `lora-affinity-scorer` plugin scores endpoints by whether the requested LoRA adapter is already active, waiting, or beyond capacity on each pod. When LoRA adapters are in use, this scorer should be part of the default profile to minimize adapter load latency and improve request routing efficiency.

## Implementation

The injection follows the existing `mutateSchedulerConfigFunc` pattern (same as `withCoreMetricsExtractorPlugin`, etc.): a new `withLoraAffinityScorer(hasLoRA bool)` function is applied unconditionally in `expectedSchedulerDeployment`. It:

1. No-ops when `hasLoRA` is false
2. Checks if `lora-affinity-scorer` is already present (idempotent)
3. Appends the plugin to the top-level `plugins` list
4. Prepends a `pluginRef` with `weight: 4` to the `default` schedulingProfile (higher than `prefix-cache-scorer`'s weight of 3, since adapter locality is the primary routing concern when LoRA adapters are in use)

## Test plan

- [ ] `TestWithLoraAffinityScorer`: 4 unit test cases covering no-op (no LoRA), injection (with LoRA), idempotency, and only-default-profile targeting
- [ ] All existing scheduler tests continue to pass

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/scheduler.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler.go
@@ -56,6 +56,7 @@ const (
 
 	precisePrefixCacheScorerPlugin = "precise-prefix-cache-scorer"
 	prefixCacheScorerPlugin        = "prefix-cache-scorer"
+	loraAffinityScorerPlugin       = "lora-affinity-scorer"
 	coreMetricsExtractorPlugin     = "model-server-protocol-metrics"
 	udsTokenizerBaseModelName      = "base"
 	udsTokenizerSocketFile         = "/tmp/tokenizer/tokenizer-uds.socket" //nolint:gosec // G101: not a credential, UDS socket path
@@ -453,6 +454,10 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 			}
 		}
 
+		if err := mutateSchedulerConfig(ctx, d, withLoraAffinityScorer(llmSvc.Spec.Model.LoRA != nil && len(llmSvc.Spec.Model.LoRA.Adapters) > 0)); err != nil {
+			return d, fmt.Errorf("failed to inject lora-affinity-scorer into scheduler config: %w", err)
+		}
+
 		// v0.7.0 migrations: version-gated (>= 0.7.0) because the v0.6 binary
 		// does not recognize the new plugin names. The v0.7 binary deprecates
 		// the old names but still accepts them.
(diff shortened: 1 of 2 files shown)


pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -838,6 +843,65 @@ func WithUdsTokenizerConfig(_ context.Context, u *unstructured.Unstructured) err
...
+		if err != nil || !found {
+			return err
+		}
+		profileList, ok := profiles.([]interface{})
+		if !ok {
+			return nil
+		}
+		for _, profile := range profileList {
+			profileMap, ok := profile.(map[string]interface{})
+			if !ok || profileMap["name"] != "default" {
```
vivekk16, line 892: Why does this only target the `default` profile? With P/D disaggregation the config has `prefill` and `decode` profiles instead, no `default` exists. Are we not supporting LoRA + P/D, or is this a gap?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -453,6 +454,10 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 			}
 		}
 
+		if err := mutateSchedulerConfig(ctx, d, withLoraAffinityScorer(llmSvc.Spec.Model.LoRA != nil)); err != nil {
```
vivekk16: Nit: `LoRA != nil` is true even for an empty `lora: {}` with no adapters. Should this also check `len(llmSvc.Spec.Model.LoRA.Adapters) > 0`?

Separate concern: if `lora: {}` with zero adapters isn't a valid use case, it might be worth rejecting it at the webhook level in `validateLoRAAdapters` (currently it silently allows it). That would be a separate PR though.

pkg/controller/v1alpha2/llmisvc/scheduler.go
vivekk16, line 1: if LoRA adapters are later removed from the spec, the scorer will persist in the config because `preserveSchedulerConfig` carries forward the existing `--config-text` and `hasLoRA=false` is just a no-op. Not a bug (scorer is harmless when idle) but worth being aware of that there is no removal path.

pkg/controller/v1alpha2/llmisvc/scheduler.go
andresllh: Fixed — changed the condition to `llmSvc.Spec.Model.LoRA != nil && len(llmSvc.Spec.Model.LoRA.Adapters) > 0`. Agreed on the webhook validation being a separate PR.

pkg/controller/v1alpha2/llmisvc/scheduler.go
andresllh: Acknowledged — added a comment to the function documenting this. The scorer is cheap when idle so leaving it in place is acceptable; a full removal path (clearing the plugin from a live scheduler config) would require additional reconciler machinery and is better handled as a follow-up if it becomes a concern.

pkg/controller/v1alpha2/llmisvc/scheduler.go
andresllh: Good catch — this is intentional scope for this PR: the goal is to enable the scorer for standard single-profile deployments. LoRA + P/D disaggregation is a separate concern; I've added a comment to the function documenting that the prefill/decode profiles are intentionally not modified here, so it's visible to future contributors.

vivekk16 (commented):
LGTM!

@pierDipi @bartoszmajsak @Jooho

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -453,6 +454,10 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 			}
 		}
 
+		if err := mutateSchedulerConfig(ctx, d, withLoraAffinityScorer(llmSvc.Spec.Model.LoRA != nil && len(llmSvc.Spec.Model.LoRA.Adapters) > 0)); err != nil {
+			return d, fmt.Errorf("failed to inject lora-affinity-scorer into scheduler config: %w", err)
+		}
```
pierDipi, line 459: I prefer that we only modify the [default configuration](https://github.com/kserve/kserve/blob/master/pkg/controller/v1alpha2/llmisvc/scheduler.go#L532-L586) instead of injecting the plugin arbitrarily on any configuration because now there is no way of overriding it.

If there will ever be a better LoRA plugin (which is being planned) we're always injecting this one now which might not be what we want

andresllh: Good point — addressed in #5675. The injection is moved into `schedulerConfigText` so it only applies to the auto-generated default config. User-provided configs (inline, ConfigMap ref, or preserved from an existing deployment) are left untouched. The `withLoraAffinityScorer` mutation function is removed entirely.

PR #5657, opened 2026-06-10
fix(ci): centralize hfserver XET/tokio env patch for helm and kustomize

vivekk16:

Moves the `HF_HUB_DISABLE_XET=1` and `TOKIO_WORKER_THREADS=1` env var patching from the kustomize-only overlay into `setup-kserve.sh` so it applies to both helm and kustomize install methods via `kubectl patch` after install.

Previously, this was only applied via a kustomize overlay patch file (#5510), which meant the helm path never got these env vars. This caused all helm-based vLLM E2E tests to fail because the huggingfaceserver pods would hang downloading models from HuggingFace Hub due to hf-xet client bugs.

- Removes `config/overlays/test/clusterresources/huggingfaceserver-tokio-patch.yaml` and its reference in `kustomization.yaml`
- Adds a centralized `kubectl patch` in `setup-kserve.sh` that runs after both helm and kustomize install the ClusterServingRuntimes
- CI-only change, does not affect production manifests or helm chart defaults


vivekk16:
This fixes the failing test-llm (helm) and test-huggingface-server-vllm (helm) tests
@pierDipi @bartoszmajsak @Jooho PTAL

Jooho:
/hold

@vivekk16 Did you add this env to ClusterServingRuntime under config/runtime? Kustomize is the source of truth  so if you add the env into the manifests under config folder, it will be added to helm chart automatically. 
With this way, only helm chart will have the information.

Jooho:
FYI, the purpose of this patch YAML is to make the image configurable through Helm --set parameters. For example:
```
      image: "{{ .Values.kserve.servingruntime.lgbserver.image }}:{{ .Values.kserve.servingruntime.lgbserver.tag | default .Values.kserve.version }}"
```
Once the patch YAML is updated, please also add the corresponding entries to values.yaml so they can be overridden by users.

Jooho:
I think I forgot to put the doc about this information. I will send the document soon

pierDipi:
@Jooho one question is why is CI not failing for that?

andresllh:
Just one note @vivekk16 
I don't think we should have XET disabled by default and limit users to 1 thread as well.  This should be a patch that gets applied to the CI only. 
See my original PR: https://github.com/kserve/kserve/pull/5510/changes
Specific patch file: https://github.com/andresllh/kserve/blob/14abd53cd9bdc0d75651f33b3498fea7a85a6fb3/config/overlays/test/clusterresources/huggingfaceserver-tokio-patch.yaml

I missed the helm side of things and should have done a patch for it as well. I was going  to include the fix as part of one of my PRs here: https://github.com/kserve/kserve/pull/5602/changes#:~:text=echo%20%22Applying,%5D%27

vivekk16:
@Jooho @andresllh Thanks for the context. I see the correct approach is a `kubectl patch` in `setup-kserve.sh` (CI-only) rather than modifying the production chart.

I can update this PR with that approach, or I can close this and let #5602 handle it. Let me know which you prefer.

andresllh:
@vivekk16 it makes more sense to have this fix on its own. I would just make the patch on this PR and get this merged asap to unblock others.

vivekk16:
@andresllh Done. I've pushed the changes. PTAL

Jooho:
This is a pr for doc: https://github.com/kserve/kserve/pull/5660

>@Jooho one question is why is CI not failing for that?

@pierDipi  If you're asking whether this did not fail because Helm and Kustomize are different, that's not really the case. We don't directly compare the two outputs.

Our process is that the output of kustomize build is automatically incorporated into Helm, so if a change is made in the Kustomize configuration, we generally assume it will be reflected in Helm as well. Kustomize is the source of truth.

If the question is why the CI itself didn't fail, I don't know for sure. My guess is that the issue happened to be resolved on the Kustomize side. In CI, we're essentially installing the same manifests using two different mechanisms: Helm and Kustomize and execute the same e2e tests.

test/scripts/gh-actions/setup-kserve.sh
```diff
@@ -62,6 +62,11 @@ if [[ $ENABLE_LLMISVC == "false" || $ENABLE_KSERVE_WITH_LLMISVC == "true" ]]; th
     export INSTALL_RUNTIMES=true
     ${REPO_ROOT}/hack/setup/infra/manage.kserve-helm.sh
     kustomize build config/overlays/test/s3-local-backend | kubectl apply --server-side --force-conflicts -f -
+    echo "Applying test env patches to ClusterServingRuntimes..."
```
Jooho: This is only set for the helm case. Why do we need this only for Helm?

Jooho (changes requested):
Please apply this change for both (helm/kustomize)

test/scripts/gh-actions/setup-kserve.sh
vivekk16: @Jooho The kustomize path already gets these env vars through the overlay patch file added in #5510:
`config/overlays/test/clusterresources/huggingfaceserver-tokio-patch.yaml`

That patch is applied at `kustomize build` time (line 78: `kubectl apply --server-side=true -k config/overlays/test/clusterresources`). So the kubectl patch is only needed for the helm path which doesn't go through kustomize overlays.

Jooho: Since we can handle this centrally in one place, I’d prefer to remove the existing env from yaml and have this script take care of it instead.

vivekk16: Alright
I'll make that change.

vivekk16: @Jooho Updated the changes. PTAL

cjohannsen-cloudera (approved):

PR #5664, opened 2026-06-11
feat(llmisvc): discover URLs for all matching listeners

bartoszmajsak:

Internal URL discovery for LLMInferenceService only advertised a single URL using the first (preferred) Gateway listener. Multi-protocol setups (e.g., HTTP + gRPC on different ports) couldn't be fully discovered by internal consumers reading `status.Addresses`.

This change generates internal URLs for every listener whose port is confirmed on the backing Service's declared ports. Listeners with unmatched ports are skipped - preventing broken URLs from heuristic port guessing. Services with no declared ports match all listeners for backward compatibility.

Also improves the dedup logic to use full-URL comparison (scheme + host + port + path) instead of the previous hostname + path check, which could suppress valid URLs when the same hostname appeared with different schemes or ports.

`DiscoverGatewayServiceHost` has been renamed to `DiscoverGatewayService` and now returns `(*corev1.Service, error)` instead of `(string, error)`. Only one internal caller - the signature change is safe.

The port matching is intentionally port-number-only (no protocol comparison) because Gateway listeners use L7 protocols (HTTP/HTTPS) while Service ports are L4 (TCP) - they operate at different layers.

**Checklist**:

```release-note
Advertise internal URLs for all matching Gateway listeners in LLMInferenceService, enabling multi-protocol discovery (e.g., HTTP + gRPC).
```


pierDipi (approved):

pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -372,24 +371,14 @@ func (r *LLMISVCReconciler) updateRoutingStatus(ctx context.Context, llmSvc *v1a
...
-			// external addresses.
-			//
-			// We prefer the path-based URL (/{ns}/{name}) over the model-routing
-			// root URL (/) for backward compatibility: status.URL was always the
-			// path-based URL before model-based routing was introduced, and
-			// existing clients may rely on that form.
-			llmSvc.Status.URL = preferPathBasedURL(discovered, llmSvc.Namespace, llmSvc.Name)
+			llmSvc.Status.URL = preferPathBasedURL(discovered, llmSvc.Namespace, llmSvc.Name, cfg.UrlScheme)
 		} else {
+			logger.Info("no public URL discovered")
```
lizzzcai: this log line should keep in line 374? Seems like it no longer fires for internal-only services.

pkg/controller/v1alpha2/llmisvc/router_discovery_status_url_test.go
```diff
@@ -0,0 +1,274 @@
...
+				),
+			},
+			services: []*corev1.Service{
+				backingService("gw-ns", "my-gateway",
+					corev1.ServicePort{Name: "http", Port: 80},
+					corev1.ServicePort{Name: "https", Port: 443},
+				),
+			},
+			urlScheme:         "https",
+			expectedStatusURL: "http://203.0.113.1/test-ns/test-svc",
```
lizzzcai, line 141: is this expected? maybe can add more comments in the code on the behavior.

bartoszmajsak: Yes, intentional. External URL selection deliberately skips scheme preference to preserve master behavior on upgrade. Adding a comment to make that clearer.

pkg/controller/v1alpha2/llmisvc/router.go
bartoszmajsak: Good catch - the log moved into the wrong branch. It should fire for all internal-only services, not just the zero-URLs case. 

Will move it back and reword to "no external URL discovered, using internal address" since the old message was misleading (we did discover URLs, just not public ones).

lizzzcai (approved):

PR #5667, opened 2026-06-11
feat(inferenceservice): expose model name and protocols in status

### LINKED ISSUE #5297: Expose protocol, model name, openapi URL in ISVC status
/kind feature

Expose the inference protocol version(s) and effective model name in `InferenceServiceStatus` so that consumers can construct working inference endpoints without external knowledge of KServe internals.

**Motivation**

Today, `status.url` provides only the host (e.g. `https://my-model.example.com`). To actually perform inference, a consumer must independently know:

1. The inference protocol (v1, v2, grpc-v2, etc.)
2. The effective model name (which may differ from `metadata.name`)
3. The correct path format for that protocol (e.g. `/v2/models/{name}/infer`)

None of this information is available in InferenceService status. The controller resolves the protocol and model name internally during reconciliation ([predictor.go#L358-L366](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L358-L366)), but only mutates `spec` in-memory -- the resolved values are never persisted to `status`. A consumer reading the ISVC from the API server cannot determine how to call the model without looking up the ServingRuntime separately.

This creates friction for:
- **New users** who deploy a model and don't know what to do with the URL they get back
- **UI/dashboard layers** that need to display a working inference endpoint but should not embed protocol-path logic
- **Automation and AI agents** (see #5295) that need structured metadata to drive inference operations

**Prior art**: #2256 / PR #2875 (May 2023) deliberately stripped the predict path from `status.address.url` to make it host-only. This was the right call -- a single baked-in URL cannot represent multiple protocols, doesn't help gRPC consumers, and only covers one endpoint type (predict) while users may also need explain, health, or metadata endpoints.

## Proposal: Add `supportedProtocols` and `modelName` to `ModelStatus`

Add two optional fields to `ModelStatus`:

```go
type ModelStatus struct {
    // (existing fields unchanged)

    // Inference protocols supported by the serving runtime (e.g. ["v2", "grpc-v2"])
    // +optional
    SupportedProtocols []InferenceServiceProtocol `json:"supportedProtocols,omitempty"`

    // The effective model name used in inference endpoint paths
    // +optional
    ModelName string `json:"modelName,omitempty"`
}
```

### How fields are populated

The predictor reconciler already resolves both values during `reconcileModel()`:

- **`supportedProtocols`**: Copied from the resolved `ServingRuntimeSpec.ProtocolVersions`. The runtime is already looked up and stored in `status.servingRuntimeName`; this avoids forcing consumers to perform that lookup themselves.
- **`modelName`**: From the existing `GetModelName()` helper ([utils.go#L115-L145](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L115-L145)), which resolves the effective model name from container args, env vars, or falls back to `isvc.Name`.

Both values are already computed in-memory. Persisting them to status requires no new watches or API calls.

### Why primitives instead of a pre-built URL

A pre-built inference URL has fundamental limitations:

| Problem | Detail |
|---------|--------|
| Protocols are plural | A runtime can serve v2 and grpc-v2 simultaneously. A single URL can only represent one. |
| gRPC has no HTTP path | `PredictPath()` returns `""` for gRPC protocols. A URL field is meaningless for gRPC consumers. |
| Only covers predict | Users may also need explain, health, metadata, or model-repository endpoints. |
| Composable is better | Given protocol + model name, a consumer can construct *any* endpoint using the well-defined [Open Inference Protocol](https://github.com/kserve/open-inference-protocol) path formats. |

### Example

Given this InferenceService status:

```yaml
status:
  url: "https://my-model.my-namespace.example.com"
  modelStatus:
    supportedProtocols: ["v2", "grpc-v2"]
    modelName: "my-model"
```

A consumer can construct any endpoint:

| Endpoint | Path |
|----------|------|
| Inference (v2) | `{url}/v2/models/{modelName}/infer` |
| Model metadata (v2) | `{url}/v2/models/{modelName}` |
| Server health (v2) | `{url}/v2/health/ready` |
| gRPC inference | Connect to `{url}` host via gRPC client |

For cluster-internal consumers, the same applies using `status.address.url` instead of `status.url`.

### Stretch goal: Surface OpenAPI spec URL

Some serving runtimes (e.g. vLLM, TorchServe) expose an OpenAPI spec at `/openapi.json`. A new optional field on `ServingRuntimeSpec` could declare this path, and the controller could surface it in status:

```go
// On ServingRuntimeSpec:
// +optional
OpenAPIPath string `json:"openApiPath,omitempty"`

// On ModelStatus:
// +optional
OpenAPIUrl string `json:"openApiUrl,omitempty"`
```

This would give consumers full self-describing API documentation (all endpoints, methods, request/response schemas) for the deployed model. However, there is no contract guaranteeing that a runtime exposes an OpenAPI endpoint, so this would be best-effort and opt-in via the ServingRuntime spec.

## Scope

- Purely additive (new optional fields), no backward compatibility risk
- No changes to existing `status.url` or `status.address.url` behavior
- Implementation is small: store two values that are already computed during reconciliation
- Lays groundwork for first-class inference support in tooling (#5295), dashboards, and CLI

Related: #5295 (MCP Server for KServe) proposes an AI-native operational interface that would directly benefit from these status fields. The protocol and model name are the minimum building blocks needed for any tool to help users perform inference without requiring deep KServe knowledge.

ItsVigneshMurugan:

`InferenceService.status` exposes only the host URL (e.g. `https://my-model.example.com`). To actually call the model, a consumer must independently know the inference protocol (v1, v2, grpc-v2, …) and the effective model name used in the endpoint path — neither of which is available in status. The controller resolves both internally during reconciliation but never persists them, so a consumer cannot construct an inference request without separately looking up the ServingRuntime.

This PR adds two optional fields to `ModelStatus`, populated in the predictor reconciler from values **already resolved in-memory** (no new watches or API calls):

- `modelName` — the effective model name from the existing `GetModelName()` helper.
- `supportedProtocols` — the backing ServingRuntime's `ProtocolVersions`.

With these, a consumer can build endpoint paths such as `/v1/models/<modelName>:predict` directly from status. As suggested in the issue, this PR is scoped to these two fields and leaves the OpenAPI-spec-URL stretch goal for a follow-up.

Fixes #5297

- [x] Verified end-to-end on a kind cluster (RawDeployment). Before the change, a deployed InferenceService's `status.modelStatus` exposed only `copies`/`states`/`transitionStatus`; after the change it also reports:
  ```json
  "modelName": "sklearn-iris",
  "supportedProtocols": ["v1", "v2"]
  ```
  matching the backing runtime's declared `protocolVersions`.
- [x] Updated the controller envtest scenarios (`controller_test.go`, `rawkube_controller_test.go`, `tests_common.go`) that assert the full status via `cmp.Diff`, so they now assert the populated `modelName`. These act as the regression test for the new behaviour. `go test ./pkg/controller/v1beta1/inferenceservice/...` and `./pkg/apis/serving/v1beta1/...` pass; `golangci-lint` reports 0 issues.
- [x] Regenerated CRDs, deepcopy, OpenAPI/swagger, and the Python SDK model.

**Special notes for reviewer**:

- `supportedProtocols` uses `+listType=atomic` (whole-list replace), matching how the controller sets it each reconcile.
- For custom predictors not backed by a ServingRuntime, `supportedProtocols` is empty and `modelName` falls back to the InferenceService name (via `GetModelName()`), so behaviour is unchanged for those.

**Checklist**:

**Release note**:
```release-note
InferenceService status now reports `status.modelStatus.modelName` and `status.modelStatus.supportedProtocols`, letting consumers construct inference endpoint paths without separately looking up the ServingRuntime.
```


vivekk16 (commented):
@ItsVigneshMurugan Please run `make precommit` before pushing the code.

ItsVigneshMurugan:
> @ItsVigneshMurugan Please run `make precommit` before pushing the code.

Ran `make precommit` locally and pushed.

PR #5671, opened 2026-06-12
fix(llmisvc): allow scheduler creation when workloads are rolling

vivekk16:

The `isWorkloadRolling` guard introduced in #5602 defers EPP deployment updates while workload pods are mid-rollout to prevent dropping in-flight streaming responses. However, it also blocks the scheduler's initial creation on new LLMInferenceService resources, creating a circular dependency where the scheduler is never created.

When a new LLMInferenceService is created, the workload deployments are immediately marked `Available=False` (pods still booting). If the scheduler reconcile runs after this condition is set, `isWorkloadRolling` returns `true` and the scheduler creation is permanently skipped. `propagateSchedulerDeploymentStatus` retries and fails with "Deployment not found" because the scheduler was never created in the first place.

The fix: only defer when the scheduler deployment already exists. If it doesn't exist, there's no running EPP to protect from a restart, so proceed with creation.

Fixes a regression introduced in #5602.


vivekk16:
PTAL @spolti

spolti:
Nice catch!

/approve

Issue #5335, opened 2026-04-02
Support `canaryTrafficPercent` for `RawDeployment` mode via Gateway API HTTPRoute weights


maskarb:
/kind feature

Here's the draft issue. Let me write it and show you before creating:

---

**Title:** Support `canaryTrafficPercent` for RawDeployment mode via Gateway API HTTPRoute weights

**Body:**

## Problem

`canaryTrafficPercent` is defined in `ComponentExtensionSpec` and works in Knative (Serverless) mode via Istio VirtualService revision weights. In RawDeployment mode, this field is **silently ignored** — there is no error, warning, or status message. Users who set `canaryTrafficPercent` on a RawDeployment InferenceService expect traffic splitting but get a full rollout instead.

Related issues:
- #4074 — User reported canary broken in RawDeployment
- #2649 — GitOps-friendly canary manifests
- #1324 — A/B/N rollouts (future extension)

## Proposal

Extend the existing `RawHTTPRouteReconciler` (gated behind `enableGatewayApi: true`) to support weighted `backendRefs` when `canaryTrafficPercent` is set.

### What already exists

- `RawHTTPRouteReconciler` creates HTTPRoutes for RawDeployment with Gateway API
- The factory in `reconcilers/factory.go` already selects between `RawIngressReconciler` and `RawHTTPRouteReconciler` based on `enableGatewayApi`
- `canaryTrafficPercent` is already in the API (`ComponentExtensionSpec`)

### What needs to change

1. **Deployment reconciler** — create a second "canary" Deployment when `canaryTrafficPercent` is set alongside a spec change. The stable Deployment continues running the previous model; the canary Deployment runs the new model.

2. **HTTPRoute reconciler** — extend `reconcilePredictorHTTPRoute()` to add weighted `backendRefs` pointing at both the stable and canary Services:
   ```yaml
   backendRefs:
   - name: isvc-predictor          # stable
     weight: 80
   - name: isvc-predictor-canary   # canary
     weight: 20
   ```

3. **Lifecycle management** — promote (canary → stable, delete canary, reset weights to 100/0), rollback (delete canary, reset weights), and status tracking (`latestRolledoutRevision`, `previousRolledoutRevision`, `traffic` array).

4. **Status reconciliation** — surface canary state in InferenceService `.status` (traffic percentages, canary readiness, canary model URI).

### Validated

[...]

PR #5672, opened 2026-06-12
feat: add canary rollout support for RawDeployment mode

maskarb:
## Summary
- Add `CanarySpec` list type for progressive model rollout with named canary deployments
- Add validation webhook enforcing canary constraints
- Add canary deployment reconciler with zero-restart promotion
- Add per-canary status reporting and aggregate `CanaryReady` condition
- Extract `buildPredictorResources` helper so stable and canary share the same annotation/podSpec/objectMeta pipeline
- Use `mergo.Merge` for canary predictor construction (overlay non-zero canary fields onto stable defaults)

## Design

Uses a named list of canaries (`[]CanarySpec`) on `InferenceServiceSpec`. Each canary gets its own Deployment and Service named `{isvc}-{name}-predictor`. Promotion means retargeting the stable predictor name to the canary — no pod restart, no redeployment.

Ref #5335

```yaml
spec:
  predictor:
    model:
      modelFormat: {name: sklearn}
    minReplicas: 4
  canary:
  - trafficPercent: 25
    predictor:
      name: v2
      model:
        modelFormat: {name: sklearn}
        storageUri: "gs://kfserving-examples/models/sklearn/1.0/model-2"
  - trafficPercent: 0
    predictor:
      name: experimental
      model:
        modelFormat: {name: sklearn}
        storageUri: "gs://kfserving-examples/models/sklearn/irisv2"
```

### API

- `CanarySpec`: `Predictor PredictorSpec` (with required `Name`), `TrafficPercent` (0-100)
- `CanaryStatus`: per-canary `Name`, `Ready`, `TrafficPercent`, `ModelStatus`
- `CanaryReady` aggregate condition (true when all canaries ready)
- `+listType=map` / `+listMapKey=name` via `predictor.name` for strategic merge patch support

### Canary predictor construction

Each canary predictor is built by overlaying the canary's non-zero fields onto a deep copy of the stable predictor using `mergo.Merge(WithOverride)`. This ensures canaries inherit stable defaults (runtime, tolerations, labels, etc.) while overriding model-specific fields. Autoscaling fields are zeroed after merge.

Both stable and canary paths use `buildPredictorResources` — a shared helper that builds annotations, pod spec, and object meta from `isvc.Spec.Predictor`. Future changes to annotation building or pod spec construction automatically apply to canary deployments.

### Replica scaling
- If `canary.predictor.minReplicas` is set: canary gets that count, stable is reduced by that amount
- If omitted: derived from `trafficPercent` ratio (e.g. 4 stable, 25% traffic → 1 canary, 3 stable)
- Total replicas always equals original `minReplicas`

### Validation rules
- Canary only in Standard (RawDeployment) mode
- `trafficPercent` 0-100, sum of all canaries <= 100
- `predictor.name` required, unique across canary entries, must not conflict with stable
- Canary must not set autoscaling fields (`maxReplicas`, `scaleTarget`, `scaleMetric`, `autoScaling`)
- Canary `minReplicas` must be less than stable `minReplicas`
- `predictor.name` change requires a model change (prevents accidental Deployment replacement)

### Promotion / Rollback
- **Promote:** set `spec.predictor.name` to the canary name and update `storageUri` to match. The canary Deployment becomes stable (same name, no pod restart). Remove canary entries — orphan cleanup deletes the old stable.
- **Rollback:** remove canary entry from list → orphan cleanup deletes its Deployment and Service, stable replicas restored.

### Follow-up work (not in this PR)
- **HTTPRoute traffic splitting:** canary Deployments and Services are created but traffic is not yet split between stable and canary at the ingress layer
- **Multi-node (WorkerSpec) canary:** leader+worker pod group scaling needs separate design — canary replica math doesn't map to multi-node topology
- **Canary ModelStatus:** per-canary `ModelStatus` (model loading state, failure info) is not yet populated — canary readiness is deployment-level only (`AvailableReplicas > 0`), so users don't get detailed model loading errors for canary variants
- **E2E tests:** automated end-to-end test coverage for canary lifecycle

## Test plan
- [x] `go build ./...` compiles
- [x] `go test ./pkg/apis/...` passes (11 validation test cases + 4 predictor name change tests)
- [x] CRD includes `canary` list under spec and `canaryStatuses` list under status
- [x] Manual testing on CRC: canary Deployments created with correct storageUri annotations
- [x] Manual testing: zero-restart promotion (canary pod UID survives, no new ReplicaSet)
- [x] Manual testing: replica scaling (25% → 1 canary, 3 stable from 4 total)
- [x] Manual testing: traffic bump (50% → 2 canary, 2 stable)
- [x] Manual testing: force stop scales down both stable and canary, resume restores them
- [x] Manual testing: CanaryReady condition shows "Stopped" reason on force stop


malhajfr:
@israel-hdez  @lizzzcai The PR is waiting on reviewer attention and a green CI run. Requesting your support to move it forward.

malhajfr:
@israel-hdez @lizzzcai could you please review to move forward for this PR ? 

@maskarb could you please check the failed tests  and re-run ?

pkg/apis/serving/v1beta1/inference_service_status.go
```diff
@@ -92,6 +96,19 @@ type ComponentStatusSpec struct {
 	Address *duckv1.Addressable `json:"address,omitempty"`
 }
 
+// CanaryStatus represents the observed state of a canary deployment.
+type CanaryStatus struct {
+	// Name of the canary variant (from predictor.name).
+	Name string `json:"name"`
+	// Ready indicates the canary deployment is available and serving traffic.
+	Ready bool `json:"ready"`
```
israel-hdez, line 104: It looks like `transformer` is out-of-scope. But I think canaries may require their own specific transformers. Thus, I would recommend making it specific, like `PredictorReady` to be future-proof.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -602,6 +683,30 @@ func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService
...
 }
 
+func validatePredictorNameChange(newIsvc *InferenceService, oldIsvc *InferenceService) error {
+	if oldIsvc.Spec.Predictor.Name == newIsvc.Spec.Predictor.Name {
+		return nil
+	}
+	oldURI := getPredictorStorageURI(oldIsvc.Spec.Predictor)
+	newURI := getPredictorStorageURI(newIsvc.Spec.Predictor)
+	if oldURI == newURI {
+		return errors.New("predictor.name change requires a model change (e.g. storageUri)")
```
israel-hdez, line 705: Why is this required?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -169,9 +172,87 @@ func validateInferenceService(isvc *InferenceService) (admission.Warnings, error
...
+			return fmt.Errorf("canary %q must not set autoScaling (canary uses fixed replicas)", canary.Predictor.Name)
+		}
+
+		if canary.Predictor.MinReplicas != nil && isvc.Spec.Predictor.MinReplicas != nil {
+			if *canary.Predictor.MinReplicas >= *isvc.Spec.Predictor.MinReplicas {
+				return fmt.Errorf("canary %q minReplicas (%d) must be less than stable predictor minReplicas (%d)",
+					canary.Predictor.Name, *canary.Predictor.MinReplicas, *isvc.Spec.Predictor.MinReplicas)
+			}
+		}
+	}
```
israel-hdez, line 259: I'm not sure I agree with all this. IMO canaries should auto-scale independently. Basically, behave almost like an individual ISVC. Thus, I'm somewhat confused about why to put a restriction over `MinReplicas`.

These validations are OK if we won't support auto-scaling in the initial iteration. Although, if we don't plan to support auto-scaling at all, I'm not very in favor or exposing fields that would never be used.

pkg/apis/serving/v1beta1/predictor.go
```diff
@@ -30,6 +30,12 @@ type PredictorImplementation interface{}
 // PredictorSpec defines the configuration for a predictor,
 // The following fields follow a "1-of" semantic. Users must specify exactly one spec.
 type PredictorSpec struct {
+	// Name optionally identifies this predictor variant. When set, the
+	// Deployment and Service are named {isvc}-{name}-predictor instead
+	// of the default {isvc}-predictor. Required when used in a canary entry.
+	// +optional
+	// +kubebuilder:validation:Pattern="^[a-z]([-a-z0-9]*[a-z0-9])?$"
+	Name string `json:"name,omitempty"`
```
israel-hdez, line 38: This adds complexity to the user. Can we omit?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -81,120 +85,121 @@ func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *
...
-		return ctrl.Result{}, err
-	}
-
 	predictor := isvc.Spec.Predictor.GetImplementation()
-
-	sourceURI := predictor.GetStorageUri()
-
-	// Knative does not support INIT containers or mounting, so we add annotations that trigger the
-	// StorageInitializer injector to mutate the underlying deployment to provision model data
-	// Only add annotations for single storage URI case. Multiple storage URIs are handled directly by reconcilers.
```
israel-hdez, line 126: This is a useful comment that should not be removed.

Also, in the previous and following lines, I see no reason for removing the comments that try to document the code:
```
// Add ModelStorageSpec annotations [...]
and
// Add agent annotations so mutator [...]
and
// Add confidential annotations [...]
etc.
```

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -81,120 +85,121 @@ func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *
...
+func (p *Predictor) Reconcile(ctx context.Context, isvc *v1beta1.InferenceService) (ctrl.Result, error) {
+	var workerPodSpec *corev1.PodSpec
+	var workerObjectMeta metav1.ObjectMeta
+	multiNodeEnabled := isvc.Spec.Predictor.WorkerSpec != nil
+	isvcGeneration := strconv.FormatInt(isvc.Generation, 10)
+
+	p.Log.V(1).Info("Predictor custom annotations", "annotations", p.inferenceServiceConfig.ServiceAnnotationDisallowedList)
+	p.Log.V(1).Info("Predictor custom labels", "labels", p.inferenceServiceConfig.ServiceLabelDisallowedList)
+
+	// Reconcile modelConfig (MMS only — canary does not use this)
```
israel-hdez: Should there be a validation in the webhook to require the main model when using `canary`? (to prevent it falling back to MMS)

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -81,120 +85,121 @@ func NewPredictor(client client.Client, clientset kubernetes.Interface, scheme *
...
+		workerObjectMeta, workerPodSpec, err = p.reconcileWorker(sRuntime, isvc, &podSpec, res.annotations, res.predictorAnnotations, isvcGeneration)
 		if err != nil {
 			isvc.Status.PropagateRawStatusWithMessages(v1beta1.PredictorComponent, v1beta1.InvalidGPUAllocation, err.Error(), corev1.ConditionFalse)
 			return ctrl.Result{}, err
 		}
 		objectMeta.Labels[constants.InferenceServiceGenerationPodLabelKey] = isvcGeneration
 	}
 
-	p.Log.Info("Resolved container", "container", predContainer, "podSpec", podSpec)
+	p.Log.Info("Resolved podSpec", "podSpec", podSpec)
```
israel-hdez: Either, revert the change, or change it to be more specific `Resolved main predictor container`

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -768,7 +776,20 @@ func (p *Predictor) reconcileRawDeployment(ctx context.Context, isvc *v1beta1.In
 		storageSpec = &modelStorageSpec.StorageSpec
 	}
 
-	r, err := raw.NewRawKubeReconciler(ctx, p.client, p.clientset, p.scheme, objectMeta, workerObjectMeta, &isvc.Spec.Predictor.ComponentExtensionSpec,
+	componentExt := isvc.Spec.Predictor.ComponentExtensionSpec
+	if len(isvc.Spec.Canary) > 0 && componentExt.MinReplicas != nil {
```
israel-hdez, line 795: For the _no autoscaler_ case, I see this OK. I saw this computation was discussed in #5335. However, given the explicit replica count is optional in the canaries, IMO the `status` of the ISVC must show the computed bounds/replicas of each workload.

When auto-scaling is enabled, as I mentioned in other comments, I think the main and the canaries should auto-scale individually. The auto-scaling behavior was not discussed in deep in #5335. I have to say that I disagree that auto-scaling is not valuable in the canaries, and also disagree that canaries aren't for production. It should be possible to do gradual promotion with the canary taking initially a small percentage of the load, and gradually taking all load to promote it. It is true that canaries are, ideally, short-lived but we cannot work with the assumption that this will always hold. The gradual promotion would not be doable if canaries does not support auto-scaling.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -848,3 +869,163 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
...
+	canaryPredictor.AutoScaling = nil
+	canaryPredictor.MinReplicas = &replicas
+
+	return canaryPredictor, nil
+}
+
+// Multi-node (WorkerSpec) is not supported for canary deployments. Multi-node
+// allocates leader+worker pod groups across GPU nodes for a single inference
+// instance — canary replica scaling doesn't map to that model. This should
+// be designed separately if needed.
```
israel-hdez: In addition to the out-of-scope, enhance the comment with the goal of the func.

Also, if canaries won't support multi-node, that suggest a check in the webhook (which I didn't see).

pkg/apis/serving/v1beta1/inference_service.go
```diff
@@ -33,6 +33,29 @@ type InferenceServiceSpec struct {
...
+	Canary []CanarySpec `json:"canary,omitempty"`
+}
+
+// CanarySpec defines a canary deployment for progressive rollout of a new
+// model version. The canary uses fixed replicas (no autoscaling).
+// The predictor.name field is required and used for Deployment/Service naming.
+type CanarySpec struct {
+	// Predictor spec for the canary variant. predictor.name is required.
+	// +required
+	Predictor PredictorSpec `json:"predictor"`
```
israel-hdez, line 51: I'm not sure using `PredictorSpec` is good here. This allows mixing model kinds; e.g. main model is ONNX and specifying OpenVINO runtime, while canary is scikit-learn with its accompanying runtime.

The changes to the reconciler seem to nicely handle the mixed configs. However, in #5335 it was mentioned that the ISVC case should deal with the simple canary case. Such case, IMO, is only for specifying a different model artifact with same model format and reusing runtime configurations. (i.e. IMO users should only specify the model URI of the canary)

Opening the possibility of a full configuration (this is almost like an ISVC inside an ISVC) adds some dimensions that I'm not sure an ISVC can cover without adding more complexity in the spec. For example, different runtimes can have different APIs... or even if same API, different way of exposing inputs/outputs, which will make impossible to transparently route requests to the different canaries.

pkg/apis/serving/v1beta1/predictor.go
maskarb: The `Name` field was added to enable zero-restart promotion.

When a canary is promoted, the goal is that the canary Deployment *becomes* the stable Deployment without restarting pods. If the canary Deployment is named something like `{isvc}-canary-predictor` and the stable is `{isvc}-predictor`, promotion would require renaming the Deployment - which Kubernetes doesn't support. You'd have to delete and recreate, which restarts pods and defeats the purpose.

With the `Name` field, the canary's deployment is `{isvc}-{name}-predictor`. We the user promotes the canary to stable, the `Name` should be transferred with the model change. This allows the stable to inherit the existing deployment (`{isvc}-{name}-predictor`) without a restart.

pkg/apis/serving/v1beta1/inference_service_status.go
maskarb: Sure, I can change this to `PredictorReady`.

pkg/apis/serving/v1beta1/inference_service_validation.go
maskarb: The `Name` is required to promote the canary to stable without triggering a restart. The name is inherently linked to the model, so they should change together. The validation simply prevents a user from accidentally changing the name which would change the names of deployments, etc.

See this comment about why the Name was added:
https://github.com/kserve/kserve/pull/5672#discussion_r3572351082

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: These were put back.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: This comment seems out of place anyway, so I removed it. And added a check in the validation.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: Made this more specific.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: Validation webhook was added to catch this.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: We will defer auto-scaling for future work.

pkg/apis/serving/v1beta1/inference_service_validation.go
maskarb: We will defer auto-scaling for future work.

pkg/apis/serving/v1beta1/inference_service.go
maskarb: I can definitely understand the hesitation to use the entire PredictorSpec, especially when that enables mixing runtimes. But at the same time, there are other fields here that may need to be tweaked for a canary model, e.g. the container spec, resources, etc. I lean towards using the full spec and adding validation to exclude things rather than building a new spec and potentially missing something. If we add to the predictor spec in the future, something that canary should use, then we'll need to add it in both places and I would rather avoid that.

e can add webhook validation to reject canary entries that specify a different model format than stable, which would prevent the ONNX-stable + sklearn-canary (or other) scenario you're describing.

VedantMahabaleshwarkar:
LGTM pending the open discussion threads with @israel-hdez

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -848,3 +882,159 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
...
+
+		replicas := canaryReplicaCount(isvc, canary)
+		canaryPredictor, err := buildCanaryPredictor(stablePredictor, *canary, replicas)
+		if err != nil {
+			return err
+		}
+
+		isvc.Spec.Predictor = canaryPredictor
+		res, err := p.buildPredictorResources(ctx, isvc, false)
+		isvc.Spec.Predictor = stablePredictor
```
mwaykole: This swap-and-restore of `isvc.Spec.Predictor` is fragile — if `buildPredictorResources` internally triggers a status update, logs with predictor details, or stores a reference to `isvc.Spec.Predictor` that outlives the call, it will see/capture the canary predictor instead of stable. Would it be cleaner to deep-copy the entire isvc for each canary iteration instead of mutating the shared object in-place?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -169,9 +172,95 @@ func validateInferenceService(isvc *InferenceService) (admission.Warnings, error
...
+			return fmt.Errorf("canary %q must not set scaleTarget (canary uses fixed replicas)", canary.Predictor.Name)
+		}
+		if ext.ScaleMetric != nil {
+			return fmt.Errorf("canary %q must not set scaleMetric (canary uses fixed replicas)", canary.Predictor.Name)
+		}
+		if ext.AutoScaling != nil {
+			return fmt.Errorf("canary %q must not set autoScaling (canary uses fixed replicas)", canary.Predictor.Name)
+		}
+
+		if canary.Predictor.MinReplicas != nil && isvc.Spec.Predictor.MinReplicas != nil {
```
mwaykole: This guard skips the check entirely when `stable.MinReplicas` is nil (which defaults to 1 at runtime). A canary could set `minReplicas: 10` and pass validation, then `canaryReplicaCount` returns 10 while stable gets `max(1-10, 1) = 1` — 11 total replicas with no warning. Should this fall back to the effective default (1) when `stable.MinReplicas` is nil?

pkg/controller/v1beta1/inferenceservice/components/predictor.go
```diff
@@ -848,3 +882,159 @@ func (p *Predictor) reconcileKnativeDeployment(ctx context.Context, isvc *v1beta
...
+
+func canaryReplicaCount(isvc *v1beta1.InferenceService, canary *v1beta1.CanarySpec) int32 {
+	if canary.Predictor.MinReplicas != nil {
+		return *canary.Predictor.MinReplicas
+	}
+	stableReplicas := int32(1)
+	if isvc.Spec.Predictor.MinReplicas != nil {
+		stableReplicas = *isvc.Spec.Predictor.MinReplicas
+	}
+	return int32(math.Ceil(float64(stableReplicas) * float64(canary.TrafficPercent) / 100))
```
mwaykole, line 894: `math.Ceil` ensures each canary gets at least 1 replica, which can inflate the total beyond the original `minReplicas`. E.g. five canaries at 1% each: `ceil(4*0.01) = 1` each, 5 total canary replicas, stable gets `max(4-5, 1) = 1` — total 6 instead of 4. If keeping a fixed replica budget matters, consider computing canary replicas proportionally and distributing the remainder to the last canary (or stable), similar to how weighted round-robin works.

mwaykole (commented):
A few observations on the canary reconciler and validation logic.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: Ok, this makes sense to me. I switched this to use a `DeepCopy`, and replace the predictor with the canary predictor spec in the `DeepCopy`. This will avoid incidental changes to the `status`.

pkg/apis/serving/v1beta1/inference_service_validation.go
maskarb: I updated this validation to account for the nil stable.MinReplicas. We'll assume the replicas are set to 1 in that case.

pkg/controller/v1beta1/inferenceservice/components/predictor.go
maskarb: If a trafficPercent is defined for a canary, then we need a deployment to route traffic to. So while yes, in your scenario we would exceed the original replica budget, by giving each canary a percentage, it needs to have a replica to serve traffic.

pkg/apis/serving/v1beta1/inference_service.go
spolti: I understand the pros and cons of having the same spec inside ot itself; however, I don't think it is a good approach. 
Limiting its update using webhook can increase the complexity of maintainability.
I am not sure how things will behave in case of duplicate fields, who takes precedence of who.

For container resources, it could inherit from the main isvc, otherwise, if the canary has different resources, what would happen when you promote the canary? This would trigger a reconcile, no?

But I am leaning more towards not having the spec inside the spec, we should allow users to override only what is really necessary, like the model version/location.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -169,9 +172,99 @@ func validateInferenceService(isvc *InferenceService) (admission.Warnings, error
...
+	if len(isvc.Spec.Canary) == 0 {
+		return nil
+	}
+
+	deploymentMode := isvc.Annotations[constants.DeploymentMode]
+	if deploymentMode != string(constants.Standard) {
+		return fmt.Errorf("canary is only supported in %s deployment mode", constants.Standard)
+	}
+
+	if isvc.Spec.Predictor.GetImplementation().GetStorageUri() == nil && isvc.Spec.Predictor.Model == nil {
```
spolti, line 193: We have some functions to validate the storageURI; isn't it better to call it directly? With it, if the URL is invalid, it will also prevent the deployment from moving forward.

pkg/apis/serving/v1beta1/inference_service.go
maskarb: If the canary needed those resources, then that would be copied into the stable. It would not trigger a reconciliation.

I'm open to changing this spec, but I need someone to indicate what it needs to be changed to.

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -169,9 +172,99 @@ func validateInferenceService(isvc *InferenceService) (admission.Warnings, error
...
+			return fmt.Errorf("canary %q predictor: %w", canary.Predictor.Name, err)
+		}
+		if err := utils.FirstNonNilError([]error{
+			canary.Predictor.GetImplementation().Validate(),
+			canary.Predictor.GetExtensions().Validate(),
+		}); err != nil {
+			return fmt.Errorf("canary %q predictor: %w", canary.Predictor.Name, err)
+		}
+
+		if canary.Predictor.WorkerSpec != nil {
```
spolti, line 231: May we move it to the very beginning?

pkg/apis/serving/v1beta1/inference_service_validation.go
spolti: I am not sure if allowing autoscale on canary deployments is a good approach, because:
- Canaries are supposed to be deterministic experiments, easy to troubleshoot, etc.
- cost/risk control
- traffic exposure: if we set 10%, how exactly should we control how new replicas will be created?

pkg/apis/serving/v1beta1/inference_service_validation.go
```diff
@@ -169,9 +172,99 @@ func validateInferenceService(isvc *InferenceService) (admission.Warnings, error
...
+		if names[canary.Predictor.Name] {
+			return fmt.Errorf("canary predictor.name %q is duplicated", canary.Predictor.Name)
+		}
+		names[canary.Predictor.Name] = true
+
+		if canary.Predictor.Name == isvc.Spec.Predictor.Name {
+			return fmt.Errorf("canary predictor.name %q conflicts with stable predictor name", canary.Predictor.Name)
+		}
+
+		if canary.TrafficPercent < 0 || canary.TrafficPercent > 100 {
```
spolti, line 216: what traffic == 0 means?

pkg/apis/serving/v1beta1/inference_service_validation_test.go
```diff
@@ -2327,3 +2327,341 @@ func TestValidateConfidential(t *testing.T) {
...
+	t.Run("Name change with model change accepted (promotion)", func(t *testing.T) {
+		oldISVC := makeISVC("", "gs://test/model-v1")
+		newISVC := makeISVC("v2", "gs://test/model-v2")
+		_, err := validator.ValidateUpdate(t.Context(), oldISVC, newISVC)
+		g.Expect(err).ShouldNot(gomega.HaveOccurred())
+	})
+
+	t.Run("Name change without model change rejected", func(t *testing.T) {
+		oldISVC := makeISVC("", "gs://test/model-v1")
+		newISVC := makeISVC("v2", "gs://test/model-v1")
```
spolti, line 2649: I see that this is easier; however, a name change does not guarantee that the model is updated.
I am wondering, for scenarios like constant training and deployment as canary deployment, without model name update, would be not allowed by this approach.

Maybe there is a better approach to determine if the model is different? Maybe require the model sha or add another field for it?

pkg/apis/serving/v1beta1/inference_service_validation.go
spolti: My comment is addressed after the discussion we had internally.

pkg/apis/serving/v1beta1/inference_service_validation.go
spolti: we would need to improve this validation to support multiple URIs.

VedantMahabaleshwarkar (approved):
/lgtm

PR #5673, opened 2026-06-12
fix(llmisvc): render llmisvc resource namespaces from release namespace

MohanKumar21:

LLMInferenceService resources hardcoded `namespace: kserve`, ignoring the install namespace. When KServe runs in any other namespace this breaks two things:

1. **`kserve-runtime-configs` chart** emitted the well-known `LLMInferenceServiceConfig` presets (scheduler, router-route, template, decode/prefill/worker-data-parallel) with a hardcoded `namespace: kserve`. The llmisvc controller resolves these presets from its own namespace (`POD_NAMESPACE`), so in any other namespace every `LLMInferenceService` fails preset combination (no scheduler Deployment, no managed HTTPRoute) with no obvious error.

2. **llmisvc CRD conversion webhook** hardcoded the webhook `clientConfig.service.namespace` and the `cert-manager.io/inject-ca-from` annotation to `kserve`, so v1alpha1↔v1alpha2 conversion pointed at a non-existent service when installed elsewhere.

Fixes # NA

- [x] **Test A — runtime-configs presets follow release namespace**
  ```
  helm template krc charts/kserve-runtime-configs --namespace my-kserve-ns --set kserve.llmisvcConfigs.enabled=true \
    | grep -A2 'kind: LLMInferenceServiceConfig'
  ```
  Result: all 8 presets render `namespace: my-kserve-ns` (was `kserve`).

- [x] **Test B — CRD conversion webhook follows release namespace (Helm path)**
  ```
  helm template t charts/kserve-llmisvc-crd --namespace inference-gateway
  helm template t charts/kserve-llmisvc-crd-minimal --namespace inference-gateway
  ```
  Result: conversion webhook `service.namespace: 'inference-gateway'` and `cert-manager.io/inject-ca-from: 'inference-gateway/llmisvc-serving-cert'`.

- Logs

- This PR does not change any image versions.
- The generated CRD chart templates use **quoted** Helm templates (`'{{ .Release.Namespace }}'`). This is required so the source kustomize patches stay valid YAML and survive `make manifests` — an unquoted `{{ ... }}` causes `kustomize build` to fail with `unable to parse SM or JSON patch`. Quoting is already the established pattern in this repo (e.g. `namespace: '{{ .GlobalConfig.IngressGatewayNamespace }}'`).
- The real source of truth is `config/crd/{full,minimal}/llmisvc/*conversion_webhook_patch.yaml`; the four `charts/kserve-llmisvc-crd[-minimal]/templates/...` files are generated outputs.

**Checklist**:

**Release note**:
```release-note
Fix LLMInferenceService presets and CRD conversion webhook to use the install namespace instead of a hardcoded `kserve`, so KServe llmisvc works when installed in any namespace.
```


MohanKumar21:
@cjohannsen-cloudera @sivanantha321 Can you retrigger the GitHub workflows? Some tests failed due to a network error with Hugging Face while downloading models. Thanks

robertobarreda:
fixes https://github.com/kserve/kserve/issues/5325

PR #5675, opened 2026-06-15
fix(llmisvc): only inject lora-affinity-scorer into default config

andresllh:
Moves the \`lora-affinity-scorer\` injection from \`mutateSchedulerConfig\` into \`schedulerConfigText\` so the scorer only appears in the auto-generated default \`EndpointPickerConfig\`, not in user-provided configs. Also extends the injection to P/D disaggregation profiles (prefill and decode), where it is equally applicable since the scorer operates on vLLM adapter metrics exposed by all pods regardless of role.

Previously (#5655) the scorer was injected via \`mutateSchedulerConfig\`, which runs on any active config — including configs provided via \`spec.router.scheduler.config.inline\`, a ConfigMap ref, or preserved from an existing deployment. This gave users no way to opt out or override the scorer placement.

Now the scorer is part of the generated default YAML only when LoRA adapters are specified (\`len(spec.model.lora.adapters) > 0\`). If the user supplies their own scheduler config through any mechanism, that config is used untouched.

**Which issue(s) this PR fixes** *(optional)*:
Follow-up to #5655 (addresses review feedback from @pierDipi)

- [x] \`TestSchedulerConfigTextLoRA\` — covers no LoRA, nil LoRA, empty adapters (no scorer), and adapters present (scorer included with weight 4 as first profile entry)
- [x] \`TestSchedulerConfigTextPDLoRA\` — covers P/D without LoRA (no scorer) and P/D with LoRA adapters (scorer in both prefill and decode profiles)
- [ ] Manual validation via e2e with LoRA adapters configured

\`withLoraAffinityScorer\` and its \`mutateSchedulerConfig\` call are removed entirely. The \`loraAffinityScorerPlugin\` constant is still used via \`fmt.Sprintf\` inside \`schedulerConfigText\` to avoid hardcoding the string.

**Checklist**:

**Release note**:
```
Auto-enable lora-affinity-scorer in the default scheduler config when LoRA adapters are specified, including in P/D disaggregation profiles
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/scheduler.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler.go
@@ -502,10 +502,6 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 			}
 		}
 
-		if err := mutateSchedulerConfig(ctx, d, withLoraAffinityScorer(llmSvc.Spec.Model.LoRA != nil && len(llmSvc.Spec.Model.LoRA.Adapters) > 0)); err != nil {
-			return d, fmt.Errorf("failed to inject lora-affinity-scorer into scheduler config: %w", err)
-		}
-
 		// v0.7.0 migrations: version-gated (>= 0.7.0) because the v0.6 binary
 		// does not recognize the new plugin names. The v0.7 binary deprecates
 		// the old names but still accepts them.
@@ -573,7 +569,12 @@ schedulingProfiles:
   - pluginRef: max-score-picker
 `
 	default:
-		return `
+		var loraPlugin, loraProfileEntry string
+		if llmSvc.Spec.Model.LoRA != nil && len(llmSvc.Spec.Model.LoRA.Adapters) > 0 {
+			loraPlugin = fmt.Sprintf("- type: %s\n", loraAffinityScorerPlugin)
+			loraProfileEntry = fmt.Sprintf("  - pluginRef: %s\n    weight: 4\n", loraAffinityScorerPlugin)
+		}
+		return fmt.Sprintf(`
 apiVersion: inference.networking.x-k8s.io/v1alpha1
 kind: EndpointPickerConfig
 plugins:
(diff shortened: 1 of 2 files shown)


andresllh:
The `test-qpext (kustomize)` failure is unrelated to this PR — it failed at the "Install KServe" infrastructure step, not in any test logic. All other checks pass including Go Test, Precommit, and the LLMInferenceService E2E suite. Could a maintainer re-run the failed job?

andresllh:
**For reviewers — design summary**

The core change is moving where the `lora-affinity-scorer` gets added:

| | Before (#5655) | After (this PR) |
|---|---|---|
| **Where** | `mutateSchedulerConfig` in `expectedSchedulerDeployment` | `schedulerConfigText` (default case only) |
| **Runs on** | Any active config, including user-provided ones | Only the auto-generated default config |
| **User override** | Not possible — scorer always injected | Fully respected — user configs untouched |

The key insight is `preserveSchedulerConfig`'s priority chain: inline config → template args → existing deployment config → **fresh default** (`schedulerConfigText`). The scorer now only appears in that last case, so any user-supplied config takes precedence automatically with no extra logic needed.

`withLoraAffinityScorer` is removed entirely — the logic is now just a conditional inside `schedulerConfigText` using `fmt.Sprintf` with the `loraAffinityScorerPlugin` constant to avoid hardcoding the string.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -573,6 +569,28 @@ schedulingProfiles:
...
+- name: default
+  plugins:
+  - pluginRef: %s
+    weight: 4
+  - pluginRef: queue-scorer
+    weight: 2
+  - pluginRef: prefix-cache-scorer
+    weight: 3
+  - pluginRef: max-score-picker
+`, loraAffinityScorerPlugin, loraAffinityScorerPlugin)
```
pierDipi: could we reduce duplication with the "else" below?

are we skipping P/D because it doesn't supported P/D case?

andresllh: Both addressed in the latest commits:

1. **Duplication**: reduced by extracting the conditional YAML fragments into variables (`loraPlugin` and `loraProfileEntry`) and using a single `fmt.Sprintf` template rather than duplicating the full config string.

2. **P/D**: not intentionally skipped — it was an oversight. Looking at the [lora-affinity-scorer implementation](https://github.com/llm-d/llm-d-router/blob/main/pkg/epp/framework/plugins/scheduling/scorer/loraaffinity/lora_affinity.go), it scores pods based on `ActiveModels`/`WaitingModels` metrics exposed by vLLM regardless of pod role, so it's equally applicable to prefill and decode pods. Updated to inject the scorer into both P/D profiles when LoRA adapters are configured.

pierDipi (approved):

PR #5678, opened 2026-06-15
feat: add latency predictor sidecar injection via plugin detection

RishabhSaini:
## Summary

Detects `predicted-latency-producer` in the EPP inline config and auto-injects a well-known `LLMInferenceServiceConfig` that provides training-server and prediction-server sidecar containers. Follows the existing well-known config pattern used by scheduler, router, prefill, and worker configs.

No CRD API changes. No hardcoded defaults in the controller.

### How it works

1. Controller detects `predicted-latency-producer` plugin in the resolved scheduler config (`combineBaseRefsConfig`)
2. Auto-injects the well-known `kserve-config-llm-scheduler-latency-predictor` config
3. The well-known config provides everything: sidecar containers, images, probes, resources, securityContext, env vars (including EPP env vars via strategic merge on `main` container), and emptyDir volumes
4. Users override any default via their own `LLMInferenceServiceConfig` and `baseRefs`

### Minimal usage

```yaml
spec:
  router:
    scheduler:
      config:
        inline: |
          plugins:
          - type: predicted-latency-producer
          - type: latency-scorer
          - type: weighted-random-picker
```

### Custom images (optional)

```yaml
spec:
  baseRefs:
  - name: my-latency-predictor-config
```

### Changes

- `config_merge.go` — detect plugin in resolved spec, auto-inject well-known config ref
- `scheduler_latency_predictor.go` — `hasLatencyProducerInSpec` and `hasPluginType` detection functions
- `charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml` — well-known `LLMInferenceServiceConfig` with sidecar containers, env vars, probes, resources, securityContext, volumes

## Test plan
- [x] 4 unit tests: plugin detection (present, absent, nil config, nil inline)
- [x] `go build ./...`, `go vet ./...` pass
- [x] Existing `TestWellKnownConfigResolver_Attach` covers versioning annotation
- [x] Existing `TestMergeSpecs` covers env var merge on containers

### DIFF
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -2076,6 +2076,192 @@ spec:
 apiVersion: serving.kserve.io/v1alpha2
 kind: LLMInferenceServiceConfig
 metadata:
+  name: kserve-config-llm-scheduler-latency-predictor
+  namespace: kserve
+spec:
+  router:
+    scheduler:
+      template:
+        containers:
+        - env:
+          - name: PREDICTION_SERVER_URL
+            value: http://localhost:8001
+          - name: TRAINING_SERVER_URL
+            value: http://localhost:8000
+          - name: LATENCY_MAX_SAMPLE_SIZE
+            value: "10000"
+          - name: LATENCY_MAX_CONCURRENT_DISPATCHES
+            value: "36"
+          - name: LATENCY_COALESCE_WINDOW_MS
+            value: "1"
+          name: main
+        - env:
+          - name: LATENCY_RETRAINING_INTERVAL_SEC
+            value: "10"
+          - name: LATENCY_MIN_SAMPLES_FOR_RETRAIN
+            value: "100"
+          - name: LATENCY_TTFT_MODEL_PATH
+            value: /models/ttft.joblib
+          - name: LATENCY_TPOT_MODEL_PATH
+            value: /models/tpot.joblib
+          - name: LATENCY_TTFT_SCALER_PATH
+            value: /models/ttft_scaler.joblib
+          - name: LATENCY_TPOT_SCALER_PATH
+            value: /models/tpot_scaler.joblib
+          - name: LATENCY_TTFT_GATED_MODEL_PATH
+            value: /models/ttft_gated.joblib
+          - name: LATENCY_TPOT_GATED_MODEL_PATH
+            value: /models/tpot_gated.joblib
+          - name: LATENCY_MODEL_TYPE
+            value: xgboost
+          - name: LATENCY_MAX_TRAINING_DATA_SIZE_PER_BUCKET
+            value: "500"
+          - name: LATENCY_OBJECTIVE_TYPE
+            value: mean
+          image: registry.k8s.io/gateway-api-inference-extension/latency-training-server:v1.5.0
+          imagePullPolicy: IfNotPresent
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8000
+            initialDelaySeconds: 30
+            periodSeconds: 20
+          name: training-server
+          ports:
+          - containerPort: 8000
+            name: training-port
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 8000
+            initialDelaySeconds: 45
+            periodSeconds: 10
+          resources:
+            limits:
+              cpu: 4000m
+              memory: 8Gi
+            requests:
+              cpu: 2000m
+              memory: 4Gi
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+              - ALL
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+          startupProbe:
+            failureThreshold: 30
+            httpGet:
+              path: /healthz
+              port: 8000
+            periodSeconds: 10
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+          - mountPath: /models
+            name: training-server-storage
+          - mountPath: /tmp
+            name: training-server-tmp
+        - env:
+          - name: TRAINING_SERVER_URL
+            value: http://localhost:8000
+          - name: LATENCY_MODEL_TYPE
+            value: xgboost
+          - name: PREDICT_HOST
+            value: 0.0.0.0
+          - name: PREDICT_PORT
+            value: "8001"
+          - name: LOCAL_TTFT_MODEL_PATH
+            value: /server_models/ttft.joblib
+          - name: LOCAL_TPOT_MODEL_PATH
+            value: /server_models/tpot.joblib
+          - name: LOCAL_TTFT_SCALER_PATH
+            value: /server_models/ttft_scaler.joblib
+          - name: LOCAL_TPOT_SCALER_PATH
+            value: /server_models/tpot_scaler.joblib
+          - name: LOCAL_TTFT_GATED_MODEL_PATH
+            value: /server_models/ttft_gated.joblib
+          - name: LOCAL_TPOT_GATED_MODEL_PATH
+            value: /server_models/tpot_gated.joblib
+          - name: UVICORN_WORKERS
+            value: "28"
+          - name: OMP_NUM_THREADS
+            value: "1"
+          - name: MODEL_SYNC_INTERVAL_SEC
+            value: "30"
+          - name: LATENCY_OBJECTIVE_TYPE
+            value: mean
+          image: registry.k8s.io/gateway-api-inference-extension/latency-prediction-server:v1.5.0
+          imagePullPolicy: IfNotPresent
+          livenessProbe:
+            failureThreshold: 5
+            httpGet:
+              path: /healthz
+              port: 8001
+            initialDelaySeconds: 15
+            periodSeconds: 15
+            timeoutSeconds: 5
+          name: prediction-server
+          ports:
+          - containerPort: 8001
+            name: predict-port
+          readinessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /readyz
+              port: 8001
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            timeoutSeconds: 5
+          resources:
+            limits:
+              cpu: 28000m
+              memory: 8Gi
+            requests:
+              cpu: 8000m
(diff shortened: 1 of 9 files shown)


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -314,6 +314,76 @@ type SchedulerSpec struct {
...
+// LatencyPredictorPredictionSpec extends LatencyPredictorServerSpec with multi-instance support.
+type LatencyPredictorPredictionSpec struct {
+	LatencyPredictorServerSpec `json:",inline"`
+
+	// Count is the number of prediction server instances to run in the pod.
+	// Each gets a separate port starting from Port (default 8001).
+	// +optional
+	// +kubebuilder:default=1
+	Count *int32 `json:"count,omitempty"`
 }
```
pierDipi: can we consider having EPP/router configuration be the source of truth for this? instead of having a separate side path to configure and override a sidecar configuration that is inconsistent with the other sidecars?

for example, I'm thinking when `predicted-latency-producer` is part of the scheduler configuration (or any other plugin that needs a sidecar), we could inject the sidecar through a well-known llmisvcconfig (so that we can prefix-based stable versioning that prevents restarts on upgrades)

RishabhSaini: I initially considered detection from the plugin config but went with an explicit CRD field for discoverability and per-service customization. Your point about consistency with the tokenizer sidecar pattern and LLMInferenceServiceConfig versioning makes sense though. 
Will rework to detect predicted-latency-producer and use LLMInferenceServiceConfig for defaults.

pkg/controller/v1alpha2/llmisvc/scheduler_latency_predictor.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	predictedLatencyProducerPlugin = "predicted-latency-producer"
+)
+
+// hasLatencyProducerInSpec checks the LLMInferenceService spec's inline config.
+// NOTE: This only checks Config.Inline, not Config.Ref (ConfigMap-based config).
```
RishabhSaini, line 31: @pierDipi 

`hasLatencyProducerInSpec` only checks `Config.Inline`, not `Config.Ref`. The ConfigMap Ref is resolved later in `expectedSchedulerDeployment`, after `combineBaseRefsConfig` has already run so we can't check it at the point where well-known configs are injected.

The implication: 
if a user specifies predicted-latency-producer via a ConfigMap reference instead of inline, the well-known config won't be auto-injected. The EPP would start with the plugin configured but no training/prediction sidecar containers the plugin would log errors and fall back to composite scoring (no latency-based routing).

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
```diff
@@ -2854,3 +2854,190 @@ spec:
     - name: tls-certs
       secret:
         secretName: '{{ ChildName .ObjectMeta.Name `-kserve-self-signed-certs` }}'
+
+---
+apiVersion: serving.kserve.io/v1alpha2
+kind: LLMInferenceServiceConfig
```
pierDipi: we need to add this to https://github.com/kserve/kserve/tree/master/config/llmisvcconfig (and regenerate the charts)

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
```diff
@@ -2854,3 +2854,190 @@ spec:
...
+          - name: TRAINING_SERVER_URL
+            value: http://localhost:8000
+          - name: LATENCY_MAX_SAMPLE_SIZE
+            value: "10000"
+          - name: LATENCY_MAX_CONCURRENT_DISPATCHES
+            value: "36"
+          - name: LATENCY_COALESCE_WINDOW_MS
+            value: "1"
+        - name: training-server
+          image: registry.k8s.io/gateway-api-inference-extension/latency-training-server:v1.5.0
```
pierDipi: interesting, this and predictor didn't move to llm-d?

pkg/controller/v1alpha2/llmisvc/scheduler_latency_predictor.go
pierDipi: for now in the 368 line block in config_merge we could re-run `hasLatencyProducerInSpec` and emit an event (with `r.EventRecorder`) on the llmisvc with that warning?

pierDipi: and let's track it as an issue to improve as a follow up

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
RishabhSaini: ghcr.io/llm-d/llm-d-latency-predictor-training-server-dev:latest
ghcr.io/llm-d/llm-d-latency-predictor-prediction-server-dev:latest

There are dev images (tagged latest + commit SHA). No versioned release tags yet.

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
RishabhSaini: ok added

pkg/controller/v1alpha2/llmisvc/scheduler_latency_predictor.go
RishabhSaini: added the warning: https://github.com/kserve/kserve/pull/5678/changes#diff-58dc6c7e777d948196095d683226e127b5a59004af289183e5790b164fbe78ceR407

charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
RishabhSaini: an 0.8.0 RC1 is available: https://github.com/llm-d/llm-d-latency-predictor/releases/tag/v0.8.0-rc.1

RishabhSaini: will switch to 0.8.0 when a candidate is chosen in a followup PR

pierDipi (commented):
Can we add some integration tests similar to others in `*_int_test.go`? and if possible (given the resource requirements might not be possible to run on GH actions) a cpu-based smoke e2e test

config/llmisvcconfig/config-llm-scheduler-latency-predictor.yaml
```diff
@@ -0,0 +1,185 @@
...
+            value: mean
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+          - name: training-server-storage
+            mountPath: /models
+          - name: training-server-tmp
+            mountPath: /tmp
+        - name: prediction-server
+          image: registry.k8s.io/gateway-api-inference-extension/latency-prediction-server:v1.5.0
```
zdtsw: but we are going to  use image from  registry.k8s.io/gateway-api-inference-extension ?

RishabhSaini: I will switch to 0.8.0 once we have a release candidate: https://github.com/llm-d/llm-d-latency-predictor/releases/tag/v0.8.0-rc.1

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
```diff
@@ -0,0 +1,182 @@
...
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(envTest.Client.Create(ctx, latencyPredictorConfig)).To(Succeed())
```
pierDipi: @bartoszmajsak aren't we creating well-known configs as part of the setup? is this needed?

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
```diff
@@ -0,0 +1,182 @@
...
+			)
+			// Set the scheduler template with a main container so the deployment can be created.
+			llmSvc.Spec.Router.Scheduler.Template = &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "main",
+						Image: "ghcr.io/llm-d/llm-d-inference-scheduler:v0.7.1",
+					},
+				},
+			}
```
pierDipi: this is enabled through WithManagedScheduler() and WithSchedulerConfigInline()
```suggestion
```

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
```diff
@@ -0,0 +1,182 @@
...
+`),
+			)
+			llmSvc.Spec.Router.Scheduler.Template = &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "main",
+						Image: "ghcr.io/llm-d/llm-d-inference-scheduler:v0.7.1",
+					},
+				},
+			}
```
pierDipi:
```suggestion
```

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
RishabhSaini: updated

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
RishabhSaini: updated

pkg/controller/v1alpha2/llmisvc/controller_latency_predictor_int_test.go
RishabhSaini: right, SharedConfigPresets creates them

hexfusion:
@pierDipi do you have any further concerns here?

config/llmisvcconfig/config-llm-scheduler-latency-predictor.yaml
RishabhSaini: ghcr.io/llm-d/llm-d-latency-predictor-prediction-server:v0.8.0 exists now. Switching and updated

pierDipi (approved):

Jooho (approved):

PR #5680, opened 2026-06-16
fix(charts/kserve-localmodel-resources): remove duplicated nodeSelector

aslafy-z:

The `kserve-localmodelnode-agent` DaemonSet had `nodeSelector: {kserve/localmodel: worker}` hardcoded in the chart's `files/resources.yaml`. The same selector is injected by `files/daemonset-patch.yaml` from `.Values.kserve.localmodelnode.controller.nodeSelector` (default `{kserve/localmodel: worker}`). Since the base and the patch are deep-merged, the hardcoded value was redundant and could not be cleared by overriding the value.

`files/resources.yaml` is generated by `hack/setup/scripts/generate_chart_manifests.sh`, so editing it directly would not survive regeneration. Instead, this PR strips the selector at generation time:

- Adds a `config/overlays/charts-localmodel` Kustomize overlay that wraps `config/localmodels` and removes `/spec/template/spec/nodeSelector` from the agent DaemonSet.
- Points the generation script at the overlay for the chart's `files/resources.yaml`.
- Regenerates `files/resources.yaml` (only the two selector lines are removed).

Direct Kustomize builds of `config/localmodels` are unaffected and keep the selector; only the Helm chart input changes, where the selector comes from values.

Note: to clear the selector, set the value to `null`. An empty map (`{}`) does not clear it because Helm merges it with the chart's default.

- [x] Rerunning `hack/setup/scripts/generate_chart_manifests.sh` reproduces the committed `files/resources.yaml` byte for byte and is stable across reruns
- [x] Default `helm template` render is identical to master, the agent DaemonSet still carries `nodeSelector: {kserve/localmodel: worker}` (from the values patch only)
- [x] `--set kserve.localmodelnode.controller.nodeSelector.foo=bar` merges with the default
- [x] `--set kserve.localmodelnode.controller.nodeSelector=null` clears the selector, which was not possible before since the hardcoded base value was merged back in

**Release note**:
```release-note
NONE
```

### DIFF
--- a/charts/kserve-localmodel-resources/files/resources.yaml
+++ b/charts/kserve-localmodel-resources/files/resources.yaml
@@ -372,8 +372,6 @@ spec:
         - mountPath: /mnt/models
           name: models
           readOnly: false
-      nodeSelector:
-        kserve/localmodel: worker
       securityContext:
         runAsNonRoot: true
         seccompProfile:
--- a/config/overlays/charts-localmodel/kustomization.yaml
+++ b/config/overlays/charts-localmodel/kustomization.yaml
@@ -1,0 +1,18 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../localmodels
+
+# Used by hack/setup/scripts/generate_chart_manifests.sh to build the
+# kserve-localmodel-resources chart's files/resources.yaml. The chart sets the
+# DaemonSet nodeSelector from .Values.kserve.localmodelnode.controller.nodeSelector
+# (files/daemonset-patch.yaml). The hardcoded selector must not remain in the
+# base: Helm would merge it back in, making it impossible to override or clear.
+patches:
+- target:
+    kind: DaemonSet
+    name: kserve-localmodelnode-agent
+  patch: |-
+    - op: remove
+      path: /spec/template/spec/nodeSelector
--- a/hack/setup/scripts/generate_chart_manifests.sh
+++ b/hack/setup/scripts/generate_chart_manifests.sh
@@ -25,7 +25,7 @@ kustomize build ${REPO_ROOT}/config/storagecontainers > ${REPO_ROOT}/charts/kser
 echo "✅ Built storagecontainer resources"
 
 # LocalModel and Common
-kustomize build ${REPO_ROOT}/config/localmodels > ${REPO_ROOT}/charts/kserve-localmodel-resources/files/resources.yaml
+kustomize build ${REPO_ROOT}/config/overlays/charts-localmodel > ${REPO_ROOT}/charts/kserve-localmodel-resources/files/resources.yaml
 
 # Generate values.yaml from common sections
 echo "Generating values.yaml files from common sections..."


aslafy-z:
Ping @Jooho, can you give a look or assign someone else? Thank you

Bafff:
Hi @aslafy-z, I reviewed this PR while investigating the same selector behavior. I reproduced it, and your change works as described right now, but it edits a generated file, so the next manifest-generation run will restore the hardcoded selector. Details below; everything reproduced against your diff applied to the `v0.20.0-rc0` tree.

## What I verified works

- Default render is unchanged: the DaemonSet still carries `nodeSelector: {kserve/localmodel: worker}`, now injected only by `files/daemonset-patch.yaml` from values.
- `--set kserve.localmodelnode.controller.nodeSelector=null` now clears the selector, which was impossible before because the hardcoded base value was merged back in.

One clarification for the PR description: to clear the selector, use `null`. An empty map (`nodeSelector: {}` in a values file, or `--set-json '...nodeSelector={}'`) does not clear it, because Helm merges it with the chart's default map. Might be worth a one-line comment in `values.yaml`.

## Issue: the edited file is generated

`charts/kserve-localmodel-resources/files/resources.yaml` is produced by `hack/setup/scripts/generate_chart_manifests.sh`:

```bash
# hack/setup/scripts/generate_chart_manifests.sh (line 61)
kustomize build ${REPO_ROOT}/config/localmodels > ${REPO_ROOT}/charts/kserve-localmodel-resources/files/resources.yaml
```

The hardcoded selector comes from `config/localmodelnodes/manager.yaml` (included via `config/localmodels`). I ran the script on a tree with this PR applied: the `nodeSelector` block comes right back into `files/resources.yaml`, and the `null` override stops working again.

Removing the selector from `config/localmodelnodes/manager.yaml` would also change existing Kustomize builds that include LocalModel: the agent would no longer be restricted to nodes labeled `kserve/localmodel=worker`.

## Suggested durable fix

Strip the selector only for the chart, with a chart-generation overlay, so Kustomize builds keep their behavior and the generated file stays direct `kustomize build` output:

```yaml
# config/overlays/charts-localmodel/kustomization.yaml (new file)
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- ../../localmodels
patches:
- target:
    kind: DaemonSet
    name: kserve-localmodelnode-agent
  patch: |-
    - op: remove
      path: /spec/template/spec/nodeSelector
```

```bash
# hack/setup/scripts/generate_chart_manifests.sh (line 61)
kustomize build ${REPO_ROOT}/config/overlays/charts-localmodel > ${REPO_ROOT}/charts/kserve-localmodel-resources/files/resources.yaml
```

Verified end to end with the script: the regenerated `files/resources.yaml` differs from the current one by exactly the two removed selector lines, the default render is unchanged, the `null` override clears the selector, and re-running the script is stable.

The same generation issue also affects #5465's direct edit to the generated `values.yaml`.

Bafff:
ping

aslafy-z:
Thanks a lot for the thorough review @Bafff, and sorry for the delay! You were right: my original edit targeted a generated file and would have been reverted by the next `generate_chart_manifests.sh` run.

I've updated the PR to use the chart-generation overlay you suggested (`config/overlays/charts-localmodel`), rebased on current master, and refreshed the PR description accordingly. Good point on `null` vs `{}` too, that's now noted in the description.

Verified end to end: regeneration is stable and reproduces the committed `files/resources.yaml` exactly, the default render is unchanged, and `--set kserve.localmodelnode.controller.nodeSelector=null` clears the selector.

Bafff:
@aslafy-z now we need to merge this :)

spolti:
/approve

Thanks!

PR #5681, opened 2026-06-16
feat: auto inject OTel sidecar annotation for PodMetric autoscaling

### LINKED ISSUE #4937: Auto-inject the annotation `sidecar.opentelemetry.io/inject` into the deployment
/kind feature

KServe could automatically inject this annotation into the Deployment when it detects:
  - PodMetricSourceType in the autoscaling spec
  - Backend == "opentelemetry"

  This would make sense because:
  1. KServe already detects this condition (in raw_kube_reconciler.go:81-96) to create the OpenTelemetryCollector CR
  2. The annotation value follows a predictable pattern: <inferenceservice-name>-predictor
  3. It would eliminate manual configuration and potential errors

  Current Workaround

  Currently, users must manually add both:
```
  metadata:
    annotations:
      sidecar.opentelemetry.io/inject: "my-model-predictor"  # Manual!
  spec:
    predictor:
      autoScaling:
        metrics:
          - type: PodMetric
            podMetric:
              metric:
                backend: opentelemetry  # This triggers CR creation
```

  This seems like a potential enhancement opportunity - KServe could auto-inject the annotation when it creates the OpenTelemetryCollector CR.

pm32900:
**Summary**
Automatically injects the sidecar.opentelemetry.io/inject annotation onto the predictor Deployment pod template when the predictor's autoscaling configuration uses type: PodMetric with backend: opentelemetry. Fixes #4937.

**Problem**
KServe already creates a OpenTelemetryCollector CR (in sidecar mode) named after the predictor component (e.g., my-model-predictor) when the ISVC autoscaling configuration contains a PodMetric with backend: opentelemetry. However, the OTel operator requires a matching annotation on the pod template to inject the sidecar:
`sidecar.opentelemetry.io/inject: "my-model-predictor"`
This annotation was never automatically set, requiring users to manually add it to their ISVC. If omitted, the sidecar is never injected, the metrics pipeline is broken, and the autoscaler cannot function.

**Solution**
Added a new addOtelSidecarAnnotation helper function in component.go, following the existing pattern of addLoggerAnnotations, addBatcherAnnotations, and addAgentAnnotations.

The function:

- Checks if AutoScaling contains any PodMetric entry with backend: opentelemetry
- Injects sidecar.opentelemetry.io/inject: <predictor-name> onto the pod template annotations
- Does not overwrite a user-provided value for that key
- Is a no-op for all other metric types and backends

The function is called from Predictor.Reconcile() immediately after the other annotation helpers and before buildObjectMeta, so the annotation flows naturally into the Deployment pod template via utils.Union.

The annotation key OTelSidecarInjectAnnotationKey is added to constants.go alongside the existing Istio sidecar constants.

**Behavior Changes**

- Before: Users must manually add sidecar.opentelemetry.io/inject: <name> to their ISVC annotation or predictor annotation when using OTel-backed autoscaling.
- After: The annotation is automatically set to the predictor name (e.g., my-model-predictor) when type: PodMetric + backend: opentelemetry is detected.
- No change for any other autoscaling configuration, backend, or deployment mode.
- No change if the user has already set sidecar.opentelemetry.io/inject manually — their value is preserved.

**Testing Done**
- Added TestAddOtelSidecarAnnotation in pkg/controller/v1beta1/inferenceservice/components/component_test.go covering:

1. Injects annotation for PodMetric + opentelemetry
2. No injection for Resource metric type
3. No injection for PodMetric with non-opentelemetry backend
4. No injection when AutoScaling is nil
5. No injection when Metrics slice is empty
6. Preserves user-provided annotation value
7. Injects when mixed metrics include one PodMetric+opentelemetry
8. No injection for External metric type

- Ran go test ./pkg/controller/v1beta1/inferenceservice/... — all pass
- Ran golangci-lint run ./pkg/... — no new issues
- Verified no diff from make generate / make manifests

**Notes for Maintainers**
- No API changes. No new CRD fields, no zz_generated files touched.
- Scope is Standard (RawKube) deployment mode only. The OTel sidecar path does not exist in Knative/serverless mode; the annotation is harmless there but will not trigger injection (no OTel CR is created in that path).
- The only structural change to predictor.go is that predictorName is computed 5 lines earlier than before (hoisted above reconcileModelConfig). This is a pure variable hoist with no semantic change.
- The constant OTelSidecarInjectAnnotationKey = "sidecar.opentelemetry.io/inject" is placed next to the Istio sidecar annotation constants in constants.go for discoverability.


pm32900:
Hi, I opened a PR for this issue.

The implementation keeps the change scoped to the OpenTelemetry autoscaling path and only injects `sidecar.opentelemetry.io/inject` when the predictor autoscaling metric uses `PodMetric` with backend `opentelemetry`.

It also preserves an existing user-provided annotation value, so explicit user configuration is not overwritten.

I’d appreciate feedback on whether the injection point and annotation value align with the expected KServe/OpenTelemetry collector naming flow.

pm32900:
Hi! Just checking in to see if there's anything I can update or clarify to help move this PR forward. Happy to make any requested changes. Thanks!

PR #5685, opened 2026-06-17
feat(llmisvc): support append merge strategy via annotation

pierDipi:
Add a merge-append-fields annotation on LLMInferenceServiceConfig (and LLMInferenceService) that declares specific array fields should be concatenated (base + override) instead of atomically replaced during strategic merge. This is useful for container args, command, and other +listType=atomic fields where an override config should extend the base rather than replace it entirely.

The annotation uses Kustomize-style field paths with kyaml Lookup for traversal:
```
  serving.kserve.io/merge-append-fields: |
    template.containers.[name=kserve-container].args
    router.scheduler.template.containers.[name=main].command
```

Without the annotation, merge behavior is unchanged (backward compatible).

**Release note**:

```release-note
feat(llmisvc): support append merge strategy via annotation. 

Add a merge-append-fields annotation on LLMInferenceServiceConfig (and LLMInferenceService) that declares specific array fields should be concatenated (base + override) instead of atomically replaced during strategic merge. This is useful for container args, command, and other +listType=atomic fields where an override config should extend the base rather than replace it entirely.

The annotation uses Kustomize-style field paths with kyaml Lookup for traversal:

  serving.kserve.io/merge-append-fields: |
    template.containers.[name=kserve-container].args
    router.scheduler.template.containers.[name=main].command

Without the annotation, merge behavior is unchanged (backward compatible).
```


bartoszmajsak:
@pierDipi precommit fails after merge, as we removed Istio Shadow Service. I think it's safe to just remove this part of the fixture from the affected tests.

```
vet: pkg/controller/v1alpha2/llmisvc/config_merge_append_int_test.go:38:45: undefined: WithIstioShadowService
```

pkg/controller/v1alpha2/llmisvc/config_merge_append.go
```diff
@@ -0,0 +1,194 @@
...
+	logger := log.FromContext(ctx)
+	out := specs[0].spec
+
+	for i := 1; i < len(specs); i++ {
+		override := specs[i]
+		appendPaths, err := v1alpha2.ParseMergeAppendFieldPaths(override.annotations)
+		if err != nil {
+			logger.V(1).Info("ignoring invalid merge-append-fields paths",
+				"error", err,
+				"configIndex", i,
```
bartoszmajsak, line 64: Is this helpful enough? Shouldn't we emit more visible signal like error that is surfaced through `PresetsCombined` condition or at least an event?

pkg/apis/serving/v1alpha2/llm_inference_service_types_func.go
```diff
@@ -195,3 +199,108 @@ func (s *LLMInferenceService) ManagedDRAContainerName() (string, bool) {
...
+// dots or double dots) are silently dropped.
+//
+//	"template.containers.[name=main].args"
+//	  -> ["template", "containers", "[name=main]", "args"]
+func ParseFieldPath(path string) ([]string, error) {
+	if path == "" {
+		return nil, errors.New("path must not be empty")
+	}
+	var segments []string
+	for _, s := range strings.Split(path, ".") {
```
bartoszmajsak: I am lazy, so can't resist to suggests using https://github.com/kubernetes-sigs/kustomize/blob/kyaml/v0.21.1/kyaml/utils/pathsplitter.go#L44

pkg/controller/v1alpha2/llmisvc/config_merge_append_int_test.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	. "github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc/fixture"
+)
+
+var _ = Describe("Merge Append Lists", func() {
+	Context("when an LLMInferenceServiceConfig carries the merge-append-fields annotation", func() {
+		It("should append container args from the override config to the base config args", func(ctx SpecContext) {
+			// given
+			svcName := "append-args"
+			testNs := NewTestNamespace(ctx, envTest, WithIstioShadowService(svcName))
```
pierDipi:
```suggestion
			testNs := NewTestNamespace(ctx, envTest)
```

pkg/controller/v1alpha2/llmisvc/config_merge_append_int_test.go
```diff
@@ -0,0 +1,337 @@
...
+					"--max-model-len=2048",
+					"--enable-lora",
+				))
+			}).WithContext(ctx).Should(Succeed())
+		})
+
+		It("should replace args when the override config has no merge-append annotation", func(ctx SpecContext) {
+			// given
+			svcName := "replace-args"
+			testNs := NewTestNamespace(ctx, envTest, WithIstioShadowService(svcName))
```
pierDipi:
```suggestion
			testNs := NewTestNamespace(ctx, envTest)
```

pkg/controller/v1alpha2/llmisvc/config_merge_append_int_test.go
```diff
@@ -0,0 +1,337 @@
...
+				g.Expect(kserveContainer).NotTo(BeNil())
+				g.Expect(kserveContainer.Args).To(ContainElement("--new-flag"))
+				g.Expect(kserveContainer.Args).NotTo(ContainElement("--old-flag"))
+			}).WithContext(ctx).Should(Succeed())
+		})
+
+		It("should handle a 3-config chain with mixed append and replace", func(ctx SpecContext) {
+			// given
+			svcName := "chain-mixed"
+			testNs := NewTestNamespace(ctx, envTest, WithIstioShadowService(svcName))
```
pierDipi:
```suggestion
			testNs := NewTestNamespace(ctx, envTest)
```

pkg/controller/v1alpha2/llmisvc/config_merge_append_int_test.go
```diff
@@ -0,0 +1,337 @@
...
+					"--appended-flag",
+					"--svc-flag",
+				))
+			}).WithContext(ctx).Should(Succeed())
+		})
+
+		It("should preserve other fields when appending args", func(ctx SpecContext) {
+			// given
+			svcName := "append-preserve"
+			testNs := NewTestNamespace(ctx, envTest, WithIstioShadowService(svcName))
```
pierDipi:
```suggestion
			testNs := NewTestNamespace(ctx, envTest)
```

pkg/controller/v1alpha2/llmisvc/config_merge_append.go
pierDipi: It's validated here at the webhook admission level https://github.com/kserve/kserve/pull/5685/changes#diff-b7e26a082143d7c2162f9b70a47a1f5dcf0fa799f40b3ea7e4a717bb3a6aa1dd

pierDipi: I guess this should never really happen if webhook is working and configured

pkg/apis/serving/v1alpha2/llm_inference_service_types_func.go
pierDipi: it seems like unit tests are equally happy

bartoszmajsak:
@pierDipi precommit among other things not happy :(

pierDipi:
Resolved

pkg/controller/v1alpha2/llmisvc/config_merge_append.go
```diff
@@ -0,0 +1,194 @@
+/*
+Copyright 2025 The KServe Authors.
```
bartoszmajsak, line 2: nit: year in all net-new files ;)

pkg/apis/serving/v1alpha2/llm_inference_service_types_func.go
```diff
@@ -195,3 +200,99 @@ func (s *LLMInferenceService) ManagedDRAContainerName() (string, bool) {
 	}
 	return kservevalidation.ManagedDRAContainerName(s.Annotations)
 }
+
+// MergeAppendFieldsAnnotation is the annotation key on LLMInferenceServiceConfig
+// (or LLMInferenceService) that lists Kustomize-style field paths whose arrays
```
bartoszmajsak, line 205: Should we take the union of append paths across all specs, rather than reading them only from the override side? `mergeAnnotatedSpecs` evaluates exactly the one merge step. 

With a preset, one baseRef and the service all setting `template.containers.[name=main].args` ([example](https://gist.github.com/bartoszmajsak/dfd939e7438d5c6d58e9f8516a444818)):

```
annotation on the config only, service also sets args  -> [--from-my-service]
annotation on the config only, service sets no args    -> [--preset-flag --from-my-config]
annotation on both the config and the service          -> [--preset-flag --from-my-config --from-my-service]
annotation on the service only                         -> [--from-my-config --from-my-service]
```

So when you annotate your config, it works, then you add an arg on the `LLMInferenceService` and the append stops - the service is the last override and carries no annotation, so it is atomic replace. 

Is this what we want?

PR #5687, opened 2026-06-17
feat(llmisvc): add speculative decoding support for LLMInferenceService

vivekk16:
## Summary

Adds first-class speculative decoding support to the LLMInferenceService CRD via a new `spec.speculator` field. The controller automatically handles model download, storage initialization, and vLLM `--speculative-config` argument injection for decode workloads.

### Supported methods
- **Eagle3** — purpose-trained speculator heads (requires separate model)
- **Draft-target** — uses a smaller model from the same family (requires separate model)
- **N-gram** — prompt-lookup based token proposal (no model needed)
- **MTP** — multi-token prediction using built-in model heads (no model needed)
- **Medusa** — multi-head decoder with tree-based attention (requires separate model)

### Changes
- **API**: Add `SpeculatorSpec` to v1alpha2 types with a dedicated `LLMSpeculatorModelSpec` and `config` (map passed to vLLM `--speculative-config`). CEL validation enforces that `config.method` is required when config is present. LoRA is structurally excluded from the speculator model type rather than relying on runtime validation.
- **v1alpha1 conversion**: Round-trip support between v1alpha1 and v1alpha2 for the speculator field.
- **Controller**: New `workload_speculator.go` handles speculator model download via a dedicated `speculator-initializer` init container (reusing the existing storage-initializer image and credential injection) and injects `--speculative-config` JSON into the main container args. Config values are coerced to native JSON types (int, float, bool) for correct vLLM parsing. Applied only to decode workloads (single-node, multi-node leader/worker); prefill pods are excluded.
- **CRDs**: Regenerated CRD manifests for all chart variants (minimal, full) and test CRDs.
- **Tests**: Unit and integration tests covering all speculative decoding methods, model download (HF, S3, OCI, PVC), credential injection, storage-initializer-disabled mode, speculator removal cleanup, and error cases.

### Release Note

```release-note
Add speculative decoding support for LLMInferenceService via a new `spec.speculator` field. Supports Eagle3, draft-target, N-gram, MTP, and Medusa methods. The controller automatically downloads speculator models (HF, S3, OCI, PVC) via a dedicated init container and injects vLLM `--speculative-config` arguments for decode workloads.

### CODE REVIEW

File: pkg/apis/serving/v1alpha1/llm_inference_service_types.go [hunk 1]
```diff
@@ -110,13 +110,31 @@ type LLMInferenceServiceSpec struct {
...
 	BaseRefs []corev1.LocalObjectReference `json:"baseRefs,omitempty"`
 }
 
+// SpeculatorSpec configures speculative decoding for the inference server.
+// +kubebuilder:validation:XValidation:rule="!has(self.model) || !has(self.model.lora)",message="speculator.model.lora is not supported; LoRA adapters apply only to the base model"
+// +kubebuilder:validation:XValidation:rule="!has(self.config) || (size(self.config) == 0) || ('method' in self.config)",message="speculator.config.method is required; it specifies the speculative decoding strategy (e.g. eagle3, draft_model, ngram, mtp)"
+type SpeculatorSpec struct {
+	// Model specification for the speculator or draft model.
+	// +optional
+	Model *LLMModelSpec `json:"model,omitempty"`
```
pierDipi: can we make a new type LLMSpeculatorModelSpec so we avoid having lora as a field of speculator?

vivekk16: Done

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -100,13 +100,87 @@ type LLMInferenceServiceSpec struct {
 	// +optional
 	Tracing *TracingSpec `json:"tracing,omitempty"`
 
+	// Speculator configures speculative decoding for the model server.
+	// When specified with a model URI, the controller creates a second storage-initializer
+	// init container to download the speculator (or draft) model and mounts it into the
+	// inference container. The config map is passed directly to vLLM's --speculative-config.
+	// +optional
+	Speculator *SpeculatorSpec `json:"speculator,omitempty"`
+
```
henrichter-sap, line 127: Would it be worth placing this under `spec.model.speculator` instead of at the top level? 

Speculative decoding decorates how the model is served. The same conceptual category as LoRA, which lives at `spec.model.lora`. Putting them side-by-side reads as "attributes of the served model", while the current top-level placement suggests they're different kinds of things.

`spec.prefill`/`spec.worker`/`spec.router` do live at the top level, but those are workload/topology fields and not model attributes.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -100,13 +100,87 @@ type LLMInferenceServiceSpec struct {
...
+	// Not required for methods that don't use a separate model (e.g., ngram, mtp).
+	// +optional
+	Model *LLMSpeculatorModelSpec `json:"model,omitempty"`
+
+	// Config provides the speculative decoding parameters passed directly to the
+	// vLLM --speculative-config JSON. Keys correspond to vLLM's speculative config schema.
+	// Common keys: method, num_speculative_tokens, max_model_len, draft_tensor_parallel_size.
+	// See https://docs.vllm.ai/en/latest/features/speculative_decoding/#-speculative-config-schema
+	// +optional
+	Config map[string]string `json:"config,omitempty"`
```
henrichter-sap, line 191: I'm wondering if strongly typed fields (like already used by lora) would be better here as well?

henrichter-sap (commented):
Great work. I was looking for something like that. Some feedback/nits from my side:

The `storage-initializer` entrypoint already accepts multiple (uri, dest) pairs, and LoRA reuses a single init container via `attachMultiStorageDownloads`. Could we pass the speculator model as an additional pair there instead of spawning a second init container?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: Thanks for the suggestion! I considered this but decided to keep it top-level for a few reasons:

Speculative decoding is only applied to decode workloads, prefill pods are explicitly excluded. This workload-scoping behavior puts it at the same conceptual level as `spec.prefill` and `spec.worker`, not alongside `spec.model.lora` which applies uniformly.

Methods like `ngram` and `mtp` don't use a separate model at all, they are pure inference-time configurations. Nesting them under `spec.model` feels misleading when no model is involved.

Moving it under spec.model would give us `spec.model.speculator.model.uri`, a model inside a model, which is confusing to read and author.

The spec.speculator structure with model and config sub-fields was proposed by @pierDipi in the predecessor PR (#5370), with the top-level placement being intentional.

Happy to discuss further if you feel strongly about it.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: This was also discussed in the predecessor PR (#5370) where @pierDipi recommended the passthrough map approach so users can reference vLLM docs directly, and the API doesn't need CRD updates as vLLM adds new methods/parameters. Different methods also use different keys (`ngram` needs `prompt_lookup_max`, `eagle3` doesn't), so a typed struct would end up with many method-specific optional fields.

vivekk16:
@henrichter-sap The speculator model download now goes through the shared storage-initializer init container via attachMultiStorageDownloads, same pattern LoRA uses. No separate init container is created. Addressed in the latest commit.

docs/samples/llmisvc/speculative-decoding/llm-inference-service-A100-draft-target.yaml
```diff
@@ -0,0 +1,34 @@
...
+  router:
+    scheduler: {}
+    route: {}
+    gateway: {}
+  template:
+    containers:
+      - name: main
+        env:
+          - name: VLLM_USE_V1
+            value: "1"
```
pierDipi, line 29: do we need this?

docs/samples/llmisvc/speculative-decoding/llm-inference-service-A100-draft-target.yaml
```diff
@@ -0,0 +1,34 @@
...
+        resources:
+          limits:
+            nvidia.com/gpu: "1"
+          requests:
+            nvidia.com/gpu: "1"
```
pierDipi, line 34: don't we need cpu and memory too?

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -127,14 +150,23 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
...
 	switch schema + "://" {
 	case constants.PvcURIPrefix:
 		if err := r.attachPVCModelArtifact(modelUri, podSpec, containerName, modelPath); err != nil {
 			return err
 		}
-		if len(loraPairs) > 0 {
-			if err := r.attachMultiStorageDownloads(ctx, serviceAccount, llmSvc, curr, podSpec, config.StorageConfig, config.CredentialConfig, containerName, loraPairs); err != nil {
+		downloadPairs := loraPairs
+		if specPair != nil {
+			downloadPairs = append(downloadPairs, *specPair)
```
KillianGolds, line 167: I may be misreading the mount plumbing, so correct me if so. This looks like a consequence of the
move to the shared init container, which seems right in general. Here's my thinking:

`attachMultiStorageDownloads` mounts one emptyDir at `FindCommonParentPath(paths)`. Once the
speculator pair joins the LoRA pairs, those paths are `/mnt/lora/<name>` and
`/mnt/speculator/model`, so the parent collapses to `/mnt`. I pulled the branch down and drove
`attachModelArtifacts` directly to check, with a `pvc://` base model, one hf LoRA adapter and an
hf speculator:

```
main container volumeMounts, in list order:
  [0] /mnt/models    kserve-pvc-source           readOnly=true
  [1] /mnt           kserve-provision-location   readOnly=true
```

The same case without LoRA gives `/mnt/speculator/model` for [1], so it is specifically the
LoRA-plus-speculator combination that widens the parent, and the emptyDir does end up as a parent
mount sitting after its child.

What I could not work out is whether that actually shadows at runtime. If mounts are applied in
list order the main container would see an empty `/mnt/models` and lose the base model, but you
would know better than me whether the kubelet or the runtime reorders nested mounts first. If it
does reorder, this is a non-issue and please ignore me.

If it does not, a test asserting the main container's mount paths would pin it down, since the
current ones only check init-container args.

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -146,8 +178,12 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
 		if err := r.attachOciNativeModelArtifact(normalizedURI, podSpec, config.StorageConfig, containerName, modelPath); err != nil {
 			return err
 		}
-		if len(loraPairs) > 0 {
-			if err := r.attachMultiStorageDownloads(ctx, serviceAccount, llmSvc, curr, podSpec, config.StorageConfig, config.CredentialConfig, containerName, loraPairs); err != nil {
+		downloadPairs := loraPairs
+		if specPair != nil {
+			downloadPairs = append(downloadPairs, *specPair)
```
KillianGolds, line 184: Same shape concern as https://github.com/kserve/kserve/pull/5687/changes#r3674168765

pkg/controller/v1alpha2/llmisvc/workload_storage.go
```diff
@@ -167,33 +203,43 @@ func (r *LLMISVCReconciler) attachModelArtifacts(ctx context.Context, serviceAcc
 				return err
 			}
 		}
-		if len(loraPairs) > 0 {
-			if err := r.attachMultiStorageDownloads(ctx, serviceAccount, llmSvc, curr, podSpec, config.StorageConfig, config.CredentialConfig, containerName, loraPairs); err != nil {
+		downloadPairs := loraPairs
+		if specPair != nil {
+			downloadPairs = append(downloadPairs, *specPair)
```
KillianGolds, line 209: Same shape concern as https://github.com/kserve/kserve/pull/5687/changes#r3674168765

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -118,13 +118,87 @@ type LLMInferenceServiceSpec struct {
...
+//
+// Example - N-gram (no model needed):
+//
+//	speculator:
+//	  config:
+//	    method: ngram
+//	    num_speculative_tokens: "4"
+//	    prompt_lookup_max: "5"
+//
+// +kubebuilder:validation:XValidation:rule="!has(self.config) || (size(self.config) == 0) || ('method' in self.config)",message="speculator.config.method is required; it specifies the speculative decoding strategy (e.g. eagle3, draft_model, ngram, mtp)"
```
KillianGolds, line 176: Since this rule is guarded on `has(self.config)`, it looks like `speculator: {model: {uri: ...}}`
with no `config` would be accepted, and then produce
`--speculative-config '{"model":"/mnt/speculator/model"}'` with no `method`.

Is that valid for vLLM, or is there a default I'm not aware of? If it isn't, requiring `method`
whenever `model` is set would move the failure to apply time:

```
// +kubebuilder:validation:XValidation:rule="!has(self.model) || (has(self.config) && 'method' in self.config)",message="speculator.config.method is required when speculator.model is set"
```

pkg/controller/v1alpha2/llmisvc/workload_speculator.go
```diff
@@ -0,0 +1,227 @@
...
+		return i
+	}
+	if f, err := strconv.ParseFloat(v, 64); err == nil {
+		return f
+	}
+	if b, err := strconv.ParseBool(v); err == nil {
+		return b
+	}
+	return v
+}
```
KillianGolds, line 223: Not reopening the typed-vs-map discussion from #5370, just a narrower thing I noticed here.

`strconv.ParseFloat` accepts `inf` and `NaN`, and `json.Marshal` errors on non-finite floats, so
a value like `num_speculative_tokens: "nan"` looks like it would fail the marshal and keep failing
on every reconcile, rather than surfacing as a validation error. Unlikely input, but the failure
mode seemed worth flagging. Would a guard before the marshal be worth adding?

pkg/controller/v1alpha2/llmisvc/workload_speculator.go
```diff
@@ -0,0 +1,227 @@
...
+	case constants.HfURIPrefix, constants.S3URIPrefix:
+		storageInitializerDisabled := llmSvc.Spec.StorageInitializer != nil &&
+			llmSvc.Spec.StorageInitializer.Enabled != nil &&
+			!*llmSvc.Spec.StorageInitializer.Enabled
+		if storageInitializerDisabled {
+			return fmt.Errorf("speculator model %q: hf:// and s3:// require the storage initializer — set storageInitializer.enabled to true", speculatorUri)
+		}
+
+	default:
+		return fmt.Errorf("unsupported scheme %q in speculator model URI %q; supported: pvc://, oci://, hf://, s3://", schema+"://", speculatorUri)
```
KillianGolds, line 83: Small question on where this lives: a typo'd scheme surfaces here as a reconcile error rather
than a rejected create, so the user sees it on the LLMInferenceService status instead of from
`kubectl apply`. `validateLoRAAdapters` in `llm_inference_service_validation.go` does its scheme
checks in the webhook, was there a reason to prefer reconcile-time here, or just that this path
already had the URI parsed? Not blocking either way, the error message itself is clear.

KillianGolds (commented):
Thanks for this, and for the storage-initializer consolidation in the latest commit. I pulled it down locally: it builds clean, the speculator unit tests pass, and the Speculative Decoding envtest specs pass too.

The shape of the feature reads well to me, and test coverage looks solid. Three questions inline,
roughly in order of how much they worry me:

1. Whether folding the speculator pair into the shared init container widens the mount root to `/mnt` for `pvc://` base models when LoRA is also configured. I confirmed the pod spec ends up with a parent mount sitting after its child, what I could not work out is whether that shadows at runtime, so this may well be a non-issue. Details inline.
2. Whether the CEL rule should also require `config.method` when `speculator.model` is set.
3. A narrow edge case in `inferJSONType` where non-finite floats would fail `json.Marshal`.

Plus a non-blocking aside about whether the URI scheme check belongs in the validating webhook.

I am not deeply familiar with this part of the codebase, so please push back where I have got it wrong. The first is the only one I would want settled before merge, and a test asserting the main container's mount paths would answer it either way.

PR #5690, opened 2026-06-17
feat(llmisvc): localModelCache support for LoRA adapters

mholder6:

LLMInferenceService already supports LocalModelCache for the base model URI. This PR extends that to LoRA adapter URIs so cached adapters are mounted from PVCs instead of being downloaded at pod startup. Adapter caching is independent of whether the base model is cached.

Changes:
- **Defaulter**: match LoRA adapter URIs against LMC/LMNC and set a `localmodel-lora` annotation with cache metadata
- **Workload reconciler**: rewrite cached adapters to PVC mounts; skip adapter URIs in storage-initializer
- **LMC/LMNC reconcilers**: index and track adapter-only LLMISVC references in cache status
- **Delete webhooks**: block LMC/LMNC deletion while an LLMISVC references a cached adapter
- **`pkg/localmodelcache`**: shared helpers for cache matching, annotation parsing, and reference checks
- **E2E**: `test_llmisvc_localmodelcache_lora_adapters` in the modelcache suite (annotations, PVC mounts, status tracking, delete protection)
- **Samples**: `docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml` and README section 4

Fixes #

- [x] Unit tests for defaulter, workload rewrite, reconciler status, and delete webhooks
- [x] E2E test `test_llmisvc_localmodelcache_lora_adapters` (runs in `e2e-test-modelcache` CI)
- [x] Manual Minikube validation (annotation, PVC mount, storage-initializer args, status tracking, delete protection)

**Reviewer notes**:

**Reviewer testing guide:** https://gist.github.com/mholder6/84574a2e662290c85d70fc850876cdf9

CI runs the new E2E via the modelcache workflow. Reviewers can also run locally:

```bash
PYTEST_ARGS="-k test_llmisvc_localmodelcache_lora" ./test/scripts/gh-actions/run-e2e-tests.sh modelcache 1
```

**Checklist**:

- [x] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

**Release note**:

```release-note
LLMInferenceService LoRA adapters can now be served from LocalModelCache and LocalModelNamespaceCache when the adapter URI matches a cache entry.
```

### DIFF
--- a/docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
+++ b/docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
@@ -1,0 +1,70 @@
+# Example: LLMInferenceService with base model and LoRA adapter served from LocalModelCache.
+#
+# Prerequisites:
+#   - LocalModelNodeGroup targeting worker nodes
+#   - LocalModelCache CRs for the base model URI and each adapter URI
+#
+# When adapter URIs match a LocalModelCache, the controller mounts cached PVCs
+# instead of downloading artifacts at pod startup.
+apiVersion: serving.kserve.io/v1alpha2
+kind: LLMInferenceService
+metadata:
+  name: llm-lora-localmodelcache
+spec:
+  model:
+    uri: hf://my-org/my-base-model
+    lora:
+      adapters:
+        - name: sql-adapter
+          uri: hf://my-org/my-sql-lora-adapter
+---
+apiVersion: serving.kserve.io/v1alpha1
+kind: LocalModelNodeGroup
+metadata:
+  name: llm-lora-nodegroup
+spec:
+  storageLimit: 50Gi
+  persistentVolumeSpec:
+    accessModes:
+      - ReadWriteOnce
+    storageClassName: standard
+    capacity:
+      storage: 50Gi
+    local:
+      path: /models
+    persistentVolumeReclaimPolicy: Delete
+    nodeAffinity:
+      required:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: node.kubernetes.io/instance-type
+                operator: In
+                values:
+                  - gpu-node
+  persistentVolumeClaimSpec:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 50Gi
+    storageClassName: standard
+---
+apiVersion: serving.kserve.io/v1alpha1
+kind: LocalModelCache
+metadata:
+  name: base-model-cache
+spec:
+  sourceModelUri: hf://my-org/my-base-model
+  modelSize: 20Gi
+  nodeGroups:
+    - llm-lora-nodegroup
+---
+apiVersion: serving.kserve.io/v1alpha1
+kind: LocalModelCache
+metadata:
+  name: sql-adapter-cache
+spec:
+  sourceModelUri: hf://my-org/my-sql-lora-adapter
+  modelSize: 1Gi
+  nodeGroups:
+    - llm-lora-nodegroup
--- a/docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
+++ b/docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
@@ -26,6 +26,8 @@ spec:
   storageLimit: 50Gi
   persistentVolumeSpec:
     accessModes:
+      # Node-local storage: each node gets its own PV (local.path below).
+      # Use ReadWriteMany/ReadOnlyMany for shared storage (e.g. NFS).
       - ReadWriteOnce
     storageClassName: standard
     capacity:
@@ -43,6 +45,7 @@ spec:
                   - gpu-node
   persistentVolumeClaimSpec:
     accessModes:
+      # Matches PV access mode; node-local storage assumption as above.
       - ReadWriteOnce
     resources:
       requests:
--- a/pkg/controller/v1alpha1/localmodel/controller_test.go
+++ b/pkg/controller/v1alpha1/localmodel/controller_test.go
@@ -494,6 +494,88 @@ var _ = Describe("CachedModel controller", func() {
 					cachedModel.Status.LLMInferenceServices[0].Namespace == llmSvcNamespace
 			}, timeout, interval).Should(BeTrue(), "Cache status should track the LLMInferenceService")
 		})
+
+		It("Should track LLMInferenceService referenced only via LoRA adapter in cluster-scoped cache status", func() {
+			defer GinkgoRecover()
+			ctx, cancel := context.WithCancel(context.Background())
+			DeferCleanup(cancel)
+			nodeGroup1 := &v1alpha1.LocalModelNodeGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "gpu1",
+				},
+				Spec: localModelNodeGroupSpec1,
+			}
+			Expect(k8sClient.Create(ctx, nodeGroup1)).Should(Succeed())
+			defer k8sClient.Delete(ctx, nodeGroup1)
+
+			adapterURI := "hf://org/lora-adapter-only"
+			baseURI := "hf://org/remote-base"
+			modelName := fmt.Sprintf("lora-adapter-cache-%d", time.Now().UnixNano())
+			llmSvcName := "test-llm-lora-adapter-only"
+			llmSvcNamespace := "default"
+			cachedModel := &v1alpha1.LocalModelCache{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: modelName,
+				},
+				Spec: v1alpha1.LocalModelCacheSpec{
+					SourceModelUri: adapterURI,
+					ModelSize:      resource.MustParse("10Gi"),
+					NodeGroups:     []string{"gpu1"},
+				},
+			}
+			Expect(k8sClient.Create(ctx, cachedModel)).Should(Succeed())
+			defer func() {
+				Expect(k8sClient.Delete(ctx, cachedModel)).Should(Succeed())
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, types.NamespacedName{Name: modelName}, cachedModel)
+					return err != nil && errors.IsNotFound(err)
+				}, timeout, interval).Should(BeTrue())
+			}()
+
+			baseModelURI, err := apis.ParseURL(baseURI)
+			Expect(err).NotTo(HaveOccurred())
+			adapterModelURI, err := apis.ParseURL(adapterURI)
+			Expect(err).NotTo(HaveOccurred())
+			llmSvc := &v1alpha2.LLMInferenceService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      llmSvcName,
+					Namespace: llmSvcNamespace,
+					Annotations: map[string]string{
+						constants.LocalModelLoRAAnnotationKey: fmt.Sprintf(
+							`{"my-adapter":{"cache":%q,"sourceUri":%q,"pvcName":%q}}`,
+							modelName, adapterURI, modelName+"-gpu1",
+						),
+					},
+				},
+				Spec: v1alpha2.LLMInferenceServiceSpec{
+					Model: v1alpha2.LLMModelSpec{
+						URI: *baseModelURI,
+						LoRA: &v1alpha2.LoRASpec{
+							Adapters: []v1alpha2.LLMModelSpec{
+								{
+									Name: ptr.To("my-adapter"),
+									URI:  *adapterModelURI,
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, llmSvc)).Should(Succeed())
+			defer k8sClient.Delete(ctx, llmSvc)
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: modelName}, cachedModel)
+				if err != nil {
+					return false
+				}
+				if len(cachedModel.Status.LLMInferenceServices) != 1 {
+					return false
+				}
+				return cachedModel.Status.LLMInferenceServices[0].Name == llmSvcName &&
+					cachedModel.Status.LLMInferenceServices[0].Namespace == llmSvcNamespace
+			}, timeout, interval).Should(BeTrue(), "Cache status should track the LLMInferenceService via LoRA adapter reference")
+		})
 	})
 
 	Context("When DisableVolumeManagement is set to true", func() {
--- a/pkg/controller/v1alpha1/localmodel/controller_test.go
+++ b/pkg/controller/v1alpha1/localmodel/controller_test.go
@@ -1342,6 +1342,93 @@ var _ = Describe("LocalModelNamespaceCache controller", func() {
 			// 2. Status correctly tracks the ISVC
 			// 3. The cleanup code path exists in ReconcileForIsvcs (verified by code inspection)
 		})
+
+		It("Should track LLMInferenceService referenced only via LoRA adapter in namespace-scoped cache status", func() {
+			defer GinkgoRecover()
+			ctx, cancel := context.WithCancel(context.Background())
+			DeferCleanup(cancel)
+
+			testNamespace := fmt.Sprintf("test-ns-lora-%d", time.Now().UnixNano())
+			namespaceObj := createTestNamespace(ctx, testNamespace)
+			defer k8sClient.Delete(ctx, namespaceObj)
+
+			nodeGroup1 := &v1alpha1.LocalModelNodeGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "gpu1",
+				},
+				Spec: localModelNodeGroupSpec1,
+			}
+			Expect(k8sClient.Create(ctx, nodeGroup1)).Should(Succeed())
+			defer k8sClient.Delete(ctx, nodeGroup1)
+
+			adapterURI := "hf://org/ns-lora-adapter-only"
+			baseURI := "hf://org/remote-base"
+			modelName := fmt.Sprintf("ns-lora-adapter-cache-%d", time.Now().UnixNano())
+			llmSvcName := "test-llm-ns-lora-adapter-only"
+			cachedModel := &v1alpha1.LocalModelNamespaceCache{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      modelName,
+					Namespace: testNamespace,
+				},
+				Spec: v1alpha1.LocalModelNamespaceCacheSpec{
+					SourceModelUri: adapterURI,
+					ModelSize:      resource.MustParse("10Gi"),
+					NodeGroups:     []string{"gpu1"},
+				},
+			}
+			Expect(k8sClient.Create(ctx, cachedModel)).Should(Succeed())
+			defer func() {
+				Expect(k8sClient.Delete(ctx, cachedModel)).Should(Succeed())
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, types.NamespacedName{Name: modelName, Namespace: testNamespace}, cachedModel)
+					return err != nil && errors.IsNotFound(err)
+				}, timeout, interval).Should(BeTrue())
+			}()
+
+			baseModelURI, err := apis.ParseURL(baseURI)
+			Expect(err).NotTo(HaveOccurred())
+			adapterModelURI, err := apis.ParseURL(adapterURI)
+			Expect(err).NotTo(HaveOccurred())
+			llmSvc := &v1alpha2.LLMInferenceService{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      llmSvcName,
+					Namespace: testNamespace,
+					Annotations: map[string]string{
+						constants.LocalModelLoRAAnnotationKey: fmt.Sprintf(
+							`{"my-adapter":{"cache":%q,"namespace":%q,"sourceUri":%q,"pvcName":%q}}`,
+							modelName, testNamespace, adapterURI, modelName+"-gpu1",
+						),
+					},
+				},
+				Spec: v1alpha2.LLMInferenceServiceSpec{
+					Model: v1alpha2.LLMModelSpec{
+						URI: *baseModelURI,
+						LoRA: &v1alpha2.LoRASpec{
+							Adapters: []v1alpha2.LLMModelSpec{
+								{
+									Name: ptr.To("my-adapter"),
+									URI:  *adapterModelURI,
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, llmSvc)).Should(Succeed())
+			defer k8sClient.Delete(ctx, llmSvc)
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: modelName, Namespace: testNamespace}, cachedModel)
+				if err != nil {
+					return false
+				}
+				if len(cachedModel.Status.LLMInferenceServices) != 1 {
+					return false
+				}
+				return cachedModel.Status.LLMInferenceServices[0].Name == llmSvcName &&
+					cachedModel.Status.LLMInferenceServices[0].Namespace == testNamespace
+			}, timeout, interval).Should(BeTrue(), "Namespace cache status should track the LLMInferenceService via LoRA adapter reference")
+		})
 	})
 })
 
--- a/pkg/controller/v1alpha1/localmodel/controller_test.go
+++ b/pkg/controller/v1alpha1/localmodel/controller_test.go
@@ -542,8 +542,8 @@ var _ = Describe("CachedModel controller", func() {
 					Namespace: llmSvcNamespace,
 					Annotations: map[string]string{
 						constants.LocalModelLoRAAnnotationKey: fmt.Sprintf(
-							`{"my-adapter":{"cache":%q,"sourceUri":%q,"pvcName":%q}}`,
-							modelName, adapterURI, modelName+"-gpu1",
+							`{"my-adapter":{"cache":%q}}`,
+							modelName,
 						),
 					},
 				},
@@ -1395,8 +1395,8 @@ var _ = Describe("LocalModelNamespaceCache controller", func() {
 					Namespace: testNamespace,
 					Annotations: map[string]string{
 						constants.LocalModelLoRAAnnotationKey: fmt.Sprintf(
-							`{"my-adapter":{"cache":%q,"namespace":%q,"sourceUri":%q,"pvcName":%q}}`,
-							modelName, testNamespace, adapterURI, modelName+"-gpu1",
+							`{"my-adapter":{"cache":%q,"namespace":%q}}`,
+							modelName, testNamespace,
 						),
 					},
 				},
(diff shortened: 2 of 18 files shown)


malhajfr:
@cmaddalozzo @israel-hdez , could you please review

pkg/localmodelcache/metadata.go
```diff
@@ -0,0 +1,225 @@
...
+func LLMISVCClusterCacheNames(labels, annotations map[string]string) []string {
+	var names []string
+	if labels != nil {
+		if name, ok := labels[constants.LocalModelLabel]; ok && name != "" {
+			if _, nsOk := labels[constants.LocalModelNamespaceLabel]; !nsOk {
+				names = append(names, name)
+			}
+		}
+	}
+	if entries, err := ParseLoRACacheAnnotation(annotations[constants.LocalModelLoRAAnnotationKey]); err == nil {
```
VedantMahabaleshwarkar: Silent parse failure here means delete protection and reconciler indexing skip LoRA references when annotation JSON is malformed, while `rewriteLoRAAdaptersFromLocalModelCache` fails reconcile — inconsistent behavior. Consider logging or fail-closed.

pkg/webhook/admission/llminferenceservice/defaulter.go
```diff
@@ -173,89 +173,82 @@ func applyDefaults(
...
+	if len(entries) == 0 {
+		clearLoRALocalModelMetadata(llmSvc)
 		return
 	}
-	if llmSvc.Labels == nil {
-		llmSvc.Labels = make(map[string]string)
+
+	raw, err := localmodelcache.MarshalLoRACacheAnnotation(entries)
+	if err != nil {
+		defaulterLogger.Error(err, "Failed to marshal LoRA local model cache annotation", "llmSvc", llmSvc.Name)
```
VedantMahabaleshwarkar, line 243: On marshal failure, consider calling `clearLoRALocalModelMetadata` or returning an error — a stale annotation could remain after adapter spec changes.

pkg/controller/v1alpha2/llmisvc/workload_lora_internal_test.go
```diff
@@ -17,100 +17,86 @@ limitations under the License.
...
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/localmodelcache"
 )
 
-func TestSanitizeLoRAPathSegment(t *testing.T) {
+func TestRewriteLoRAAdaptersFromLocalModelCache(t *testing.T) {
```
VedantMahabaleshwarkar, line 138: Please restore the removed tests for `sanitizeLoRAPathSegment`, `addLoRAVLLMArgs`, and `userSuppliedLoRAConfig` alongside these new cache-rewrite tests.

VedantMahabaleshwarkar (changes requested):
mostly looks good, a few minor changes

andresllh (commented):
**Note: pre-existing bug, not introduced here — but this PR is a good opportunity to fix it.**

`BuildCachedPVCURI` uses `strings.CutPrefix(storageURI, sourceURI)` without normalizing trailing slashes on `sourceURI`. The same pattern lives inline in `workload_storage.go` on `master` (line 113), so the base-model path has the same bug.

The failure case: if a `LocalModelCache.spec.sourceModelUri` is stored with a trailing slash (e.g. `hf://org/model/`), `MatchStorageURI` trims before comparing so the match succeeds, but `CutPrefix` performs a literal prefix check — it fails silently (the ignored `ok` return), `subPath` falls back to the full original URI, and the resulting path is `pvc://pvcName/models/<hash>//hf://org/model` — an unresolvable mount that prevents pod startup.

Suggested fix (one line, fixes both base model and LoRA paths):
```go
subPath, _ := strings.CutPrefix(storageURI, strings.TrimSuffix(sourceURI, "/"))
```

andresllh (commented):
**Minor: `ValidateDelete` ISVC loop doesn't handle `NotFound`, but the new LLMISVC loop below it does.**

In `localmodelcache_validator.go` (and the namespace-cache equivalent), the classic `InferenceService` loop propagates any `v.Get` error — including `NotFound`:

```go
if err := v.Get(ctx, client.ObjectKey(isvcMeta), &isvc); err != nil {
    return nil, err  // NotFound blocks deletion
}
```

The new `LLMInferenceService` loop added by this PR correctly skips stale status entries:

```go
if apierrors.IsNotFound(err) {
    continue
}
```

If a classic ISVC is force-deleted while its entry lingers in `Status.InferenceServices`, the delete webhook permanently blocks `LocalModelCache` deletion until the status is manually patched. The fix is already right there — just apply the same `IsNotFound` skip to the ISVC loop too.

pkg/localmodelcache/metadata.go
mholder6: Fixed in latest commit by making `LLMISVCReferencesClusterCache` and `LLMISVCReferencesNamespaceCache` fail-closed: if the `localmodel-lora` annotation is non-empty but unparseable, we treat it as a reference and block cache deletion until the annotation is fixed or removed. Added unit tests for the malformed-annotation case.

pkg/webhook/admission/llminferenceservice/defaulter.go
mholder6: Updated `setLoRALocalModelMetadata` to call `clearLoRALocalModelMetadata(llmSvc)` on marshal failure before returning

pkg/controller/v1alpha2/llmisvc/workload_lora_internal_test.go
mholder6: Restored `TestSanitizeLoRAPathSegment`, `TestAddLoRAVLLMArgs`, and `TestUserSuppliedLoRAConfig` alongside the new cache-rewrite tests in latest commit. Updated the `addLoRAVLLMArgs` expectations to match current behavior. (LoRA args prepended before user args, modules single-quoted).

mholder6:
> **Note: pre-existing bug, not introduced here — but this PR is a good opportunity to fix it.**
> 
> `BuildCachedPVCURI` uses `strings.CutPrefix(storageURI, sourceURI)` without normalizing trailing slashes on `sourceURI`. The same pattern lives inline in `workload_storage.go` on `master` (line 113), so the base-model path has the same bug.
> 
> The failure case: if a `LocalModelCache.spec.sourceModelUri` is stored with a trailing slash (e.g. `hf://org/model/`), `MatchStorageURI` trims before comparing so the match succeeds, but `CutPrefix` performs a literal prefix check — it fails silently (the ignored `ok` return), `subPath` falls back to the full original URI, and the resulting path is `pvc://pvcName/models/<hash>//hf://org/model` — an unresolvable mount that prevents pod startup.
> 
> Suggested fix (one line, fixes both base model and LoRA paths):
> 
> ```go
> subPath, _ := strings.CutPrefix(storageURI, strings.TrimSuffix(sourceURI, "/"))
> ```

@andresllh thanks for pointing this out -
`BuildCachedPVCURI` now normalizes `sourceURI` with `strings.TrimSuffix(sourceURI, "/")` before `CutPrefix`, which covers both the base-model path in workload_storage.go and LoRA adapter rewrites. Added `TestBuildCachedPVCURI_TrailingSlashSourceURI` in latest commit.

mholder6:
> **Minor: `ValidateDelete` ISVC loop doesn't handle `NotFound`, but the new LLMISVC loop below it does.**
> 
> In `localmodelcache_validator.go` (and the namespace-cache equivalent), the classic `InferenceService` loop propagates any `v.Get` error — including `NotFound`:
> 
> ```go
> if err := v.Get(ctx, client.ObjectKey(isvcMeta), &isvc); err != nil {
>     return nil, err  // NotFound blocks deletion
> }
> ```
> 
> The new `LLMInferenceService` loop added by this PR correctly skips stale status entries:
> 
> ```go
> if apierrors.IsNotFound(err) {
>     continue
> }
> ```
> 
> If a classic ISVC is force-deleted while its entry lingers in `Status.InferenceServices`, the delete webhook permanently blocks `LocalModelCache` deletion until the status is manually patched. The fix is already right there — just apply the same `IsNotFound` skip to the ISVC loop too.

applied the same `apierrors.IsNotFound` skip to the classic InferenceService loop in both localmodelcache_validator.go and local_model_namespace_cache_validation.go, matching what the LLMISVC loop already did. Added `TestDeleteLocalModelCacheWithStaleIsvcStatusEntry` in latest commit.

VedantMahabaleshwarkar (approved):
/lgtm

mholder6:
@cmaddalozzo @israel-hdez — original review feedback is addressed and @VedantMahabaleshwarkar has lgtm'd. Could one of you please take a look and approve if you're good with it? Thanks!

andresllh (commented):
Thorough and well-factored PR — the `pkg/localmodelcache` extraction is clean and the fail-closed annotation handling is a good defensive choice. A few minor comments inline.

pkg/controller/v1alpha2/llmisvc/workload_lora_internal_test.go
```diff
@@ -42,24 +50,17 @@ func TestSanitizeLoRAPathSegment(t *testing.T) {
 func TestAddLoRAVLLMArgs(t *testing.T) {
 	t.Parallel()
 
-	t.Run("all params set with model-routing aliases", func(t *testing.T) {
+	t.Run("all params set", func(t *testing.T) {
```
andresllh: The test subtest was renamed from `"all params set with model-routing aliases"` to `"all params set"` and now passes simplified `name=path` strings instead of JSON. The function itself is format-agnostic (just wraps each entry in single-quotes), so this still tests the quoting behaviour correctly — but the function comment on `addLoRAVLLMArgs` (below in `workload_lora.go`) still says _"Each loraModules entry is a JSON object"_, which now mismatches the test. If the JSON format is the real production contract (the caller still generates JSON via `loraModuleJSON`), update the test to use JSON fixtures; otherwise drop the JSON claim from the comment.

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
```diff
@@ -356,19 +339,20 @@ func (c *LocalModelNamespaceCacheReconciler) SetupWithManager(mgr ctrl.Manager)
 
 	llmIsvcPredicates := predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			oldModel := e.ObjectOld.GetLabels()[constants.LocalModelLabel]
-			newModel := e.ObjectNew.GetLabels()[constants.LocalModelLabel]
-			oldNsLabel := e.ObjectOld.GetLabels()[constants.LocalModelNamespaceLabel]
-			newNsLabel := e.ObjectNew.GetLabels()[constants.LocalModelNamespaceLabel]
-			return oldModel != newModel || oldNsLabel != newNsLabel
+			old := e.ObjectOld.(*v1alpha2.LLMInferenceService)
```
andresllh, line 342: Minor naming nit: `ClusterCacheNamesEqual` is used here to compare **namespace** cache name lists, but the name implies cluster-scoped equality. Since the helper is just comparing two sorted string slices, renaming it to `CacheNamesEqual` (or `CacheNameListsEqual`) would make it read correctly at both call sites without implying scope.

pkg/webhook/admission/localmodelcache/localmodelcache_validator.go
```diff
@@ -128,11 +132,7 @@ func (v *LocalModelCacheValidator) ValidateDelete(ctx context.Context, obj runti
 			localModelCacheValidatorLogger.Error(err, "Error getting LLMInferenceService", "name", llmIsvcMeta.Name)
 			return nil, err
 		}
-		modelName, ok := llmIsvc.Labels[constants.LocalModelLabel]
-		if !ok {
-			continue
-		}
-		if modelName == localModelCache.Name {
+		if localmodelcache.LLMISVCReferencesClusterCache(localModelCache.Name, llmIsvc.Labels, llmIsvc.Annotations) {
```
andresllh, line 138: Worth noting in a comment: delete protection here (and in the `InferenceService` loop above) relies on `localModelCache.Status.LLMInferenceServices` being up-to-date at the time of the delete request. If a user creates an LLMISVC with a cached LoRA adapter and immediately tries to delete the LMC before the reconciler writes the adapter reference into the status, the webhook sees an empty list and allows the deletion. This was already true for base-model references; the adapter path inherits the same gap. It's not a regression, but a comment here would help future readers understand the protection isn't race-free.

pkg/controller/v1alpha1/localmodel/controller_test.go
```diff
@@ -494,6 +494,88 @@ var _ = Describe("CachedModel controller", func() {
 					cachedModel.Status.LLMInferenceServices[0].Namespace == llmSvcNamespace
 			}, timeout, interval).Should(BeTrue(), "Cache status should track the LLMInferenceService")
 		})
+
+		It("Should track LLMInferenceService referenced only via LoRA adapter in cluster-scoped cache status", func() {
```
andresllh, line 498: Good to have this envtest-level test for cluster-scoped cache tracking via LoRA adapter. The symmetric `LocalModelNamespaceCache` + LoRA adapter path (annotation entry with a non-empty `namespace` field) is only covered by `metadata_test.go` unit tests, not by a reconciler-level integration test. Consider adding a parallel `It` case here for the namespace-scoped reconciler so both code paths get the same level of integration coverage.

docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
```diff
@@ -0,0 +1,70 @@
...
+---
+apiVersion: serving.kserve.io/v1alpha1
+kind: LocalModelNodeGroup
+metadata:
+  name: llm-lora-nodegroup
+spec:
+  storageLimit: 50Gi
+  persistentVolumeSpec:
+    accessModes:
+      - ReadWriteOnce
```
andresllh, line 31: `ReadWriteOnce` allows only a single node to mount the PV at a time. This works for node-local storage (one independent PV per node), which appears to be the intent here given the `local.path` field. A short comment noting this assumption (e.g. `# node-local storage: each node gets its own PV; use ReadWriteMany/ReadOnlyMany for shared storage`) would prevent readers from expecting this sample to work with NFS or other multi-node-writable volumes.

pkg/controller/v1alpha2/llmisvc/workload_lora_internal_test.go
mholder6: Good catch — JSON is the production contract (`loraModuleJSON` in the caller). Restored the JSON fixtures (including model-routing aliases) in `TestAddLoRAVLLMArgs` so the test matches the comment and what we actually pass to `--lora-modules`.

pkg/controller/v1alpha1/localmodel/reconcilers/localmodelnamespacecache_reconciler.go
mholder6: Agreed — renamed `ClusterCacheNamesEqual` to `CacheNamesEqual` in `pkg/localmodelcache` and updated both the LMC and LMNC reconciler call sites.

pkg/webhook/admission/localmodelcache/localmodelcache_validator.go
mholder6: Added a comment above the ISVC/LLMISVC delete-protection loops in both the LMC and LMNC validators documenting that protection depends on status being up-to-date and can race with a newly created consumer before the reconciler writes the reference.

pkg/controller/v1alpha1/localmodel/controller_test.go
mholder6: Added a parallel envtest `It` under the `LocalModelNamespaceCache controller` suite that creates an LMNC for the adapter URI and an LLMISVC whose `localmodel-lora` annotation includes a non-empty `namespace` field, then asserts the LLMISVC appears in `Status.LLMInferenceServices`.

docs/samples/llmisvc/lora-adapters/llm-inference-service-lora-localmodelcache.yaml
mholder6: Added comments on the PV/PVC `accessModes` in the sample noting the node-local storage assumption (`local.path`) and that RWX/ROX should be used for shared storage such as NFS.

VedantMahabaleshwarkar: This is fine for modelcache since the models are ultimately written to the node storage so the usual PV accessMode consideration doesn't apply here.

pkg/localmodelcache/metadata.go
```diff
@@ -0,0 +1,259 @@
...
+
+// BuildCachedPVCURI rewrites storageURI to a local model cache PVC path.
+// subPath under sourceURI is preserved (e.g. hf://org/model/subdir).
+func BuildCachedPVCURI(sourceURI, pvcName, storageURI string) string {
+	normalizedSourceURI := strings.TrimSuffix(sourceURI, "/")
+	subPath, _ := strings.CutPrefix(storageURI, normalizedSourceURI)
+	if !strings.HasPrefix(subPath, "/") {
+		subPath = "/" + subPath
+	}
+	storageKey := v1alpha1.GetStorageKey(normalizedSourceURI)
```
bartoszmajsak: Maybe we can normalize in GetStorageKey itself so all callers hash consistently?

pkg/localmodelcache/metadata_test.go
bartoszmajsak, line 1: We probably can align with other tests that are pure functions and use:

- `t.Parallel()` 
-  TestLLMISVCClusterCacheNames inlines four scenarios sequentially instead of using `t.Run` subtests - if
  assertion 2 fails, 3 and 4 don't run.

pkg/localmodelcache/metadata.go
```diff
@@ -0,0 +1,259 @@
...
+		return true
+	}
+	return false
+}
+
+func uniqueSorted(names []string) []string {
+	if len(names) == 0 {
+		return nil
+	}
+	slices.Sort(names)
```
bartoszmajsak, line 249: Maybe we should keep `names` untouched and work on the copy by calling `slices.Clone`?

pkg/localmodelcache/metadata.go
```diff
@@ -0,0 +1,259 @@
...
+// CacheMatch describes a LocalModelCache or LocalModelNamespaceCache hit for a storage URI.
+type CacheMatch struct {
+	Name      string
+	Namespace string // set for LocalModelNamespaceCache; empty for cluster-scoped LocalModelCache
+	SourceURI string
+	PVCName   string
+}
+
+// LoRACacheEntry is one adapter's local model cache metadata stored in the LoRA JSON annotation.
+type LoRACacheEntry struct {
```
bartoszmajsak: Is this specific to LoRA? I wonder if we can't just have one `CacheEntry` struct instead of `LoRa*` and `*Match`.

bartoszmajsak (commented):
I am not sure I fully grasp the annotation bridge, but I wonder if we cannot simplify it to just store:

```
 {"adapter1": {"cache": "my-cache"}, "adapter2": {"cache": "other-cache",
  "namespace": "ns"}}
```

Controller does at reconcile time:
```
cache := &v1alpha1.LocalModelCache{}
client.Get(ctx, types.NamespacedName{Name: entry.Cache}, cache)
// now you have cache.Spec.SourceModelUri and the PVC name from the nodegroup
```

This way we have a single source of truth for PVC name and source URI (the cache object itself,  not a snapshot in an annotation).

pkg/localmodelcache/metadata.go
mholder6: @bartoszmajsak Agree that normalizing once in `GetStorageKey` would be the clean systemic fix so download, injector, and PVC rewrites all hash the same. We deliberately did not do that in this PR: `GetStorageKey` is shared by the node agent / download path and existing tests treat `s3://`foo vs `s3://foo/` as different keys (on disk folder names). Changing it here would change storage identity for any cache whose `sourceModelUri` already has a trailing slash.

The trim in `BuildCachedPVCURI` is mainly for the `CutPrefix` bug @andresllh  called out (match succeeds, prefix strip fails). Happy to track normalizing inside `GetStorageKey` as a follow up so we can do it with a clear migration/compatibility story.

pkg/localmodelcache/metadata.go
mholder6: @bartoszmajsak They are intentionally separate. `CacheMatch` is the in memory result of URI to cache matching (used by the defaulter for base model and adapters). `LoRACacheEntry` is the JSON schema for the multi adapter `localmodel-lora` annotation (json tags, map keyed by adapter name).

Base model already uses the same split conceptually (match then snapshot into annotations). `LoRACacheEntryFromMatch` is the bridge. Combining them would either put serialization tags on the match type or force LoRA and base model through one annotation shape. I would rather keep the match type free of annotation concerns unless we are redesigning the base model bridge too.

mholder6:
> I am not sure I fully grasp the annotation bridge, but I wonder if we cannot simplify it to just store...
@bartoszmajsak 
The richer annotation mirrors the existing base model LocalModelCache bridge, which already snapshots `localmodel-sourceuri` and `localmodel-pvc-name` in the defaulter (not just the cache name). LoRA does the same per adapter so rewrite can build `pvc://…` without re resolving node group / PVC naming in the LLMISVC controller.

Looking up the cache at reconcile would recompute PVC name and LocalModelCache vs LocalModelNamespaceCache, which can diverge from what the defaulter chose. Indexing and delete protection only need the cache name; `sourceUri`/`pvcName` keep rewrite a pure annotation to URI transform.

Single source of truth via `Get` is attractive, but I would rather change base model and LoRA together in a follow up than give LoRA a different contract here. What do you think?

pkg/localmodelcache/metadata_test.go
mholder6: Done. Split `TestLLMISVCClusterCacheNames` (and `TestLLMISVCNamespaceCacheNames`) into `t.Run` subtests with `t.Parallel()`.

pkg/localmodelcache/metadata.go
mholder6: Done. `uniqueSorted` now clones before sort/compact so it does not mutate the caller's slice.

pkg/localmodelcache/metadata.go
bartoszmajsak: Fair point on the migration risk. But does it mean `BuildCachedPVCURI` is now producing a hash that doesn't match what the download job wrote?

pkg/localmodelcache/metadata.go
bartoszmajsak: I am not sure if  `CacheMatch` with JSON tags isn't simpler. For base models they're harmless. And we could keep it named as general purpose entity (`CacheEntry`), I don't see anything LoRA specific yet.

bartoszmajsak:
> > I am not sure I fully grasp the annotation bridge, but I wonder if we cannot simplify it to just store...
> > @bartoszmajsak
> > The richer annotation mirrors the existing base model LocalModelCache bridge, which already snapshots `localmodel-sourceuri` and `localmodel-pvc-name` in the defaulter (not just the cache name). LoRA does the same per adapter so rewrite can build `pvc://…` without re resolving node group / PVC naming in the LLMISVC controller.
> 
> Looking up the cache at reconcile would recompute PVC name and LocalModelCache vs LocalModelNamespaceCache, which can diverge from what the defaulter chose. Indexing and delete protection only need the cache name; `sourceUri`/`pvcName` keep rewrite a pure annotation to URI transform.
> 
> Single source of truth via `Get` is attractive, but I would rather change base model and LoRA together in a follow up than give LoRA a different contract here. What do you think?

I wonder what would diverge? IIUC PVC name is `cacheName + "-" + nodeGroup`, and `sourceUri` comes from `cache.Spec.SourceModelURI`.  These are the same for webhook and reconcile time. 

A slimmer annotation with just cache refs (name + namespace) seems simpler and sufficient. If we need to keep LoRA entries distinct from the base model, an "adapters" top-level key can be good option.

pkg/localmodelcache/metadata.go
mholder6: Yes, that was a real bug in the earlier trim-before-hash version. Fixed now: `BuildCachedPVCURI` trims only for `CutPrefix`, and hashes `sourceURI` as stored via `GetStorageKey`, same as the download job. A=I added a regression assert in `TestBuildCachedPVCURI_TrailingSlashSourceURI`.. Wondering if we can leave normalization inside `GetStorageKey` as a follow-up?

pkg/localmodelcache/metadata.go
mholder6: After deeper investigation I agree with you. I collapsed `CacheMatch` / `LoRACacheEntry` into a single `CacheEntry`

mholder6:
> > > I am not sure I fully grasp the annotation bridge, but I wonder if we cannot simplify it to just store...
> > > @bartoszmajsak
> > > The richer annotation mirrors the existing base model LocalModelCache bridge, which already snapshots `localmodel-sourceuri` and `localmodel-pvc-name` in the defaulter (not just the cache name). LoRA does the same per adapter so rewrite can build `pvc://…` without re resolving node group / PVC naming in the LLMISVC controller.
> > 
> > 
> > Looking up the cache at reconcile would recompute PVC name and LocalModelCache vs LocalModelNamespaceCache, which can diverge from what the defaulter chose. Indexing and delete protection only need the cache name; `sourceUri`/`pvcName` keep rewrite a pure annotation to URI transform.
> > Single source of truth via `Get` is attractive, but I would rather change base model and LoRA together in a follow up than give LoRA a different contract here. What do you think?
> 
> I wonder what would diverge? IIUC PVC name is `cacheName + "-" + nodeGroup`, and `sourceUri` comes from `cache.Spec.SourceModelURI`. These are the same for webhook and reconcile time.
> 
> A slimmer annotation with just cache refs (name + namespace) seems simpler and sufficient. If we need to keep LoRA entries distinct from the base model, an "adapters" top-level key can be good option.

After further investigation, agreed. Slimmed to `{cache, namespace?}` and resolve source URI / PVC via `Get` at reconcile. Skipped a top-level `adapters` key because `localmodel-lora` is already LoRA-only, so the whole JSON is the adapter map; base model identity stays on the separate label/annotations.

pierDipi:
@mholder6 can you rebase please? there are conflicts

pkg/webhook/admission/llminferenceservice/defaulter.go
```diff
@@ -173,89 +173,84 @@ func applyDefaults(
...
+
+	entries := make(map[string]localmodelcache.CacheEntry)
+	for _, adapter := range llmSvc.Spec.Model.LoRA.Adapters {
+		if adapter.Name == nil {
+			continue
+		}
+		adapterURI := adapter.URI.String()
+		if match := localmodelcache.MatchCacheForURI(adapterURI, nodeGroup, nodeGroupExists, models, nsModels); match != nil {
+			// Persist only cache identity; reconcile-time Get derives sourceUri/PVC.
+			entries[*adapter.Name] = localmodelcache.AnnotationRef(*match)
```
bartoszmajsak: nit: why not `entries[*adapter.Name] = *match`? This would eliminate `localmodelcache.AnnotationRef`

bartoszmajsak (commented):
Non-blocking comment, but as you need to rebase anyway maybe worth a shot!

Thanks for addressing all my previous comments ❤️

pkg/webhook/admission/llminferenceservice/defaulter.go
mholder6: Good call, done. Assigned `*match` directly and dropped `AnnotationRef`. `json:"-"` on SourceURI/PVCName already keeps those fields out of the annotation.

bartoszmajsak (approved):

PR #5691, opened 2026-06-17
fix(llmisvc): scope InferencePool readiness to gw refs

bartoszmajsak:

InferencePool.Status.Parents is managed by external gateway controllers, not by KServe. When evaluating pool readiness, only parents matching gateways currently referenced by the HTTPRoute are relevant; entries from previously referenced gateways should not affect the readiness decision.

The existing check iterated all parent entries and treated any rejected parent as a hard failure, causing `InferencePoolNotReady` after switching gateways until the external gateway controller reconciles the pool - even though the current gateway accepted the pool.

Scopes `InferencePool` readiness evaluation to only the gateways resolved from the current `HTTPRoute#parentRefs`.

Also moves all `v1alpha2.InferencePool` readiness helpers into a dedicated `router_discovery_giev1alpha2.go` file so dropping `v1alpha2` support will be a single file deletion.

- [x] Unit tests
- [x] Integration tests

**Checklist**:

```release-note
Fixes stale InferencePool parent entries from previously-referenced gateways, blocking LLMInferenceService readiness after gateway ref changes.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router_discovery.go
+++ b/pkg/controller/v1alpha2/llmisvc/router_discovery.go
@@ -40,8 +40,6 @@ import (
 	igwapi "sigs.k8s.io/gateway-api-inference-extension/api/v1"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
-	igwapiv1alpha2 "github.com/kserve/kserve/pkg/apis/gie/v1alpha2pool"
-
 	"github.com/kserve/kserve/pkg/constants"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/controller/v1alpha2/llmisvc/router_discovery_giev1alpha2.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func hasMatchingV1Alpha2PoolParent(gw types.NamespacedName, parents []igwapiv1alpha2.PoolStatus) bool {
+	for _, parent := range parents {
+		if gw == v1alpha2ParentKey(parent, gw.Namespace) {
+			return true
+		}
+	}
+	return false
+}
```
KillianGolds: Really nitty I think, possibly intentional: the default namespace for parent matching is inconsistent between the v1 and v1alpha2 paths. `filterRelevantV1Alpha2Parents` (lines 112-120) defaults an omitted parent namespace to `pool.Namespace`, but this function defaults it to `gw.Namespace`. The v1 equivalents both use `pool.Namespace`, so only this path differs, and it can report a pool ready for a gateway that never accepted it.

If the pool lives in namespace `P` and has one parent entry for a gateway named `gw1` with no namespace set, marked accepted. Now the route points at two gateways that happen to share the name `gw1`, one in `P` and one in `Q`. The `gw1` in `Q` has never touched this pool, but because the match falls back to the gateway's own namespace, it lines up against the `P` parent anyway and the pool gets reported ready for it. It can only ever over-report (never the reverse), and you need two gateways with the same name in different namespaces, so it's possibly super edge case rather than something most setups would hit.

Mirroring the v1 matcher I think would fix it:

```go
func hasMatchingV1Alpha2PoolParent(gw types.NamespacedName, parents []igwapiv1alpha2.PoolStatus, defaultNS string) bool {
	for _, parent := range parents {
		if gw == v1alpha2ParentKey(parent, defaultNS) {
			return true
		}
	}
	return false
}
```

Callers would pass the pool namespace: line 85 -> `(gw, relevant, pool.Namespace)`, router.go:716 -> `(gw, relevant, v1alpha2Pool.Namespace)`.

pkg/controller/v1alpha2/llmisvc/router_discovery.go
KillianGolds, line 722: Non-blocking cleanup: now that the controller uses the `...ForGateways` variants, `IsInferencePoolReady` (and its helper at 727-747) is only called from tests. Same for `IsInferencePoolV1Alpha2Ready` in router_discovery_giev1alpha2.go:36-69. The unused linter won't catch them since the wrappers are exported.

Side effect: the integration helper at `controller_int_test.go:2294` checks readiness via `IsInferencePoolReady`, so it's testing semantics the controller no longer uses rather than the live path.

Might be worth dropping the v1 pair and pointing that helper at the scoped check. Fine to keep the v1alpha2 pair if it's staying for the planned single-file deletion.

KillianGolds (commented):
Nice fix. build/vet/tests are green locally.

Left two non-blocking comments: a namespace-defaulting inconsistency in the v1alpha2 matcher that can over-report readiness in a narrow same-name-across-namespaces case, and a cleanup note about the now-unused whole-pool readiness functions and a test helper that drifted off the live path. Neither blocks merge from my side.

/lgtm

pkg/controller/v1alpha2/llmisvc/router_discovery_giev1alpha2.go
bartoszmajsak: Good catch! My split brain didn't fix v1alpha2 path.

pkg/controller/v1alpha2/llmisvc/router_discovery.go
bartoszmajsak: Maybe we can clean it up as a follow-up and preserve the current behaviour for now? It's already been quite a big PR on its own.

KillianGolds: Sounds good.

pierDipi (approved):

Jooho (approved):

PR #5695, opened 2026-06-19
fix(llmisvc): reduce K8s object noise in test assertion output

pierDipi:
Assert on pod names instead of full V1Pod objects so pytest's assert introspection shows useful info rather than pages of managed_fields and None values. Strip managed_fields from diagnostic dumps as well.

An example https://github.com/kserve/kserve/actions/runs/27759842580/job/82132722768?pr=5400

### DIFF
--- a/test/e2e/llmisvc/diagnostic.py
+++ b/test/e2e/llmisvc/diagnostic.py
@@ -22,6 +22,16 @@ from typing import Callable, Optional
 _log = logging.getLogger(__name__)
 
 
+def strip_managed_fields(d):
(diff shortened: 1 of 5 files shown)


bartoszmajsak (approved):

bartoszmajsak:
/rerun-failed

@pierDipi precommit ain't happy

pierDipi:
precommit is due to unrelated changes, fix in https://github.com/kserve/kserve/pull/5696

bartoszmajsak (approved):

Jooho (approved):

PR #5696, opened 2026-06-19
fix(ci): regenerate uv-lock with make uv-lock

pierDipi:
Fix regression on generated files due to multiple parallel merges.


bartoszmajsak (approved):
/lgtm

@pierDipi title needs adjustement as per gh check :'

PR #5698, opened 2026-06-19
fix(llmisvc): default EPP port to 9002 when endpointPickerRef omits port

pierDipi:
The v1 InferencePool CRD (GIE v1.5.0) has a CEL validation rule requiring port when kind is "Service" or unspecified. Configs created before GIE v1.2.0 omit the port field entirely, causing SchedulerReconcileError on upgrade.

Default port to 9002 (standard EPP gRPC port) in two places:
- v1alpha2-to-v1 InferencePool conversion
- combineBaseRefsConfig after merging all config layers

Both sites check Kind=="Service" and Kind=="" because the kubebuilder default is only applied server-side during admission, not during in-process deserialization.


pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -397,6 +397,22 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
...
+	// unspecified (defaults to 'Service')". We check both Kind=="Service" and Kind==""
+	// because the kubebuilder default is only applied server-side during admission, not
+	// during in-process deserialization.
+	if llmSvcCfg.Spec.Router != nil &&
+		llmSvcCfg.Spec.Router.Scheduler != nil &&
+		llmSvcCfg.Spec.Router.Scheduler.Pool != nil &&
+		llmSvcCfg.Spec.Router.Scheduler.Pool.Spec != nil &&
+		(llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Port == nil || llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Port.Number == 0) &&
+		(llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Kind == "Service" || llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Kind == "") {
+		llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Port = ptr.To(igwapi.Port{Number: 9002})
```
KillianGolds, line 413: nit / possible alternative, likely out of scope: both sites hardcode `9002`, which is correct for the default config. The EPP gRPC port is in principle customizable via the scheduler template's named `grpc` container port, so for a non-default port this would default the pool to the wrong port and break ext-proc routing (a quiet 500) rather than failing loudly.

One option would be to source the port from the template's `grpc` container port when present and fall back to `9002` otherwise. I looked at this downstream in opendatahub-io/kserve#1097 by extracting the gRPC port from the scheduler template, and that logic would carry over fairly cleanly since `combineBaseRefsConfig` already has the template via `Router.Scheduler.Template`. Totally fine to skip if custom EPP ports aren't a supported case, just noting it.

Also minor: `9002` is now duplicated across both sites plus tests (and already a literal in `sample.go` and the scheduler config YAML), so a shared `const` could be a nice cleanup down the line.

pkg/apis/gie/v1alpha2pool/inferencepool_conversion.go
```diff
@@ -257,6 +257,13 @@ func convertExtensionRefToV1(src *Extension) (v1.EndpointPickerRef, error) {
 	}
 	endpointPickerRef.FailureMode = failureMode
 
+	// The v1 InferencePool CRD requires port when kind is "Service" or unspecified. Old
+	// v1alpha2 pools may omit PortNumber, which would make the converted v1 object fail
+	// the CEL rule. Default to the standard EPP gRPC port (9002).
+	if (endpointPickerRef.Port == nil || endpointPickerRef.Port.Number == 0) && (endpointPickerRef.Kind == "Service" || endpointPickerRef.Kind == "") {
+		endpointPickerRef.Port = ptr.To(v1.Port{Number: 9002})
```
KillianGolds, line 264: wrt: [nit / possible alternative, likely out of scope: both sites hardcode `9002`](https://github.com/kserve/kserve/pull/5698/changes#r3443701686)

KillianGolds (commented):
Looks good, the diagnosis is spot on. The guard `Kind == "Service" || Kind == ""` mirrors the CRD's CEL rule exactly, and handling both the conversion path and `combineBaseRefsConfig` covers the distinct inputs. Tests are thorough, with the explicit-port-preservation cases as the key regression guard.

Left a couple of non-blocking nits inline (sourcing the port from the template vs. hardcoding, and a possible shared const). LGTM either way.

pkg/controller/v1alpha2/llmisvc/config_merge.go
pierDipi: I think when changing ports the endpoint picker server, I'd expect the `llmSvcCfg.Spec.Router.Scheduler.Pool.Spec.EndpointPickerRef.Port` to be set and not enter this case

bartoszmajsak (approved):
Not a huge fan of those excessive ifs but that's parallel thing :)

/lgtm

Jooho (approved):

PR #5703, opened 2026-06-20
fix: multi-node InferenceService Ready condition stuck at Unknown

mwaykole:
## What

Fixes a bug where multi-node InferenceService `PredictorReady` condition gets stuck at `Unknown` despite all pods running and deployments being Available.

## Why

Two issues in `PropagateRawStatus`:

1. **Exact string match failure**: The check `progressingCondition.Reason == "NewReplicaSetAvailable"` fails for multi-node deployments where `getDeploymentCondition` concatenates reasons with container name prefixes (e.g., `"predictor-container: NewReplicaSetAvailable, worker-container: NewReplicaSetAvailable"`).

2. **Available=True downgrade race**: When `Progressing=True` with an intermediate reason (e.g., `ReplicaSetUpdated` during rollout) while `Available=True` for all deployments, the ISVC was incorrectly set to `Unknown`. The component IS ready and should not be downgraded.

Additionally, `getDeploymentCondition` had a bug where `reasons = append(...)` was placed after `break`, making it unreachable for multi-node deployments.

## Changes

- Use `strings.Contains` instead of exact match for `NewReplicaSetAvailable` reason detection
- Add guard to prevent Unknown downgrade when `Available=True` is already confirmed
- Fix `reasons` append placement in `getDeploymentCondition` (move before `break`)
- Move `condition.Reason = strings.Join(...)` inside the `len(statuses) == 2` block
- Add unit tests covering multi-node Available+Progressing race scenarios

## Testing

- Unit tests: All 7 new/modified tests pass
- On-cluster validation: Deployed patched controller via DevSpace cluster with 2 GPU nodes. Multi-node ISVC correctly transitioned to `Ready: True` and maintained that state through rolling updates.

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_status.go
+++ b/pkg/apis/serving/v1beta1/inference_service_status.go
@@ -380,7 +380,7 @@ func (ss *InferenceServiceStatus) PropagateRawStatus(
 				}
 			}
 		} else {
-			// If progressing condition is True, and the reason is set to NewReplicaSetAvailable, override component as ready.
+			// If progressing condition is True, and the reason is NewReplicaSetAvailable, override component as ready.
 			// This is because progressing condition doesn't get set to false when deployment is complete.
 			// See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment
 			if progressingCondition.IsTrue() {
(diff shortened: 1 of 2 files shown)


yashaswipiplani (commented):
LGTM!

Jooho:
Nice fix, the three changes make sense together.

Tiny nit on the comment at the `Available=True` guard — "race" might be slightly misleading since it's not really a concurrency race. It's more of a transient state during rollout where head and worker deployments are at different stages (one already `NewReplicaSetAvailable`, the other still `ReplicaSetUpdated`). Something like "timing window" or "transient mismatch" might be clearer, but totally non-blocking either way.

PR #5705, opened 2026-06-22
feat: add namespace-scoped StorageContainer CRD (#5482)

### LINKED ISSUE #5482: Add namespace-scoped StorageContainer to complement cluster-scoped ClusterStorageContainer
/kind feature

ClusterStorageContainer is cluster-scoped only, which creates limitations in multi-tenant environments:

- Requires cluster-admin privileges to create/modify
- No namespace isolation - all namespaces share same configuration
- Different teams may have conflicting requirements 

Add a namespace-scoped StorageContainer resource following Kubernetes conventions (like Role vs ClusterRole).

Lookup precedence:
1. Namespace-scoped StorageContainer (highest priority)
2. Cluster-scoped ClusterStorageContainer
3. Built-in defaults (lowest priority)

Example:
```
# Namespace-scoped (namespace admins can create)
kind: StorageContainer
metadata:
  name: team-a-config
  namespace: team-a
spec:
  supportedUriFormats:
    - prefix: hf://
  container:
    env:
      - name: HF_HUB_DISABLE_XET
        value: "1"
```

Benefits:
- Multi-tenant isolation - teams can have different configs
- Self-service - namespace admins don't need cluster-admin privileges
- Backward compatible - existing ClusterStorageContainer resources unchanged
- Follows Kubernetes patterns

mwaykole:

Adds a namespace-scoped `StorageContainer` so teams can configure InferenceService storage initialization within their own namespace.

Automatic URI matching prefers an eligible namespaced container over a cluster container. Explicit `storageContainerName` selection checks the namespace first and uses the same-named cluster container only when the namespaced resource is absent. An ineligible explicitly named resource produces an error.

Existing ClusterStorageContainers remain supported. This change does not add storage configuration watches or extend namespaced lookup to LLMInferenceService and local model download jobs.

Usage examples and selection rules: [namespace-scoped storage containers](docs/samples/storage/namespace-scoped-storage-container/README.md).

Closes #5482

- Same-prefix namespaced precedence, named shadowing and deletion fallback.
- Namespace isolation and empty-namespace fallback.
- Disabled, wrong-workload and unsupported-URI behavior for named and automatic selection.
- Passed: 311 tests across `pkg/webhook/admission/pod`, `pkg/controller/v1beta1/inferenceservice/utils`, and `pkg/controller/v1beta1/inferenceservice/components` using Kubernetes 1.36.2 envtest.
- Passed: `make precommit` for code generation, manifests, formatting, linting and e2e test collection.

Rebased onto current master. Removed unused localmodelnode permissions and the implementation design report; the user guide lives with storage samples. LLMInferenceService support remains outside this PR.

## Checklist

- [x] Tests cover the selection behavior.
- [x] User documentation and comments are included.
- [ ] Website documentation: the usage guide is included in this repository; no separate website PR is included.

## Release Notes

```release-note
Add namespace-scoped StorageContainer resources for InferenceService storage initialization, with namespace-first lookup and ClusterStorageContainer fallback.
```


pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -587,6 +588,86 @@ func clusterServingRuntimesPredicate() predicate.Funcs {
 	}
 }
 
+func (r *InferenceServiceReconciler) storageContainerFunc(ctx context.Context, obj client.Object) []reconcile.Request {
```
spolti: Is there a reason to add reconcilers for this?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -722,6 +813,16 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployCo
 		r.Log.Info("The InferenceService controller won't watch serving.kserve.io/v1alpha1/ClusterServingRuntime resources because the CRD is not available.")
 	}
 
+	cscFound, err := utils.IsCrdAvailable(r.ClientConfig, v1alpha1.SchemeGroupVersion.String(), "ClusterStorageContainer")
+	if err != nil {
```
spolti: See the other question, o do t think it is needed.

pkg/controller/v1beta1/inferenceservice/storage_container_watch_test.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2025 The KServe Authors.
```
spolti: Year.

spolti (commented):
Thanks @mwaykole 
Just one comment for now.

mwaykole (commented):
**🤖 PR Health Monitor Report** — E2E tests are failing on `test-predictor`, `test-llm`, and `test-huggingface-server-vllm`; three inline review comments from `spolti` are unaddressed; and two assigned reviewers (`greenmoon55`, `johnugeorge`) have not yet responded. The PR is also 7 days stale.

pkg/controller/v1beta1/inferenceservice/controller.go
mwaykole: Yes — the watch is needed so that when a namespace-scoped `StorageContainer` is created, updated, or deleted, all ISVCs in that namespace get requeued and pick up the new storage configuration. Without it, changes to a `StorageContainer` would only take effect on the next unrelated reconcile trigger (e.g. the ISVC being touched), which breaks the expected "apply and see it reflected" user experience. The predicate ensures we only requeue on spec/disabled changes, not on status updates, keeping the blast radius small.

pkg/controller/v1beta1/inferenceservice/controller.go
mwaykole: Same reasoning as above — `ClusterStorageContainer` needs a watch for the same reason: changes to a cluster-wide storage config should immediately requeue all ISVCs so the new spec is applied without waiting for an unrelated reconcile. Happy to remove both if the preference is to rely on periodic re-sync instead, but that would mean a delay before storage config changes take effect.

pkg/controller/v1beta1/inferenceservice/storage_container_watch_test.go
mwaykole: Good catch — fixed in latest push (2025 → 2026).

pkg/controller/v1beta1/inferenceservice/controller.go
israel-hdez: It would be great if we could re-queue only the impacted ISVCs. But that would make the logic of this func more complex, probably in an unnecessary way if the main Reconciler keeps non-impacted ISVCs stable

docs/namespace-scoped-storage-container.md
```diff
@@ -0,0 +1,364 @@
+# Design: Namespace-Scoped StorageContainer
```
israel-hdez: This looks like a file that we don't need to commit.

israel-hdez (commented):
Basically, this PR adds a new CRD, and adapts resolving the Storage Container configuration to prefer namespaced one first, then cluster scope.

I left one minor comment. Everything else looks good to me.

config/crd/full/serving.kserve.io_storagecontainers.yaml
Jooho, line 1: Don't we need this CRD for llmisvc?

Jooho (commented):
I have 1 question. Do we need this only for predictive ai? or we need this for llmisvc as well? current PR include the crd only for kserve.

pkg/controller/v1beta1/inferenceservice/controller.go
kliukovkin: One thing I noticed while tracing the effect of the watch: for the single-storageUri case the reconcile only sets the source-uri annotation, and the actual init container is resolved later by the pod webhook. So requeuing the ISVC on a StorageContainer change doesn't produce a template diff — existing pods won't roll, and the new spec only takes effect on the next pod churn. The plural StorageUris path does embed the container spec at reconcile time, so the watch delivers the "apply and see it reflected" behavior there, but not for the common single-URI case. Might be worth noting that limitation, or scoping the watch to the path where it actually triggers a rollout.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
```diff
@@ -5617,3 +5624,169 @@ func TestApplyConfidentialConfig(t *testing.T) {
...
+	// namespace="" → no namespaced SCs to list, then ClusterStorageContainer List returns NoMatchError
+	spec, err := GetStorageContainerSpec(ctx, "", "s3://bucket/path", nil, fakeClient)
+	assert.Nil(t, spec)
+	assert.NoError(t, err)
+}
+
+// TestGetStorageContainerSpecNamespaceScoped verifies the two-tier lookup:
+// namespace-scoped StorageContainer is found first; falls back to ClusterStorageContainer
+// when no namespace-scoped match; returns nil when neither tier matches.
+func TestGetStorageContainerSpecNamespaceScoped(t *testing.T) {
```
kliukovkin, line 5899: Could we add a case where a namespace-scoped SC and a ClusterStorageContainer both support the same prefix? That's the headline behavior of this PR (namespace wins over cluster), but the current tests only cover disjoint prefixes (custom:// vs s3://), so the actual precedence is never exercised. Looks like the unchecked "verify precedence" box in the test plan is exactly this.

kliukovkin (commented):
Took a closer look at the lookup and watch paths since this touches the same storage code I've been working in. A few questions and test suggestions, nothing blocking from my side.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -153,39 +182,66 @@ func GetStorageContainerSpec(ctx context.Context, storageUri string, storageCont
...
+			if !supported {
+				return nil, fmt.Errorf("StorageContainer %q does not support storageUri %q", name, storageUri)
+			}
+			return &sc.Spec, nil
+		}
+		if !apierrors.IsNotFound(err) && !apimeta.IsNoMatchError(err) {
+			return nil, fmt.Errorf("failed to fetch StorageContainer %q: %w", name, err)
+		}
+		// Not found in namespace (or CRD not installed) — fall through to ClusterStorageContainer.
+		// Note: this allows a ClusterStorageContainer with the same name to substitute.
```
kliukovkin, line 216: The fallthrough is documented here, but there's no test for the shadowing direction: an ISVC references a ClusterStorageContainer via storageContainerName, then someone creates a namespace SC with the same name and the resolution silently changes. A test pinning that down would help, especially since disabled/wrong-workloadType fail hard instead of falling through — that asymmetry is easy to trip on.

config/rbac/localmodelnode/role.yaml
```diff
@@ -50,6 +50,7 @@ rules:
   resources:
   - clusterstoragecontainers
   - localmodelnodegroups
+  - storagecontainers
```
kliukovkin, line 53: The localmodelnode role gets get/list/watch on storagecontainers here, but the download path (getContainerSpecForStorageUri) still only lists ClusterStorageContainers, so the grant is unused. Wiring namespaced SCs into localmodel downloads has its own design questions (which namespace's SC applies to a job running in the central job namespace?), so it might be cleaner to drop the grant for now and keep the role minimal.

pkg/webhook/admission/pod/storage_initializer_injector.go
```diff
@@ -122,18 +122,47 @@ type StorageInitializerParams struct {
...
+func GetStorageContainerSpec(ctx context.Context, namespace string, storageUri string, storageContainerName *string, c client.Client) (*v1alpha1.StorageContainerSpec, error) {
+	// Named lookup takes priority — skip auto-match entirely
 	if storageContainerName != nil && *storageContainerName != "" {
-		return GetStorageContainerSpecByName(ctx, *storageContainerName, storageUri, client)
+		return GetStorageContainerSpecByName(ctx, namespace, *storageContainerName, storageUri, c)
+	}
+
+	// Auto-match: check namespace-scoped StorageContainer first
+	namespacedList := &v1alpha1.StorageContainerList{}
+	if err := c.List(ctx, namespacedList, client.InNamespace(namespace)); err != nil {
```
kliukovkin: nit: if namespace ever comes in empty here, InNamespace("") lists StorageContainers across all namespaces. Current callers always pass one, but the named-lookup path guards with namespace != "" — same guard here would keep the two paths consistent.

config/crd/full/serving.kserve.io_storagecontainers.yaml
mwaykole: Good question @Jooho! This PR is intentionally scoped to `InferenceService` only. The `StorageContainer` CRD itself is cluster-installed, so llmisvc could use it in a follow-up PR. Wiring the namespace-scoped lookup into the llmisvc storage path has its own design questions (which namespace applies for a centralized job?) that are better handled separately. Happy to open a follow-up issue for that if useful.

pkg/controller/v1beta1/inferenceservice/controller.go
mwaykole: Thanks for the detailed analysis. You are right — the watch was removed in the latest push based on reviewer feedback. The periodic re-sync is sufficient for this use case.

pkg/webhook/admission/pod/storage_initializer_injector_test.go
mwaykole: Good catch — adding a test that exercises the namespace-wins case (both namespace SC and ClusterStorageContainer support the same prefix) is on the list. Will add it in the next push.

pkg/webhook/admission/pod/storage_initializer_injector.go
mwaykole: Agreed — will add a test that covers the shadowing direction: `storageContainerName` set, namespace SC exists with that name and takes precedence over a same-named ClusterStorageContainer.

config/rbac/localmodelnode/role.yaml
mwaykole: Valid point — the localmodelnode download path does not yet use namespace-scoped StorageContainers, so the grant is premature. Will remove it to keep the role minimal.

pkg/webhook/admission/pod/storage_initializer_injector.go
mwaykole: Good nit — will add the `namespace != ""` guard in the auto-match path to match the named-lookup path and prevent an accidental cross-namespace list on empty namespace.

config/crd/full/serving.kserve.io_storagecontainers.yaml
mwaykole:
```
LLMInferenceService → StorageInitializerSpec (its own type)
                       ↓
                    builds init container directly
                    (NOT through GetStorageContainerSpec)

InferenceService → GetStorageContainerSpec → StorageContainer/ClusterStorageContainer
                   ↓
                webhook injection
```

pkg/controller/v1beta1/inferenceservice/controller.go
spolti: thanks @mwaykole.

@israel-hdez it can be done by using the name of the StorageContainer in use by that isvc. This was recently added. If no name is configured, we could choose a default action, just ignore or reconcile.

spolti:
@mwaykole just a side note, we also need to create the docs for this feature.

Once I have some free time, I'll test this as well.

mwaykole:
@kliukovkin  @spolti @Jooho  can i get approval is there anything missing ?

kliukovkin:
Re-checked — watch removal actually resolves my single-URI concern (no more inconsistent rollout between the two paths), and InNamespace("") guard got fixed too, nice catch via the e2e test.

Two nits still open — no test for a namespace SC and ClusterStorageContainer supporting the same prefix (the actual precedence claim), and shadowing fallthrough is documented in a comment but still untested. also the localmodelnode storagecontainers RBAC grant looks unused, download path still only lists ClusterStorageContainerList.

Nothing blocking though.

/lgtm

PR #5710, opened 2026-06-23
fix(llmisvc): skip absent InferencePool CRDs during scheduling

### LINKED ISSUE #5708: LLMInferenceService scheduler: reconcileV1Alpha2InferencePool fails when v1alpha2 CRD is absent, despite "if the CRD is available" doc comment
/kind bug

### Summary

In v0.19.0, `pkg/controller/v1alpha2/llmisvc/scheduler.go` dual-writes InferencePool objects under both API groups:

    func (r *LLMISVCReconciler) reconcileSchedulerInferencePool(...) error {
        ...
        if err := r.reconcileV1InferencePool(ctx, llmSvc, shouldDelete); err != nil {
            return err
        }
        if err := r.reconcileV1Alpha2InferencePool(ctx, llmSvc, shouldDelete); err != nil {
            return err
        }
    }

Both per-group methods carry a doc comment promising graceful behavior when the corresponding CRD is missing:

    // reconcileV1InferencePool reconciles the v1 InferencePool if the CRD is available.
    // reconcileV1Alpha2InferencePool reconciles the v1alpha2 InferencePool if the CRD is available.

But neither implementation actually checks CRD availability. Each just calls the shared `Reconcile` helper, which propagates the API server `NoKindMatchError` / `NotFound`. The whole scheduler reconcile then fails.

### Reproducer

1. Install `kserve-llmisvc-resources` v0.19.0 with `kserve.llmisvc.createGIECRDs=false` and install **only** the stable `inferencepools.inference.networking.k8s.io` CRD.
2. Create an LLMInferenceService.
3. Observe scheduler status:

       Ready=False  SchedulerReconcileError:
         Failed to reconcile scheduler: failed to create v1alpha2.InferencePool
         <ns>/<name>: the server could not find the requested resource
         (post inferencepools.inference.networking.x-k8s.io)

   `HTTPRoutesReady=False` and `RouterReady=False` propagate from this. The main workload deployment never reaches Ready because the scheduler stage gates downstream reconciliation.

### Why this matters

Operators who deliberately want to run on the stable `inference.networking.k8s.io/v1` CRD only (e.g., to avoid carrying the experimental CRD on a production cluster) have no way to do so today, even though the doc comments suggest it is supported. The chart flag `createGIECRDs=false` (#5544) is currently misleading: it only stops the chart from installing the GIE CRDs but the controller still mandates both CRDs at runtime.

### Suggested fix

Match the doc comment: check CRD availability via the RESTMapper before calling `Reconcile`, and skip the branch (with a log line) when the corresponding CRD is not registered. Roughly:

    func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(...) error {
        if !r.isCRDAvailable(igwapiv1alpha2.GroupVersion.WithKind("InferencePool")) {
            log.V(2).Info("skipping v1alpha2 InferencePool reconcile: CRD not installed")
            return nil
        }
        ...
    }

Same fix on the v1 side for symmetry.

### Relation to existing issues

- #5605 (open) covers the routing sidecar `--pool-group` flag and `EndpointPickerConfig` apiVersion migration — distinct scope.
- #5592 (open) tracks the wider migration toward stable / `llm-d.ai` groups — long-term direction.

This issue is the smaller "honor the doc comment" change that unblocks stable-only deployments today.

### Environment

- KServe: v0.19.0
- Stable CRD: `inferencepools.inference.networking.k8s.io/v1` installed
- Experimental CRD: removed
- GKE, GatewayClass `gke-l7-rilb`

Fahmid-Arman:
* Resolves a failure in the `LLMInferenceService` controller where the scheduler reconciliation loop completely fails if either the `v1` or `v1alpha2` InferencePool CRD is absent from the API server.
* The internal `Reconcile()` CRUD helper utilizes `client.IgnoreNotFound()`, which only filters out `apierrors.IsNotFound` but propagates `meta.NoKindMatchError`. This behavior causes unintended reconciliation failures for cluster configurations that deliberately omit experimental or legacy CRDs.
* Modifies both `reconcileV1InferencePool` and `reconcileV1Alpha2InferencePool` in `scheduler.go` to explicitly intercept `meta.IsNoMatchError`. If triggered, the controller logs an information line at level V(2) and exits the branch with a `nil` return state, allowing downstream scheduling logic to execute uninterrupted.
* Introduces a comprehensive test suite `scheduler_no_crd_test.go` utilizing mock interceptors to simulate missing CRD states and guarantee graceful termination.

Fixes #5708

- [x] Added `TestReconcileV1Alpha2InferencePool_WhenCRDNotInstalled_ShouldNotFail`
- [x] Added `TestReconcileV1Alpha2InferencePool_WhenCRDNotInstalled_DeleteShouldNotFail`
- [x] Added `TestReconcileV1InferencePool_WhenCRDNotInstalled_ShouldNotFail`
- [x] Added `TestReconcileV1InferencePool_WhenCRDNotInstalled_DeleteShouldNotFail`
- [x] Added `TestReconcileSchedulerInferencePool_WhenV1Alpha2CRDNotInstalled_ShouldNotFail`

This brings the reconciliation path in line with the `Delete` path in `lifecycle_crud.go` (which already handles `NoMatchError`) and the conditional CRD watch registration in `SetupWithManager`.

**Checklist**:

**Release note**:
```release-note
Fix: Allow `LLMInferenceService` scheduling to proceed gracefully when either `v1` or `v1alpha2` InferencePool CRDs are not installed on the cluster.

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_loader.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader.go
@@ -66,6 +66,13 @@ type SchedulerConfig struct {
 	// determining whether a certificate secret has expired. The first entry is the
 	// write key (set on newly created secrets); all entries are read keys.
 	ExpirationAnnotations []string `json:"expirationAnnotations,omitempty"`
+
+	// AllowSkippingV1Alpha2InferencePool controls whether the scheduler silently skips
+	// the v1alpha2 InferencePool reconciliation when the CRD is not installed.
+	// When false (default), a missing v1alpha2 CRD causes a reconciliation error,
+	// preserving backward-compatible behavior. Set to true in environments where
+	// only the stable v1 InferencePool CRD is installed.
+	AllowSkippingV1Alpha2InferencePool bool `json:"allowSkippingV1Alpha2InferencePool,omitempty"`
 }
 
 // NewSchedulerConfig parses the "scheduler" key from the inferenceservice-config
--- a/pkg/controller/v1alpha2/llmisvc/scheduler.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler.go
@@ -32,6 +32,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -199,7 +200,7 @@ func (r *LLMISVCReconciler) reconcileSchedulerInferencePool(ctx context.Context,
 	return nil
 }
 
-// reconcileV1InferencePool reconciles the v1 InferencePool if the CRD is available.
+// reconcileV1InferencePool reconciles the v1 InferencePool.
 func (r *LLMISVCReconciler) reconcileV1InferencePool(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, shouldDelete bool) error {
 	expected := r.expectedSchedulerInferencePool(ctx, llmSvc)
 	if shouldDelete {
@@ -209,12 +210,30 @@ func (r *LLMISVCReconciler) reconcileV1InferencePool(ctx context.Context, llmSvc
 }
 
 // reconcileV1Alpha2InferencePool reconciles the v1alpha2 InferencePool if the CRD is available.
+// If the CRD is not installed (meta.NoKindMatchError) and Config.AllowSkippingV1Alpha2InferencePool
+// is true, the reconciliation is skipped gracefully. When the flag is false (default),
+// the error is returned to preserve backward-compatible behavior.
 func (r *LLMISVCReconciler) reconcileV1Alpha2InferencePool(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService, shouldDelete bool) error {
 	expected := r.expectedSchedulerInferencePoolV1Alpha2(ctx, llmSvc)
+	var err error
 	if shouldDelete {
-		return Delete(ctx, r, llmSvc, expected)
+		err = Delete(ctx, r, llmSvc, expected)
+	} else {
+		err = Reconcile(ctx, r, llmSvc, &igwapiv1alpha2.InferencePool{}, expected, semanticInferencePoolV1Alpha2IsEqual)
 	}
-	return Reconcile(ctx, r, llmSvc, &igwapiv1alpha2.InferencePool{}, expected, semanticInferencePoolV1Alpha2IsEqual)
+	if meta.IsNoMatchError(err) {
+		cfg, cfgErr := r.loadConfig(ctx)
+		if cfgErr != nil {
+			log.FromContext(ctx).Error(cfgErr, "Failed to load config while checking AllowSkippingV1Alpha2InferencePool")
+			return err
+		}
+		if cfg.SchedulerConfig != nil && cfg.SchedulerConfig.AllowSkippingV1Alpha2InferencePool {
+			log.FromContext(ctx).V(2).Info("v1alpha2 InferencePool CRD not available, skipping (AllowSkippingV1Alpha2InferencePool=true)",
+				"group", igwapiv1alpha2.SchemeGroupVersion.Group, "version", igwapiv1alpha2.SchemeGroupVersion.Version)
+			return nil
+		}
+	}
+	return err
 }
 
 func (r *LLMISVCReconciler) reconcileSchedulerService(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
--- a/pkg/controller/v1alpha2/llmisvc/scheduler_no_crd_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler_no_crd_test.go
@@ -1,0 +1,252 @@
+/*
+Copyright 2026 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: pkg/controller/v1alpha2/llmisvc/scheduler.go [hunk 1]
```diff
@@ -200,21 +201,37 @@ func (r *LLMISVCReconciler) reconcileSchedulerInferencePool(ctx context.Context,
...
 	if shouldDelete {
-		return Delete(ctx, r, llmSvc, expected)
+		err = Delete(ctx, r, llmSvc, expected)
+	} else {
+		err = Reconcile(ctx, r, llmSvc, &igwapiv1alpha2.InferencePool{}, expected, semanticInferencePoolV1Alpha2IsEqual)
+	}
+	if meta.IsNoMatchError(err) {
+		log.FromContext(ctx).V(2).Info("v1alpha2 InferencePool CRD not available, skipping", "group", igwapiv1alpha2.SchemeGroupVersion.Group, "version", igwapiv1alpha2.SchemeGroupVersion.Version)
+		return nil
 	}
```
pierDipi, line 235: can we have feature flags with this silent skip and be disabled by default to ensure backward compatibility?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -200,21 +201,37 @@ func (r *LLMISVCReconciler) reconcileSchedulerInferencePool(ctx context.Context,
...
 	if shouldDelete {
-		return Delete(ctx, r, llmSvc, expected)
+		err = Delete(ctx, r, llmSvc, expected)
+	} else {
+		err = Reconcile(ctx, r, llmSvc, &igwapi.InferencePool{}, expected, semanticInferencePoolIsEqual)
+	}
+	if meta.IsNoMatchError(err) {
+		log.FromContext(ctx).V(2).Info("v1 InferencePool CRD not available, skipping", "group", igwapi.GroupVersion.Group, "version", igwapi.GroupVersion.Version)
+		return nil
 	}
```
pierDipi: are we expected that none of the 2 CRDs are installed? afaik, the creation of the InferencePool objects is drived by `scheduler: {}` if none are installed by having `scheduler` field unspecified it should not create one

pierDipi: I see the original issue only wants to skip v1alpha2 infPool object

pierDipi: I'd suggest to leave v1 as is and handle this only for v1alpha2, the other area to double check if needs updates is the readiness and migration to v1 logic

Fahmid-Arman:
@pierDipi Thanks for the review. I have refactored the implementation to address all three points:

1. **Reverted `v1`:** `reconcileV1InferencePool` has been reverted to its original state. It no longer intercepts `NoMatchError`.
2. **Feature Flag (`v1alpha2`):** Introduced `AllowSkippingV1Alpha2InferencePool` (default: `false`) into `SchedulerConfig` (parsed from the `scheduler` key in the `inferenceservice-config` ConfigMap). `reconcileV1Alpha2InferencePool` now explicitly checks this flag before deciding to suppress `meta.IsNoMatchError`. If the flag is disabled, the error propagates exactly as it did before.
3. **Readiness & Migration Logic:** I verified `EvaluateInferencePoolConditions` in `router.go`. It already handles the `v1alpha2` CRD absence correctly (Lines 728-732). It treats `meta.IsNoMatchError` identically to `NotFound` and will mark the condition as `Ready` as long as the `v1` InferencePool successfully initializes. No changes were required there.

The test suite in `scheduler_no_crd_test.go` has also been rewritten to explicitly assert the behavior for both the enabled and disabled states of the new feature flag.

bartoszmajsak:
@Fahmid-Arman there are compilation failures for tests

```
make test
go fmt ./pkg/... ./cmd/... && cd qpext && go fmt ./...
pkg/openapi/openapi_generated.go
go vet ./pkg/... ./cmd/... && cd qpext && go vet ./...
# github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc [github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc.test]
pkg/controller/v1alpha2/llmisvc/scheduler_no_crd_test.go:136:12: cannot use v1alpha2.InferencePoolSpec{} (value of struct type "github.com/kserve/kserve/pkg/apis/serving/v1alpha2".InferencePoolSpec) as *"github.com/kserve/kserve/pkg/apis/serving/v1alpha2".InferencePoolSpec value in struct literal
make: *** [Makefile:78: vet] Error 1
```

and it would be good to run `make precommit` on those changes.

pkg/controller/v1alpha2/llmisvc/config_loader.go
```diff
@@ -66,6 +66,13 @@ type SchedulerConfig struct {
 	// determining whether a certificate secret has expired. The first entry is the
 	// write key (set on newly created secrets); all entries are read keys.
 	ExpirationAnnotations []string `json:"expirationAnnotations,omitempty"`
+
+	// AllowSkippingV1Alpha2InferencePool controls whether the scheduler silently skips
+	// the v1alpha2 InferencePool reconciliation when the CRD is not installed.
+	// When false (default), a missing v1alpha2 CRD causes a reconciliation error,
+	// preserving backward-compatible behavior. Set to true in environments where
+	// only the stable v1 InferencePool CRD is installed.
+	AllowSkippingV1Alpha2InferencePool bool `json:"allowSkippingV1Alpha2InferencePool,omitempty"`
```
bartoszmajsak, line 75: We already know it at startup time

https://github.com/kserve/kserve/blob/52552b71dc918b34e730509561080a4ad2a2197c/pkg/controller/v1alpha2/llmisvc/controller.go#L440-L443

so instead of adding configmap property, I think we are fine with pushing this info down to reconciler as an internal boolean flag (`r.v1alpha2PoolCRDAvailable` for example).

This would simplify the code even further as we can have a guard in `reconcileV1Alpha2InferencePool` returning early.

bartoszmajsak:
FWIW I opened new PR based on this work and discussions #5890 to ship it for the upcoming release.

bartoszmajsak:
Incorporated in #5890

PR #5712, opened 2026-06-24
feat(llmisvc): replace UDS tokenizer sidecar with vLLM render deployment

vivekk16:
## What

Replaces the UDS tokenizer sidecar (bundled inside the scheduler pod) with a standalone tokenizer Deployment and ClusterIP Service that runs vLLM's `/v1/completions/render` endpoint over HTTP, and decomposes the monolithic `precise-prefix-cache-scorer` plugin into the new 3-plugin pipeline (`token-producer` + `precise-prefix-cache-producer` + `prefix-cache-scorer`).

## Why

The llm-d router v0.9.0+ deprecates the UDS-based tokenizer sidecar in favor of a standalone vLLM instance serving the render endpoint over HTTP. This enables:
- Independent scaling, lifecycle management, and health checking of the tokenizer
- The new 3-plugin pipeline that replaces the monolithic `precise-prefix-cache-scorer`
- Cleaner separation of concerns: tokenization (HTTP) is decoupled from cache indexing and scoring

### API & Types
- **`v1alpha2/llm_inference_service_types.go`** / **`v1alpha1/llm_inference_service_types.go`**: Added `TokenizerSpec` struct and `Tokenizer *TokenizerSpec` field to `SchedulerSpec` for optional operational overrides
- **`v1alpha2/llm_inference_service_lifecycle.go`**: New `TokenizerReady` condition with `MarkTokenizerReady` / `MarkTokenizerNotReady` / `MarkTokenizerUnset` helpers; included in `DetermineRouterReadiness` aggregation
- **`v1alpha1/llm_inference_service_conversion.go`**: Roundtrip conversion for the tokenizer field between v1alpha1 ↔ v1alpha2

### Controller
- **New `tokenizer.go`**: Reconciler that creates a dedicated tokenizer Deployment + ClusterIP Service. Auto-deploys when `token-producer` or `precise-prefix-cache-scorer` plugins are detected in `config.inline`, or when `spec.router.scheduler.tokenizer` is explicitly set
- **`scheduler.go`**:
  - Strips the legacy UDS tokenizer sidecar from the scheduler pod when standalone tokenizer is active (`withStripUdsTokenizerSidecar`)
  - Decomposes `precise-prefix-cache-scorer` into the 3-plugin pipeline with `token-producer` pointing at the tokenizer Service URL (`withDecomposePrecisePrefixCacheScorer`)
  - Injects `modelName` and `vllm.url` into bare `token-producer` plugins (`withInjectTokenProducerConfig`)
  - Migrates parameters from the old monolithic plugin to `precise-prefix-cache-producer`, stripping `tokenizersPoolConfig` (`migrateProducerParams`)
  - Emits deprecation warning event when legacy `precise-prefix-cache-scorer` is detected
  - All new transforms are version-gated to EPP ≥ 0.9.0
- **`config_merge.go`**: Added `kserve-config-llm-tokenizer` to well-known configs; auto-includes the tokenizer config ref when tokenizer-requiring plugins are detected

### Config Templates
- **New `config-llm-tokenizer.yaml`**: Well-known tokenizer config with vLLM render container (`docker.io/vllm/vllm-openai-cpu:v0.23.0` pinned to SHA256 digest), resource requests/limits, readiness probe, and security context
- **`config-llm-scheduler.yaml`**: Removed UDS tokenizer sidecar container, associated volumes (`tokenizer-uds`, `tokenizer-tmp`, `tokenizer-cache`), and volume mounts from the scheduler pod template
- **`resources.yaml`**: Same UDS sidecar removal; added standalone tokenizer `LLMInferenceServiceConfig`

### Samples
- **`llmisvc_config_prefix_cache.yaml`**: Rewritten to use decomposed 3-plugin pipeline; removed custom scheduler template (tokenizer is now auto-deployed)
- **`precise-prefix-kv-cache-routing/`**: Updated both CPU simulator and GPU samples to use `token-producer` + `precise-prefix-cache-producer` + `prefix-cache-scorer`
- **READMEs**: Updated architecture descriptions to reflect the 3-plugin pipeline and standalone tokenizer

### Tests
- **`tokenizer_test.go`**: Unit tests for tokenizer deployment/service construction, plugin detection, deletion, model attachment, and status propagation
- **`controller_int_tokenizer_test.go`**: Integration tests for clean-path (explicit `token-producer` in inline), migration-path (legacy `precise-prefix-cache-scorer` auto-decomposed), and cleanup when tokenizer is removed
- **`test_llm_inference_service.py`** (e2e):
  - Clean-path test: `token-producer` in inline → tokenizer auto-deployed
  - Migration-path test: legacy `precise-prefix-cache-scorer` → auto-decomposed + tokenizer auto-deployed
  - `assert_tokenizer_pipeline_wired`: Verifies scheduler's `--config-text` has `modelName` and tokenizer Service URL
  - `TokenizerReady` condition asserted as True when present

### Backward Compatibility
- The legacy UDS sidecar path is preserved for scheduler versions < 0.9.0
- Existing `precise-prefix-cache-scorer` configs are automatically migrated with a deprecation warning event


config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -104,37 +104,37 @@ spec:
...
-                name: tokenizer-uds
           - name: tokenizer
-            ports:
-              - containerPort: 8082
-                name: health
-                protocol: TCP
-            image: ghcr.io/llm-d/llm-d-uds-tokenizer:vllm-v0.19.1
+            image: vllm/vllm-openai-cpu:v0.23.0
             imagePullPolicy: IfNotPresent
+            command: ["vllm"]
```
zdtsw: not sure why we did in this way. 
cannot just remove it from the this  yaml
and skip `stripTokenizerSidecar() `totally?

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,264 @@
...
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      tokenizerDeploymentName(llmSvc),
+			Namespace: llmSvc.GetNamespace(),
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(llmSvc, v1alpha2.LLMInferenceServiceGVK),
+			},
+			Labels: labels,
+		},
+		Spec: appsv1.DeploymentSpec{
```
zdtsw, line 174: i think the main idea to have it run as standalone so we can easily scale it up.
i guess for now, user will need to manually scale up on the pod
would be good (in the later release) add support for this

config/llmisvcconfig/config-llm-scheduler.yaml
zdtsw: this wont be enough i think,
1cpu + 4G mem will be a good start point.

config/llmisvcconfig/config-llm-scheduler.yaml
vivekk16: The tokenizer container spec is kept in the scheduler config as a single source of truth for both code paths. When the scheduler version is < 0.9.0, it is used as a UDS sidecar in the scheduler pod (legacy). When >= 0.9.0, `extractTokenizerContainer()` reads it to build the standalone Deployment, then `stripTokenizerSidecar()` removes it from the scheduler. Removing it from the YAML would break backward compatibility with older router versions.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Agreed, standalone tokenizer enables independent scaling, which is one of the main motivations. For this initial PR, it defaults to 1 replica and users can manually scale. I'll follow up with configurable replicas / HPA support in a subsequent PR.

config/llmisvcconfig/config-llm-scheduler.yaml
vivekk16: @zdtsw Where are these values coming from? Do you have any benchmark?

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -448,6 +450,60 @@ func isUsingTokenizerSidecar(spec v1alpha2.LLMInferenceServiceSpec) bool {
...
+		}
+	}
+	var filtered []corev1.Volume
+	for _, v := range tmpl.Volumes {
+		if !mountNames[v.Name] || usedByOthers[v.Name] {
+			filtered = append(filtered, v)
+		}
+	}
+	tmpl.Volumes = filtered
+}
```
pierDipi: if we have a well-known config that gets injected to set `tokenizer: ...` field, we wouldn't need this and it's more explicit

pierDipi: actually even now, why is this needed?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -451,9 +461,6 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 				return d, fmt.Errorf("failed to attach model artifacts to scheduler deployment: %w", err)
 			}
 
-			// Mutate scheduler config: inject UDS tokenizer settings, migrate
-			// tokenProcessorConfig from indexerConfig to top-level parameters, and
-			// rename deprecated blockSize to blockSizeTokens (schema changes in v0.6.0).
```
pierDipi, line 456: why comments are gone?

pierDipi (commented):
can you update the precise samples?

pkg/controller/v1alpha2/llmisvc/config_merge.go
vivekk16: Dead code removed.

pkg/controller/v1alpha2/llmisvc/scheduler.go
vivekk16: retrieved the comments.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,227 @@
...
+// is set on the scheduler spec AND the scheduler version is >= 0.9.0. The version
+// gate ensures that old scheduler binaries (which only understand UDS sidecars)
+// are not paired with a standalone tokenizer Deployment.
+func shouldDeployStandaloneTokenizer(spec v1alpha2.LLMInferenceServiceSpec) bool {
+	return spec.Router != nil &&
+		spec.Router.Scheduler != nil &&
+		spec.Router.Scheduler.Tokenizer != nil &&
+		spec.Router.Scheduler.Tokenizer.Template != nil &&
+		schedulerVersionAtLeast(spec, tokenizerVersionGate)
+}
```
pierDipi: should we optimize this and say `&& "only when precise-prefix-scorer is used"`?

pierDipi: or if this check is used for other places, we can add that && in `reconcileTokenizer...`

vivekk16: Makes sense. I will add the plugin check in `reconcileTokenizerDeployment/reconcileTokenizerService` by checking for the plugin presence.

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -104,79 +104,73 @@ spec:
...
+                  path: /health
+                  port: 8000
+                periodSeconds: 10
+                initialDelaySeconds: 5
+                timeoutSeconds: 5
+                failureThreshold: 60
+              resources:
+                requests:
+                  cpu: 256m
+                  memory: 500Mi
```
pierDipi: we probably want to define some limits here, is llm-d defining them that we can re-use?

config/llmisvcconfig/config-llm-scheduler.yaml
zdtsw: https://github.com/llm-d/llm-d/blob/main/guides/precise-prefix-cache-routing/render/deployment.yaml#L53

config/llmisvcconfig/config-llm-scheduler.yaml
pierDipi: https://github.com/llm-d/llm-d/blob/main/guides/precise-prefix-cache-routing/render/deployment.yaml#L53-L61

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,256 @@
...
+
+// tokenizerEndpointURL returns the in-cluster URL for the tokenizer Service.
+func tokenizerEndpointURL(llmSvc *v1alpha2.LLMInferenceService) string {
+	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d",
+		tokenizerServiceName(llmSvc),
+		llmSvc.GetNamespace(),
+		tokenizerServicePort,
+	)
+}
```
pierDipi: we need to use the `network.GetClusterDomainName()` function to not hardcode cluster.local as it might not be the cluster domain, see other functions where we do this.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -451,9 +468,6 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
...
```
pierDipi, line 456: comments are still removed, why?

vivekk16: Missed pushing the commit that restored the commit. My bad.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Done

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,256 @@
...
+		spec.Router.Scheduler.Tokenizer != nil &&
+		spec.Router.Scheduler.Tokenizer.Template != nil &&
+		schedulerVersionAtLeast(spec, tokenizerVersionGate)
+}
+
+// hasPrefixCachePlugin reports whether the scheduler's EndpointPickerConfig
+// contains the precise-prefix-cache-scorer plugin. It returns true when the
+// plugin is present or when the config cannot be inspected (e.g. ConfigMap ref),
+// so callers default to deploying the tokenizer in ambiguous cases.
+func hasPrefixCachePlugin(spec v1alpha2.LLMInferenceServiceSpec) bool {
```
pierDipi: nit: maybe `hasPrecisePrefixCachePlugin`?

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if isStopped := utils.GetForceStopRuntime(llmSvc); isStopped || !shouldDeployStandaloneTokenizer(llmSvc.Spec) || !hasPrefixCachePlugin(llmSvc.Spec) {
+		return Delete(ctx, r, llmSvc, expected)
+	}
+
+	if err := Reconcile(ctx, r, llmSvc, &appsv1.Deployment{}, expected, semanticDeploymentIsEqual, PreserveDeploymentReplicas()); err != nil {
+		return fmt.Errorf("failed to reconcile tokenizer deployment %s/%s: %w", expected.GetNamespace(), expected.GetName(), err)
+	}
+	return nil
```
pierDipi: should we report readiness of this deployment?

vivekk16: Yes. I will do it in a follow up PR if that's okay. Didn't want to bloat the changes.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Sure. I will make the change.

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -107,79 +104,76 @@ spec:
...
+                  memory: 4Gi
+                limits:
+                  cpu: "4"
+                  memory: 12Gi
+              terminationMessagePath: /dev/termination-log
+              terminationMessagePolicy: FallbackToLogsOnError
+              securityContext:
+                allowPrivilegeEscalation: false
+                runAsNonRoot: true
+                runAsUser: 1000
```
pierDipi: do we need this?

config/llmisvcconfig/config-llm-scheduler.yaml
```diff
@@ -107,79 +104,76 @@ spec:
...
-          - name: tokenizer-cache
-            emptyDir: {}
         dnsPolicy: ClusterFirst
         restartPolicy: Always
         terminationGracePeriodSeconds: 60 # must be >= workload terminationGracePeriodSeconds
+      tokenizer:
+        template:
+          containers:
+            - name: tokenizer
+              image: vllm/vllm-openai-cpu:v0.23.0
```
pierDipi:
```suggestion
              image: docker.io/vllm/vllm-openai-cpu:v0.23.0@sha256:89e1fbe829038b1a560601990c8ec89625c622a969fcbc8f10ce48b6139795be
```

docs/samples/llmisvc/e2e-gpt-oss/llmisvc_config_prefix_cache.yaml
```diff
@@ -45,13 +45,15 @@ spec:
           periodSeconds: 10
   router:
     scheduler:
+      tokenizer: {}
```
pierDipi: is this needed?

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -412,7 +442,11 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
 			injectSchedulerTracing(llmSvc.Spec.Tracing, llmSvc.GetNamespace(), llmSvc.GetName(), mainContainer)
 		}
 
-		if isUsingTokenizerSidecar(llmSvc.Spec) {
+		if isTokenizerEnabled(llmSvc.Spec) {
+			// Strip the legacy UDS tokenizer sidecar if it was included via
+			// the well-known config or versioned config pinning.
+			withStripUdsTokenizerSidecar(d)
```
pierDipi: when is this case possible?

pierDipi: oh I see, the logic is changed quite a bit from last review

docs/samples/llmisvc/e2e-gpt-oss/llmisvc_config_prefix_cache.yaml
vivekk16: Updated the logic as per our conversation.

config/llmisvcconfig/config-llm-scheduler.yaml
vivekk16: Done

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -412,7 +443,11 @@ func (r *LLMISVCReconciler) expectedSchedulerDeployment(ctx context.Context, llm
...
```
pierDipi: do we need this logic? the well-known config doesn't have the sidecar anymore in 0.9+ configs so we don't need to strip the sidecar as it won't be there anyway?

pierDipi (approved):

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,60 @@
...
+    scheduler:
+      tokenizer:
+        template:
+          automountServiceAccountToken: false
+          containers:
+            - name: vllm-render
+              image: docker.io/vllm/vllm-openai-cpu:v0.23.0@sha256:89e1fbe829038b1a560601990c8ec89625c622a969fcbc8f10ce48b6139795be
+              imagePullPolicy: IfNotPresent
+              command: ["vllm", "launch", "render"]
+              args:
```
pierDipi: I missed this but we need to pass TLS flags when enabled (+ volume mount similar to the main workload)

https://github.com/kserve/kserve/blob/bacf590e37b8daa8160b30f46e592842439bccac/config/llmisvcconfig/config-llm-template.yaml#L172-L174

https://github.com/kserve/kserve/blob/master/config/llmisvcconfig/config-llm-template.yaml#L204-L226

https://github.com/kserve/kserve/blob/bacf590e37b8daa8160b30f46e592842439bccac/config/llmisvcconfig/config-llm-template.yaml#L251-L253

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -65,6 +65,7 @@ const (
 	configRouterSchedulerNameSuffix           = "config-llm-scheduler"
 	configRouterRouteNameSuffix               = "config-llm-router-route"
 	configSchedulerLatencyPredictorNameSuffix = "config-llm-scheduler-latency-predictor"
+	configTokenizerNameSuffix                 = "config-llm-tokenizer" // #nosec G101
```
bartoszmajsak, line 69: @vivekk16 is it complaining b/c of the word `token` here? Why is this comment only here?

config/llmisvcconfig/config-llm-tokenizer.yaml
pierDipi: can we also use command for our well-known configs and leave args empty for users to use?

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,274 @@
...
+func tokenizerDeploymentName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return kmeta.ChildName(llmSvc.GetName(), "-tokenizer")
+}
+
+func tokenizerServiceName(llmSvc *v1alpha2.LLMInferenceService) string {
+	return kmeta.ChildName(llmSvc.GetName(), "-tokenizer")
+}
+
+func tokenizerServiceURL(llmSvc *v1alpha2.LLMInferenceService) string {
+	return fmt.Sprintf("http://%s.%s.svc.%s:%d",
```
pierDipi: see comment on TLS

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,60 @@
...
+metadata:
+  name: kserve-config-llm-tokenizer
+spec:
+  router:
+    scheduler:
+      tokenizer:
+        template:
+          automountServiceAccountToken: false
+          containers:
+            - name: vllm-render
```
pierDipi: using main as the name might be more consistent with the rest (easier to remember/guess as a user)

pkg/controller/v1alpha2/llmisvc/config_merge.go
vivekk16: Yes @bartoszmajsak. It flags `configTokenizerNameSuffix` and `tokenizerRenderContainerName` because of "`token`izer". It only applies to const/var with string literal assignments. Function names, struct fields don't trigger it, which is why the comment isn't elsewhere and just these two lines.

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  value: /tmp/hf
+                - name: DO_NOT_TRACK
+                  value: "1"
+              resources:
+                requests:
+                  cpu: "1"
+                  memory: 4Gi
+                limits:
+                  cpu: "4"
+                  memory: 12Gi
```
bartoszmajsak, line 40: Curious - where are these numbers coming from?

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,60 @@
...
+                httpGet:
+                  path: /health
+                  port: render-http
+                initialDelaySeconds: 30
+                periodSeconds: 10
+                timeoutSeconds: 5
+                failureThreshold: 60
+              securityContext:
+                allowPrivilegeEscalation: false
+                readOnlyRootFilesystem: false
```
bartoszmajsak: Is this necessary?

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  memory: 12Gi
+              readinessProbe:
+                httpGet:
+                  path: /health
+                  port: render-http
+                initialDelaySeconds: 30
+                periodSeconds: 10
+                timeoutSeconds: 5
+                failureThreshold: 60
+              securityContext:
```
bartoszmajsak, line 50: Can we harden it?

```suggestion
              securityContext:
                runAsNonRoot: true
```

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -1434,6 +1477,234 @@ func withRemoveUnnecessaryTokenizer(d *appsv1.Deployment) mutateSchedulerConfigF
...
+
+		// Remove any existing prefix-cache-scorer to avoid duplicates after
+		// the decomposed pipeline injects its own prefix-cache-scorer.
+		remaining := make([]interface{}, 0, len(plugins))
+		for i, plugin := range plugins {
+			if i == idx {
+				continue
+			}
+			pluginMap, ok := plugin.(map[string]interface{})
+			if ok && pluginMap["type"] == prefixCacheScorerPlugin {
```
bartoszmajsak, line 1659: To be on the safe side maybe we can dedup every plugin?

pkg/controller/v1alpha2/llmisvc/controller_int_tokenizer_test.go
```diff
@@ -0,0 +1,527 @@
...
+					Name:      tokenizerName,
+					Namespace: testNs.Name,
+				}, dep)).To(Succeed())
+
+				g.Expect(dep).To(BeOwnedBy(llmSvc))
+				containerNames := containerNameList(dep)
+				g.Expect(containerNames).To(ContainElement("vllm-render"))
+			}).WithContext(ctx).Should(Succeed())
+
+			// Tokenizer Service should also exist
```
bartoszmajsak, line 300: Would it make sense to check 3-plugin pipeline contract here?

pkg/controller/v1alpha2/llmisvc/tokenizer_test.go
```diff
@@ -0,0 +1,632 @@
...
+}
+
+func TestDecomposePluginPipeline(t *testing.T) {
+	tests := []struct {
+		name           string
+		configYAML     string
+		tokenizerURL   string
+		validateConfig func(g Gomega, configText string)
+	}{
+		{
```
bartoszmajsak, line 205: Is it valid case?

```suggestion
		{
      name: "promotes tokenProcessorConfig from inside indexerConfig to top-level producer params",
      configYAML: `apiVersion: inference.networking.x-k8s.io/v1alpha1
  kind: EndpointPickerConfig
  plugins:
  - type: single-profile-handler
  - type: precise-prefix-cache-scorer
    parameters:
      indexerConfig:
        tokenizersPoolConfig:
          modelName: base
          uds:
            socketFile: /tmp/tokenizer/tokenizer-uds.socket
        tokenProcessorConfig:
          blockSize: 16
        kvBlockIndexConfig:
          enableMetrics: true
  - type: max-score-picker
  schedulingProfiles:
  - name: default
    plugins:
    - pluginRef: precise-prefix-cache-scorer
      weight: 3
    - pluginRef: max-score-picker
  `,
      tokenizerURL: "http://my-llm-tokenizer.default.svc.cluster.local:8000",
      validateConfig: func(g Gomega, configText string) {
          g.Expect(configText).NotTo(ContainSubstring("tokenizersPoolConfig"))

          u := unstructured.Unstructured{}
          g.Expect(yaml.Unmarshal([]byte(configText), &u)).To(Succeed())
          val, _, _ := unstructured.NestedFieldNoCopy(u.Object, "plugins")
          plugins := val.([]interface{})
          producerPlugin := plugins[2].(map[string]interface{})
          g.Expect(producerPlugin["type"]).To(Equal(precisePrefixCacheProducerPlugin))
          params := producerPlugin["parameters"].(map[string]interface{})

          // tokenProcessorConfig should be promoted to top-level params,
          // not left nested inside indexerConfig where the producer won't find it
          g.Expect(params).To(HaveKey("tokenProcessorConfig"))
          tpc := params["tokenProcessorConfig"].(map[string]interface{})
          g.Expect(tpc["blockSize"]).To(BeEquivalentTo(16))

          // indexerConfig should still have kvBlockIndexConfig but not tokenProcessorConfig
          indexerConfig := params["indexerConfig"].(map[string]interface{})
          g.Expect(indexerConfig).To(HaveKey("kvBlockIndexConfig"))
          g.Expect(indexerConfig).NotTo(HaveKey("tokenProcessorConfig"))
      },
  },
  {
```

pkg/controller/v1alpha2/llmisvc/controller_int_tokenizer_test.go
```diff
@@ -0,0 +1,527 @@
...
+*/
+
+package llmisvc_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
```
bartoszmajsak: Perhaps we only need one?

bartoszmajsak (commented):
Couple of questions and suggestions on my end

config/llmisvcconfig/config-llm-tokenizer.yaml
zdtsw: it is from the upstream guide https://github.com/llm-d/llm-d/blob/main/guides/precise-prefix-cache-routing/render/deployment.yaml#L56

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: Yes. I'll update to use commands and leave args empty.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Passing the TLS flags when enabled. Made the relevant change here too.

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: Sure.

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: `readOnlyRootFilesystem: false` is the k8s default. The upstream llm-d render deployment doesn't set it either. Happy to remove it to keep the YAML clean.

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: Done.

pkg/controller/v1alpha2/llmisvc/controller_int_tokenizer_test.go
vivekk16: Removed the duplicate.

pkg/controller/v1alpha2/llmisvc/scheduler.go
vivekk16: Duplicates can't occur under normal usage. `withDecomposePrecisePrefixCacheScorer` only runs when `precise-prefix-cache-scorer` is present. If the user already has the new 3 plugin pipeline, there's nothing to decompose and the function no ops. The one real collision (`prefix-cache-scorer`) is already explicitly stripped before injection.

The only way to get a duplicate is if a user puts both `precise-prefix-cache-scorer` (old) and `token-producer` or  `precise-prefix-cache-producer` (new) in the same config. That's not a realistic scenario.

Happy to add a dedup function for plugins across entire migration logic if it's still necessary as a follow-up PR.

pkg/controller/v1alpha2/llmisvc/tokenizer_test.go
vivekk16: Good catch. I verified in llm-d source code that the old `kvcache.Config` (the type behind indexerConfig) used to include `tokenProcessorConfig` as a field, but the current version removed it. The new `precise-prefix-cache-producer PluginConfig` expects `tokenProcessorConfig` at the top level, so if it stays nested after migration it gets silently ignored and the producer falls back to defaults.

Fixed `migrateProducerParams` to extract `tokenProcessorConfig` from `indexerConfig` and promote it to top-level parameters during migration. If a top-level one already exists it takes priority. Added unit tests covering promotion, priority, and no-op cases, and strengthened the existing integration test assertions to verify the values are preserved.

pkg/controller/v1alpha2/llmisvc/controller_int_tokenizer_test.go
vivekk16: Yes. Added an assertion for this.

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,73 @@
...
+              imagePullPolicy: IfNotPresent
+              command:
+                - vllm
+                - launch
+                - render
+                - /mnt/models/base
+                - "--port=8000"
+                - '{{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }}'
+                - '{{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt{{- end }}'
+                - '{{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key{{- end }}'
```
pierDipi: we need to include the tokenizer in SANs and IPs of the certificate too, I think IPs should be covered already through consistent labels (let's double check), the service address is not

https://github.com/kserve/kserve/blob/97895ee00320681f6ae7a7d790fc31d165bdf5c8/pkg/controller/v1alpha2/llmisvc/workload_tls_self_signed.go#L275-L303

https://github.com/kserve/kserve/blob/97895ee00320681f6ae7a7d790fc31d165bdf5c8/pkg/controller/v1alpha2/llmisvc/workload_tls_self_signed.go#L305-L354

vivekk16: tokenizer pod IPs and service ClusterIP are already covered by `collectIPAddresses`. Added the tokenizer service DNS to `collectDNSNames` (gated on isTokenizerEnabled) since it was the only missing piece.

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,73 @@
...
+              env:
+                - name: HF_HOME
+                  value: /tmp/hf
+                - name: DO_NOT_TRACK
+                  value: "1"
+              resources:
+                requests:
+                  cpu: "1"
+                  memory: 4Gi
+                limits:
```
mwaykole, line 38: 4Gi request / 12Gi limit for a CPU-only tokenizer (vLLM render endpoint) seems high. The old UDS sidecar used 256m CPU + 500Mi memory. Is the 24× memory increase based on measured usage, or a conservative estimate? If the tokenizer is purely doing tokenization + render, models like opt-125m shouldn't need more than ~1-2Gi.

If this is sized for large-vocab models, a comment explaining the sizing rationale would help downstream operators avoid over-provisioning.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -467,6 +467,23 @@ type SchedulerSpec struct {
...
+	// token-producer or precise-prefix-cache-scorer plugins are detected in the
+	// scheduler config.inline. Use this field to customize the tokenizer pod
+	// template (resources, image overrides, etc.).
+	// +optional
+	Tokenizer *TokenizerSpec `json:"tokenizer,omitempty"`
+}
+
+// TokenizerSpec configures a standalone tokenizer deployment.
+type TokenizerSpec struct {
+	// Template for the tokenizer pod spec. This is merged on top of the
```
mwaykole, line 483: The docstring says this template is "merged on top of the well-known tokenizer config defaults" but doesn't describe the merge strategy. Is this a strategic merge patch (partial overrides work), or does the user-provided template completely replace the default containers list?

If a user only wants to change resource limits, do they need to provide the full container spec, or can they provide just `containers: [{name: main, resources: ...}]`?

pkg/apis/serving/v1alpha2/llm_inference_service_lifecycle.go
```diff
@@ -321,6 +340,7 @@ func (in *LLMInferenceService) DetermineRouterReadiness() {
 		in.GetStatus().GetCondition(HTTPRoutesReady),
 		in.GetStatus().GetCondition(InferencePoolReady),
 		in.GetStatus().GetCondition(SchedulerWorkloadReady),
+		in.GetStatus().GetCondition(TokenizerReady),
```
mwaykole, line 343: Since `TokenizerReady` is now included in `DetermineRouterReadiness`, the tokenizer reconciler needs to actively call `MarkTokenizerUnset()` when the tokenizer is no longer needed. Otherwise a stale `TokenizerReady=False` from a previous config could block `RouterReady` even after the user removes tokenizer-requiring plugins.

Is there a test covering the transition from tokenizer-enabled to tokenizer-disabled config?

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: values are from the upstream guide https://github.com/llm-d/llm-d/blob/main/guides/precise-prefix-cache-routing/render/deployment.yaml#L56.

pkg/apis/serving/v1alpha2/llm_inference_service_lifecycle.go
vivekk16: When the tokenizer is no longer needed (`shouldDeleteTokenizer`), `reconcileTokenizerDeployment` calls `MarkTokenizerUnset() `before deleting the Deployment, which clears the TokenizerReady condition. `DetermineRouterReadiness` skips nil conditions, so a stale `TokenizerReady=False` should not keep blocking `RouterReady` after disable.

I also extended the enabled → disabled int tests to assert that `TokenizerReady` is cleared after:

removing `spec.router.scheduler.tokenizer`
removing the `token-producer` plugin from the inline scheduler config

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
vivekk16: When the tokenizer is enabled, the well-known `config-llm-tokenizer` preset is injected and merged with the user LLMInferenceService spec via MergeSpecs (strategic merge patch), with the user spec applied last. So `tokenizer.template` is merged the same way as other nested `PodSpec` fields. Containers are merged by name, so a partial override like containers: [{name: main, resources: ...}] is enough.  I left the godoc as-is since this is the general well-known-preset merge behavior rather than something unique to the tokenizer.

config/llmisvcconfig/config-llm-tokenizer.yaml
bartoszmajsak: @vivekk16 I was wondering if we can improve security posture here. Would it work or this is necessary?

pkg/controller/v1alpha2/llmisvc/scheduler.go
bartoszmajsak: Fair, that defend us from nothing :)

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: Oh i mistook it. 

Tested this locally by running the `vllm/vllm-openai-cpu:v0.23.0` image with --read-only via podman. Since the tokenizer uses the CPU only image, the nvidia-cdi-hook constraint that forces `readOnlyRootFilesystem: false` on the GPU templates doesn't apply here.

I set `readOnlyRootFilesystem: true` and widened the emptyDir mount from /tmp/hf to /tmp to cover Python's tempfile and other runtime temp writes. HF_HOME=/tmp/hf is unchanged, HuggingFace creates the subdirectory automatically under the mount. vllm import, tokenizer download, and server startup all worked without filesystem permission errors.

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,74 @@
...
+                initialDelaySeconds: 30
+                periodSeconds: 10
+                timeoutSeconds: 5
+                failureThreshold: 60
+              securityContext:
+                allowPrivilegeEscalation: false
+                readOnlyRootFilesystem: true
+                runAsNonRoot: false
+                capabilities:
+                  drop:
```
mwaykole, line 55: Commit `a354383` reverts this to `runAsNonRoot: false`, but the earlier review thread with @bartoszmajsak was resolved based on `runAsNonRoot: true`. If the vllm-openai-cpu image actually requires root, that's understandable, but the resolved thread should be reopened or a comment left explaining why — otherwise the reviewer's approval was based on a security posture that no longer holds.

config/llmisvcconfig/config-llm-tokenizer.yaml
```diff
@@ -0,0 +1,74 @@
...
+              terminationMessagePath: /dev/termination-log
+              terminationMessagePolicy: FallbackToLogsOnError
+              volumeMounts:
+                - mountPath: /tmp
+                  name: tmp-dir
+                - mountPath: /var/run/kserve/tls
+                  name: tls-certs
+                  readOnly: true
+          volumes:
+            - name: tmp-dir
```
mwaykole, line 68: The `tls-certs` volume (line 72) is unconditionally defined as a Secret, but the `command` only uses TLS flags when `{{ if .GlobalConfig.EnableTLS }}`. If TLS is disabled and the self-signed certs secret isn't created, this will block pod startup. Should the volume + volumeMount be wrapped in the same TLS conditional, or is the secret always guaranteed to exist?

mwaykole (commented):
Re-review of changes since July 15. The TLS support, `readOnlyRootFilesystem: true`, `tokenProcessorConfig` migration, and test improvements all look solid. Two items flagged inline.

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: The `vllm/vllm-openai-cpu:v0.23.0` image doesn't include a non-root user entry in `/etc/passwd` (unlike GPU images which publish `vllm-openai-nonroo`t variants with `UID 2000`). Setting `runAsNonRoot: true` + `runAsUser: 1000` caused a `KeyError: 'getpwuid(): uid not found: 1000' `crash at startup. `runAsNonRoot: false` is required until upstream publishes a non-root CPU image. r`eadOnlyRootFilesystem: true` is still kept for defense-in-depth. @bartoszmajsak tagging for visibility.

config/llmisvcconfig/config-llm-tokenizer.yaml
vivekk16: The self-signed cert secret is always reconciled unconditionally by the workload reconciler (before the tokenizer is created), regardless of the TLS setting. This follows the same pattern as all other templates (scheduler, main vLLM, decode, prefill). The volume mount is harmless when TLS is disabled -- the files are mounted but unused.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// hasTokenProducerPlugin checks if the scheduler config contains the token-producer
+// plugin, which requires a standalone tokenizer deployment to serve tokenization
+// requests over HTTP (vLLM render endpoint).
+func hasTokenProducerPlugin(spec v1alpha2.LLMInferenceServiceSpec) bool {
+	if spec.Router == nil || spec.Router.Scheduler == nil || spec.Router.Scheduler.Config == nil || spec.Router.Scheduler.Config.Inline == nil {
+		return false
+	}
+	return bytes.Contains(spec.Router.Scheduler.Config.Inline.Raw, []byte(tokenProducerPlugin))
```
bartoszmajsak: Could we align this with how plugin detection works elsewhere in the controller? We already have [`hasPluginType`](https://github.com/kserve/kserve/blob/85065ab905/pkg/controller/v1alpha2/llmisvc/scheduler_latency_predictor.go#L47-L65) which unmarshals and does exact `type` field matching - [`hasLatencyProducerInSpec`](https://github.com/kserve/kserve/blob/85065ab905/pkg/controller/v1alpha2/llmisvc/scheduler_latency_predictor.go#L36-L45) is already using it.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,279 @@
...
+					Port:       tokenizerServicePort,
+					TargetPort: intstr.FromString(tokenizerPortName),
+					Protocol:   corev1.ProtocolTCP,
+				},
+			},
+		},
+	}
+}
+
+func (r *LLMISVCReconciler) propagateTokenizerDeploymentStatus(ctx context.Context, expected *appsv1.Deployment, llmSvc *v1alpha2.LLMInferenceService) error {
```
bartoszmajsak: [`propagateSchedulerDeploymentStatus`](https://github.com/kserve/kserve/blob/85065ab905/pkg/controller/v1alpha2/llmisvc/scheduler.go#L810) already accepts  `ready`/`notReady` callbacks, so it could be reused here:
 
```go
return r.propagateSchedulerDeploymentStatus(ctx, expected, llmSvc.MarkTokenizerReady, llmSvc.MarkTokenizerNotReady)
```

Maybe rename to `propagateDeploymentStatus` to avoid confusing name.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,279 @@
...
+func (r *LLMISVCReconciler) reconcileTokenizerDeployment(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) error {
+	expected, err := r.expectedTokenizerDeployment(ctx, llmSvc)
+	if err != nil {
+		return fmt.Errorf("failed to build expected tokenizer deployment: %w", err)
+	}
+
+	if shouldDeleteTokenizer(llmSvc) {
+		llmSvc.MarkTokenizerUnset()
+		return Delete(ctx, r, llmSvc, expected)
+	}
```
bartoszmajsak, line 143: Should we align with the ForceStop handling in [`reconcileSchedulerDeployment`](https://github.com/kserve/kserve/blob/85065ab905e44febc1c5c96bc51be23ee9609ec6/pkg/controller/v1alpha2/llmisvc/scheduler.go#L174-L181)? 

The scheduler distinguishes `MarkNotReady("Stopped", ...)` from `MarkUnset()`, so users see a clear reason across all components. Right now the tokenizer condition just disappears on stop, which could be confusing when someone is reading the conditions list.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return Reconcile(ctx, r, llmSvc, &corev1.Service{}, expected, semanticServiceIsEqual)
+}
+
+func (r *LLMISVCReconciler) expectedTokenizerDeployment(ctx context.Context, llmSvc *v1alpha2.LLMInferenceService) (*appsv1.Deployment, error) {
+	labels := TokenizerLabels(llmSvc)
+	replicas := int32(1)
+
+	d := &appsv1.Deployment{
```
bartoszmajsak, line 165: The scheduler deployment carries `app.kubernetes.io/version` on its pod template - if it's helpful for this deployment too to gate version-specific migrations we could follow-up.

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -72,6 +75,34 @@ func (r *LLMISVCReconciler) reconcileScheduler(ctx context.Context, llmSvc *v1al
...
 		return err
 	}
+
+	// Emit a deprecation warning when the legacy monolithic precise-prefix-cache-scorer
+	// plugin is detected. It will be automatically decomposed into the 3-plugin pipeline
+	// (token-producer + precise-prefix-cache-producer + prefix-cache-scorer) during
+	// version-gated migration. Users should migrate their inline config to use the
+	// decomposed plugins directly.
+	if hasPrecisePrefixCachePlugin(llmSvc.Spec) {
+		log.FromContext(ctx).Info("DEPRECATION: precise-prefix-cache-scorer plugin detected in config.inline. " +
```
bartoszmajsak: This fires on every rencocile - maybe we are good with event only. Seems to be sufficient signal for operators to take action?

pkg/apis/serving/v1alpha2/llm_inference_service_lifecycle.go
bartoszmajsak, line 350: We are adding `TokenizerReady` to the sub-conditions list and we only seem to handle `IsFalse()`.

That would mean Unknown shouldn't block readiness - is it what we want?

bartoszmajsak (commented):
Few minor comments

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: - Removed `propagateTokenizerDeploymentStatus` from `tokenizer.go` and reused the scheduler's version with `MarkTokenizerReady`/`MarkTokenizerNotReady` callbacks, as suggested.

- Renamed `propagateSchedulerDeploymentStatus` → `propagateComponentDeploymentStatus` since it's now shared between scheduler and tokenizer.

- Also renamed the existing `propagateDeploymentStatus` in `workload_single_node.go` → `propagateWorkloadDeploymentStatus` so the naming is consistent across the board and avoids confusion:

1.      propagateComponentDeploymentStatus` — scheduler, tokenizer
2.     propagateWorkloadDeploymentStatus` — main vLLM, prefill

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Done.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Yes. I've added the ForceStop handling to Tokenizer deployment as well.

pkg/apis/serving/v1alpha2/llm_inference_service_lifecycle.go
vivekk16: `TokenizerReady` follows the same pattern as all other sub-conditions in both `DetermineRouterReadiness` and `DetermineWorkloadReadiness`. Only `IsFalse()` blocks, nil is skipped.

`Unknown` can't occur for `TokenizerReady` because they are not registered as dependents in the `LivingConditionSet` (only WorkloadReady and RouterReady are), so `InitializeConditions` never touches them. They only transition between `nil` (cleared/unset), `True` (via `MarkTrue`), and `False` (via `MarkFalse`). No code path calls `MarkUnknown` on them.

pkg/controller/v1alpha2/llmisvc/tokenizer.go
vivekk16: Makes sense.The tokenizer has no version-gated migrations today (it's a single config shape), so there's nothing to gate on yet. Happy to add it when we introduce the first tokenizer migration (hopefully shouldn't have one soon xD).

pkg/controller/v1alpha2/llmisvc/scheduler.go
vivekk16: Agreed, logging on every reconcile is noisy here. 

Downgraded the log to V(1) so it only appears at elevated verbosity. The event remains the primary operator-facing signal.

bartoszmajsak (approved):

bartoszmajsak:
/lgtm

Thanks for this work!

PR #5714, opened 2026-06-24
fix(agent): contain object storage downloads

jeffspahr:

This remediates ADA-KUBEFL-03 from the Ada Logics Kubeflow security audit for KServe's object-storage downloaders.

GCS, S3, and Azure object names are remote-controlled path input. The downloaders previously combined those names with the shared model directory using `filepath.Join`, then created or replaced the resulting file. Parent-directory components such as `../../outside.txt` could escape the intended model location. A lexical containment check alone was insufficient because a model-server container sharing the volume could place a symlink in the destination path and redirect a later write outside the volume. Normalizing backslashes also caused distinct object keys such as `a\b` and `a/b` to collide on Linux.

The downloaders now create files through a common helper backed by `os.Root`. Object names must be valid slash-separated relative paths, parent traversal is rejected, and every directory and file operation stays rooted under the trusted model directory even when a path contains an escaping symlink. Literal backslashes remain part of Linux object names instead of being rewritten as separators. The same protected write path is used by GCS, S3, and Azure, including the S3 variant explicitly called out by the audit.

Related tracking issue: https://github.com/kubeflow/community/issues/992#issuecomment-4785057657

Related to kubeflow/community#992 (ADA-KUBEFL-03). The linked issue tracks multiple projects and findings, so this PR must not close it.

- [x] `go test -vet=off -p=1 ./pkg/agent/storage`
- [x] `go vet ./pkg/agent/storage`
- [x] `git diff --check`
- [x] Regression coverage for GCS, S3, and Azure traversal attempts
- [x] Regression coverage for symlink escape and literal backslash preservation

This completes the known KServe code remediation for ADA-KUBEFL-03 across the object-storage downloaders identified during review. After merge, the remaining work is tracking only: link the merged PR from the original community issue as remediation evidence. ADA-KUBEFL-02 is handled separately by #5715.

**Checklist**:

- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)? Not applicable; this changes internal path handling without changing the user-facing storage API.

**Release note**:

```release-note
Prevent GCS, S3, and Azure model object names and symlinks from escaping the configured model download directory.
```

### DIFF
--- a/pkg/agent/storage/azure.go
+++ b/pkg/agent/storage/azure.go
@@ -18,8 +18,8 @@ package storage
 
 import (
 	"context"
+	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
@@ -58,16 +58,11 @@ func (a AzureProvider) DownloadModel(modelDir string, modelName string, storageU
 			return err
 		}
 		for _, _blob := range resp.Segment.BlobItems {
-			fileName := filepath.Join(modelDir, modelName, *_blob.Name)
-			log.Info("Downloading blob %s", "fileName", fileName)
-			file, err := os.Create(filepath.Clean(fileName))
+			file, fileName, err := createLocalModelFile(modelDir, modelName, *_blob.Name)
 			if err != nil {
 				return err
 			}
-			defer func(destFile *os.File) {
-				err = destFile.Close()
-				log.Error(err, "Error closing file")
-			}(file)
+			log.Info("Downloading blob %s", "fileName", fileName)
 
 			_, err = a.Client.DownloadFile(ctx, bucket, prefix, file,
 				&azblob.DownloadFileOptions{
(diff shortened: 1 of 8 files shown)


jeffspahr:
This security audit remediation remains relevant. The branch has been refreshed onto current master, the storage regression tests and vet pass, and the refreshed DCO, style, license, and security checks are green. Marking it to remain open for maintainer review.

spolti:
This is nice, Thanks!
/approve
/lgtm

bartoszmajsak:
@jeffspahr there are some related `make precommit` linter finds worth addressing.

jeffspahr:
Thanks @bartoszmajsak!  Addressing them now.

jeffspahr:
Addressed the precommit findings and the Go Test regression in 49e6e3401. The test files now use restrictive fixture permissions and narrowly scoped `gosec` suppressions for paths rooted in `t.TempDir()`. GCS whole-bucket downloads again support an empty model name, exact-object URIs retain a valid basename, and directory marker objects are skipped; traversal containment remains enforced in each case. The branch is also rebased onto current `master`. Focused storage and agent tests, `go vet`, and diff checks pass locally.

jeffspahr:
The refreshed fork workflow runs are currently awaiting maintainer approval before their jobs can start. DCO and the external Snyk checks are green, and the existing approval/lgtm state is retained.

jeffspahr:
@bartoszmajsak I think this just needs a fresh approval from you.

PR #5715, opened 2026-06-24
fix(agent): prevent HTTP storage SSRF

jeffspahr:

This remediates ADA-KUBEFL-02 from the Ada Logics Kubeflow security audit for KServe's HTTP(S) storage provider.

The provider previously passed a user-controlled model URL directly to `http.Client`. That allowed requests to loopback, private, link-local, metadata, reserved, and other non-public destinations, including through redirects. An initial validation followed by a normal hostname dial would still be vulnerable to DNS rebinding because the connection could resolve the hostname again and use a different address.

Default HTTP(S) providers now use a restricted transport that validates the initial URL and every redirect, rejects any hostname if one of its DNS answers is unsafe, and validates the concrete socket address through `net.Dialer.ControlContext` immediately before connection. Using the standard dialer preserves its shared timeout and IPv4/IPv6 address racing while preventing the address actually used by the socket from bypassing validation.

The destination policy now covers the missed IPv4 special-purpose range and fails closed for IPv6 addresses outside `2000::/3`, with explicit rejection for special allocations inside that space. A 30-second response-header timeout prevents a reachable public endpoint from holding the downloader indefinitely before sending headers; model body transfers remain unbounded so large downloads are not cut off.

Related tracking issue: https://github.com/kubeflow/community/issues/992#issuecomment-4785057657

Related to kubeflow/community#992 (ADA-KUBEFL-02). The linked issue tracks multiple projects and findings, so this PR must not close it.

- [x] `go test -vet=off -p=1 ./pkg/agent/storage`
- [x] `go vet ./pkg/agent/storage`
- [x] `git diff --check`
- [x] Regression coverage for unsafe direct destinations, redirects, mixed DNS answers, socket-address validation, zoned IPv6 addresses, and omitted/reserved ranges
- [x] Regression coverage for allowed public IPv4 and IPv6 destinations

Environment proxy routing is intentionally disabled for this client. Go's standard HTTP proxy path delegates target DNS resolution to the proxy, so KServe cannot validate or pin the destination address with the same guarantee. Clusters that require an HTTP proxy for all egress will need an explicitly trusted transport rather than the default provider client.

This PR is separate from ADA-KUBEFL-03 / #5714. It preserves both existing `http://` and `https://` storage schemes and does not impose a deployment-specific hostname allowlist.

**Checklist**:

- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)? Not applicable; this changes the default provider's security policy without adding a new API.

**Release note**:

```release-note
HTTP(S) model downloads now reject non-public destinations and redirects, validate the connected socket address, and time out stalled response headers. Environment proxy routing is disabled for the default HTTP(S) storage client.
```

### DIFF
--- a/pkg/agent/storage/http_client.go
+++ b/pkg/agent/storage/http_client.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2026 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"net/netip"
+	"net/url"
+	"time"
+)
+
+const maxHTTPSRedirects = 10
+
+var blockedHTTPDestinationPrefixes = []netip.Prefix{
+	netip.MustParsePrefix("0.0.0.0/8"),
+	netip.MustParsePrefix("10.0.0.0/8"),
+	netip.MustParsePrefix("100.64.0.0/10"),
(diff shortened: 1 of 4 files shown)


jeffspahr:
This security audit remediation remains relevant. The branch has been refreshed onto current master, the storage regression tests and vet pass, and the refreshed DCO, style, license, and security checks are green. Marking it to remain open for maintainer review.

PR #5716, opened 2026-06-24
fix(llmisvc): disable secure serving on EPP when TLS is off

bartoszmajsak:

The llm-d-router EPP binary is [secure by design](https://github.com/llm-d/llm-d-router/blob/47020924/pkg/epp/server/options.go#L140) - it defaults `--secure-serving` to `true`. The scheduler template only passed `--secure-serving=true` when `EnableTLS` was set, leaving the flag unset when TLS was disabled. Since the binary default kicks in, the EPP served TLS unconditionally - even when `enableLLMInferenceServiceTLS` is false.

This breaks gRPC connectivity from gateways that connect over plaintext (e.g. Istio on kind without mesh mTLS), producing `upstream connect error or disconnect/reset before headers` in the gateway proxy logs with no surface-level indication of the cause.

Explicitly passes `--secure-serving` with the actual `EnableTLS` value so the EPP correctly serves plain gRPC when TLS is not configured.

One-line template change. The rest is the test expectation update.

**Checklist**:

```release-note
NONE
```


bartoszmajsak:
Holding this PR - the fix breaks EG+AIEG.

## Details

The AI Gateway extension server hardcodes TLS on the ext_proc cluster connection to the EPP [`buildExtProcClusterForInferencePoolEndpointPicker`](https://github.com/envoyproxy/ai-gateway/blob/v1.0.0/internal/extensionserver/inferencepool.go#L206-L210) - applies `UpstreamTlsContext` unconditionally). There's no annotation or config to disable it.

This creates a coupling: the EPP **must** serve TLS when running behind EG+AIEG, regardless of global TLS config. 

The current template behavior (omitting `--secure-serving`, which defaults to `true` in [llm-d-router](https://github.com/llm-d/llm-d-router/blob/47020924/pkg/epp/server/options.go#L140)) works because both sides agree on TLS.

With this fix applied (`--secure-serving=false`), the EPP serves plaintext but the gateway proxy still connects via TLS - connection reset.

## Compatibility issue

| Setup | EPP serves | Gateway ext_proc connects | Result |
|-------|-----------|--------------------------|--------|
| EG+AIEG (upstream CI) | TLS (default) | TLS (AIEG hardcodes) | Works |
| Istio (no fix) | TLS (default) | Plaintext (native) | Broken |
| Istio (with fix) | Plaintext | Plaintext (native) | Works |
| EG+AIEG (with fix) | Plaintext | TLS (AIEG hardcodes) | Broken |

## Path forward

The proper fix needs coordination across repos:
1. `envoyproxy/ai-gateway` - respect an annotation (e.g. `aigateway.envoyproxy.io/ext-proc-tls: "false"`) to skip `UpstreamTlsContext` on the ext_proc cluster
2. `kserve/kserve` - set that annotation on InferencePool when `enableLLMInferenceServiceTLS` is false, and pass `--secure-serving=false` to the EPP

Until the AI Gateway side lands, this fix can only be applied in environments that don't use the AI Gateway extension server (e.g. Istio-based deployments).

PR #5722, opened 2026-06-25
feat(llmisvc): migrate WVA from VA CRD to annotation-based discovery

vivekk16:

WVA v0.8.0 deprecates the VariantAutoscaling CRD in favor of annotation-based discovery (https://github.com/llm-d/llm-d-workload-variant-autoscaler/releases/tag/v0.8.0). Instead of watching VA custom resources, WVA now discovers workloads via annotations on HPA/ScaledObject and synthesizes in-memory VAs from them.

This PR aligns KServe with that change. The controller no longer creates VA CRDs and instead annotates HPA/ScaledObject with `llm-d.ai/managed`, `llm-d.ai/model-id`, and `llm-d.ai/variant-cost` for WVA discovery.

Key changes:
- Remove VA CRD creation, RBAC, scheme registration, and WVA Go module dependency
- Add WVA discovery annotations and accelerator label to HPA/ScaledObject
- Auto-cleanup legacy VA CRs on reconcile (with log warning requiring WVA >= v0.8.0)
- Migrate WVA install from Helm to Kustomize (Helm chart deprecated in v0.8.0)
- Bump WVA_VERSION to v0.8.0

- [x] Unit tests (`scaling_test.go`) — assert WVA annotations, accelerator labels, metric selectors
- [x] Integration tests (`scaling_int_test.go`) — verify annotations on live HPA/ScaledObject, no VA CRDs
- [x] E2E HPA test (`test_llm_autoscaling_hpa_deployment`) — passed on minikube locally
- [x] E2E KEDA test (`test_llm_autoscaling_keda_deployment`) — passed on minikube locally

- WVA >= v0.8.0 is now required. Older versions will not discover annotated actuators.
- Legacy VA CRs are deleted on first reconcile; a log warning is emitted.
- The `variant_name` metric label now uses the HPA/ScaledObject name instead of the old VA CR name.

**Checklist**:

```release-note
Migrate WVA integration from VariantAutoscaling CRD to annotation-based discovery, aligning with VA deprecation in WVA v0.8.0. HPA/ScaledObject are now annotated with llm-d.ai/managed and llm-d.ai/model-id for WVA workload discovery. Legacy VA CRs are automatically cleaned up. WVA install migrated from Helm to Kustomize. ACTION REQUIRED: update Prometheus queries/dashboards — wva_desired_replicas variant_name now uses the HPA/ScaledObject name (e.g. *-kserve-hpa / *-kserve-keda) instead of the old VA CR name (*-kserve-va).

### CODE REVIEW

File: pkg/controller/v1alpha2/llmisvc/scaling.go [hunk 1]
```diff
@@ -119,32 +125,30 @@ func prefillWorkloadScalingParams(llmSvc *v1alpha2.LLMInferenceService) workload
 	}
 }
 
-// reconcileWorkloadScaling reconciles all scaling resources (VA, actuator) and propagates
-// status for a single workload (main or prefill). The workloadScalingParams struct captures
-// all per-workload differences so this method does not need to know which workload it is handling.
+// reconcileWorkloadScaling reconciles the scaling actuator and propagates status for a single
```
mwaykole, line 122: Minor ordering concern: `cleanupLegacyVA` runs before `reconcileActuator`, so there's a window where neither the VA nor the WVA annotations exist on the actuator. If `reconcileActuator` fails after this point, WVA loses track of the workload until next reconcile.

Would it be safer to reconcile the actuator (with annotations) first and delete the VA after? That way WVA always has at least one discovery source.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -452,77 +454,37 @@ func prometheusTrigger(cfg *WVAAutoscalingConfig, query string) kedav1alpha1.Sca
...
-	if val, ok := workloadLabels[acceleratorNameLabelKey]; ok && val != "" {
-		accelerator = val
+	if apimeta.IsNoMatchError(err) {
+		r.legacyVACRDAbsent.Store(true)
+		return nil
 	}
-	labels[acceleratorNameLabelKey] = accelerator
-
-	modelID := llmSvc.Spec.Model.URI.String()
-	if llmSvc.Spec.Model.Name != nil {
```
mwaykole, line 493: Since this uses unstructured delete (no typed client), the `legacyVACRDAbsent` cache is the only thing preventing an API call per reconcile per workload. Once the CRD is absent this is efficient, but during the migration window (CRD present, VAs being cleaned up), every LLMISVC reconcile makes a delete call per workload. For clusters with many LLMISVCs, that's 2× delete calls per reconcile (main + prefill). Probably fine for one release cycle, just flagging it.

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -191,15 +198,10 @@ func expectedHPA(llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.Scaling
 							Name: wvaDesiredReplicasMetricName,
```
mwaykole, line 188: The `variant_name` metric selector is now keyed by the HPA/ScaledObject name instead of the VA name. This is a user-visible change: any existing Prometheus dashboards or alerting rules filtering on `variant_name=<va-name>` will silently stop matching. Might be worth calling this out in the release note section — the current note mentions the annotation change but not the metric label rename.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Good point. I already looked at swapping to annotate-first / delete-VA-second.

I kept the current order on purpose. WVA v0.8 is dual-mode (VA CR + annotated HPA/ScaledObject). The names differ (…-kserve-va vs …-kserve-hpa / …-kserve-keda), so annotate then delete would briefly register two variants for the same workload and risk competing `wva_desired_replicas` series.

Delete then annotate avoids that overlap. The empty-discovery window is only until `reconcileActuator` succeeds (same reconcile on the happy path; retry on failure), which I think is preferable to dual registration during migration.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Updated the release notes.

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: Thanks for flagging. Yeah this should be just a couple of extra calls during migration window.

pkg/controller/v1alpha2/llmisvc/scheduler_tokenizer_test.go
```diff
@@ -109,6 +109,7 @@ func TestAttachStorageInitializer_TargetContainer(t *testing.T) {
 			initContainer := findContainer(podSpec.InitContainers, constants.StorageInitializerContainerName)
 			if initContainer == nil {
 				t.Fatal("storage-initializer init container not found")
+				return
```
bartoszmajsak: Do we need it?

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -452,77 +454,37 @@ func prometheusTrigger(cfg *WVAAutoscalingConfig, query string) kedav1alpha1.Sca
...
-// workloadLabels are the labels from the WorkloadSpec for the specific workload (decode or prefill).
-// If the workload labels contain inference.optimization/acceleratorName, that value is propagated
-// to the VA label. Otherwise the label is set to "unknown".
-func expectedVA(llmSvc *v1alpha2.LLMInferenceService, scaling *v1alpha2.ScalingSpec, scaleTargetRef autoscalingv2.CrossVersionObjectReference, vaName string, workloadLabels map[string]string) *wvav1alpha1.VariantAutoscaling {
-	labels := scalingLabels(llmSvc)
-	accelerator := "unknown"
-	if val, ok := workloadLabels[acceleratorNameLabelKey]; ok && val != "" {
-		accelerator = val
+	if apimeta.IsNoMatchError(err) {
+		r.legacyVACRDAbsent.Store(true)
```
bartoszmajsak: This is smart optimization. I wonder if we could also avoid attempts to `DELETE` VA CRs when CRD is still there but we already cleaned-up?

Maybe on the top of the reconcile we could do best-effort cleanup call to not block the rest of the reconcilation. Sth like this comes to mind at the beginning of the flow:

```golang
if !r.legacyVACleanupDone.Load() {
      if err := r.cleanupAllLegacyVAs(ctx, llmSvc); err != nil {
          log.FromContext(ctx).Error(err, "legacy VA cleanup will retry next
  reconcile")
		 // best-effort, non-blocking attempt
      }
  }
```

bartoszmajsak (commented):
Few minor comments, also maybe we should make sure we do not reference obsolete CR in any godoc like API.

bartoszmajsak:
@vivekk16 we could also solve it as runnable on startup like we do for storage migration https://github.com/kserve/kserve/blob/master/cmd/llmisvc/main.go#L345

Maybe as a follow-up

pkg/controller/v1alpha2/llmisvc/scaling.go
vivekk16: I've updated the logic to avoid extra DELETE API calls in every reconcile.

pkg/controller/v1alpha2/llmisvc/scheduler_tokenizer_test.go
vivekk16: Removed them.

vivekk16:
@bartoszmajsak

 The startup runnable pattern works well for storage migration because it's self contained. Here the VA cleanup is coupled with annotating the actuators, both need to happen together per LLMInferenceService to avoid a WVA discovery gap. 
 
 If we delete VAs at startup before the actuators are annotated, WVA has nothing to discover during that window. Reimplementing the naming and annotation logic in a startup runnable would duplicate what reconcileScaling already does. The legacyVACleanupDone flag ensures cleanup is a one time cost that doesn't repeat after the first successful reconcile.

bartoszmajsak:
> @bartoszmajsak
> 
>  The startup runnable pattern works well for storage migration because it's self contained. Here the VA cleanup is coupled with annotating the actuators, both need to happen together per LLMInferenceService to avoid a WVA discovery gap. 
>  
>  If we delete VAs at startup before the actuators are annotated, WVA has nothing to discover during that window. Reimplementing the naming and annotation logic in a startup runnable would duplicate what reconcileScaling already does. The legacyVACleanupDone flag ensures cleanup is a one time cost that doesn't repeat after the first successful reconcile.

Not sure it's reimplemeting, maybe just moving the logic to another place :)

bartoszmajsak (approved):

pkg/controller/v1alpha2/llmisvc/scaling.go
```diff
@@ -119,32 +128,26 @@ func prefillWorkloadScalingParams(llmSvc *v1alpha2.LLMInferenceService) workload
 	}
 }
 
-// reconcileWorkloadScaling reconciles all scaling resources (VA, actuator) and propagates
-// status for a single workload (main or prefill). The workloadScalingParams struct captures
-// all per-workload differences so this method does not need to know which workload it is handling.
+// reconcileWorkloadScaling reconciles the scaling actuator and propagates status for a single
```
mwaykole, line 122: `legacyVACleanupDone` is a single bool on the reconciler, but `cleanupAllLegacyVAs` is called with a specific `llmSvc` and only deletes that instance's VAs (names derived from `llmSvc.GetName()`). First LLMISVC to reconcile successfully sets the flag to `true`, and all other LLMISVCs' legacy VAs are never cleaned up.

Either track per-LLMISVC cleanup (e.g. a `sync.Map` of cleaned names), or do a single label-selector List+DeleteAll on first reconcile instead of per-instance deletes.

mwaykole (commented):
Re-review of changes since July 15. The annotation-based discovery, best-effort cleanup at reconcile start, and `legacyVACRDAbsent` optimization are clean. One issue flagged inline with the `legacyVACleanupDone` scope.

pkg/controller/v1alpha2/llmisvc/scaling.go
bartoszmajsak: Good catch, this flag is not necessary. I was thinking about List+Delete All when reviewing earlier.

@vivekk16 I think we should fix it before shipping

bartoszmajsak:
@vivekk16 I think [this claim](https://github.com/kserve/kserve/pull/5722#issuecomment-4995207202) does not hold and is based on the premise that VA deletion and annotation stamping are coupled. AFAICT the annotations are derived from the LLMISVC spec in the reconcile loop (which will be triggered on manager startup), completely independent of whether any VA ever existed.

I think having this clean-up logic as best-effort non-blocking process at manager startup is simpler. Both from the code base perspective, but also when thinking about removing it entirely in the future release cycle.

I took a stab and [drafted an alternative approach](https://gist.github.com/bartoszmajsak/20d0fd8013dbdaf6ef9ec881c2b8ff13) discussed earlier. WDYT?

For your convenience:

```sh
curl -sL https://gist.githubusercontent.com/bartoszmajsak/20d0fd8013dbdaf6ef9ec881c2b8ff13/raw/va-cleanup-startup-runnable.patch | git apply -
```

> [!NOTE]
> You might still need to run `make precommit`.

bartoszmajsak:
@vivekk16 I wonder about one thing you mentioned in the PR description

> Auto-cleanup legacy VA CRs on reconcile (with log warning requiring WVA >= v0.8.0)

What if we still have WVA not migrated to 0.8.0? Is it hard requirement for upcoming release to migrate? If not I think we should deal with it somewhat gracefully.

vivekk16:
Thanks @mwaykole for flagging the per-instance cleanup bug. Good catch.

@bartoszmajsak My earlier concern about the discovery gap was that deleting VAs at startup before the reconciler stamps annotations could leave WVA with nothing to discover. But thinking about it more, both the startup runnable and the reconcile loop are triggered concurrently, so the window is at most a few seconds and WVA's polling interval, HPA minReplicas, and KEDA cooldown all ensure the workload isn't disrupted during that brief gap.

I'll incorporate your patch into the PR.

vivekk16:
@bartoszmajsak WVA 0.8.0 is a hard requirement for this change, but in practice the upgrade is coupled this PR bumps WVA_VERSION to v0.8.0 and migrates the install to Kustomize, so upgrading KServe automatically upgrades WVA. 

Given that the user base for LLMInferenceService + WVA is still small, I think it's better to migrate now rather than carry dual-mode complexity. Aligning with the standard WVA annotation-based discovery sooner means we can adopt new WVA features as they land without being held back by legacy CRD compatibility. The longer we defer, the more users we'd eventually need to migrate.

If we want to be extra safe, I can add a startup log line making the version requirement explicit (e.g. "annotation-based WVA discovery active; requires WVA >= v0.8.0"), so anyone running an independently managed WVA gets a clear signal.

bartoszmajsak:
Thanks for comprehensive explanaition. I was not aware it's hard req, but it makes things easier :)

bartoszmajsak (approved):

Jooho (approved):

PR #5735, opened 2026-06-29
fix: honor disableHTTPRouteTimeout in LLMInferenceService router

### LINKED ISSUE #5707: LLMInferenceService: disableHTTPRouteTimeout is silently ignored — HTTPRoute still emits timeouts (rejected by GKE Gateway)
/kind bug

### Summary

PR #5313 added the `disableHTTPRouteTimeout` config flag to fix #5311 (GKE Gateway controller rejects HTTPRoutes with `spec.rules.timeouts`). The fix only touched the v1beta1 `InferenceService` ingress reconciler (`pkg/controller/.../httproute_reconciler.go`).

The v1alpha2 `LLMInferenceService` router (`pkg/controller/v1alpha2/llmisvc/`) has no equivalent code path. The flag is silently ignored for LLMISVC. The `kserve-llmisvc-resources` chart still exposes the value (`values.yaml`, README), which makes the omission confusing — operators set the flag expecting it to apply.

### What I expected

`kserve.controller.gateway.disableHTTPRouteTimeout=true` causes the generated HTTPRoute to omit `spec.rules[*].timeouts`, regardless of whether the resource is `InferenceService` or `LLMInferenceService`.

### What I observed (v0.19.0)

1. Installed `kserve-llmisvc-resources` v0.19.0 with:
   `--set kserve.controller.gateway.disableHTTPRouteTimeout=true`
2. Created an `LLMInferenceService` whose `router.gateway.refs` point at a GKE `gke-l7-rilb` Gateway.
3. The generated HTTPRoute has `spec.rules[*].timeouts: {request: 0s, backendRequest: 0s}` on every rule.
4. GKE rejects:

       Error GWCER104: HTTPRoute "<ns>/<name>" is misconfigured,
       err: timeouts are not supported.

   GKE Gateway translation is all-or-nothing, so every other HTTPRoute attached to the same gateway also fails to program. The parent Gateway status shows `Programmed: False`.

### Root cause

    $ gh api 'search/code?q=DisableHTTPRouteTimeout+repo:kserve/kserve+path:pkg/controller/v1alpha2/llmisvc'
    (empty)

The chart writes the value into `inferenceservice-config`, but only the v1beta1 reconciler reads it.

### Workaround

Override `router.route.http.spec` with a hand-written `HTTPRouteSpec` omitting `timeouts:`. Fragile — see #5706 about `router.route.http.spec` being a full-replace, which means the override has to duplicate the entire preset.

### Suggested fix

Wire `disableHTTPRouteTimeout` through `pkg/controller/v1alpha2/llmisvc/router.go` and the `kserve-config-llm-router-route` preset rendering path. The existing test fixture `fixture/gwapi_builders.go` already builds routes with timeouts and can be extended.

This is part of a broader pattern: LLMISVC route synthesis appears to assume Envoy Gateway semantics (e.g., the `timeouts` field is implemented there). First-class GKE Gateway support would benefit from a configurable route renderer.

### Environment

- KServe: v0.19.0
- GKE GatewayClass: `gke-l7-rilb` (regional internal)
- Reproducible on `gke-l7-regional-external-managed` per #5311.

nikolauspschuetz:

PR #5313 wired the `disableHTTPRouteTimeout` ingress flag into the **v1beta1** ingress reconciler only. The **v1alpha2 `LLMInferenceService`** router never read it.

The router copies the managed HTTPRoute spec from the `config-llm-router-route` preset verbatim, and that preset hardcodes `spec.rules.timeouts: {backendRequest: 0s, request: 0s}` on **every** rule. Because the flag was never consulted, generated HTTPRoutes always emitted `timeouts`, and Gateway implementations that do not support the field (e.g. GKE Gateway) rejected the entire route with `GWCER104`. Users had no escape hatch.

This threads the existing `IngressConfig.DisableHTTPRouteTimeout` flag through the llmisvc `Config` and strips `spec.rules.timeouts` from the generated HTTPRoute when the flag is set — mirroring `resolveTimeout` in the v1beta1 ingress reconciler (which omits the field entirely).

Custom routes supplied via `route.http.refs` are user-owned and left untouched; only the managed `route.http.spec` is affected.

Fixes #5707

## Test plan

- New unit test `TestExpectedHTTPRouteDisableTimeout` (`router_timeout_internal_test.go`) asserts the rendered HTTPRoute **omits** `timeouts` when `disableHTTPRouteTimeout=true` and **preserves** them when `false`. Verified failing before the fix, passing after.
- Full package suite green: `make test TEST_PKGS="./pkg/controller/v1alpha2/llmisvc/..."`.

```release-note
Fixed the `disableHTTPRouteTimeout` ingress option being silently ignored by the LLMInferenceService (v1alpha2) router. Generated HTTPRoutes now omit `spec.rules.timeouts` when the flag is set, so they are no longer rejected by Gateway implementations (e.g. GKE) that do not support the timeouts field.
```

### Checklist

- [x] Tests added/updated
- [x] Linked issue

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_loader.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader.go
@@ -94,6 +94,11 @@ type Config struct {
 	UrlScheme               string `json:"urlScheme,omitempty"`
 	EnableTLS               bool   `json:"enableTLS,omitempty"`
 
+	// DisableHTTPRouteTimeout omits the spec.rules.timeouts field from generated
+	// HTTPRoutes. Required for Gateway implementations (e.g. GKE) that reject the
+	// timeouts field. Mirrors the v1beta1 ingress reconciler's behavior.
+	DisableHTTPRouteTimeout bool `json:"disableHTTPRouteTimeout,omitempty"`
+
 	ModelBasedRoutingHeaderName string                `json:"modelBasedRoutingHeaderName,omitempty"`
 	ModelBasedRoutingMode       ModelBasedRoutingMode `json:"modelBasedRoutingMode,omitempty"`
 
@@ -170,6 +175,7 @@ func NewConfig(ingressConfig *v1beta1.IngressConfig, storageConfig *types.Storag
 		IngressGatewayName:          igwName,
 		UrlScheme:                   ingressConfig.UrlScheme,
 		EnableTLS:                   ingressConfig.EnableLLMInferenceServiceTLS,
+		DisableHTTPRouteTimeout:     ingressConfig.DisableHTTPRouteTimeout,
 		ModelBasedRoutingHeaderName: ingressConfig.ModelBasedRoutingHeaderName,
 		ModelBasedRoutingMode:       parseModelBasedRoutingMode(ingressConfig.ModelBasedRoutingMode),
 		StorageConfig:               storageConfig,
--- a/pkg/controller/v1alpha2/llmisvc/config_loader_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader_test.go
@@ -125,3 +125,48 @@ func TestLoadConfig(t *testing.T) {
 		t.Fatal("SchedulerConfig = nil, want populated config")
 	}
 }
+
+// TestLoadConfigDisableHTTPRouteTimeout closes the ConfigMap -> Config loop for the
+// disableHTTPRouteTimeout ingress flag: it asserts that the value set on the
+// inferenceservice-config ConfigMap propagates through NewConfig into Config, which the
+// router consumes to strip spec.rules.timeouts. Without this, a mis-wire of the
+// propagation would silently resurface #5707 while the router-level test still passes.
+func TestLoadConfigDisableHTTPRouteTimeout(t *testing.T) {
+	tests := []struct {
+		name    string
+		ingress string
+		want    bool
+	}{
+		{
+			name:    "flag true propagates",
+			ingress: `{"kserveIngressGateway": "kserve/kserve-ingress-gateway", "ingressGateway": "knative-serving/knative-ingress-gateway", "disableHTTPRouteTimeout": true}`,
+			want:    true,
+		},
+		{
+			name:    "flag omitted defaults to false",
+			ingress: `{"kserveIngressGateway": "kserve/kserve-ingress-gateway", "ingressGateway": "knative-serving/knative-ingress-gateway"}`,
+			want:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Start from the full fixture (valid storageInitializer/credentials blocks)
+			// and override only the ingress block so LoadConfig parses end to end.
+			configMap := fixture.InferenceServiceCfgMap(constants.KServeNamespace)
+			configMap.Data["ingress"] = tt.ingress
+			c := fake.NewClientBuilder().
+				WithScheme(clientgoscheme.Scheme).
+				WithObjects(configMap).
+				Build()
+
+			got, err := llmisvc.LoadConfig(t.Context(), c)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if got.DisableHTTPRouteTimeout != tt.want {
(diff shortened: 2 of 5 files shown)


nikolauspschuetz:
This fixes #5707 by threading the existing `disableHTTPRouteTimeout` ingress flag through the v1alpha2 LLMInferenceService router, mirroring the v1beta1 `resolveTimeout` behavior so GKE Gateway no longer rejects the route.

cc @sophieliu15 (authored the v1beta1 template this mirrors) @pierDipi @israel-hdez @bartoszmajsak for review 🙏

pkg/controller/v1alpha2/llmisvc/router_timeout_internal_test.go
```diff
@@ -0,0 +1,117 @@
...
+							},
+						},
+					},
+				},
+			}
+
+			r := &LLMISVCReconciler{}
+			cfg := &Config{DisableHTTPRouteTimeout: tt.disableHTTPRouteTimeout}
+
+			route := r.expectedHTTPRoute(context.Background(), llmSvc, cfg)
```
bartoszmajsak: nit


```suggestion
			route := r.expectedHTTPRoute(t.Context(), llmSvc, cfg)
```

pkg/controller/v1alpha2/llmisvc/router_timeout_internal_test.go
```diff
@@ -0,0 +1,117 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+)
+
+// ruleWithTimeouts builds a managed HTTPRoute rule carrying timeouts, mirroring the
+// shape produced by the config-llm-router-route preset (timeouts: 0s on every rule).
+func ruleWithTimeouts(path string) gwapiv1.HTTPRouteRule {
```
bartoszmajsak: Maybe we can use existing fixtures?

```
HTTPRouteRule(
      WithMatches(PathMatch(gwapiv1.PathMatchPathPrefix, "/v1/completions")),
      WithBackendRefs(BackendRefService("svc")),
      WithTimeouts("0s", "0s"),
  )
```

https://github.com/kserve/kserve/blob/3a87283e03c88f4ff3284fc41c07c0753212bc25/pkg/controller/v1alpha2/llmisvc/fixture/gwapi_builders.go#L331-L338

bartoszmajsak (commented):
Thanks for this improvement @nikolauspschuetz!

Two non-blocking comments.

pkg/controller/v1alpha2/llmisvc/router_timeout_internal_test.go
nikolauspschuetz: Good call, thanks @bartoszmajsak — switched to the shared fixture builders: `HTTPRouteRule(WithMatches(PathPrefixMatch("/v1/completions")), WithBackendRefs(BackendRefService("svc")), WithTimeouts("0s", "0s"))`. (Used the existing `PathPrefixMatch` helper rather than a `PathMatch(...)` one, which the fixture package does not have.) One wrinkle: the `fixture` package imports `llmisvc` for its envtest harness, so an in-package test importing it forms an import cycle. I moved the test to `package llmisvc_test` and reach `expectedHTTPRoute` through a tiny `ExpectedHTTPRouteForTest` export shim, matching the existing `*_export_test.go` pattern in this package.

pkg/controller/v1alpha2/llmisvc/router_timeout_internal_test.go
nikolauspschuetz: Applied, thanks @bartoszmajsak — now `route := r.ExpectedHTTPRouteForTest(t.Context(), llmSvc, cfg)` (using `t.Context()` instead of `context.Background()`; the call goes through the export shim noted above).

bartoszmajsak (approved):
Thanks!

nikolauspschuetz:
@bartoszmajsak apologies for the churn — I rebased onto master (was 111 behind) so the E2E runs against current main, which dropped the `lgtm`.

The workflows are sitting at `action_required`, so they need your approval to start. Once they're green, could you re-add `/lgtm` and `/approve`?

Also flagging that the stale bot is at warning 3 and may auto-close this before then.

PR #5736, opened 2026-06-29
fix(ci): reduce max_tokens for LoRA llmisvc tests

pierDipi:
100 max tokens just slows down CI runs since we're not testing performance nor we're using accelerators.


pierDipi:
unrelated tests are failing

bartoszmajsak:
we cannot rerun b/c `The logs for this run have expired and are no longer available.`

bartoszmajsak (approved):

PR #5739, opened 2026-06-29
feat(storage): add oci+fetch:// KServe-side OCI image pull (Step 3, #4083)

kliukovkin:
Step 3 of #4083 (the 4-step OCI harmonization roadmap from @rhuss). Depends on #5558 (Step 2: oci+native://); base will switch to master once #5558 merges.

This PR introduces oci+fetch:// — runtime-agnostic KServe-side OCI image pull and extraction. Where oci+native:// delegates to kubelet's ImageVolume (KEP-4639), oci+fetch:// uses an init container that pulls the image with oras-py, extracts the /models/ subtree to a shared emptyDir volume, and bypasses container-runtime dependencies (works on containerd's overlay snapshotter, older K8s, restricted environments).

Commits (4):

- Go controller wiring + imagePullSecrets→docker config mount (registers `oci+fetch://` in `SupportedStorageURIPrefixList`, dispatches through `ConfigureOciFetchToContainer` in the storage initializer injector, mounts imagePullSecrets as projected volume at `/mnt/oci-fetch-auth` signaled via `KSERVE_OCI_DOCKER_CONFIG` env var)
- Python `_download_oci` handler via oras-py (multi-arch index resolution, `/models/` subpath extraction, `REQUESTS_CA_BUNDLE` for private-registry TLS, `_login_from_docker_config` for private auth)
- E2E pytest + sample InferenceService + README
- Test fix on llmisvc native test (asserted mount on wrong container name; surfaced when local CI started installing LLMISVC for the storage marker — see commit message for context)

Local validation: storage marker 4/4 PASS against a kind cluster with KServe + LLMISVC + ImageVolume gate enabled. Real pull from `ghcr.io/kliukovkin/kserve-oci-test-fixture:v1` (public multi-arch test fixture; placeholder until kserve org publishes a canonical one).

Step 4 (arbitrary OCI artifacts incl. ModelPack — github.com/modelpack/model-spec) builds on this fetch infrastructure and follows in a separate PR — see the design dialogue in kserve/kserve#5558 (comment 4811818974) and @rhuss's original 4-step plan in #4083.

Deferred for follow-up:

- LLMInferenceService fetch wiring in `workload_storage.go` (currently silent modelcar fallback for `ociModelMode=fetch`; flagged in code)
- Multi-secret merging for imagePullSecrets (first secret only; documented limitation)
- Legacy `kubernetes.io/dockercfg` secret support (use modern `kubernetes.io/dockerconfigjson`)

cc @lizzzcai @rhuss @brettmthompson @tarilabs

### DIFF
--- a/docs/samples/storage/oci-fetch/README.md
+++ b/docs/samples/storage/oci-fetch/README.md
@@ -1,0 +1,133 @@
+# OCI fetch storage for InferenceService
+
+This sample shows how to load a model from an OCI container image using the
+`oci+fetch://` storageUri scheme (KServe issue [#4083](https://github.com/kserve/kserve/issues/4083)).
+
+With `oci+fetch://`, the KServe **storage-initializer** pulls the image's layers
+(using [oras-py](https://github.com/oras-project/oras-py)) and extracts the
+`/models/` subtree into the shared `/mnt/models` volume. This is runtime-agnostic:
+unlike `oci+native://` it does not require Kubernetes ImageVolume support, and
+unlike the default `modelcar` mode it does not run a sidecar for the pod's
+lifetime — the image is pulled once in an init container.
+
+## Prerequisites
+
+| Requirement | Notes |
+|---|---|
+| KServe | this branch or later, with `enableOciModelSupport: true` |
+| Model image layout | model files under `/models/` in the image (modelcar convention) |
+| Registry access | public images pull anonymously; private images need an `imagePullSecret` |
+
+## OCI image layout convention
+
+The image must store model files under `/models/` (the modelcar convention). The
+handler pulls the image, extracts only the `/models/` subtree into `/mnt/models`,
+and discards the rest of the container rootfs. An image without a `/models/`
+directory fails the pull with a clear error.
+
+Multi-arch images (an OCI image index / Docker manifest list, as produced by
+`docker buildx`) are resolved to the platform manifest matching the init
+container's architecture (`linux/<GOARCH>`); a tag-only target therefore pulls
+the correct single-arch model layers.
+
+## How to apply
+
+1. Replace the `storageUri` in `inference_service.yaml` with your OCI model image
+   reference. The image must contain model files under `/models/`.
+
+2. (Private registries only) create a docker-registry secret in the
+   `kserve-test` namespace and uncomment both the `Secret` block and the
+   `predictor.imagePullSecrets` field in `inference_service.yaml`:
+   ```bash
+   kubectl create secret docker-registry oci-fetch-registry-creds \
+     --docker-server=ghcr.io --docker-username=<user> \
+     --docker-password=<token> -n kserve-test
+   ```
+
+3. Apply the manifest:
+   ```bash
+   kubectl apply -f inference_service.yaml
+   ```
+
+## How to verify
+
+Check that the InferenceService is created:
+```bash
+kubectl get inferenceservice sklearn-oci-fetch -n kserve-test
+```
+
+The storage-initializer init container performs the pull. Confirm it completed
+successfully:
+```bash
+kubectl get pod -n kserve-test -l serving.kserve.io/inferenceservice=sklearn-oci-fetch \
+  -o jsonpath='{.items[0].status.initContainerStatuses[?(@.name=="storage-initializer")].state.terminated.exitCode}'
+# -> 0
+```
+
+Inspect the init container logs to see the pull:
+```bash
+kubectl logs -n kserve-test \
+  -l serving.kserve.io/inferenceservice=sklearn-oci-fetch -c storage-initializer
+```
+
+Confirm the model files landed in the shared mount:
+```bash
+kubectl exec -n kserve-test \
+  $(kubectl get pod -n kserve-test \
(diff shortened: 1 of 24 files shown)


kliukovkin:
Ready for review. Step 3 of #4083 (oci+fetch://) — 4 commits, CI green.

cc @lizzzcai @rhuss @brettmthompson @tarilabs (requested review at draft open; pinging again now that it's ready).

kliukovkin:
Pushed two commits addressing performance and private-registry support, with benchmarks.

Profiling the previous implementation on large images showed three sequential full-size passes: download the blob to a temp file, extract it, then move the extracted tree into the target volume — with peak transient disk usage of ~2x the image size. The first commit streams each layer straight from the registry response into a tarfile streaming reader and extracts directly into the destination: one pass, decompression overlapped with download, ~1x disk. As a side benefit, extraction now uses tarfile's filter="data", which also hardens against tar-slip (needs Python >= 3.11.4, satisfied by the initializer image).

Numbers (single-node kind, i4i.2xlarge with local NVMe, local registry; modelcar-layout images; median cold time-to-first-prediction): 14 GB: 118.4s -> 38.9s (-67%). 140 GB: 1082.7s -> 281.7s (-74%). Peak transient disk: ~2x image size -> ~1.0x (df-sampled). For reference, a plain s3:// download of the same artifacts on the same rig lands at 55.1s / 414.0s — so oci+fetch is no longer the slow path.

The second commit adds storageInitializer.ociInsecureRegistry (default false) wired to the initializer via KSERVE_OCI_INSECURE_REGISTRY, mirroring the CA-bundle pattern — needed for plain-HTTP internal registries; keyed off explicit config, never inferred. One more note for anyone testing this PR: scheme validation lives in the controller webhook, so testing requires rebuilding the controller from this branch too, not just the initializer.

Follow-up candidate (not in this PR): concurrent layer fetch for multi-layer images, analogous to containerd's max_concurrent_downloads.

cjohannsen-cloudera (changes requested):
## Hardcoded `r|gz` decompression breaks zstd-compressed layers

In `_download_oci`, every layer blob is opened with `tarfile.open(fileobj=resp.raw, mode="r|gz")`, unconditionally assuming gzip compression:

```python
with tarfile.open(fileobj=resp.raw, mode="r|gz") as tar:
```

OCI images commonly use zstd-compressed layers (`application/vnd.oci.image.layer.v1.tar+zstd`), which is the default in newer containerd, `docker buildx` outputs, and many hosted registries (GHCR, ECR, etc.). Opening a zstd blob with `mode="r|gz"` will raise a gzip decompression error (or in the worst case silently extract zero members). Since the error surfaces mid-stream rather than at a manifest-read stage, it results in a partially-extracted directory with no useful diagnostic.

The layer's `mediaType` field is already present in the manifest — this should be checked before opening the blob:

```python
LAYER_MEDIA_TYPE_MODES = {
    "application/vnd.oci.image.layer.v1.tar+gzip": "r|gz",
    "application/vnd.docker.image.rootfs.diff.tar.gzip": "r|gz",
    "application/vnd.oci.image.layer.v1.tar": "r|",
}

for layer in manifest.get("layers", []):
    media_type = layer.get("mediaType", "")
    mode = LAYER_MEDIA_TYPE_MODES.get(media_type)
    if mode is None:
        if media_type.endswith("+zstd"):
            raise RuntimeError(
                "Layer mediaType %s uses zstd compression, which is not supported by this handler. "
                "Rebuild the image with gzip-compressed layers (e.g. BUILDKIT_COLORS=0 docker buildx build --output type=oci,compression=gzip)."
                % media_type
            )
        # skip non-tar layer types (attestation blobs, config, etc.)
        continue
    digest = layer["digest"]
    with client.get_blob(target, digest, stream=True) as resp:
        resp.raise_for_status()
        with tarfile.open(fileobj=resp.raw, mode=mode) as tar:
            ...
```

(Native zstd support in `tarfile` requires Python 3.14+; for now, a clear `RuntimeError` with a remediation hint is far better than a silent or cryptic failure.)

This is worth blocking on: the test fixture appears to be gzip-compressed, so the issue won't surface in the current E2E suite but will hit users in the field whose images were built with any modern toolchain defaulting to zstd.

kliukovkin:
Applied verbatim — thanks. `_LAYER_MEDIA_TYPE_MODES` dispatches by `mediaType`; zstd rejected with an actionable error (stdlib `tarfile` gained zstd only in 3.14, initializer runs 3.11); non-tar layers skipped.

New unit tests in `python/storage/test/test_oci_storage.py`: `test_oci_uncompressed_tar_layer_extracts` (real tar round-trip — fails "not a gzip file" under the old hardcoded mode, so it genuinely pins the `r|` dispatch), `test_oci_zstd_layer_rejected`, `test_oci_non_tar_layer_skipped`. Existing gzip happy-path tests now carry an explicit gzip `mediaType` on the layer so they exercise the dispatch too.

Native zstd support (either `zstandard` runtime dep or 3.14 base image) is real follow-up — I'll file it separately and link here.

cjohannsen-cloudera (approved):

kliukovkin:
Filed the zstd follow-up: #6014

PR #5740, opened 2026-06-29
feat(llmisvc): add secondary filesystem tiers to KV cache offloading

andresllh:
## Summary

Extends \`KVCacheOffloadingSpec\` with a \`secondary []SecondaryTierSpec\` array, enabling a **GPU → CPU → POSIX disk** KV cache cascade via vLLM's \`OffloadingConnector\`.

Each \`SecondaryTierSpec\` entry holds a \`fileSystem\` backend — one of three Kubernetes volume primitives:

| Backend | Volume kind | Use case |
|---|---|---|
| \`emptyDir\` | \`emptyDir\` with \`sizeLimit\` | Node-local ephemeral, no StorageClass required |
| \`pvc\` | Generic ephemeral (\`ephemeral\`) | Controller-managed PVC per pod, pod-lifetime, safe for multi-replica/multi-node |
| \`ref\` | \`persistentVolumeClaim\` | User-managed PVC; use RWX to share cache across replicas |

The \`secondary\` array shape is designed to accommodate object-store tiers in a future follow-up without an API break.

## Changes

- **API** (\`v1alpha2\`): \`SecondaryTierSpec\`, \`FileSystemTierSpec\`, \`EmptyDirTierSpec\`, \`PVCTierSpec\`, \`PVCRefTierSpec\` added to \`KVCacheOffloadingSpec.Secondary\`; deepcopy regenerated
- **Config rendering** (\`config_merge.go\`): \`kvTransferConfig\` template func now builds \`secondary_tiers[]\` in the vLLM \`--kv-transfer-config\` JSON
- **Volume injection** (\`workload_kvcache.go\`): new file wires volumes/mounts for all three backends into single-node and multi-node pod specs (leader + workers, decode + prefill)
- **Validation** (\`llm_inference_service_validation.go\`): enforces mutual-exclusivity of backends and requires \`cpu\` when \`secondary\` is set
- **CRDs/charts**: regenerated via \`make precommit\`
- **Sample YAML**: \`docs/samples/llmisvc/kv-cache-offloading/\` with all three backend options shown

## Test plan

- [ ] Unit tests for \`kvTransferConfig\` secondary tier rendering pass (\`config_merge_test.go\`)
- [ ] Unit tests for volume injection pass for all three backends (\`workload_kvcache_test.go\`)
- [ ] Validation unit tests pass (\`llm_inference_service_validation_test.go\`)
- [ ] \`go build ./...\` succeeds
- [ ] \`make precommit\` succeeds (CRDs/charts regenerated cleanly)
- [ ] Deploy with \`emptyDir\` secondary tier and verify vLLM receives correct \`--kv-transfer-config\` with \`secondary_tiers\`


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,78 @@ type KVCacheOffloadingSpec struct {
...
+	// +optional
+	FileSystem *FileSystemTierSpec `json:"fileSystem,omitempty"`
+}
+
+// FileSystemTierSpec configures a POSIX disk secondary tier.
+// Exactly one of EmptyDir, PVC, or Ref must be set.
+type FileSystemTierSpec struct {
+	// MountPath is the path inside the container where the volume is mounted and
+	// the path passed to vLLM as secondary_tiers[i].path. Defaults to
+	// /mnt/kv-cache-<index> when not specified.
```
zdtsw: the index is depends on how user config filesystem in order, either make it clear, or just take as the comments here which is quite confusing

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,78 @@ type KVCacheOffloadingSpec struct {
...
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
+
+	// EmptyDir uses a node-local ephemeral volume. No StorageClass required.
+	// Each pod gets its own independent disk cache.
+	// +optional
+	EmptyDir *EmptyDirTierSpec `json:"emptyDir,omitempty"`
+
+	// PVC causes the controller to create a generic ephemeral PVC per pod
+	// (one PVC per pod, pod-lifetime). Safe for multi-replica and multi-node.
```
zdtsw: i thought to use PVC is to keep warm cache even after pod restart

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -645,6 +645,27 @@ func ReplaceVariables(llmSvc *v1alpha2.LLMInferenceService, llmSvcCfg *v1alpha2.
...
+				for i, s := range kv.Secondary {
+					if s.FileSystem == nil {
+						continue
+					}
+					mountPath := s.FileSystem.MountPath
+					if mountPath == "" {
+						mountPath = fmt.Sprintf("/mnt/kv-cache-%d", i)
+					}
+					entry := map[string]any{
+						"spec_name": "DiskOffloadingSpec",
```
zdtsw: do we still use `spec_name:DiskOffloadingSpec`?

pkg/controller/v1alpha2/llmisvc/workload_kvcache.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// attachFileSystemKVCacheTier adds a single filesystem-backed KV cache volume to podSpec.
+func attachFileSystemKVCacheTier(podSpec *corev1.PodSpec, fs *v1alpha2.FileSystemTierSpec, volumeName, mountPath, containerName string) {
+	var volumeSource corev1.VolumeSource
+	var subPath string
+
+	switch {
+	case fs.EmptyDir != nil:
+		sizeLimit := fs.EmptyDir.Size.DeepCopy()
```
zdtsw, line 49:
```suggestion
		sizeLimit := fs.EmptyDir.Size.DeepCopy()
		ephemeralStorageRequest = &sizeLimit
```

then later in loop, set something like resources.requests.ephemeral-storage

to prevent user set a high emptydir.size which the node does not have it and enter into CrashLoopBackOff

docs/samples/llmisvc/kv-cache-offloading/llm-inference-service-kv-cache-offloading.yaml
```diff
@@ -0,0 +1,48 @@
...
+            memory: 32Gi
+            nvidia.com/gpu: "1"
+  workload:
+    kvCacheOffloading:
+      # Primary tier: CPU RAM (GPU overflows here first)
+      cpu: "10Gi"
+      evictionPolicy: lru
+      # Secondary tiers: disk (GPU -> CPU -> disk cascade)
+      secondary:
+      # Option A: node-local ephemeral disk, no StorageClass required
```
zdtsw: i think Option A B C might be misleading, as it actually support mutliple/mixed usage with all 3 cases.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,78 @@ type KVCacheOffloadingSpec struct {
...
+	// Size is the maximum storage capacity for this tier (maps to emptyDir.sizeLimit).
+	Size resource.Quantity `json:"size"`
+}
+
+// PVCTierSpec configures a controller-managed generic ephemeral PVC as a KV cache tier.
+// One PVC is created per pod automatically; PVCs are deleted when the pod is deleted.
+type PVCTierSpec struct {
+	// AccessModes for the ephemeral PVC, e.g. [ReadWriteOnce].
+	// +optional
+	AccessModes []corev1.PersistentVolumeAccessMode `json:"accessModes,omitempty"`
```
zdtsw: if this is a ephemeral volumes, isnt RWO the only mode make sense? why user might wanna config to RWX

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
```diff
@@ -754,6 +755,58 @@ func (l *LLMInferenceServiceValidator) validateManagedDRAAnnotations(llmSvc *LLM
...
+		if set == 0 {
+			allErrs = append(allErrs, field.Required(fsp,
+				"exactly one of emptyDir, pvc, or ref must be set"))
+		} else if set > 1 {
+			allErrs = append(allErrs, field.Invalid(fsp, fs,
+				"exactly one of emptyDir, pvc, or ref must be set; multiple are set"))
+		}
+		if fs.Ref != nil && fs.Ref.Name == "" {
+			allErrs = append(allErrs, field.Required(fsp.Child("ref", "name"), "name is required"))
+		}
```
zdtsw, line 823: why not use a switch here?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Done — updated the comment to say `/mnt/kv-cache-N where N is the zero-based position of this entry in the secondary array`.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Good point — clarified in the doc: `pvc` creates an ephemeral PVC that is deleted when the pod terminates (not for warm cache across restarts). Updated the comment to explicitly say "use `ref` for a cache that survives restarts".

pkg/controller/v1alpha2/llmisvc/workload_kvcache.go
andresllh: Done — implemented the ephemeral-storage resource request. When `emptyDir` is used, the controller now adds an `ephemeral-storage` request equal to the emptyDir size to the main container so the scheduler accounts for local disk capacity.

docs/samples/llmisvc/kv-cache-offloading/llm-inference-service-kv-cache-offloading.yaml
andresllh: Done — restructured the sample so all three backends are shown as independent additive entries (not mutually exclusive options). The active emptyDir entry is uncommented; pvc and ref are commented out as additional tiers that can be mixed freely.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Agreed — removed `accessModes` from `PVCTierSpec` entirely. The controller now hardcodes `ReadWriteOnce` since that's the only access mode that makes sense for per-pod ephemeral PVCs.

pkg/apis/serving/v1alpha2/llm_inference_service_validation.go
andresllh: Done — refactored to use a `switch` on the count of set backends.

pkg/controller/v1alpha2/llmisvc/config_merge.go
andresllh: Update: we verified against the vLLM OffloadingConnector docs and already corrected this. The top-level `spec_name: "TieringOffloadingSpec"` was already correct. The secondary tier entries don't use `spec_name` at all — they use `type: "fs"` and `root_dir` per the vLLM FS tier API. The invented `spec_name: "DiskOffloadingSpec"` and `disk_bytes_to_use` fields have been removed.
Source: https://github.com/vllm-project/vllm/blob/main/docs/features/kv_offloading_usage.md

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,83 @@ type KVCacheOffloadingSpec struct {
...
+	// Size is the maximum storage capacity for this tier (maps to emptyDir.sizeLimit).
+	// The controller also requests this amount as ephemeral-storage on the main container.
+	Size resource.Quantity `json:"size"`
+}
+
+// PVCTierSpec configures a controller-managed generic ephemeral PVC as a KV cache tier.
+// One PVC is created per pod automatically with ReadWriteOnce access mode; PVCs are
+// deleted when the pod is deleted.
+type PVCTierSpec struct {
+	// StorageClassName selects the StorageClass for the ephemeral PVC.
```
zdtsw: if we wanna advtise cephfs, should we set default to ocs-storagecluster-cephfs

andresllh: We intentionally left `storageClassName` without a default to keep the upstream API portable — `ocs-storagecluster-cephfs` is ODF/OpenShift-specific and would not exist on vanilla Kubernetes clusters. Users on ODF can set `storageClassName: ocs-storagecluster-cephfs` explicitly. That said, the `pvc` backend is per-pod ephemeral RWO, so the more typical target here is a fast local storage class (e.g. NVMe-backed). For a CephFS-backed shared cache across replicas, the `ref` backend with a pre-existing RWX PVC is the better fit.

pkg/controller/v1alpha2/llmisvc/workload_kvcache_test.go
```diff
@@ -0,0 +1,212 @@
...
+			name:          "multiple tiers get indexed names and default mountPaths",
+			containerName: "main",
+			secondary: []v1alpha2.SecondaryTierSpec{
+				{FileSystem: &v1alpha2.FileSystemTierSpec{
+					EmptyDir: &v1alpha2.EmptyDirTierSpec{Size: resource.MustParse("100Gi")},
+				}},
+				{FileSystem: &v1alpha2.FileSystemTierSpec{
+					EmptyDir: &v1alpha2.EmptyDirTierSpec{Size: resource.MustParse("200Gi")},
+				}},
+			},
```
zdtsw, line 150: can we make this mutlipe test with  a mix of  index 0 "ref"  index 1 "pvc"

docs/samples/llmisvc/kv-cache-offloading/llm-inference-service-kv-cache-offloading.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: serving.kserve.io/v1alpha1
```
zdtsw, line 1: can we also add a short README.md in this folder
main idea is to help user make good call which mode they should use.
1. i have a ceph and i run on ocp, => prefer ref
2. i have single replica and i dont wanna set csi => emptydir is the first to start with
etc

zdtsw (commented):
in general lgtm, with 2 comments

pkg/controller/v1alpha2/llmisvc/workload_kvcache_test.go
andresllh: Done — added a test case mixing `ref` at index 0 (pre-existing shared PVC with subPath) and `pvc` at index 1 (ephemeral with custom StorageClass). Verifies both volume sources and mounts are correctly indexed.

docs/samples/llmisvc/kv-cache-offloading/llm-inference-service-kv-cache-offloading.yaml
andresllh: Done — added a README.md to this folder with a decision guide covering the three main cases:
1. Ceph/ODF on OpenShift → `ref` with a pre-existing RWX CephFS PVC (shared across replicas)
2. Single replica, no persistence needed → `emptyDir` (zero provisioning overhead, node-local)
3. Dedicated StorageClass (e.g. local NVMe) without managing PVCs → `pvc` (controller-managed ephemeral, pod-lifetime)

Also includes a mixing example showing ref + emptyDir as two consecutive tiers.

zdtsw (commented):
/LGTM

thanks

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,83 @@ type KVCacheOffloadingSpec struct {
...
+	// survives restarts, use Ref with a pre-existing persistent PVC instead.
+	// +optional
+	PVC *PVCTierSpec `json:"pvc,omitempty"`
+
+	// Ref mounts a pre-existing user-managed PVC as the disk cache. The PVC must
+	// already exist. For a cache shared across replicas, provision an RWX PVC;
+	// for per-pod persistent cache, provision a separate PVC per pod and reference
+	// it by name.
+	// +optional
+	Ref *PVCRefTierSpec `json:"ref,omitempty"`
```
pierDipi: should we have this to be consistent with other types?
```
pvc: // either spec or ref 
  spec:
  ref:
```

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,83 @@ type KVCacheOffloadingSpec struct {
...
+type EmptyDirTierSpec struct {
+	// Size is the maximum storage capacity for this tier (maps to emptyDir.sizeLimit).
+	// The controller also requests this amount as ephemeral-storage on the main container.
+	Size resource.Quantity `json:"size"`
+}
+
+// PVCTierSpec configures a controller-managed generic ephemeral PVC as a KV cache tier.
+// One PVC is created per pod automatically with ReadWriteOnce access mode; PVCs are
+// deleted when the pod is deleted.
+type PVCTierSpec struct {
```
pierDipi: should we simply use the core type to support all kinds of PVCs?

pierDipi (commented):
can we add e2e tests for both emptyDir and pvcs?

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -190,6 +190,83 @@ type KVCacheOffloadingSpec struct {
...
+
+// FileSystemTierSpec configures a POSIX disk secondary tier.
+// Exactly one of EmptyDir, PVC, or Ref must be set.
+type FileSystemTierSpec struct {
+	// MountPath is the path inside the container where the volume is mounted and
+	// the path passed to vLLM as secondary_tiers[i].path. When omitted, defaults
+	// to /mnt/kv-cache-N where N is the zero-based position of this entry in the
+	// secondary array.
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
```
pierDipi: can we avoid exposing this for now? I'm thinking there could potentially conflicting mountPath(s) on different tiers if we manage that automatically we ensure that doesn't happen

pierDipi: I see mountPath as an implementation detail that doesn't need to be expose (at least I'm not aware of a use case for changing that for now)

andresllh: Done — `MountPath` has been removed from `FileSystemTierSpec`. Mount paths are now always auto-assigned as `/mnt/kv-cache-N` (where N is the zero-based index into the secondary array), so path conflicts between tiers can't occur.

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Done — restructured to a single `pvc` field with `spec` and `ref` sub-fields:
- `pvc.spec` — provisions a per-pod ephemeral PVC (controller-managed, deleted with the pod)
- `pvc.ref` — references a pre-existing PVC by name (persists across pod restarts)

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
andresllh: Done — replaced the custom struct with `corev1.PersistentVolumeClaimSpec`. This gives users access to all standard PVC parameters (storage class, access modes, volume modes, data sources, etc.) without us having to enumerate them.

andresllh:
Done — added `controller_int_kvcache_test.go` with three integration tests using envtest:
- `emptyDir` tier: verifies an `EmptyDir` volume with `SizeLimit` is injected
- `pvc.spec` tier: verifies an `Ephemeral` volume (with the full `PersistentVolumeClaimSpec`) is injected
- `pvc.ref` tier: verifies a `PersistentVolumeClaim` volume (with `ClaimName` and `SubPath`) is injected

Each test creates a full `LLMInferenceService` and asserts the volumes and mounts on the resulting Deployment.

Jooho (approved):

PR #5741, opened 2026-06-29
fix(e2e): unique model names to prevent collisions

jlost:
## Summary

Parallel pytest workers create LLMInferenceServices that share the same gateway. When two services use the same model name (e.g. `facebook/opt-125m`), the gateway's header-based routing rules (`X-Gateway-Model-Name`) can match the wrong backend, causing flaky assertion failures.

### Approach: per-namespace isolation for model-routing tests

Instead of mangling model names with hash suffixes, model-routing tests now get their own Kubernetes namespace. This solves the collision naturally:

- **Path-based routing** (`/{namespace}/{name}/v1/completions`) is unique because each test has a different namespace
- **Header-based routing** (`X-Gateway-Model-Name: publishers/{namespace}/models/{model}`) is unique for the same reason
- **No model name changes** needed -- the original model names stay intact

Non-model-routing tests continue using the shared `kserve-ci-e2e-test` namespace. They were never affected by the collision because their path-based routes already included unique service names.

### Why `response_assertion_factory` is still needed

Per-namespace isolation makes model-routing tests use a dynamically-generated namespace (e.g. `e2e-llmisvc-abc12345`). Some assertions need to reference this namespace -- for example, the LoRA `/v1/models` test checks that the response contains model names like `publishers/{namespace}/models/lora-adapter-1`.

These assertions are Python callables built at parametrize time (module load), before the namespace exists. The factory pattern defers construction to fixture setup time, when both the model name and namespace are known:

```python
# At module load: define what to assert, not the concrete values
response_assertion_factory=lambda m, ns: assert_models_contains(
    m,
    f"publishers/{ns}/models/{m}",
    "lora-adapter-1",
    f"publishers/{ns}/models/lora-adapter-1",
)

# At fixture time: factory is called with resolved (model_name, namespace)
tc.response_assertion = tc.response_assertion_factory(tc.model_name, namespace)
```

Tests that don't need namespace in their assertions still use `response_assertion` directly. The factory is only used by the subset of model-routing tests that reference the `publishers/{namespace}/models/...` format.

### Changes

- **`fixtures.py`**: Namespace lifecycle helpers (`create_test_namespace`, `delete_test_namespace`, `_copy_secret`), `_substitute_namespace` for config dict fixup, per-namespace creation in `test_case` fixture when `url_getter` is set
- **`test_llm_inference_service.py`**: `namespace` field on TestCase, `response_assertion_factory` takes `(model_name, namespace)`, LoRA test uses factory
- **CI scripts**: Label `kserve.io/e2e-test=true` on default namespace, label-based diagnostics in `status-check.sh`

## Test plan
- [ ] CI e2e-llm-inference-service passes with parallel workers
- [ ] Model-routing tests create/cleanup per-test namespaces
- [ ] LoRA and model-routing tests produce correct assertions with dynamic namespace
- [ ] Non-model-routing tests unaffected (same shared namespace)


test/e2e/llmisvc/fixtures.py
```diff
@@ -1367,23 +1368,41 @@ def _setup_test_case_service(kserve_client, tc, test_node_name, peer_index=None)
...
+        unique_model_name = f"{config_model_name}-{model_name_suffix}"
+    else:
+        unique_model_name = config_model_name
     if tc.model_name == "default/model":
-        tc.model_name = _get_model_name_from_configs(tc.base_refs)
+        tc.model_name = unique_model_name
+
+    if tc.url_getter is not None and tc.extra_headers is None:
+        tc.extra_headers = {
+            "X-Gateway-Model-Name": f"publishers/{KSERVE_TEST_NAMESPACE}/models/{tc.model_name}",
```
bartoszmajsak: Do we always want to send it? This way we are not covering other rules that allow model access. https://github.com/kserve/kserve/blob/master/config/llmisvcconfig/config-llm-router-route.yaml

pierDipi: I agree, we probably don't

jlost:
@pierDipi @bartoszmajsak I reworked it to use separate namespaces, so that the X-Gateway-Model-Name header is no longer globally applied.

test/e2e/llmisvc/fixtures.py
jlost: Reworked it to use separate namespaces, so that the X-Gateway-Model-Name header is no longer globally applied.

test/e2e/llmisvc/fixtures.py
```diff
@@ -1429,36 +1439,57 @@ def _setup_test_case_service(kserve_client, tc, test_node_name, peer_index=None)
...
+    elif KSERVE_TEST_NAMESPACE in tc.model_name:
+        tc.model_name = tc.model_name.replace(KSERVE_TEST_NAMESPACE, namespace)
+
+    if tc.response_assertion_factory is not None:
+        tc.response_assertion = tc.response_assertion_factory(tc.model_name, namespace)
+
+    if tc.url_getter is not None and tc.extra_headers is None:
+        tc.extra_headers = {
+            "X-Gateway-Model-Name": f"publishers/{namespace}/models/{tc.model_name}",
+        }
```
bartoszmajsak: Why are we doing implicit header injection?

There are other ways to access the model which we should cover in tests too https://github.com/kserve/kserve/blob/master/config/llmisvcconfig/config-llm-router-route.yaml

jlost:
superseded by #5832

PR #5323, opened 2026-03-31
fix(build): pin setup-envtest

bartoszmajsak:
**What this PR does / why we need it**:

setup-envtest@latest now resolves to a pseudo-version from the main branch (v0.0.0-20260331131016-598e330bda55) that bumped the minimum Go requirement to 1.26.0. Any build running go1.25.x with GOTOOLCHAIN=local fails immediately when fetching the tool.

Pins ENVTEST_VERSION to release-0.19, matching the controller-runtime replace target (v0.19.7) in go.mod, which is known to work within the current go1.25.x toolchain constraint.

**Release note**:
```release-note
NONE
```

PR #5744, opened 2026-06-30
deps: upgrade Go to 1.26, KEDA, otel-operator controller-runtime 

rajatvig:

Upgrades KEDA and the OpenTelemetry Operator (and the dependencies they pull in), removing the previous `replace` workaround that pinned k8s and controller-runtime to older versions for KEDA compatibility. All four upgraded modules require `go 1.26.0`, so this PR also moves the build to Go 1.26.

**Dependency upgrades**
- `github.com/kedacore/keda/v2` v2.18.0 → v2.20.1
- `github.com/open-telemetry/opentelemetry-operator` v0.113.0 → `github.com/open-telemetry/opentelemetry-operator/apis` v0.153.0 (the operator split `apis/` into a separate module)
- `sigs.k8s.io/controller-runtime` v0.22.5 → v0.24.1
- `k8s.io/*` v0.35.3 → v0.36.0 (required by KEDA v2.20.1 via MVS)
- Dropped the `replace` blocks that pinned k8s/controller-runtime to the older KEDA-compatible versions

**Go 1.26 toolchain**
- `go.mod` / `.golangci.yml` move to 1.26; the CI build images bump `golang:1.25` → `golang:1.26` (main `Dockerfile`, `tools/tf2openapi`, `docs/apis`). The `setup-go` jobs already read `go-version` from `go.mod`; the `golang` images ship go1.25.x with `GOTOOLCHAIN=local` and were failing `go.mod requires go >= 1.26.0`.
- `setup-envtest` pin `release-0.19` → `release-0.24` (matches controller-runtime v0.24.x). `release-0.19` cannot provide the k8s 1.36 envtest control-plane binaries that `k8s.io/api` v0.36 requires; the quick-install scripts that embed the pin are regenerated.

**Source changes required by the upgrades**
- Migrated all `admission.Validator[T]` / `admission.Defaulter[T]` implementations to the controller-runtime v0.23+ generic webhook API (concrete types instead of `runtime.Object`), dropping the now-redundant `utils.Convert` calls and the invalid-object-type tests the compiler now enforces.
- Added `corev1.PodSchedulingGroup` (`schedulingGroup`) to the kserve `PodSpec` to match the k8s v0.36.0 core `PodSpec` struct layout (the `corev1.PodSpec` cast in the custom predictor/transformer/explainer needs field parity).

**Migrated off APIs deprecated by controller-runtime v0.24.1 / Go 1.26** (rather than suppressing the new staticcheck/vet diagnostics)
- `scheme.Builder` → apimachinery `runtime.SchemeBuilder` with a consolidated `addKnownTypes` per package, matching the existing `pkg/apis/gie/v1alpha2pool/register.go`. This also drops the controller-runtime dependency from the api packages and dedupes the previously double-registered `InferenceGraph` type.
- `mgr.GetEventRecorderFor` → a `record.Broadcaster` recorder wired to the API server via `StartRecordingToSink` in the two controller suite tests, mirroring `cmd/manager/main.go` (the manager-provided recorder persisted events to the API server, so the sink is needed for the event-assertion tests).
- `webhook.CustomDefaulter` assertion dropped from the shared, test-only `LLMInferenceServiceDefaulter` (it dispatches across both API versions via a type switch and is never registered with the webhook framework).
- `EventRecorder.Eventf` calls pass the value as an argument to a `%s`/`%v` format string to satisfy the Go 1.26 printf vet check.

**Regenerated artifacts** (`make precommit`)
- CRDs / openapi / python models pick up the new `schedulingGroup` field and the v0.36 `PodCertificateProjection` `userAnnotations` field propagated through pod volumes.
- `defaulter-gen` v0.36.0 no longer emits an empty `RegisterDefaults` stub, so the no-op `zz_generated.defaults.go` files for `v1alpha1`/`v1alpha2` are removed (nothing references `RegisterDefaults`).
- Regenerated clients/informers, helm/quick-install manifests, and quick-install scripts.
- `generate-versions-from-gomod.py` points the otel-operator module at the new `/apis` path so `sync-deps` resolves its version after the module split.

Fixes #

- [x] `make precommit` passes cleanly with no residual file changes.
- [x] envtest suites pass on Go 1.26 with the k8s 1.36 control-plane binaries:
  - `go test ./pkg/controller/v1alpha1/inferencegraph/...` — 20/20 specs pass
  - `go test ./pkg/controller/v1beta1/inferenceservice/...` (controller suite + reconciler subpackages) pass

- This raises the project's minimum Go to **1.26**. It is not optional: KEDA v2.20.1, k8s v0.36.0, controller-runtime v0.24.1, and otel-operator `apis` v0.153.0 all declare `go 1.26.0`. The `golang` image bump and the `setup-envtest` `release-0.24` pin follow from that — note this intentionally lifts the `release-0.19` pin added in #5323 that kept builds on go1.25.x with `GOTOOLCHAIN=local`.

**Checklist**:

**Release note**:
```release-note
Upgrade KEDA to v2.20.1, opentelemetry-operator to apis/v0.153.0, controller-runtime to v0.24.1, and k8s.io/* to v0.36.0. This raises the minimum Go version to 1.26.
```


go.mod
```diff
@@ -19,39 +19,39 @@ require (
...
-	k8s.io/apiextensions-apiserver v0.35.3
-	k8s.io/apimachinery v0.35.3
-	k8s.io/client-go v0.35.3
-	k8s.io/code-generator v0.35.3
+	k8s.io/api v0.36.0
+	k8s.io/apiextensions-apiserver v0.36.0
+	k8s.io/apimachinery v0.36.0
+	k8s.io/client-go v0.36.0
+	k8s.io/code-generator v0.36.0
 	k8s.io/component-helpers v0.35.3
```
cjohannsen-cloudera: k8s.io/component-helpers should be set to the same version as k8s.io/api to avoid minor version mismatch errors.

rajatvig: Updated

rajatvig:
@cjohannsen-cloudera Could you review again? The Go tests failures seems to be flaky but unsure how to retrigger them

bartoszmajsak:
@rajatvig I think it would be helpful to land it soon. Can you check why `make precommit` is failing? I also feel like making split for Go+ctrl-runtime/KEDA/otel as dedicated PRs would help reviewing and landing them.

pkg/apis/serving/v1alpha1/inference_graph.go
```diff
@@ -358,7 +358,3 @@ type InferenceGraphList struct {
 	// +listType=set
 	Items []InferenceGraph `json:"items"`
 }
-
-func init() {
-	SchemeBuilder.Register(&InferenceGraph{}, &InferenceGraphList{})
-}
```
bartoszmajsak: Are these necessary for the bump?

rajatvig:
@bartoszmajsak I'll merge head to my branch again and verify on precommit. The Keda upgrade cannot be split as the go.mod dependencies then get mixed as the older Keda uses older controller runtimes.

rajatvig:
@bartoszmajsak I merged current main in and fixed the precommit issues. It was due to a mismatched Python version on my local shell. 

Am unsure why some of the docker builds were cancelled.

rajatvig:
@bartoszmajsak @cjohannsen-cloudera When you have time, could you please review the changes?

PR #5745, opened 2026-06-30
docs(llmsivc): add sample for llmisvc to run latency-predictor

zdtsw:
Add docs sample to enable latency predictor feature in llmisvc
follow up PR https://github.com/kserve/kserve/pull/5678

cc @RishabhSaini @pierDipi

### DIFF
--- a/docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
+++ b/docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
@@ -1,0 +1,48 @@
+apiVersion: serving.kserve.io/v1alpha1
+kind: LLMInferenceService
+metadata:
+  name: qwen2-7b-latency-predictor
+spec:
+  model:
+    uri: hf://Qwen/Qwen2.5-7B-Instruct
+    name: Qwen/Qwen2.5-7B-Instruct
+  replicas: 3
+  router:
+    scheduler:
+      config:
+        inline:
+          apiVersion: inference.networking.x-k8s.io/v1alpha1
+          kind: EndpointPickerConfig
+          plugins:
+            - type: predicted-latency-producer
+            - type: latency-scorer
+            - type: weighted-random-picker
+          schedulingProfiles:
+            - name: default
+              plugins:
+                - pluginRef: predicted-latency-producer
+                - pluginRef: latency-scorer
+                - pluginRef: weighted-random-picker
+    route: { }
+    gateway: { }
+  template:
+    containers:
+      - name: main
+        resources:
+          limits:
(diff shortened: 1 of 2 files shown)


docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: qwen2-7b-latency-predictor
+spec:
+  model:
+    uri: hf://Qwen/Qwen2.5-7B-Instruct
+    name: Qwen/Qwen2.5-7B-Instruct
+  replicas: 3
+  router:
+    scheduler:
+      config:
+        inline:
```
RishabhSaini: since the inline config is a map
```
inline: |
      kind: EndpointPickerConfig
      plugins:
```

docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
```diff
@@ -0,0 +1,48 @@
...
+          apiVersion: inference.networking.x-k8s.io/v1alpha1
```
RishabhSaini: alpha1 vs alpha2?

docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
```diff
@@ -0,0 +1,48 @@
...
+          kind: EndpointPickerConfig
+          plugins:
+            - type: predicted-latency-producer
```
RishabhSaini, line 20: worth taking example from: https://github.com/llm-d/llm-d/blob/main/guides/predicted-latency-routing/router/predicted-latency.values.yaml

docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
zdtsw: related change https://github.com/kserve/kserve/pull/5774

docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
zdtsw: updated with full plugins from guide

docs/samples/llmisvc/latency-predictor/llm-inference-service-latency-predictor.yaml
zdtsw: updated

pierDipi (approved):

PR #5749, opened 2026-07-01
feat(e2e): add autogluon pytest marker to all autogluon tests

pierDipi:
Register an `autogluon` marker in both unit and e2e pytest configs and apply it to every autogluon test, enabling selective test runs via `pytest -m autogluon`.


bartoszmajsak (approved):
That's helpful! Thanks!

Jooho (approved):
/lgtm

PR #5571, opened 2026-05-22
feat(deps): upgrade GIE to v1.5.0 with local v1alpha2 InferencePool shim

vivekk16:
## Summary

Closes #5570

- Upgrade `sigs.k8s.io/gateway-api-inference-extension` from v1.3.1 to v1.5.0.
- Pin `k8s.io/*` to v0.34.x and `controller-runtime` to v0.22.5 to avoid incompatibility with KEDA v2.18's controller-runtime v0.19.x requirement (which causes informer sync crashes with k8s.io v0.35).
- Introduce a local v1alpha2 InferencePool shim (`pkg/apis/gie/v1alpha2pool/`) with conversion helpers to/from upstream v1 types, preserving backward compatibility for the LLMISvc controller.
- Fix deprecated `result.Requeue` usage (staticcheck SA1019) introduced by controller-runtime v0.22.
- Fix `hack/update-codegen.sh` to respect `replace` directives.
- Regenerate CRDs, OpenAPI specs, Python SDK docs, and update `kserve-deps.env`.

## Changes

- `go.mod` / `go.sum`: bump GIE to v1.5.0, add replace directives pinning `k8s.io/*` to v0.34.x and `controller-runtime` to v0.22.5
- `kserve-deps.env`: `GIE_VERSION=v1.5.0`
- `pkg/apis/gie/v1alpha2pool/`: new local shim (types, conversion, deepcopy, registration)
- `pkg/apis/serving/v1alpha2/`, `pkg/controller/v1alpha2/llmisvc/`: update imports to use local shim
- `pkg/apis/serving/v1beta1/podspec.go`: remove v0.35-only `WorkloadRef` field
- `pkg/client/`: restore generated informer/clientset code to v0.34 patterns
- `pkg/controller/v1beta1/inferenceservice/`: replace deprecated `result.Requeue` with `RequeueAfter`
- `hack/update-codegen.sh`: use `go list -m` with replace-aware template
- `config/llmisvc/gateway-inference-extension.yaml`, `test/crds/gateway-inference-extension.yaml`: regenerated CRDs
- `hack/setup/`: update InferencePool CRD name from `x-k8s.io` to `k8s.io` (GA domain)

```release-note
Upgrade gateway-api-inference-extension to v1.5.0, pin k8s.io/* to v0.34.x and controller-runtime to v0.22.5 to avoid KEDA v2.18 incompatibility, add v1alpha2 InferencePool shim for backward compatibility.

PR #5751, opened 2026-07-01
fix(llmisvc): add v1alpha2 InferencePool CRD to quick-install

### LINKED ISSUE #5750: llmisvc quick-install script missing v1alpha2 InferencePool CRD
/kind bug

1. Installed KServe LLMInferenceService on a cluster using hack/setup/quick-install/llmisvc-full-install-with-manifests.sh
2. The script installs upstream GIE v1.5.0 CRDs, which only include inferencepools.inference.networking.k8s.io (v1 GA)
3. Created an LLMInferenceService CR with spec.router.scheduler
4. Controller fails reconciliation with:
```
failed to reconcile networking: failed to reconcile scheduler: failed to get v1alpha2pool.InferencePool llm-d-tool-calling/nemotron-ultra-pd-inference-pool: no matches for kind "InferencePool" in version "inference.networking.x-k8s.io/v1alpha2"
```

The controller should either:
- Gracefully skip v1alpha2 InferencePool reconciliation when the CRD is not present (matching the pattern already used in `SetupWithManager()`), or
- The quick-install script should install the v1alpha2 InferencePool CRD (which is bundled in
`install/v0.19.0/kserve_kubeflow.yaml` but not applied by the script)

N/A. This is an LLMInferenceService.

- The `Delete()` function in lifecycle_crud.go:59-100 already handles missing CRDs gracefully via `meta.IsNoMatchError(err)`, but `Reconcile()` in the same file does not.
- Workaround: manually extract the v1alpha2 InferencePool CRD from `install/v0.19.0/kserve_kubeflow.yaml` and apply it.

**Environment:**

- Istio Version: 
- Knative Version:
- KServe Version: latest main
- Kubeflow version:
- Cloud Environment:[k8s_istio/istio_dex/gcp_basic_auth/gcp_iap/aws/aws_cognito/ibm] CoreWeave
- Minikube/Kind version:
- Kubernetes version: (use `kubectl version`): v1.36.1
- OS (e.g. from `/etc/os-release`): Ubuntu 24.04.3 LTS

tessapham:
v1alpha2 InferencePool CRD is not installed, leading to controller hard-fail.

The `llmisvc-full-install-with-manifests.sh` quick-install script does not install the `inferencepools.inference.networking.x-k8s.io` (v1alpha2) CRD. The LLMInferenceService controller unconditionally reconciles both v1 and v1alpha2 InferencePools in `reconcileSchedulerInferencePool()`, so when only the upstream GIE v1 CRDs are present, reconciliation hard-fails with:

```
no matches for kind "InferencePool" in version "inference.networking.x-k8s.io/v1alpha2"
```

This PR:
1. Adds the v1alpha2 InferencePool CRD to `config/crd/full/gie/` and includes it in the install script generator's CRD dirs, so the regenerated quick-install script embeds and installs it
2. Wraps `reconcileV1InferencePool()` and `reconcileV1Alpha2InferencePool()` with `utils.IsCrdAvailable()` checks, matching the existing pattern in `SetupWithManager()` (controller.go:395-403)

Fixes #5750.

- [x] Regenerated install script embeds the v1alpha2 InferencePool CRD (verified via `grep inferencepools.inference.networking.x-k8s.io`)
- [x] Controller change follows established `IsCrdAvailable()` pattern already used for watches in `SetupWithManager()`
- [x] Pre-existing controller unit test failure (`BeforeSuite` setup issue) confirmed unrelated to this change

The v1alpha2 InferencePool CRD was added to `config/llmisvc/gateway-inference-extension.yaml` in PR #5571 (GIE v1.5.0 upgrade). The install script generator's `filter_out_crds()` strips all CRDs from the kustomize overlay output, and the CRD was not added to the CRD source directories – so it was silently dropped from the generated install script.

**Checklist**:

**Release note**:

```
Fix LLMInferenceService controller crash when v1alpha2 InferencePool CRD is not installed; add CRD to quick-install script.
```


tessapham:
cc @vivekk16 (this is related to the v1alpha2 InferencePool shim) @pierDipi

henrichter-sap:
We're trying to adopt KServe and found the bundling of the GAIE CRDs very surprising (rationale in [#5543](https://github.com/kserve/kserve/issues/5543#issuecomment-4875485608)). If the controller can now handle a broader version range, can't we remove them entirely and keep them as a user-required prerequisite only?

tessapham:
@henrichter-sap Good point – the `IsCrdAvailable()` checks added here do make it safe to run without the bundled CRDs. This PR is scoped to fixing the hard crash (#5750), but it seems like #5544 already made the GIE CRDs optional.

cc @Jooho @bartoszmajsak

pkg/controller/v1alpha2/llmisvc/scheduler.go
```diff
@@ -188,12 +188,16 @@ func (r *LLMISVCReconciler) reconcileSchedulerDeployment(ctx context.Context, ll
...
+			return err
+		}
 	}
 
-	if err := r.reconcileV1Alpha2InferencePool(ctx, llmSvc, shouldDelete); err != nil {
-		return err
+	if ok, err := utils.IsCrdAvailable(r.Config, igwapiv1alpha2.GroupVersion.String(), "InferencePool"); ok && err == nil {
+		if err := r.reconcileV1Alpha2InferencePool(ctx, llmSvc, shouldDelete); err != nil {
+			return err
+		}
```
neelabhsinha: Is the `IsCrdAvailable` check necessary here? This runs on every reconcile, so if either CRD is absent later for any other reason, the `InferencePool` creation will be silently skipped without any visibility/action. The user's EPP pod may fail to start or misbehave because the pool it expects doesn't exist.

tessapham: Single-version installs are valid, so if user has InferencePool v1 but not v1alpha2 it should still succeed. We don't want InferencePool creation be silently skipped when no version is installed. I can add a check to harden this case, e.g., return error if no version is available at all. Wdyt?

neelabhsinha: Yes, this should be okay I think. Thanks.

Jooho:
I don't think the `IsCrdAvailable` skip pattern fits here. InferencePool is a required dependency for LLMISVC, so if the CRD is missing, LLMISVC can't function correctly. Unlike Envoy AI Gateway, where InferencePool is optional, GIE is a hard dependency for LLMISVC. I'd rather fail fast with a clear error than silently skip reconciliation.

GIE is updated very frequently, so the process of downloading the GIE manifests and placing them in the appropriate directory is already automated. Could you take a look at how it's done here?
https://github.com/kserve/kserve/blob/master/Makefile#L142-L153

pierDipi:
Related to https://github.com/kserve/kserve/pull/5710

tessapham:
@Jooho Thanks for pointing that out. I've added the GIE CRDs to the Makefile automation. The `IsCrdAvailable` checks meant to solve the same issue as in #5710 (as @pierDipi mentioned) – since it is now handled there, I've dropped these checks from the scheduler.

Makefile
```diff
@@ -151,6 +151,8 @@ manifests: controller-gen kustomize yq
 	cp config/llmisvc/gateway-inference-extension.yaml test/crds/gateway-inference-extension.yaml
 	cat test/crds/gateway-inference-extension-v1alpha2pool.yaml >> config/llmisvc/gateway-inference-extension.yaml
 	cat test/crds/gateway-inference-extension-v1alpha2pool.yaml >> test/crds/gateway-inference-extension.yaml
+	cp test/crds/gateway-inference-extension-v1alpha2pool.yaml config/crd/full/gie/inference.networking.x-k8s.io_inferencepools.yaml
+	cp test/crds/gateway-inference-extension-v1alpha2pool.yaml config/crd/minimal/gie/inference.networking.x-k8s.io_inferencepools.yaml
```
vivekk16: This copy creates a file that nothing consumes.` minimal-crdgen.sh` skips it (filename doesn't match `serving.kserve.io*`), no Helm chart target copies from it, and `config/crd/minimal/kustomization.yaml` doesn't reference `./gie`. Consider dropping this line and the `config/crd/minimal/gie/` directory to keep things clean.

henrichter-sap:
I don't think this is the best solution to the actual problem. 

Essentially this PR is just a downgrade of the gaie release from 1.5 to 1.4 to keep using the old group.

tessapham:
@henrichter-sap This PR doesn't introduce the v1alpha2 support (that was #5571), which already bumped GIE to v1.5 but also introduced v1alpha2 shim (from older version) for local backward compability. This PR just ensures the CRD that's already required by the controller actually gets installed, or else the quick-install script will fail.

henrichter-sap:
Hm. Maybe Im doing something wrong then. But the latest release (which should also include #5571) does not work with just the gaie v1.5. You can check yourself by doing this minimal setup. Here everything comes up, but due to missing/incompatible CRDs the LLMIsvc never gets ready.

```bash
$ kind create cluster
$ kubectl apply --server-side -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.0/standard-install.yaml
$ kubectl apply --server-side -f https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v1.5.0/manifests.yaml
$ helm install \
  cert-manager oci://quay.io/jetstack/charts/cert-manager \
  --version v1.20.3 \
  --namespace cert-manager \
  --create-namespace \
  --set crds.enabled=true
$ helm install kserve-llmisvc-crd \
  oci://ghcr.io/kserve/charts/kserve-llmisvc-crd \
  --namespace kserve \
  --create-namespace \
  --version v0.19.0
$ helm install kserve-llmisvc-resources \
  oci://ghcr.io/kserve/charts/kserve-llmisvc-resources \
  --namespace kserve \
  --version v0.19.0 \
  --set kserve.llmisvc.createGIECRDs=false
$ helm install kserve-runtime-configs \
  oci://ghcr.io/kserve/charts/kserve-runtime-configs \
  --namespace kserve \
  --version v0.19.0 \
  --set kserve.llmisvcConfigs.enabled=true
$ kubectl apply -f - <<'EOF'
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: kserve-ingress-gateway
  namespace: kserve
spec:
  gatewayClassName: dummy
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
EOF
$ kubectl apply -f - <<EOF
kind: LLMInferenceService
metadata:
spec:
  model:
    uri: hf://facebook/opt-125m
    name: facebook/opt-125m

  replicas: 1

  template:
    containers:
      - name: main
        image: quay.io/pierdipi/vllm-cpu:latest
        securityContext:
          runAsNonRoot: false  # Image requires root
        env:
          - name: VLLM_LOGGING_LEVEL
            value: DEBUG
        resources:
          limits:
            cpu: '1'
            memory: 10Gi
          requests:
            cpu: '100m'
            memory: 8Gi
        livenessProbe:
          initialDelaySeconds: 30
          periodSeconds: 30
          timeoutSeconds: 30
          failureThreshold: 5

  router:
    gateway: {}
    route: {}
    scheduler: {}
EOF
$ kubectl logs -n kserve deployments/llmisvc-controller-manager | grep error | tail -1
{"level":"error","ts":"2026-07-08T15:42:37Z","msg":"Reconciler error","controller":"llminferenceservice","controllerGroup":"serving.kserve.io","controllerKind":"LLMInferenceService","LLMInferenceService":{"name":"facebook-opt-125m-single","namespace":"default"},"namespace":"default","name":"facebook-opt-125m-single","reconcileID":"1c3cc316-74dc-4ffa-b762-5b7f1f0d2499","error":"failed to reconcile networking: failed to reconcile scheduler: failed to get v1alpha2.InferencePool default/facebook-opt-125m-single-inference-pool: no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\"","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.7/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.7/pkg/internal/controller/controller.go:263\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.7/pkg/internal/controller/controller.go:224"}
```

From my understanding this would've been fixed with the initial idea of handling both versions.

EDIT: Actually this is probably a different but related issue. If you prefer I can open it as a separate issue. However I think the project shouldn't ship those CRDs anyways (https://github.com/kserve/kserve/issues/5543#issuecomment-4875485608)

tessapham:
@henrichter-sap Thanks for posting the reproduction. I think your scenario (Helm with `createGIECRDs=false` + upstream GIE v1.5.0) is a valid case but would now be handled by #5710, which adds a config flag to gracefully skip v1alpha2 when the CRD isn't present. This PR fixes the _quick-install path_ specifically.

To drop the GIE CRDs completely is a different issue, which you can feel free to open for discussion.

henrichter-sap:
Ah. I haven't seen the other PR. That makes the scope of this PR much more clear. Thanks for pointing out!

Makefile
tessapham: Gotcha, removed.

tessapham: `sync_kustomize_patches()` in `minimal-crdgen.sh` still copies from `full/` to `minimal/` (it copies everything _but_ `serving.kserve.io*` and `kustomization.yaml` files, so still including the GIE CRDs) and will error out if minimal directory doesn't exist. I think I'll need to restore this directory.

vivekk16 (commented):
LGTM
cc @Jooho @spolti

config/crd/minimal/gie/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- inference.networking.x-k8s.io_inferencepools.yaml
```
bartoszmajsak, line 5: @tessapham For consistency - don't we need to install v1 this way too, or do we assume it's provided in the target env as pre-req?

cc @Jooho

tessapham: @bartoszmajsak Sorry just getting to this question now. v1 is not bundled – it's fetched directly from the upstream GIE release at install time. v1alpha2 needs to be bundled because it's a kserve-local shim that doesn't exist upstream anymore.

PR #5755, opened 2026-07-03
fix(storage): honor CA bundle for https and hf downloads

KillianGolds:

The CA bundle configured via `storageInitializer.caBundleConfigMapName` gets mounted into the storage initializer, but only the `s3://` download path actually uses it. Downloads from `https://` and `hf://` URIs use requests / huggingface_hub default trust (certifi), so models behind a private CA fail with `SSL: CERTIFICATE_VERIFY_FAILED` even when the bundle is configured correctly.

The change adds one step at the start of `Storage.download()` for remote URIs: if the CA bundle is configured, combine it with the system and certifi trust stores into a single file and export it via `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` (only when those aren't already set). The download paths themselves are unchanged; requests, huggingface_hub, google-auth, azure-core, and hf_transfer all pick these variables up on their own.

Notes on behavior:
- No CA bundle configured: no change at all.
- Configured: trust is only ever extended, never replaced, so downloads that verify today keep working. Existing `REQUESTS_CA_BUNDLE` / `SSL_CERT_FILE` values are left alone.
- If the bundle file is missing or doesn't parse, a warning is logged and the default trust store stays in effect.
- The `s3://` path is untouched. The export is skipped for `hdfs://`/`webhdfs://`, since those configure TLS explicitly (`TLS_CA` / `TLS_SKIP_VERIFY`) and `REQUESTS_CA_BUNDLE` would override their session settings.

`certifi` is added to the dependency list because it's now imported directly. It was already installed via requests, so nothing new gets installed; the other `uv.lock` changes are just that declaration propagating to packages that depend on kserve-storage.

No existing issue that I could find; hit this deploying a model from an `https://` URI behind a private CA.

- [x] Unit tests: no-op when unconfigured, bundle exported for `https://` and `hf://`, pre-set env values preserved, missing/malformed bundle falls back to default trust, hdfs excluded. Full storage suite passes.
- [x] Tested manually against a self-signed HTTPS server: fails with `CERTIFICATE_VERIFY_FAILED` on master, succeeds with this change; a publicly-signed endpoint still verifies with the bundle configured; no change when unconfigured. Also checked that hf_transfer honors `SSL_CERT_FILE` on Linux.
- [x] `make precommit` passes.

- Logs

Before (bundle mounted and env vars set):
```
storage.initializer ERROR ... HTTP error for https://model-server.example.svc/model.bin: ... SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate')
```
After (same setup):
```
storage.initializer INFO Global ca bundle file(/etc/ssl/custom-certs/cabundle.crt) combined with the default trust store at /tmp/tmpXXXX.crt
storage.initializer INFO Successfully copied https://model-server.example.svc/model.bin to /mnt/models
```

- I combined the custom CA with the default trust stores instead of replacing them (the way s3 does), because replacing would break downloads from publicly-signed endpoints on clusters that have a bundle configured. Happy to align with s3 semantics instead if you'd prefer consistency.
- A missing bundle file logs a warning rather than raising, since pods can end up with the env vars but no mount. The s3 path's existing raise is unchanged.
- hdfs and git keep their own TLS handling and are out of scope here. I didn't add an e2e test because CI doesn't have a TLS storage fixture today; if there's interest, that could be a follow-up.

**Checklist**:

**Release note**:
```release-note
Fixed the storage initializer ignoring the CA bundle configured via caBundleConfigMapName for https:// and hf:// storage URIs; it was previously only applied to s3:// downloads.
```


mwaykole (commented):
looks good  the changes makes sense

python/storage/test/test_storage.py
```diff
@@ -120,6 +121,178 @@ def test_https_uri_path(_):
...
+        "CA_BUNDLE_CONFIGMAP_NAME": "cabundle",
+        "CA_BUNDLE_VOLUME_MOUNT_POINT": str(tmp_path),
+    }
+    with mock.patch.dict(os.environ, env):
+        os.environ.pop("REQUESTS_CA_BUNDLE", None)
+        os.environ.pop("SSL_CERT_FILE", None)
+        # The hdfs session pins verify from TLS_CA / TLS_SKIP_VERIFY and
+        # REQUESTS_CA_BUNDLE would override it, so the global CA bundle
+        # must not be exported for hdfs downloads.
+        Storage.download("webhdfs://host/model", out_dir=str(tmp_path / "out"))
```
cjohannsen-cloudera: Please also explicitly test `hdfs://` to ensure compatibility.

KillianGolds: Good call, added in the latest push.

- The exclusion test is now parametrized over both `hdfs://` and `webhdfs://`, asserting the global CA bundle env vars are never exported for either scheme.
- Also verified end-to-end on a cluster with an image built from this branch: an `hdfs://` download with a pinned `TLS_CA` (cert not present in the global bundle) succeeds with `caBundleConfigMapName` configured, and a control pod that forces `REQUESTS_CA_BUNDLE` (simulating the absence of the exclusion) fails TLS verification on the first WebHDFS call, which is the override this guards against.

cjohannsen-cloudera (approved):

PR #5756, opened 2026-07-03
feat: add spell checks with codespell

### LINKED ISSUE #2493: Add spelling check for KServe docs
/kind feature

Add spelling checks for KServe docs, related to https://github.com/kserve/kserve/pull/1970

tessapham:
Add a spell-check tool to the repo.

Fixes #2493.

**Release note**:
```release-note
NONE.
```


tessapham:
Typos are fixed in #5757.

tessapham:
cc @yuzisun

Jooho:
codespell is added to .pre-commit-config.yaml, but the KServe CI `Precommit Check` workflow runs `make check` -> `make precommit` (Go/Python linters), not `pre-commit run`. So codespell only catches typos for devs who ran `pre-commit install` locally — it won't block typos in CI.

  Was the intent to enforce this in CI as well? If so, I think we should add this as part of precommit tasks.

tessapham:
@Jooho Good callout, thanks! Ideally this check should be available in both local and CI precommit checks. I've added spell-check to the CI as well.

tessapham:
CI check fails as expected due to typos – getting #5757 in will fix.

Makefile.quality.mk
```diff
@@ -118,8 +118,11 @@ sync-helm-common-helpers:
 		echo "  ✓ Copied to charts/$$chart/templates/_utils.tpl"; \
 	done
 
+spell-check: $(CODESPELL)
```
Jooho, line 122: Could you please add a comment on which file we can update for spell-check? (should be setup.cfg, right?)

setup.cfg
```diff
@@ -0,0 +1,15 @@
...
+    ./install,./hack/setup/quick-install,
+# Tokenizer vocabulary – subword tokens trigger false positives
+    ./python/custom_tokenizer/vocab.txt,
+# Sample training data
+    ./docs/samples/v1beta1/sklearn
+# Go CamelCase identifiers: WithT (gomega helper), UpToDate/NotIn (constants), allReady, caNew, VAs
+# Auto-generated Python: querys (OpenAPI Generator api_client.py), maxium (datamodel-codegen openapi.py)
+# Valid terms: re-use, keep-alives (HTTP idle timeout), ist/Worte (German in test assertions)
+# Log output abbreviation: ERRO (vllm log level)
+ignore-words-list = witht,uptodate,notin,querys,maxium,re-use,keep-alives,ist,worte,vas,allready,canew,erro,unparseable
```
spolti, line 14: are all of these valid exceptions instead typos?

tessapham: Yeah these are all false positives, they're either variable names or abbreviations.

Jooho (commented):
nit

Makefile.quality.mk
tessapham: Good point, added!

Jooho (commented):
/lgtm

spolti:
/approve

thanks @tessapham!

setup.cfg
bartoszmajsak: This is a bit all-or-nothing exclusion which might be growing over the time and we will potentially miss typos in the docs sections. I wonder if we could consider swapping to https://github.com/crate-ci/typos/ which seems code-aware and more configurable. WDYT?

tessapham: @bartoszmajsak Thanks for the suggestion! I just tried out https://github.com/crate-ci/typos/ and compared it to codespell:

| | Raw (no config) | After excluding generated files |
  |---|---|---|
  | **codespell** | 816 | 103 |
  | **typos** | 819 | 409 |

So without any exclusion config, codespell produces fewer false positives on source code. typos' code-awareness seems to be more aggressive overall even though it helps with some.

Either tool needs its own all-or-nothing exclusion config file.

codespell seems to be simpler and more widely adopted. I'm open to revisiting tool choice later if the ignore world list grows significantly, but for now the config is still pretty minimal while still covering our needs.

tessapham:
@bartoszmajsak Please lmk if there're any other comments, otherwise feel free to merge if all looks good to you!

PR #5768, opened 2026-07-06
fix(llmisvc): honor disableHTTPRouteTimeout in v1alpha2 router

### LINKED ISSUE #5707: LLMInferenceService: disableHTTPRouteTimeout is silently ignored — HTTPRoute still emits timeouts (rejected by GKE Gateway)

chethanuk:

PR #5313 added the `disableHTTPRouteTimeout` flag so that generated HTTPRoutes omit `spec.rules[*].timeouts` (GKE Gateway rejects routes that set timeouts). That fix only touched the v1beta1 `InferenceService` ingress reconciler. The v1alpha2 `LLMInferenceService` router ignored the flag: its route preset bakes `timeouts` into every rule and `expectedHTTPRoute` copied them verbatim, so the flag was silently a no-op for LLMISVC even though the chart still exposes it.

This wires the read side:

- `config_loader.go`: plumb `DisableHTTPRouteTimeout` from the shared `IngressConfig` into the llmisvc `Config` (no CRD change — the flag already exists in `IngressConfig`).
- `router.go`: in `expectedHTTPRoute`, clear each rule's `Timeouts` when the flag is set. This single choke point sits inside the `HasSpec` block, after model-based routing rules are applied and before the migration block, so it covers both the non-migration and managed-pool migration InferencePool paths. BYO routes via `Refs` are untouched.

```mermaid
flowchart TD
    A[expectedHTTPRoute] --> B{Router/Route/HasSpec?}
    B -- no --> R[return route]
    B -- yes --> C[build rules from spec]
    C --> D{DisableHTTPRouteTimeout?}
    D -- yes --> E[Timeouts = nil on every rule]
    D -- no --> F[keep timeouts]
    E --> G{managed pool<br/>Scheduler set, no Ref?}
    F --> G
    G -- no --> R
    G -- yes --> H[migration block:<br/>set BackendRef.Group only]
    H --> R
```

Fixes #5707

Table-driven unit test `TestExpectedHTTPRoute_DisableHTTPRouteTimeout` calls `expectedHTTPRoute` directly (no envtest/apiserver needed):

- flag set -> `timeouts` omitted from the rule
- flag unset -> `timeouts` retained
- flag set on the managed-pool migration path (non-nil `Scheduler`, fake client) -> `timeouts` still omitted, and the migration block is proven to run (default backendRef `Group` is set), guarding against a future regression if that block is extended to touch `Timeouts`.

- Logs

```
$ gofmt -l pkg/controller/v1alpha2/llmisvc/{config_loader,router,router_disable_timeout_test}.go   # clean
$ go vet ./pkg/controller/v1alpha2/llmisvc/                                                          # no issues
$ go test ./pkg/controller/v1alpha2/llmisvc/ -run TestExpectedHTTPRoute_DisableHTTPRouteTimeout      # PASS (3 cases)
```

Managed routes only; BYO routes via `Refs` are unaffected by design. A full envtest integration test asserting the reconcile path is a reasonable follow-up.

**Checklist**:

**Release note**:
```release-note
Fix: the v1alpha2 LLMInferenceService router now honors `disableHTTPRouteTimeout`, omitting `spec.rules[*].timeouts` from generated HTTPRoutes so gateways that reject route timeouts (e.g. GKE Gateway) accept them.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/config_loader.go
+++ b/pkg/controller/v1alpha2/llmisvc/config_loader.go
@@ -94,6 +94,11 @@ type Config struct {
 	UrlScheme               string `json:"urlScheme,omitempty"`
 	EnableTLS               bool   `json:"enableTLS,omitempty"`
 
+	// DisableHTTPRouteTimeout, when true, causes generated HTTPRoutes to omit
+	// spec.rules[*].timeouts. Some Gateway implementations (e.g. GKE Gateway)
+	// reject routes that set timeouts. Mirrors the v1beta1 ingress reconciler.
+	DisableHTTPRouteTimeout bool `json:"disableHTTPRouteTimeout,omitempty"`
+
 	ModelBasedRoutingHeaderName string                `json:"modelBasedRoutingHeaderName,omitempty"`
 	ModelBasedRoutingMode       ModelBasedRoutingMode `json:"modelBasedRoutingMode,omitempty"`
 
@@ -170,6 +175,7 @@ func NewConfig(ingressConfig *v1beta1.IngressConfig, storageConfig *types.Storag
 		IngressGatewayName:          igwName,
 		UrlScheme:                   ingressConfig.UrlScheme,
 		EnableTLS:                   ingressConfig.EnableLLMInferenceServiceTLS,
+		DisableHTTPRouteTimeout:     ingressConfig.DisableHTTPRouteTimeout,
 		ModelBasedRoutingHeaderName: ingressConfig.ModelBasedRoutingHeaderName,
 		ModelBasedRoutingMode:       parseModelBasedRoutingMode(ingressConfig.ModelBasedRoutingMode),
 		StorageConfig:               storageConfig,
--- a/pkg/controller/v1alpha2/llmisvc/router.go
+++ b/pkg/controller/v1alpha2/llmisvc/router.go
@@ -227,6 +227,15 @@ func (r *LLMISVCReconciler) expectedHTTPRoute(ctx context.Context, llmSvc *v1alp
 				cfg.ModelBasedRoutingHeaderName,
 			)
 		}
+
+		// Some Gateway implementations (e.g. GKE Gateway) reject HTTPRoutes that
+		// set spec.rules[*].timeouts. When disableHTTPRouteTimeout is configured,
+		// omit timeouts from every rule. Mirrors the v1beta1 ingress reconciler.
+		if cfg.DisableHTTPRouteTimeout {
+			for i := range httpRoute.Spec.Rules {
+				httpRoute.Spec.Rules[i].Timeouts = nil
+			}
+		}
 	}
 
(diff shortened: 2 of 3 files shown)


bartoszmajsak:
Thanks for this contribution, but it's a duplicate of #5735

PR #5771, opened 2026-07-07
refactor(e2e): group LLMISVC tests with positive markers

vivekk16:
## What this PR does

Introduces positive pytest markers for LLMISVC e2e test grouping, replacing the negative exclusion pattern (`not autoscaling and not tracing`) used in CI job marker expressions.

### New markers added
- `llmisvc_core` -- applied to all 15 core LLMISVC tests (model routing, lora, pvc, config deletion, rolling upgrade, etc.)
- `autoscaling_wva` -- replaces `autoscaling` on all 12 LLMISVC autoscaling tests, scoped to WVA-based scaling
- `llmisvc_autoscaling` -- additional grouping marker on the same 12 autoscaling tests

### CI workflow changes
- Core job: `"llminferenceservice and llmisvc_core and cluster_cpu"` (was `not autoscaling and not tracing`)
- HPA job: `"llminferenceservice and autoscaling_hpa and cluster_cpu"` (unchanged behavior)
- KEDA job: `"llminferenceservice and autoscaling_keda and cluster_cpu"` (unchanged behavior)
- Tracing job: `"llminferenceservice and tracing and cluster_cpu"` (unchanged)

## Why

The previous approach required adding `not <marker>` exclusions to the core job every time a new test category with additional infrastructure was introduced. This was error-prone and caused CI failures in midstream environments where the required infrastructure was not available. Positive markers make each job self-contained and allow midstream CI to compose test groups independently using `or` expressions (e.g., `llmisvc_core or tracing`).

## Impact
- No behavioral change to upstream CI -- same tests run in the same jobs


vivekk16:
@bartoszmajsak PTAL
Grouped the tests to avoid never ending list of negative markers

bartoszmajsak (commented):
Thanks for cleaning it up @vivekk16. A few ideas for you to consider:

### Auto markers

I think it would be good to auto-mark tests based on naming convention that we already established. This will help lazy folks like me, but also we won't need to remember about it while reviewing new tests.

The file naming convention already encodes the grouping - `test_llm_autoscaling*.py = autoscaling`, everything else = core. 

A small addition to the existing `pytest_collection_modifyitems` in conftest eliminates ~40 decorators and makes the system self-maintaining:

```python
  LLMISVC_GROUPS = {
      "test_llm_autoscaling": pytest.mark.llmisvc_autoscaling,
  }

  # in pytest_collection_modifyitems, before the existing nodeid renaming:
  for item in items:
      path = item.path.name
      matched = False
      for prefix, marker in LLMISVC_GROUPS.items():
          if path.startswith(prefix):
              item.add_marker(marker)
              matched = True
              break
      if not matched:
          item.add_marker(pytest.mark.llmisvc_core)
```

I think this is evaluated before `-m` flag is processed, so it *should* work.

New core test files auto-get `llmisvc_core`. New autoscaling files (e.g. `test_llm_autoscaling_keda.py`) auto-match via `startswith`. One dict entry to maintain instead of decorators on every test function.

### Markers clean-up (might be follow-up)

We declare markers in both `conftest.py` and `pytest.ini`. I wonder if we can't just pick one as a single source of truth. Seems a bit redundant right now.

I'd keep pytest_configure in the same conftest since that's where the hook lives - keeps the "what markers exist" next to "how they're assigned."

WDYT?

vivekk16:
Thanks for the suggestions @bartoszmajsak! Great idea, the auto-marker approach is much better for maintainability. I'll implement marker clean up in a follow up PR.

jlost:
Nice, this was much needed! 
Have you considered a module-based approach instead of the filename matching? It would be more conventional to python and pytest. The downside is more changes. Example: 
```
test/e2e/llmisvc/
  conftest.py                        # shared fixtures (test_case factory, node-id rewrite)
  core/
    conftest.py                      # auto-applies llmisvc_core
    test_llm_inference_service.py
    test_llm_lora_adapters.py
    test_pod_watch.py
    ...
  autoscaling/
    conftest.py                      # auto-applies llmisvc_autoscaling
    test_llm_autoscaling_wva.py      # WVA tests (HPA/KEDA backends inside)
  tracing/
    conftest.py                      # auto-applies tracing
    test_llm_tracing.py
```

```
# test/e2e/llmisvc/core/conftest.py
import pytest

def pytest_collection_modifyitems(config, items):
    for item in items:
        item.add_marker(pytest.mark.llmisvc_core)
```
```
# test/e2e/llmisvc/autoscaling/conftest.py
import pytest

def pytest_collection_modifyitems(config, items):
    for item in items:
        item.add_marker(pytest.mark.llmisvc_autoscaling)
```

vivekk16:
Thanks for the suggestion @jlost! The module based approach is definitely more conventional for pytest.

For the current scale (~13 test files, 3 groups), I think the filename convention keeps things simpler. It avoids moving files, adding new directories, and maintaining multiple conftest files. The grouping logic lives in one place and adding a new autoscaling variant is just naming the file correctly with no structural changes needed.

If the test suite grows significantly or groups start needing isolated fixtures/setup, the directory approach would be a natural next step. For now I'd prefer to keep it flat.

@bartoszmajsak WDYT? Do you see value in restructuring into subdirectories at this point, or is the filename convention sufficient for now?

jlost:
Fwiw, other than that, LGTM.

bartoszmajsak:
Not a Python expert, but 13 files in a flat structure look good to me (maybe not a huge fan of the number itself, but that's irrelevant ;)).

bartoszmajsak (approved):

Jooho (approved):

PR #5774, opened 2026-07-07
chore(llmisvc): migrate EndpointPickerConfig apiversion

zdtsw:

- from `inference.networking.x-k8s.io/v1alpha1 `to `llm-d.ai/v1alpha1`
- mostly for tests and docs/sample
- schedulerConfigText() change wont impact existing llmisvc with old apiversion if epp pod is running with old image (<0.9.0)  otherwise, migrate from old API to new llm-d.ai/v1alpha1

cc @pierDipi

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/scheduler_test.go
+++ b/pkg/controller/v1alpha2/llmisvc/scheduler_test.go
@@ -1404,7 +1404,7 @@ plugins:
 }
 
 func TestSchedulerTransform(t *testing.T) {
-	oldConfigYAML := `apiVersion: inference.networking.x-k8s.io/v1alpha1
+	oldConfigYAML := `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
@@ -1498,7 +1498,7 @@ func TestSchedulerTransformThreshold(t *testing.T) {
 	}{
 		{
 			name: "migrates non-zero threshold in full transform",
-			configYAML: `apiVersion: inference.networking.x-k8s.io/v1alpha1
+			configYAML: `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
@@ -1525,7 +1525,7 @@ plugins:
 		},
 		{
 			name: "migrates non-zero threshold with deciderPluginName in full transform",
-			configYAML: `apiVersion: inference.networking.x-k8s.io/v1alpha1
+			configYAML: `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
@@ -1551,7 +1551,7 @@ plugins:
 		},
 		{
 			name: "migrates threshold 0 in full transform",
-			configYAML: `apiVersion: inference.networking.x-k8s.io/v1alpha1
+			configYAML: `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
@@ -1609,7 +1609,7 @@ plugins:
 
 func TestFullMigrationPipeline(t *testing.T) {
 	// Realistic old v0.6 config with all deprecated features.
-	oldConfigYAML := `apiVersion: inference.networking.x-k8s.io/v1alpha1
+	oldConfigYAML := `apiVersion: llm-d.ai/v1alpha1
 kind: EndpointPickerConfig
 plugins:
 - type: prefill-header-handler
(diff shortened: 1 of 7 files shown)


vivekk16 (commented):
Do we need a migration in `schedulerTransform` to rewrite `apiVersion` for existing deployments? Right now `preserveSchedulerConfig()` will keep the old `inference.networking.x-k8s.io/v1alpha1` forever on previously-created deployments.

zdtsw:
> Do we need a migration in `schedulerTransform` to rewrite `apiVersion` for existing deployments? Right now `preserveSchedulerConfig()` will keep the old `inference.networking.x-k8s.io/v1alpha1` forever on previously-created deployments.

i was thinking to keep what has been in the cluster as-is
just to follow the same pattern that we do not change after upgrade.
mainly, that could trigger reconcile, right?

pierDipi:
> > Do we need a migration in `schedulerTransform` to rewrite `apiVersion` for existing deployments? Right now `preserveSchedulerConfig()` will keep the old `inference.networking.x-k8s.io/v1alpha1` forever on previously-created deployments.
> 
> i was thinking to keep what has been in the cluster as-is just to follow the same pattern that we do not change after upgrade. mainly, that could trigger reconcile, right?

I think we want this but gated on the scheduler version, if people upgrade their specific service from let's say scheduler 0.5 to 0.9+, we migrate to the new api version.

zdtsw:
> > > Do we need a migration in `schedulerTransform` to rewrite `apiVersion` for existing deployments? Right now `preserveSchedulerConfig()` will keep the old `inference.networking.x-k8s.io/v1alpha1` forever on previously-created deployments.
> > 
> > 
> > i was thinking to keep what has been in the cluster as-is just to follow the same pattern that we do not change after upgrade. mainly, that could trigger reconcile, right?
> 
> I think we want this but gated on the scheduler version, if people upgrade their specific service from let's say scheduler 0.5 to 0.9+, we migrate to the new api version.

updated

pkg/controller/v1alpha2/llmisvc/scheduler_test.go
```diff
@@ -1609,7 +1684,7 @@ plugins:
 
 func TestFullMigrationPipeline(t *testing.T) {
 	// Realistic old v0.6 config with all deprecated features.
-	oldConfigYAML := `apiVersion: inference.networking.x-k8s.io/v1alpha1
+	oldConfigYAML := `apiVersion: llm-d.ai/v1alpha1
```
mwaykole, line 1692: The comment says "Realistic old v0.6 config with all deprecated features" but the apiVersion is now already `llm-d.ai/v1alpha1`. A real v0.6 scheduler config would have had `inference.networking.x-k8s.io/v1alpha1`. Leaving the old apiVersion here would make this a true end-to-end migration test — verifying the apiVersion rewrite composes correctly with all the other transforms (plugin renames, threshold migration, etc.) in a single `schedulerTransform` call. The new `TestSchedulerTransformMigratesLLMDAPIVersion` covers the migration in isolation but not in composition with the rest of the pipeline.

test/e2e/llmisvc/fixtures.py
```diff
@@ -1856,7 +1856,7 @@ def delete_scheduler_configmap():
             raise
 
 
-INFERENCE_OBJECTIVE_GROUP = "inference.networking.x-k8s.io"
+INFERENCE_OBJECTIVE_GROUP = "llm-d.ai"
```
mwaykole, line 1859: The PR description says this migrates the `EndpointPickerConfig` apiVersion, but this also changes the `InferenceObjective` API group (`v1alpha2`, separate CRD). Does `InferenceObjective` also live under `llm-d.ai` now, or is this an unintentional side-effect? If it did migrate, might be worth calling that out in the PR description since it's a different CRD with a different version.

mwaykole (commented):
Migration function looks clean. Couple of test coverage observations.

test/e2e/llmisvc/fixtures.py
zdtsw: with https://github.com/kserve/kserve/pull/5835 already in i think this is just duplicated change

pkg/controller/v1alpha2/llmisvc/scheduler_test.go
zdtsw: i updated comments to make it clear the test is still on the migration feature but the api version is bumped to the new one, which does not impact the test itself

mwaykole (commented):
The updates look good. The version-gated `withMigrateLLMDAPIVersion` transform addresses the concern about existing deployments staying on the old apiVersion forever — cleanly gated on >= 0.9.0 and tested in isolation with the three-case `TestSchedulerTransformMigratesLLMDAPIVersion`. The `TestFullMigrationPipeline` comment update also clarifies the intent.

bartoszmajsak (approved):

bartoszmajsak:
/lgtm

Thanks!

Issue #5729, opened 2026-06-26
Feature request: Make HTTPRoute backendRef kind/group configurable for InferenceService


mgonzalezo:
/kind feature

**Describe the solution you'd like**

When KServe creates HTTPRoutes for an InferenceService, the backendRef is hardcoded to `kind: Service`. This makes it difficult for AI gateways (like [agentgateway](https://github.com/agentgateway/agentgateway)) to apply LLM-aware features such as token-based rate limiting, cost tracking, and guardrails, since those require the route to point to a gateway-specific backend type instead of a plain Service.

Allow users to configure the backendRef kind and group used when KServe auto-generates HTTPRoutes. For example, via annotation or global config:

```yaml
# Option A: Per-InferenceService annotation
metadata:
  annotations:
    serving.kserve.io/backend-kind: AgentgatewayBackend
    serving.kserve.io/backend-group: gateway.kgateway.dev

# Option B: Global KServe config
kserve.controller.gateway.backendRef:
  kind: AgentgatewayBackend
  group: gateway.kgateway.dev
```

The core ask is an extension point so the auto-generated HTTPRoute can reference a backend type other than Service when the gateway requires it.

This aligns with Gateway API's support for arbitrary backendRef kinds ([GEP-1742](https://gateway-api.sigs.k8s.io/geps/gep-1742/)) and keeps the solution generic and gateway-neutral.

**Anything else you would like to add:**

The current workaround requires users to manually create two extra resources per InferenceService ([documented here](https://agentgateway.dev/docs/kubernetes/latest/integrations/kserve/#step-5-create-a-backend)):

- An AgentgatewayBackend CRD declaring the Service as an OpenAI-compatible LLM
- A second HTTPRoute with a more specific path match to override KServe's auto-generated route

This came from a discussion in [agentgateway/agentgateway#2323](https://github.com/agentgateway/agentgateway/issues/2323) where both agentgateway maintainers agreed that the cleanest fix lives on the KServe side rather than building per-platform plugins in the gateway.

**Links to the design documents:**

N/A — looking for maintainer feedback on the right abstraction before drafting a design doc.

mgonzalezo:
cc @yuzisun @sivanantha321 @bartoszmajsak , would appreciate your thoughts on whether this direction makes sense for KServe's HTTPRoute generation.

bartoszmajsak:
[...]

PR #5776, opened 2026-07-07
fix(llmisvc): skip scheduler for custom backendRefs

bartoszmajsak:

When an LLMInferenceService overrides HTTPRoute `backendRefs` to a non-InferencePool kind (e.g. `AgentgatewayBackend`), the controller still creates the full scheduler subsystem (EPP deployment, service, InferencePool) and gates `Ready` on InferencePool acceptance. Gateways that don't implement the GIE InferencePool resource - like AgentGateway - never accept the pool, blocking `Ready` indefinitely.

This change detects when the effective (post-merge) HTTPRoute backendRefs don't target the managed InferencePool and:

- Clears `SchedulerWorkloadReady` so the scheduler doesn't block readiness
- Surfaces a `SchedulerBackendRefMismatch` condition when scheduler is configured but no backendRef references the pool, giving the user a clear signal to remove `spec.router.scheduler`
- Always calls `reconcileScheduler` so its internal guards clean up stale resources on transition (e.g. switching from InferencePool to a custom backend)

The validation runs post-merge (in `validateRouterReferences`, not the webhook) because `baseRefs` are not resolved at admission time - the backendRefs may come entirely from an `LLMInferenceServiceConfig` preset.

Ref https://github.com/kserve/kserve/issues/5729

- [x] Unit tests for `validateSchedulerBackendRefConsistency` covering: no scheduler, matching InferencePool, custom backendRef bypass, external pool ref, no route spec
- [x] Envtest integration tests: custom backendRef reaches `RouterReady` without InferencePool; default InferencePool path creates scheduler resources as before

Discovered while validating the [LLMInferenceService + AgentGateway integration guide](https://github.com/kserve/website/pull/697) on a kind cluster. AgentGateway doesn't watch InferencePool resources unless `inferenceExtension.enabled=true` is set, and even then only writes parent status for pools directly referenced by an HTTPRoute - not through an `AgentgatewayBackend` chain.

**Checklist**:
- [x] Tests added
- [x] Existing tests pass

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router.go
+++ b/pkg/controller/v1alpha2/llmisvc/router.go
@@ -89,7 +89,6 @@ func (r *LLMISVCReconciler) reconcileRouter(ctx context.Context, llmSvc *v1alpha
 		return err
 	}
 
-	// Reconcile the scheduler component that manages inference pools
 	if err := r.reconcileScheduler(ctx, llmSvc); err != nil {
 		llmSvc.MarkSchedulerWorkloadNotReady("SchedulerReconcileError", "Failed to reconcile scheduler: %v", err.Error())
 		return fmt.Errorf("failed to reconcile scheduler: %w", err)
--- a/pkg/controller/v1alpha2/llmisvc/router_validation.go
+++ b/pkg/controller/v1alpha2/llmisvc/router_validation.go
@@ -24,6 +24,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
@@ -31,7 +32,8 @@ import (
 )
 
 const (
-	RefsInvalidReason = "RefsInvalid"
+	RefsInvalidReason                 = "RefsInvalid"
+	SchedulerBackendRefMismatchReason = "SchedulerBackendRefMismatch"
 )
 
 // ValidationError represents a validation failure that should be reported via conditions
@@ -91,6 +93,12 @@ func (r *LLMISVCReconciler) validateRouterReferences(ctx context.Context, llmSvc
 			return err
 		}
 		return fmt.Errorf("managed HTTPRoute spec validation failed: %w", err)
+	}
+
+	if err := r.validateSchedulerBackendRefConsistency(llmSvc); err != nil {
+		llmSvc.MarkSchedulerWorkloadNotReady(SchedulerBackendRefMismatchReason, err.Error())
+		logger.Info("Scheduler/backendRef consistency validation failed", "error", err)
+		return nil
 	}
 
 	if llmSvc.Spec.Router != nil && llmSvc.Spec.Router.Route != nil && llmSvc.Spec.Router.Route.HTTP.HasRefs() {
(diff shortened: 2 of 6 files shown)


pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -89,7 +89,9 @@ func (r *LLMISVCReconciler) reconcileRouter(ctx context.Context, llmSvc *v1alpha
 		return err
 	}
 
-	// Reconcile the scheduler component that manages inference pools
+	if !r.httpRouteReferencesInferencePool(llmSvc) {
+		llmSvc.MarkSchedulerWorkloadUnset()
+	}
```
pierDipi: should this be within `reconcileScheduler` to encapsulate it?

pkg/controller/v1alpha2/llmisvc/router_validation.go
```diff
@@ -93,6 +95,12 @@ func (r *LLMISVCReconciler) validateRouterReferences(ctx context.Context, llmSvc
 		return fmt.Errorf("managed HTTPRoute spec validation failed: %w", err)
 	}
 
+	if err := r.validateSchedulerBackendRefConsistency(llmSvc); err != nil {
+		llmSvc.MarkSchedulerWorkloadNotReady(SchedulerBackendRefMismatchReason, err.Error())
+		logger.Info("Scheduler/backendRef consistency validation failed", "error", err)
+		return nil
+	}
```
pierDipi, line 102: should this be a warning event rather than an hard failure/stop?

pkg/controller/v1alpha2/llmisvc/router_validation.go
```diff
@@ -267,3 +275,53 @@ func (r *LLMISVCReconciler) validateManagedHTTPRouteSpec(ctx context.Context, ll
...
+// httpRouteReferencesInferencePool reports whether any backendRef in the effective
+// HTTPRoute spec targets the managed InferencePool. Returns true when no spec is
+// present (controller-generated default always targets the pool) or no scheduler is configured.
+func (r *LLMISVCReconciler) httpRouteReferencesInferencePool(llmSvc *v1alpha2.LLMInferenceService) bool {
+	if llmSvc.Spec.Router == nil || llmSvc.Spec.Router.Route == nil || !llmSvc.Spec.Router.Route.HTTP.HasSpec() {
+		return true
+	}
+	if llmSvc.Spec.Router.Scheduler == nil {
+		return true
+	}
```
pierDipi: is this case really true?

pkg/controller/v1alpha2/llmisvc/router_validation.go
pierDipi: because I'm wondering if in theory there could GW implementations that have other ways of resolving the InferencePool behind the scene and so allowing that might help in those cases

pierDipi: maybe this is more like a foundational question on the PR in general

bartoszmajsak: My thinking here was that you could have another backendRef in place and having scheduler not attached to any httproute requires user intervention, so surfacing it through the condition is the right signal.

As for the follow-up question, in case of AgentGateway InferencePool can be used under the hood - https://agentgateway.dev/docs/kubernetes/latest/inference/#ai-policies

so that would be an implementation detail and we perhaps shouldn't create our own scheduler.

pkg/controller/v1alpha2/llmisvc/router_validation.go
bartoszmajsak: I think it's misleading name and the code - it's checking either if scheduler is not used at all OR if used, it's properly referenced. I think it's too blurry and deserves cleanup.

pkg/controller/v1alpha2/llmisvc/router.go
bartoszmajsak: It could, good point.

pkg/controller/v1alpha2/llmisvc/router_validation.go
bartoszmajsak: If your point is that it might be premature to have validation like this, right now it only triggers when the user explicitly configuress `spec.router.scheduler`. In the AgentGateway chained topology the gateway manages its own endpoinnt pick er- the user wouldn't set `scheduler: {}` because KServe's EPP isn't in the data path.

So this catches one specific case: user asks for KServe's scheduler but provides backendRefs that bypass it. That's always a misconfiguration regardless of gateway implementation. If the gateway handles

An event would be even less intrusive - but then the user has to watch events instead of just checking conditions, which feels worse for discoverability. WDYT?

danehans:
> Gateways that don't implement the GIE InferencePool resource - like AgentGateway - never accept the pool, blocking Ready indefinitely.

Note that agentgateway fully supports the InferencePool resource:

- https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/conformance/reports/v1.4.0/gateway/agentgateway
- https://agentgateway.dev/docs/kubernetes/latest/inference/

An `AgentgatewayBackend` is only needed when AI policies, i.e. token-based rate limits, are required.

danehans:
The referenced issue (https://github.com/kserve/kserve/issues/5729) has been closed. Is this PR still required?

PR #5779, opened 2026-07-07
feat: improve storage-initializer/modelcare configuration

### LINKED ISSUE #5769: LLMIsvc: Improve `storage-initializer` configuration
/kind feature

Kservers provides a [`ClusterStorageContainer`](https://kserve.github.io/website/docs/next/model-serving/storage/storage-containers) to configure the `storage-initializer`. In this CRD users can pre-configure the `storage-initializer` `initContainer`. 

However when using the `LLMIsvc` this CR gets silently ignored and replaced by some `ConfigMap` content. This is unexpected for users which try to configure the container.

I was running into the issue when the initializer was OOMing and found no intuitive way of configuration.

https://github.com/kserve/kserve/blob/c14d8b6864f12a4286f6de1b561b12b5f90b806f/charts/kserve-llmisvc-resources/files/common/configmap-patch.yaml#L101-L114

I would propose to properly create the `storage-initalizer` container based on the `ClusterStorageContainer` instead of just improving the helm templating.

Id be happy to contribute this feature.

henrichter-sap:
> PR description was drafted with AI assistance.

Every new operator-facing knob on the LLMISVC storage-initializer used to require a three-layer change: Helm value → `inferenceservice-config` ConfigMap → Go parsing. Anything not already in the ConfigMap schema (envFrom, imagePullPolicy, lifecycle, extra volume mounts, custom securityContext, …) meant a KServe code change. Meanwhile the `ClusterStorageContainer` CRD,  the documented way to configure the storage-initializer, was silently ignored by the LLMISVC controller.

This PR routes the LLMISVC storage flow through `ClusterStorageContainer` end-to-end. Everything on `spec.container` is used verbatim; adding a new knob is now zero code changes.

Helm values are unchanged. The chart now also renders them into two CSCs — `default` (for `s3://`, `hf://`, `https://`, …) and `modelcar` (for `oci://`, gated on `enableModelcar: true`). Modelcar sidecars keep four code-owned fields (name, image from the URI, the `ln -sf /proc/$$/root/models` args, the mandatory shared volume mount); everything else on the CSC container flows through.

Scoped to LLMISVC. The classic v1beta1 InferenceService controller, the pod-mutating webhook, and the local-model reconciler still consume the ConfigMap-based path and are unchanged. Chart additions ship on both charts so a follow-up migration doesn't need to touch operator-facing config.

Fixes #5769

- [x] Unit tests for the CSC-verbatim modelcar helpers (`go test ./pkg/utils/...`) — asserts env / envFrom / resources / securityContext / lifecycle / imagePullPolicy flow verbatim; name / image / args are code-owned; nil-CSC nil-safety.
- [x] Unit tests for `resolveStorageContainerSpec` (auto-match, no-match, disabled, wrong workload type, 5 by-name error paths) and `extractCaBundleConfig` (CSC vs credential-builder precedence, kserve-ns vs user-ns).
- [x] All 241 existing Ginkgo integration specs still pass under envtest (`make setup-envtest && KUBEBUILDER_ASSETS=... go test ./pkg/controller/v1alpha2/llmisvc/...`).
- [x] `helm lint` clean on both charts; `helm template` renders 2 CSCs by default, 1 when `enableModelcar=false`, and omits the CA env block when both CA values are empty.
- [x] `hack/setup/scripts/generate_chart_manifests.sh` re-run is a no-op — `_common` source and per-chart mirrors stay in sync.

CA bundle env vars (`CA_BUNDLE_CONFIGMAP_NAME`, `CA_BUNDLE_VOLUME_MOUNT_POINT`) on the CSC drive `cabundleconfigmap.ReconcileForSource` — a new source-name-explicit form of the existing reconciler so llmisvc-only clusters get the cross-namespace CA-bundle copy without depending on the v1beta1 controller running. The existing `Reconcile(ctx, ns)` signature is unchanged.

**Checklist**:

**Release note**:

```release-note
LLMInferenceService: storage-initializer configuration is now sourced from a matching `ClusterStorageContainer` end-to-end. Operators can customize every field of the init container (env, envFrom, imagePullPolicy, securityContext, lifecycle, volumeMounts, …) via the CSC without KServe code changes. Modelcar (`oci://`) is gated on a `ClusterStorageContainer` matching `oci://`; the Helm chart now renders a `modelcar` CSC when `enableModelcar` is set. Classic InferenceService, the pod-mutating webhook, and localmodel-node continue to use the existing ConfigMap-based configuration.
```

### DIFF
--- a/charts/_common/common-patches/storagecontainer-patch.yaml
+++ b/charts/_common/common-patches/storagecontainer-patch.yaml
@@ -13,3 +13,16 @@ spec:
       {{- with .Values.kserve.storage.containerSecurityContext}}
       {{- toYaml . | nindent 6 }}
       {{- end }}
+    {{- $caCM := .Values.kserve.storage.caBundleConfigMapName }}
+    {{- $caPath := .Values.kserve.storage.caBundleVolumeMountPath }}
+    {{- if or $caCM $caPath }}
+    env:
+      {{- if $caCM }}
+      - name: CA_BUNDLE_CONFIGMAP_NAME
+        value: "{{ $caCM }}"
+      {{- end }}
+      {{- if $caPath }}
+      - name: CA_BUNDLE_VOLUME_MOUNT_POINT
+        value: "{{ $caPath }}"
+      {{- end }}
+    {{- end }}
--- a/charts/kserve-llmisvc-resources/files/common/storagecontainer-patch.yaml
+++ b/charts/kserve-llmisvc-resources/files/common/storagecontainer-patch.yaml
@@ -13,3 +13,16 @@ spec:
       {{- with .Values.kserve.storage.containerSecurityContext}}
       {{- toYaml . | nindent 6 }}
       {{- end }}
+    {{- $caCM := .Values.kserve.storage.caBundleConfigMapName }}
+    {{- $caPath := .Values.kserve.storage.caBundleVolumeMountPath }}
+    {{- if or $caCM $caPath }}
+    env:
+      {{- if $caCM }}
+      - name: CA_BUNDLE_CONFIGMAP_NAME
+        value: "{{ $caCM }}"
+      {{- end }}
+      {{- if $caPath }}
+      - name: CA_BUNDLE_VOLUME_MOUNT_POINT
+        value: "{{ $caPath }}"
+      {{- end }}
+    {{- end }}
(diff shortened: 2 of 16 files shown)


henrichter-sap:
Most of it is implemented by #5614 already

Issue #5773, opened 2026-07-07
Single-node LLMInferenceService silently drops spec.parallelism (tensor/pipeline) with no warning


okhs0712:
/kind bug

**What steps did you take and what happened:**

On a single-node `LLMInferenceService` (no `spec.worker`, no `spec.prefill`) I set `spec.parallelism.tensor: 2` and allocated `nvidia.com/gpu: "2"`, expecting tensor parallelism across both GPUs. The field is accepted by the API (`spec.parallelism.tensor` has `+kubebuilder:validation:Minimum=1`), but it is **silently dropped** — the rendered `vllm serve` command never receives `--tensor-parallel-size`, so vLLM runs with the default `TP=1` on a single GPU while the second GPU sits idle. The resource reports `Ready=True` with no event or warning, so there is no indication anything was ignored.

Verified on a live cluster:

1. The workload pod's container `command` renders the single-node `vllm serve` block with an empty line exactly where `--tensor-parallel-size` would be:

   ```
   eval "exec vllm serve /mnt/models \
     --served-model-name ... \
     --port 8000 \
     ${ACCESS_LOG_ARGS} \
     ${SHUTDOWN_TIMEOUT_ARGS} \
      \                          <-- empty; no --tensor-parallel-size
      \
      \
     ${VLLM_ADDITIONAL_ARGS} \
     $@"
   ```

2. `status.appliedConfigs` lists only `kserve-config-llm-router-route` and `kserve-config-llm-template` (single-node base) — none of the `*-worker-data-parallel` configs.

3. vLLM logs confirm no TP: `Model loading took 27.5 GiB memory`, weights loaded only on `rank_0`. Pod is `1/1 Running` and serving `200 OK`; the 14B model happens to fit on one GPU, so nothing surfaces the problem. A model that did not fit on one GPU would instead OOM at startup — again with no hint that the ignored `parallelism.tensor` was the cause.

[...]

PR #5782, opened 2026-07-07
fix(llmisvc): render tensor parallelism flags for single-node

bartoszmajsak:

Single-node LLMInferenceService deployments silently drop `spec.parallelism.tensor` - the CRD accepts these fields, the resource reports `Ready=True`, but the rendered `vllm serve` command never includes `--tensor-parallel-size`.

This happens because the single-node config templates (base, decode, prefill) never had these flags - only the multi-node worker templates did. Since tensor is valid on a single multi-GPU node, the single-node templates should render it too.

Related to #5773

- [x] `TestSingleNodeParallelismRendered` - verifies all three single-node templates render `--tensor-parallel-size`  when set
- [x] Negative case confirms flags are omitted when `spec.parallelism` is nil (no panic on nil pointer)
- [x] Existing `TestPresetFiles` suite passes - no regressions in multi-node template rendering

Each template uses `and` to nil-guard the `Parallelism` pointer since single-node deployments may not have parallelism set at all. The multi-node worker templates don't need this guard because they're only selected when `Worker != nil`, which implies `Parallelism` is always set.

**Checklist**:

```release-note
Renders spec.parallelism.tensor into single-node vllm serve commands, fixing silent field drop on LLMInferenceService without spec.worker.
```

### DIFF
--- a/config/llmisvcconfig/config-llm-decode-template.yaml
+++ b/config/llmisvcconfig/config-llm-decode-template.yaml
@@ -169,6 +169,7 @@ spec:
               ${ACCESS_LOG_ARGS} \
               ${SHUTDOWN_TIMEOUT_ARGS} \
               ${KV_TRANSFER_ARGS} \
+              {{- if and .Spec.Parallelism .Spec.Parallelism.Tensor }} --tensor-parallel-size {{ .Spec.Parallelism.Tensor }}{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key{{- end }} \
--- a/config/llmisvcconfig/config-llm-template.yaml
+++ b/config/llmisvcconfig/config-llm-template.yaml
@@ -169,6 +169,7 @@ spec:
               ${ACCESS_LOG_ARGS} \
               ${SHUTDOWN_TIMEOUT_ARGS} \
               ${KV_TRANSFER_ARGS} \
+              {{- if and .Spec.Parallelism .Spec.Parallelism.Tensor }} --tensor-parallel-size {{ .Spec.Parallelism.Tensor }}{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--ssl-certfile /var/run/kserve/tls/tls.crt{{- end }} \
               {{ if .GlobalConfig.EnableTLS }}--ssl-keyfile /var/run/kserve/tls/tls.key{{- end }} \
--- a/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
+++ b/charts/kserve-runtime-configs/files/llmisvcconfigs/resources.yaml
@@ -164,6 +164,7 @@ spec:
           ${ACCESS_LOG_ARGS} \
           ${SHUTDOWN_TIMEOUT_ARGS} \
           ${KV_TRANSFER_ARGS} \
+          {{- if and .Spec.Parallelism .Spec.Parallelism.Tensor }} --tensor-parallel-size {{ .Spec.Parallelism.Tensor }}{{- end }} \
           {{ if .GlobalConfig.EnableTLS }}--enable-ssl-refresh{{- end }} \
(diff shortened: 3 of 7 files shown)


config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -169,6 +169,8 @@ spec:
               ${ACCESS_LOG_ARGS} \
               ${SHUTDOWN_TIMEOUT_ARGS} \
               ${KV_TRANSFER_ARGS} \
+              {{- if and .Spec.Parallelism .Spec.Parallelism.Expert }} --enable-expert-parallel{{- end }} \
```
pierDipi: Expert should be only applicable to DataParallelism

bartoszmajsak: Good catch, I got it wrong :)

pierDipi (approved):
/lgtm

config/llmisvcconfig/config-llm-decode-template.yaml
```diff
@@ -169,6 +169,7 @@ spec:
               ${ACCESS_LOG_ARGS} \
               ${SHUTDOWN_TIMEOUT_ARGS} \
               ${KV_TRANSFER_ARGS} \
+              {{- if and .Spec.Parallelism .Spec.Parallelism.Tensor }} --tensor-parallel-size {{ .Spec.Parallelism.Tensor }}{{- end }} \
```
neelabhsinha, line 172: Hi @bartoszmajsak , technically vLLM allows `--pipeline-parallel-size` in single node deployments also. While this will mostly be slower, should be still enable it for the `LLMIsvcConfig` to be flexible for users?

config/llmisvcconfig/config-llm-template.yaml
```diff
@@ -169,6 +169,7 @@ spec:
               ${ACCESS_LOG_ARGS} \
               ${SHUTDOWN_TIMEOUT_ARGS} \
               ${KV_TRANSFER_ARGS} \
+              {{- if and .Spec.Parallelism .Spec.Parallelism.Tensor }} --tensor-parallel-size {{ .Spec.Parallelism.Tensor }}{{- end }} \
```
neelabhsinha, line 172: Same point as above: should be enable other parallelisms also?

config/llmisvcconfig/config-llm-decode-template.yaml
bartoszmajsak: Thanks for the review @neelabhsinha! Currently, pipeline parallelism is targeting multi-node topology and it requires `spec.worker`. The webhook enforces that direction - worker needs `data` or `pipeline`. 

I looked at the code and setting `parallelism.pipeline` without `spec.worker` is accepted by the webhook but doesn't describe a meaningful topology - maybe the webhook should reject that too, but that's a separate concern.

If there's a need to support it in the single node topology, feel free to open an issue and we can take it from there :)

neelabhsinha: @bartoszmajsak thanks for addressing the comment.
I believed that setting `parallelism.pipeline` without `spec.worker` is accepted because that's an acceptable configuration, and allowed by vLLM. vLLM never ties `--pipeline-parallel-size` to nodes and `--tensor-parallel-size` to GPUs per node. That's done by `--nnodes` parameter in case of `mp` distributed backend and via the Ray cluster for the `ray` backend. One can still have single-node-multi-GPU deployments with pipeline parallel. This might arguably be even fine in cases where user doesn't have high-speed communication like NVLink between the GPUs of same node, and one-time communication of activation transfer is working better. Is my understanding incorrect?

bartoszmajsak: You're absolutely right @neelabhsinha :) Maybe I wasn't clear in my previous response - the current kserve LLMInferenceService API associates pipeline parallelism with the `spec.worker` multi-node topology - so that's also a gap worth fixing.

Let's keep this PR focused on fixing the silent drop of tensor parallelism in the single-node templates. The original issue mentions pipeline in the title, but the description is primarily focused on tensor and I totally missed it :) Thanks for being persistent :) I will follow up with the pipeline fix.

neelabhsinha: Understood. Thanks @bartoszmajsak . We can take this up separately. Thanks for taking out the time to address the concerns.

bartoszmajsak: As discussed, I will follow-up with pipeline parallelism in dedicated PR

config/llmisvcconfig/config-llm-template.yaml
bartoszmajsak: Same comment - it will be shipped in dedicated PR

PR #5785, opened 2026-07-08
fix(rbac): harden controller manager ClusterRoles

mholder6:

Hardens RBAC for the `kserve-controller-manager` and `llmisvc-controller-manager` service accounts with a minimal, review-friendly change:

1. **Remove unused webhook RBAC** — Drop unused `MutatingWebhookConfiguration`/`ValidatingWebhookConfiguration` permissions from `kserve-manager-role`.
2. **Trim LLMISVC secret verbs** — Keep cluster-scoped secret access (needed for `Owns()` watches of TLS cert secrets in workload namespaces), but drop unused `patch`/`delete` verbs. Remaining: `get;list;watch;create;update`.
3. **KServe controller secrets** — Remain `get` only (already the case for ISVC/TrainedModel credential reads).

**Design notes (per review feedback)**:

Earlier drafts explored namespace-scoped secret Roles with `escalate`/`bind`. That approach was dropped because:
- `escalate` can allow a compromised SA to grant itself arbitrary Role permissions, which is arguably worse than cluster-wide secret access
- narrowing the secret informer breaks `Owns(&Secret{})` watches in workload namespaces
- midstream (`cmd/llmisvc/manager_options_odh.go`) depends on the existing label-filtered AllNamespaces secret cache

**Intentionally deferred**:
- Shared scheduler auth-delegator ClusterRoleBinding
- Namespace-scoped secret Roles without runtime `escalate`

- `make manifests` / chart + quick-install regeneration
- Go build of affected packages

**Checklist**:

**Release note**:
```release-note
Harden KServe controller manager RBAC by removing unused webhook permissions and trimming LLMISVC ClusterRole secret verbs while preserving secret watches needed for TLS cert reconciliation.
```


mholder6:
hey, @bartoszmajsak @VedantMahabaleshwarkar can I get a final review here?

cmd/llmisvc/main.go
```diff
@@ -194,10 +194,27 @@ func main() {
...
 				&corev1.Secret{}: {
-					Label: llmSvcCacheSelector,
+					// Limit the secret informer to the controller namespace where bootstrap
+					// RBAC is installed. Workload namespaces are accessed via the API reader
+					// after namespace-scoped Roles are reconciled.
+					Namespaces: map[string]cache.Config{
+						constants.KServeNamespace: {
+							LabelSelector: llmSvcCacheSelector,
+						},
+					},
```
bartoszmajsak: Wouldn't narrowing the cache this render watch of TLS cert secrets in workloads ns blind? I think we need to watch secrets there too and trigger reconcile 
https://github.com/kserve/kserve/blob/09ade9b873f360b7c5fedbb4a5e483deddeed581/pkg/controller/v1alpha2/llmisvc/controller.go#L376

Before this PR, the secret informer was cluster-wide (filtered by label), so `Owns()` worked for all namespaces and I think we should preserve this contract.

bartoszmajsak (commented):
I am not sure about the core trade-off this PR makes. 

Today the controller has secrets CRUD cluster-wide via ClusterRole -  and this is broad, no question about it. But the PR replaces it with `escalate+bind` on roles/rolebindings, which is arguably more risky in my eyes. `escalate` lets the SA create a Role with any permissions, not just secrets. A compromised SA can stuff arbitrary permissions into the named Role and bind it to any subject.

If the goal is trimming secret access, maybe the simplest path is narrowing the existing ClusterRole verbs to what's actually needed (create, get, update) and dropping the rest. What we need to make sure is if we still need a Watch on secrets.

charts/kserve-resources/files/kserve/resources.yaml
```diff
@@ -99,19 +99,6 @@ rules:
   - serviceaccounts
   verbs:
   - get
-- apiGroups:
-  - admissionregistration.k8s.io
```
spolti, line 103: this sounds out of scope, so, the controller will not be able to deploy webhooks if installed with charts?

mholder6: @spolti  the removal is only manager RBAC; webhook YAML is still in the chart/kustomize install path, so installs are unaffected

spolti: by removing it, you are revoking the permission the controller to handle the webhooks, is this really wanted?

mholder6: yes, it is intentional. we’re dropping unused cluster-scoped CRUD on webhook configs; the controller still serves webhooks, it just doesn’t create/update/delete the Configuration resources.

pkg/controller/v1alpha2/llmisvc/controller.go
```diff
@@ -150,13 +148,14 @@ type LLMISVCReconciler struct {
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=leaderworkerset.x-k8s.io,resources=leaderworkersets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// Secret access stays cluster-scoped so Owns() can watch TLS cert secrets in workload namespaces.
+// Verbs are trimmed to what reconcile uses (create/update/delete on force-stop) plus list/watch for the informer.
+//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;delete
```
bartoszmajsak: I think `patch` would be helpful, especially that we allow `update` too.

mholder6: Addressed in 5896ccb — restored `patch` alongside `update`/`delete` on secrets.

bartoszmajsak (approved):
/lgtm

PR #5743, opened 2026-06-30
feat: set NextProtos (ALPN) on logger TLS client transport

ugiordan:
## Summary

- Sets `NextProtos: ["h2", "http/1.1"]` on the logger TLS client transport for proper ALPN negotiation
- Removes the vestigial `enableHTTP2` flag from llmisvc. CVE-2023-44487 (HTTP/2 Rapid Reset) is fixed in Go's net/http since Go 1.21.3, and kserve uses Go 1.25.8. The flag defaulted to false, which set `NextProtos: ["http/1.1"]` and blocked HTTP/2 unnecessarily. Now always sets `NextProtos: ["h2", "http/1.1"]` for proper ALPN compliance.

## Changes

- `pkg/logger/worker.go`: Set `NextProtos: ["h2", "http/1.1"]` on logger HTTP client TLS config
- `cmd/llmisvc/main.go`: Remove `--enable-http2` CLI flag, `enableHTTP2` Options field, and `disableHTTP2` function. Always enable HTTP/2 via ALPN.

## Motivation

Without explicit `NextProtos`, Go's TLS client does not advertise protocol preferences during the TLS handshake. This can cause protocol mismatches with servers that require ALPN (like OCP infrastructure services).

The `enableHTTP2` flag was introduced as a mitigation for CVE-2023-44487, but Go 1.21.3+ already includes the fix in the stdlib. On Go 1.25.8, the flag just unnecessarily blocks HTTP/2.

PR #5791, opened 2026-07-09
feat: integrate with cluster TLS security profile

ugiordan:

Adds TLS configurability to all 4 kserve entry points via CLI flags (`--tls-min-version`, `--tls-cipher-suites`). This allows deploying operators or platforms to inject TLS configuration externally without any platform-specific code in kserve.

When no flags are set, defaults to hardened TLS 1.2 with ALPN (h2, http/1.1). The `pkg/tls/` package is a pure flag-to-`tls.Config` mapper with no API calls and no external dependencies.

Also removes the `enableHTTP2` flag from llmisvc (CVE-2023-44487 fixed in Go 1.21.3+, ref: https://go.dev/issue/63417).

Supersedes #5743.

**Changes:**
- New `pkg/tls/` package: parses `--tls-min-version` and `--tls-cipher-suites` flags into `tls.Config` options
- `pkg/tls/tls.go`: `Resolve` and `MustResolve` helpers, cipher suite validation, version parsing
- `cmd/manager/main.go`: TLS flags added, applied to webhook + metrics servers
- `cmd/llmisvc/main.go`: TLS flags added, enableHTTP2 flag removed
- `cmd/localmodel/main.go`: TLS flags added
- `cmd/localmodelnode/main.go`: TLS flags added
- 23 table-driven unit tests (version parsing, cipher suite parsing, full resolve)

- All 4 binaries build
- 23 unit tests pass
- Same approach as kubeflow/pipelines#13609

```release-note
Add TLS version and cipher suite CLI flags for webhook and metrics servers across all kserve controllers. Remove the --enable-http2 flag from llmisvc (CVE-2023-44487 fixed in Go 1.21.3+, HTTP/2 always enabled via ALPN).
```

### DIFF
--- a/cmd/llmisvc/main.go
+++ b/cmd/llmisvc/main.go
@@ -56,6 +56,7 @@ import (
 	"github.com/kserve/kserve/pkg/constants"
 	"github.com/kserve/kserve/pkg/controller/v1alpha2/llmisvc"
 	kservescheme "github.com/kserve/kserve/pkg/scheme"
+	kservetls "github.com/kserve/kserve/pkg/tls"
 	llmisvcwebhook "github.com/kserve/kserve/pkg/webhook/admission/llminferenceservice"
 )
 
@@ -81,11 +82,13 @@ type Options struct {
 	metricsAddr           string
 	webhookPort           int
 	enableLeaderElection  bool
+	enableHTTP2           bool
 	probeAddr             string
 	metricsSecure         bool
-	enableHTTP2           bool
 	migrationTimeout      time.Duration
 	migrationPollInterval time.Duration
+	tlsMinVersion         string
+	tlsCipherSuites       string
 	zapOpts               zap.Options
 }
 
@@ -96,7 +99,6 @@ func DefaultOptions() Options {
 		enableLeaderElection:  false,
 		probeAddr:             ":8081",
 		metricsSecure:         true,
-		enableHTTP2:           false,
 		migrationTimeout:      1 * time.Hour,
 		migrationPollInterval: 30 * time.Second,
 		zapOpts:               zap.Options{},
@@ -113,7 +115,9 @@ func GetOptions() Options {
 			"Enabling this will ensure there is only one active kserve controller manager.")
 	flag.StringVar(&opts.probeAddr, "health-probe-addr", opts.probeAddr, "The address the probe endpoint binds to.")
 	flag.BoolVar(&opts.metricsSecure, "metrics-secure", opts.metricsSecure, "Whether to serve metric via HTTPS.")
-	flag.BoolVar(&opts.enableHTTP2, "enable-http2", false, "If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	flag.BoolVar(&opts.enableHTTP2, "enable-http2", false, "Deprecated: CVE-2023-44487 is fixed in Go 1.21.3+. Use --tls-min-version and --tls-cipher-suites instead.")
+	flag.StringVar(&opts.tlsMinVersion, "tls-min-version", opts.tlsMinVersion, "Minimum TLS version (VersionTLS12, VersionTLS13). Defaults to VersionTLS12.")
+	flag.StringVar(&opts.tlsCipherSuites, "tls-cipher-suites", opts.tlsCipherSuites, "Comma-separated list of TLS cipher suites (Go names). If empty, Go defaults are used.")
 	flag.DurationVar(&opts.migrationTimeout, "storage-migration-timeout", opts.migrationTimeout, "Total retry budget for storage version migration.")
 	flag.DurationVar(&opts.migrationPollInterval, "storage-migration-poll-interval", opts.migrationPollInterval, "Polling interval for storage version migration retries after initial backoff.")
 	opts.zapOpts.BindFlags(flag.CommandLine)
@@ -153,24 +157,37 @@ func main() {
 		os.Exit(1)
 	}
 
-	// http/2 should be disabled due to its vulnerabilities. More specifically, disabling http/2 will
-	// prevent from being vulnerable to the HTTP/2 Stream Cancellation and
-	// Rapid Reset CVEs. For more information see:
-	// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
-	// - https://github.com/advisories/GHSA-4374-p667-p6c8
-	disableHTTP2 := func(c *tls.Config) {
-		setupLog.Info("disabling http/2")
-		c.NextProtos = []string{"http/1.1"}
-	}
+	var enableHTTP2Set bool
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == "enable-http2" {
+			enableHTTP2Set = true
+		}
+	})
 
 	var tlsOpts []func(*tls.Config)
-	if !options.enableHTTP2 {
-		tlsOpts = append(tlsOpts, disableHTTP2)
+	switch {
+	case options.tlsMinVersion != "" || options.tlsCipherSuites != "":
+		var err error
+		tlsOpts, err = kservetls.Resolve(ctx, cfg, options.tlsMinVersion, options.tlsCipherSuites)
+		if err != nil {
+			setupLog.Error(err, "unable to resolve TLS configuration")
+			os.Exit(1)
+		}
+	case enableHTTP2Set:
+		setupLog.Info("WARNING: --enable-http2 is deprecated and will be removed in a future release. " +
+			"CVE-2023-44487 is fixed in Go 1.21.3+. Use --tls-min-version and --tls-cipher-suites instead.")
+		if !options.enableHTTP2 {
+			tlsOpts = kservetls.LegacyHTTP2TLSOpts()
+		}
+	default:
+		var err error
+		tlsOpts, err = kservetls.Resolve(ctx, cfg, "", "")
+		if err != nil {
+			setupLog.Error(err, "unable to resolve TLS configuration")
+			os.Exit(1)
+		}
 	}
-	// Metrics endpoint is enabled in 'config/default/kustomization.yaml'. The Metrics options configure the server.
-	// More info:
-	// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.21.0/pkg/metrics/server
-	// - https://book.kubebuilder.io/reference/metrics.html
+
 	metricsServerOptions := metricsserver.Options{
 		BindAddress:   options.metricsAddr,
 		SecureServing: options.metricsSecure,
@@ -178,10 +195,6 @@ func main() {
 	}
 
 	if options.metricsSecure {
-		// FilterProvider is used to protect the metrics endpoint with authn/authz.
-		// These configurations ensure that only authorized users and service accounts
-		// can access the metrics endpoint. The RBAC are configured in 'config/rbac/kustomization.yaml'. More info:
-		// https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.21.0/pkg/metrics/filters#WithAuthenticationAndAuthorization
 		metricsServerOptions.FilterProvider = filters.WithAuthenticationAndAuthorization
 	}
 
--- a/cmd/localmodel/main.go
+++ b/cmd/localmodel/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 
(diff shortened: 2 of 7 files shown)


cmd/localmodel/main.go
```diff
@@ -99,14 +101,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Resolve the cluster TLS security profile before creating the manager.
```
andresllh: This function is used in several files. Can it be moved to a utils file and be called from there?

pkg/tls/tls.go
```diff
@@ -0,0 +1,180 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// +kubebuilder:rbac:groups=config.openshift.io,resources=apiservers,verbs=get
```
andresllh: Is there a way to make this cluster agnostic?

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,158 @@
...
+limitations under the License.
+*/
+
+package tls
+
+import (
+	"crypto/tls"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
```
andresllh: Can this be made cluster agnostic as well?

cmd/localmodel/main.go
ugiordan: Added a `MustResolve` wrapper that logs and exits on error.

andresllh (commented):
/lgtm

bartoszmajsak:
/hold 

This shouldn't have openshift-specific code paths in upstream. This can be solved as distro-specific hooks midstream.

https://github.com/kserve/kserve/pull/5791/changes#diff-0a7bfc6f78e5b5125c95da2d7297aa54835e746f7528ce570268dcc65cc64dbfR70-R74

pkg/tls/tls.go
```diff
@@ -0,0 +1,228 @@
...
+var tlsVersionMap = map[string]uint16{
+	"VersionTLS10": tls.VersionTLS10,
+	"VersionTLS11": tls.VersionTLS11,
+	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
+}
+
+var apiServerGVR = schema.GroupVersionResource{
+	Group:    "config.openshift.io",
+	Version:  "v1",
```
andresllh: Openshift specific

ugiordan:
> /hold
> 
> This shouldn't have openshift-specific code paths in upstream. This can be solved as distro-specific hooks midstream.
> 
> https://github.com/kserve/kserve/pull/5791/changes#diff-0a7bfc6f78e5b5125c95da2d7297aa54835e746f7528ce570268dcc65cc64dbfR70-R74

Reworked based on your feedback. Removed all platform-specific code. The PR now adds CLI flags (`--tls-min-version`, `--tls-cipher-suites`) to all 4 entry points, same approach as kubeflow/pipelines#13609. `pkg/tls/` is a pure flag-to-config mapper with no API calls. Defaults to TLS 1.2 with ALPN when no flags are set.

cmd/llmisvc/main.go
```diff
@@ -113,9 +113,10 @@ func GetOptions() Options {
 			"Enabling this will ensure there is only one active kserve controller manager.")
 	flag.StringVar(&opts.probeAddr, "health-probe-addr", opts.probeAddr, "The address the probe endpoint binds to.")
 	flag.BoolVar(&opts.metricsSecure, "metrics-secure", opts.metricsSecure, "Whether to serve metric via HTTPS.")
-	flag.BoolVar(&opts.enableHTTP2, "enable-http2", false, "If set, HTTP/2 will be enabled for the metrics and webhook servers")
```
bartoszmajsak, line 116: Isn't it potentially breaking change?

go.mod
```diff
@@ -156,6 +156,7 @@ require (
 	github.com/oasdiff/yaml v0.0.0-20250309154309-f31be36b4037 // indirect
 	github.com/oasdiff/yaml3 v0.0.0-20250309153720-d2182401db90 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
+	github.com/openshift/api v0.0.0-20260709102940-580f1c1ba691 // indirect
```
bartoszmajsak: Do we need this? Maybe `make precommit` can clean it all up?

cmd/llmisvc/main.go
andresllh: https://github.com/kserve/kserve/pull/5743#discussion_r3537687696

bartoszmajsak: Somewhat fair, but worth noting in release notes if we really want to remove it.

I was thinking about someone's deployment manifest explicitly passing `--enable-http2=false` and then updating the image. The binary would fail to start with "unknown flag."  Probably it's a minor risk but worth calling out.

andresllh: Added as release note by @ugiordan

bartoszmajsak: Maybe  we can use `flag.CommandLine.MarkDeprecated` instead of removing?

pkg/tls/tls.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, nil
+	}
+
+	allCiphers := make(map[string]uint16)
+	for _, cs := range tls.CipherSuites() {
+		allCiphers[cs.Name] = cs.ID
+	}
+	for _, cs := range tls.InsecureCipherSuites() {
+		allCiphers[cs.Name] = cs.ID
+	}
```
bartoszmajsak: Is it desired to accept insecure cipher suites?

cmd/llmisvc/main.go
```diff
@@ -113,9 +113,10 @@ func GetOptions() Options {
...
 	flag.DurationVar(&opts.migrationTimeout, "storage-migration-timeout", opts.migrationTimeout, "Total retry budget for storage version migration.")
 	flag.DurationVar(&opts.migrationPollInterval, "storage-migration-poll-interval", opts.migrationPollInterval, "Polling interval for storage version migration retries after initial backoff.")
+	flag.StringVar(&opts.tlsMinVersion, "tls-min-version", opts.tlsMinVersion, "Minimum TLS version (VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13). Defaults to VersionTLS12.")
+	flag.StringVar(&opts.tlsCipherSuites, "tls-cipher-suites", opts.tlsCipherSuites, "Comma-separated list of TLS cipher suites (Go names). If empty, Go defaults are used.")
```
bartoszmajsak, line 120: Is it really true that we accept `VersionTLS10` and `VersionTLS11`? I don't see them in the map in `tls.go`

Could be confusing help msg in all binaries.

pkg/tls/tls.go
```diff
@@ -0,0 +1,135 @@
...
+
+var tlsVersionMap = map[string]uint16{
+	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
+}
+
+// Result holds the resolved TLS configuration.
+type Result struct {
+	TLSOpts []func(*tls.Config)
+}
```
bartoszmajsak: Unless there's a plan to enrich this struct maybe we can just return slices and remove it?

bartoszmajsak (commented):
Few minor things. Thanks for this work!

cmd/llmisvc/main.go
ugiordan: already fixed (removed from map), but need to update the flag help text

cmd/llmisvc/main.go
```diff
@@ -153,35 +157,23 @@ func main() {
...
 	var tlsOpts []func(*tls.Config)
-	if !options.enableHTTP2 {
-		tlsOpts = append(tlsOpts, disableHTTP2)
+	if options.tlsMinVersion != "" || options.tlsCipherSuites != "" {
+		tlsResult := kservetls.MustResolve(options.tlsMinVersion, options.tlsCipherSuites)
+		tlsOpts = tlsResult
+	} else if !options.enableHTTP2 {
+		setupLog.Info("WARNING: --enable-http2 is deprecated and will be removed in a future release. " +
+			"CVE-2023-44487 is fixed in Go 1.21.3+. Use --tls-min-version and --tls-cipher-suites instead.")
+		tlsOpts = kservetls.LegacyHTTP2TLSOpts()
```
bartoszmajsak: This is kinda noisy, because it will be logged on every run when none of the new flags are enabled. To warn on actualy flag use we could do:

```golang
		var enableHTTP2Set bool
        flag.Visit(func(f *flag.Flag) {
                if f.Name == "enable-http2" {
                        enableHTTP2Set = true
                }
        })

        var tlsOpts []func(*tls.Config)
        switch {
        case options.tlsMinVersion != "" || options.tlsCipherSuites != "":
                tlsOpts = kservetls.MustResolve(options.tlsMinVersion, options.tlsCipherSuites)
        case enableHTTP2Set:
                setupLog.Info("WARNING: --enable-http2 is deprecated and will be removed in a future release. " +
                        "CVE-2023-44487 is fixed in Go 1.21.3+. Use --tls-min-version and --tls-cipher-suites instead.")
                if !options.enableHTTP2 {
                        tlsOpts = kservetls.LegacyHTTP2TLSOpts()
                }
        default:
                tlsOpts = kservetls.MustResolve("", "")
        }
```

ugiordan: Valid point, fixing it.

cmd/llmisvc/main.go
ugiordan: Added and deprecated the flag. 
All kserve binaries use the stdlib [flag](https://github.com/kserve/kserve/blob/master/cmd/llmisvc/main.go#L22), not pflag. Switching to pflag just for `MarkDeprecated` would mean changing the flag library for llmisvc. Is it worth the switch, or can we keep as-is?

andresllh:
@bartoszmajsak The current implementation uses stdlib `flag` (not pflag) with `flag.Visit` to detect explicit use of `--enable-http2` and logs a runtime deprecation warning, plus the deprecation note in the flag's help text. Since switching to pflag just for `MarkDeprecated` would mean changing the flag library for all llmisvc flags — is the `flag.Visit` approach acceptable, or would you prefer we make the switch to pflag?

andresllh:
The PR looks good to me overall — the hook split (`tls_default.go` with `//go:build !distro`), the `--enable-http2` deprecation path, insecure cipher removal, and the corrected flag help text all look solid. Just waiting on @bartoszmajsak's response on the stdlib `flag` vs pflag question above before this is ready.

pkg/tls/tls_default.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// MustResolve calls Resolve and exits the process if an error occurs.
+// In the default (upstream) build, ctx and cfg are unused; TLS settings
+// come from the minVersion and cipherSuites flag values.
+func MustResolve(_ context.Context, _ *rest.Config, minVersion, cipherSuites string) []func(*tls.Config) {
+	tlsOpts, err := Resolve(context.Background(), nil, minVersion, cipherSuites)
+	if err != nil {
+		log.Error(err, "Unable to resolve TLS configuration")
+		os.Exit(1)
```
bartoszmajsak: maybe panic?

pkg/tls/tls_default.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+
+	"k8s.io/client-go/rest"
+)
+
+// MustResolve calls Resolve and exits the process if an error occurs.
+// In the default (upstream) build, ctx and cfg are unused; TLS settings
+// come from the minVersion and cipherSuites flag values.
+func MustResolve(_ context.Context, _ *rest.Config, minVersion, cipherSuites string) []func(*tls.Config) {
+	tlsOpts, err := Resolve(context.Background(), nil, minVersion, cipherSuites)
```
bartoszmajsak: If we just delegate here, I think `MustResolve` can be a shared func and only `Resolve` doesn't care for ctx and config in non-distro path?

bartoszmajsak:
> @bartoszmajsak The current implementation uses stdlib `flag` (not pflag) with `flag.Visit` to detect explicit use of `--enable-http2` and logs a runtime deprecation warning, plus the deprecation note in the flag's help text. Since switching to pflag just for `MarkDeprecated` would mean changing the flag library for all llmisvc flags — is the `flag.Visit` approach acceptable, or would you prefer we make the switch to pflag?

@andresllh I think we settled on `flag.Visit`, no need to make more changes for flags handling as part of this PR

pkg/tls/tls_default.go
bartoszmajsak: Or maybe we don't need `MustResolve` at all? Caller can handle it:

```golang
tlsOpts, err := kservetls.Resolve("", "")
if err != nil {
    setupLog.Error(err, "unable to resolve TLS configuration")
    os.Exit(1)
}
```
 I would say we can delete `MustResolve` entirely.

ugiordan: Works for me. I'll remove `MustResolve` and update all 4 call sites.

bartoszmajsak (approved):
/lgtm

Thanks!

andresllh (commented):
@Jooho @spolti can we get an approval to trigger the CI?

PR #5797, opened 2026-07-09
ci: parallelize predictor-runtime-build with matrix strategy

maskarb:
## Summary
Parallelize the `kserve-image-build` (6 images) and `predictor-runtime-build` (11 images) jobs by building each image on its own runner via matrix strategies.

## Problem
Both jobs sequentially build all their Docker images on a single `ubuntu-latest` runner:
- **Slow**: `predictor-runtime-build` takes 30-55 minutes, `kserve-image-build` takes 15-25 minutes
- **Disk pressure**: ~14GB free on the runner; 11+ uncompressed images frequently exceed this
- **Fragile**: a transient `apt-get` mirror failure on one image wastes the time spent building all prior images

### `build-server-runtimes.sh` and `build-images.sh`
Add an image registry (associative array) mapping short names to dockerfiles. Both scripts now accept:
- Group names (`predictor`, `transformer`, `explainer`, `kserve`, `llmisvc`) — builds all images in that group (backward compatible)
- A specific image name like `sklearn` or `controller` — builds just that one

### `e2e-test.yml`
Replace both monolithic jobs with matrix strategies. Each matrix entry builds one image on its own runner and uploads one artifact. `fail-fast: false` ensures other images keep building if one fails.

Artifact names are preserved identically so **no downstream test jobs need changes** — they already use `pattern: base-*` / `pred-*` with `merge-multiple: true`.

## Expected improvement
| | Before | After |
|---|---|---|
| Wall clock (predictor-runtime-build) | ~45 min | ~5 min (longest single image) |
| Wall clock (kserve-image-build) | ~20 min | ~5 min (longest single image) |
| Disk per runner | ~15GB (all images) | ~1-2GB (1 image) |
| Blast radius | 1 failure kills all images | 1 failure affects 1 image |

## Test plan
- [ ] All 17 matrix jobs (6 + 11) complete successfully
- [ ] Downstream test jobs (`test-predictor`, `test-raw`, etc.) download and load all images
- [ ] `./build-server-runtimes.sh predictor,transformer` still works (backward compat)
- [ ] `./build-images.sh kserve` still works (backward compat)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -67,18 +67,21 @@ jobs:
             echo 'install_methods=["kustomize"]' >> $GITHUB_OUTPUT
           fi
 
+  kserve-image-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.gen.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - id: gen
+        run: echo "matrix=$(./test/scripts/gh-actions/build-images.sh --list-json kserve)" >> "$GITHUB_OUTPUT"
+
   kserve-image-build:
+    needs: kserve-image-matrix
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-        image:
-          - {name: controller,            image_env: CONTROLLER_IMG}
-          - {name: localmodel-controller, image_env: LOCALMODEL_CONTROLLER_IMG}
-          - {name: localmodel-agent,      image_env: LOCALMODEL_AGENT_IMG}
-          - {name: agent,                 image_env: AGENT_IMG}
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -93,64 +103,37 @@ jobs:
...
+          path: ${{ env.DOCKER_IMAGES_PATH }}/${{ env[matrix.image.image_env] }}-${{ env.TAG }}
           compression-level: 0
           if-no-files-found: error
 
   predictor-runtime-build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
```
lizzzcai: is there a way to detect and avoid some images (new) are missing?

maskarb: I will update this so that the image matrix lives in the build script, then have an initial action workflow step that dynamically builds the matrix from the list of images in the build script. That way, we would only need to update the list of images in a single place.

PR #5727, opened 2026-06-25
feat(llmisvc): add traffic splitting API for controlled deployment

bartoszmajsak:
**What this PR does / why we need it**:

Adds the API surface for controlled deployment - progressive traffic splitting across LLMInferenceService versions. This is the first PR in a stack of four that delivers the feature incrementally.

This PR introduces:

- `group` and `weight` fields on `GatewayRoutesSpec` for declaring traffic splitting group membership
- `GroupReady` and `GroupDegraded` conditions (independent of `RouterReady` rollup - a broken group doesn't block the resource from being Ready)
- `GroupStatus` and `GroupMemberStatus` types with `stopped` field for force-stop visibility
- Webhook validation (group requires weight, weight requires group)
- `v1alpha1` parity with bidirectional conversion
- Controller-managed `serving.kserve.io/routing-group` label synced from spec on every admission

No controller logic yet - that comes in the next PR. This PR is purely types, validation, conditions, and CRD schema.

**Which issue(s) this PR fixes**:
Ref #5725 

**Feature/Issue validation/testing**:

- [x] v1alpha2 validation tests for all group/weight combinations
- [x] v1alpha1 validation tests with same coverage
- [x] Conversion round-trip preserves group/weight fields
- [x] `make precommit` passes (codegen, CRD schema, lint)

**Special notes for your reviewer**:

This is the first of several stacked PRs, see [#5725](https://github.com/kserve/kserve/issues/5725) for the full design with examples, rollout lifecycle, and design decisions.

Key API decisions worth reviewing:
- `GroupReady`/`GroupDegraded` are independent of the `Ready` rollup (a broken group doesn't mean the resource can't serve inference)
- Group membership is always explicit (never inferred from `spec.model.name`)
- `GroupMemberStatus.Stopped` surfaces force-stop state so operators see declared weight alongside the stopped flag

**Checklist**:

- [x] Have you added unit/e2e tests that prove your fix is effective or that this feature works?
- [ ] Have you made corresponding changes to the [documentation](https://github.com/kserve/website)?

```release-note
Add group and weight fields to LLMInferenceService GatewayRoutesSpec for progressive traffic splitting (controlled deployment). Includes GroupReady/GroupDegraded conditions, webhook validation, v1alpha1 parity, and routing-group label.
```

Issue #5725, opened 2026-06-25
Controlled Deployment for LLMInferenceService


bartoszmajsak:
/kind feature

## Problem

Runtime upgrades, model version rotations, and configuration changes on LLM inference workloads are currently all-or-nothing. There's no way to validate a change on a fraction of production traffic without risking the take-down of the entire serving endpoint.

## Proposal

Add progressive traffic shifting via a symmetric `group` + `weight` model. Each participant independently declares its routing group membership and traffic share - there's no baseline/canary distinction in the API. The controller discovers group members, translates weights into Gateway API HTTPRoute weighted `backendRef` entries, and manages the routing lifecycle.

This enables:

- **Zero-downtime validation** - run a new version alongside production, send it a fraction of live traffic, compare per-version metrics before committing
- **Fast rollback** - adjust weights back at any point, even after stopping the old version (pods restart, traffic resumes)
- **GPU reclamation** - stop old versions to free GPU while preserving them for fast rollback
- **Simple promotion** - deploy a second LLMISVC with the same group, shift weights, remove the old one once confident

### Non-goals

This proposal does not attempt to provide:

- a durable endpoint object independent of backend lifecycle
- atomic multi-backend traffic-policy updates
- ownership/RBAC for a group separate from namespace LLMInferenceService permissions
- header/cohort routing, mirroring, A/B assignment, or rollout analysis
- a unified traffic API across InferenceService and LLMInferenceService
- a replacement for bring-your-own HTTPRoute via `route.ref`

## API Contract

- `group` is a namespace-scoped membership key. It is not an endpoint resource and does not own a routing policy independently of the member LLMInferenceServices.

- `weight` is a per-member proportional share using Gateway API backendRef weight semantics. The effective distribution is derived from all current group members and surfaced in status.

Updates are eventually consistent. When multiple members are edited, intermediate effective distributions may exist until all resources reconcile. This is acceptable for progressive rollout workflows; strict atomic routing-policy updates are out of scope for this primitive.

## Example

Both versions declare the same group. Each carries its own weight:

```yaml
# v1 (production)
apiVersion: serving.kserve.io/v1alpha2
kind: LLMInferenceService
metadata:
  name: llama-70b-v1
spec:
  model:
    name: llama-70b
    uri: hf://meta-llama/Llama-3-70B
  template:
    containers:
      - name: kserve-container
        image: vllm/vllm:v2.1
  router:
    route:
      group: llama-70b
      weight: 9
    scheduler: {}
---
[...]

PR #5798, opened 2026-07-09
feat(llmisvc): add group routing machinery for traffic splitting

bartoszmajsak:

Implements the controller logic that translates per-member `group` + `weight` fields into Gateway API HTTPRoute weighted `backendRef` entries. This is the core machinery that makes traffic splitting work - group discovery, backend resolution, weight injection, and lifecycle management.

Handles the full group lifecycle: member join/leave, weight changes, force-stop (GPU reclamation with preserved weights), model name validation via strict majority, and symmetric `GroupReady`/`GroupDegraded` condition reporting across all members. 

A finalizer-based cleanup ensures peer routes converge before a deleted member's resources are garbage-collected, preventing the 5xx window.

Each member's HTTPRoute carries weighted backendRefs for ALL group members so the gateway distributes traffic proportionally when using header-based routing. 

Per-participant rules (`/<ns>/<name>/...`) keep their single `backendRef` for direct version access.

Part of #5725

- [x] Unit tests for group member resolution, model name validation (majority-wins), backend ref matching
- [x] Unit tests for event handler fan-out (create/update/delete, old-group cleanup)  
- [x] Unit tests for `trafficFieldsChanged` predicate (weight, group, force-stop, deletion, baseRefs, scheduler changes)
- [x] Unit tests for HTTPRoute comparator with grouped route rules
- [x] Envtest integration tests: group formation, weight changes, model name ambiguity, force-stop with stopped status, member leaving group, member deletion with finalizer convergence, three-member groups

Stacked on #5727 (API surface, merged). The group routing code is split across focused files:
- `group.go` - core reconcile loop, finalization, member resolution
- `group_routing.go` - HTTPRoute rule rewriting (shared vs per-participant)
- `group_handler.go` - event handler for peer fan-out, change predicate
- `group_validation.go` - model name majority validation

**Checklist**:

```release-note
Add group routing machinery for LLMInferenceService traffic splitting - translates per-member weights into Gateway API HTTPRoute weighted backendRefs with full lifecycle management, including force-stop, model name validation, and finalizer-based safe deletion.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router_group.go
+++ b/pkg/controller/v1alpha2/llmisvc/router_group.go
@@ -1,0 +1,283 @@
+/*
+Copyright 2026 The KServe Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package llmisvc
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"slices"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	kmeta "knative.dev/pkg/kmeta"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	v1alpha2 "github.com/kserve/kserve/pkg/apis/serving/v1alpha2"
+	"github.com/kserve/kserve/pkg/constants"
+	"github.com/kserve/kserve/pkg/utils"
+)
+
+const reasonFinalizationPending = "FinalizationPending"
+
+const groupFieldIndex = ".spec.router.route.group"
+
+// resolvedMember holds the effective weight and backend for a single group member.
+type resolvedMember struct {
+	name       string
+	modelNames []string
+	weight     int32
+	stopped    bool
+	backendRef gwapiv1.BackendObjectReference
+}
+
+// setupGroupFieldIndex registers a field indexer on spec.router.route.group
+// for efficient group member discovery without listing all LLMISVCs in a namespace.
+func setupGroupFieldIndex(ctx context.Context, mgr client.FieldIndexer) error {
+	return mgr.IndexField(
+		ctx,
+		&v1alpha2.LLMInferenceService{},
+		groupFieldIndex,
+		func(obj client.Object) []string {
+			llmSvc := obj.(*v1alpha2.LLMInferenceService)
+			if g := llmSvc.Spec.Router.Group(); g != nil {
+				return []string{llmSvc.Namespace + "/" + *g}
+			}
+			return nil
+		},
+	)
+}
+
+// injectGroupBackendRefs post-processes the template-rendered HTTPRoute to add
+// weighted backendRefs for all group members. Each member's route carries
+// backendRefs for ALL members so the gateway can distribute traffic proportionally.
+//
+// Returns a groupResult for the caller to apply status after the route write.
+func (r *LLMISVCReconciler) injectGroupBackendRefs(
+	ctx context.Context,
+	llmSvc *v1alpha2.LLMInferenceService,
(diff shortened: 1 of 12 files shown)


pkg/controller/v1alpha2/llmisvc/router_group.go
pierDipi, line 1: should we call all group files as `router_group*` or `router_traffic_group*`?

bartoszmajsak: `router` and `traffic` seem a bit overlapping, but we can go either way. No hard preference here

pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -0,0 +1,350 @@
...
+// for efficient group member discovery without listing all LLMISVCs in a namespace.
+func setupGroupFieldIndex(ctx context.Context, mgr client.FieldIndexer) error {
+	return mgr.IndexField(
+		ctx,
+		&v1alpha2.LLMInferenceService{},
+		groupFieldIndex,
+		func(obj client.Object) []string {
+			llmSvc := obj.(*v1alpha2.LLMInferenceService)
+			if g := llmSvc.Spec.Router.Group(); g != nil {
+				return []string{*g}
```
pierDipi: should this be `<namespace>/<group>` or something segregated per namespace? perhaps it's already handled but maybe extra care doesn't hurt

pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return fmt.Errorf("injecting group backends: %w", err)
+	}
+
+	members := filterActiveMembers(allMembers)
+
+	resolved, err := r.resolveGroupMembers(members, route)
+	if err != nil {
+		r.Eventf(llmSvc, corev1.EventTypeWarning, reasonBackendResolutionFailed,
+			"Failed to resolve group member backend: %v", err)
```
pierDipi: the returned error will emit an event from my understanding, can we avoid duplicated (but slightly different events) and just let the caller do the emission?

pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -0,0 +1,350 @@
...
+
+	if len(divergentModels) > 0 || adapterMismatch {
+		msg := divergenceMessage(divergentModels, adapterMismatch)
+		llmSvc.MarkGroupDegraded(reasonMemberDivergence, msg)
+		r.Eventf(llmSvc, corev1.EventTypeWarning, reasonMemberDivergence,
+			"Group %q has members with divergent model.name or LoRA adapter sets",
+			*llmSvc.Spec.Router.Route.Group)
+	} else {
+		llmSvc.MarkGroupNotDegraded()
+	}
```
pierDipi: don't we end up rewriting route rules to backends with divergent models? (same for group status)

pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -0,0 +1,350 @@
...
+	route *gwapiv1.HTTPRoute,
+) error {
+	allMembers, err := r.listGroupMembers(ctx, llmSvc)
+	if err != nil {
+		return fmt.Errorf("injecting group backends: %w", err)
+	}
+
+	members := filterActiveMembers(allMembers)
+
+	resolved, err := r.resolveGroupMembers(members, route)
```
pierDipi: nit perhaps better naming this "groupMembers" or simply `members`?

pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -0,0 +1,350 @@
...
+				divergentModels = append(divergentModels, m.modelName)
+			} else {
+				adapterMismatch = true
+			}
+		}
+	}
+	resolved = matching
+
+	rewriteRulesForGroup(route, llmSvc, resolved)
+	updateGroupStatus(llmSvc, resolved)
```
pierDipi: ideally this is done based on the "current" httproute similar to "readiness" logic after we have reconciled the HTTPRoute? reason is that if it fails to update the HTTPRoute, our status is not accurate

pkg/controller/v1alpha2/llmisvc/router_group.go
pierDipi: ah no I missed `resolved = matching`

pkg/controller/v1alpha2/llmisvc/router_group.go
bartoszmajsak: Good call, right now we just filter ns while listing. It won't harm to make it more precise.

pierDipi (approved):

pierDipi (approved):

Jooho (approved):

PR #5799, opened 2026-07-10
fix(webhook): allow first-reconcile update when DeploymentMode is empty

### LINKED ISSUE #5793: validateDeploymentMode rejects first-reconcile finalizer patch because ParseDeploymentMode("") returns Standard
## /kind bug

**Affected versions:** v0.19.0 (regression from #5025, still present on `master` at time of writing).

## Summary

On a fresh install, a newly-created `InferenceService` with `serving.kserve.io/deploymentMode: Knative` can fail its very first reconcile with:

```
admission webhook "inferenceservice.kserve-webhook-server.validator" denied the request:
update rejected: deploymentMode cannot be changed from 'Standard' to 'Knative'
```

The controller never wrote `Standard` anywhere — the validator invents it by normalizing an empty `Status.DeploymentMode` through `ParseDeploymentMode`.

## Root cause

`pkg/apis/serving/v1beta1/inference_service_validation.go`:

```go
func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService) error {
    statusDeploymentMode := string(constants.ParseDeploymentMode(oldIsvc.Status.DeploymentMode))
    if len(statusDeploymentMode) != 0 {
        annotations := newIsvc.Annotations
        annotationDeploymentMode, ok := annotations[constants.DeploymentMode]
        if ok && annotationDeploymentMode != statusDeploymentMode {
            return fmt.Errorf("update rejected: deploymentMode cannot be changed from '%s' to '%s'",
                statusDeploymentMode, annotationDeploymentMode)
        }
    }
    return nil
}
```

`pkg/constants/constants.go`:

```go
const DefaultDeployment DeploymentModeType = Standard

func ParseDeploymentMode(mode string) DeploymentModeType {
    if mode == "" {
        return DefaultDeployment  // "Standard"
    }
    ...
}
```

When `oldIsvc.Status.DeploymentMode == ""` (the state of every ISVC before its first `updateStatus`), the validator sees `statusDeploymentMode = "Standard"`. The `len(...) != 0` guard was intended to skip validation in this case, but `ParseDeploymentMode` defeats it by materializing a value out of thin air.

## Why the controller trips it on the first reconcile

`pkg/controller/v1beta1/inferenceservice/controller.go` adds the finalizer via `r.Patch(...)` **before** `updateStatus` runs. That Patch is an UPDATE, so the ValidatingWebhook fires with `oldIsvc.Status.DeploymentMode == ""`. If the ISVC's annotation is `Knative` (or anything other than `Standard`), the reconcile fails; the finalizer is never persisted, `updateStatus` never runs, and every requeue hits the same wall.

Prior to #5025 the validator used `oldIsvc.Status.DeploymentMode` directly, and the `len != 0` guard did what it was supposed to.

## Reproduction

1. Install KServe v0.19.0 with `deploy.defaultDeploymentMode: Knative`.
2. Create an ISVC with `annotations: serving.kserve.io/deploymentMode: Knative` (matching the default).
3. Observe the controller error log immediately after create:
   ```
   error: admission webhook "inferenceservice.kserve-webhook-server.validator" denied the request:
   update rejected: deploymentMode cannot be changed from 'Standard' to 'Knative'
   ```

The failure appears intermittent in practice: sometimes the finalizer patch lands before the `ValidatingWebhookConfiguration` is fully enforced on the API server (racy on fresh installs), the finalizer sticks, `updateStatus` populates `Status.DeploymentMode`, and everything works from then on. When the webhook wins the race, the ISVC is permanently stuck on that reconcile.

## Expected behavior

The validator should treat an unpopulated `oldIsvc.Status.DeploymentMode` as "no prior state to protect" and let the update proceed, matching pre-#5025 behavior. Only *actually recorded* status values should be compared against the incoming annotation.

## Suggested fixes (either works)

**Option A** — check emptiness before normalizing (smallest change):

```go
raw := oldIsvc.Status.DeploymentMode
if raw == "" {
    return nil
}
statusDeploymentMode := string(constants.ParseDeploymentMode(raw))
```

**Option B** — make `ParseDeploymentMode` caller-agnostic for empty input (return `""` for empty; move the `DefaultDeployment` fallback into the reconciler paths that actually want a default).

## References

- Regressed by #5025: normalized status values on both sides but did not account for the empty-string branch of `ParseDeploymentMode`.
- #5427 normalized legacy values in status but did not address the empty-status case.
- Related symptom described in #4798.

cjohannsen-cloudera:
## Summary

- Fixes #5793: `validateDeploymentMode` was rejecting the finalizer patch on the very first reconcile of an `InferenceService` whenever its `serving.kserve.io/deploymentMode` annotation was anything other than `Standard` (e.g. `Knative`).
- Root cause: the validator normalized `oldIsvc.Status.DeploymentMode` through `constants.ParseDeploymentMode`, which folds `""` to `DefaultDeployment` (`Standard`). The `len(...) != 0` guard therefore never fired on a freshly-created ISVC, and the annotation was compared against a synthesized `Standard`.
- The fix short-circuits when `oldIsvc.Status.DeploymentMode == ""`: an unpopulated status has no prior state to protect, so the update is allowed. This restores the pre-#5025 behavior for the empty-status case while keeping legacy-value normalization intact for the populated case.
- Regression tests added covering empty status crossed with `Knative`, `LegacyServerless`, `Standard`, and no annotation — the `Knative` case reproduces the reported failure without the fix.

## Test plan

- [x] `go test ./pkg/apis/serving/v1beta1/... -run TestDeploymentModeUpdate` fails before the fix with the exact reported error, passes after
- [x] `go test ./pkg/apis/serving/v1beta1/...` (full package) passes
- [x] `make test`
- [x] `make test-qpext`
- [x] `make py-lint`
- [x] `make vet` / `make go-lint`
- [x] E2E smoke on `kind-kserve` (k8s 1.33.1, Knative-mode install):
  - Bug reproduced against **released** webhook: `admission webhook "inferenceservice.kserve-webhook-server.validator" denied the request: update rejected: deploymentMode cannot be changed from 'Standard' to 'Knative'` on every reconcile of a fresh ISVC annotated `serving.kserve.io/deploymentMode: Knative`.
  - Controller re-rolled from a `ko`-built image with this patch applied, same ISVC recreated: zero admission rejections, `status.deploymentMode=Knative` populated, finalizer added, ISVC progresses normally through Knative Revision creation.

### DIFF
--- a/pkg/apis/serving/v1beta1/inference_service_validation.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation.go
@@ -589,15 +589,20 @@ func validateCollocationStorageURI(predictorSpec PredictorSpec) error {
 	return nil
 }
 
-// validates if the deploymentMode specified in the annotation is not different from the one recorded in the status
+// validates if the deploymentMode specified in the annotation is not different from the one recorded in the status.
+//
+// An empty oldIsvc.Status.DeploymentMode means the controller has not yet run its first updateStatus
+// (e.g. the finalizer patch on first reconcile). In that case there is no prior state to protect, so
+// the update is allowed. Passing "" through ParseDeploymentMode would fold to DefaultDeployment
+// ("Standard") and reject any non-Standard annotation, breaking first-reconcile on Knative installs.
 func validateDeploymentMode(newIsvc *InferenceService, oldIsvc *InferenceService) error {
+	if oldIsvc.Status.DeploymentMode == "" {
+		return nil
+	}
 	statusDeploymentMode := string(constants.ParseDeploymentMode(oldIsvc.Status.DeploymentMode))
-	if len(statusDeploymentMode) != 0 {
-		annotations := newIsvc.Annotations
-		annotationDeploymentMode, ok := annotations[constants.DeploymentMode]
-		if ok && annotationDeploymentMode != statusDeploymentMode {
-			return fmt.Errorf("update rejected: deploymentMode cannot be changed from '%s' to '%s'", statusDeploymentMode, annotationDeploymentMode)
-		}
+	annotationDeploymentMode, ok := newIsvc.Annotations[constants.DeploymentMode]
+	if ok && annotationDeploymentMode != statusDeploymentMode {
+		return fmt.Errorf("update rejected: deploymentMode cannot be changed from '%s' to '%s'", statusDeploymentMode, annotationDeploymentMode)
 	}
 	return nil
 }
--- a/pkg/apis/serving/v1beta1/inference_service_validation_test.go
+++ b/pkg/apis/serving/v1beta1/inference_service_validation_test.go
@@ -1471,6 +1471,48 @@ func TestDeploymentModeUpdate(t *testing.T) {
 	// During deletion, deploymentMode mismatch should be allowed
 	g.Expect(warnings).Should(gomega.BeEmpty())
 	g.Expect(err).Should(gomega.Succeed())
+
+	// Test: Empty Status.DeploymentMode with a non-Standard annotation should be accepted.
+	// This is the first-reconcile finalizer patch case: the controller adds a finalizer via
+	// r.Patch(...) before updateStatus runs, so oldIsvc.Status.DeploymentMode is "". The
+	// validator must treat that as "no prior state" and not synthesize "Standard" from it.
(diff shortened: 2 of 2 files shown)


andresllh (commented):
/lgtm

PR #5800, opened 2026-07-10
feat(llmisvc): warn when model-routing names collide across LLMISVCs

bartoszmajsak:

If two LLMISVCs in the same namespace produce identical model-routing header values (`publishers/{ns}/models/{name}`), Gateway API picks the oldest route and the other one just stops getting traffic silently.

This PR adds a collision check at reconcile time that emits a `ModelNameCollision` warning event when overlapping model-routing names are detected. Compares self's effective model names (base model + LoRA adapters) from post-merge spec against each peer's `status.addresses[].models` - the resolved state that already includes baseRef overrides. Catches both base-model and LoRA adapter collisions.

Does not check whether the peer has model-based routing enabled - that state can flip at any time (annotation, gateway config), but a conservative warning beats is still helpful.

Follow-up from the discussion in #5521.

- [x] Unit tests for `findModelNameCollisions` covering: no collision, single/multiple collisions, same group excluded, asymmetric group membership, nil model.name defaulting, deleted peer filtering, empty status (not yet reconciled), LoRA adapter vs base model collision (both directions), conservative warning when peer lacks model-routing annotation
- [x] All existing tests pass (`go test ./pkg/controller/v1alpha2/llmisvc/...`)
- [x] `make precommit` passes (lint, vet, codegen)

The check is intentionally conservative - warns based on model-routing name overlap without verifying per-peer enablement. Peers that haven't been reconciled yet (empty status) are silently skipped and caught on the next reconcile.

**Checklist**:

```release-note
Emit a warning event when LLMInferenceService resources in the same namespace produce overlapping model-routing header values (base model or LoRA adapters), which causes the gateway to silently shadow one service.
```

### DIFF
--- a/pkg/controller/v1alpha2/llmisvc/router.go
+++ b/pkg/controller/v1alpha2/llmisvc/router.go
@@ -24,6 +24,7 @@ import (
 	"slices"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"knative.dev/pkg/apis"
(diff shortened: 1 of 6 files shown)


pkg/controller/v1alpha2/llmisvc/router.go
```diff
@@ -160,6 +161,21 @@ func (r *LLMISVCReconciler) reconcileHTTPRoutes(ctx context.Context, llmSvc *v1a
...
+		// activating a dormant collision.
+		if r.isModelBasedRoutingEnabled(ctx, llmSvc, cfg) {
+			others := &v1alpha2.LLMInferenceServiceList{}
+			if err := r.List(ctx, others, client.InNamespace(llmSvc.Namespace)); err != nil {
+				logger.Error(err, "failed to list LLMISVCs for model name collision check")
+			} else if collisions := findModelNameCollisions(llmSvc, others.Items); len(collisions) > 0 {
+				r.Eventf(llmSvc, corev1.EventTypeWarning, "ModelNameCollision",
+					"model-routing header values overlap with %s in namespace %s; "+
+						"identical header matches cause the gateway to shadow one service",
+					strings.Join(collisions, ", "), llmSvc.Namespace)
```
pierDipi: should we additionally add "you can continue accessing the service directly through the `<direct address name>` address, and you can ignore if you intend to enable traffic group splitting later" (or similar verbiage)?

bartoszmajsak: I like it, that's helpful.

pierDipi: idk if there is a limit of text though, let's double check

bartoszmajsak: @pierDipi I added extra info based on your suggestion. Thanks!

pierDipi (approved):

bartoszmajsak:
/hold

better to merge #5798 first

PR #5802, opened 2026-07-10
fix(autogluon): prevent path traversal in model loading

Wojciech-Rebisz:
## Summary
- Add `_validate_model_path()` to `AutoGluonModelRepository` to reject empty model names, absolute paths, and resolved paths that escape the configured models directory.
- Use the validated canonical path when loading models via `load_model()`, preventing path traversal attacks (e.g. `../etc/passwd`, nested `..` segments, and symlink-based escapes).
- Add unit tests covering traversal attempts, invalid names, and valid model loading behavior.
## Motivation
Previously, model names were joined directly with `models_dir` without validation. A malicious or malformed model name could resolve outside the intended mount directory and expose arbitrary filesystem paths during model load.
## Test plan
- [ ] Run autogluon server unit tests: `pytest python/autogluonserver/tests/test_autogluon_model_repository.py`
- [ ] Verify valid model names (e.g. `valid-model-123`) still load successfully
- [ ] Verify traversal attempts raise `ValueError` with appropriate messages:
  - `../etc/passwd`, `../../tmp/evil`
  - absolute paths (`/etc/passwd`, Windows drive paths)
  - nested traversal (`model/../../../etc/passwd`)
  - empty and whitespace-only names

Assisted with <Claude Sonnet 4.5>

### DIFF
--- a/python/autogluonserver/autogluonserver/autogluon_model_repository.py
+++ b/python/autogluonserver/autogluonserver/autogluon_model_repository.py
@@ -26,8 +26,31 @@ class AutoGluonModelRepository(ModelRepository):
     async def load(self, name: str) -> bool:
         return self.load_model(name)
 
+    def _validate_model_path(self, name: str) -> str:
+        """Validate model name to prevent path traversal attacks."""
+        if not name or not name.strip():
+            raise ValueError("Model name cannot be empty")
+
+        if os.path.isabs(name):
+            raise ValueError(f"Model name cannot be an absolute path: {name}")
+
+        models_dir_real = os.path.realpath(self.models_dir)
+        candidate_path = os.path.join(models_dir_real, name)
+        candidate_path_real = os.path.realpath(candidate_path)
+
+        try:
+            common_path = os.path.commonpath([models_dir_real, candidate_path_real])
+        except ValueError:
+            raise ValueError(f"Model path traversal detected: {name}")
+
+        if common_path != models_dir_real:
+            raise ValueError(f"Model path traversal detected: {name}")
+
+        return candidate_path_real
+
     def load_model(self, name: str) -> bool:
-        model = create_autogluon_model(name, os.path.join(self.models_dir, name))
+        model_path = self._validate_model_path(name)
+        model = create_autogluon_model(name, model_path)
         if model.load():
             self.update(model)
         return model.ready
--- a/python/autogluonserver/tests/test_autogluon_model_repository.py
+++ b/python/autogluonserver/tests/test_autogluon_model_repository.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import pytest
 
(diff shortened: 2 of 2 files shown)


Wojciech-Rebisz:
@Jooho, @lizzzcai @greenmoon55  can I ask for review?

python/autogluonserver/autogluonserver/autogluon_model_repository.py
```diff
@@ -26,8 +26,31 @@ def __init__(self, model_dir: str = MODEL_MOUNT_DIRS):
...
+
+        if os.path.isabs(name):
+            raise ValueError(f"Model name cannot be an absolute path: {name}")
+
+        models_dir_real = os.path.realpath(self.models_dir)
+        candidate_path = os.path.join(models_dir_real, name)
+        candidate_path_real = os.path.realpath(candidate_path)
+
+        try:
+            common_path = os.path.commonpath([models_dir_real, candidate_path_real])
```
spolti, line 42: Wouldn't `return os.path.relpath(os.path.join("/", path), "/")` be enough for this?

Wojciech-Rebisz: I tried your approach, but `relpath` silently rewrites `../etc/passwd` to `etc/passwd` instead of rejecting it, see sample below:
```python
import os
print(os.path.relpath(os.path.join("/", "../etc/passwd"), "/"))
# etc/passwd
```
The second problem are symlinks, if we assume that we have a simlink:
```
models_dir/malicious_link ->  /etc/
```
then `malicious_link/passwd` looks perfectly fine after normalization, but `realpath()` follows the symlink and reveals the actual target `/etc/passwd`, which `commonpath` catches.

spolti: nice catch, thanks.

spolti:
@Wojciech-Rebisz can you please send an empty commit to trigger CI? I can't rerun it.

Wojciech-Rebisz:
Hi @spolti, thank you for comments, I rerun the CI checks and these passed. Will this merge automatically after this?

mwaykole:
The core `realpath` + `commonpath` check looks solid. One issue with the new tests: `test_path_traversal_blocked_url_encoded` and `test_path_traversal_blocked_mixed_separators` assert `pytest.raises((ValueError, Exception))`, so they still pass if validation never fires and model load fails for some other reason (e.g. `..%2F..%2Ftmp%2Fevil` is just a weird relative name under `models_dir`, not traversal). Please tighten those to expect `ValueError` with the traversal message, or drop them if they aren't real attack vectors on Unix.

Wojciech-Rebisz:
Hi @mwaykole, thanks for comment, I removed these tests as they seem not to be real attack vectors.

mwaykole:
@Wojciech-Rebisz Thanks — dropping those two cases makes sense if they are not real traversal vectors on Unix.

PR #5806, opened 2026-07-11
fix: check sRuntime.WorkerSpec for nil before dereferencing in reconc…

### LINKED ISSUE #5805: multiNodeProcess and reconcileWorker dereference sRuntime.WorkerSpec before nil check, causing a panic
In pkg/controller/v1beta1/inferenceservice/components/predictor.go, both reconcileWorker and multiNodeProcess dereference sRuntime.WorkerSpec before checking whether it's nil.

WorkerSpec is an optional pointer field on ServingRuntimeSpec:

```go
WorkerSpec *WorkerSpec `json:"workerSpec,omitempty"`
```

reconcileWorker accesses it immediately on entry:

```go
sRuntimeWorkerAnnotations := sRuntime.WorkerSpec.Annotations
sRuntimeWorkerLabels := sRuntime.WorkerSpec.Labels
```

It then calls multiNodeProcess, which dereferences sRuntime.WorkerSpec several more times (setting PipelineParallelSize, TensorParallelSize, and reading isvc.Spec.Predictor.WorkerSpec fields to assign into it) before finally reaching its own nil check:

```go
if sRuntime.WorkerSpec == nil {
    errMsg := "you cannot set WorkerSpec in the InferenceService if the ServingRuntime does not have a WorkerSpec"
    isvc.Status.PropagateRawStatusWithMessages(v1beta1.PredictorComponent, v1beta1.InvalidWorkerSpecNotSet, errMsg, corev1.ConditionFalse)
    return nil, errors.New(errMsg)
}
```

Since sRuntime.WorkerSpec is used multiple times before this check runs, any InferenceService that sets predictor.workerSpec while referencing a ServingRuntime with no workerSpec configured will cause a nil pointer 
dereference panic in the controller, rather than returning the clean error this check is clearly intended to produce.

Proposed fix: move the nil check to the top of both reconcileWorker and multiNodeProcess, before any field access on sRuntime.WorkerSpec. This is a reordering only — the error message and control flow already exist 
and are correct, they just run too late.

jahnavigajjala-3:

Fixes a nil pointer dereference panic in reconcileWorker and multiNodeProcess. Both functions previously dereferenced sRuntime.WorkerSpec (an optional pointer field) before checking whether it was nil. Any InferenceService that sets 
predictor.workerSpec while referencing a ServingRuntime with no workerSpec configured would cause the controller to panic instead of returning the ntended error message.

This PR moves the existing nil check to the top of both functions, before any field access on sRuntime.WorkerSpec. No new logic or error messages are introduced — the correct error handling already existed, it just ran too late to prevent the panic.

Fixes #5805

- [x] Verified go build ./... succeeds with the change
- [x] Verified existing tests in pkg/controller/v1beta1/inferenceservice/components/... still pass
- [x] Added a new test case with a ServingRuntime that has no WorkerSpec, 
      confirming a clean error is returned instead of a panic

This is a reordering-only fix — the error messages and control flow that already existed are unchanged, they now just run before the WorkerSpec field accesses instead of after.

**Checklist**:

**Release note**:
```release-note
Fixed a nil pointer dereference panic in the InferenceService controller when a multi-node predictor references a ServingRuntime without a WorkerSpec.
```


mwaykole:
Moving the nil `WorkerSpec` check above the `PipelineParallelSize` / `TensorParallelSize` defaulting is the real fix for the panic — good. The same guard is now duplicated in `reconcileWorker` and `multiNodeProcess`; consider keeping status propagation only in `reconcileWorker` and letting `multiNodeProcess` just return the error so the helper doesn't side-effect status from every call path.

PR #5808, opened 2026-07-13
feat(llmisvc): add structured model identity to status

bartoszmajsak:

`status.addresses[].models[]` is a flat list providing both base model names and LoRA adapter names with no simple way to distinguish them. The same name appears in both publisher-qualified (`publishers/ns/models/name`) and short forms, depending on address type. 

This adds `status.model` which is a structured field carrying the post-config-merge model identity: the effective runtime name and sorted LoRA adapter set. Provides a single canonical source for "what model does this resource serve" without understanding baseRef merge internals or parsing address lists.

Authoritative when `PresetsCombined=True`. Nil when merge has never succeeded (`ConfigNotFound`, `CombineBaseError`). Retains the last successful value on force-stop.

- [x] Unit tests for `observeModelStatus` covering: explicit model + adapters, name defaulting to service name, nil-named adapter skipping (validation bypass), empty LoRA spec
- [x] Integration test assertions in ConfigNotFound/recovery flow: `PresetsCombined=False` => `status.model == nil`, recovery to `PresetsCombined=True` => `status.model.name` populated with expected value
- [x] All existing tests pass (`go test ./pkg/controller/v1alpha2/llmisvc/...`)
- [x] CRD codegen, deepcopy updated

`status.model` is deliberately separate from `addresses[].models` because they serve different purposes. The address models list carries routable aliases (both qualified and unqualified forms) per endpoint. `status.model` is resource-level identity. This provides one canonical name, adapters cleanly separated, no per-address variation.

The `PresetsCombined` guard on the call site prevents writing unmerged identity on the force-stop path, where `reconcileBaseRefs` returns a raw user spec copy. On early failures (`loadConfig`), status fields are left as-is and `observedGeneration` lag signals staleness. This is consistent with how every other status field behaves.

**Checklist**:

```release-note
Add `status.model` to LLMInferenceService with the effective model name and LoRA adapter set after config merge, giving consumers a structured alternative to parsing the flat `addresses[].models` list.
```


pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -863,6 +863,19 @@ type AppliedConfigRef struct {
...
+// after applying base configurations and service-specific overrides.
+type ObservedModelStatus struct {
+	// Name is the effective model name configured for the inference runtime.
+	Name string `json:"name"`
+
+	// Adapters lists the effective names of the configured LoRA adapters,
+	// sorted lexicographically for stable identity comparisons.
+	// +optional
+	// +listType=atomic
+	Adapters []string `json:"adapters,omitempty"`
```
pierDipi, line 876: problem is that some model names are `SourcedAddress`-specific

bartoszmajsak: I thought it would be helpful to have "effective spec" in status so we could always know what's defined without triggering merge logic.

We still have address-specific models discoverable in the existing `addresses` struct.

Is it misleading with this addition?

pierDipi: > we could always know what's defined without triggering merge logic

we in this case is the controller? or a user?

bartoszmajsak: Pretty much both :)

bartoszmajsak:
Closing, as we have all model names in the addresses / routing discovery. Adding this is not necessary now and can be problematic.

PR #5820, opened 2026-07-13
feat(llmisvc): gate group traffic splitting on member readiness

bartoszmajsak:

When a group member's backend is not yet ready (no available pods, routing infrastructure not created), the gateway routes its share of traffic to a dead endpoint and might return 503s.

Introduces a readiness gate for group membership: members are only included in peers' HTTPRoute weighted backendRefs when they have `RouterReady=True` and at least one available replica on every configured workload (primary, prefill, scheduler).

Adds `ObservedWorkloadStatus` to surface `availableReplicas` from Deployment and LeaderWorkerSet status, giving external consumers workload health visibility without querying individual deployments. Backward compatible - existing JSON fields unchanged, `availableReplicas` is additive.

Related to #5725

- [x] `TestFilterEligibleMembers` - 8 cases: terminating, no RouterReady, zero replicas, zero prefill replicas, nil scheduler replicas, force-stopped, self always included, routable peers included
- [x] `TestTrafficFieldsChanged` - peer becomes routable, peer becomes unroutable, replica count change without routable transition (no false positive)
- [x] `TestObserveWorkloadStatus` - 10 existing cases pass with new type
- [x] Envtest type fixes for `ObservedWorkloadStatus` in multi-node, scheduler config, and main integration tests

Verified on kind cluster:
- `RouterReady` stays True during deployment scaling (1->2, 2->1)
- `availableReplicas` stays >0 during scale-up (no flapping)
- Only drops to 0 when ALL pods are down

`ObservedWorkloadStatus` embeds `corev1.TypedLocalObjectReference` inline, so `status.workloads.primary.name` still works. The new `availableReplicas` field is optional and additive - no breaking change for existing consumers.

The readiness gate cascades correctly: member A's replica change triggers re-enqueue of peers B/C, they update their HTTPRoutes, but their own routable state doesn't change so no further fan-out. Confirmed by both golang-reviewer and dev-k8s-ctrl agents independently.

**Checklist**:

- [x] Existing unit tests updated for new type
- [x] New unit tests for readiness gate logic
- [x] CRD manifests regenerated


pkg/controller/v1alpha2/llmisvc/router_group.go
```diff
@@ -276,12 +276,36 @@ func (r *LLMISVCReconciler) listGroupMembers(
...
+// filterEligibleMembers returns non-terminating, routable members. Self always passes.
+func filterEligibleMembers(members []v1alpha2.LLMInferenceService, selfName string) []v1alpha2.LLMInferenceService {
+	eligible := make([]v1alpha2.LLMInferenceService, 0, len(members))
 	for i := range members {
-		if members[i].DeletionTimestamp == nil {
-			active = append(active, members[i])
+		if !members[i].DeletionTimestamp.IsZero() {
+			continue
+		}
+		if members[i].Name != selfName && !memberIsRoutable(&members[i]) {
```
spolti: maybe change to `isMemberRoutable`?

bartoszmajsak: why? It's not Java and this is not JavaBean :D

bartoszmajsak: But I see `isXYZ` is used elsewhere, so that's fine, I can adjust

spolti (approved):

pierDipi (approved):

PR #5822, opened 2026-07-13
feat(llmisvc): add stable and version-independent URLs

bartoszmajsak:

Adds a stable, version-independent URL path for LLMInferenceService endpoints using `spec.model.name` instead of `metadata.name`.

Today, every participant's URL includes its `metadata.name` - which changes across versions during controlled deployments. Consumers need a stable endpoint that survives model upgrades. The publisher path (`/publishers/{namespace}/models/{model-name}`) provides exactly that - a single URL that follows the model identity rather than the participant identity, and naturally participates in weighted traffic splitting when multiple grouped members serve the same model. This approach is aligned with header-based routing that we already have in place.

Also lifts the model name collision warning out of the model-routing-only gate. Since publisher paths are always present on managed routes, collisions matter regardless of whether header-based model routing is enabled.

## Publisher path and LoRA adapters

The HTTPRoute exposes one stable publisher-qualified path for the base model. It   does not create a separate path for each LoRA adapter. Clients invoke an adapter through the base model's publisher path and select it using the request body's `model` field:

```http
POST /publishers/ns/models/base-model/v1/chat/completions
{
  "model": "publishers/ns/models/sql-expert"
}
```

Explicit per-adapter publisher paths are intentionally out of scope. Preserving the same behavior as the base publisher path would require five rules per adapter - four inference endpoint rewrites and one Service catch-all. Gateway API requires every rule using ReplacePrefixMatch to contain exactly one PathPrefix match, so these rules cannot be combined. Since the base route already uses 15 of the 16 available rules, even one fully mirrored adapter path would require HTTPRoute sharding. Sharding can be introduced separately if dedicated adapter URLs become a requirement

Related to #5725 

- [x] Unit tests for `extractRoutePaths` covering publisher path bucketing (5 cases: all-three-present, publisher-without-model-routing, no-publisher, publisher-only, service-backed preference)
- [x] Existing `findModelNameCollisions` tests pass unchanged - detection logic is shared

The publisher path reuses the same `fullyQualifiedModelName` identity (`publishers/{ns}/models/{name}`) already used by model-routing headers. This means a single collision check covers both routing surfaces.

**Checklist**:

```release-note
Add publisher-qualified model path (/publishers/{namespace}/models/{model-name}) providing stable, version-independent URLs for LLMInferenceService endpoints during controlled deployments.
```

### DIFF
--- a/config/llmisvcconfig/config-llm-router-route.yaml
+++ b/config/llmisvcconfig/config-llm-router-route.yaml
@@ -223,6 +223,119 @@ spec:
               timeouts:
                 backendRequest: 0s
                 request: 0s
+            # Publisher-qualified model paths provide a stable, version-independent
+            # endpoint using spec.model.name instead of metadata.name.
+            # The /publishers/ path prefix is reserved for this purpose.
+            # These mirror the per-participant endpoint rules above.
+            - name: v1-completions-publisher-path
+              backendRefs:
+                - group: inference.networking.k8s.io
+                  kind: InferencePool
+                  name: |-
+                    {{ ChildName .ObjectMeta.Name `-inference-pool` }}
+                  port: 8000
+                  weight: 1
+              matches:
+                - path:
+                    type: PathPrefix
+                    value: |-
+                      /publishers/{{ .ObjectMeta.Namespace }}/models/{{ .Spec.Model.Name }}/v1/completions
+              filters:
+                - type: URLRewrite
+                  urlRewrite:
+                    path:
+                      type: ReplacePrefixMatch
+                      replacePrefixMatch: /v1/completions
+              timeouts:
+                backendRequest: 0s
+                request: 0s
+            - name: v1-chat-completions-publisher-path
+              backendRefs:
+                - group: inference.networking.k8s.io
+                  kind: InferencePool
+                  name: |-
+                    {{ ChildName .ObjectMeta.Name `-inference-pool` }}
+                  port: 8000
+                  weight: 1
+              matches:
+                - path:
+                    type: PathPrefix
+                    value: |-
+                      /publishers/{{ .ObjectMeta.Namespace }}/models/{{ .Spec.Model.Name }}/v1/chat/completions
+              filters:
+                - type: URLRewrite
+                  urlRewrite:
+                    path:
+                      type: ReplacePrefixMatch
+                      replacePrefixMatch: /v1/chat/completions
+              timeouts:
+                backendRequest: 0s
+                request: 0s
+            - name: v1-responses-publisher-path
+              backendRefs:
+                - group: inference.networking.k8s.io
+                  kind: InferencePool
+                  name: |-
+                    {{ ChildName .ObjectMeta.Name `-inference-pool` }}
+                  port: 8000
+                  weight: 1
+              matches:
+                - path:
+                    type: PathPrefix
+                    value: |-
+                      /publishers/{{ .ObjectMeta.Namespace }}/models/{{ .Spec.Model.Name }}/v1/responses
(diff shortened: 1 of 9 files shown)


config/llmisvcconfig/config-llm-router-route.yaml
```diff
@@ -223,6 +223,32 @@ spec:
...
+            # Participates in weighted splitting for grouped members.
+            # The /publishers/ path prefix is reserved for this purpose.
+            - name: v1-publisher-path-catch-all
+              backendRefs:
+                - group: inference.networking.k8s.io
+                  kind: InferencePool
+                  name: |-
+                    {{ ChildName .ObjectMeta.Name `-inference-pool` }}
+                  port: 8000
+                  weight: 1
```
pierDipi: catch-all usually goes to Service

config/llmisvcconfig/config-llm-router-route.yaml
```diff
@@ -223,6 +223,32 @@ spec:
...
+              matches:
+                - path:
+                    type: PathPrefix
+                    value: |-
+                      /publishers/{{ .ObjectMeta.Namespace }}/models/{{ .Spec.Model.Name }}
```
pierDipi, line 329: did we try with having additional matches on existing rules?

bartoszmajsak: We are capped at [8 per rule](https://github.com/kubernetes-sigs/gateway-api/issues/1485) and 16 matches overall. Probably at this point worth noting as limitation?

I wonder if we really need LoRA path expansion - maybe for now it's sufficient to assume we hit canonical path and LoRA is part of the payload as model id? Or are we thinking of LoRA being entirely separate entity and shouldn't be coupled with the base model?

bartoszmajsak: I will try to collapse some of them to make more headroom, but I still think it's worth calling out the limitation in the docs and plan additional logic for extra `HTTPRoute`s when set of models+LoRAs will exceed it.

pierDipi: it got increased? https://github.com/kubernetes-sigs/gateway-api/pull/3205

pierDipi: https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/httproute_types.go#L134

```
While 16 rules and 64 matches per rule are allowed, the total number of matches across all rules in a route must be less than 128
```

bartoszmajsak: > did we try with having additional matches on existing rules?

Doesn't look like we can:

```
When using URLRewrite filter with path.replacePrefixMatch, exactly one PathPrefix match must be specified
```

https://github.com/kubernetes-sigs/gateway-api/blob/d9ccc3f62e8ff65fc024bd0f776e4f229aa33c29/apis/v1/httproute_types.go#L144

pierDipi: that is very unfortunate

config/llmisvcconfig/config-llm-router-route.yaml
bartoszmajsak: added separate `v1-publisher-path-catch-all`

config/llmisvcconfig/config-llm-router-route.yaml
bartoszmajsak: Captured the outcome of our discussion in PR description and will add to the docs too.

pierDipi (approved):

PR #5827, opened 2026-07-14
fix(llmisvc): overlay timeout-only HTTPRoute rule patches onto presets

### LINKED ISSUE #5706: LLMInferenceService: router.route.http.spec is a full replace — supplying hostnames drops all rules and backendRefs
/kind bug

### Summary

`pkg/controller/v1alpha2/llmisvc/router.go:212` performs

    httpRoute.Spec = *llmSvc.Spec.Router.Route.HTTP.Spec.DeepCopy()

— a full replacement, not a merge. Any field supplied in `router.route.http.spec` causes the controller to drop all preset-generated rules (InferencePool backendRefs, model-name routing, path-prefix variants, catch-all). A user who wants to add only a `hostnames` field has to re-supply the entire 6-8-rule HTTPRouteSpec.

This is undocumented as far as I can tell, and the obvious interpretation of `router.route.http.spec` is "additional overrides on top of the defaults", which is what users keep attempting.

### Reproducer (v0.19.0)

1. Create an LLMInferenceService with defaults:

       router:
         gateway:
           refs:
           - {name: my-gateway, namespace: gw}
         route: {}
         scheduler: {}

   Controller generates ~6 rules covering `/v1/completions`, `/v1/chat/completions`, `/v1/responses`, model-name match, catch-all.

2. Edit only to add a hostname:

       router:
         route:
           http:
             spec:
               hostnames:
               - my-svc.example.com

3. Regenerated HTTPRoute has only `hostnames`. `rules`, `parentRefs`, `backendRefs` are gone. Traffic stops.

### Suggested fix

Two options, in order of preference:

1. **Merge top-level fields.** Treat `hostnames` and `parentRefs` as additive when the user does not supply `rules`. When the user supplies `rules`, replace. (Matches user expectation.)
2. **Document the full-replace contract.** Add to the field godoc and the chart README that `router.route.http.spec` is exclusive — provide a complete spec or none. (Cheap; does not fix the cliff.)

### Environment

- KServe: v0.19.0

sahilleth:

Fixes a sharp edge with `router.route.http.spec` composition for LLMInferenceService (#5706).

`MergeSpecs` uses Kubernetes strategic merge. For `HTTPRouteSpec.Rules`, that means:

- **Hostnames / parentRefs only** already preserve preset rules (verified).
- **Any non-empty `rules` override replaces the entire Rules list**, so the docs "Custom Timeouts" pattern (a rule with only `timeouts`) wiped backendRefs/matches and broke routing.

This PR:

1. Detects timeout-only rule patches and overlays those timeouts onto the preset rules instead of replacing them.
2. Clarifies the `HTTPRouteSpec.Spec` godoc so merge semantics are explicit.
3. Adds regression tests for hostnames-only merge, timeout-only overlay, and full Rules replacement.

Fixes #5706

- [x] `go test ./pkg/controller/v1alpha2/llmisvc/ -run TestMergeSpecs_HTTPRoute -count=1`
- [x] Existing merge behavior for complete Rules overrides still replaces (covered by test)

While investigating #5706 I could **not** reproduce hostnames-only wiping rules on current `master` — strategic merge already preserves them. The reproducible cliff matching user expectation / docs examples is timeout-only (and similarly incomplete) `rules` patches. Happy to adjust if maintainers prefer docs-only clarification for hostnames and a different approach for timeouts.

## Checklist

**Follow-up:** website docs for "Custom HTTPRoute Spec" should call out merge semantics (I can open a `kserve/website` PR if desired).

## Release note

```release-note
Fixed LLMInferenceService HTTPRoute composition so timeout-only rule patches overlay preset rules instead of replacing them and dropping backendRefs.
```


sahilleth:
Docs companion PR: https://github.com/kserve/website/pull/702

pkg/apis/serving/v1alpha2/llm_inference_service_types.go
```diff
@@ -412,6 +412,13 @@ type HTTPRouteSpec struct {
 
 	// Spec allows for providing a custom specification for an HTTPRoute.
 	// If provided, the controller will create and manage an HTTPRoute with this specification.
+	//
+	// Spec fields are strategically merged with the well-known router-route preset:
+	// top-level fields such as hostnames and parentRefs can be set alone without losing
+	// preset rules/backendRefs. Supplying Rules replaces the entire preset Rules list
+	// (provide a complete list). An exception is timeout-only rule patches (Rules entries
+	// that set only timeouts), which overlay those timeouts onto the preset rules instead
+	// of replacing them.
```
pierDipi: I'm wondering if we should define a field like `ruleDefaults` instead so that usage looks like this:

```yaml
spec:
  router:
    route:
      ruleDefaults:
        timeouts: { ... }
        retry: ...
        sessionPersistence: ...
```

sahilleth: I like the idea. ruleDefaults would definitely provide a cleaner and more extensible API. Since this PR is intended as a targeted fix for timeout overlays, would you prefer I keep that refactor in this PR or address it in a follow-up?

bartoszmajsak: I like `ruleDefaults` idea, and I think it's worth addressing this way, Maybe we can do it in this PR since it seems to be a fix for a broader problem?

sahilleth:
Thanks for the review - @bartoszmajsak , @pierDipi 
I've updated the PR based on your feedback:

- Applied the requested ruleDefaults merge behavior.
- Removed the timeout-only special case.
- Dropped unrelated fixture and dependency changes to keep the PR scoped.

pkg/controller/v1alpha2/llmisvc/fixture/required_resources.go
```diff
@@ -59,6 +61,32 @@ func RequiredResources(ctx context.Context, c client.Client, ns string) {
 	gomega.Expect(c.Create(ctx, DefaultGatewayClass())).To(gomega.Succeed())
 }
 
+func IstioShadowService(name, ns string) *corev1.Service {
```
bartoszmajsak: This has been removed as unnecessary - you can adjust affected tests by removing a call to this fixture and they should still pass.

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -829,9 +829,51 @@ func mergeSpecs(ctx context.Context, base, override v1alpha2.LLMInferenceService
 	if err := json.Unmarshal(mergedJSON, &finalSpec); err != nil {
 		return v1alpha2.LLMInferenceServiceSpec{}, fmt.Errorf("could not unmarshal merged spec: %w", err)
 	}
+
+	// Apply rule defaults onto preset HTTPRoute rules.
+	applyHTTPRouteRuleDefaults(&base, &override, &finalSpec)
```
bartoszmajsak: Is it the right place to call it?

bartoszmajsak (changes requested):

pkg/controller/v1alpha2/llmisvc/fixture/required_resources.go
bartoszmajsak: Maybe all you need is a rebase :)

pkg/controller/v1alpha2/llmisvc/config_merge.go
pierDipi: +1, this is better placed in combineBaseRefsConfig after merge

sahilleth:
Thanks for the review — addressed both points in the latest push:

1. **Removed `IstioShadowService`** from `fixture/required_resources.go` (leftover from the rebase; it wasn’t used anywhere).

2. **Moved `ruleDefaults` application out of `mergeSpecs`** — `ApplyHTTPRouteRuleDefaults` now runs in `combineBaseRefsConfig` after the full preset chain + user spec merge (`MergeSpecs` on presets only for the base rules, then overlay user `ruleDefaults` onto the merged spec). Pairwise `mergeSpecs` stays a generic strategic merge.

Rebased onto current `master` earlier (CRD conflicts resolved); `ruleDefaults` API is in place as discussed — timeout-only `spec.rules` patches still replace preset rules via strategic merge (no special case), with `ruleDefaults` as the supported overlay path.

@bartoszmajsak @pierDipi

pkg/controller/v1alpha2/llmisvc/config_merge.go
```diff
@@ -299,6 +299,12 @@ func (r *LLMISVCReconciler) combineBaseRefsConfig(ctx context.Context, llmSvc *v
 		return nil, fmt.Errorf("failed to merge specs: %w", err)
 	}
 
+	presetSpec, err := MergeSpecs(ctx, specs...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to merge preset specs: %w", err)
+	}
```
pierDipi, line 305: why do we need to merge specs again? aren't they merged just before this line?

pierDipi:
@sahilleth would you mind rebasing please? there are conflicts

bartoszmajsak:
Friendly reminder @sahilleth. Can you rebase and resolve conflicts?

